PR #5, opened 2017-05-11
Initial work

viniciusccarvalho:
Basic infrastructure:

Still needs:

- More comprehensive readme
- Google dependencies are a bit messy, but it's something the google team could help us with
- Does not have the starters, wanted to circle back on that
- pubsub is just some experimentation around a reactive pubsub client, can be removed if we decide not to proceed


pom.xml
```diff
@@ -0,0 +1,183 @@
...
+	<profiles>
+		<profile>
+			<id>integration-test</id>
+			<activation>
+				<property>
+					<name>els.config.dir</name>
+				</property>
+			</activation>
+			<modules>
+				<module>spring-cloud-aws-integration-test</module>
```
meltsufin: Probably don't want this in there. :)

pom.xml
```diff
@@ -0,0 +1,183 @@
...
+			</modules>
+		</profile>
+		<profile>
+			<id>spring</id>
+			<repositories>
+				<repository>
+					<id>spring-snapshots</id>
+					<name>Spring Snapshots</name>
+					<url>https://repo.spring.io/libs-snapshot-local</url>
```
meltsufin: It looks like we need this repo outside of the profile to let Maven find `spring-cloud-build:1.3.1.BUILD-SNAPSHOT` parent.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/converters/support/BooleanConverter.java
```diff
@@ -0,0 +1,29 @@
+package org.springframework.cloud.gcp.pubsub.converters.support;
+
+import com.sun.tools.javac.util.Convert;
```
meltsufin: Unused import

pom.xml
markfisher: we could activate the `spring` profile by default, but this works now:

```
mvn -Pspring clean package
```

(although my current build from scratch had a test failure because of an assumption that credentials are defined)

Also, we should go ahead and add the mvn wrapper (in a separate PR would be fine).

spring-cloud-gcp-core/src/main/java/org/springframework/cloud/gcp/autoconfigure/auth/CredentialsAutoConfiguration.java
```diff
@@ -0,0 +1,40 @@
+package org.springframework.cloud.gcp.autoconfigure.auth;
```
markfisher: missing copyright header

spring-cloud-gcp-dependencies/pom.xml
```diff
@@ -0,0 +1,148 @@
...
+			</dependency>
+			<dependency>
+				<groupId>io.grpc</groupId>
+				<artifactId>grpc-auth</artifactId>
+				<version>${grpc.version}</version>
+			</dependency>
+			<dependency>
+				<groupId>io.projectreactor</groupId>
+				<artifactId>reactor-core</artifactId>
+				<version>3.0.5.RELEASE</version>
```
markfisher: latest stable is `3.0.7.RELEASE`

markfisher:
LGTM; merged along with https://github.com/spring-cloud/spring-cloud-gcp/commit/3198f76c2a4234e7acf251f65b7a34fe818e8d41

PR #6, opened 2017-05-22
Initial version of Google Cloud Pub/Sub Spring Integration channel adapters

joaoandremartins:
This is the very early stage of the inbound and outbound adapters.

There is still a lot of work required here, I'm just looking for some early validation and eyes on the code. There are also some hacks that I added just so the project would build and run.

The prototype does work on basic use cases, though its usage is still a bit cumbersome on the user side (e.g., no credentials support, clunky project ID injection, etc.)


spring-cloud-gcp-pubsub/pom.xml
```diff
@@ -42,5 +42,34 @@
 			<artifactId>spring-test</artifactId>
 			<scope>test</scope>
 		</dependency>
+		<!-- SI channel adapters -->
+		<dependency>
+			<groupId>org.springframework.integration</groupId>
+			<artifactId>spring-integration-core</artifactId>
```
artembilan: If we are sure that SI Channel Adapters should be as a part of the `spring-cloud-gcp-pubsub` module (not like `spring-cloud-gcp-integrtaion`), these dependencies for SI should be `optional`

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapter.java
```diff
@@ -0,0 +1,53 @@
...
+    this.subscriptionName = subscriptionName;
+  }
+
+  @Override
+  protected void onInit() {
+    super.onInit();
+
+    subscriber = Subscriber.defaultBuilder(SubscriptionName.create(
+        projectId, this.subscriptionName), this::receiveMessage).build();
+    subscriber.startAsync();
```
artembilan: Well, this hard connection logic should be moved to the `doStart()`.
It's not good to perform low-level resource interaction during Spring container initialization.

On the other hand with the `Lifecycle` in hands I even can decide do not connection in the beginning of application life.
Or can decide to `stop()` such a component at runtime.
More info in the Spring Reference Manual: http://docs.spring.io/spring/docs/current/spring-framework-reference/html/beans.html#beans-factory-lifecycle

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapter.java
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  private void receiveMessage(PubsubMessage message, AckReplyConsumer consumer) {
+    Map<String, Object> messageHeaders = new HashMap<>();
+    message.getAttributesMap().forEach(messageHeaders::put);
+
+    sendMessage(getMessagingTemplate().getMessageConverter().toMessage(
+        message.getData(),
+        new MessageHeaders(messageHeaders)));
+    consumer.ack();
```
artembilan: The `AckReplyConsumer` should go to message headers.
I can decide downstream in the flow when and how to `ack()` it. 
See similar logic in the `KinesisMessageDrivenChannelAdapter`: https://github.com/spring-projects/spring-integration-aws/blob/master/src/main/java/org/springframework/integration/aws/inbound/kinesis/KinesisMessageDrivenChannelAdapter.java#L852
or `KafkaMessageDrivenChannelAdapter`: https://github.com/spring-projects/spring-integration-kafka/blob/master/src/main/java/org/springframework/integration/kafka/inbound/KafkaMessageDrivenChannelAdapter.java#L304

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapter.java
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  @Override
+  protected void doStop() {
+    if (subscriber != null) {
+      // TODO(joaomartins): This doesn't seem like the best place to stop.
```
artembilan: No, it's exactly the best place to do this. See my comment in the `onInit()`

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/outbound/PubSubOutboundChannelAdapter.java
```diff
@@ -0,0 +1,32 @@
...
+
+  public PubSubOutboundChannelAdapter(String projectId, String topicName) throws IOException {
+    publisher = Publisher.defaultBuilder(TopicName.create(
+        projectId, topicName)).build();
+  }
+
+  @Override
+  protected void handleMessageInternal(Message<?> message) throws Exception {
+    if (!(message.getPayload() instanceof PubsubMessage)) {
+      throw new IllegalArgumentException("This channel adapter cannot send messages other than"
```
artembilan: That's fully dead end. If it can't it must create one.
That's why in the Desing Doc I stand out against `PubSubMessagingTemplate` idea. It doesn't provide expected flexibility and makes end-users to confuse.

See `KinesisMessageHandler` for example how the target entity can be built: https://github.com/spring-projects/spring-integration-aws/blob/master/src/main/java/org/springframework/integration/aws/outbound/KinesisMessageHandler.java

spring-cloud-gcp-pubsub/pom.xml
viniciusccarvalho: Correct, two ways we can skin this cat, either bundle all SI modules in one spring-cloug-gcp-integration or the former which is have integration as a package within each of the services (storage, pubsub,spanner). I think we would be better off with the first approach, we should make this optional (although we may end up needing it for pubsub, since most cases will be as a gateway or perhaps a binder one day)

artembilan: I don't mind to keep everything in this `spring-cloud-gcp` project. Only my concern that it might be better to keep SI Channel Adapters as a separate module, therefore separate artifact (jar) in the end. 
The name I suggest is `spring-cloud-gcp-integrtaion` because we might have adapters not only just for PubSub.
Thanks

joaoandremartins: How about if we create another `spring-cloud-gcp-integration` module, alongside the existing `spring-cloud-gcp-core`, `spring-cloud-gcp-dependencies`, etc.?
I kinda like it, although there will be a little redundancy. Users (like me) might ask why the PubSub SI channel adapters aren't in the `spring-cloud-gcp-pubsub` module. Logically, it could make more sense to have `spring-cloud-gcp-pubsub/integration`, `spring-cloud-gcp-pubsub/cloud`, etc. They could still be arranged as different modules and it seems like the most coherent arrangement to me (unless there is a case where users could use importing all SI adapters at a time to their projects, in which case having a separate `integration` module could make more sense?).

Would be great to agree on a decision here asap!

artembilan: Well, we usually don't provide nested modules and I keen like we can keep out of that here as well.

My idea is that standard Spring API over the target protocol may really be severed by the technology - storage, pubsub, spanner. But SI Channel Adapters, `LockRegistry`, `MetadataStore`, Leader Election etc., can really be combined in the separate `-integration` module. That can be considered as an optional addition if you would like to get a gain of Spring Integration in your application.

If you decide to go `integration` sub-package, you should be figure out in the future where to place those mentioned components implementations.

Hope I am clear

meltsufin: If we take the GCP-centered view, which is the current structure, then `spring-cloud-gcp-integration` doesn't make sense because all of the spring integrations would go into the respective GCP service modules like `spring-cloud-gcp-pubsub` or `spring-cloud-gcp-bigquery`. We can also take the corss-product view, which will result in modules like `spring-cloud-gcp-pubsub-integration` and `spring-cloud-gcp-biquery-integration` -- this might be too much.

artembilan: OK. I made my vote.

Can leave with the `integration` sub-package in the each module if that is the decision...

spring-cloud-gcp-pubsub/pom.xml
```diff
@@ -42,5 +42,34 @@
...
+      <version>1.0.0</version>
+    </dependency>
 	</dependencies>
+
+	<!-- Must also get rid of this -->
+	<build>
+		<plugins>
+			<plugin>
+				<groupId>org.codehaus.mojo</groupId>
+				<artifactId>animal-sniffer-maven-plugin</artifactId>
```
meltsufin: @joaoandremartins What is this for?

joaoandremartins: This plugin is being invoked by some dependency that I couldn't locate, to inspect language level. In this case, it was failing the build because it found Java 8 constructs and it was configured for language level 7.

viniciusccarvalho: This may be due usage of jdk8 as requirement on the pom. You need to add an empty .jdk8 file to the project so our build tools ignore jdk7 enforcement

viniciusccarvalho:
@joaoandremartins Please take a look at the very simple PubSubMessageTemplate. That's an example, not a final class. 

Things that needs to be take care of:

- Converters : I'd vote we use a type of `CompositeMessageConverter` with a list of pre defined message converters. We have converters for String payload such as `JSON` and `XML` that do not need the payload to be String, they can be Pojos that Jackson can marshall into json for us. We will need to add header support for content type inference as well. For binary types we can just add support for java serialization or kryo (both supported ATM by integration), those usually yield byte[] as the result of the conversion, and we can use that as an input to ByteString. I recall using that and Avro successfully in the past.

- The send method needs better exception handling, the way it's there now it was meant just to compile

- With this approach, we now have  MessageTemplate that can send to any topic and we share the executor. Another easy optimization is adding something like `String[] topics;` as a member variable and on `afterPropertiesSet` we can pre-populate the Publishers map. Would be useful for the outbound gateways that can set that since they will mostly just attach to a fixed size of topics.

Let me know if you have questions with this approach, we can move to how to handle conversion next.

artembilan:
Well, I think the best sample of the `template` pattern impl to talk about separation of concerns and more granual API is `KafkaTemplate`: https://github.com/spring-projects/spring-kafka/blob/master/spring-kafka/src/main/java/org/springframework/kafka/core/KafkaTemplate.java. Also see its converters infrastructure: https://github.com/spring-projects/spring-kafka/tree/master/spring-kafka/src/main/java/org/springframework/kafka/support/converter.

With that in hands it's not so hard to guess that the SI `MessageHandler` implementation on the matter is pretty simple: https://github.com/spring-projects/spring-integration-kafka/blob/master/src/main/java/org/springframework/integration/kafka/outbound/KafkaProducerMessageHandler.java

@viniciusccarvalho , please, don't confuse people. We don't talk here about `MessageTemplate`. Let's call things with their arranged names - `PubSubTemplate`! 

The other concern and we had some discussion in the design document is about the for the class which represents Outbound Channel Adapter. Well, that concept exists only for the SI XML config for conveninece. When we come to Java there is only Consumer Endpoint to get data from the `MessageChannel` and `MessageHandler` to accept that data. There is no such a real solution like `OutboundChannelAdapter` class. If you meet some of the `MessageHandler` implementations with such a name, that is wrong. Because the Outbound Channel Adapter principle implementation with Java can be reached only artificially or, if you wish, by convention. No one stops you to return something from your `@ServiceActivator` and send it into some channel.

What I mean with that, that we talk about some `MessageHandler` implementation which just `void` and send to the PubSub and forget. There is no need to confuse with the *outbound gateway*. That is my point.

viniciusccarvalho:
@artembilan I forget about the KaftaTemplate, indeed a very clean implementation. We really should follow, even the `sendDefault` would work nicely as well as having a `PublisherFactory`, thanks for advising

meltsufin:
Regarding merging, someone with admin on the repo can make "Squash & Merge" the default. We do it on our other repos as well.

artembilan:
Well, "Squash & Merge" is not enough in most cases. We also do `rebase` before merging to support linear commits history in the `master`:
linear_history

viniciusccarvalho:
Regarding the rebase usually the person approving the PR will take care of it.

Also @joaoandremartins I noticed your classes are missing the ASL header, please look at some of the classes int the `spring-cloud-gcp-core` module for examples.

Also please add the `@author` in the class javadoc. We usually follow firstname lastname conventions, examples on the module as well

meltsufin:
I might be missing something here, but my understanding is that "Squash & Merge" adds the squashed commit on top of the last commit in `master`, maintaining the linear commit history. You might be required to rebase against `master` or merge `master` into your branch before using "Squash & Merge", but that's enforced by GitHub.

viniciusccarvalho:
We do rebasing, I just call it squash and merge, but in this case we would do `git rebase -i HEAD~X` and then squash all commits until the latest one, and finally push this to master.

@artembilan Gary mentioned how you guys work together, I have never yet worked with a direct branch, and I do have a commit in the logs now. I honestly don't mind having it squashed along everything Joao is submitting, but how do you guys solve that?

meltsufin:
I see, you do it manually. We're talking about the [GitHub PR feature](https://help.github.com/articles/about-pull-request-merges/#squash-and-merge-your-pull-request-commits) that does it for you.

viniciusccarvalho:
@meltsufin Someone once explained to me why we don't use that feature, but I forgot and I just continue using it manually 😄

artembilan:
One more time: that "Squash & Merge" does **not** do `rebase`. 
To avoid all those conflicts we rebase and squash manually or using IDE features and only after that `merge` into master: https://github.com/spring-projects/spring-integration/wiki/Administrator-Guidelines

joaoandremartins:
Apologies, I still don't get it. If the branch isn't in sync with master, then you can't merge the PR to begin with. You must perform a manual merge or rebase, whichever flavour suits you the best, push and merge.

But the squashing part can be done automatically by GitHub. I'd prefer using this automated tool rather than doing the interactive rebase manually, it feels like a retrocess.

artembilan:
Well, our experience shows that it is much better to pull your PR locally and make a proper review and testing there. And since your branch copy is in my local repository it isn't so hard to perform a couple Git commands to have it merged to master properly.

During local review I may make some refactoring and small improvements. For example those missed Copyrights and `@author` in your new classes. Mostly we don't see reason to make a new commits for that and just do polishing during merge process. No reason to jump to GH for squashing and back to local version if you can do everything locally and have a proper commit in the `master`. Another polishing I usually do is final commit message. There is an epic article from my former colleague on the matter: https://chris.beams.io/posts/git-commit/. You can't do that from GH.

joaoandremartins:
Regarding the copyright, I noticed that it contains `Copyright 2017 original author or authors.`.

How do you feel about adding Google Inc. either here, or on a new line?

viniciusccarvalho:
I don't think we have that as a practice - Adding the company name anywhere in the license.
Which reminds me, have you signed our CLA?

artembilan:
This is Pivotal's project, so you should follow with our License agreement.

joaoandremartins:
I haven't signed the CLA, yet. Where can I do it?

We're contributing to this project as part of Google Inc. As I understand, the copyrights will always be with Google, regardless of whether the notice says that or not. It might be OK to omit Google Inc. from the notice.

viniciusccarvalho:
Oops, that's a big red flag to us. @joaoandremartins @meltsufin and any others that would like to make commits into the project, we need you guys to sign the CLA: https://github.com/spring-cloud/spring-cloud-dataflow/blob/master/spring-cloud-dataflow-docs/src/main/asciidoc/appendix-contributing.adoc 

From your comments Joao, I'd say check with your legal team, what's Google policy on allowing employees making contributions to open source projects, we don't want to get in any sort of problems later.

artembilan:
@viniciusccarvalho ,

we have to add Webhook to the [Pivotal CLA](https://cla.pivotal.io/) for this project, so no one contributor will miss CLA signing in the future. 
Not sure how so...

pivotal-issuemaster:
@joaoandremartins Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-cloud/spring-cloud-gcp&pullRequestId=6) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@joaoandremartins Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-gcp&pullRequestId=6)!

viniciusccarvalho:
@joaoandremartins Somehow my project is not compiling anymore, it states that `subscriber.stopAsync();` is not a valid method. We were based on 0.17.2 right have we moved to 0.18.0? 

You don't need to change anything, this is something I can modify on the PR review and rebase as well as formatting of the code.

I just noticed that there are tons of duplicated libraries such as guava 18 and 19. I'll see what I can do before reaching out to you guys.

viniciusccarvalho:
Close via dd44846bcf29663c28e6e7e0a494fd96ce2e6b50 and 2d28b5748cd285b5be6de373675f107ab6f2ff13

joaoandremartins:
Are you using IJ? I got the same issue there and I think it's an IJ bug, since mvn clean install builds successfully.

https://youtrack.jetbrains.com/oauth?state=%2Fissue%2FIDEA-173421

viniciusccarvalho:
It looks more like an issue on google-cloud-java, reverting to 0.17.2-alpha fixes the problem on my end. I'm simplifying the dependencies for now, but we need to discuss further is quite a messy web of libraries in google-cloud-java

meltsufin:
There is an effort to improve the dependency hell on the client lib side. If you find any specific issues, please file them on [google-cloud-java](https://github.com/GoogleCloudPlatform/google-cloud-java/issues).

Issue #10, opened 2017-05-24
Enhance the PubsubTemplate


viniciusccarvalho:
This is more of a placeholder for whomever is going to work in this story. Just make sure we have a good implementation with reasonable defaults for users to springify their pubsub experience ;)

joaoandremartins:
Would be great to eventually come down with a list of requirements for the template so we can close this off.

joaoandremartins:
In lieu of specific requirements here, I suggest we follow up on actual actionable issues and close this one out.
Thoughts?

PR #14, opened 2017-05-24
Makes PubSubMessageHandler use PubSubTemplate

joaoandremartins:
Partially addresses https://github.com/spring-cloud/spring-cloud-gcp/issues/10

### DIFF
--- a/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
+++ b/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
@@ -24,13 +24,13 @@ import com.google.api.gax.grpc.ExecutorProvider;
 import com.google.api.gax.grpc.InstantiatingExecutorProvider;
 import com.google.auth.oauth2.GoogleCredentials;
 import com.google.cloud.pubsub.spi.v1.Publisher;
-import com.google.protobuf.ByteString;
 import com.google.pubsub.v1.PubsubMessage;
 import com.google.pubsub.v1.TopicName;
 
 import org.springframework.beans.factory.InitializingBean;
 import org.springframework.cloud.gcp.pubsub.integration.converters.SimpleMessageConverter;
 import org.springframework.messaging.Message;
+import org.springframework.messaging.converter.MessageConversionException;
 import org.springframework.messaging.converter.MessageConverter;
 
 /**
@@ -38,48 +38,60 @@ import org.springframework.messaging.converter.MessageConverter;
  */
 public class PubSubTemplate implements PubSubOperations, InitializingBean {
 
-	private final String projectId;
+  private final String projectId;
 
-	private final GoogleCredentials credentials;
+  private final GoogleCredentials credentials;
 
-	private ConcurrentHashMap<String, Publisher> publishers = new ConcurrentHashMap<>();
+  private ConcurrentHashMap<String, Publisher> publishers = new ConcurrentHashMap<>();
 
-	private ExecutorProvider executorProvider;
+  private ExecutorProvider executorProvider;
 
-	private MessageConverter converter;
+  private MessageConverter messageConverter;
 
-	private int concurrentProducers = 1;
+  private int concurrentProducers = 1;
 
-	public PubSubTemplate(GoogleCredentials credentials, String projectId){
-		this.projectId = projectId;
-		this.credentials = credentials;
-		this.executorProvider = InstantiatingExecutorProvider.newBuilder().setExecutorThreadCount(concurrentProducers).build();
-		this.converter = new SimpleMessageConverter();
-	}
+  public PubSubTemplate(GoogleCredentials credentials, String projectId) {
+    this(credentials, projectId, new SimpleMessageConverter());
+  }
 
-	@Override
-	public String send(final String topic, Message message) throws RuntimeException{
+  public PubSubTemplate(GoogleCredentials credentials, String projectId,
+      MessageConverter messageConverter) {
+    this.projectId = projectId;
+    this.credentials = credentials;
+    this.executorProvider = InstantiatingExecutorProvider.newBuilder()
+        .setExecutorThreadCount(concurrentProducers).build();
+    this.messageConverter = messageConverter;
+  }
 
-		Publisher publisher = publishers.computeIfAbsent(topic, s -> {
-			try {
-				return Publisher.defaultBuilder(TopicName.create(projectId,topic)).setExecutorProvider(executorProvider).build();
-			}
-			catch (IOException e) {
-				e.printStackTrace();
-			}
-			return null;
-		});
+  @Override
+  public String send(final String topic, Message message) throws RuntimeException {
 
-		try {
-			return publisher.publish(PubsubMessage.newBuilder().setData(ByteString.copyFrom(message.getPayload().toString().getBytes())).build()).get();
-		}
-		catch (Exception e) {
-			throw new RuntimeException(e);
-		}
-	}
+    Publisher publisher = publishers.computeIfAbsent(topic, s -> {
+      try {
+        return Publisher.defaultBuilder(TopicName.create(projectId, topic))
+            .setExecutorProvider(executorProvider).build();
+      } catch (IOException e) {
+        e.printStackTrace();
+      }
+      return null;
+    });
 
-	@Override
-	public void afterPropertiesSet() throws Exception {
+    Object pubsubMessageObject = messageConverter.fromMessage(message, PubsubMessage.class);
 
-	}
+    if (!(pubsubMessageObject instanceof PubsubMessage)) {
+      throw new MessageConversionException("The specified converter must produce"
+          + "PubsubMessages to send to Google Cloud Pub/Sub.");
+    }
+
+    try {
+      return publisher.publish((PubsubMessage) pubsubMessageObject).get();
+    } catch (Exception e) {
+      throw new RuntimeException(e);
+    }
+  }
+
+  @Override
+  public void afterPropertiesSet() throws Exception {
+
+  }
 }
--- a/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/outbound/PubSubMessageHandler.java
+++ b/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/outbound/PubSubMessageHandler.java
@@ -16,47 +16,28 @@
 
 package org.springframework.cloud.gcp.pubsub.integration.outbound;
 
-import com.google.cloud.pubsub.spi.v1.Publisher;
-import com.google.pubsub.v1.PubsubMessage;
-import com.google.pubsub.v1.TopicName;
 import java.io.IOException;
-import org.springframework.cloud.gcp.pubsub.integration.converters.SimpleMessageConverter;
+import org.springframework.cloud.gcp.pubsub.core.PubSubTemplate;
 import org.springframework.integration.handler.AbstractMessageHandler;
 import org.springframework.messaging.Message;
-import org.springframework.messaging.converter.MessageConversionException;
-import org.springframework.messaging.converter.MessageConverter;
 
 /**
- * Converts from internal Spring message to GCP Pub/Sub message and publishes the Pub/Sub
- * message to a Pub/Sub topic.
+ * Sends messages to Google Cloud Pub/Sub by delegating to {@link PubSubTemplate}.
  *
  * @author João André Martins
  */
 public class PubSubMessageHandler extends AbstractMessageHandler {
 
-  private Publisher publisher;
-  private MessageConverter messageConverter;
+  private PubSubTemplate pubsubTemplate;
+  private String topic;
 
-  public PubSubMessageHandler(String projectId, String topicName) throws IOException {
-    this(projectId, topicName, new SimpleMessageConverter());
-  }
-
-  public PubSubMessageHandler(String projectId, String topicName,
-      MessageConverter messageConverter) throws IOException {
-    publisher = Publisher.defaultBuilder(TopicName.create(
-        projectId, topicName)).build();
-    this.messageConverter = messageConverter;
+  public PubSubMessageHandler(String projectId, String topic) throws IOException {
+    pubsubTemplate = new PubSubTemplate(null, projectId);
+    this.topic = topic;
   }
 
   @Override
   protected void handleMessageInternal(Message<?> message) throws Exception {
-    Object pubsubMessageObject = messageConverter.fromMessage(message, PubsubMessage.class);
-
-    if (!(pubsubMessageObject instanceof PubsubMessage)) {
-      throw new MessageConversionException("The specified converter must produce"
-          + "PubsubMessages to send to Google Cloud Pub/Sub.");
-    }
-
-    publisher.publish((PubsubMessage) pubsubMessageObject);
+    pubsubTemplate.send(topic, message);
   }
 }


spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
```diff
@@ -24,62 +24,74 @@
...
+import org.springframework.messaging.converter.MessageConversionException;
 import org.springframework.messaging.converter.MessageConverter;
 
 /**
  * @author Vinicius Carvalho
  */
 public class PubSubTemplate implements PubSubOperations, InitializingBean {
 
-	private final String projectId;
+  private final String projectId;
```
artembilan: We use **tab** for indents.
Otherwise it's going to provide a lot of mess and inconvenience during incorporation.

@viniciusccarvalho , WDYT?

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
```diff
@@ -24,62 +24,74 @@
...
-		this.converter = new SimpleMessageConverter();
-	}
+  public PubSubTemplate(GoogleCredentials credentials, String projectId) {
+    this(credentials, projectId, new SimpleMessageConverter());
+  }
 
-	@Override
-	public String send(final String topic, Message message) throws RuntimeException{
+  public PubSubTemplate(GoogleCredentials credentials, String projectId,
+      MessageConverter messageConverter) {
```
artembilan: I'd say `messageConverter` should be optional and supplied by default with something. 
Therefore not a ctor arg.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
```diff
@@ -24,62 +24,74 @@
...
-		try {
-			return publisher.publish(PubsubMessage.newBuilder().setData(ByteString.copyFrom(message.getPayload().toString().getBytes())).build()).get();
-		}
-		catch (Exception e) {
-			throw new RuntimeException(e);
-		}
-	}
+    Publisher publisher = publishers.computeIfAbsent(topic, s -> {
+      try {
+        return Publisher.defaultBuilder(TopicName.create(projectId, topic))
```
artembilan: What is this `projectId`?
According this code I'd say that `projectId` should be carried together with the `topic` from the client.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
viniciusccarvalho: I'm with you, I just don't try to jump 4 steps of a staircase at each time like Richard Hendricks have once proved to us :)

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/outbound/PubSubMessageHandler.java
```diff
@@ -16,47 +16,28 @@
...
-  }
-
-  public PubSubMessageHandler(String projectId, String topicName,
-      MessageConverter messageConverter) throws IOException {
-    publisher = Publisher.defaultBuilder(TopicName.create(
-        projectId, topicName)).build();
-    this.messageConverter = messageConverter;
+  public PubSubMessageHandler(String projectId, String topic) throws IOException {
+    pubsubTemplate = new PubSubTemplate(null, projectId);
+    this.topic = topic;
```
artembilan: Well, we really have to consider scenarios to resolve at least `topic` against incoming `message`. 
That is where `KafkaProducerMessageHandler` comes to the rescue as a sample with all its Expression goodies: https://github.com/spring-projects/spring-integration-kafka/blob/master/src/main/java/org/springframework/integration/kafka/outbound/KafkaProducerMessageHandler.java

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
meltsufin: I think it's auto-format at work. We should follow Spring's style, which is tabs.

viniciusccarvalho:
Which reminds me @artembilan , that I need to add a **checkstyle** file to the project

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
viniciusccarvalho: @artembilan so ProjectId is google's interpretation for multi tenancy (if I can put in this terms). Think of it like vhosts in rabbitmq. This is why I think it should be at the constructor level, whereas the topic could be per invocation. Each project id can (and probably should) have several topics associated within it. Does it make sense?

artembilan: Thanks, @viniciusccarvalho . I'm fine with the analogy with `vhost` or connection URL analogy, but looks like it is similar to AWS Stack:
https://cloud.google.com/resource-manager/docs/creating-managing-projects
http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/stacks.html

So, anyway I'm good with the ctor arg, therefore `PubSubTemplate` (`PubSubMessageHandler`) per project.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
joaoandremartins: It is optional, you can use the first constructor if you don't want to specify a converter.
I didn't specify a setter so that the object is immutable.

viniciusccarvalho: @joaoandremartins please hold any changes, I'm modifying all the directories, modules, and moving a lot of things around, it will probably cause you a lot of pain when merging changes

viniciusccarvalho: And regarding the comments, I'd also rather have it as a setter if it's not required. As far as I remember having a single constructor makes autowiring automatic for spring beans as well

joaoandremartins:
> I'm not sure that it would be so easy to work this way with the PR for PRs...
> We prefer just only additional commits to existing PRs.

@artembilan what did you mean?

artembilan:
I just described what you did, @joaoandremartins .

You had a PR against `master` and then you have opened one more against that one.
I'm pretty sure that it wouldn't be possible if the fist PR wasn't from the `upstream`, but from your `origin`.

I understand that it is possible by the Git nature, but I don't have such an experience, so that is a bit confusing and makes things much complicated. Just IMO, BTW...

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/outbound/PubSubMessageHandler.java
joaoandremartins: Apologies, but I'm not very clear on the requirement.
Do you mean that the adapter should (also) get the topic from the message, instead of being bound to a single topic?

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
joaoandremartins: @viniciusccarvalho did you close this PR so we can resume after you've moved things around?

joaoandremartins:
@artembilan I think I see what you mean now.

I'm used to rebasing the second PR in case the first one (and its branch) doesn't exist anymore. I prefer to split work among several PRs so that the PRs don't become too hard to read. I'll see what I can do in the future.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/outbound/PubSubMessageHandler.java
artembilan: Right, that's the point. Imagine we have gazillions of messages and somehow thousands of topics.
So, do we need to configure such an amount of channel adapters for them?

Yes, that is possible as well, like for Binder use-case, but we really should be flexible and let end-user to determine the target destination on his/her message. The simple sample is `gcp_topic` header populated to the message before sending to this `PubSubMessageHandler`

artembilan:
Well, we just add more commits to the PR.
But I see you point, too.
I think I'll get used to work your way as well.

viniciusccarvalho:
I did not close the PR, I guess git did when we closed the first one (don't ask me why)

Anyways you'd be in a hell of trouble for merging. I would honestly stash the work, get the new folder strucuture, replace the contents of the MessageHandler if I were you

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/outbound/PubSubMessageHandler.java
viniciusccarvalho: My 2c @joaoandremartins is to not confuse the PubSubTemplate with the Handler. What @artembilan is describing is a a very known pattern in SI. Topics would be in the header, now, the Handler has the decision what to do with that topic name, and in our case it would just pass it to the `send(topic, PubSubMessage)` inside the PubsubTemplate, right?

The internals of the PubSubTemplate will need to handle how create publishers and cache them, but this is a different story.

artembilan: That's correct, @viniciusccarvalho .

I have just pointed to that as a future improvement. 
We can live with the current state for some time any way.

meltsufin: +1 on having `PubSubTemplate.send(topic, PubSubMessage)`

joaoandremartins: Do you mean `PubSubTemplate.send(topic, Message)` as we have today?

Regarding the earlier suggestion, one thing I fail to understand is at what point we add this header to the message. From the user's point of view, they just do `messagingGateway.send(String)`, which sends the message to a `MessageChannel`, which is intercepted by `PubSubMessageHandler`. We only control the last bit, `PubSubMessageHandler`, unless there's any magic we can do at the `MessagingGateway` level where we can add the name of the `defaultRequestChannel` as the topic header.

artembilan: The `PubSubTemplate.send()` should be extended to more APIs. See `KafkaOperations` for example: https://github.com/spring-projects/spring-kafka/blob/master/spring-kafka/src/main/java/org/springframework/kafka/core/KafkaOperations.java#L92.

Yes, we really can have there `PubSubMessage`-based variant as well.

Forget a bit about that `messagingGateway`. Even `@Controller` can be replaced by the Spring Integration HTTP Inbound Channel Adapter: http://docs.spring.io/spring-integration/reference/html/http.html#http-inbound

Also consider the case:
we poll data from the RDBMS and each record has its' own option about the target PubSub topic. We just store that info in the header after splitting the list of records.
Another situation that we may determine the target topic depending on the system state or some other external influence. That's how the SpEL expression (or Java `Function`) against each message comes to the rescue.

Again. Kafka is the best sample for us here: https://github.com/spring-projects/spring-integration-kafka/blob/master/src/main/java/org/springframework/integration/kafka/outbound/KafkaProducerMessageHandler.java#L147

viniciusccarvalho: @joaoandremartins I think it may prove to be more useful if we have a screenshare one day to go over this. The base to grasp is that the integration outbound/inbound channels should only deal with Message from SI. I think this is where you are having some trouble understanding. In SI land Message and Channels are the key constructs that allows components to form a flow.

The PubsubTemplate is just a client to PubSub with some added spring goodness, that the outbound (not so sure about inbound) happens to use.

So think about the oubound part now. You can have a poller like Artem mentioned that polls records from the DB, those are transformed into Message (the currency of SI) and sent to channels, you maybe perhaps have a `header-enricher` somewhere down your flow that adds a header `TOPIC_NAME` to the message. And this message lands on our outbound. The Outbound channel now needs to convert from Message into whatever the contract PubSubTemplate offers (could as well be PubSubMessage) and it then dispatches it via the send method.

The other way is having something like a MessageListenerContainer (as Artem pointed, not needed actually) but an EventDriven source that receives messages from PubSub and then translates them into Message from SI and forwards to the output channel (which would connect the flow to downstream consumers)

joaoandremartins: Thanks a lot for the explanation, it's a lot clearer now.

From my point of view, I think it would be great to focus on landing the work on everything SI needs to operate right now, and file non-SI use cases as issue that we could address later.

joaoandremartins: I guess my earlier point was also that, while it is possible to have a composite flow in SI like you guys described, with hooked `header-enrichers`, it's also possible to have a simpler flow where we can never set the headers (or it would be to cumbersome for users to do so).

For example, the following is a valid use case. It sounds like the SI adapter could support both cases simultaneously? Then we'd need to come with some sort of order for topic resolution, e.g., the topic passed in the header would have precedence over the one in the ctor/setter, or vice-versa.

```
  @Bean
  public MessageHandler messageSender(@Qualifier("gcpProjectId") String projectId)
      throws IOException {
    return new PubSubMessageHandler(projectId, "test");
  }

  @Bean
  public MessageChannel pubsubOutputChannel() {
    return new PublishSubscribeChannel();
  }

  @MessagingGateway(defaultRequestChannel = "pubsubOutputChannel")
  public interface PubsubOutboundGateway {
    void sendToPubsub(String text);
  }
```

artembilan: We can't assume how complex is end-user flow and can't guess what is the source of metadata for incoming data. But any way your component (`PubSubMessageHandler`) should be ready for that and provide some flexibility and choice for end-user. That's why I always show that `KafkaProducerMessageHandler` as a sample with its flexible expressions. The same I can say about `KinesisMessageHandler`: https://github.com/spring-projects/spring-integration-aws/blob/master/src/main/java/org/springframework/integration/aws/outbound/KinesisMessageHandler.java#L93, where we have the logic like:
```
	String stream = message.getHeaders().get(AwsHeaders.STREAM, String.class);
		if (!StringUtils.hasText(stream) && this.streamExpression != null) {
			stream = this.streamExpression.getValue(this.evaluationContext, message, String.class);
		}
		Assert.state(stream != null, "'stream' must not be null for sending a Kinesis record. " +
				"Consider configuring this handler with a 'stream'( or 'streamExpression') or supply an " +
				"'aws_stream' message header.");

```

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
joaoandremartins: I made this change in https://github.com/viniciusccarvalho/spring-cloud-gcp/pull/2

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/outbound/PubSubMessageHandler.java
joaoandremartins: Filed this as an [issue](https://github.com/spring-cloud/spring-cloud-gcp/issues/19).

PR #27, opened 2017-05-30
Makes PubSubMessageHandler use PubSubTemplate

joaoandremartins:


spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
```diff
@@ -72,15 +78,18 @@ public String send(final String topic, Message message) throws RuntimeException
...
-		try {
-			return publisher.publish(PubsubMessage.newBuilder()
-					.setData(ByteString
-							.copyFrom(message.getPayload().toString().getBytes()))
-					.build()).get();
-		}
-		catch (Exception e) {
-			throw new RuntimeException(e);
-		}
+    Object pubsubMessageObject = messageConverter.fromMessage(message, PubsubMessage.class);
```
meltsufin: Spaces -> tabs

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
```diff
@@ -46,16 +47,21 @@
 
 	private ExecutorProvider executorProvider;
 
-	private MessageConverter converter;
+	private MessageConverter messageConverter;
 
 	private int concurrentProducers = 1;
 
-	public PubSubTemplate(GoogleCredentials credentials, String projectId) {
+  public PubSubTemplate(GoogleCredentials credentials, String projectId) {
```
meltsufin: Spaces -> tabs

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
```diff
@@ -46,16 +47,21 @@
...
+  }
+
+	public PubSubTemplate(GoogleCredentials credentials, String projectId,
+      MessageConverter messageConverter) {
 		this.projectId = projectId;
 		this.credentials = credentials;
 		this.executorProvider = InstantiatingExecutorProvider.newBuilder()
 				.setExecutorThreadCount(concurrentProducers).build();
-		this.converter = new SimpleMessageConverter();
+		this.messageConverter = new SimpleMessageConverter();
```
joaoandremartins: This is a mistake, probably bad merge.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
```diff
@@ -46,16 +46,21 @@
 
 	private ExecutorProvider executorProvider;
 
-	private MessageConverter converter;
+	private MessageConverter messageConverter;
 
 	private int concurrentProducers = 1;
 
 	public PubSubTemplate(GoogleCredentials credentials, String projectId) {
```
meltsufin: `projectId` should be optional

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
```diff
@@ -65,20 +70,22 @@ public String send(final String topic, Message message) throws RuntimeException
 			try {
 				return Publisher.defaultBuilder(TopicName.create(projectId, topic))
```
meltsufin: Use `Topic createTopic(TopicName name)`, if `projectId` is not set.

meltsufin: Note, that the client will still have to know the project id to make it part of the `name` specification. However, hopefully that will change in the future after [this issue](https://github.com/GoogleCloudPlatform/google-cloud-java/issues/2094) is fixed..

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
```diff
@@ -65,20 +70,22 @@ public String send(final String topic, Message message) throws RuntimeException
...
+		Object pubsubMessageObject = messageConverter.fromMessage(message, PubsubMessage.class);
+
+		if (!(pubsubMessageObject instanceof PubsubMessage)) {
+			throw new MessageConversionException("The specified converter must produce"
+					+ "PubsubMessages to send to Google Cloud Pub/Sub.");
 		}
-		catch (Exception e) {
+
+		try {
+			return publisher.publish((PubsubMessage) pubsubMessageObject).get();
```
artembilan: What is this `.get()` about?
If we talk about something like `Future<?>`, then let's return that exactly!
Or even better `ListenableFuture`.
The end-user will make decision to block the result or handle asynchronously.

The good sample for that is the same `KafkaTemplate.send()`

viniciusccarvalho:
@artembilan PubSubTemplate is a stub class, nothing there should be taken seriously. It was just to show how to separate the client (pubsub publisher) form message handlers and endpoints. 

What we will need is a Future Adapter, as pubsub returns something called ApiFuture which is based out of guava. I think we should adapt that to a ListanableFuture, like you said, probably a settablefuture via a callback registered at the ApiFuture.

artembilan:
> probably a settablefuture via a callback registered at the ApiFuture.

Sounds like a plan a really can be done in the separate issue.
Nevertheless this `PubSubTemplate` should evolve any way...

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
viniciusccarvalho: Why has the import order changed? That's not what our checkstyle format supports. Here's our current model. I know we are missing the checkstyle file, but that's not the right import order.

java.*
<blank line>
javax.*
<blank line>
all other imports
<blank line>
org.springframework.*
<blank line>
all other static imports

artembilan: No asterisk imports at all.
No `static` imports in `main` classes.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
joaoandremartins: I'm pretty sure I had addressed that already, a commit must have been lost when I rebased onto master, after your rearrangement PR. Sorry about that!

joaoandremartins:
Regarding `send()`'s return, I suggest to change the method signature to ListenableFuture<ApiFuture<String>> in this PR and implement the ApiFuture<String> -> String adapter in another PR. Is this suitable?

artembilan:
`ListenableFuture<ApiFuture>` doesn't sounds good, so I'd prefer do nothing here on the matter and just deffer the feature to another issue.

viniciusccarvalho:
I second that, either make a change to a correct Future Implementation, or just leave it be for now, and address on another PR (I'd prefer this approach, otherwise this PR will drag on)

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
joaoandremartins: Regarding message conversion chaining, I'm not so sure how that would work. All the examples I've seen use a single converter, so I suggest to just add a setter/getter in this PR and add chaining later?

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
```diff
@@ -65,20 +64,22 @@ public String send(final String topic, Message message) throws RuntimeException
...
-		try {
-			return publisher.publish(PubsubMessage.newBuilder()
-					.setData(ByteString
-							.copyFrom(message.getPayload().toString().getBytes()))
-					.build()).get();
+		Object pubsubMessageObject = messageConverter.fromMessage(message, PubsubMessage.class);
+
+		if (!(pubsubMessageObject instanceof PubsubMessage)) {
+			throw new MessageConversionException("The specified converter must produce"
+					+ "PubsubMessages to send to Google Cloud Pub/Sub.");
```
artembilan: There is no space between `produce` and `PubsubMessages` words

spring-integration-gcp/src/main/java/org/springframework/integration/gcp/outbound/PubSubMessageHandler.java
```diff
@@ -17,50 +17,29 @@
...
  */
 public class PubSubMessageHandler extends AbstractMessageHandler {
 
-	private Publisher publisher;
-	private MessageConverter messageConverter;
-
-	public PubSubMessageHandler(String projectId, String topicName) throws IOException {
-		this(projectId, topicName, new SimpleMessageConverter());
-	}
+	private PubSubTemplate pubsubTemplate;
```
artembilan: `final`

spring-integration-gcp/src/main/java/org/springframework/integration/gcp/outbound/PubSubMessageHandler.java
```diff
@@ -17,50 +17,29 @@
...
+	private String topic;
 
-	public PubSubMessageHandler(String projectId, String topicName,
-			MessageConverter messageConverter) throws IOException {
-		publisher = Publisher.defaultBuilder(TopicName.create(projectId, topicName))
-				.build();
-		this.messageConverter = messageConverter;
+	public PubSubMessageHandler(String projectId, String topic, GoogleCredentials credentials) {
+		pubsubTemplate = new PubSubTemplate(credentials, projectId);
+		this.topic = topic;
```
artembilan: topic must go to the setter since we are going to add expression resolution based on the `requestMessage` context.

spring-integration-gcp/src/main/java/org/springframework/integration/gcp/outbound/PubSubMessageHandler.java
```diff
@@ -17,50 +17,29 @@
...
-		Object pubsubMessageObject = messageConverter.fromMessage(message,
-				PubsubMessage.class);
-
-		if (!(pubsubMessageObject instanceof PubsubMessage)) {
-			throw new MessageConversionException("The specified converter must produce"
-					+ "PubsubMessages to send to Google Cloud Pub/Sub.");
-		}
-
-		publisher.publish((PubsubMessage) pubsubMessageObject);
+		pubsubTemplate.send(topic, message);
```
artembilan: `this.` is missed.

joaoandremartins: So if I understand correctly, `this` should always be used to refer to the class members, as per your style?

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
joaoandremartins: This PR is about to get merged. @viniciusccarvalho since you have the most context on what to do here, would you mind filing an issue to make sure this is addressed? Thanks!

spring-integration-gcp/src/main/java/org/springframework/integration/gcp/outbound/PubSubMessageHandler.java
artembilan: Yes, `this.` should be used for properties of the class, but not for methods.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
joaoandremartins: I just realised you suggested a CompositeMessageConverter [here](https://github.com/spring-cloud/spring-cloud-gcp/pull/6#issuecomment-303567611), I'll be sure to have a look at this soon!

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
artembilan: Any my latest concerns have not been addressed yet.
Thanks

joaoandremartins: Sorry, I thought you wanted to do those changes on merge. I'll do them right away.

artembilan: Oh! Sorry I really told that yesterday. Late work in the evening eliminates any amount of coffee in the morning 😄 

So, I'm waiting for your commit and rebase and merge.
Thanks

spring-integration-gcp/src/main/java/org/springframework/integration/gcp/outbound/PubSubMessageHandler.java
joaoandremartins: I noticed that in the Kafka world, you let the Kafka producer fail due to the lack of a topic. So I'm doing the same here, and not checking for the existence of a topic before `send()`.

spring-integration-gcp/src/main/java/org/springframework/integration/gcp/outbound/PubSubMessageHandler.java
joaoandremartins: Just for reference, the checkstyle file isn't enforcing `this` on property references, I wonder if it supports this.

joaoandremartins:
@artembilan changes applied. Feel free to merge any time!

artembilan:
Merged as https://github.com/spring-cloud/spring-cloud-gcp/commit/03d95f87300b32e75a487655d7f1937acfd3c8c8

PR #28, opened 2017-05-30
Adds GcpProperties to get GCP configuration

### LINKED ISSUE #13: Create project ID and GcpProject configuration provider
On a Spring Boot scenario, it would be nice to have things like the GCP project ID injected as a `String` bean, as well as a potential `GcpProject` or even `GcpConfiguration` `@ConfigurationProperties` object.

This would be useful even in non Spring Boot scenarios, like SI or even plain Spring Framework.

joaoandremartins:
This would enable the following user experience:

```
  @Bean
  public MessageHandler messageSender(@Qualifier("gcpProjectId") String projectId)
      throws IOException {
    return new PubSubMessageHandler(projectId, "test");
  }
```

However, it might be lacking stuff such as @conditionalonclass, we might want to rename GcpProject to GcpConfiguration, etc.

Fixes #13


spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/src/main/java/org/springframework/cloud/gcp/core/autoconfig/GcpProject.java
```diff
@@ -0,0 +1,20 @@
+package org.springframework.cloud.gcp.core.autoconfig;
```
meltsufin: Copyright header missing

spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/src/main/java/org/springframework/cloud/gcp/core/autoconfig/GcpProject.java
```diff
@@ -0,0 +1,20 @@
...
+
```
meltsufin: Copyright header missing

spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/src/main/java/org/springframework/cloud/gcp/core/autoconfig/GcpProjectAutoConfiguration.java
```diff
@@ -0,0 +1,50 @@
+package org.springframework.cloud.gcp.core.autoconfig;
```
meltsufin: Copyright header missing

joaoandremartins:
@viniciusccarvalho the issue with injecting a `GcpProperties` object and then extracting the project ID from it is that we can't get the project ID from the credentials or a metadata server in the fallback, since that's not supported by `@ConfigurationProperties`.

Also, I think we had discussed in the design document that we should try to get credentials from the environment variable before the property. [@ConfigurationProperties gets it the other way around](https://docs.spring.io/spring-boot/docs/current/reference/html/boot-features-external-config.html).

This PR enables configuring the project ID in both ways. We can put off the String injection until we actually consider those two use cases if you prefer -- they're just TODOs in the code now.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/src/main/java/org/springframework/cloud/gcp/core/autoconfig/GcpProjectAutoConfiguration.java
```diff
@@ -0,0 +1,50 @@
...
+ * @author João André Martins
+ */
+@Configuration
+@EnableConfigurationProperties(GcpProject.class)
+public class GcpProjectAutoConfiguration {
+
+  private static final Logger LOGGER =
+      Logger.getLogger(GcpProjectAutoConfiguration.class.getName());
+
+  private static final String PROJECT_ID_ENV_VAR_NAME = "GCP_PROJECT_ID";
```
meltsufin: Let's use `GOOGLE_CLOUD_PROJECT` as done by the [client libraries](https://github.com/GoogleCloudPlatform/google-cloud-java/#specifying-a-project-id).

viniciusccarvalho:
This whole approach does not look right.

We are not getting from the credentials, but we should not have one class for the project that is for sure. For one property we have an entire mapping class?

there should be a GCPProperties for common stuff, one is auth location (if present) the other is projectID (also if present)

Those can be then used later to configure the appropriate beans.

We do not need an autoconfiguration to configure a projectId, this is overkill

meltsufin:
@viniciusccarvalho I think @joaoandremartins intended `GcpProject` to include more than the `id` in the future -- equivalent to what I think you're getting at with `GCPProperties`.

viniciusccarvalho:
What else there can be in a project. I think we should have project as a property on GCPProperties not having it as a first class citizen

joaoandremartins:
It seems like we can do away with just renaming GcpProject to GcpProperties and adding a credentials location, as I had previously suggested.
I'll go ahead and make that change.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/src/main/java/org/springframework/cloud/gcp/core/autoconfig/GcpProjectAutoConfiguration.java
```diff
@@ -0,0 +1,50 @@
...
+
+  private GcpProject gcpProjectFromProperties;
+
+  @Autowired
+  public GcpProjectAutoConfiguration(GcpProject gcpProjectFromProperties) {
+    this.gcpProjectFromProperties = gcpProjectFromProperties;
+  }
+
+  @Bean
+  public String gcpProjectId() {
```
meltsufin: You can also get project ID from `ServiceOptions.getDefaultProjectId()`. I think it will check the environment variable for you.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/src/main/java/org/springframework/cloud/gcp/core/autoconfig/GcpProperties.java
```diff
@@ -0,0 +1,46 @@
...
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+/**
+ * @author João André Martins
+ */
+@ConfigurationProperties("cloud.gcp")
+public class GcpProperties {
+
+	private String projectId;
```
joaoandremartins: Ideally, we'd have cloud.gcp.project.id instead of cloud.gcp.projectId. Not sure if there's a way to make this happen. Maybe nested configurations or a Project/Credentials object?

spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/src/main/java/org/springframework/cloud/gcp/core/autoconfig/GcpPropertiesAutoConfiguration.java
```diff
@@ -0,0 +1,28 @@
...
+
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.context.annotation.Configuration;
+
+/**
+ * @author João André Martins
+ */
+@Configuration
+@EnableConfigurationProperties(GcpProperties.class)
+public class GcpPropertiesAutoConfiguration {
```
joaoandremartins: I'm assuming we want `@Configuration`/`@EnableConfigurationProperties` in a separate class.

joaoandremartins:
I pushed another commit that defers getting the project ID from credentials/metadata server to another time.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/src/main/java/org/springframework/cloud/gcp/core/autoconfig/GcpProperties.java
artembilan: For that purpose you should have a structure like:
```
@ConfigurationProperties("cloud.gcp")
public class GcpProperties {
     
     private Project project = new Project();

     public static class Project {
            private String id;
     }
}
```

The good simple sample is `SecurityProperties`: https://github.com/spring-projects/spring-boot/blob/master/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/security/SecurityProperties.java#L83

viniciusccarvalho: I don't see a reason for that, unless project is a POJO which I hardly believe it will be. Let's think a little bit about what gcpproperties should have:
- Project ID
- Region ?
- Auth file

Those are perhaps the common things across all projects we may need. Adding a new dot would only make boot's relaxed binding even more confusing

spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/src/main/java/org/springframework/cloud/gcp/core/autoconfig/GcpPropertiesAutoConfiguration.java
viniciusccarvalho: Nope....

I think it will be more productive accepting this, and fixing it to show what we meant by properties and configuration.

I think the auto configuration concept is not clear to you it. You don't need one auto configuration file for every properties file, or one autoconfiguration file for every feature.

I guess the best course is to rename the CredentialsAutoConfiguration to be something like GoogleCloudContextAutoConfiguration.

I'll take this on another PR, and clean a bit the project

spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/src/main/java/org/springframework/cloud/gcp/core/autoconfig/GcpProperties.java
joaoandremartins: I see a few reasons, but none of them very strong. 1) We might be interested in adding things to a project in the future, and then project.property would be recommended, which would be a configuration breaking change, 2) project.id looks better than projectId, 3) there are indeed other things in a project, some of them could be useful for other "legacy" APIs like the project number (but none that's critical besides ID). To have an idea of what a GCP project can have, you can refer to [this page](https://cloud.google.com/resource-manager/reference/rest/v1beta1/projects). There are things like the IAM policy that we might want to use from Spring and that it's its own separate `Policy` entity in the GCP world, but would belong to a `Project` in Spring configuration.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/src/main/java/org/springframework/cloud/gcp/core/autoconfig/GcpPropertiesAutoConfiguration.java
joaoandremartins: Wouldn't `@ConditionalOnClass(GoogleCredentials.class)` be on our way? We want to enable `GcpProperties` regardless of whether `GoogleCredentials` is in the classpath, no?

spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/src/main/java/org/springframework/cloud/gcp/core/autoconfig/GcpProperties.java
artembilan: That is something not related to the `@ConfigurationProperties`.
That is what can be obtained from the GCP Client by the `projectId`: https://cloud.google.com/resource-manager/reference/rest/v1beta1/projects/get

What is the point to have IAM Policy as a read only configuration property?
We don't manage Project from arbitrary Spring Boot application.
If you need that logic, it is already a part of your target application, not Spring Boot auto-configuration.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/src/main/java/org/springframework/cloud/gcp/core/autoconfig/GcpPropertiesAutoConfiguration.java
viniciusccarvalho: AFAIK GoogleCredentials is required for anything we do. Just like on AWS you need your access_id and secrets. I don't think there's anything that should not depend on it.

Like I said before, what's need to be done is a bit more elaboration on the gcp-core to look more like a context/core project that provide such things.

viniciusccarvalho:
@joaoandremartins I've said we can have the renaming as a separate issue, if we continue to keep pushing commits and modifying new classes the review will go on forever, we are already at an absurd 23 messages for a simple commit. I'm closing this PR

joaoandremartins:
Forgot to merge?

viniciusccarvalho:
No @joaoandremartins  this PR was rejected due the number changes, there's a new PR already approved by Artem at #32

joaoandremartins:
It seems like this could've been merged and you could've added your refactoring on top. In any case, let's proceed with #32.

viniciusccarvalho:
Not after so many changes, in any case, when we merge and need to make changes, the process is usually a bit more complex. It was going to be way easier to just restart this

joaoandremartins:
Just so I'm clear in the future and this doesn't happen again, what impact does the number of changes have? Would having rebased into a single commit helped?

The only real added value of #32 over this PR is the change to use `GcpProperties` instead of `Environment`. I thought of making that change as a separate issue, can't see great complexity in this.

It would be great for us to converge on the same process so we're all in the same page and this doesn't happen again.

PR #30, opened 2017-05-30
Renames CONSUMER -> ACKNOWLEDGEMENT for more accurate naming

joaoandremartins:


joaoandremartins:
@artembilan changed the PR title, which will be the commit name if Squash commits is selected.

artembilan:
Well, that's not,
 @joaoandremartins 

squash

More over the headline of the squashed commit must not exceed 50 symbols.

Not a big deal, but let's work out good habits at least to avoid similar discussions.

I have shared not one time the reasons why I insist to make commit messages properly.

So, right now I would like to hear from you: 

what is the reason do not make  commit messages more informative?

Thanks

joaoandremartins:
@artembilan I'll keep that in mind. Usually, I find details like that in the PR description itself. This was a very basic renaming, so I didn't think more detail was necessary.

When you try to squash and commit, is the first commit text selected instead of the title? Now it's clearer why you always manually squash and commit before merging in GitHub.

artembilan:
Right, as you see by that screenshot, the commit message has been selected for squash.

Well, when you scan commits locally in Git or in the IDE, there is no any links to PR. That's how more info in the commit helps to understand the progress.

Merging though with your "squash" suggestion. Let's see what will happen!

PR #29, opened 2017-05-30
Pass in credentials to PubSub's Publisher and Subscriber.

joaoandremartins:
And possibly makes use of CredentialsAutoConfiguration.

Fixes #31

PR #32, opened 2017-05-30
Renames core classes

viniciusccarvalho:
- Adding new Property binding class
 - Renaming base autoconfiguration class to reflect a more general
   context configuration, not only credentials

### DIFF
--- a/spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/src/main/java/org/springframework/cloud/gcp/core/GCPProperties.java
+++ b/spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/src/main/java/org/springframework/cloud/gcp/core/GCPProperties.java
@@ -23,7 +23,7 @@ import org.springframework.boot.context.properties.ConfigurationProperties;
  * @author Vinicius Carvalho
  */
 @ConfigurationProperties("cloug.gcp")
-public class GoogleCloudProperties {
+public class GCPProperties {
 
 	private String projectId;
 
--- a/spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/src/main/java/org/springframework/cloud/gcp/core/GoogleCloudProperties.java
+++ b/spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/src/main/java/org/springframework/cloud/gcp/core/GoogleCloudProperties.java
@@ -1,0 +1,47 @@
+/*
+ *  Copyright 2017 original author or authors.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ */
+
+package org.springframework.cloud.gcp.core;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+/**
+ * @author Vinicius Carvalho
+ */
+@ConfigurationProperties("cloug.gcp")
+public class GoogleCloudProperties {
+
+	private String projectId;
+
+	private String jsonKeyFile;
+
+	public String getProjectId() {
+		return projectId;
+	}
+
+	public void setProjectId(String projectId) {
+		this.projectId = projectId;
+	}
+
+	public String getJsonKeyFile() {
+		return jsonKeyFile;
+	}
+
+	public void setJsonKeyFile(String jsonKeyFile) {
+		this.jsonKeyFile = jsonKeyFile;
+	}
+}
--- a/spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/src/main/java/org/springframework/cloud/gcp/core/autoconfig/GCPContextAutoConfiguration.java
+++ b/spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/src/main/java/org/springframework/cloud/gcp/core/autoconfig/GCPContextAutoConfiguration.java
@@ -5,11 +5,9 @@ import java.io.FileInputStream;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
-import org.springframework.cloud.gcp.core.GoogleCloudProperties;
-import org.springframework.context.EnvironmentAware;
+import org.springframework.cloud.gcp.core.GCPProperties;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
-import org.springframework.core.env.Environment;
 import org.springframework.core.io.ResourceLoader;
 import org.springframework.util.StringUtils;
 
@@ -19,24 +17,24 @@ import com.google.auth.oauth2.GoogleCredentials;
  * @author Vinicius Carvalho
  *
  * Base starter for Google Cloud Projects. Provide defaults for {@link GoogleCredentials}.
- * Binds properties from {@link org.springframework.cloud.gcp.core.GoogleCloudProperties}
+ * Binds properties from {@link GCPProperties}
  *
  */
 @Configuration
 @ConditionalOnClass(GoogleCredentials.class)
-@EnableConfigurationProperties(GoogleCloudProperties.class)
-public class GoogleCloudContextAutoConfiguration {
+@EnableConfigurationProperties(GCPProperties.class)
+public class GCPContextAutoConfiguration {
 
 
 	@Autowired
-	private GoogleCloudProperties googleCloudProperties;
+	private GCPProperties gcpProperties;
 
 	@Bean
 	public GoogleCredentials googleCredentials(ResourceLoader resourceLoader)
 			throws Exception {
-		if (!StringUtils.isEmpty(this.googleCloudProperties.getJsonKeyFile())) {
+		if (!StringUtils.isEmpty(this.gcpProperties.getJsonKeyFile())) {
 			return GoogleCredentials.fromStream(new FileInputStream(resourceLoader
-					.getResource(this.googleCloudProperties.getJsonKeyFile())
+					.getResource(this.gcpProperties.getJsonKeyFile())
 					.getFile()));
 		}
 		else {
--- a/spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/src/main/java/org/springframework/cloud/gcp/core/autoconfig/GoogleCloudContextAutoConfiguration.java
+++ b/spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/src/main/java/org/springframework/cloud/gcp/core/autoconfig/GoogleCloudContextAutoConfiguration.java
@@ -2,7 +2,10 @@ package org.springframework.cloud.gcp.core.autoconfig;
 
 import java.io.FileInputStream;
 
+import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.cloud.gcp.core.GoogleCloudProperties;
 import org.springframework.context.EnvironmentAware;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
@@ -14,26 +17,26 @@ import com.google.auth.oauth2.GoogleCredentials;
 
 /**
  * @author Vinicius Carvalho
+ *
+ * Base starter for Google Cloud Projects. Provide defaults for {@link GoogleCredentials}.
+ * Binds properties from {@link org.springframework.cloud.gcp.core.GoogleCloudProperties}
+ *
  */
 @Configuration
 @ConditionalOnClass(GoogleCredentials.class)
-public class CredentialsAutoConfiguration implements EnvironmentAware {
+@EnableConfigurationProperties(GoogleCloudProperties.class)
+public class GoogleCloudContextAutoConfiguration {
 
-	private static final String AUTH_LOCATION_KEY = "cloud.gcp.auth.location";
 
-	private Environment environment;
-
-	@Override
-	public void setEnvironment(Environment environment) {
-		this.environment = environment;
-	}
+	@Autowired
+	private GoogleCloudProperties googleCloudProperties;
 
 	@Bean
 	public GoogleCredentials googleCredentials(ResourceLoader resourceLoader)
 			throws Exception {
-		if (!StringUtils.isEmpty(this.environment.getProperty(AUTH_LOCATION_KEY))) {
+		if (!StringUtils.isEmpty(this.googleCloudProperties.getJsonKeyFile())) {
 			return GoogleCredentials.fromStream(new FileInputStream(resourceLoader
-					.getResource(this.environment.getProperty(AUTH_LOCATION_KEY))
+					.getResource(this.googleCloudProperties.getJsonKeyFile())
 					.getFile()));
 		}
 		else {
--- a/spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/src/main/resources/META-INF/spring.factories
+++ b/spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/src/main/resources/META-INF/spring.factories
@@ -1,2 +1,2 @@
 org.springframework.boot.autoconfigure.EnableAutoConfiguration=\
-org.springframework.cloud.gcp.core.autoconfig.CredentialsAutoConfiguration
+org.springframework.cloud.gcp.core.autoconfig.GoogleCloudContextAutoConfiguration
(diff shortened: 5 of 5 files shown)


spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/src/main/java/org/springframework/cloud/gcp/core/GoogleCloudProperties.java
```diff
@@ -0,0 +1,47 @@
...
+ */
+
+package org.springframework.cloud.gcp.core;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+/**
+ * @author Vinicius Carvalho
+ */
+@ConfigurationProperties("cloug.gcp")
```
artembilan: Right, it is a common properties in the `core` module.

Well, I think that it must be `spring.cloud.gcp`. We may have too many ENV props and that would be great to distinguish those we are responsible.

meltsufin: I was just pointing out the typo in "cloud", but `spring.cloud.gcp` sounds good as well.

joaoandremartins:
Fixes #13

spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/src/main/java/org/springframework/cloud/gcp/core/autoconfig/GoogleCloudContextAutoConfiguration.java
```diff
@@ -14,26 +17,26 @@
...
+ *
+ * Base starter for Google Cloud Projects. Provide defaults for {@link GoogleCredentials}.
+ * Binds properties from {@link org.springframework.cloud.gcp.core.GoogleCloudProperties}
+ *
  */
 @Configuration
 @ConditionalOnClass(GoogleCredentials.class)
-public class CredentialsAutoConfiguration implements EnvironmentAware {
+@EnableConfigurationProperties(GoogleCloudProperties.class)
+public class GoogleCloudContextAutoConfiguration {
```
joaoandremartins: I don't know that this name is a very big improvement over GcpPropertiesAutoConfiguration. It confuses me greatly, but I'm no SI expert.
At a minimum, this should be GcpContextAutoConfiguration.

viniciusccarvalho: Naming is not a big issue, and it's the most useless waste of anyone's effort going on this. That is the very single thing that will take me off my base if we need to start having discussions around, because *NO ONE* gets naming right.

A few things on the naming:

- Usually the AutoConfiguration is related to some feature you want to add, We are not adding `Properties` on the auto configuration module, `Properties` are a feature that those modules can use to abstract variables from the system (environment, config files, config server, you name it). So `XXXPropertiesAutoConfiguration` doesn't make much sense.

- So the question is are we using `GoogleCloud` or `Gcp` which is the same as debate `Tomato or Tomato`. If the others think that Gcp is preferred than `GoogleCloud` by no means I'll make a problem out of it.

 I'd prefer to use GoogleCloud to be more specific about what this is about. Gcp could akso mean `Great Circus Parade` and we don't want that to confuse our users

artembilan: It's Spring Boot stuff. SI is far away from there.

Well, auto-config means provisioning of some beans into the application context and it make it based on some properties. We really don't configure here some `Properties` bean. 
I guess we are going to provide some common beans in this auto-config which really are related to the GCP Context. 
I don't have particular preferences to the name, but seems for me the whole name sounds much official than abbreviation.

joaoandremartins: I'm just afraid that using Gcp in some places and GoogleCloud in others can be a little confusing. I myself got confused with Spring-related naming before, so maybe it's not that small deal.
In any case, I too don't want to waste anyone's time over this. It seems GoogleCloud is the consensual choice, so be it.

meltsufin: For what it's worth, I'd cast my vote for using GCP. GCP is as much of a brand as AWS.

artembilan: Makes sense, too. Won't mind

viniciusccarvalho: GCP shall it be

spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/src/main/java/org/springframework/cloud/gcp/core/GCPProperties.java
```diff
@@ -0,0 +1,47 @@
...
+ */
+
+package org.springframework.cloud.gcp.core;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+/**
+ * @author Vinicius Carvalho
+ */
+@ConfigurationProperties("cloug.gcp")
```
artembilan: But you didn't fix this `cloug`! 😄

spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/src/main/java/org/springframework/cloud/gcp/core/GCPProperties.java
```diff
@@ -0,0 +1,47 @@
...
+public class GCPProperties {
+
+	private String projectId;
+
+	private String jsonKeyFile;
+
+	public String getProjectId() {
+		return projectId;
```
artembilan: What is our code style rules about `this.` prefix ?

spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/src/main/java/org/springframework/cloud/gcp/core/GCPProperties.java
joaoandremartins: Don't forget about this typo.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/src/main/java/org/springframework/cloud/gcp/core/GCPProperties.java
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * @author Vinicius Carvalho
+ */
+@ConfigurationProperties("cloug.gcp")
+public class GCPProperties {
+
+	private String projectId;
+
+	private String jsonKeyFile;
```
joaoandremartins: FYI keys don't need to be JSON, they can also be P12 for backwards compatibility.
https://developers.google.com/identity/protocols/OAuth2ServiceAccount

artembilan: M-m-m. Why it's not `Resource` from first of all?
`file://`, `classpath:` etc.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/src/main/java/org/springframework/cloud/gcp/core/autoconfig/GCPContextAutoConfiguration.java
```diff
@@ -2,38 +2,39 @@
...
 import org.springframework.core.io.ResourceLoader;
 import org.springframework.util.StringUtils;
 
 import com.google.auth.oauth2.GoogleCredentials;
 
 /**
  * @author Vinicius Carvalho
+ *
+ * Base starter for Google Cloud Projects. Provide defaults for {@link GoogleCredentials}.
+ * Binds properties from {@link GCPProperties}
```
joaoandremartins: Nit: period missing.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/src/main/java/org/springframework/cloud/gcp/core/autoconfig/GCPContextAutoConfiguration.java
```diff
@@ -2,38 +2,39 @@
...
 
-	private static final String AUTH_LOCATION_KEY = "cloud.gcp.auth.location";
 
-	private Environment environment;
-
-	@Override
-	public void setEnvironment(Environment environment) {
-		this.environment = environment;
-	}
+	@Autowired
```
joaoandremartins: I thought field injection in this way is discouraged, and we should instead use constructor or setter autowiring?

spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/src/main/java/org/springframework/cloud/gcp/core/autoconfig/GCPContextAutoConfiguration.java
```diff
@@ -2,38 +2,39 @@
...
 	public GoogleCredentials googleCredentials(ResourceLoader resourceLoader)
 			throws Exception {
-		if (!StringUtils.isEmpty(this.environment.getProperty(AUTH_LOCATION_KEY))) {
+		if (!StringUtils.isEmpty(this.gcpProperties.getJsonKeyFile())) {
 			return GoogleCredentials.fromStream(new FileInputStream(resourceLoader
-					.getResource(this.environment.getProperty(AUTH_LOCATION_KEY))
+					.getResource(this.gcpProperties.getJsonKeyFile())
 					.getFile()));
 		}
 		else {
```
joaoandremartins: Redundant else.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/src/main/java/org/springframework/cloud/gcp/core/autoconfig/GCPContextAutoConfiguration.java
```diff
@@ -2,38 +2,39 @@
...
+	private GCPProperties gcpProperties;
 
 	@Bean
 	public GoogleCredentials googleCredentials(ResourceLoader resourceLoader)
 			throws Exception {
-		if (!StringUtils.isEmpty(this.environment.getProperty(AUTH_LOCATION_KEY))) {
+		if (!StringUtils.isEmpty(this.gcpProperties.getJsonKeyFile())) {
 			return GoogleCredentials.fromStream(new FileInputStream(resourceLoader
-					.getResource(this.environment.getProperty(AUTH_LOCATION_KEY))
+					.getResource(this.gcpProperties.getJsonKeyFile())
```
joaoandremartins: Not using Resource.getInputStream(). Which is fine, I'm doing that change in #29 anyway.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/src/main/java/org/springframework/cloud/gcp/core/autoconfig/GCPContextAutoConfiguration.java
artembilan: Well, I'm always with this style, much simpler to type and keep me away from setters and let me stick with default ctor.

Would you mind, @joaoandremartins , to point me out where it says that field autowiring is evil?
Thanks

spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/src/main/java/org/springframework/cloud/gcp/core/autoconfig/GCPContextAutoConfiguration.java
artembilan: Not really. This is Spring style, even if we have one like of code.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/src/main/java/org/springframework/cloud/gcp/core/autoconfig/GCPContextAutoConfiguration.java
artembilan: Correct. That is what I was going to do during merge.
But right now with your `jsonFile` concern I have placed comment about `Resource` directly.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/src/main/java/org/springframework/cloud/gcp/core/autoconfig/GCPContextAutoConfiguration.java
joaoandremartins: screenshot from 2017-05-31 18 13 18

spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/src/main/java/org/springframework/cloud/gcp/core/autoconfig/GCPContextAutoConfiguration.java
joaoandremartins: If the else being there or not is the same, I'd rather get rid of the else. But if it's your style, then let's keep it. I also wouldn't argue with Vinicius if he didn't think this PR was the right place to change it.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/src/main/java/org/springframework/cloud/gcp/core/autoconfig/GCPContextAutoConfiguration.java
artembilan: 😄 , funny, but I'm not aware of that.
OTOH it says "in your beans". Here this is Spring own classes 😉 

Would be great to understand background.
Thank you for the pointer!

spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/src/main/java/org/springframework/cloud/gcp/core/GCPProperties.java
viniciusccarvalho: it's missing, you are right, those were generated by intellij which we should probably set to use `this.` as convention from Spring guidelines

spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/src/main/java/org/springframework/cloud/gcp/core/GCPProperties.java
joaoandremartins: I didn't know Spring automatically transformed resource descriptors from the property into `Resource` objects. If it does that, then it's pretty neat! Saves us some trouble below.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/src/main/java/org/springframework/cloud/gcp/core/autoconfig/GCPContextAutoConfiguration.java
viniciusccarvalho: FWIW: The general recommendation for setters and constructors is to aid in tests. We have a saying, every time you autowire a property a unit test dies. But as pointed by Artem, this is an internal class, much like when we create Configuration classes for tests, we don't mind injecting the dependencies on the test itself.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/src/main/java/org/springframework/cloud/gcp/core/GCPProperties.java
artembilan: It does. See `KafkaProperties.Ssl`:
```
/**
 * Location of the key store file.
 */
private Resource keystoreLocation;
```

artembilan:
So, if we agreed on all those concerns, can I go ahead and fix on merge?

@viniciusccarvalho , otherwise, please consider `spring.cloud.gcp` for properties prefix.

viniciusccarvalho:
Yes we should use `spring.cloud.gcp` I just forgot at all about changing the typo. I did no change the Resource since we have an issue to fix that.

artembilan:
OK. Merging with simple polishing then.
Would be great to have a cross link to that `Resource` issue though.
Just for better traceability to determine who we're doing here.

Thanks

artembilan:
Merged as https://github.com/spring-cloud/spring-cloud-gcp/commit/b0864e3c90e8f58585162a00a6ee7b46eeda611e

spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/src/main/java/org/springframework/cloud/gcp/core/autoconfig/GCPContextAutoConfiguration.java
markfisher: I think @joaoandremartins was referring to the fact that the `if` block above had a `return`, not just the fact that there's one line in the `else` block.

PR #38, opened 2017-06-01
Add code formatting rules

### LINKED ISSUE #15: Add checkstyle file to the project

viniciusccarvalho:
- new etc/ folder with eclipse code formatter and checkstyle rules
- refactoring of classes where import order / copyright were not
  following spec
- adding section on readme on how to enable eclipse code formatter
- fixes #15

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -1,0 +1,15 @@
+# spring-cloud-gcp
+Integration for Google Cloud Platform APIs with Spring
+
+=== Code formatting guidelines
+
+* The directory ./etc/eclipse has two files for use with code formatting, `eclipse-code-formatter.xml` for the majority of the code formatting rules and `eclipse.importorder` to order the import statements.
+
+* In eclipse you import these files by navigating `Windows -> Preferences` and then the menu items `Preferences > Java > Code Style > Formatter` and `Preferences > Java > Code Style > Organize Imports` respectfully.
+
+* In `IntelliJ`, install the plugin `Eclipse Code Formatter`.  You can find it by searching the "Browse Repositories" under the plugin option within `IntelliJ` (Once installed you will need to reboot Intellij for it to take effect).
+Then navigate to `Intellij IDEA > Preferences` and select the Eclipse Code Formatter.
+Select the `eclipse-code-formatter.xml` file for the field `Eclipse Java Formatter config file` and the file `eclipse.importorder` for the field `Import order`.
+Enable the `Eclipse code formatter` by clicking `Use the Eclipse code formatter` then click the *OK* button.
+
+** NOTE: If you configure the `Eclipse Code Formatter` from `File > Other Settings > Default Settings` it will set this policy across all of your Intellij projects.
--- a/README.adoc
+++ b/README.adoc
@@ -5,9 +5,9 @@ Integration for Google Cloud Platform APIs with Spring
 
 * The directory ./etc/eclipse has two files for use with code formatting, `eclipse-code-formatter.xml` for the majority of the code formatting rules and `eclipse.importorder` to order the import statements.
 
-* In eclipse you import these files by navigating `Windows -> Preferences` and then the menu items `Preferences > Java > Code Style > Formatter` and `Preferences > Java > Code Style > Organize Imports` respectfully.
+* In Eclipse you import these files by navigating `Windows -> Preferences` and then the menu items `Preferences > Java > Code Style > Formatter` and `Preferences > Java > Code Style > Organize Imports` respectfully.
 
-* In `IntelliJ`, install the plugin `Eclipse Code Formatter`.  You can find it by searching the "Browse Repositories" under the plugin option within `IntelliJ` (Once installed you will need to reboot Intellij for it to take effect).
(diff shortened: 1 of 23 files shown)


pom.xml
```diff
@@ -97,6 +97,31 @@
...
+				<groupId>org.apache.maven.plugins</groupId>
+				<artifactId>maven-checkstyle-plugin</artifactId>
+				<version>2.17</version>
+				<executions>
+					<execution>
+						<id>checkstyle-validation</id>
+						<phase>validate</phase>
+						<configuration>
+							<encoding>UTF-8</encoding>
+							<!--<failsOnError>true</failsOnError>-->
```
joaoandremartins: Uncomment or remove.

meltsufin:
Thanks for adding this! The formatter works for me, although I had to disable an internal code-assist plugin.

README.adoc
```diff
@@ -0,0 +1,15 @@
+# spring-cloud-gcp
+Integration for Google Cloud Platform APIs with Spring
+
+=== Code formatting guidelines
+
+* The directory ./etc/eclipse has two files for use with code formatting, `eclipse-code-formatter.xml` for the majority of the code formatting rules and `eclipse.importorder` to order the import statements.
+
+* In eclipse you import these files by navigating `Windows -> Preferences` and then the menu items `Preferences > Java > Code Style > Formatter` and `Preferences > Java > Code Style > Organize Imports` respectfully.
```
artembilan: Eclipse? 
Capitalized, I mean

s/respectfully/respectively ? 😄

README.adoc
```diff
@@ -0,0 +1,15 @@
...
+
+* In `IntelliJ`, install the plugin `Eclipse Code Formatter`.  You can find it by searching the "Browse Repositories" under the plugin option within `IntelliJ` (Once installed you will need to reboot Intellij for it to take effect).
```
artembilan: No reason to wrap IntelliJ to the code. And brand name is **IntelliJ IDEA**

And this is official open for everyone document, so some typos like `Intellij` aren't appropriate

pom.xml
artembilan: Yes, uncomment and fix Checkstyle vulnerabilities.

etc/checkstyle/checkstyle.xml
```diff
@@ -0,0 +1,167 @@
...
+			<property name="max" value="3"/>
+		</module>
+		<module name="NestedIfDepth">
+			<property name="max" value="3"/>
+		</module>
+		<module name="NestedTryDepth">
+			<property name="max" value="3"/>
+		</module>
+		<module name="MultipleVariableDeclarations"/>
+		<module name="RequireThis">
```
meltsufin: Is the idea to enable checks like this one gradually?

artembilan: +1
Since the project is still small that would be better to enable them all and fix all the vulnerability. So, all subsequent new members will suffer from all the rules already.

So, yes, let's switch them on right now already!

Is that hard for you @viniciusccarvalho ?

The question from me: how to use the Ccheckstyle task during project build?
In Gradle that is pretty easy - just `check` instead of `test`.
Thanks

viniciusccarvalho: if you run a full build like clean install, I know it runs the checkstyle goal.

This file have been copied from what we currently use as default on `spring-cloud-stream` and `acceptance-tests`. Ilaya is probably the best person to discuss all the flags that were decided to enable/disable

artembilan: I'm not sure why we should involve Ilaya if this project is new can be the best from the beginning.
Those rules are not switched on those projects because they are so big and it is really harmful to fix all the problem. Although we still chase the goal in Spring Integration to add JavaDocs to all the public API to meet as much as possible Checkstyle rules.

pom.xml
```diff
@@ -97,6 +97,30 @@
...
+		<plugins>
+			<plugin>
+				<groupId>org.apache.maven.plugins</groupId>
+				<artifactId>maven-checkstyle-plugin</artifactId>
+				<version>2.17</version>
+				<executions>
+					<execution>
+						<id>checkstyle-validation</id>
+						<phase>validate</phase>
+						<configuration>
```
meltsufin: If you move this configuration outside of the `<execution>`, you'll be able to run checkstyle with just `mvn checkstyle:check`.

etc/checkstyle/checkstyle.xml
viniciusccarvalho: I don't think we need to involve him, I was just saying I'm not sure what was the decision behind choosing this specific set of rules. I pair with Syer on my feelings about checkstyle. I really hate it, but ...

artembilan: Well, here is Spring Kafka sample: https://github.com/spring-projects/spring-kafka/blob/master/src/checkstyle/checkstyle.xml

No comments any more. 
I'm fine to leave without rules, but I will always spam your email box with  the "`this.` is required"...

joaoandremartins:
Opened #55 with several fixes on this.

pom.xml
joaoandremartins: What do you mean? Move the `<execution>` block out of `<plugin>`?

PR #40, opened 2017-06-02
Changes jsonKeyFile from String to Resource and renames it

### LINKED ISSUE #37: Change GCPProperties.jsonKeyFile type to Resource
It should probably also be renamed to something else, since JSON isn't the only format supported by Google's OAuth2 credentials.

joaoandremartins:
GCP Credentials aren't necessarily JSON, they can be P12 too.
Leverages Spring resource loading.

Fixes #37

### DIFF
--- a/spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/src/main/java/org/springframework/cloud/gcp/core/GCPProperties.java
+++ b/spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/src/main/java/org/springframework/cloud/gcp/core/GCPProperties.java
@@ -18,6 +18,7 @@
 package org.springframework.cloud.gcp.core;
 
 import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.core.io.Resource;
 
 /**
  * @author Vinicius Carvalho
@@ -27,7 +28,7 @@ public class GCPProperties {
 
 	private String projectId;
 
-	private String jsonKeyFile;
+	private Resource credentialsLocation;
 
 	public String getProjectId() {
 		return this.projectId;
@@ -37,12 +38,12 @@ public class GCPProperties {
 		this.projectId = projectId;
 	}
 
-	public String getJsonKeyFile() {
-		return this.jsonKeyFile;
+	public Resource getCredentialsLocation() {
+		return this.credentialsLocation;
 	}
 
-	public void setJsonKeyFile(String jsonKeyFile) {
-		this.jsonKeyFile = jsonKeyFile;
+	public void setCredentialsLocation(Resource credentialsLocation) {
+		this.credentialsLocation = credentialsLocation;
 	}
 
 }
--- a/spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/src/main/java/org/springframework/cloud/gcp/core/autoconfig/GCPContextAutoConfiguration.java
+++ b/spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/src/main/java/org/springframework/cloud/gcp/core/autoconfig/GCPContextAutoConfiguration.java
@@ -6,7 +6,6 @@ import org.springframework.boot.context.properties.EnableConfigurationProperties
 import org.springframework.cloud.gcp.core.GCPProperties;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
-import org.springframework.core.io.ResourceLoader;
 import org.springframework.util.StringUtils;
 
 import com.google.auth.oauth2.GoogleCredentials;
@@ -23,16 +22,14 @@ import com.google.auth.oauth2.GoogleCredentials;
 @EnableConfigurationProperties(GCPProperties.class)
 public class GCPContextAutoConfiguration {
 
-
 	@Autowired
 	private GCPProperties gcpProperties;
 
 	@Bean
-	public GoogleCredentials googleCredentials(ResourceLoader resourceLoader) throws Exception {
-		if (!StringUtils.isEmpty(this.gcpProperties.getJsonKeyFile())) {
+	public GoogleCredentials googleCredentials() throws Exception {
+		if (!StringUtils.isEmpty(this.gcpProperties.getCredentialsLocation())) {
 			return GoogleCredentials.fromStream(
-					resourceLoader.getResource(this.gcpProperties.getJsonKeyFile())
-							.getInputStream());
+					this.gcpProperties.getCredentialsLocation().getInputStream());
 		}
 		else {
 			return GoogleCredentials.getApplicationDefault();


spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/src/main/java/org/springframework/cloud/gcp/core/GCPProperties.java
```diff
@@ -27,7 +28,7 @@
 
 	private String projectId;
 
-	private String jsonKeyFile;
+	private Resource credentialKeyFile;
```
artembilan: M-m-m. Maybe would be better `credentialsFile` ?

spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/src/main/java/org/springframework/cloud/gcp/core/autoconfig/GCPContextAutoConfiguration.java
```diff
@@ -23,16 +22,14 @@
...
 
 	@Bean
-	public GoogleCredentials googleCredentials(ResourceLoader resourceLoader) throws Exception {
-		if (!StringUtils.isEmpty(this.gcpProperties.getJsonKeyFile())) {
+	public GoogleCredentials googleCredentials() throws Exception {
+		if (!StringUtils.isEmpty(this.gcpProperties.getCredentialKeyFile())) {
 			return GoogleCredentials.fromStream(
-					resourceLoader.getResource(this.gcpProperties.getJsonKeyFile())
-							.getInputStream());
+					this.gcpProperties.getCredentialKeyFile().getInputStream());
```
artembilan: Are we going to have tests on the matter?
I don't exactly for this change, but the whole test suit for the `GCPContextAutoConfiguration` at least on mocks.

Just an idea, no need to address in this PR.
Thanks

spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/src/main/java/org/springframework/cloud/gcp/core/GCPProperties.java
meltsufin: The corresponding environment variable is called `GOOGLE_APPLICATION_CREDENTIALS`. So, maybe just `Resource credentials`?

artembilan: Fine with me

joaoandremartins: Actually, I'm thinking it might be confusing for users to have two places to get "credentials" from -- `GcpProperties` and the injected `GoogleCredentials` object (by `GcpContextAutoConfiguration`).

In the following example, the user would get the project ID from `GcpProperties` and the credentials from a different `GoogleCredentials` object, even though `GcpProperties` would contain a credentials field.

```
  @Bean
  public MessageHandler messageSender(GCPProperties gcpProperties,
      GoogleCredentials credentials) throws IOException {
    PubSubMessageHandler outboundAdapter =
        new PubSubMessageHandler(gcpProperties.getProjectId(), credentials);
    outboundAdapter.setTopic("test");
    return outboundAdapter;
  }
```

I'm wondering if it's possible to turn this field into `GoogleCredentials`, instead, and perform some kind of magic similar to the `String` -> `Resource` conversion. WDYT @artembilan ?

spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/src/main/java/org/springframework/cloud/gcp/core/autoconfig/GCPContextAutoConfiguration.java
joaoandremartins: Agreed, I actually intended to file a new issue to address any unit test needs.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/src/main/java/org/springframework/cloud/gcp/core/GCPProperties.java
viniciusccarvalho: They should not be getting credentials from the Properties. In boot, the context auto configuration should provide the credentials.

Now, how does that object get created? Either via defaults or, if someone set the path to the credentials file, it comes from that file.

meltsufin: What about the project ID? Should they get it from the `GcpProperties`?

artembilan: Well, here is a story about fundamental purpose of Spring Boot's auto-configuration and dependency injection. That `@ServiceActivator` bean definition is a good sample of dependency injection. And here it doesn't care from where those dependency beans come.

The auto-config is *auto* only if there is no user-defined beans. So, if you decide to declare, e.g. `MockGoogleCredentials`, the auto-config won't look into that property. But this `@ServiceActivator` will be the same and free of the inversion control machinery. 

Does it make sense to you now?

viniciusccarvalho: Tha analogy is AWS. There's an AWSCredentials object that gets created from SECRET_KEY and ACCESS_KEY. Those two strings are represented in a Properties class. The auto configuration creates the AWSCredentials using those two provided properties.

Could users use the values from the properties to get the credentials? Yes, if they are willing to go and create a new Credentials object, but why would they if there's already a bean created for them.

Same applies for a datasource for example. you have a property called `datasource.username` & `datasource.password` that is available for anyone via properties, but Boot will create a DataSource object for you so you don't need to create yourself

joaoandremartins: Absolutely agreed. Just to clarify, I'm not against injecting `GoogleCredentials` and using it in the adapters.

My only concern is that it might be confusing for users that project ID is gotten from `GcpProperties`, which also contains a `credentials` named field, but they must get the credentials from somewhere else.

Another way to resolve this confusion is name the field `credentialsLocation` and document how channel adapters must be created.

viniciusccarvalho: @meltsufin Anything that we believe needs to be configurable via Boot's abstraction I think should be on the GCPProperties. Things like : projectId, Region, credentials.

If there's a way to obtain the projectId from the credentials (I think I asked that before and got a negative reply)

Then we could make it a default value being the one coming from the credentials.

The sole reason of the Properties file is to be a placeholder to carry config options. Nothing more.

AutoConfiguration classes than rely on that to create the beans as well as to make decision on Conditional. There's a `ConditionalOnProperty` that can assist on activation of beans depending on properties being set or not.

meltsufin: @viniciusccarvalho Make sense. I just wasn't sure if using `GcpProperties.projectId` outside of auto-configuration machinery is ok, as in @joaoandremartins example. I guess the answer is yes.

artembilan: Well, that's maybe how `ProjectIdProvider` would fix the concern. That is absolutely what we have with the `RegionProvider` in the SC AWS. But yes, that is fine to inject `GcpProperties` in to target code and perform some logic. 
In the end the `GcpProperties` is a bean anyway!

meltsufin: Ahhh the `ProjectIdProvider` was the missing piece here. We should auto-configure `ProjectIdProvider` based on `GcpProperties.projectId` or other possible sources, and then client code can use that. `PubSubMessageHandler` and `PubSubTemplate` would be configured with `ProjectIdProvider` instead of `String projectId`. Am I understanding this correctly?

viniciusccarvalho: Let's separate two things: How to use the lib in boot and how to use it as standalone

So, the standalone lib, or the template, should just have a dependency on credentials and project id.

How they get populated, it will depend on how the user is using SI. For example someone may still be relying on XML to configure the beans, and it may turn out that they use this:
```
<bean id="pubsubTemplate" class "...">
<property name="credentials" ref""/>
<property name="projectId" value="mygooglecloudprojectid"/>
</bean>```

This is an example, I know we are using constructors, it's just to show that how someone creates a credential object or pass the project is highly dependable on how they are configuring their spring project.

Boot users, on the other hand, can benefit from autoconfiguration (many times they choose not to BTW) and that's where GCPProperties live.

Does it makes sense now?

Comment (meltsufin) [inline spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/src/main/java/org/springframework/cloud/gcp/core/GCPProperties.java]: Sort of. So for @joaoandremartins' example. Is this how we envision it?

```java
  @Bean
  public MessageHandler messageSender(ProjectIdProvider projectIdProvider,
      GoogleCredentials credentials) throws IOException {
    PubSubMessageHandler outboundAdapter =
        new PubSubMessageHandler(projectIdProvider.getProjectId(), credentials);
    outboundAdapter.setTopic("test");
    return outboundAdapter;
  }
```

viniciusccarvalho: @meltsufin Depends where are you creating that `MessageHandler`, this is what I'm trying to explain.

So let's get some basic scenarios perhaps

1: Your constructor is `PubSubMessageHandler(String projectId, GoogleCredentials credentials)

Now, let's say you are using the starter package on a Boot app:


```java
@Autowired 
private GcpProperties properties;

@Bean
  public MessageHandler messageSender(
      GoogleCredentials credentials) throws IOException {
    PubSubMessageHandler outboundAdapter =
        new PubSubMessageHandler(properties.getProjectId(), credentials);
    outboundAdapter.setTopic("test");
    return outboundAdapter;
  }
```

That makes sense for a Boot app.

But what if your project is not a boot app?

```java
//note you could use @Value too
@Autowired
private Environment env;
  @Bean
  public GoogleCloudCredentials credentials(){
// create credentials manually here, somehow
}
  @Bean
  public MessageHandler messageSender(GoogleCredentials credentials) throws IOException {
    PubSubMessageHandler outboundAdapter =
        new PubSubMessageHandler(env.getProperty("MY_ENV_PROPERTY_FOR_PROJECT_ID"), credentials);
    outboundAdapter.setTopic("test");
    return outboundAdapter;
  }
```

My main concern is explaining how you keep those two concerns separate (AutoConfiguration module along with the Property binding) and your MessageHandler code.

Hope this helps

meltsufin: @viniciusccarvalho This definitely helps. I think the part I was missing is that in non-boot Spring, there is no `@ConfigurationProperties`. Consequently, using `GcpProperties` in `MessageHandler` would make it specific to boot. The part I'm still not completely sure on is `GcpProjectIdProvider`. Where would that be useful?

joaoandremartins: From my point of view, there are a few advantages to GcpProjectIdProvider.
1) It reduces confusion by not requiring the user to get project ID from GcpProperties (and being confused with the pseudo-credentials in it),
2) It helps with testing, with the benefit that we don't need to inject/mock Strings but rather get something more strongly typed.

In any case, it seems better not to worry about this too much in this PR. I might file an issue to look into this later.

joaoandremartins:
I'm starting to notice that the squash -> force push approach always clears active comments in a PR.

Do you know any way around this? Only squash at the end?

meltsufin:
@joaoandremartins Why bother squashing before you're ready to merge PR?

joaoandremartins:
Hmm I thought it would make it easier for the reviewer to merge. However, perhaps it should be up to the reviewer to ultimately squash and merge? This has the downside that the reviewer has to checkout the PR author's fork IIUC.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/src/main/java/org/springframework/cloud/gcp/core/GCPProperties.java
viniciusccarvalho: I don't think that a GcpProjectIdProvider makes sense, boot users are used to retrieve information from properties, that's the Spring way. Having a provider that provides a String is overkill as we mentioned here before. The Environment abstraction was created for that.

If project is a top level construct, with more information required for the utilization of the libraries (I have not seen any indication so far) than it could be something to be explored.

The idea of a auto configuration for a **Context** is to provide all the required dependencies for a cloud context to be available. `Credentials`, `Region`, `ProjectId`.

@meltsufin you would not use GcpProperties in the message handler, that's the whole point here. The message handler needs a `Credentials` object and a `String` that represents the projectId. How that is passed down will depend on how you use the library.

If you guys are still confused about this, I suggest opening a few AutoConfiguration classes from Boot, and check how the Properties from Boot play along with classes such as Redis,Rabbit fora example. The spring-data projects will require certain properties like the host and port where redis is located, they do not depend on the RedisProperties class, instead on Strings and Integers. The Autoconfiguration classes will depend on the Properties and use them to configure the connection on redis for example.

joaoandremartins: @viniciusccarvalho I think we're on the same page on the Boot and non-Boot question.

The issue that opened this whole discussion, however, is still unanswered: Boot users are expected to get the project ID from `GcpProperties` but not `GoogleCredentials`, even though `GcpProperties` would contain a `credential` member.

To tackle that a little bit, I renamed the member to `credentialsLocation`, so that it's a little bit clearer to the users that they can't use that member to feed into the adapters (if the `Resource`/`GoogleCredentials` type mismatch wasn't already strong a hint enough :) ). It sounds there isn't much more to do there.

viniciusccarvalho: Yes, and that's not an issue. Think about Datasources again, simple plain old JDBC datasources.

When you import spring-data-jdbc it will look for a property called `spring.datasource.username` and `spring.datasource.password`, at the same time a full blown DataSource object is created in the context. 

Now, the users can if they want to access the username/password (analogy to the resourceFile) and create their own datasource (analogy to GoogleCredentials)

And as a matter of fact, this is how bypassing autoconfiguration works, someone may decide to create the Credentials in a different manner.

So having the constructor primitive for your credentials and a credentials object is not a problem at all.

meltsufin: My concern was more about supporting autoconfiguration of project ID in Boot. Created issue #44 to track and discuss that.

artembilan:
This PR has conflicts with `master` and can't be rebased properly.
This is contributor responsibility to address this kind of conflicts.
Thanks

spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/src/main/java/org/springframework/cloud/gcp/core/GcpProperties.java
```diff
@@ -27,7 +28,7 @@
 
 	private String projectId;
 
-	private String jsonKeyFile;
+	private Resource credentialsLocation;
```
artembilan: One of cool Spring Boot features is `Meta-Data assistance`: http://docs.spring.io/spring-boot/docs/1.5.3.RELEASE/reference/htmlsingle/#configuration-metadata
Therefor these properties must be with JavaDocs.

In the next section you can find how to generate such a meta-data automatically: 
http://docs.spring.io/spring-boot/docs/1.5.3.RELEASE/reference/htmlsingle/#configuration-metadata-annotation-processor

See #47

artembilan:
Merged as https://github.com/spring-cloud/spring-cloud-gcp/commit/4f7697fa8a4c45c3f8cbea23a03df32b2e1de708

PR #43, opened 2017-06-05
Makes PubSubOperations.send return ListenableFuture

### LINKED ISSUE #36: Implement ApiFuture -> ListenableFuture adapter for PubSubTemplate's send()
Currently, `PubSubTemplate.send()` returns a String with `PubsubMessage`'s ID. We block the call to publish() and, ideally, we'd return a future instead.

We should implement a PubSub `ApiFuture` adapter to Spring's `ListenableFuture`, so the user doesn't have to use anything PubSub.

joaoandremartins:
Gives control to the user to make a blocking call or not wait for its
result.

Fixes #36

### DIFF
--- a/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubOperations.java
+++ b/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubOperations.java
@@ -18,12 +18,13 @@
 package org.springframework.cloud.gcp.pubsub.core;
 
 import org.springframework.messaging.Message;
+import org.springframework.util.concurrent.ListenableFuture;
 
 /**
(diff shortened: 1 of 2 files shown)


joaoandremartins:
After this, we might want `PubSubMessageHandler` to block on `send()` and throw a `PubSubException` if an exception occurred.

artembilan:
Merged as https://github.com/spring-cloud/spring-cloud-gcp/commit/21a0a5d7eac6e01b3655e2ab0a76205b82755489.

> After this, we might want PubSubMessageHandler to block on send() and throw a PubSubException if an exception occurred.

No, we don't. See the same `KafkaProducerMessageHandler` with its `sync` option: https://github.com/spring-projects/spring-integration-kafka/blob/master/src/main/java/org/springframework/integration/kafka/outbound/KafkaProducerMessageHandler.java#L104

Issue #26, opened 2017-05-26
Create a PubSubAdmin utility


viniciusccarvalho:
Much like we have today with RabbitMQ, it would be nice to extract all the admin stuff from pubsub into a separate class.

One of the main reasons, besides good separation of concerns, is that we could use it to create topics from the storage module when users decide to enable notification from a bucket.

joaoandremartins:
In addition, this could also be where the [helpers](https://github.com/spring-cloud/spring-cloud-gcp/pull/61#issuecomment-312044648) that create `PublisherFactory`, `SubscriberFactory`, etc. objects will live, for non-Boot users.

PR #50, opened 2017-06-06
Sets client lib name and version

### LINKED ISSUE #12: Investigate metadata to correlate utilization
The idea is to provide a default header in our PubSubMessageTemplate that could be use to track utilization. Something on the same aspects as the `Agent` header of HTTP.

Let's take this as a Spike to verify first the feasibility of this approach

joaoandremartins:
Allows library usage to be tracked from the server. Some polishing on
the side.

Fixes #12

### DIFF
--- a/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
+++ b/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
@@ -51,30 +51,56 @@ public class PubSubTemplate implements PubSubOperations, InitializingBean {
 
 	private ConcurrentHashMap<String, Publisher> publishers = new ConcurrentHashMap<>();
 
-	private ExecutorProvider executorProvider;
-
 	private MessageConverter messageConverter = new SimpleMessageConverter();
 
-	private int concurrentProducers = 1;
+	/**
+	 * The Spring component a Google Cloud Pub/Sub call is originated from.
+	 *
+	 * <p>Spring Integration callers should set this to "spring-integration", Spring Cloud to
+	 * "spring-cloud", etc.
+	 */
+	private String sourceName = "spring";
+
+	/**
+	 * The version of the Spring component a Google Cloud Pub/Sub call is originated from.
+	 *
+	 * <p>By default, the version is obtained from this package's manifest file, which is mirrored
+	 * from this package's pom.xml file. Ideally, this would be the version of the library invoking
+	 * this call (e.g., Spring Integration, Spring Cloud, etc.).
+	 */
+	private String sourceVersion;
+
+	/**
+	 * Aimed at reusing the same executor across every {@link Publisher}, so the number of allocated
+	 * threads doesn't blow out of proportion.
+	 */
+	private ExecutorProvider executorProvider;
+
+	/**
+	 * Number of threads to be used by the shared executor.
+	 *
+	 * <p>If not set, uses the default value from the gax library.
+	 */
+	private Integer executorThreadCount;
 
 	public PubSubTemplate(GoogleCredentials credentials, String projectId) {
 		this.projectId = projectId;
 		this.credentials = credentials;
-		this.executorProvider = InstantiatingExecutorProvider.newBuilder()
-				.setExecutorThreadCount(this.concurrentProducers).build();
+		this.sourceVersion = this.getClass().getPackage().getImplementationVersion();
 	}
 
 	@Override
 	public ListenableFuture<String> send(final String topic, Message message) {
-
-		Publisher publisher = this.publishers.computeIfAbsent(topic, s -> {
+		// Reuse existing publisher if one already exists.
+		Publisher publisher = this.publishers.computeIfAbsent(topic, key -> {
 			try {
-				return Publisher.defaultBuilder(TopicName.create(this.projectId, topic))
+				return Publisher.defaultBuilder(TopicName.create(this.projectId, key))
 						.setExecutorProvider(this.executorProvider)
 						.setChannelProvider(
 								TopicAdminSettings
 										.defaultChannelProviderBuilder()
 										.setCredentialsProvider(() -> this.credentials)
+										.setClientLibHeader(this.sourceName, this.sourceVersion)
 										.build())
 						.build();
 			}
@@ -84,13 +110,16 @@ public class PubSubTemplate implements PubSubOperations, InitializingBean {
 			}
 		});
 
-		Object pubsubMessageObject = this.messageConverter.fromMessage(message, PubsubMessage.class);
+		// Convert from payload into PubsubMessage.
+		Object pubsubMessageObject =
+				this.messageConverter.fromMessage(message, PubsubMessage.class);
 
 		if (!(pubsubMessageObject instanceof PubsubMessage)) {
 			throw new MessageConversionException("The specified converter must produce "
 					+ "PubsubMessages to send to Google Cloud Pub/Sub.");
 		}
 
+		// Send message to Google Cloud Pub/Sub.
 		ApiFuture<String> publishFuture = publisher.publish((PubsubMessage) pubsubMessageObject);
 
 		final SettableListenableFuture<String> settableFuture = new SettableListenableFuture<>();
@@ -113,7 +142,14 @@ public class PubSubTemplate implements PubSubOperations, InitializingBean {
 
 	@Override
 	public void afterPropertiesSet() throws Exception {
+		InstantiatingExecutorProvider.Builder executorProviderBuilder =
+				InstantiatingExecutorProvider.newBuilder();
 
+		if (this.executorThreadCount != null) {
+			executorProviderBuilder.setExecutorThreadCount(this.executorThreadCount);
+		}
+
+		this.executorProvider =	executorProviderBuilder.build();
 	}
 
 	public MessageConverter getMessageConverter() {
@@ -124,4 +160,11 @@ public class PubSubTemplate implements PubSubOperations, InitializingBean {
 		this.messageConverter = messageConverter;
 	}
 
+	public int getExecutorThreadCount() {
+		return this.executorThreadCount;
+	}
+
+	public void setExecutorThreadCount(int executorThreadCount) {
+		this.executorThreadCount = executorThreadCount;
+	}
 }
--- a/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
+++ b/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
@@ -85,7 +85,7 @@ public class PubSubTemplate implements PubSubOperations, InitializingBean {
 		this.credentials = credentials;
 		this.sourceVersion = this.getClass().getPackage().getImplementationVersion();
 		this.executorProvider = InstantiatingExecutorProvider.newBuilder()
-				.setExecutorThreadCount(concurrentProducers).build();
+				.setExecutorThreadCount(this.concurrentProducers).build();
 	}
 
 	@Override
--- a/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
+++ b/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
@@ -45,11 +45,19 @@ import com.google.pubsub.v1.TopicName;
  */
 public class PubSubTemplate implements PubSubOperations, InitializingBean {
 
+	private static final String DEFAULT_SOURCE_NAME = "spring";
+
 	private final String projectId;
 
 	private final GoogleCredentials credentials;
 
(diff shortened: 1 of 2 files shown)


joaoandremartins:
As discussed in the meeting, I couldn't come up with a way to add headers to the Spring `Message` in `PubSubMessageHandler`, so I'm leaving changing `sourceName` and `sourceVersion` out of this PR. Will file a new issue after this PR is merged so this gets addressed eventually.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
```diff
@@ -124,4 +160,11 @@ public void setMessageConverter(MessageConverter messageConverter) {
 		this.messageConverter = messageConverter;
 	}
 
+	public int getExecutorThreadCount() {
+		return this.executorThreadCount;
+	}
+
+	public void setExecutorThreadCount(int executorThreadCount) {
+		this.executorThreadCount = executorThreadCount;
```
meltsufin: If the thread count is set after the bean is initialized, it will have no effect. So, you either have to update the thread pool setting, or only allow the variable to be set in the constructor.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
```diff
@@ -51,30 +51,56 @@
...
 	private MessageConverter messageConverter = new SimpleMessageConverter();
 
-	private int concurrentProducers = 1;
+	/**
+	 * The Spring component a Google Cloud Pub/Sub call is originated from.
+	 *
+	 * <p>Spring Integration callers should set this to "spring-integration", Spring Cloud to
+	 * "spring-cloud", etc.
+	 */
+	private String sourceName = "spring";
```
meltsufin: Looks like it should be a constant - `SOURCE_NAME`, and the same for `SOURCE_VERSION`.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
```diff
@@ -51,30 +51,56 @@
...
 	@Override
 	public ListenableFuture<String> send(final String topic, Message message) {
-
-		Publisher publisher = this.publishers.computeIfAbsent(topic, s -> {
+		// Reuse existing publisher if one already exists.
+		Publisher publisher = this.publishers.computeIfAbsent(topic, key -> {
 			try {
-				return Publisher.defaultBuilder(TopicName.create(this.projectId, topic))
+				return Publisher.defaultBuilder(TopicName.create(this.projectId, key))
 						.setExecutorProvider(this.executorProvider)
```
meltsufin: Won't this actually cause a new thread pool of size `executorThreadCount` be created for each Publisher/topic?

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
```diff
@@ -51,30 +51,56 @@
...
 						.setChannelProvider(
 								TopicAdminSettings
 										.defaultChannelProviderBuilder()
 										.setCredentialsProvider(() -> this.credentials)
+										.setClientLibHeader(this.sourceName, this.sourceVersion)
```
artembilan: Looking at this closely, it is definitely correct to have it as setters on the component which is closer to this `InstantiatingChannelProvider`. Therefore in our case `PubSubTemplate`.

That would be pointless to reconfigure it against header from each message to send.

OTOH let's imaging that we will have some Spring Boot auto-config for these option, too.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
meltsufin: +1 I would replace the `executorThreadCount` with just `executorProvider`.

joaoandremartins: This one gave me a bit to think about. I don't really like updating the thread pool setting, since that would mean we'd have to rebuild every existing publisher to use the new settings.

OTOH also not a big fan of having that setting in the constructor. It's what makes more sense, but it could confuse users?

We could have two constructors, one with and one without `executorThreadCount`, which just uses the default. Would that be reasonable?

joaoandremartins: I'd like to support both: specifying both `executorThreadCount` and `executorProvider`.

I'm not sure how acceptable it would be to allow Spring users to specify Google Cloud Pub/Sub stuff into the template. Perhaps we could make use of `PubSubAdmin` in #26 to help us create PubSub executors? Then we'd have Spring idioms only.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
joaoandremartins: In the future, we'd want these to change, right? In that case, should we leave them as variables as is, so we don't have to change back in the future?

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
artembilan: ??? That doesn't fit our Spring goal. We provide here some high level lightweight wrapping API to fit it into dependency injection idioms. But if there is some helpers API (builders ?) which can be used directly as beans and fully feasible in the inversion of control container that would be overhead to wrap that API to something else.
Right, if we look into the XML configuration, where builder API isn't executable, we need to wrap it into `FactoryBean` implementation. But here it is fully fine to let end-user to deal with the `InstantiatingExecutorProvider.newBuilder()` directly and inject the result into this `PubSubTemplate`. Having that I don't see reason to overhead the `PubSubTemplate` configuration with wider option and restricted one. Therefore my vote only for the `executorProvider` option.

meltsufin: How about just accepting a `java.util.concurrent.ScheduledExecutorService` and wrapping it into the `com.google.api.gax.grpc.ExecutorProvider`?

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
artembilan: No, @meltsufin 's concern that "magic value" must move to the constant:
```
public static final String  SOURCE_NAME = "spring";
...
private String sourceName = SOURCE_NAME;
```

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
artembilan: @meltsufin ,

what is the reason if that can be done externally and an appropriate `ExecutorProvider` will be injected here?

I don't mind about something default internally, but programming interface of this class (contract) should be as transparent as possible. And that way we sever responsibility and chase the flexibility of the dependency injection.

meltsufin: I was thinking more in terms of trying to hide the implementation details of `PubSubTemplate`. However, relying on DI to obtain an instance of the `com.google.api.gax.grpc.ExecutorProvider` is more flexible and transparent.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
joaoandremartins: Can you elaborate on the usage of Spring Boot here? I'm not sure the call source should be easily settable by users. Adding a setter here seems necessary since the header way seems non-feasible, at least with SI, but there is the risk of the user setting this value to something that won't be meaningful to us (like gibberish).

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
joaoandremartins: I'll need to look closely into this.
InstantiatingExecutorProvider.getExecutor() seems to create a new ScheduledThreadPoolExecutor(getExecutorThreadCount()), so maybe this code doesn't do what it's intended to do.

joaoandremartins:
Regarding the whole `ExecutorProvider` issue, it turned out to be a bit bigger than we thought and not really in the scope of this PR, so I opened #53 to address it.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
artembilan: I don't understand this concern since you still have this `.setClientLibHeader(this.sourceName, this.sourceVersion)` API in the GCP and it is fully public and can be specified by user application either way. 

Not sure why you treat Spring here as something different...
What we do here is just provide dependency injection options via setters.
They can be easily extracted from any Spring Boot configuration properties during this `PubSubTemplate` autoconfiguration.

If you want to protect yourself somehow from any vulnerability you should request something like signed certificate via SSL.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
```diff
@@ -45,36 +45,61 @@
...
 	@Override
 	public ListenableFuture<String> send(final String topic, Message message) {
-
-		Publisher publisher = this.publishers.computeIfAbsent(topic, s -> {
+		// Reuse existing publisher if one already exists.
+		Publisher publisher = this.publishers.computeIfAbsent(topic, key -> {
 			try {
-				return Publisher.defaultBuilder(TopicName.create(this.projectId, topic))
+				return Publisher.defaultBuilder(TopicName.create(this.projectId, key))
 						.setExecutorProvider(this.executorProvider)
```
meltsufin: @artembilan Correct me if I'm not understanding correctly, but the whole concern of creating a `Publisher` should be delegated to a Publisher factory, which in turn would depend on a `ChannelProvider`, and calling the `setClientLibHeader` would be the responsibility of whatever is creating the `ChannelProvider`. Basically, the whole creation of the `Publisher` should be done through dependency injection.

artembilan: H-m. I didn't think so far, but if that is the way you, guys, see it, let's go that way then!
We can merge as is right now and address `PublisherFactory` in the separate issue.

Yes, that was the reason why we chose `Factory` abstraction for Apache Kafka `Producer` and `Consumer`. Jut because they are volatile and non-inversion of control entities.

artembilan:
I can merge now.
Just let me know!

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
meltsufin: Noted `PublisherFactory` in the issue that @joaoandremartins created.

meltsufin:
Let's give @joaoandremartins a chance to merge it tomorrow morning. My only slight nit in this PR is that `sourceName` and `sourceVersion` don't have accessors, but they will probably go away in the future PR anyway.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
joaoandremartins: Say a non-Boot non-Integration plain Spring user calls `setSourceName` to "foo" and `setSourceVersion` to "bar". We'd see calls coming from a "foo.bar" client lib version on our side, which wouldn't be very helpful. One way this could be addressed is to, instead of setting `sourceName`, adding to it. In that case, the calls would be coming from "spring-foo", and we could always check on the first 6 characters to check that the call indeed came from Spring.

joaoandremartins:
@meltsufin sorry if I'm missing anything, but why would `sourceName` and `sourceVersion` go away?

I didn't think accessors would be necessary, but I can still add them if you want.

meltsufin:
`sourceName` and `sourceVersion` would just be handled somewhere in the `PublisherFactory`

PR #58, opened 2017-06-09
Adds manual Pub/Sub message acking mode

### LINKED ISSUE #8: Provide ACK control
Spring Messaging in general has the notion of a MessageListenerContainer, which allows configuration of different settings such as ACK mode, concurrent listeners.

We need to add same behavior here not only to be consistent with the Spring ecosystem, but to allow our users to control those parameters.

As usual, let's use the spring-kafka project (https://github.com/spring-projects/spring-kafka/tree/master/spring-kafka/src/main/java/org/springframework/kafka/listener) as source of inspiration, since it's a new take on a well stablished concept

joaoandremartins:
Users can now choose between getting the AckReplyConsumer and acking
the messages themselves, or having the channel adapters acking for
them.

Fixes #8.

### DIFF
--- a/spring-integration-gcp/src/main/java/org/springframework/integration/gcp/inbound/PubSubInboundChannelAdapter.java
+++ b/spring-integration-gcp/src/main/java/org/springframework/integration/gcp/inbound/PubSubInboundChannelAdapter.java
@@ -83,8 +83,15 @@ public class PubSubInboundChannelAdapter extends MessageProducerSupport {
 		}
 
 
-		sendMessage(getMessagingTemplate().getMessageConverter()
-				.toMessage(message.getData(), new MessageHeaders(messageHeaders)));
+		try {
+			sendMessage(getMessagingTemplate().getMessageConverter()
+					.toMessage(message.getData(), new MessageHeaders(messageHeaders)));
+		} catch (RuntimeException re) {
+			if (this.ackMode == AckMode.AUTO) {
+				consumer.nack();
+			}
+			throw re;
+		}
 
 		if (this.ackMode == AckMode.AUTO) {
 			consumer.ack();
--- a/spring-integration-gcp/src/main/java/org/springframework/integration/gcp/inbound/PubSubInboundChannelAdapter.java
+++ b/spring-integration-gcp/src/main/java/org/springframework/integration/gcp/inbound/PubSubInboundChannelAdapter.java
@@ -30,6 +30,7 @@ import org.springframework.cloud.gcp.pubsub.support.GcpHeaders;
 import org.springframework.integration.endpoint.MessageProducerSupport;
 import org.springframework.integration.gcp.AckMode;
 import org.springframework.messaging.MessageHeaders;
+import org.springframework.util.Assert;
 
 /**
  * Converts from GCP Pub/Sub message to Spring message and sends the Spring message to the
@@ -104,6 +105,7 @@ public class PubSubInboundChannelAdapter extends MessageProducerSupport {
 	}
 
 	public void setAckMode(AckMode ackMode) {
+		Assert.notNull(ackMode, "The acknowledgement mode can't be null.");
 		this.ackMode = ackMode;
 	}
 }
--- a/spring-integration-gcp/src/main/java/org/springframework/integration/gcp/inbound/PubSubInboundChannelAdapter.java
+++ b/spring-integration-gcp/src/main/java/org/springframework/integration/gcp/inbound/PubSubInboundChannelAdapter.java
@@ -28,6 +28,7 @@ import com.google.pubsub.v1.SubscriptionName;
 
 import org.springframework.cloud.gcp.pubsub.support.GcpHeaders;
 import org.springframework.integration.endpoint.MessageProducerSupport;
+import org.springframework.integration.gcp.AckMode;
 import org.springframework.messaging.MessageHeaders;
 
 /**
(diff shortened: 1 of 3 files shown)


joaoandremartins:
We thought of nacking messages if `sendMessage()` throws a `RuntimeException`. However, I find the behaviour becomes a bit too erratic, since it depends on whether or not an `errorChannel` is associated with `PubSubMessageHandler`. If there is an `errorChannel`, then `RuntimeExceptions` will go there and the message won't get nacked.

An `errorChannel` might also be shared among different channels, so if we were thinking of nacking when `RuntimeExceptions` are caught there, it's possible to nack messages based on errors thrown from other unrelated channels.

Auto nacking based on user processing failure also appears to be not possible, since we don't have access to exceptions thrown in that context at the `PubSubMessageHandler` level, and users would be better off advised to use manual acking instead of auto.

The creation of an `AckMode` enum was also inspired on other examples (Kafka, AMQP, etc.). We currently only have two states for acking, so this could be `boolean`, but we might want more states in the future?

meltsufin:
@artembilan @viniciusccarvalho Can you guys take a quick look at this as well, please? Thanks!

artembilan:
@joaoandremartins , 

thank you for such a great investigation in the (n)ack area!
Let me try to answer and elaborate on some your concerns.

Yes, we should worry about any other possible ack modes. We really might want more in the future.

Answering to all other problems at once, I'd suggest to investigate recently introduced:
```
/**
 * Build an error message for the exception and message using the configured
 * {@link ErrorMessageStrategy}.
 * @param message the message.
 * @param exception the exception.
 * @return the error message.
 * @since 4.3.10
 */
protected final ErrorMessage buildErrorMessage(Message<?> message, RuntimeException exception) {
	return this.errorMessageStrategy.buildErrorMessage(exception,
			getErrorMessageAttributes(message));
}
```
I think in case of `MANUAL` ack we still have `GcpHeaders.ACKNOWLEDGEMENT` header in the `MessagingException.failedMessage` and can rely on it for nacking (or even acking if user prefers).

In case of `AUTO` I think we should perform `nack()` on the `ThreadLocal<AckReplyConsumer>` from that `getErrorMessageAttributes()` implementation.

We might consider similar improvement in the Kafka and AMQP as well.

OTOH I think we are already in stage when we should introduce a general `Acknowledgement` contract:
```
interface Acknowledgement {
     
     void ack();

     void nack();

}
```

and provide its implementation in target modules (here, Kafka, AMQP, Kinesis etc.).
I think you have already mention such a lack of an abstraction to let to use Spring API more clearly.

This way we will be able to have more powerful SCSt application when we switch Binders without breaking changes.

WDYT, @garyrussell , @mbogoevici ?

spring-integration-gcp/src/main/java/org/springframework/integration/gcp/inbound/PubSubInboundChannelAdapter.java
```diff
@@ -85,4 +98,12 @@ protected void doStop() {
 
 		super.doStop();
 	}
+
+	public AckMode getAckMode() {
+		return this.ackMode;
+	}
+
+	public void setAckMode(AckMode ackMode) {
+		this.ackMode = ackMode;
```
artembilan: Should be with the `Assert.notNull()` otherwise you have an `NPE` on the `this.ackMode == AckMode.MANUAL` condition above.

meltsufin:
FYI @joaoandremartins is on vacation this week. So, he might not be responding until next week. :)

garyrussell:
>If there is an errorChannel, then RuntimeExceptions will go there and the message won't get nacked.

That is a consistent pattern across all Integration adapters.

It's like a try/catch in Java.

If the error flow "consumes" the exception (exits normally), the message delivery is considered successful and ack'd; if the error flow re-throws (or throws another) exception, then the container ack behavior is performed - depending on the container configuration.

For MANUAL-style acks, the user is always responsible for the ack/nack either in mainline code or on the error flow.

@artembilan I have no objection to introducing a common interface across all technologies for manual ack/nack.

spring-integration-gcp/src/main/java/org/springframework/integration/gcp/inbound/PubSubInboundChannelAdapter.java
joaoandremartins: I think the condition would not evaluate to true in that scenario, without the `NPE`, but I added the guard anyways.

joaoandremartins:
@artembilan I suggest we handle auto-nacking in a new PR. WDYT?

artembilan:
>  I suggest we handle auto-nacking in a new PR.

Well, I don't think so. 

According proper Gary's explanation we just need to add `try...catch` around `sendMessage()`. Nack and re-throw exception.

artembilan:
I have just approved PR, but I can't confirm that it works well because I didn't pull it locally for testing or at least verifying the Checkstyle.
First: we definitely need tests in any PR. That may become a bad habit to merge everything without testing.
Second: we can live without CI any more. Having this style of review/merging we can't be sure what's going on in project.

Thanks

meltsufin:
@artembilan Completely agree that tests should be included with new code in the same PR. Regarding the CI, should we just run on Travis for now?

artembilan:
> should we just run on Travis for now?

I think so, but we need somehow who has admin rights here in project and knows how to do that.

/CC @viniciusccarvalho , @markfisher

Issue #46, opened 2017-06-05
Support injection of Google Storage objects as Resource


meltsufin:
The `spring-cloud-gcp-starter-storage` already supports loading `gs://...` resources using `DefaultResourceLoader`, but direct injection using `@Value` does not work.

```java
@Value("gs://<bucket>/<object>")
private Resource gcsResource;
```

Currently, for some reason it tries to resolve the resource via ServletContext.
```
java.io.FileNotFoundException: Could not open ServletContext resource [/gs://<bucket>/<object>]
	at org.springframework.web.context.support.ServletContextResource.getInputStream(ServletContextResource.java:141) ~[spring-web-4.3.7.RELEASE.jar:4.3.7.RELEASE]
	at com.example.HelloController.readGcs(BootDemoApplication.java:164) ~[classes/:na]
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method) ~[na:1.8.0_73]
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62) ~[na:1.8.0_73]
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43) ~[na:1.8.0_73]
	at java.lang.reflect.Method.invoke(Method.java:497) ~[na:1.8.0_73]
```

artembilan:
Well, actually it works. See `GoogleStorageTests`:
```
@Value("gs://test-spring/images/spring.png")
private Resource remoteResource;
...
@Configuration
static class GoogleStorageTestsConfiguration implements ResourceLoaderAware{

	private ResourceLoader defaultResourceLoader;
        ...
        @Bean
	public GoogleStorageProtocolResolver googleStorageProtocolResolver(Storage storage) {
		return new GoogleStorageProtocolResolver(this.defaultResourceLoader, storage);
	}
```
The main trick here is exactly that `GoogleStorageProtocolResolver` bean definition.
The `StorageAutoConfiguration` does the stuff for us when we use Boot and have `spring-cloud-gcp-starter-storage` in the classpath. Otherwise we have to declare that `GoogleStorageProtocolResolver` bean manually in one of our configuration.

Does it make sense to you?

meltsufin:
[...]

PR #59, opened 2017-06-09
GH-46: Fix GoogleStorageProtocolResolver handling

artembilan:
Fixes spring-cloud/spring-cloud-gcp#46

Since `@Value("gs://<bucket>/<file>")` maybe performed pretty early
in the application context lifecycle (e.g. in the `@Configuration`),
it is impossible to make `GoogleStorageProtocolResolver` and let it
to be registered with the `DefaultResourceLoader`

* Introduce `GoogleStorageProtocolResolverBeanFactoryPostProcessor`
to register `GoogleStorageProtocolResolver` into the `DefaultResourceLoader`
* Make `GoogleStorageProtocolResolver` as package protected class do not
encourage end-users to try to register it as a bean
* Add `GoogleStorageAutoConfigurationTests` to check the functionality
against full Spring Boot Web application
* Polishing `GoogleStorageTests` for unnecessary mocking
* Polishing `StorageAutoConfiguration`  to deal only with the `Storage`
bean and `@Import(GoogleStorageProtocolResolverBeanFactoryPostProcessor.class)`


spring-cloud-gcp-starters/spring-cloud-gcp-starter-storage/pom.xml
```diff
@@ -26,5 +26,10 @@
 			<groupId>org.springframework.cloud</groupId>
 			<artifactId>spring-cloud-gcp-storage</artifactId>
 		</dependency>
```
viniciusccarvalho: @artembilan I don't understand why do we need the web dependency here, was it just for a test as a resource server? Was it the only way to test? I don't oppose this, I just find it a bit peculiar the need for a servlet engine just for testing.

artembilan: See the issue. We have a problem exactly in the servlet-based application context

viniciusccarvalho: So the Servlet application context was failing then? Does it have a different mechanism to load resources?

artembilan: Well, I don't if that is the point, but I decided to restore the whole scenario from the issue. But, yeah, sometimes we may not meet issue until we make the test web-based.

The problem here is fundamental: having the `ProtocolResolver` as a bean is too late for resource resolving.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageProtocolResolver.java
```diff
@@ -17,51 +17,35 @@
...
-import org.springframework.beans.factory.InitializingBean;
-import org.springframework.core.io.DefaultResourceLoader;
 import org.springframework.core.io.ProtocolResolver;
 import org.springframework.core.io.Resource;
 import org.springframework.core.io.ResourceLoader;
 import org.springframework.util.Assert;
 
 /**
+ * A {@link ProtocolResolver} implementation for the {@code gs://} protocol.
+ *
```
meltsufin: Checkstyle doesn't like the trailing space at the end of this line.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageProtocolResolverBeanFactoryPostProcessor.java
```diff
@@ -0,0 +1,77 @@
...
+
+/**
+ * A {@link BeanFactoryPostProcessor} implementation to register
+ * {@link GoogleStorageProtocolResolver} with the {@link DefaultResourceLoader}. The
+ * registration is performed only if {@link Storage} bean is declared in the application
+ * context.
+ * <p>
+ * This class must be used as a {@code static @Bean} method definition or just
+ * {@code @Import}ed on some {@code @Configuration} class.
+ *
```
meltsufin: Again checkstyle trailing white space issue. Maybe we should adjust the rule not to apply to comments.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-storage/src/main/java/org/springframework/cloud/gcp/storage/autoconfig/StorageAutoConfiguration.java
```diff
@@ -22,35 +22,33 @@
...
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
-import org.springframework.core.io.ResourceLoader;
+import org.springframework.context.annotation.Import;
 
 /**
+ * An auto-configuration for Google {@link Storage} bean definition. Also it
+ * {@link Import} a {@link GoogleStorageProtocolResolverBeanFactoryPostProcessor} to
+ * register {@code GoogleStorageProtocolResolver} with the {@code DefaultResourceLoader}.
+ *
```
meltsufin: trailing space

spring-cloud-gcp-starters/spring-cloud-gcp-starter-storage/src/main/java/org/springframework/cloud/gcp/storage/autoconfig/StorageAutoConfiguration.java
```diff
@@ -22,35 +22,33 @@
...
+import org.springframework.context.annotation.Import;
 
 /**
+ * An auto-configuration for Google {@link Storage} bean definition. Also it
+ * {@link Import} a {@link GoogleStorageProtocolResolverBeanFactoryPostProcessor} to
+ * register {@code GoogleStorageProtocolResolver} with the {@code DefaultResourceLoader}.
+ * 
  * @author Vinicius Carvalho
+ * @author Artem Bilan
+ *
```
meltsufin: trailing space

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageProtocolResolver.java
```diff
@@ -17,51 +17,35 @@
...
-		}
-	}
-
-	@Override
 	public Resource resolve(String location, ResourceLoader resourceLoader) {
-		if (location.startsWith("gs://")) {
+		if (location.startsWith(PROTOCOL)) {
 			return new GoogleStorageResource(this.storage, location);
 		}
 		return null;
```
joaoandremartins: Not for this PR, but eventually I'd like us to start looking into using Java 8's `Optional` to avoid returning nulls, and explicitly unwrapping results.

I've been using `Optionals` in other projects and I found it usually improves code quality.

meltsufin:
@artembilan ping

artembilan:
@meltsufin ,

sorry, we are busy on the team meeting this week.
Will come back from next Monday.

meltsufin:
@artembilan Sounds good, no worries. Thanks!

artembilan:
> Can you add the null check there?

I don't see that problem, but when I have `GcpContextAutoConfiguration` on board without any options I end up with the:
```
Caused by: java.io.IOException: The Application Default Credentials are not available. They are available if running in Google Compute Engine. Otherwise, the environment variable GOOGLE_APPLICATION_CREDENTIALS must be defined pointing to a file defining the credentials. See https://developers.google.com/accounts/docs/application-default-credentials for more information.
	at com.google.auth.oauth2.DefaultCredentialsProvider.getDefaultCredentials(DefaultCredentialsProvider.java:109)
	at com.google.auth.oauth2.GoogleCredentials.getApplicationDefault(GoogleCredentials.java:103)
	at com.google.auth.oauth2.GoogleCredentials.getApplicationDefault(GoogleCredentials.java:76)
	at org.springframework.cloud.gcp.core.autoconfig.GcpContextAutoConfiguration.googleCredentials(GcpContextAutoConfiguration.java:51)
```
which clearly doesn't have any `NPE` issue related.

I think that logic in the `GcpContextAutoConfiguration` is good enough and doesn't need any improvements. Although I will add `@ConditionalOnMissingBean` to the `googleCredentials()` since we really should treat end-user preference as his/her own choice.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageProtocolResolver.java
artembilan: That's not correct.

The `DefaultResourceLoader` has a logic like:
```
Resource resource = protocolResolver.resolve(location, this);
if (resource != null) {
	return resource;
}
```
So, it really relies on the fact of `null` from the delegate.

OTOH the `Optional` is an object (unlike simple `if...else`). It requires some memory to allocate and afterward some GC effort to clean up afterwards.

meltsufin:
@artembilan You might be getting that exception because your Cloud SDK CLI (gcloud) is not initialized. Try running `gcloud init` first, which will "provide" the local default credentials.

For the `NPE`, are you defining any of the `spring.cloud.gcp.*` properties in your `application.properties`?

artembilan:
> Cloud SDK CLI (gcloud) is not initialized.

That's correct. But I'm not interested in that in my test-case.

>  are you defining any of the spring.cloud.gcp.* properties in your application.properties?

No, I'm not. But that shouldn't be a problem at all. Any `@ConfigurationProperties` are beans anyway, independently of the provided (or not) `application.properties`. That's one of the main Spring Boot points - provide some default configuration when no any options from end-user.

If you have an `NPE` somewhere here in my PR, would be great to share that as a test-case. 
Otherwise, please, raise a separate GH issue with more thorough description and more info to reproduce.
Thanks

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageProtocolResolver.java
artembilan: That's not a Checkstyle issue. The real problem that Eclipse formatter doesn't have ability to remove trailing space. Native Eclipse requires special plugin on the matter: http://andrei.gmxhome.de/anyedit/

Not sure what we can do to unless manually switch to the IDEA formatter after having Checkstyle report.

OTOH if we all use IDEA, there might not be reason to use this Eclipse formatter but configure the IDEA native to closer style.

(I'll fix all whitespaces issue in this PR eventually)

artembilan:
@viniciusccarvalho ,

Can you tell, please, what is that Codacy, why it is based on your account (viniciusccarvalho/spring-cloud-gcp) and why it doesn't like our `import` code style?
```

Extra separation in import group before 'org.junit.Test'
25

import org.junit.Test;
```
That sounds suspicious...

Thanks


@meltsufin , thanks for the review. Does your confirmation mean that I can merge PR? (According to mentioned policy before, I mean)

meltsufin:
Yep, you can "Squash and merge". I see that "Rebase and Merge" is still available as an option too. Someone with admin access should disable it. @markfisher

Issue #53, opened 2017-06-07
Allow ExecutorProvider to be passed in to PubSubTemplate


joaoandremartins:
In #50, some suggested polishing changes surfaced a bigger issue. We might want to either use the default `ExecutorProvider` (with 4 threads by default) or, instead, allow the user to specify their own `ExecutorProvider` implementation.

We should look into injecting the `ExecutorProvider` into the `PubSubTemplate` which, right now, isn't possible in SI, since `PubSubMessageHandler` uses `new PubSubTemplate()` instead of injecting it.

Another issue is that re-using the same `ExecutorProvider` does not guarantee that only one `Executor` is generated, thereby reducing the number of created threads, which was the motivation behind the reuse.

meltsufin:
As mentioned in the comments on the PR #50. We should probably just create a `PublisherFactory` abstraction and make `PubSubTemplate` depend on it.

joaoandremartins:
The same should probably also apply for the `Subscriber` in `PubSubInboundChannelAdapter`.

artembilan:
The `PubsubTemplate` is based now on the `PublisherFactory`, where the `DefaultPublisherFactory` is supplied with the `ExecutorProvider`, so, I consider this issue as fixed.
(Can't find the commit, but fix is in the place any way...)

joaoandremartins:
FTR the fix is #61 (GitHub doesn't appear to understand "Fixes #x, #y" notation)

PR #61, opened 2017-06-29
Adds missing unit tests

### LINKED ISSUE #45: Investigate and add missing unit tests
Now that we have the project base and some code already in, we should look into missing unit tests and add them as soon as possible.

joaoandremartins:
Also creates PublisherFactory and SubscriberFactory. Polishes code.

Fixes #45, #53


joaoandremartins:
Few notes on this PR:
1) Unit tests for `PubSubInboundChannelAdapter` started getting too complicated, and I eventually dropped them. Reason is that `Subscriber` won't take in the right mocks to simulate an incoming message (I tried and gave up). The alternative is to run a local Google Cloud Pub/Sub emulator, which is prohibitive due to the requirement on Cloud SDK. Maybe you've got any hints for other avenues worth exploring?

2) `DefaultPublisherFactory` will create duplicate `Publishers` for the same topic, one per different `sourceVersion` (e.g., `"spring"`, `"spring-boot"`, `"spring-integration"`, etc.). It's still unclear whether we can use versions like this, that aren't semantically versioned, so some work around specifying `sourceVersion` might need to be rolled back.

3) I'm now wondering if `PubSubMessageHandler` should depend on a `PublisherFactory`, the same way `PubSubInboundChannelAdapter` depends on a `SubscriberFactory`.

joaoandremartins:
Just added a commit to get rid of `clientLibVersions` like `"integration\0.0.1"`, and stick only to semver `"0.0.1"`. That allowed us to make `PubSubMessageHandler` depend solely on `PubSubTemplate`.

For Spring Boot users, all should be fine and dandy, but I'm really worried about non-Boot users, and the amount of manual configuration they will need to do to set up a `PublisherFactory` and `SubscriberFactory`. Maybe we should provide alternative constructors for users of those classes, like `PubSubMessageHandler`, or create a helper to create them for us.

Also, I'm using the current import file but, when I optimize imports, I'm getting e.g., `import org.mockito.Mockito.*`, which is a bit annoying.

joaoandremartins:
I'm requesting the google-cloud-pubsub team for a Subscriber that takes mocks which let us simulate an incoming message in https://github.com/GoogleCloudPlatform/google-cloud-java/issues/2205.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/DefaultSubscriberFactory.java
```diff
@@ -0,0 +1,58 @@
...
+		this.credentials = credentials;
+		this.executorProvider = executorProvider;
+	}
+
+	@Override
+	public Subscriber getSubscriber(String subscriptionName, MessageReceiver receiver) {
+		return Subscriber.defaultBuilder(
+				SubscriptionName.create(this.projectId, subscriptionName), receiver)
+				.setChannelProvider(
+						SubscriptionAdminSettings.defaultChannelProviderBuilder()
```
meltsufin: `ChannelProvider` should be a constructor argument just like in the `DefaultPublisherFactory`. You also wouldn't need the `GoogleCredentials` to be passed in.

spring-cloud-gcp-pubsub/src/test/java/org/springframework/cloud/gcp/pubsub/converters/SimpleMessageConverterTest.java
```diff
@@ -39,10 +38,16 @@
...
+
+		Map<String, Object> headers = new HashMap<>();
+		headers.put("h1", "value");
+		headers.put("h2", 23);
+		Message<?> message = new GenericMessage<Object>("test payload", headers);
 		PubsubMessage pubsubMessage = (PubsubMessage) converter.fromMessage(message,
 				PubsubMessage.class);
 		assertEquals("test payload", pubsubMessage.getData().toStringUtf8());
+		assertEquals("value", pubsubMessage.getAttributesOrDefault("h1", "def"));
+		assertEquals("23", pubsubMessage.getAttributesOrDefault("h2", "def"));
```
meltsufin: Maybe also verify the count of attributes if possible?

spring-cloud-gcp-starters/spring-cloud-gcp-starter-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/autoconfig/GcpPubsubAutoConfiguration.java
```diff
@@ -0,0 +1,96 @@
...
+/**
+ * @author João André Martins
+ */
+@Configuration
+@ComponentScan(basePackageClasses =
+		org.springframework.cloud.gcp.core.autoconfig.GcpContextAutoConfiguration.class)
+public class GcpPubsubAutoConfiguration {
+
+	public static final String DEFAULT_SOURCE_NAME = "spring";
+	public static final int DEFAULT_EXECUTOR_THREADS = 4;
```
meltsufin: I think this should be configurable through a property. Also, we should fall back to the default defined in the client library.

spring-cloud-gcp-pubsub/src/test/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplateTest.java
```diff
@@ -0,0 +1,129 @@
...
+
+		this.siMessage = new GenericMessage<>("testBody", headers);
+	}
+
+	@Test
+	public void testSend() throws ExecutionException, InterruptedException {
+		ListenableFuture<String> future = this.pubSubTemplate.send("testTopic", this.siMessage);
+		this.settableApiFuture.set("result");
+
+		assertEquals("result", future.get());
```
artembilan: 1. We prefer to have `...Tests` names for test classes. You really have many of them here anyway..
2. You have a race condition. The `mockPublisher.publish()` may return early than you set value to the `settableApiFuture`. So, move the value setting *before* publishing.

spring-cloud-gcp-pubsub/src/test/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplateTest.java
```diff
@@ -0,0 +1,129 @@
...
+	}
+
+	@Test
+	public void testSend_onFailure() {
+		ListenableFuture<String> future = this.pubSubTemplate.send("testTopic", this.siMessage);
+		this.settableApiFuture.setException(new Exception("future failed."));
+
+		try {
+			future.get();
+			TestCase.fail("Test should fail.");
```
artembilan: Please, don't use `junit.framework` pacakge. Don't remember why, but `fail()` can simply be imported statically from the `Assert` class.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/DefaultSubscriberFactory.java
```diff
@@ -0,0 +1,58 @@
...
+			ExecutorProvider executorProvider) {
+		this.projectId = projectId;
+		this.credentials = credentials;
+		this.executorProvider = executorProvider;
+	}
+
+	@Override
+	public Subscriber getSubscriber(String subscriptionName, MessageReceiver receiver) {
+		return Subscriber.defaultBuilder(
+				SubscriptionName.create(this.projectId, subscriptionName), receiver)
```
artembilan: Have just noticed and can't calculate that from docs: https://cloud.google.com/pubsub/docs/subscriber.
So, the question is: how do we determine to which topic do we subscribe?
Is that what `subscriptionName` for?

spring-cloud-gcp-starters/spring-cloud-gcp-starter-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/autoconfig/GcpPubsubAutoConfiguration.java
```diff
@@ -0,0 +1,96 @@
...
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.ComponentScan;
+import org.springframework.context.annotation.Configuration;
+
+/**
+ * @author João André Martins
+ */
+@Configuration
+@ComponentScan(basePackageClasses =
+		org.springframework.cloud.gcp.core.autoconfig.GcpContextAutoConfiguration.class)
```
artembilan: No, `@ComponentScan`.
`@AutoConfigureAfter(GcpContextAutoConfiguration.class)` instead.
Plus we need to have `META-INF/spring.factories` with an appropriate entry.
See similar for aforementioned `GcpContextAutoConfiguration`.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/autoconfig/GcpPubsubAutoConfiguration.java
```diff
@@ -0,0 +1,96 @@
...
+public class GcpPubsubAutoConfiguration {
+
+	public static final String DEFAULT_SOURCE_NAME = "spring";
+	public static final int DEFAULT_EXECUTOR_THREADS = 4;
+
+	@Bean
+	@ConditionalOnMissingBean
+	public ExecutorProvider executorProviderForPublisher() {
```
artembilan: That's not good bean name. Should be just noun: `publisherExecutorProvider`

spring-cloud-gcp-starters/spring-cloud-gcp-starter-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/autoconfig/GcpPubsubAutoConfiguration.java
```diff
@@ -0,0 +1,96 @@
...
+
+	@Bean
+	@ConditionalOnMissingBean
+	public ExecutorProvider executorProviderForSubscriber() {
+		return FixedExecutorProvider.create(
+				Executors.newScheduledThreadPool(DEFAULT_EXECUTOR_THREADS));
+	}
+
+	@Bean
+	@ConditionalOnClass({GcpProperties.class, GoogleCredentials.class})
```
artembilan: This condition is somehow non-relevant to the bean it is declared for.
If we `@AutoConfigureAfter(GcpContextAutoConfiguration.class)`, we definitely have both classes already.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/autoconfig/GcpPubsubAutoConfiguration.java
```diff
@@ -0,0 +1,96 @@
...
+	public PubSubTemplate pubsubTemplate(PublisherFactory publisherFactory) {
+		return new PubSubTemplate(publisherFactory);
+	}
+
+	@Bean
+	@ConditionalOnClass({GcpProperties.class, GoogleCredentials.class})
+	@ConditionalOnMissingBean
+	public SubscriberFactory defaultSubscriberFactory(GcpProperties gcpProperties,
+			GoogleCredentials credentials,
+			@Qualifier("executorProviderForSubscriber") ExecutorProvider executorProvider) {
```
artembilan: You can just use method directly instead of arguments:
```
return new DefaultSubscriberFactory(
	gcpProperties.getProjectId(), credentials, executorProviderForSubscriber());
```

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/DefaultSubscriberFactory.java
joaoandremartins: A `Subscription` is associated to a topic at subscription build time. You can create a new subscription from the Developer's Console (like I did) or programatically, through `SubscriptionAdminClient` (we should provide functionality similar to this on [PubsubAdmin](https://github.com/spring-cloud/spring-cloud-gcp/issues/26)).

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/DefaultSubscriberFactory.java
joaoandremartins: I just noticed we never passed `clientLibName`/`Version` to the `Subscriber`, which we probably should, although I don't know how that would show on the pull method.

An issue with wildly accepting `ChannelProviders` is that users might provide new ones that don't contain the correct `clientLibName`/`Version`. However, this issue exists regardless, since users can also just provide their own `PublisherFactory`/`SubscriberFactory`, so this might be acceptable after all.

spring-cloud-gcp-pubsub/src/test/java/org/springframework/cloud/gcp/pubsub/converters/SimpleMessageConverterTest.java
joaoandremartins: SI adds ID and timestamp attributes to the SI message. As long as they don't provide any more/less, we should be fine with an attribute count check here.

But that brings up a question: I wonder if we should be sending those two generated attributes to Pubsub?

spring-cloud-gcp-starters/spring-cloud-gcp-starter-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/autoconfig/GcpPubsubAutoConfiguration.java
joaoandremartins: FixedExecutorProvider doesn't specify any default for threads. I copied this value from InstantiatingExecutorProvider.DEFAULT_EXECUTOR_THREADS.

As for the property, should this be called `spring.cloud.gcp.pubsub.publisher.executorThreads`?

spring-cloud-gcp-starters/spring-cloud-gcp-starter-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/autoconfig/GcpPubsubAutoConfiguration.java
joaoandremartins: `@ComponentScan` is required so that this class knows about the injected `GoogleCredentials` and `GcpProperties` beans. Is there a better way to do this?

spring-cloud-gcp-starters/spring-cloud-gcp-starter-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/autoconfig/GcpPubsubAutoConfiguration.java
joaoandremartins: But then, without injection, the user wouldn't be able to specify their own executors.

Even though having to specify them with a method named `subscriberExecutorProvider` is also kind of weird?

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/DefaultSubscriberFactory.java
artembilan: Cool! Can't we leverage that `SubscriptionAdminClient`  functionality directly here in the `DefaultSubscriberFactory`? 
Or does that require special permissions, since it is admin?

spring-cloud-gcp-pubsub/src/test/java/org/springframework/cloud/gcp/pubsub/converters/SimpleMessageConverterTest.java
artembilan: That is good point! And right, we don't send them in other protocols, unless end-user requests that via `HeaderMapper`.

joaoandremartins:
@artembilan 
1) local emulator: we'll probably have to resort to something similar for integration tests. But for unit tests, spinning one up is probably too heavy, and it could be prohibitive for users who want to build this from source and don't have Cloud SDK installed.

2) client name/version: this is just a string like "spring\[lib semantic version]" that we send in a header, so we can quantify the number of RPC calls originating from our Spring integrations. You're right that there are ways in which a user can override our ChannelProvider with theirs without this data, but I don't that's very likely to happen, and the amount of insight from the other users who didn't build the libs on their own is still very meaningful.

3) helpers: I'm not sure if having an abstraction such as SpringChannelProvider would be useful here since the concept of channel provider is specific to the Google Cloud Pubsub API (that is, they couldn't be easily reused by other Spring components/integrations). I still want to take a look at using helpers to aid non-Boot users, but this could probably be left to another PR since this one is already pretty sizeable?

spring-cloud-gcp-starters/spring-cloud-gcp-starter-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/autoconfig/GcpPubsubAutoConfiguration.java
artembilan: It is. It is called Spring Boot auto-configuration. I explained you how to reach it in your case.
We definitely don't need `@ComponentScan`. Otherwise you have missed something else...

spring-cloud-gcp-starters/spring-cloud-gcp-starter-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/autoconfig/GcpPubsubAutoConfiguration.java
artembilan: Oh! I missed that fact that you auto-configure it. Never mind. Sorry for the noise.
So, yes `@ConfigurationProperties` on the matter would be useful:
```
spring.cloud.gcp.pubsub.publisher.executorThreads
spring.cloud.gcp.pubsub.subscriber.executorThreads
```
as you mentioned above.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/autoconfig/GcpPubsubAutoConfiguration.java
```diff
@@ -48,49 +51,55 @@
 	public static final int DEFAULT_EXECUTOR_THREADS = 4;
 
 	@Bean
-	@ConditionalOnMissingBean
-	public ExecutorProvider executorProviderForPublisher() {
+	public ExecutorProvider publisherExecutorProvider() {
```
artembilan: Sorry, I misled you a bit with method reference dependency inject.
So, right, the use may provide his/her own `ExecutorProvider` on the matter.
So, we should keep `@ConditionalOnMissingBean` and use argument injection as you had before.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/autoconfig/GcpPubsubAutoConfiguration.java
```diff
@@ -48,49 +51,55 @@
...
-	@ConditionalOnMissingBean
-	public ExecutorProvider executorProviderForSubscriber() {
+	public ExecutorProvider subscriberExecutorProvider() {
 		return FixedExecutorProvider.create(
 				Executors.newScheduledThreadPool(DEFAULT_EXECUTOR_THREADS));
 	}
 
 	@Bean
-	@ConditionalOnClass({GcpProperties.class, GoogleCredentials.class})
+	public ChannelProvider subscriberChannelProvider(GoogleCredentials credentials) {
```
artembilan: I think `ChannelProvider`s can be specified by end-user as well, therefore `@ConditionalOnMissingBean`, too.

artembilan:
@joaoandremartins ,

1. Well, that happens in many other our projects anyway. We run some kind of integration tests together with the unit tests and guard integration tests with mentioned `@Rule`s. I have ran RabbitMQ, Kafka, MongoDB, Redis, MySQL etc. all the time on my machine. So, when I develop the project and run test I'm sure that everything works as expected because all those services. If there are no services, like in our CI env, the integration tests are ignored. The same I suggest here with the mentioned emulator. If some other developer runs our project without  Cloud SDK installed, he just won't have problems because of JUnit Rule.

2. OK. Fine with me to keep it as is.

3. `SpringChannelProvider` is not an abstraction. It is wrapper around target `ChannelProvider` (`SpringChannelProvider implements ChannelProvider`) with particular clientLib properties injection.
Really, let's deffer this discussion to the separate issue.

(This hope story about `clientLib properties` is annoying me a bit. We spend too much time around the fact that it can be simply overridden...)

spring-cloud-gcp-starters/spring-cloud-gcp-starter-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/autoconfig/GcpPubsubAutoConfiguration.java
meltsufin: So, why duplicate this constant here and try to keep them in sync?

spring-cloud-gcp-pubsub/src/test/java/org/springframework/cloud/gcp/pubsub/converters/SimpleMessageConverterTest.java
meltsufin: Should we create a separate issue to stop adding ID and timestamp to pubsub message headers?

artembilan: I think so.
And any `GcpHeaders` should not be included on sending part.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/autoconfig/GcpPubsubAutoConfiguration.java
joaoandremartins: It's not externally visible. https://github.com/googleapis/gax-java/blob/master/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingExecutorProvider.java#L46

meltsufin: I see, but can you just use the `InstantiatingExecutorProvider` to create the executor and utilize the default indirectly?

joaoandremartins: We can't, because that would defeat the purpose of having a single shared executor.
I'm not using `InstantiatingExecutorProvider` here on purpose, but rather `FixedExecutorProvider`, so that a new executor is not created for each `Publisher`/`Subscriber`.

spring-cloud-gcp-pubsub/src/test/java/org/springframework/cloud/gcp/pubsub/converters/SimpleMessageConverterTest.java
joaoandremartins: Opened https://github.com/spring-cloud/spring-cloud-gcp/issues/63

spring-cloud-gcp-starters/spring-cloud-gcp-starter-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/autoconfig/GcpPubsubAutoConfiguration.java
meltsufin: I see. I guess another, slightly ugly alternative, would be `InstantiatingExecutorProvider.newBuilder().getExecutorThreadCount()`. I don't know if it's worth it though. :)

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/DefaultSubscriberFactory.java
joaoandremartins: Indeed, the used credentials would need to have certain permissions for things like subscription creation to work.
But, before getting there, how did you idealize subscription creation would work here?

spring-cloud-gcp-starters/spring-cloud-gcp-starter-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/autoconfig/GcpPubsubAutoConfiguration.java
joaoandremartins: I added `@AutoConfigureAfter`, as well as the META-INF/spring.factories file, if that's what you mean, but IDEA still warns me that no bean of that type is being injected.
`@ComponentScan` fixes the issue. Let me know if I'm missing anything.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/DefaultSubscriberFactory.java
meltsufin: A Subscription is an important concept in Pub/Sub and is managed similarly to a Topic. A subscription can be bound to multiple competing clients, where only one will receive the message. Whereas, if you want broadcast behavior, each client has to bind to a separate subscription.

I don't think we should automatically creating subscription because those survive beyond the execution of the application. However, it might make sense to maybe autoconfigure a `SubscriptionAdminClient`.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/autoconfig/GcpPubsubAutoConfiguration.java
joaoandremartins: Interestingly, my test app works without @ComponentScan. It seems that only the IDE requires it for some reason. I'll remove it.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/autoconfig/GcpPubsubAutoConfiguration.java
joaoandremartins: Here I was thinking of

```
@Value("${spring.cloud.gcp.pubsub.subscriber.executorThreads : 4}")
private int subscriberExecutorThreads;
```

Can you clarify why `@ConfigurationProperties` would be better? Also, wouldn't the namespace clash with the existing `@ConfigurationProperties("spring.cloud.gcp")`?

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/DefaultSubscriberFactory.java
artembilan: Well, correct, 
I think not here, but in the `PubSubInboundChannelAdapter` with an appropriate `createSubscription=true/false` option.
But that really should be fully different story. Not for this PR...

spring-cloud-gcp-starters/spring-cloud-gcp-starter-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/autoconfig/GcpPubsubAutoConfiguration.java
artembilan: >  IDEA still warns me 

Why should end-user care about problems of our IDE?
Right, my one complains on the matter, too, but the code compiles and tests well.
So, what is the point to support IDE features on the code level?

spring-cloud-gcp-starters/spring-cloud-gcp-starter-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/autoconfig/GcpPubsubAutoConfiguration.java
artembilan: This comment has not been addressed yet.
Thanks

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/DefaultSubscriberFactory.java
joaoandremartins: Yes, that would probably require the `PubSubInboundChannelAdapter` to also take in a topic name (to which to bind the subscription), which could complicate things a little...

spring-cloud-gcp-starters/spring-cloud-gcp-starter-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/autoconfig/GcpPubsubAutoConfiguration.java
joaoandremartins: I thought the code wouldn't work because of the IDE complaint. It seems I was wrong. :)

spring-cloud-gcp-starters/spring-cloud-gcp-starter-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/autoconfig/GcpPubsubAutoConfiguration.java
joaoandremartins: IIUC, we can't use `@ConditionalOnMissingBean` here because one of these `ExecutorProviders` gets created first, and the other one never gets created. I think I ran into this issue a few days ago. I don't think `@ConditionalOnMissingBean` takes into account the bean's qualified name, maybe you can confirm?

spring-cloud-gcp-starters/spring-cloud-gcp-starter-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/autoconfig/GcpPubsubAutoConfiguration.java
joaoandremartins: See my comment above on `@ConditionalOnMissingBean`.

Also, on my test app, if I specify the same `@Bean`, mine is used, not this one. Is there any case in which my defined bean doesn't take precedence over this one from the library?

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/DefaultSubscriberFactory.java
artembilan: Well, we can go the way we have in Spring AMQP - `RabbitAdmin` and specific beans to represent auto-declaration entities on the Broker: http://docs.spring.io/spring-amqp/reference/html/_reference.html#broker-configuration

joaoandremartins:
@artembilan I will look into optionally running integration tests depending on whether Cloud SDK is installed in another issue. I like the SpringChannelProvider idea, we should also look into it in another issue.

Let me know if I missed any comment.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/autoconfig/GcpPubsubAutoConfiguration.java
```diff
@@ -0,0 +1,111 @@
...
+	public static final String DEFAULT_SOURCE_NAME = "spring";
+
+	@Value("${spring.cloud.gcp.pubsub.subscriber.executorThreads:4}")
+	private int subscriberExecutorThreads;
+
+	@Value("${spring.cloud.gcp.pubsub.publisher.executorThreads:4}")
+	private int publisherExecutorThreads;
+
+	@Bean
+	public ExecutorProvider publisherExecutorProvider() {
```
artembilan: The `@ConditionalOnMissingBean` must be here anyway.
Since we talk about *named* beans, like you have above with the `@Qualifier`, we have to use its opposite here as well:
```
/**
	 * The names of beans to check. The condition matches when each bean name specified is
	 * missing in the {@link ApplicationContext}.
	 * @return the name of beans to check
	 */
String[] name() default {};
```

spring-cloud-gcp-starters/spring-cloud-gcp-starter-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/autoconfig/GcpPubsubAutoConfiguration.java
```diff
@@ -0,0 +1,111 @@
...
+	}
+
+	@Bean
+	public ExecutorProvider subscriberExecutorProvider() {
+		return FixedExecutorProvider.create(
+				Executors.newScheduledThreadPool(this.subscriberExecutorThreads));
+	}
+
+	@Bean
+	public ChannelProvider subscriberChannelProvider(GoogleCredentials credentials) {
```
artembilan: The same `@ConditionalOnMissingBean(name = "subscriberChannelProvider")` approach here, as well.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/autoconfig/GcpPubsubAutoConfiguration.java
```diff
@@ -0,0 +1,111 @@
...
+		return SubscriptionAdminSettings.defaultChannelProviderBuilder()
+				.setCredentialsProvider(() -> credentials)
+				.setClientLibHeader(DEFAULT_SOURCE_NAME,
+						this.getClass().getPackage().getImplementationVersion())
+				.build();
+	}
+
+	@Bean
+	public ChannelProvider publisherChannelProvider(GoogleCredentials credentials) {
```
artembilan: Would you mind to elaborate more why do we need separate channel providers for subscriber and  publisher.

joaoandremartins: Good question. Because the Google Pubsub API provides distinct ways to get channel providers for both publishers and subscribers, even though I think the final object is the same. Still, I thought it would be a good practice to use both ways, depending on the case.

In any case, I'm preparing another PR on the side where `PublisherFactory` and `SubscriberFactory` will no longer take a `ChannelProvider`. The `ChannelProvider` used to be the way in which the user provided the `GoogleCredentials` for the API calls, but that is no longer the case -- `Publisher` and `Subscriber's` Builders now take a `CredentialsProvider`, and there is nothing else in the `ChannelProvider` left for users to configure that comes to mind.

PR #65, opened 2017-07-06
Adds CredentialProviders to Subscribers and Publishers directly

### LINKED ISSUE #64: Replace ChannelProvider.Builder.setCredentialsProvider by new method
`setCredentialsProvider` is deprecated on google-cloud-java 0.20.1-alpha and there is a new way to provide credentials.

joaoandremartins:
Setting CredentialsProviders in the ChannelProvider is deprecated since
google-cloud-pubsub version 0.20.1-alpha. That allows us to not require
a ChannelProvider from users, which better allows us to enforce that
clientLibName and clientLibVersion are filled with our data.

Fixes #64

### DIFF
--- a/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/DefaultPublisherFactory.java
+++ b/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/DefaultPublisherFactory.java
@@ -20,9 +20,11 @@ import java.io.IOException;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 
+import com.google.api.gax.core.CredentialsProvider;
 import com.google.api.gax.grpc.ChannelProvider;
 import com.google.api.gax.grpc.ExecutorProvider;
 import com.google.cloud.pubsub.v1.Publisher;
+import com.google.cloud.pubsub.v1.TopicAdminSettings;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.pubsub.v1.TopicName;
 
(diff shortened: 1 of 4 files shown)


spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/DefaultPublisherFactory.java
```diff
@@ -53,11 +55,17 @@
...
 
+	private final CredentialsProvider credentialsProvider;
+
 	public DefaultPublisherFactory(String projectId, ExecutorProvider executorProvider,
-			ChannelProvider channelProvider) {
+			CredentialsProvider credentialsProvider) {
 		this.projectId = projectId;
 		this.executorProvider = executorProvider;
-		this.channelProvider = channelProvider;
+		this.channelProvider = TopicAdminSettings.defaultChannelProviderBuilder()
```
meltsufin: I understand that the `credentialsProvider` no longer has to be set on the `channelProvider`, but why would we not want to allow the users to configure their own `channelProvider` and keep its auto-configuration?

In other words, why not have the constructor take both `ChannelProvider` and `CredentialsProvider`?

joaoandremartins: The idea is that there wouldn't be anything else in `ChannelProvider` that users would want to override. This would also help ensure that the `clientLib` header would be correctly filled.

meltsufin: I see that the `ChannelProvider` has [multiple implementations](http://googleapis.github.io/gax-java/0.2.0/apidocs/com/google/api/gax/grpc/ChannelProvider.html), but I'm not sure what the use-case would be for anything but the default. Maybe for testing?

artembilan: OK. One more time: we develop here Open Source and that wouldn't be so friendly to hide from end-user the possibility to inject any existing public `ChannelProvider` implementation. Even if that would be only for testing (mock), that is already end-user choice.

As far as I would insist to stop to chase that `clientLib` header. It is still open source and that is already end-user choice to show or hide such an option from the target platform.

Anyway he can't do that so clearly because he/she will always authenticate to the platform and we there always can track and trace what and how he/she does with our platform. As I suggested before, something like `SpringChannelProvider` wrapper may server the `clientLib` wishes, but can't be as a strong requirement.

If you really want to force that option, the GCP should reconsider it to something like a special file in the classpath. Or as I suggested once, too - via SSL key what end-users get from the platform together with the account.

PR #66, opened 2017-07-06
Renames PubSub classes to Pubsub

### LINKED ISSUE #56: Use Pubsub instead of PubSub in identifiers
Example: `PubsubTemplate` rather than `PubSubTemplate`.

joaoandremartins:
To keep in line with the branding on the Google side.

Fixes #56


spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/AbstractPubSubSender.java
```diff
@@ -18,7 +18,7 @@
```
artembilan: I can't point you to the line of code, but this class is still `AbstractPubSubSender`

I think you can just make a search through the project for the `*.java` files and the `PubSub` criteria match case:

image

joaoandremartins: Good catch, thanks!

PR #69, opened 2017-07-11
Adds PubsubAdmin

### LINKED ISSUE #26: Create a PubSubAdmin utility

joaoandremartins:
A Spring abstraction to simplify Google Cloud Pub/Sub topic and
subscription programmatic management. Also adds helpers to instantiate
PublisherFactory and SubscriberFactory.

Fixes #26

### DIFF
--- a/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/PubsubAdmin.java
+++ b/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/PubsubAdmin.java
@@ -29,6 +29,7 @@ import com.google.pubsub.v1.SubscriptionName;
 import com.google.pubsub.v1.Topic;
 import com.google.pubsub.v1.TopicName;
 
+import org.springframework.beans.factory.annotation.Value;
 import org.springframework.cloud.gcp.core.GcpProjectIdProvider;
 import org.springframework.util.Assert;
 
@@ -46,7 +47,8 @@ public class PubsubAdmin {
 	private SubscriptionAdminClient subscriptionAdminClient;
 
 	/** Default inspired in the subscription creation web UI. */
-	private static final int DEFAULT_ACK_DEADLINE = 10;
+	@Value("${spring.cloud.gcp.pubsub.subscriber.ackDeadline:10}")
+	private int defaultAckDeadline;
 
 	public PubsubAdmin(GcpProjectIdProvider projectIdProvider, TopicAdminClient topicAdminClient,
 			SubscriptionAdminClient subscriptionAdminClient) {
@@ -151,7 +153,7 @@ public class PubsubAdmin {
 		Assert.hasText(subscriptionName, "No subscription name was specified.");
 		Assert.hasText(topicName, "No topic name was specified.");
 
-		int finalAckDeadline = DEFAULT_ACK_DEADLINE;
+		int finalAckDeadline = this.defaultAckDeadline;
 		if (ackDeadline != null) {
 			Assert.isTrue(ackDeadline >= 0,
 					"The acknowledgement deadline value can't be negative.");
--- a/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/PubsubAdmin.java
+++ b/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/PubsubAdmin.java
@@ -80,9 +80,15 @@ public class PubsubAdmin {
 	 * @return the created topic
 	 */
 	public Topic createTopic(String topicName) {
-		Assert.hasText(topicName, "The topic name wasn't specified.");
+		Assert.hasText(topicName, "No topic name was specified.");
 
 		return this.topicAdminClient.createTopic(TopicName.create(this.projectId, topicName));
+	}
+
+	public void deleteTopic(String topicName) {
+		Assert.hasText(topicName, "No topic name was specified.");
+
+		this.topicAdminClient.deleteTopic(TopicName.create(this.projectId, topicName));
 	}
 
 	public Subscription createSubscription(String subscriptionName, String topicName) {
@@ -112,8 +118,8 @@ public class PubsubAdmin {
 	 */
 	public Subscription createSubscription(String subscriptionName, String topicName,
 			Integer ackDeadline, String pushEndpoint) {
-		Assert.hasText(subscriptionName, "The subscription name wasn't specified.");
-		Assert.hasText(topicName, "The topic name wasn't specified.");
+		Assert.hasText(subscriptionName, "No subscription name was specified.");
+		Assert.hasText(topicName, "No topic name was specified.");
 
 		int finalAckDeadline = DEFAULT_ACK_DEADLINE;
 		if (ackDeadline != null) {
@@ -134,5 +140,12 @@ public class PubsubAdmin {
 				TopicName.create(this.projectId, topicName),
 				pushConfigBuilder.build(),
 				finalAckDeadline);
+	}
+
+	public void deleteSubscription(String subscriptionName) {
+		Assert.hasText(subscriptionName, "No subscription name was specified");
+
+		this.subscriptionAdminClient.deleteSubscription(
+				SubscriptionName.create(this.projectId, subscriptionName));
 	}
 }
--- a/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/PubsubAdmin.java
+++ b/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/PubsubAdmin.java
@@ -195,7 +195,7 @@ public class PubsubAdmin {
 	}
 
 	public int getDefaultAckDeadline() {
-		return defaultAckDeadline;
+		return this.defaultAckDeadline;
 	}
 
 	public void setDefaultAckDeadline(int defaultAckDeadline) {
--- a/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/PubsubAdmin.java
+++ b/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/PubsubAdmin.java
@@ -22,13 +22,13 @@ import com.google.cloud.pubsub.v1.PagedResponseWrappers;
 import com.google.cloud.pubsub.v1.SubscriptionAdminClient;
 import com.google.cloud.pubsub.v1.TopicAdminClient;
 import com.google.common.collect.Lists;
-
 import com.google.pubsub.v1.ProjectName;
 import com.google.pubsub.v1.PushConfig;
 import com.google.pubsub.v1.Subscription;
 import com.google.pubsub.v1.SubscriptionName;
 import com.google.pubsub.v1.Topic;
 import com.google.pubsub.v1.TopicName;
+
 import org.springframework.cloud.gcp.core.GcpProjectIdProvider;
 import org.springframework.util.Assert;
 
--- a/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/PubsubAdmin.java
+++ b/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/PubsubAdmin.java
@@ -1,0 +1,138 @@
+/*
+ *  Copyright 2017 original author or authors.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package org.springframework.cloud.gcp.pubsub;
+
+import java.io.IOException;
+
+import com.google.cloud.pubsub.v1.SubscriptionAdminClient;
+import com.google.cloud.pubsub.v1.TopicAdminClient;
+import com.google.pubsub.v1.PushConfig;
+import com.google.pubsub.v1.Subscription;
+import com.google.pubsub.v1.SubscriptionName;
+import com.google.pubsub.v1.Topic;
+import com.google.pubsub.v1.TopicName;
+import org.apache.commons.validator.UrlValidator;
+
+import org.springframework.cloud.gcp.core.GcpProjectIdProvider;
+import org.springframework.util.Assert;
+
+/**
+ * Pub/Sub admin utility that creates new topics and subscriptions on Google Cloud Pub/Sub.
+ *
+ * @author João André Martins
+ */
+public class PubsubAdmin {
+
+	private final String projectId;
+
+	private TopicAdminClient topicAdminClient;
+
+	private SubscriptionAdminClient subscriptionAdminClient;
+
+	/** Default inspired in the subscription creation web UI. */
+	private static final int DEFAULT_ACK_DEADLINE = 10;
+
+	public PubsubAdmin(GcpProjectIdProvider projectIdProvider) throws IOException {
+		this(projectIdProvider, TopicAdminClient.create(), SubscriptionAdminClient.create());
+	}
+
+	public PubsubAdmin(GcpProjectIdProvider projectIdProvider, TopicAdminClient topicAdminClient)
+			throws IOException {
+		this(projectIdProvider, topicAdminClient, SubscriptionAdminClient.create());
+	}
+
+	public PubsubAdmin(GcpProjectIdProvider projectIdProvider,
+			SubscriptionAdminClient subscriptionAdminClient) throws IOException {
+		this (projectIdProvider, TopicAdminClient.create(), subscriptionAdminClient);
+	}
+
+	public PubsubAdmin(GcpProjectIdProvider projectIdProvider, TopicAdminClient topicAdminClient,
+			SubscriptionAdminClient subscriptionAdminClient) {
+		Assert.notNull(projectIdProvider, "The project ID provider can't be null.");
+		Assert.notNull(topicAdminClient, "The topic administration client can't be null");
+		Assert.notNull(subscriptionAdminClient,
+				"The subscription administration client can't be null");
+
+		this.projectId = projectIdProvider.getProjectId();
+		Assert.hasText(this.projectId, "The project ID can't be null or empty.");
+		this.topicAdminClient = topicAdminClient;
+		this.subscriptionAdminClient = subscriptionAdminClient;
+	}
+
+	/**
+	 * Creates a new topic on Google Cloud Pub/Sub.
+	 *
+	 * @param topicName the name for the new topic
+	 * @return the created topic
+	 */
+	public Topic createTopic(String topicName) {
+		Assert.hasText(topicName, "The topic name wasn't specified.");
+
+		return this.topicAdminClient.createTopic(TopicName.create(this.projectId, topicName));
+	}
+
+	public Subscription createSubscription(String subscriptionName, String topicName) {
+		return createSubscription(subscriptionName, topicName, null, null);
+	}
+
+	public Subscription createSubscription(String subscriptionName, String topicName,
+			Integer ackDeadline) {
+		return createSubscription(subscriptionName, topicName, ackDeadline, null);
+	}
+
+	public Subscription createSubscription(String subscriptionName, String topicName,
+			String pushEndpoint) {
+		return createSubscription(subscriptionName, topicName, null, pushEndpoint);
+	}
+
+	/**
+	 * Creates a new subscription on Google Cloud Pub/Sub.
+	 *
+	 * @param subscriptionName the name of the new subscription
+	 * @param topicName the name of the topic being subscribed to
+	 * @param ackDeadline deadline in seconds before a message is resent. If not provided, set to
+	 *                    default of 10 seconds
+	 * @param pushEndpoint URL of the service receiving the push messages. If not provided, set to
+	 *                    Google Cloud Pub/Sub default
+	 * @return the created subscription
+	 */
+	public Subscription createSubscription(String subscriptionName, String topicName,
+			Integer ackDeadline, String pushEndpoint) {
+		Assert.hasText(subscriptionName, "The subscription name wasn't specified.");
+		Assert.hasText(topicName, "The topic name wasn't specified.");
+
+		int finalAckDeadline = DEFAULT_ACK_DEADLINE;
+		if (ackDeadline != null) {
+			Assert.isTrue(ackDeadline >= 0,
+					"The acknowledgement deadline value can't be negative.");
+			finalAckDeadline = ackDeadline;
+		}
+
+		PushConfig.Builder pushConfigBuilder = PushConfig.newBuilder();
+		if (pushEndpoint != null) {
+			Assert.isTrue(new UrlValidator().isValid(pushEndpoint),
+					"The specified URL is invalid.");
+			pushConfigBuilder.setPushEndpoint(pushEndpoint);
+		}
+
+		return this.subscriptionAdminClient.createSubscription(
+				SubscriptionName.create(this.projectId, subscriptionName),
+				TopicName.create(this.projectId, topicName),
+				pushConfigBuilder.build(),
+				finalAckDeadline);
+	}
+}
--- a/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/PubsubAdmin.java
+++ b/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/PubsubAdmin.java
@@ -17,9 +17,13 @@
 package org.springframework.cloud.gcp.pubsub;
 
 import java.io.IOException;
+import java.util.List;
 
+import com.google.cloud.pubsub.v1.PagedResponseWrappers;
 import com.google.cloud.pubsub.v1.SubscriptionAdminClient;
 import com.google.cloud.pubsub.v1.TopicAdminClient;
+import com.google.common.collect.Lists;
+import com.google.pubsub.v1.ProjectName;
 import com.google.pubsub.v1.PushConfig;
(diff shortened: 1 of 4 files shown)


joaoandremartins:
There were questions on our side about whether this should be two separate classes: `TopicAdmin` and `SubscriptionAdmin`. We went with single class `PubsubAdmin` just for ease of use, but I can be persuaded to go a different way.

joaoandremartins:
Also, it just occurred to me that I need to add topic and subscription delete functionality to this class too.

meltsufin:
Also, please add `.travis.yml`.

joaoandremartins:
@meltsufin will do in another PR

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/PubsubAdmin.java
```diff
@@ -0,0 +1,151 @@
...
+
+	private TopicAdminClient topicAdminClient;
+
+	private SubscriptionAdminClient subscriptionAdminClient;
+
+	/** Default inspired in the subscription creation web UI. */
+	private static final int DEFAULT_ACK_DEADLINE = 10;
+
+	public PubsubAdmin(GcpProjectIdProvider projectIdProvider) throws IOException {
+		this(projectIdProvider, TopicAdminClient.create(), SubscriptionAdminClient.create());
```
artembilan: Why do we have `IOException` in the ctor?

It is really anti-pattern to get access to low level resources from ctors

Also how those admins work without credentials?
I mostly mean you auto-config below 

Thanjs

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/PubsubAdmin.java
```diff
@@ -0,0 +1,151 @@
...
+		int finalAckDeadline = DEFAULT_ACK_DEADLINE;
+		if (ackDeadline != null) {
+			Assert.isTrue(ackDeadline >= 0,
+					"The acknowledgement deadline value can't be negative.");
+			finalAckDeadline = ackDeadline;
+		}
+
+		PushConfig.Builder pushConfigBuilder = PushConfig.newBuilder();
+		if (pushEndpoint != null) {
+			Assert.isTrue(new UrlValidator().isValid(pushEndpoint),
```
artembilan: Why really `new Url()` isn't enough here?

Why do we need to extra dependencies almost for nothing?

joaoandremartins: I share your pain with the extra dependencies for seemingly little benefit..

`new URL()` doesn't really cut it because it requires us to catch an exception instead of just returning a boolean (and the consequent performance hit, although that's not really important) and it doesn't perform validation by itself as [evidenced by this comment](http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/6-b14/java/net/URL.java#407).

Looking around the internet, this seems to be the most popular way of validating URLs.

We can also just let the Pubsub library blow up on invalid endpoint. I just tested and it does complain:

```
Exception in thread "main" com.google.api.gax.grpc.ApiException: io.grpc.StatusRuntimeException: INVALID_ARGUMENT: Invalid push endpoint given (endpoint=asdasd). Refer to https://cloud.google.com/pubsub/subscriber#create for more information.
	at com.google.api.gax.grpc.ExceptionTransformingCallable$ExceptionTransformingFuture.onFailure(ExceptionTransformingCallable.java:108)
...
```

spring-cloud-gcp-starters/spring-cloud-gcp-starter-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/autoconfig/GcpPubsubAutoConfiguration.java
```diff
@@ -112,4 +115,15 @@ public PublisherFactory defaultPublisherFactory(GcpProjectIdProvider projectIdPr
 		return new DefaultPublisherFactory(projectIdProvider, subscriberProvider, channelProvider,
 				credentialsProvider);
 	}
+
+	@Bean
+	@ConditionalOnMissingBean
+	public PubsubAdmin pubsubAdmin(GcpProjectIdProvider projectIdProvider) {
+		try {
+			return new PubsubAdmin(projectIdProvider);
```
artembilan: ??? There is no such a ctor any more...

joaoandremartins: Oops! Should've installed it locally before..

spring-cloud-gcp-starters/spring-cloud-gcp-starter-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/autoconfig/GcpPubsubAutoConfiguration.java
```diff
@@ -112,4 +117,16 @@ public PublisherFactory defaultPublisherFactory(GcpProjectIdProvider projectIdPr
 		return new DefaultPublisherFactory(projectIdProvider, subscriberProvider, channelProvider,
 				credentialsProvider);
 	}
+
+	@Bean
+	@ConditionalOnMissingBean
+	public PubsubAdmin pubsubAdmin(GcpProjectIdProvider projectIdProvider) {
```
artembilan: BTW, the real auto-config is missed.
That is what I mean by the comment here: https://github.com/GoogleCloudPlatform/java-docs-samples/pull/755#discussion_r127274615

joaoandremartins: Sorry, not sure what you mean with the autoconfig is missed?

I removed `@IntegrationComponentScan` and `@ComponentScan` and `PubsubAdmin` is correctly autowired. You were right that I didn't need them, they must had been there from previous experiments...

artembilan: OK. Sorry. I thought the `PubsubAdmin` is in its own auto-config class.
But that is good to hear that everything is fine there already.
Since that is a sample it would be good to keep it as simple as possible to avoid unexpected questions in the future.

joaoandremartins: Totally agreed. It was exactly that kind of feedback I was seeking from you in the sample PR!

artembilan: OK. Sorry. I thought the `PubsubAdmin` is in its own auto-config class.
But that is good to hear that everything is fine there already.
Since that is a sample it would be good to keep it as simple as possible to avoid unexpected questions in the future.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/autoconfig/GcpPubsubAutoConfiguration.java
```diff
@@ -112,4 +117,16 @@ public PublisherFactory defaultPublisherFactory(GcpProjectIdProvider projectIdPr
...
+		try {
+			return new PubsubAdmin(projectIdProvider, TopicAdminClient.create(),
+					SubscriptionAdminClient.create());
```
artembilan: ???
I asked before security.
How all these `.create()` understand with what GC user they work?

Why this default, auto-configured `PubsubAdmin` doesn't depend on the `CredentialsProvider`?

That was really my first concern when I saw the same void `create()` in the `PubsubAdmin` ctors before.

Thanks

joaoandremartins:
@artembilan @meltsufin I think I addressed all the comments.

There is the question of whether this should be two classes (a `SubscriptionAdmin` and a `TopicAdmin`), but I think this simplifies usage and we can probably go with this for now. Later we might even want to make two classes, and one central class that takes the two as inputs.

artembilan:
I would say that I don't need this wrapper - the `SubscriptionAdminClient` and `TopicAdminClient` looks good to me as they are. 
Only a gain of this wrapper that it utilize `projectId`. So be it.

See my comment about `CredentialsProvider`.

Maybe we really ca consider a variant where we hide `SubscriptionAdminClient` and `TopicAdminClient` instances, but let end-user to specify their options externally?

I think I would go only with a single class anyway, if that is about my opinion.

joaoandremartins:
Good point on the credentials aspect -- it was my oversight.

I think `PubsubAdmin` actually fits really well within the Spring context. The Google Pub/Sub library uses objects like `SubscriptionName`, `TopicName`, etc., that we can easily spare the end user and instead just have them give us Strings.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/PubsubAdmin.java
```diff
@@ -0,0 +1,196 @@
...
+public class PubsubAdmin {
+
+	private final String projectId;
+
+	private TopicAdminClient topicAdminClient;
+
+	private SubscriptionAdminClient subscriptionAdminClient;
+
+	/** Default inspired in the subscription creation web UI. */
+	private static final int DEFAULT_ACK_DEADLINE = 10;
```
meltsufin: Would it make sense to expose it as a configuration property?
Something like `@Value("${spring.cloud.gcp.pubsub.defaultAckDeadline:10}")`.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/PubsubAdmin.java
```diff
@@ -0,0 +1,196 @@
...
+	/**
+	 * Returns every topic in a project.
+	 *
+	 * <p>If there are multiple pages, they will all be merged into the same result.
+	 */
+	public List<Topic> listTopics() {
+		PagedResponseWrappers.ListTopicsPagedResponse topicListPage =
+				this.topicAdminClient.listTopics(ProjectName.create(this.projectId));
+
+		return Lists.newArrayList(topicListPage.iterateAll());
```
meltsufin: The name of the local variable `topicListPage` seems to imply that it only contains a single page of results. Are you sure that `iterateAll()` actually return *all* results rather than a single page? If so, maybe rename `topicListPage` to just `topicsList`.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/PubsubAdmin.java
joaoandremartins: The `createSubscription()` API lets you specify an `ackDeadline` parameter. I didn't add the value here because the user can do that by themselves in another class (i.e., they can get it from the properties or any other source).

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/PubsubAdmin.java
joaoandremartins: It does only contain a single page of results. `iterateAll()` returns the full collection. Or, better yet, the full collection **counting from the current page**.

https://github.com/googleapis/gax-java/blob/master/gax-grpc/src/main/java/com/google/api/gax/grpc/AbstractPage.java#L135
https://github.com/googleapis/gax-java/blob/master/gax-grpc/src/main/java/com/google/api/gax/grpc/AbstractPage.java#L173

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/PubsubAdmin.java
meltsufin: I understand that the user can call the version of the method that takes `ackDeadline` but being able to configure the default externally could still be useful. Hardcoded defaults seem to be a bit contrary to Spring's philosophy.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/PubsubAdmin.java
meltsufin: Ok, my comment about the variable name still applies I think.

joaoandremartins: It would make sense for a local variable containing the result of `iterateAll()` to be called `topicsList`, not the `topicAdminClient.listTopics()` result. Also, the object type is `ListTopicsPagedResponse`, which also hints to it being a page, rather than the whole thing.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/PubsubAdmin.java
artembilan: I agree that we should have `setDefaultAckDeadline()`.
Like we have `KafkaTemplate.setDefautTopic()` or `RabbitTemplate.setReplyTimeout()`

spring-cloud-gcp-starters/spring-cloud-gcp-starter-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/autoconfig/GcpPubsubAutoConfiguration.java
```diff
@@ -112,4 +117,42 @@ public PublisherFactory defaultPublisherFactory(GcpProjectIdProvider projectIdPr
...
+	}
+
+	@Bean
+	@ConditionalOnMissingBean
+	public SubscriptionAdminClient subscriptionAdminClient(
+			CredentialsProvider credentialsProvider) {
+		try {
+			return SubscriptionAdminClient.create(
+					SubscriptionAdminSettings.defaultBuilder()
+							.setCredentialsProvider(credentialsProvider)
```
artembilan: Can't we have an additional `PubsubAdmin` ctor based on the `CredentialsProvider` to instantiate `TopicAdminClient` and `SubscriptionAdminClient` internally similar way?

Just in case for flexibility to let end-user to avoid this boilerplate code

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/PubsubAdmin.java
joaoandremartins: defaultAckDeadline is now configurable through properties.
Does that address your concerns?

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/PubsubAdmin.java
meltsufin: "paged response" does not necessarily imply "page". You can also call it verbosely `listTopicsPagedResponse`. In any case, it's so minor, it's not worth discussing further. :)

spring-cloud-gcp-starters/spring-cloud-gcp-starter-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/autoconfig/GcpPubsubAutoConfiguration.java
joaoandremartins: Well, I also like that kind of simplicity, but we would then need to have that kind of low level resource handling in the constructor you were opposed to earlier. i.e., we would need to instantiate `TopicAdminClient` and `SubscriptionAdminClient` in the constructor, and they can throw `IOException`.

But if we're OK with it, I can add it. I personally don't mind the `IOException` in the constructor.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/PubsubAdmin.java
```diff
@@ -46,7 +47,8 @@
 	private SubscriptionAdminClient subscriptionAdminClient;
 
 	/** Default inspired in the subscription creation web UI. */
-	private static final int DEFAULT_ACK_DEADLINE = 10;
+	@Value("${spring.cloud.gcp.pubsub.subscriber.ackDeadline:10}")
```
artembilan: No, that's wrong. 
To have `@Value` resolved your must have specific configuration in the application context.
Also this looks like a mix of auto-config and regular components concerns.
This is fully fine to have this property resolution in the auto-config, but here it must be setter.

joaoandremartins: At one point, I used a setter, but got rid of it because I don't want PubsubAdmin to have state. `PubsubAdmin` basically becomes not threadsafe.

artembilan: ??? So, in your terms anything with setter is evil?

All the components in Spring are based on setters. 
Right, you should always keep in mind if you can change the state of an object it isn't thread-safe.
That's not only for Spring classes.

I insist that we should not use `@Value` in target components. That is end-user application feature or maximum auto-config.

And why do you care too much that it won't be thread-safe?.. More over the story is only about an `int` value, which can be `volatile` and that's all...

joaoandremartins: I don't know about other Spring components, but IMO `PubsubAdmin` being this central class with a bunch of functionality and singleton by default, it's likely to be used from several different threads and inconsistency here is an actual hazard. Also, the added value of setting different defaults hardly justifies risking inconsistency.

I'm not opposed to getting rid of the `@Value` either -- setting a new default is a nice to have, but not _that_ useful, and users have the ability to call `createSubscription()` with a different default they have declared somewhere else.

Again, this is all IMO -- if you feel strongly about your point of view, I can add the setter...

meltsufin: FWIW I think the setter is fine. We make no claims about this class being thread-safe. Besides, this value would typically only be set on app initialization anyway.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/PubsubAdmin.java
```diff
@@ -0,0 +1,207 @@
...
+				this.subscriptionAdminClient.listSubscriptions(ProjectName.create(this.projectId));
+
+		return Lists.newArrayList(subscriptionsPage.iterateAll());
+	}
+
+	public int getDefaultAckDeadline() {
+		return this.defaultAckDeadline;
+	}
+
+	public void setDefaultAckDeadline(int defaultAckDeadline) {
```
artembilan: JavaDocs for public API.

joaoandremartins: For setters and getters?

spring-cloud-gcp-starters/spring-cloud-gcp-starter-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/autoconfig/GcpPubsubAutoConfiguration.java
artembilan: Looking to the `private SubscriptionAdminSettings(Builder settingsBuilder) throws IOException {` it is really unclear why an `IOException` is there from the day first at all.

So, having all these builders I don't think we really get access to low-level resource.
Anyway what you have here via dependency injection is still " low level resource handling" if you are interested in my opinion.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/PubsubAdmin.java
```diff
@@ -194,10 +194,18 @@ public void deleteSubscription(String subscriptionName) {
...
 
+	/**
+	 * @return the default acknowledgement deadline value in seconds
+	 */
 	public int getDefaultAckDeadline() {
 		return this.defaultAckDeadline;
 	}
 
+	/**
+	 * Sets the default acknowledgement deadline value.
```
artembilan: Nit-pick.
Method JavaDocs should be *imperative* not descriptive.
In other words: `Set the default ....`

spring-cloud-gcp-starters/spring-cloud-gcp-starter-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/autoconfig/GcpPubsubAutoConfiguration.java
artembilan: This is only concern left.
At least give some answer, please.
Thanks

joaoandremartins: Sorry, didn't mean to not reply here, just went grab lunch in the meantime. :)

`IOException` is thrown because of the channel creation, and maybe other cases, although I'm not sure about the exact circumstances it can happen (internet endpoint is unreachable?). Because of that, I think you do have access to low level resources, like a TCP connection, when you're creating a `Subscription`.

I really like this way because you still have to handle low level resources, but you do that away from the constructor. If you need to handle the `IOException`, you do it away from the constructor. And it should still be easy enough for users to just call `new PubsubAdmin(projectIdProvider, TopicAdminClient.create(), SubscriptionAdminClient.create())`.

Although, to be honest, I'm not opposed to offer that constructor on the side too.

artembilan: OK. Fine with me. Looks like we get `IOException` from the `CredentialsProvider.getCredentials()` when we open a file on the matter.

No more discussion on the matter! I can live with this. 😄

santhh:
I was trying the GCP pub/sub from a spring boot app and getting following error relate to pubsub admin in the auto configuration class. Any idea what may be causing this? Thanks..
=============
Error starting ApplicationContext. To display the auto-configuration report re-run your application with 'debug' enabled.
**2018-01-14 21:51:23.201 ERROR 8904 --- [           main] o.s.boot.SpringApplication               : Application startup failed

org.springframework.beans.factory.UnsatisfiedDependencyException: Error creating bean with name 'pubSubAdmin' defined in class path resource [org/springframework/cloud/gcp/pubsub/autoconfig/GcpPubSubAutoConfiguration.class]:** Unsatisfied dependency expressed through method 'pubSubAdmin' parameter 0; nested exception is org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'topicAdminClient' defined in class path resource [org/springframework/cloud/gcp/pubsub/autoconfig/GcpPubSubAutoConfiguration.class]: Bean instantiation via factory method failed; nested exception is org.springframework.beans.BeanInstantiationException: Failed to instantiate [com.google.cloud.pubsub.v1.TopicAdminClient]: Factory method 'topicAdminClient' threw exception; nested exception is java.lang.NoClassDefFoundError: org/threeten/bp/Duration
	at org.springframework.beans.factory.support.ConstructorResolver.createArgumentArray(ConstructorResolver.java:749) ~[spring-beans-4.3.13.RELEASE.jar:4.3.13.RELEASE]
	at org.springframework.beans.factory.support.ConstructorResolver.instantiateUsingFactoryMethod(ConstructorResolver.java:467) ~[spring-beans-4.3.13.RELEASE.jar:4.3.13.RELEASE]
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.instantiateUsingFactoryMethod(AbstractAutowireCapableBeanFactory.java:1173) ~[spring-beans-4.3.13.RELEASE.jar:4.3.13.RELEASE]
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.createBeanInstance(AbstractAutowireCapableBeanFactory.java:1067) ~[spring-beans-4.3.13.RELEASE.jar:4.3.13.RELEASE]
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.doCreateBean(AbstractAutowireCapableBeanFactory.java:513) ~[spring-beans-4.3.13.RELEASE.jar:4.3.13.RELEASE]
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.createBean(AbstractAutowireCapableBeanFactory.java:483) ~[spring-beans-4.3.13.RELEASE.jar:4.3.13.RELEASE]

santhh:
I was trying GCP pub sub form a spring boot app and get this error message on start up. It's relate to pub sub admin and auto configuration class. Has anyone seen this before? 
===============
org.springframework.beans.factory.UnsatisfiedDependencyException: Error creating bean with name 'pubSubAdmin' defined in class path resource [org/springframework/cloud/gcp/pubsub/autoconfig/GcpPubSubAutoConfiguration.class]: Unsatisfied dependency expressed through method 'pubSubAdmin' parameter 0; nested exception is org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'topicAdminClient' defined in class path resource [org/springframework/cloud/gcp/pubsub/autoconfig/GcpPubSubAutoConfiguration.class]:

artembilan:
@santhh,

Please, don’t repeat your problem in different places. We got your issue #307 and will come back to you as soon as possible, next week. Right now it is weekend and even I am typing here from the phone.

Thank you for understanding

Issue #21, opened 2017-05-26
Add Continuous Integration to GitHub repo


joaoandremartins:
We've been using self hosted Jenkins in other neighbour projects and I think it's been good so far.

viniciusccarvalho:
We have a hosted jenkins (https://jenkins.spring.io) that we can configure to build and publish artifacts

artembilan:
Even if we have already added Travis to PR's, we still need CI to publish snapshots and releasing eventually.

meltsufin:
@mbogoevici You mentioned you know someone who might be able to help us get this set up. Can you please point them here? Thanks!

markfisher:
I've added @sobychacko to the dev team here.

meltsufin:
Hi Soby! Thanks for volunteering to help us! Please let us know what you need from us to get the CI setup started.

sobychacko:
@meltsufin I will set it up soon.

sobychacko:
@meltsufin here is the URL for the Jenkins build: https://jenkins.spring.io/job/spring-cloud-gcp-master-ci/

meltsufin:
That was really fast! Thanks! I don't have read permissions though, it looks like.

sobychacko:
hmm - I thought everyone has read access with the GitHub logins. @trevormarshall can help with that.

meltsufin:
We now have access and the snapshots are published! Thanks guys!
@joaoandremartins I think you can close this issue now.

artembilan:
I'm reopening this issue because I see this in the PR:

![travis](https://user-images.githubusercontent.com/1071590/28464149-eb23e82e-6df1-11e7-9231-b28e4809ca47.png)

Can some one explain why do we we have two of them?
Plus that looks like they both perform the same task.
So, in my feeling we have to wait twice more time to determine that PR is OK.

Thanks

meltsufin:
One build is for the branch and another for the PR merge into master. [See this explanation on stackoverflow](https://stackoverflow.com/questions/34974925/travis-ci-creates-two-buids-for-each-github-commit-push/).
I think one way to disable the "useless" push build is to limit to master branch:
```yaml
branches:
  only:
    - master
```
Also, can we get the newly setup Jenkins CI to report build status to GitHub?

artembilan:
OK. Thank you, @meltsufin !

[...]

PR #71, opened 2017-07-11
Adds .travis.yml file

joaoandremartins:
Makes Travis build Java instead of Ruby.

Helps with #21

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,1 +1,3 @@
 language: java
+jdk:
+  - oraclejdk8
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,1 @@
+language: java


artembilan:
In the Gradle projects we have:
```
script:
    - ./gradlew check --refresh-dependencies --no-daemon
```
How does this "empty" config understand that it must call `./mvnw package` ?

Thanks

joaoandremartins:
It's calling `./mvnw test` by default. And it also calls `install` before to install dependencies.

https://docs.travis-ci.com/user/languages/java/#Default-script-Command

Do we want it to do anything differently?

meltsufin:
On other projects we do something like this, but I think the default is fine for now.

```yaml
script:
  # execute regular local build
  - mvn clean verify -B -V -q
```

artembilan:
Why did you merge this?
I didn't approve.

I really would like to see Checkstyle results in the Travis as well.

Thanks

joaoandremartins:
Sorry @artembilan I thought we could get any changes merged to `.travis.yml` as we needed them.

`./mvn test` runs checkstyle, it will fail if there are any issues.

artembilan:
Yeah... Confirmed:
```
D:\SpringIO\spring-cloud\spring-cloud-gcp>mvnw test
[INFO] Scanning for projects...
...
[ERROR] src\main\java\org\springframework\cloud\gcp\pubsub\support\SubscriberFactory.java:[29,54] (whitespace) WhitespaceAfter: ',' is not followed by whitespace.
```

I was under impression that we couldn't manage to work Checkstyle in some other project only via just `test` goal. We have to run `package` there...

Cool! Thank your for the patience! And sorry for extra noise.

Issue #22, opened 2017-05-26
Optimize GoogleStorageResource


artembilan:
The `blob` property must be `volatile`.
The `resolve()` method should not be `synchronized` from the first row.
We can use " Double-Checked Locking" pattern do not block all the threads if `blob` is already initialized:
```
  class Foo {
        private volatile Helper helper = null;
        public Helper getHelper() {
            if (helper == null) {
                synchronized(this) {
                    if (helper == null)
                        helper = new Helper();
                }
            }
            return helper;
        }
    }
```

https://www.cs.umd.edu/~pugh/java/memoryModel/DoubleCheckedLocking.html

meltsufin:
I think caching should be remove altogether because the `blob` can be modified from an external source. In addition, since blobs are immutable, if you write to it, you will get into a situation where your `blob` object actually points to an non-existing resource.

artembilan:
Closing according @meltsufin 's comment about `Blob` immutability. So, caching even can be evil in this case.

See linked #73 for more information.

PR #73, opened 2017-07-12
Extend GoogleStorageResource to allow writing

### LINKED ISSUE #72: Upgrade GoogleStorageResource to WritableResource

meltsufin:
In addition to being able to read objects in Google Cloud Storage
using Spring's Resource abstraction, we can now write resources
as well. GoogleStorageResource class now implements WritableResource
to support this. Note that if using @Value("gs://path-to/some-file.txt")
to inject a resource, it still needs to be cast from Resource to
WritableResource.

Fixes #72.


spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageOutputStream.java
```diff
@@ -0,0 +1,61 @@
...
+
+import java.io.IOException;
+import java.io.OutputStream;
+import java.nio.ByteBuffer;
+
+import com.google.cloud.WriteChannel;
+import com.google.cloud.storage.Blob;
+
+/**
+ * @author Mike Eltsufin
```
artembilan: Add JavaDocs, please

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResource.java
```diff
@@ -29,21 +30,22 @@
 import com.google.cloud.storage.Storage;
 
 import org.springframework.core.io.Resource;
+import org.springframework.core.io.WritableResource;
 import org.springframework.util.Assert;
 
 /**
+ * Implements {link WritableResource} for reading and writing objects in Google Cloud
```
artembilan: Missed `@` before `link` word

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResource.java
```diff
@@ -88,22 +90,16 @@ public URI getURI() throws IOException {
 	}
 
 	private Blob resolve() throws IOException {
-		synchronized (this.monitor) {
-			if (this.blob == null) {
```
artembilan: Any explanation why did you remove caching capability?

Thanks

artembilan: See #22 for some related issue

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageOutputStream.java
meltsufin: Will do. Note that the class is package private. I wasn't too sure about making it public. WDYT?

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResource.java
meltsufin: Sorry, I forgot to explain that in the commit. Basically, I think of the `Resource` object as simply a light reference to the remote resource. Since this remote resource can be modified from anywhere, we can't really control the consistency of our cache.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResource.java
```diff
@@ -88,22 +90,16 @@ public URI getURI() throws IOException {
...
-							"Failed to open remote connection to " + this.location, e);
-				}
-			}
+		try {
+			URI uri = getURI();
+			BlobId blobId = BlobId.of(uri.getHost(),
+					uri.getPath().substring(1, uri.getPath().length()));
+			return this.storage.get(blobId);
+		}
+		catch (Exception e) {
```
joaoandremartins: This catch is a bit strange. Only getURI() throws an `IOException`, yet we're catching every `Exception`.

Do we really need that extra error message? If so, I'd suggest we only catch `IOException`, instead. Otherwise, get rid of the catch altogether.

spring-cloud-gcp-storage/src/test/java/org/springframework/cloud/gcp/storage/GoogleStorageTests.java
```diff
@@ -50,6 +55,32 @@ public void testValidObject() throws Exception {
...
+	public void testGcsOutputStream() throws Exception {
+		Blob blob = Mockito.mock(Blob.class);
+		WriteChannel writeChannel = Mockito.mock(WriteChannel.class);
+		Mockito.when(blob.writer()).thenReturn(writeChannel);
+
+		GoogleStorageOutputStream os = new GoogleStorageOutputStream(blob);
+
+		// test single byte
+		os.write(99);
+		Mockito.verify(writeChannel).write(ByteBuffer.wrap(new byte[] { 99 }));
```
joaoandremartins: Would be interesting to read and then write. I guess that makes more sense on an integration test, though.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageOutputStream.java
```diff
@@ -0,0 +1,66 @@
...
+class GoogleStorageOutputStream extends OutputStream {
+
+	private WriteChannel writeChannel;
+
+	GoogleStorageOutputStream(Blob blob) {
+		this.writeChannel = blob.writer();
+	}
+
+	@Override
+	public void write(int b) throws IOException {
```
meltsufin: Just being consistent with the interface declaration.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageOutputStream.java
```diff
@@ -0,0 +1,66 @@
...
+		write(new byte[] { (byte) b });
```
meltsufin: It's part of the contract of this method in the interface: "The byte to be written is the eight low-order bits of the argument b. The 24 high-order bits of b are ignored." See: [OutputStream](https://docs.oracle.com/javase/8/docs/api/java/io/OutputStream.html#write-int-).

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResource.java
meltsufin: I agree with just removing the try-catch.

spring-cloud-gcp-storage/src/test/java/org/springframework/cloud/gcp/storage/GoogleStorageTests.java
meltsufin: Yep, we should add that when we add integration tests.

meltsufin:
@joaoandremartins I agree that `GoogleStorageResource.isOpen()` always returning false is a bit odd, but it's something that really doesn't apply to GCS. So, returning `false` is a reasonable compromise imo.

meltsufin:
@artembilan @joaoandremartins I think all of the review comments have been addressed. Please take one more look. Thanks!

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageOutputStream.java
joaoandremartins: You're not forced to adopt the name from the interface. I typically use a more descriptive name in these cases, but this isn't worth a big discussion anyway :)

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageOutputStream.java
joaoandremartins: The things you know :) Interesting that the behaviour of casting an integer to a byte is to just use the least order 8 bits and discard the remaining ones, instead of a plain fail.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResource.java
```diff
@@ -146,4 +136,14 @@ public String getDescription() {
...
 	}
+
+	@Override
+	public boolean isWritable() {
+		return true;
+	}
+
+	@Override
+	public OutputStream getOutputStream() throws IOException {
+		return new GoogleStorageOutputStream(resolve());
```
artembilan: I wonder why doesn't `return Channels.newOutputStream(resolve().writer());` fit here?
Why do we need this custom `OutputStream` impl?

Thanks

meltsufin: I didn't realize that the `com.google.cloud.WriteChannel` actually implements `java.nio.channels.WritableByteChannel` and we can use the `Channels.newOutputStream`. This is actually much better! Thanks!

PR #77, opened 2017-07-24
Fixes dependency hell by not including google-cloud dependency

### LINKED ISSUE #16: Fix dependency hell
Right now dependency management is very chaotic. Depending on the `google-cloud-java` is a problem for modules that really do not depend on all the libs. 

Just to make the project build, right now all projects depend on `google-cloud-java` which is an abuse of dependencies.

joaoandremartins:
Imports single dependencies, instead. Also fixes other nits with
pom.xml files.

Fixes #16

### DIFF
--- a/spring-cloud-gcp-core/pom.xml
+++ b/spring-cloud-gcp-core/pom.xml
@@ -10,12 +10,13 @@
 	</parent>
 
 	<artifactId>spring-cloud-gcp-core</artifactId>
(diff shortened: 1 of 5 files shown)


joaoandremartins:
Hopefully this addresses the bigger issue with dependencies. I don't think we can do any finer grain than this, but let me know if I missed anything.

PR #78, opened 2017-07-25
Google Cloud SQL Boot starter

### LINKED ISSUE #74: Add Cloud SQL Boot starter
Still thinking what this will entail.

joaoandremartins:
Provides a better user experience where the user only provides the
instance and database names, along with username and password, and we
construct the URL for them.

Fixes #74


spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/test/java/org/springframework/cloud/gcp/sql/autoconfig/GcpCloudSqlAutoConfigurationTests.java
```diff
@@ -0,0 +1,158 @@
...
+	@Mock
+	private SQLAdmin.Instances mockInstances;
+
+	@Mock
+	private SQLAdmin.Instances.Get mockGet;
+
+	private DatabaseInstance databaseInstance = new DatabaseInstance();
+
+	@Before
+	public void setUp() throws IOException {
```
joaoandremartins: @artembilan I'm having some issues here and could use your help.

The `cloudSqlDataSource()` bean in `GcpCloudSqlAutoConfiguration` is being configured before the `@Before` block is run, therefore, `sqlAdmin.instances()` blows up with an NPE.

I'm not sure what the best way to proceed here, since I seem to be following everything from https://spring.io/blog/2016/04/15/testing-improvements-in-spring-boot-1-4 (except I'm not testing as a web application but rather just pulling the bean from the boot app, but it should be the same).

Also, it seems that the only way to test with different property settings is to have multiple test classes for multiple property test cases? The only place I've seen where I can specify properties is at the class level...

spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/src/main/java/org/springframework/cloud/gcp/core/autoconfig/GcpContextAutoConfiguration.java
```diff
@@ -76,4 +79,19 @@ public GcpProjectIdProvider gcpProjectIdProvider() {
...
 				: new DefaultGcpProjectIdProvider();
 	}
+
+	/**
+	 * Google Client Library APIs (i.e., non-Veneer/google-cloud-java) require
+	 * {@link GoogleCredential} instead of {@link GoogleCredentials}.
+	 */
+	@Bean
+	@ConditionalOnMissingBean
+	public Supplier<GoogleCredential> googleCredentialSupplier() throws IOException {
```
artembilan: ??? It's not good. No body would be happy to deal with the `Supplier` bean. With this you try to eliminate the good search improvement via `BeanFactory.getBean(Class<T> requiredType)`

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/pom.xml
```diff
@@ -0,0 +1,57 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  ~  Copyright 2017 original author or authors.
```
artembilan: No Copyrights in the non-Java files.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/autoconfig/GcpCloudSqlAutoConfiguration.java
```diff
@@ -0,0 +1,127 @@
...
+import org.springframework.cloud.gcp.sql.GcpCloudSqlProperties;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.util.Assert;
+import org.springframework.util.StringUtils;
+
+/**
+ * Google Cloud SQL starter.
+ *
+ * <p>>It simplifies database configuration in that only an instance and database name are provided,
```
artembilan: Typo: extra `>`

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/test/java/org/springframework/cloud/gcp/sql/autoconfig/GcpCloudSqlAutoConfigurationTests.java
artembilan: Consider to create some `@Configuration` class for test beans. And that `mock(SQLAdmin.class)` you can move there to the `@Bean` definition which overrides your `createSqlAdminService()` (what the name for the bean?!) definition.

The `@MockBean` is intended for target application beans, but not for mocking auto-config.

joaoandremartins: The same issue happens: I'm autowiring the subject `DataSource` object and this happens at `GcpCloudSqlAutoConfigurationTests` creation time, therefore, `@Before` still hasn't run, so the behaviour wasn't mocked yet when `DataSource` is created.

The only solution I can see is to move the setup mocking logic to inside the `sqlAdminService` `@Bean` generator, although I don't think this is very nice...

artembilan: ??? Show, please, your `@Bean` for the `mock(SQLAdmin.class)`.
Or would be better to push what you have so far.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/pom.xml
```diff
@@ -0,0 +1,41 @@
...
+		<dependency>
+			<groupId>com.google.apis</groupId>
+			<artifactId>google-api-services-sqladmin</artifactId>
+			<version>v1beta4-rev45-1.22.0</version>
+		</dependency>
+
+		<!-- https://mvnrepository.com/artifact/mysql/mysql-connector-java -->
+		<dependency>
+			<groupId>mysql</groupId>
+			<artifactId>mysql-connector-java</artifactId>
```
artembilan: Why do we need `mysql` though?

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/autoconfig/GcpCloudSqlAutoConfiguration.java
```diff
@@ -0,0 +1,127 @@
...
+	@Bean
+	@ConditionalOnMissingBean(DataSource.class)
+	public DataSource cloudSqlDataSource(SQLAdmin sqlAdmin)
+			throws GeneralSecurityException, IOException {
+		Assert.hasText(this.gcpCloudSqlProperties.getDatabaseName(),
+				"spring.cloud.gcp.sql.databaseName name can't be empty.");
+		Assert.hasText(this.gcpCloudSqlProperties.getInstanceName(),
+				"spring.cloud.gcp.sql.instanceName name can't be empty.");
+		Assert.hasText(this.gcpCloudSqlProperties.getUserName(),
+				"spring.cloud.gcp.sql.userName name can't be null.");
```
artembilan: See http://docs.spring.io/spring-boot/docs/1.5.4.RELEASE/reference/html/boot-features-external-config.html#boot-features-external-config-validation instead.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/pom.xml
```diff
@@ -0,0 +1,41 @@
...
+		<artifactId>spring-cloud-gcp-starters</artifactId>
+		<groupId>org.springframework.cloud</groupId>
+		<version>1.0.0.BUILD-SNAPSHOT</version>
+	</parent>
+	<modelVersion>4.0.0</modelVersion>
+
+	<artifactId>spring-cloud-gcp-starter-sql</artifactId>
+
+	<dependencies>
+		<dependency>
```
artembilan: I think:
```
<dependency>
			<optional>true</optional>
</dependency>
```
is missed everywhere we have `@ConfigurationProperties`: http://docs.spring.io/spring-boot/docs/1.5.4.RELEASE/reference/htmlsingle/#configuration-metadata-annotation-processor

spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/src/main/java/org/springframework/cloud/gcp/core/autoconfig/GcpContextAutoConfiguration.java
artembilan: Looks like we don't need this at all:
````java
protected SQLAdmin sqlAdminService(CredentialsProvider googleCredentials)
			throws GeneralSecurityException, IOException {
	HttpTransport httpTransport = GoogleNetHttpTransport.newTrustedTransport();
	JsonFactory jsonFactory = JacksonFactory.getDefaultInstance();

	return new SQLAdmin.Builder(httpTransport, jsonFactory,
			new HttpCredentialsAdapter(googleCredentials.getCredentials()))
					.setApplicationName("Google-SQLAdminSample/0.1")
					.build();
}
````

File: spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/test/java/org/springframework/cloud/gcp/sql/autoconfig/GcpCloudSqlAutoConfigurationTests.java (hunk 1 above)
Comment (artembilan) [inline spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/test/java/org/springframework/cloud/gcp/sql/autoconfig/GcpCloudSqlAutoConfigurationTests.java]: I don't understand your concern here.
The `SQLAdmin` is indeed mock:

image

with your latest changes.

We have a problem a bit later for `socketFactory`, but that already different story. 
The `GcpCloudSqlTestConfiguration` does its trick.

File: spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/src/main/java/org/springframework/cloud/gcp/core/autoconfig/GcpContextAutoConfiguration.java [hunk 8]
```diff
@@ -76,4 +79,19 @@ public GcpProjectIdProvider gcpProjectIdProvider() {
 				? () -> this.gcpProperties.getProjectId()
 				: new DefaultGcpProjectIdProvider();
 	}
+
+	/**
+	 * Google Client Library APIs (i.e., non-Veneer/google-cloud-java) require
```
meltsufin: Please don't use the term "Veneer".

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/autoconfig/GcpCloudSqlAutoConfiguration.java
```diff
@@ -0,0 +1,127 @@
...
+
+		// Look for the region if the user didn't specify one.
+		String region = this.gcpCloudSqlProperties.getRegion();
+		if (!StringUtils.hasText(region)) {
+			DatabaseInstance instance =	sqlAdmin.instances().get(
```
meltsufin: Odd tab after the `=`?

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/autoconfig/GcpCloudSqlAutoConfiguration.java
```diff
@@ -0,0 +1,127 @@
...
+					this.gcpProperties.getProjectId(),
+					this.gcpCloudSqlProperties.getInstanceName()).execute();
+
+			region = instance.getRegion();
```
meltsufin: Why not combine with the previous line and avoid using a local var for `instance`?

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/autoconfig/GcpCloudSqlAutoConfiguration.java
```diff
@@ -0,0 +1,127 @@
...
+		String jdbcUrl = String.format("jdbc:mysql://google/%s?cloudSqlInstance=%s&"
+						+ "socketFactory=com.google.cloud.sql.mysql.SocketFactory",
+				this.gcpCloudSqlProperties.getDatabaseName(),
+				instanceConnectionName);
+
+		// TODO(joaomartins): Set the credential factory to be used by SocketFactory when
+		// https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/issues/41 is
+		// resolved. For now, it uses application default creds.
+
+		HikariConfig hikariConfig = new HikariConfig();
```
meltsufin: I feel like this should be a separated out from auto-construction of the `jdbcUrl`. There's too much going on in this method. It would be nice to let the users overwrite which connection pool to use without sacrificing the auto-construction of the JDBC URL.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/src/main/java/org/springframework/cloud/gcp/core/autoconfig/GcpContextAutoConfiguration.java
artembilan: When we go to the `HttpCredentialsAdapter`, we won't need this bean definition at all. Therefore "Veneer" will go.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/test/java/org/springframework/cloud/gcp/sql/autoconfig/GcpCloudSqlAutoConfigurationTests.java
joaoandremartins: It only works because I applied that "hack" I mentioned earlier -- I'm mocking the `SQLAdmin` behaviour in the configuration, not in the `@Before` `setUp()` method as is best done.

artembilan: ??? It isn't going to work in the `@Before` any way.

The `ApplicationContext` is fired up much earlier before `@Before`. And need to understand that `@Before` is called before each test method. So, that is again not nice to re-mock for each method when we need only once. 

That is absolutely normal to produce full stub in the `@Bean` method. More over it is an infrastructure bean, not end-user logic when we can simply use `@MockBean`.

Does it make sense for you?

joaoandremartins: OK that sounds good.

I guess the other question I have is about how to test with different configurations. `@SpringBootTest` takes in a properties attribute, but it is a class level annotation. If I want to test a different configuration (e.g., in this case, I want to test a configuration with/without a region), it seems like I have to create a new class, whereas I would like to just add a new method to this class.

I've seen this before with that `loadEnvironment()` and `EnvironmentTestUtils`, but I couldn't get my test configuration to work well that way and I'm curious if there is another way of using `@SpringBootTest`.

artembilan: Yeah... Right. Unfortunately there is no other way, unless a new test class.
That is exactly how we do in Spring Cloud Stream App Starters.
For example S3 Source tests: https://github.com/spring-cloud-stream-app-starters/aws-s3/blob/master/spring-cloud-starter-stream-source-s3/src/test/java/org/springframework/cloud/stream/app/s3/source/AmazonS3SourceMockTests.java

We organize the hierarchy for them do not copy/paste shared configs.
Pay attention to the `@TestPropertySource` on each sub-class.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/src/main/java/org/springframework/cloud/gcp/core/autoconfig/GcpContextAutoConfiguration.java
joaoandremartins: Good stuff!

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/pom.xml
joaoandremartins: Why do you say so? I'm able to install everything correctly, it must be a transitive dependency? Do you still require us to explicitly state it in the pom?

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/pom.xml
joaoandremartins: Well, this has to do with one open issue -- right now, in order for there to be any registered driver, the user must actually add this dependency on their pom.xml file for the static code that registers the driver to be run.

Ideally, the user wouldn't need to do that, and adding it here would register it, but it's not working for some reason. I need to look more into this.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/test/java/org/springframework/cloud/gcp/sql/autoconfig/GcpCloudSqlAutoConfigurationTests.java
joaoandremartins: I like it! I'll do this tomorrow.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/pom.xml
artembilan: M-m-m. Why do you never read links I point?..
That Boot reference says clearly:
> You can easily generate your own configuration meta-data 

We need this for the compile time to generate particular JSON data in the META-INF of our jar.
When end-user uses our jar as dependency, the IDE is able to read that metadata and allow to auto-complete `application.properties`.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/autoconfig/GcpCloudSqlAutoConfiguration.java
```diff
@@ -0,0 +1,118 @@
...
+ *
+ * <p>It simplifies database configuration in that only an instance and database name are provided,
+ * instead of a JDBC URL. If the instance region isn't specified, this starter attempts to get it
+ * through the Cloud SQL API.
+ *
+ * @author João André Martins
+ */
+@Configuration
+@EnableConfigurationProperties(GcpCloudSqlProperties.class)
+@ConditionalOnClass({GcpProperties.class, GoogleCredential.class})
```
artembilan: I think we don't need `GoogleCredential` any more.
We might rely on some other more SQL-specific class.
How about `SQLAdmin`?

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/test/java/org/springframework/cloud/gcp/sql/autoconfig/GcpCloudSqlTestConfiguration.java
```diff
@@ -0,0 +1,53 @@
...
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.Primary;
+
+import static org.mockito.Matchers.eq;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+/**
+ * Created by joaomartins on 7/25/17.
```
artembilan: Typo.
Must be `João André Martins`

Or have you changed your name ? 😉

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/autoconfig/GcpCloudSqlAutoConfiguration.java
joaoandremartins: Hmm if the user provides their own region, `SQLAdmin` isn't necessary. I'm not sure if it's worth adding.

artembilan: Well, the main concern here that we can have a jar with the `org.springframework.cloud.gcp.sql.autoconfig` in the classpath, but we may not have some transitive dependency. Therefore we have to use some class from there (`HikariDataSource`?) do not trigger this auto-config.

joaoandremartins: OK, then `SQLAdmin` sounds like an adequate choice.

I'm also wondering if we should add `GoogleCredentials` to the list. It is kind of a transitive dependency, because `GcpProperties` is enabled by `GcpContextAutoConfiguration`, and that depends on `GoogleCredentials`. But it would probably be cautious to add it here, too. (e.g., in case `GcpContextAutoConfiguration` no longer depends on `GoogleCredentials` in the future).

artembilan: Well, instead we can just add `@AutoConfigureAfter(GcpContextAutoConfiguration.class)`

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/CloudSqlJdbcUrlProvider.java
```diff
@@ -0,0 +1,29 @@
...
+package org.springframework.cloud.gcp.sql;
+
+/**
+ * Implementations of this interface typically construct a JDBC URL for Cloud SQL from a GCP
+ * project ID and an instance name.
+ *
+ * @author João André Martins
+ */
+@FunctionalInterface
+public interface CloudSqlJdbcUrlProvider {
```
joaoandremartins: I added this in the SQL Boot starter. I didn't create a new module because I thought it wasn't worth it, but let me know if you think otherwise.

artembilan: Fine with me.
Although that looks like we don't need it at all.
Simple `@PostConstruct` or `private` method to build the URL from the properties looks pretty enough for me

meltsufin: We wanted to separate the creation of the URL so that users can take advantage of it while still customizing the JDBC connection pool.

joaoandremartins: The motivation for this was to split the `DataSource` `@Bean` method in two -- constructing the JDBC URL and creating a `DataSource` with it. We wanted to give users the flexibility of using their own `DataSources` but using our URL builder.

This interface is a way to avoid a `String` `@Bean`.

artembilan: OK. Fine with me.
I wanted to be sure that there is no other way to avoid `String`-wrapper bean.

Can't we rely on the Boot `DataSourceProperties` BTW?

joaoandremartins: We thought about it, but I don't think we can because there are several fields that are specific to our DataSource configuration: project ID, region name and instance name. Project ID is already in `spring.cloud.gcp`, but we needed another namespace to create the other two (`spring.cloud.gcp.sql`). Also using `DataSourceProperties` would mix namespaces, and this could be confusing to users.

We could use `DataSourceProperties` as a POJO to feed into `cloudSqlDataSource()`, and then just do the field mapping inside the method, but I don't see many advantages in this -- the POJO would just do to carry the URL since the username and password are already in the `GcpCloudSqlProperties` object.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/test/java/org/springframework/cloud/gcp/sql/autoconfig/GcpCloudSqlAutoConfigurationMockTests.java
```diff
@@ -0,0 +1,96 @@
...
+
+	@Autowired
+	protected DataSource dataSource;
+
+	@Autowired
+	protected CloudSqlJdbcUrlProvider urlProvider;
+
+	public abstract void test();
+
+	public static class CloudSqlJdbcUrlProviderTest extends GcpCloudSqlAutoConfigurationMockTests {
```
joaoandremartins: @artembilan I noticed that only this first test gets run, the other two don't.

I looked around a bit and didn't find out why. Do you have an idea?

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/CloudSqlJdbcUrlProvider.java
artembilan: Fair enough!

Let it be string provider then! 😉

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/test/java/org/springframework/cloud/gcp/sql/autoconfig/GcpCloudSqlAutoConfigurationMockTests.java
artembilan: I don't know what you are doing there. Works for me:
image


Also the `GcpCloudSqlAutoConfigurationTests_WithRegion` ends up with the `no running methods`.
You are really missing there `@Test` on the `test()`.

The Maven doesn't see some classes because they don't have `Test(s)` suffix 😄

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/pom.xml
joaoandremartins: Neat! Is there anything similar for `@Value`? I noticed that the Pubsub publisher and subscriber configuration isn't recognised by IDEA.
Cool that we can do "project-id" instead of "projectId" in the configuration.

artembilan: That's different story.
We talk here about `application.properties/yml` definition.
The `@Value` is a user of properties already.
And that is already up to IntelliJ question... Far away outside of scope of this PR.

Only what I want here is to have that dependency.
Thanks

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/test/java/org/springframework/cloud/gcp/sql/autoconfig/GcpCloudSqlAutoConfigurationMockTests.java
joaoandremartins: D'oh! Thanks

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/pom.xml
joaoandremartins: I'm slightly concerned that the applications.properties/yml properties will work like "projectId" or "project-id", but `@Value` ones only work like "executorThreads" and the IDE doesn't even suggest them.

Nothing that documentation can't fix, I guess.

joaoandremartins:
I also made the Boot starter register the MySQL driver in my last commit, so I think this is ready for final review on my side.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/autoconfig/GcpCloudSqlAutoConfiguration.java
```diff
@@ -0,0 +1,123 @@
...
+	@Bean
+	@ConditionalOnMissingBean(DataSource.class)
+	public DataSource cloudSqlDataSource(CloudSqlJdbcUrlProvider cloudSqlJdbcUrlProvider)
+			throws GeneralSecurityException, IOException, ClassNotFoundException {
+		// TODO(joaomartins): Set the credential factory to be used by SocketFactory when
+		// https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/issues/41 is
+		// resolved. For now, it uses application default creds.
+
+		// Load MySQL driver to register it, so the user doesn't have to.
+		Class.forName("com.mysql.jdbc.Driver");
```
joaoandremartins: Yeah, you're right actually. The magic we're really adding is the URL builder, the user should have more freedom to choose their `DataSource`.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/src/main/java/org/springframework/cloud/gcp/core/autoconfig/GcpContextAutoConfiguration.java
```diff
@@ -47,7 +48,12 @@
 @EnableConfigurationProperties(GcpProperties.class)
 public class GcpContextAutoConfiguration {
 
-	private static final String GCP_PUBSUB_SCOPE = "https://www.googleapis.com/auth/pubsub";
+	private static final String PUBSUB_SCOPE = "https://www.googleapis.com/auth/pubsub";
+
+	private static final String SQLADMIN_SCOPE =
+			"https://www.googleapis.com/auth/sqlservice.admin";
+
+	private static final List<String> SCOPES_LIST = ImmutableList.of(PUBSUB_SCOPE, SQLADMIN_SCOPE);
```
joaoandremartins: Hum I'm not sure that's worth the trouble TBH... It's more complexity for the user with questionable benefit.

Unless a user wants to make sure that their app never does anything Pub/Sub, and they don't specify the scope? But then, I think they should just disable that permission on IAM instead, not providing a scope to achieve that would be the wrong way IMO.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/autoconfig/GcpCloudSqlAutoConfiguration.java
```diff
@@ -0,0 +1,123 @@
...
+	@Bean
+	@ConditionalOnMissingBean
+	protected SQLAdmin sqlAdminService(CredentialsProvider credentialsProvider)
+			throws GeneralSecurityException, IOException {
+		HttpTransport httpTransport = GoogleNetHttpTransport.newTrustedTransport();
+		JsonFactory jsonFactory = JacksonFactory.getDefaultInstance();
+
+		return new SQLAdmin.Builder(httpTransport, jsonFactory,
+				new HttpCredentialsAdapter(credentialsProvider.getCredentials()))
+				.setApplicationName("Google-SQLAdminSample/0.1")
```
joaoandremartins: Oh wow, can't believe this slipped... I never read this correctly and always thought it was a required control field. I removed it and it works fine.

Although this begs the question -- like we do for Pub/Sub, we should also be getting metrics in the Storage and Cloud SQL side. I think this deserves a new issue.

joaoandremartins: #83

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/GcpCloudSqlProperties.java
```diff
@@ -0,0 +1,95 @@
...
+	@NotEmpty
+	private String instanceName;
+
+	@NotEmpty
+	private String databaseName;
+
+	private String region;
+
+	@NotEmpty
+	private String userName;
```
meltsufin: Can we default to `root`?

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/autoconfig/GcpCloudSqlAutoConfiguration.java
```diff
@@ -0,0 +1,123 @@
...
+		// https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/issues/41 is
+		// resolved. For now, it uses application default creds.
+
+		// Load MySQL driver to register it, so the user doesn't have to.
+		Class.forName("com.mysql.jdbc.Driver");
+
+		// Look for the region if the user didn't specify one.
+		String region = this.gcpCloudSqlProperties.getRegion();
+		if (!StringUtils.hasText(region)) {
+			region = sqlAdmin.instances().get(this.gcpProperties.getProjectId(),
```
meltsufin: As we just talked about, let's use `ProjectIdProvider` instead.

joaoandremartins:
@artembilan do you mind taking another look at the two latest commits? Sorry and thanks!

joaoandremartins:
@meltsufin please approve if you're OK with these changes and I'll merge

Issue #89, opened 2017-08-07
Add a conditional to use a different JDBC URL connection string & Driver if running in App Engine Standard


saturnism:
See https://cloud.google.com/appengine/docs/standard/java/cloud-sql/

```
if (System
        .getProperty("com.google.appengine.runtime.version").startsWith("Google App Engine/")) {
  // inside App Engine runtime
  // JDBC Driver: com.mysql.jdbc.GoogleDriver
 // JDBC URL: jdbc:google:mysql://${INSTANCE_CONNECTION_NAME}/${database}?user=${user}&amp;password=${password}
} else {
  // JDBC Driver/URL: Regular MYSQL Driver w/ SSL Socket Factory
}
```

PR #90, opened 2017-08-07
Add a conditional to use a different JDBC URL connection string & Driver if running in App Engine Standard

saturnism:
For issue #89

### DIFF
--- a/spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/src/main/java/org/springframework/cloud/gcp/core/AppEngineCondition.java
+++ b/spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/src/main/java/org/springframework/cloud/gcp/core/AppEngineCondition.java
@@ -1,0 +1,43 @@
+/*
+ *  Copyright 2017 original author or authors.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package org.springframework.cloud.gcp.core;
+
+import org.springframework.context.annotation.Condition;
+import org.springframework.context.annotation.ConditionContext;
+import org.springframework.core.type.AnnotatedTypeMetadata;
+import org.springframework.util.StringUtils;
+
+/**
+ * Determine whether this is running inside of an App Engine environment by looking into
+ * the System properties `com.google.appengine.runtime.version`.
+ * @author
+ */
+public class AppEngineCondition implements Condition {
+	public static final String APPENGINE_RUNTIME_PROPERTY = "com.google.appengine.runtime.version";
+
+	public static final String APPENGINE_RUNTIME_PREFIX = "Google App Engine/";
+
+	@Override
+	public boolean matches(ConditionContext conditionContext, AnnotatedTypeMetadata annotatedTypeMetadata) {
+		String appEngineVersion = System.getProperty(APPENGINE_RUNTIME_PROPERTY);
+		if (StringUtils.isEmpty(appEngineVersion)) {
+			return false;
+		}
+
+		return appEngineVersion.startsWith(APPENGINE_RUNTIME_PREFIX);
+	}
+}
--- a/spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/src/main/java/org/springframework/cloud/gcp/core/AppEngineCondition.java
+++ b/spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/src/main/java/org/springframework/cloud/gcp/core/AppEngineCondition.java
@@ -24,7 +24,8 @@ import org.springframework.util.StringUtils;
 /**
  * Determine whether this is running inside of an App Engine environment by looking into
  * the System properties `com.google.appengine.runtime.version`.
- * @author
+ *
+ * @author Ray Tsang
  */
 public class AppEngineCondition implements Condition {
 	public static final String APPENGINE_RUNTIME_PROPERTY = "com.google.appengine.runtime.version";
--- a/spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/CloudSqlJdbcUrlProvider.java
+++ b/spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/CloudSqlJdbcUrlProvider.java
@@ -22,8 +22,7 @@ package org.springframework.cloud.gcp.sql;
  *
  * @author João André Martins
  */
-@FunctionalInterface
 public interface CloudSqlJdbcUrlProvider {
-
+	String getJdbcDriverClass();
 	String getJdbcUrl();
 }
--- a/spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/GcpCloudSqlProperties.java
+++ b/spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/GcpCloudSqlProperties.java
@@ -38,6 +38,8 @@ public class GcpCloudSqlProperties {
 
 	private String region;
 
+	private String instanceConnectionName;
+
 	private String userName = "root";
 
 	private String password = "";
@@ -66,6 +68,14 @@ public class GcpCloudSqlProperties {
 
 	public void setRegion(String region) {
 		this.region = region;
+	}
+
+	public String getInstanceConnectionName() {
+		return this.instanceConnectionName;
+	}
+
+	public void setInstanceConnectionName(String instanceConnectionName) {
+		this.instanceConnectionName = instanceConnectionName;
 	}
 
 	public String getUserName() {
--- a/spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/autoconfig/AppEngineCloudSqlJdbcUrlProvider.java
+++ b/spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/autoconfig/AppEngineCloudSqlJdbcUrlProvider.java
@@ -1,0 +1,60 @@
+/*
+ *  Copyright 2017 original author or authors.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package org.springframework.cloud.gcp.sql.autoconfig;
+
+import org.springframework.cloud.gcp.sql.CloudSqlJdbcUrlProvider;
+import org.springframework.cloud.gcp.sql.GcpCloudSqlProperties;
+import org.springframework.util.Assert;
+import org.springframework.util.StringUtils;
+
+/**
+ * Provides default JDBC driver class name and constructs the JDBC URL for Cloud SQL v2
+ * when running inside of Google App Engine Standard environment.
+ *
+ * @author Ray Tsang
+ */
+public class AppEngineCloudSqlJdbcUrlProvider implements CloudSqlJdbcUrlProvider {
+	private final String projectId;
+
+	private final GcpCloudSqlProperties properties;
+
+	public AppEngineCloudSqlJdbcUrlProvider(String projectId, GcpCloudSqlProperties properties) {
+		this.projectId = projectId;
+		this.properties = properties;
+		Assert.hasText(projectId,
+				"A project ID must be provided.");
+		Assert.hasText(properties.getDatabaseName(), "Database Name is required");
+		if (StringUtils.isEmpty(properties.getInstanceConnectionName())) {
+			Assert.hasText(properties.getInstanceName(),
+					"Instance Name is required, or specify Instance Connection Name directly");
+			Assert.hasText(properties.getRegion(), "Region is required, or specify Instance Connection Name directly");
+			properties.setInstanceConnectionName(
+					String.format("%s:%s:%s", projectId, properties.getRegion(), properties.getInstanceName()));
+		}
+	}
+
+	@Override
+	public String getJdbcDriverClass() {
+		return "com.mysql.jdbc.GoogleDriver";
+	}
+
+	@Override
+	public String getJdbcUrl() {
+		return String.format("jdbc:google:mysql://%s/%s",
+				this.properties.getInstanceConnectionName(), this.properties.getDatabaseName());
+	}
+}
--- a/spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/autoconfig/DefaultCloudSqlJdbcUrlProvider.java
+++ b/spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/autoconfig/DefaultCloudSqlJdbcUrlProvider.java
@@ -1,0 +1,85 @@
+/*
+ *  Copyright 2017 original author or authors.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package org.springframework.cloud.gcp.sql.autoconfig;
+
+import java.io.IOException;
+
+import com.google.api.services.sqladmin.SQLAdmin;
+
+import org.springframework.cloud.gcp.sql.CloudSqlJdbcUrlProvider;
+import org.springframework.cloud.gcp.sql.GcpCloudSqlProperties;
+import org.springframework.util.Assert;
+import org.springframework.util.StringUtils;
+
+/**
+ * Provides default JDBC driver class name and constructs the JDBC URL for Cloud SQL v2
+ * when running on local laptop, or in a VM-based environment (e.g., Google Compute
+ * Engine, Google Container Engine).
+ *
+ * @author Ray Tsang
+ */
+public class DefaultCloudSqlJdbcUrlProvider implements CloudSqlJdbcUrlProvider {
+	private final String projectId;
+
+	private final SQLAdmin sqlAdmin;
+
+	private final GcpCloudSqlProperties properties;
+
+	public DefaultCloudSqlJdbcUrlProvider(String projectId, SQLAdmin sqlAdmin, GcpCloudSqlProperties properties) {
+		this.projectId = projectId;
+		this.sqlAdmin = sqlAdmin;
+		this.properties = properties;
+		Assert.hasText(projectId,
+				"A project ID must be provided.");
+		if (StringUtils.isEmpty(properties.getInstanceConnectionName())) {
+			Assert.hasText(this.properties.getInstanceName(),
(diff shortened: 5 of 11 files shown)


pivotal-issuemaster:
@saturnism Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-gcp&pullRequestId=90)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-cloud/spring-cloud-gcp&pullRequestId=90) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@saturnism Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-gcp&pullRequestId=90)!

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/autoconfig/GcpCloudSqlAutoConfiguration.java
```diff
@@ -75,44 +76,36 @@ protected SQLAdmin sqlAdminService(CredentialsProvider credentialsProvider)
...
-
-		// Load MySQL driver to register it, so the user doesn't have to.
-		Class.forName("com.mysql.jdbc.Driver");
-
-		// Look for the region if the user didn't specify one.
-		String region = this.gcpCloudSqlProperties.getRegion();
-		if (!StringUtils.hasText(region)) {
-			region = sqlAdmin.instances().get(projectIdProvider.getProjectId(),
-					this.gcpCloudSqlProperties.getInstanceName()).execute().getRegion();
+		String appEngineVersion = System.getProperty("com.google.appengine.runtime.version");
```
meltsufin: Is this the best way to detect App Engine Standard? Is it documented somewhere?

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/autoconfig/GcpCloudSqlAutoConfiguration.java
```diff
@@ -75,44 +76,36 @@ protected SQLAdmin sqlAdminService(CredentialsProvider credentialsProvider)
...
-				this.gcpCloudSqlProperties.getDatabaseName(),
-				instanceConnectionName);
 	}
 
 	@Bean
 	@ConditionalOnMissingBean(DataSource.class)
 	public DataSource cloudSqlDataSource(CloudSqlJdbcUrlProvider cloudSqlJdbcUrlProvider)
-			throws GeneralSecurityException, IOException {
+			throws GeneralSecurityException, IOException, ClassNotFoundException {
+		Class.forName(cloudSqlJdbcUrlProvider.getJdbcDriverClass());
```
joaoandremartins: The issue with loading the class here is that it won't be loaded if the user wants to use their own `DataSource`, instead of HikariCP.

It should be moved back to `cloudSqlJdbcUrlProvider()`.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/GcpCloudSqlProperties.java
```diff
@@ -68,6 +70,14 @@ public void setRegion(String region) {
...
 	}
 
+	public String getInstanceConnectionName() {
+		return this.instanceConnectionName;
+	}
+
+	public void setInstanceConnectionName(String instanceConnectionName) {
+		this.instanceConnectionName = instanceConnectionName;
+	}
+
```
meltsufin: What do you think of adding an optional `projectId` property here in case the database you're connecting to is in a different project?

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/autoconfig/AppEngineCloudSqlJdbcUrlProvider.java
```diff
@@ -0,0 +1,60 @@
...
+					"Instance Name is required, or specify Instance Connection Name directly");
+			Assert.hasText(properties.getRegion(), "Region is required, or specify Instance Connection Name directly");
+			properties.setInstanceConnectionName(
+					String.format("%s:%s:%s", projectId, properties.getRegion(), properties.getInstanceName()));
+		}
+	}
+
+	@Override
+	public String getJdbcDriverClass() {
+		return "com.mysql.jdbc.GoogleDriver";
```
meltsufin: Is this driver in the same jar as the non-AE one?

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/autoconfig/DefaultCloudSqlJdbcUrlProvider.java
```diff
@@ -0,0 +1,85 @@
...
+
+	private final GcpCloudSqlProperties properties;
+
+	public DefaultCloudSqlJdbcUrlProvider(String projectId, SQLAdmin sqlAdmin, GcpCloudSqlProperties properties) {
+		this.projectId = projectId;
+		this.sqlAdmin = sqlAdmin;
+		this.properties = properties;
+		Assert.hasText(projectId,
+				"A project ID must be provided.");
+		if (StringUtils.isEmpty(properties.getInstanceConnectionName())) {
```
meltsufin: This whole block of code for determining the connection name is duplicated between the two JDBC URL providers. It should be factored out.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/autoconfig/AppEngineCloudSqlJdbcUrlProvider.java
```diff
@@ -0,0 +1,60 @@
...
+	public AppEngineCloudSqlJdbcUrlProvider(String projectId, GcpCloudSqlProperties properties) {
+		this.projectId = projectId;
+		this.properties = properties;
+		Assert.hasText(projectId,
+				"A project ID must be provided.");
+		Assert.hasText(properties.getDatabaseName(), "Database Name is required");
+		if (StringUtils.isEmpty(properties.getInstanceConnectionName())) {
+			Assert.hasText(properties.getInstanceName(),
+					"Instance Name is required, or specify Instance Connection Name directly");
+			Assert.hasText(properties.getRegion(), "Region is required, or specify Instance Connection Name directly");
```
joaoandremartins: Should we also try to discover the region here?

https://cloud.google.com/appengine/docs/admin-api/reference/rest/v1/apps.locations/get

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/autoconfig/GcpCloudSqlAutoConfiguration.java
joaoandremartins: I was thinking we could leverage `@ConditionalOnProperty(name, matchIfMissing)` and `@AutoConfigureBefore` in another `AutoConfiguration` file.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/CloudSqlJdbcUrlProvider.java
```diff
@@ -22,8 +22,7 @@
  *
  * @author João André Martins
  */
-@FunctionalInterface
 public interface CloudSqlJdbcUrlProvider {
```
joaoandremartins: How about renaming to `CloudSqlJdbcProvider` or `CloudSqlJdbcInfoProvider`?

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/autoconfig/AppEngineCloudSqlJdbcUrlProvider.java
joaoandremartins: No, it's in a different one:

```
<dependency>
  <artifactId>mysql-socket-factory-connector-j-6</artifactId>
  <version>1.0.3</version>
</dependency>
```

I wonder if it would make sense to log an error pointing to https://cloud.google.com/appengine/docs/flexible/java/using-cloud-sql if we fail to load this class.

meltsufin: So, the assumption is that this dependency is provided at runtime? Maybe we can then auto-configure the JDBC URL based on the presence of this class rather than trying to detect AE Standard some other way.

joaoandremartins: I don't think that works if the dependency exists, but the code isn't running on App Engine Standard.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/autoconfig/GcpCloudSqlAutoConfiguration.java
artembilan: Agreed. Something like `AppEngineCloudSqlAutoConfiguration`. 
Then if `AppEngineCloudSqlJdbcUrlProvider` bean is configured by the condition, this one won't be registered because of `@ConditionalOnMissingBean`

artembilan:
All other comments are fine with me.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/autoconfig/AppEngineCloudSqlJdbcUrlProvider.java
saturnism: I tried it out but it wasn't doable from the SQLAdmin API. Was getting permission denied. :(

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/autoconfig/GcpCloudSqlAutoConfiguration.java
saturnism: In #89 it points to https://cloud.google.com/appengine/docs/standard/java/cloud-sql/ which has the snippet.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/autoconfig/DefaultCloudSqlJdbcUrlProvider.java
saturnism: it can be a necessary evil, since each of the providers should do its own validation. We can make a base class, but that seems slightly trivial in this case.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/autoconfig/GcpCloudSqlAutoConfiguration.java
saturnism: Nice, can the conditional match the string prefix too? This is needed because you can have the property, but w/ a different value, which means different thing. It has to match based on the string prefix.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/CloudSqlJdbcUrlProvider.java
saturnism: That sounds reasonable if we don't use a functional interface.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/autoconfig/GcpCloudSqlAutoConfiguration.java
```diff
@@ -75,44 +76,36 @@ protected SQLAdmin sqlAdminService(CredentialsProvider credentialsProvider)
 
 		return new SQLAdmin.Builder(httpTransport, jsonFactory,
 				new HttpCredentialsAdapter(credentialsProvider.getCredentials()))
-				.build();
+						.build();
```
joaoandremartins: Does this pass code check?

artembilan: I think there is just no rule on the matter and it even can't be possible IMO.

OTOH this looks better than previous one

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/autoconfig/GcpCloudSqlAutoConfiguration.java
artembilan: We always can do custom `Conditional` if that. However maybe that really would be better to make a separate `starter` at all as @meltsufin suggests: different environment - their own stuff

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/GcpCloudSqlProperties.java
joaoandremartins: That's a good point. In fact, every starter should probably specify their own project ID in case that service is running on a different project that's not the default one.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/autoconfig/GcpCloudSqlAutoConfiguration.java
joaoandremartins: I actually don't think it should be a separate starter. This is already the "gcp" starter, it makes sense that it's smart enough to distinguish between our different compute environments (but maybe in different AutoConfiguration objects and I mentioned before). For discovering the project ID if the user doesn't specify one, we're already trying out a bunch of different locations depending on the compute environment, by delegating to `ServiceOptions`.

For environments where a lot of things change, like PCF, it could probably make sense to have a new starter, although I'm still not too convinced, since that starter would likely span a bunch of different products (SQL, Pub/Sub, Core, etc.)...

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/autoconfig/GcpCloudSqlAutoConfiguration.java
saturnism: it was auto-formatted w/ the Eclipse rules :)

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/GcpCloudSqlProperties.java
saturnism: +1

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/autoconfig/GcpCloudSqlAutoConfiguration.java
saturnism: if we move this back to cloudSqlJdbcUrlProvider(), then the question is who loads the class if someone else provides a Provider instance? The lifecycle needs to be defined clearly. Class must be loaded when we configure a new data source, but not when we provide information.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/autoconfig/GcpCloudSqlAutoConfiguration.java
```diff
@@ -75,46 +76,35 @@ protected SQLAdmin sqlAdminService(CredentialsProvider credentialsProvider)
...
-		String region = this.gcpCloudSqlProperties.getRegion();
-		if (!StringUtils.hasText(region)) {
-			region = sqlAdmin.instances().get(projectIdProvider.getProjectId(),
-					this.gcpCloudSqlProperties.getInstanceName()).execute().getRegion();
-		}
-
-		String instanceConnectionName = projectIdProvider.getProjectId() + ":" + region + ":"
-				+ this.gcpCloudSqlProperties.getInstanceName();
+	@ConditionalOnMissingBean(CloudSqlJdbcInfoProvider.class)
+	@Conditional(AppEngineCondition.class)
```
saturnism: @artembilan added AppEngineCondition here, but just wnat to make sure this override makes sense, since we also have another ConditionalOnMissingBean just below it.

artembilan: That's easy to check:
````java
@SpringBootConfiguration
public class OverrideApplication {

	public static void main(String[] args) {
		ConfigurableApplicationContext applicationContext = new SpringApplicationBuilder(OverrideApplication.class)
				.run(args);
		Foo bean = applicationContext.getBean(Foo.class);
		System.out.println(bean.getFoo());

		applicationContext = new SpringApplicationBuilder(OverrideApplication.class)
				.properties("bar=bar")
				.run(args);
		bean = applicationContext.getBean(Foo.class);
		System.out.println(bean.getFoo());
	}

	@Bean
	@ConditionalOnMissingBean
	@ConditionalOnProperty("bar")
	public Foo fooConditional() {
		return new Foo("bar");
	}


	@Bean
	@ConditionalOnMissingBean
	public Foo fooDefault() {
		return new Foo("foo");
	}

	public static final class Foo {

		private final String foo;

		public Foo(String foo) {
			this.foo = foo;
		}

		public String getFoo() {
			return this.foo;
		}

	}

}
````

The first one prints `foo` and second `bar` - as expected.
Therefore your solution is correct.

File: spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/src/main/java/org/springframework/cloud/gcp/core/AppEngineCondition.java [hunk 10]
```diff
@@ -0,0 +1,43 @@
...
+
+import org.springframework.context.annotation.Condition;
+import org.springframework.context.annotation.ConditionContext;
+import org.springframework.core.type.AnnotatedTypeMetadata;
+import org.springframework.util.StringUtils;
+
+/**
+ * Determine whether this is running inside of an App Engine environment by looking into
+ * the System properties `com.google.appengine.runtime.version`.
+ * @author
```
joaoandremartins: Fix author

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/autoconfig/GcpCloudSqlAutoConfiguration.java
```diff
@@ -75,46 +76,35 @@ protected SQLAdmin sqlAdminService(CredentialsProvider credentialsProvider)
...
 
 	@Bean
-	@ConditionalOnMissingBean(CloudSqlJdbcUrlProvider.class)
-	public CloudSqlJdbcUrlProvider cloudSqlJdbcUrlProvider(SQLAdmin sqlAdmin,
-			GcpProjectIdProvider projectIdProvider)
-			throws IOException, ClassNotFoundException {
-		Assert.hasText(projectIdProvider.getProjectId(),
-				"A project ID must be provided.");
-
-		// TODO(joaomartins): Set the credential factory to be used by SocketFactory when
```
joaoandremartins: Can you leave this TODO?

saturnism: oh yes! i'll add this back in.

PR #99, opened 2017-08-11
Adds logging to Boot auto-discovery

### LINKED ISSUE #87: Add log statements to warn users about auto-discovered values
This is valid for project ID, Cloud SQL instance region, credentials, etc.
We should warn the user by logging the auto-discovered values that are being used when an app is run.

joaoandremartins:
Helps users making sure that they are connecting to the desired
components, such as project IDs, SQL instances, etc.

Fixes #87

### DIFF
--- a/spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/src/main/java/org/springframework/cloud/gcp/core/autoconfig/GcpContextAutoConfiguration.java
+++ b/spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/src/main/java/org/springframework/cloud/gcp/core/autoconfig/GcpContextAutoConfiguration.java
@@ -85,16 +85,18 @@ public class GcpContextAutoConfiguration {
 
 		Credentials credentials = credentialsProvider.getCredentials();
 
-		if (credentials instanceof UserCredentials) {
-			LOGGER.info("Default credentials provider for user "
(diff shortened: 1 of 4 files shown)


spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/src/main/java/org/springframework/cloud/gcp/core/autoconfig/GcpContextAutoConfiguration.java
```diff
@@ -55,22 +61,43 @@
...
+			credentialsProvider = GoogleCredentialsProvider.newBuilder()
+					.setScopesToApply(SCOPES_LIST)
+					.build();
+		}
+
+		Credentials credentials = credentialsProvider.getCredentials();
+
+		if (credentials instanceof UserCredentials) {
+			LOGGER.info("Default credentials provider for user "
+					+ ((UserCredentials) credentials).getClientId());
```
artembilan: everything less then `warn` should be wrapped to the `LOGGER.is...Enable()` to avoid extra computation for the string concatenation.

joaoandremartins:
@markfisher in this case, Travis has failed but Github still let's me squash&merge.

Can you turn on the option to only allow merges if CI doesn't fail.

PR #102, opened 2017-08-14
Documents undocumented parts of spring-cloud-gcp

joaoandremartins:
Things like property names aren't configured anywhere yet. This is one
step while we don't have a reference document with more comprehensive
documentation.

Fixes #100

Issue #47, opened 2017-06-05
Generate meta-data for the @ConfigurationProperties


artembilan:
http://docs.spring.io/spring-boot/docs/1.5.3.RELEASE/reference/htmlsingle/#configuration-metadata-annotation-processor

joaoandremartins:
I think this is currently happening for the two places we have `@ConfigurationProperties` at: our Core and SQL starters.

I'm thinking we should probably transform the `@Values` we have in https://github.com/spring-cloud/spring-cloud-gcp/blob/master/spring-cloud-gcp-starters/spring-cloud-gcp-starter-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/autoconfig/GcpPubSubAutoConfiguration.java#L56 to `@ConfigurationProperties`.

`@Value` doesn't do that trick where `executorThreads` is aliased to `executor-threads`, which is the rule with our other properties.

meltsufin:
What do you mean by aliasing to `executor-threads`? If the meta-data is not generated for `@Value`, let's just switch to structured properties like you're suggesting.

artembilan:
I think you need to take a look here: https://docs.spring.io/spring-boot/docs/1.5.7.RELEASE/reference/html/boot-features-external-config.html#boot-features-external-config-vs-value

The most important feature there is about https://docs.spring.io/spring-boot/docs/1.5.7.RELEASE/reference/html/boot-features-external-config.html#boot-features-external-config-relaxed-binding

joaoandremartins:
Yes, relaxed binding is exactly what I meant. I didn't know there were even other options like underscore and upper case notation. Thanks @artembilan !

I might file a PR on this soon so we get parity across all our config.

PR #106, opened 2017-08-15
Adds Configuration Client Starter for Google Runtime Configurator : spring-cloud-gcp-starter-config

### LINKED ISSUE #95: Implement `spring-cloud-config` client adapter for Google Cloud Runtime Configurator
[Runtime Configurator API](https://cloud.google.com/deployment-manager/runtime-configurator/reference/rest/) allows for dynamic application configuration.
An adapter to [spring-cloud-config](https://cloud.spring.io/spring-cloud-config/) client will allow developers to easily use the Runtime Configurator as their configuration server.
A prototype implementation currently exists [here](https://github.com/GoogleCloudPlatform/rcloadenv/pull/2).

jabubake:
- TODO : Unit Tests
- Also, I'm not clear if this needs a starter, or how that would need to be done, given it needs to be provided in the `BootstrapConfiguration`.
- I introduced `google-cloud-gcp-examples` to describe the usage of `google-cloud-gcp-config` : might not the best place, do you have a planned location to place these samples ? 

Fixes #95


joaoandremartins:
I'm not at all against a `spring-cloud-gcp-examples` module, I actually think it's a good idea.

spring-cloud-gcp-config/README.md
```diff
@@ -0,0 +1,13 @@
+# Spring Cloud Config Client integration for Google Cloud Runtime Configurator API
```
joaoandremartins: Sorry for the annoyance but we're going with asciidoc for READMEs. I think that's a Spring standard. Can you convert this?

spring-cloud-gcp-config/pom.xml
```diff
@@ -0,0 +1,50 @@
...
+        <version>1.0.0.BUILD-SNAPSHOT</version>
+    </parent>
+    <artifactId>spring-cloud-gcp-config</artifactId>
+    <name>Spring Cloud GCP Configuration Client</name>
+    <description>Spring Cloud GCP Configuration Client for Runtime Configurator</description>
+    <properties>
+        <main.basedir>${basedir}/../..</main.basedir>
+    </properties>
+
+    <dependencyManagement>
```
joaoandremartins: I think this should already be done automatically with our parent pom?

spring-cloud-gcp-config/pom.xml
```diff
@@ -0,0 +1,50 @@
...
+
+    <dependencies>
+        <dependency>
+            <groupId>org.springframework.cloud</groupId>
+            <artifactId>spring-cloud-starter-config</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>com.google.cloud</groupId>
+            <artifactId>google-cloud-core</artifactId>
+            <version>1.2.3</version>
```
joaoandremartins: These versions are either not needed or should be in our BOM instead.

spring-cloud-gcp-config/src/main/java/org/springframework/cloud/gcp/config/GoogleConfigEnvironment.java
```diff
@@ -0,0 +1,35 @@
...
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package org.springframework.cloud.gcp.config;
+
+import java.util.List;
+
+final class GoogleConfigEnvironment {
```
joaoandremartins: The Spring style guide mandates that you at least add Javadoc with `@author`.

spring-cloud-gcp-config/src/main/java/org/springframework/cloud/gcp/config/GoogleConfigPropertySourceLocator.java
```diff
@@ -0,0 +1,171 @@
...
+
+	private ConfigClientProperties defaultProperties;
+
+	private GoogleCredentials credentials;
+
+	private String projectId;
+
+	public GoogleConfigPropertySourceLocator(ConfigClientProperties defaultProperties) throws IOException {
+		this.defaultProperties = defaultProperties;
+		this.credentials = ServiceAccountCredentials.getApplicationDefault()
```
joaoandremartins: Should we instead add a `CredentialsProvider` to the ctor args and use that? In order to use the credentials specified by the user.

spring-cloud-gcp-config/src/main/java/org/springframework/cloud/gcp/config/GoogleConfigPropertySourceLocator.java
```diff
@@ -0,0 +1,171 @@
...
+				.createScoped(Collections.singleton(RUNTIMECONFIG_SCOPE));
+		this.projectId = ServiceOptions.getDefaultProjectId();
```
joaoandremartins: Likewise for `projectId`.

spring-cloud-gcp-config/src/main/java/org/springframework/cloud/gcp/config/Variable.java
```diff
@@ -0,0 +1,80 @@
...
+ *  limitations under the License.
+ */
+
+package org.springframework.cloud.gcp.config;
+
+import java.nio.charset.StandardCharsets;
+
+import javax.xml.bind.DatatypeConverter;
+
+final class Variable {
```
joaoandremartins: Javadoc here. Also, not immediately clear what the purpose of this class is.

spring-cloud-gcp-examples/spring-cloud-gcp-config-example/README.md
```diff
@@ -0,0 +1,83 @@
+# Spring Cloud Config Client for Google Cloud Runtime Configurator API
```
joaoandremartins: Ditto for adoc. Should we also rename the module to example**s**?

spring-cloud-gcp-examples/spring-cloud-gcp-config-example/README.md
```diff
@@ -0,0 +1,83 @@
...
+
+This sample application demonstrates using the `Spring Cloud Config Client for GCP` in your code.
+
+## Setup
+1. Install the [Google Cloud SDK](https://cloud.google.com/sdk/). You
```
joaoandremartins: This isn't actually required, for example, if users use the core Boot starter and specify the credentials using `spring.cloud.gcp.credentials-location` property.

spring-cloud-gcp-examples/spring-cloud-gcp-config-example/README.md
```diff
@@ -0,0 +1,83 @@
...
+```
+  gcloud beta runtime-config configs variables set \
+    queue_size 25 \
+    --config-name myapp_prod
+  gcloud beta runtime-config configs variables set \
+    feature_x_enabled true \
+    --config-name myapp_prod
+```
+
+1. In your Spring Boot application directory ,
```
joaoandremartins: It sounds very much like we could turn this into a Boot starter.

spring-cloud-gcp-examples/spring-cloud-gcp-config-example/README.md
```diff
@@ -0,0 +1,83 @@
...
+add the following line to [src/main/resources/META-INF/spring.factories](src/main/resources/META-INF/spring.factories) :
+```
+  org.springframework.cloud.bootstrap.BootstrapConfiguration=org.springframework.cloud.gcp.config.GoogleConfigPropertySourceLocator
+```
+
+1. Add the following dependency to your [pom.xml](pom.xml):
+```
+  <dependency>
```
joaoandremartins: Suggestion: you might want to follow the pattern of the Maven and Gradle snippet in the other READMEs (#102). If a user uses our BOM, they won't need the `version` either.

spring-cloud-gcp-examples/spring-cloud-gcp-config-example/README.md
```diff
@@ -0,0 +1,83 @@
...
+```
+
+1. (Optional) [Spring Boot Actuator](http://cloud.spring.io/spring-cloud-static/docs/1.0.x/spring-cloud.html#_endpoints)
+ provides support to have configuration parameters be reloadable with the POST `/refresh` endpoint.
+    
+    1. Add the following dependency to your `pom.xml`:
+        ```
+          <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-actuator</artifactId>
```
joaoandremartins: No version here comprehends the user should be using a BOM?

spring-cloud-gcp-examples/spring-cloud-gcp-config-example/pom.xml
```diff
@@ -0,0 +1,44 @@
...
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <groupId>com.example</groupId>
+    <artifactId>spring-gcp-config-example</artifactId>
+    <version>1.0.0.BUILD-SNAPSHOT</version>
+
+    <parent>
+        <groupId>org.springframework.boot</groupId>
```
joaoandremartins: How about using no parent at all?

spring-cloud-gcp-examples/spring-cloud-gcp-config-example/pom.xml
```diff
@@ -0,0 +1,44 @@
...
+            <version>1.0.0.BUILD-SNAPSHOT</version>
+        </dependency>
+
+        <!-- Optional dependency to enable runtime property refresh -->
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-actuator</artifactId>
+        </dependency>
+    </dependencies>
+    <build>
```
joaoandremartins: Add our BOM somewhere?

jabubake:
@joaoandremartins : I still need to add tests. I redid this as a starter

spring-cloud-gcp-starters/spring-cloud-gcp-starter-config/src/main/java/org/springframework/cloud/gcp/config/GoogleConfigPropertySourceLocator.java
```diff
@@ -0,0 +1,138 @@
...
+	private GoogleConfigEnvironment getRemoteEnvironment(RestTemplate restTemplate, ConfigClientProperties properties)
+			throws IOException, HttpClientErrorException {
+		String name = properties.getName();
+		String profile = properties.getProfile();
+
+		ResponseEntity<GoogleConfigEnvironment> response = restTemplate.exchange(
+				RUNTIMECONFIG_API_ROOT + ALL_VARIABLES_PATH, HttpMethod.GET,
+				null, GoogleConfigEnvironment.class, this.projectId, name, profile);
+
+		Assert.isTrue(response != null && response.getStatusCode().is2xxSuccessful(),
```
joaoandremartins: Should we throw an exception other than `IllegalArgumentException` here?

spring-cloud-gcp-starters/spring-cloud-gcp-starter-config/src/main/java/org/springframework/cloud/gcp/config/GoogleConfigPropertySourceLocator.java
```diff
@@ -0,0 +1,138 @@
...
+				"Invalid response from Runtime Configurator API");
+		return response.getBody();
+	}
+
+	@Override
+	public PropertySource<?> locate(Environment environment) {
+		ConfigClientProperties properties = this.defaultProperties.override(environment);
+		Map<String, Object> config = Collections.emptyMap();
```
joaoandremartins: Redundant `Collections.emptyMap()`.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-config/src/main/java/org/springframework/cloud/gcp/config/GoogleConfigPropertySourceLocator.java
```diff
@@ -0,0 +1,138 @@
...
+		Assert.notNull(projectIdProvider, "Project ID provider cannot be null");
+		Assert.notNull(defaultProperties, "Config client properties cannot be null");
+		this.credentials = credentialsProvider.getCredentials();
+		this.projectId = projectIdProvider.getProjectId();
+		this.defaultProperties = defaultProperties;
+	}
+
+	private RestTemplate getSecureRestTemplate() throws Exception {
+		SimpleClientHttpRequestFactory requestFactory = new SimpleClientHttpRequestFactory();
+		requestFactory.setReadTimeout((60 * 1000 * 3) + 5000);
```
joaoandremartins: Should the timeout be configurable?

spring-cloud-gcp-starters/spring-cloud-gcp-starter-config/src/main/java/org/springframework/cloud/gcp/config/GoogleConfigPropertySourceLocator.java
```diff
@@ -0,0 +1,138 @@
...
+
+	private String projectId;
+
+	public GoogleConfigPropertySourceLocator(CredentialsProvider credentialsProvider,
+			GcpProjectIdProvider projectIdProvider, ConfigClientProperties defaultProperties) throws Exception {
+		Assert.notNull(credentialsProvider, "Credentials provider cannot be null");
+		Assert.notNull(projectIdProvider, "Project ID provider cannot be null");
+		Assert.notNull(defaultProperties, "Config client properties cannot be null");
+		this.credentials = credentialsProvider.getCredentials();
+		this.projectId = projectIdProvider.getProjectId();
```
joaoandremartins: Should we check that the returned `credentials` and `projectId` are also not null?

spring-cloud-gcp-starters/spring-cloud-gcp-starter-config/src/main/resources/META-INF/spring.factories
```diff
@@ -0,0 +1,4 @@
+#org.springframework.boot.autoconfigure.EnableAutoConfiguration=\
```
joaoandremartins: Remove comments?

pom.xml
```diff
@@ -49,6 +48,13 @@
 				<type>pom</type>
 				<scope>import</scope>
 			</dependency>
+			<dependency>
```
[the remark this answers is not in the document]
jabubake: Looks like I had made a mess and not done it the expected way in BOM, I've cleaned it up, PTAL

spring-cloud-gcp-examples/spring-cloud-gcp-config-example/README.adoc
```diff
@@ -0,0 +1,98 @@
...
++
+....
+  curl https://sdk.cloud.google.com | bash
+....
+2.  Create, enable billing on a project on the
+https://console.cloud.google.com[Google Cloud Console].
+3.  Enable the
+https://console.cloud.google.com/flows/enableapi?apiid=runtimeconfig.googleapis.com[Runtime
+Config API]
+4.  https://cloud.google.com/authentication/docs[Set up authentication]
```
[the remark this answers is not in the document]
jabubake: I see how it looked confusing, added comment on why we ask users to install Cloud SDK.
Moved most of this content to the starter README

spring-cloud-gcp-examples/spring-cloud-gcp-config-example/README.adoc
```diff
@@ -0,0 +1,98 @@
...
+<dependency>
+  <groupId>org.springframework.cloud</groupId>
+  <artifactId>spring-cloud-gcp-config-starter</artifactId>
+  <version>1.0.0.BUILD-SNAPSHOT</version>
+</dependency>
+....
+3.  Update src/main/resources/application.properties:
++
+....
+spring.cloud.config.name = "myapp" // (optional, "spring.application.name" is used, if not provided)
```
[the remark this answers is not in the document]
jabubake: Added, will fallback to spring.cloud.config.{name,profile}

spring-cloud-gcp-examples/spring-cloud-gcp-config-example/pom.xml
```diff
@@ -0,0 +1,42 @@
...
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-web</artifactId>
+        </dependency>
+
+        <!-- Configuration Client Starter for Google Cloud Runtime Configurator API -->
+        <dependency>
+            <groupId>org.springframework.cloud</groupId>
+            <artifactId>spring-cloud-gcp-starter-config</artifactId>
+            <version>${project.version}</version>
```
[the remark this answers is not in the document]
jabubake: Fixed in BOM

spring-cloud-gcp-starters/spring-cloud-gcp-starter-config/README.adoc
```diff
@@ -0,0 +1,21 @@
+[[spring-cloud-config-client-integration-for-google-cloud-runtime-configurator-api]]
+Spring Cloud Config Client integration for Google Cloud Runtime
+Configurator API
+--------------------------------------------------------------------------------
```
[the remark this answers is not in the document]
jabubake: Fixed.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-config/README.adoc
```diff
@@ -0,0 +1,21 @@
...
+
+https://cloud.google.com/deployment-manager/runtime-configurator/[Runtime
+Configuration API] on Google Cloud Platform allows dynamic configuration
+of your services.
+
+https://cloud.spring.io/spring-cloud-config/[Spring Cloud Config Client]
+provides developers the ability to bootstrap their application
+configuration with property sources.
```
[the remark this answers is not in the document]
jabubake: not necessarily, can be from file-based sources too. If there is better language to capture that, can add : still mulling over a better way to state this.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-config/README.adoc
```diff
@@ -0,0 +1,21 @@
...
+
+This integration uses the
+https://cloud.google.com/deployment-manager/runtime-configurator/reference/rest/[Runtime
+Configurator REST SDK] to
+[bootstrap](http://projects.spring.io/spring-cloud/spring-cloud.html#customizing-bootstrap-property-sources[integrates]
```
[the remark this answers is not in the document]
jabubake: fixed

spring-cloud-gcp-starters/spring-cloud-gcp-starter-config/src/main/java/org/springframework/cloud/gcp/config/GoogleConfigPropertySourceLocator.java
```diff
@@ -0,0 +1,138 @@
...
+import org.springframework.web.client.HttpClientErrorException;
+import org.springframework.web.client.RestTemplate;
+
+/**
+ * Custom property source locator for Google Cloud Runtime Configurator.
+ *
+ * @author Jisha Abubaker
+ */
+@Configuration
+@ComponentScan("org.springframework.cloud.gcp.core")
```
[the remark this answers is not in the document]
jabubake: This was a bug, updated to include the autoconfig classes in the bootstrap configuration.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-config/src/main/java/org/springframework/cloud/gcp/config/GoogleConfigPropertySourceLocator.java
jabubake: done.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-config/src/main/java/org/springframework/cloud/gcp/config/GoogleConfigPropertySourceLocator.java
jabubake: done.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-config/src/main/java/org/springframework/cloud/gcp/config/GoogleConfigPropertySourceLocator.java
jabubake: Agree, fixed.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-config/src/main/java/org/springframework/cloud/gcp/config/GoogleConfigPropertySourceLocator.java
jabubake: removed.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-config/src/main/resources/META-INF/spring.factories
jabubake: updated

spring-cloud-gcp-dependencies/pom.xml
```diff
@@ -91,8 +92,14 @@
 				<version>1.0.4</version>
 			</dependency>
 
-			<!--Starters-->
+      <!-- spring-cloud-gcp-starter-config -->
+      <dependency>
```
joaoandremartins: Nit: checkstyle isn't complaining probably because this is an xml file. Can you fix the indentation here?

spring-cloud-gcp-examples/spring-cloud-gcp-config-example/src/main/resources/bootstrap.properties
```diff
@@ -0,0 +1,16 @@
...
+
+# `spring.cloud.config.name` will be used if spring.cloud.gcp.config.name is not provided
+# spring.application.name is used if neither spring.cloud.config.name or spring.cloud.gcp.config.name is provided
+spring.cloud.gcp.config.name=myapp
+
+# `spring.cloud.config.profile` will be used if spring.cloud.gcp.config.profile is not provided
+spring.cloud.gcp.config.profile=prod
+
+# Config server API timeout in milliseconds, default : 18000 ( 3 minutes)
+spring.cloud.gcp.config.timeout = 12000
```
joaoandremartins: Crazy nit: this equal separator has spaces around and the previous two don't. Sorry :)

spring-cloud-gcp-starters/spring-cloud-gcp-starter-config/src/main/java/org/springframework/cloud/gcp/config/GcpConfigProperties.java
```diff
@@ -0,0 +1,59 @@
...
+package org.springframework.cloud.gcp.config;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+/**
+ * Configuration for {@link GoogleConfigPropertySourceLocator}.
+ *
+ * @author Jisha Abubaker
+ */
+@ConfigurationProperties("spring.cloud.gcp.config")
```
joaoandremartins: Is the duplication of name and profile from here and `ConfigClientProperties` intended?

joaoandremartins: If we're also using `ConfigClientProperties`, wouldn't it be better to just stick to it for `name` and `profile`?
I noticed that `ConfigClientProperties` doesn't contain a `timeout` field, I think it's fair to use `GcpConfigProperties` just for that bit (since arguably, it's the gcp-specific configuration).

spring-cloud-gcp-starters/spring-cloud-gcp-starter-config/README.adoc
```diff
@@ -0,0 +1,134 @@
...
+....
+  gcloud beta runtime-config configs variables set \
+queue_size 25 \
+--config-name myapp_prod
+  gcloud beta runtime-config configs variables set \
+feature_x_enabled true \
+--config-name myapp_prod
+....
+
+3.  Create / update `src/main/resources/bootstrap.properties` to specify the configuration name, profile a:
```
joaoandremartins: Is the sentence incomplete? Did you mean "profile and timeout"? You'll also need another new line to the "Note:" shows up in a different paragraph.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-config/src/main/java/org/springframework/cloud/gcp/config/GcpConfigProperties.java
```diff
@@ -0,0 +1,39 @@
...
+public class GcpConfigProperties {
+
+	// Config server API time out in milliseconds, default = 3 minutes
+	private int timeout = 180000;
```
joaoandremartins: I wonder if the timeout should be something shorter, like a minute, instead?

spring-cloud-gcp-starters/spring-cloud-gcp-starter-config/README.adoc
```diff
@@ -0,0 +1,134 @@
...
+Note: If using the
+link:../../spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/README.adoc[GCP Starter Core properties]
+to configure the GCP project and credentials, they must be included
+in the `boostrap.properties`, and not `application.properties`.
```
joaoandremartins: Do you mean the project ID and credentials should also go in `bootstrap.properties`? Or only there?

spring-cloud-gcp-starters/spring-cloud-gcp-starter-config/src/main/resources/META-INF/spring.factories
```diff
@@ -0,0 +1,4 @@
+org.springframework.cloud.bootstrap.BootstrapConfiguration=\
```
joaoandremartins: I'm still wrapping my head around the bootstrap context concept which I hadn't seen before. I guess we need this here because of that, but I think it would also be nice to add `org.springframework.boot.autoconfigure.EnableAutoConfiguration` to this file, so that the IDEs can auto-complete the name of the properties.

I tried this locally but the IDE still isn't auto-completing the property name. I'm still trying to figure out why...

spring-cloud-gcp-starters/spring-cloud-gcp-starter-config/README.adoc
```diff
@@ -0,0 +1,134 @@
...
++
+....
+  gcloud beta runtime-config configs variables set \
+     queue_size 200 \
+     --config-name myapp_prod
+....
+4.  Send a POST request to the `/refresh` endpoint:
++
+....
+  curl -XPOST http://myapp.host.com/refresh`
```
joaoandremartins: Nit: extra ` at the end

spring-cloud-gcp-starters/spring-cloud-gcp-starter-config/README.adoc
```diff
@@ -0,0 +1,134 @@
...
+The  default value of `spring.cloud.config.profile` is `default`.
+
+`spring.cloud.gcp.config.timeout` configures the timeout in milliseconds for requests sent to the
+Google Cloud Runtime Configurator API. This field is optional, and uses a default of 3 minutes.
+
+
+4.  Add Spring style configuration variables to your `@Configuration` annotated class.
++
+....
+@Value(${queue_size}) private int queueSize;
```
joaoandremartins: I think ${queue_size} has double quotes around, right?

joaoandremartins:
I'm trying this out and getting "c.c.c.ConfigServicePropertySourceLocator : Fetching config from server at: http://localhost:8888". I think the spring-cloud-starter-config module might be trying this?

jabubake:
It does that by default , there is a way to suppress it in the config, but I felt it was a harmless log statement and would not require additional config overhead. LMK.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-config/src/main/resources/META-INF/spring.factories
jabubake: For the config client, there is no value in providing the `EnableAutoConfiguration`, since its done in the application context, post bootstrap
Are you talking of providing an empty org.springframework.boot.autoconfigure.EnableAutoConfiguration ?

spring-cloud-gcp-starters/spring-cloud-gcp-starter-config/src/main/java/org/springframework/cloud/gcp/config/GcpConfigProperties.java
jabubake: Works for me, will update. I just picked up what the default config client uses.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-config/README.adoc
jabubake: Tx, good catch. fixed.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-config/README.adoc
jabubake: fixed

spring-cloud-gcp-starters/spring-cloud-gcp-starter-config/src/main/java/org/springframework/cloud/gcp/config/GoogleConfigPropertySourceLocator.java
```diff
@@ -0,0 +1,150 @@
...
+import org.springframework.util.Assert;
+import org.springframework.web.client.HttpClientErrorException;
+import org.springframework.web.client.RestTemplate;
+
+/**
+ * Custom {@link PropertySourceLocator} for Google Cloud Runtime Configurator API.
+ *
+ * @author Jisha Abubaker
+ */
+@Configuration
```
joaoandremartins: I don't think this `@Configuration` is necessary here. It's working well for me without it.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-config/src/main/resources/META-INF/spring.factories
joaoandremartins: If I recall correctly, pointing `EnableAutoConfiguration` here to our auto configuration class makes IDEs auto-suggest property names when you're typing.

But when I'm editing `application.properties`, if I add `spring.cloud.gcp.c`, the editor doesn't auto-suggest the option (as it does for sql and core).

joaoandremartins: I'm not sure why after adding the following to this file, the IDE still won't recognise the spring.cloud.gcp.config property prefix.

```
org.springframework.boot.autoconfigure.EnableAutoConfiguration=\
  org.springframework.cloud.gcp.config.autoconfig.GcpConfigAutoConfiguration
```

@artembilan do you have any idea why?

spring-cloud-gcp-starters/spring-cloud-gcp-starter-config/src/main/java/org/springframework/cloud/gcp/config/GoogleConfigPropertySourceLocator.java
jabubake: Took it out, I had based it off of the example [here](http://cloud.spring.io/spring-cloud-static/spring-cloud.html#customizing-bootstrap-property-sources), but looks like it works without it too.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-config/README.adoc
jabubake: I see the text was confusing, updated.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-config/src/main/java/org/springframework/cloud/gcp/config/GcpConfigProperties.java
jabubake: Updated to a minute.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-config/README.adoc
jabubake: Done.

spring-cloud-gcp-examples/spring-cloud-gcp-config-example/README.adoc
```diff
@@ -0,0 +1,98 @@
...
+[[testing]]
+Testing
+~~~~~~~
+
+1.  Create a configuration using the
+https://cloud.google.com/sdk/[Google Cloud SDK]. The configuration name
+should be in the format `config`_`profile`, for example : `myapp_prod`
++
+....
+gcloud beta runtime-config configs create myapp_prod
```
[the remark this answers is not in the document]
jabubake: Added note to state why we ask to install Cloud SDK.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-config/src/main/java/org/springframework/cloud/gcp/config/GoogleConfigPropertySourceLocator.java
```diff
@@ -0,0 +1,148 @@
...
+		Assert.notNull(this.configClientProperties, "Client Configuration must not be null");
+		Assert.notNull(this.timeout, "Runtime Configurator API timeout must not be null");
+	}
+
+	private RestTemplate getSecureRestTemplate() throws Exception {
+		SimpleClientHttpRequestFactory requestFactory = new SimpleClientHttpRequestFactory();
+		requestFactory.setReadTimeout(this.timeout);
+		RestTemplate template = new RestTemplate(requestFactory);
+		Map<String, String> headers = new HashMap<>();
+		Map<String, List<String>> requestMetadata = this.credentials.getRequestMetadata();
```
joaoandremartins: Can we log somewhere in this class (maybe here? or in the ctor?) what user we're authenticating as? Kind of like we do [here](https://github.com/spring-cloud/spring-cloud-gcp/blob/master/spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/src/main/java/org/springframework/cloud/gcp/core/autoconfig/GcpContextAutoConfiguration.java#L99).

I'm debugging an issue where the application default credentials are giving me 403 Forbiddens and knowing who the users are authenticating as here would be really useful to debug issues.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-config/src/main/resources/META-INF/spring.factories
spencergibb: auto-configuration doesn't enable ide autocompetion of properties. Property metadata, does that.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-config/src/main/java/org/springframework/cloud/gcp/config/autoconfig/GcpConfigAutoConfiguration.java
```diff
@@ -34,5 +37,12 @@
 
 	@Autowired
 	GcpConfigProperties gcpConfigProperties;
+
+	// Disabling the default Spring cloud config property source locator
+	@Bean
```
jabubake: @joaoandremartins @spencergibb @saturnism 
Please see if this is an acceptable solution to disable the config property source locator.

spencergibb: I'm confused at why you import spring-cloud-starter-config at all. I don't think you need it. Bootstrap comes from spring-cloud-commons which you could get with spring-cloud-starter

joaoandremartins: We want to reuse the same property namespace as `ConfigClientProperties`. We could import `spring-cloud-config-client` instead of the whole `spring-cloud-config-starter`, but the same problem would pose -- `ConfigServicePropertySourceLocator` would also be imported from the same package and it will try to run unless we tell it not to (and we don't want our users to do it through properties).

Basically, if our starter is active, we want to default `ConfigClientProperties.enabled` to false. Is this a good way to do so?

spencergibb: Reusing the same namespace seems like a bad idea.

jabubake: @spencergibb : If that is the case, I can simplify what's done here easily.

spring-cloud-gcp-dependencies/pom.xml
```diff
@@ -91,8 +92,14 @@
 				<version>1.0.4</version>
 			</dependency>
 
-			<!--Starters-->
+			<!-- spring-cloud-gcp-starter-config -->
+			<dependency>
+				<groupId>org.springframework.cloud</groupId>
+				<artifactId>spring-cloud-context</artifactId>
+				<version>${spring-cloud-context.version}</version>
```
jabubake: Depending on the release version here, PTAL @spencergibb @joaoandremartins 
Is there a better way to use a managed version of  `org.springframework.cloud:spring-cloud-context` ?

spring-cloud-gcp-examples/spring-cloud-gcp-config-example/README.adoc
```diff
@@ -0,0 +1,53 @@
+= Google Cloud Spring Config Client Starter Example
```
joaoandremartins: Spring Cloud GCP Config Client Starter Example?

spring-cloud-gcp-starters/spring-cloud-gcp-starter-config/README.adoc
```diff
@@ -0,0 +1,136 @@
+= Google Cloud Spring Config Client Starter
```
joaoandremartins: Spring Cloud GCP Config Client Starter?

jabubake: done

spring-cloud-gcp-examples/spring-cloud-gcp-config-example/README.adoc
jabubake: done

spring-cloud-gcp-dependencies/pom.xml
joaoandremartins: @jabubake what do you mean?

spring-cloud-gcp-starters/spring-cloud-gcp-starter-config/src/main/resources/META-INF/spring.factories
joaoandremartins: Says [here](https://docs.spring.io/spring-boot/docs/current/reference/html/configuration-metadata.html#configuration-metadata) that `@ConfigurationProperties` should take care of generating the configuration metadata, but my browser isn't reading it..

spencergibb: https://docs.spring.io/spring-boot/docs/current/reference/html/configuration-metadata.html#configuration-metadata-annotation-processor

spring-cloud-gcp-dependencies/pom.xml
jabubake: @joaoandremartins  : sorry for lack of context.
@spencergibb suggested using https://mvnrepository.com/artifact/org.springframework.cloud/spring-cloud-starter

Given we only need the bootstrap support, I've only used `spring-cloud-context` here.  Just confirming that this is fine and we don't need to include the starter in the `spring-cloud-gcp-dependencies` and use `spring-cloud-context` without a version in our GCP config starter.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-config/pom.xml
```diff
@@ -0,0 +1,33 @@
...
+			<artifactId>spring-cloud-context</artifactId>
+		</dependency>
+		<dependency>
+			<groupId>org.springframework</groupId>
+			<artifactId>spring-web</artifactId>
+		</dependency>
+		<dependency>
+			<groupId>org.springframework.cloud</groupId>
+			<artifactId>spring-cloud-gcp-starter-core</artifactId>
+		</dependency>
```
joaoandremartins: Let's add this here to generate property metadata:

		
		<dependency>
			<optional>true</optional>
		</dependency>

spring-cloud-gcp-dependencies/pom.xml
joaoandremartins: No version is fine, since it's in our BOM already. Also +1 for not using the starter but only what we need in this case.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-config/pom.xml
joaoandremartins: I think we're good on my part after this bit.

Just one more thing, are you thinking of adding unit tests here?

jabubake: Done. Looks like I have a merge conflict, fixing.
Also added an enabled flag (default = true), if users need to shut this off for local development

joaoandremartins:
Can you please also add a little entry to the top-level README for Spring Runtime Config?

I'll try to make one final pass today, since you've submitted unit tests.

jabubake:
@joaoandremartins Done.

joaoandremartins:
@jabubake it doesn't look like the tests were added, but let me know if I'm missing anything.

meltsufin:
@ryanjbaxter / @spencergibb,
@markfisher suggested that you might have some interest in reviewing this. PTAL, if you can. Thanks!

ryanjbaxter:
@meltsufin seems pretty nice.  Not sure if I have any use for it at the moment, but if you need any help integrating this with other Spring Cloud Core components let me know.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-config/src/main/resources/META-INF/spring.factories
artembilan: +1 what Spencer says. 
In our language it sounds like: "I told you!" 

https://github.com/spring-cloud/spring-cloud-gcp/issues/47

joaoandremartins:
It would be great if you could add something to our reference document after merging this PR. :)

jabubake:
Addresses #95 
@joaoandremartins, @meltsufin  : Please merge, I don't have permissions

joaoandremartins:
Thanks and well done @jabubake !

PR #108, opened 2017-08-16
Passes user credentials to Cloud SQL

### LINKED ISSUE #104: Send our credentials provider to the Cloud SQL socket factory
Now that https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/issues/41 is closed, we should fix the TODO in the [Cloud SQL Boot starter](https://github.com/spring-cloud/spring-cloud-gcp/blob/master/spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/autoconfig/DefaultCloudSqlJdbcInfoProvider.java#L81).

It's still unclear where the fix should go. I don't think the current location of the TODO is the right place to fix. It should go somewhere in `GcpCloudSqlAutoConfiguration`. The user can provide both their own `CloudSqlJdbcInfoProvider` and `DataSource`, but I would argue the Google magic is added in the `CloudSqlJdbcInfoProvider`, so we should probably add the fix to both the `Default` and `AppEngineCloudSqlJdbcInfoProvider`.

joaoandremartins:
Currently, Cloud SQL uses application default credentials. They just
implemented a fix on their end to allow external credentials to be
provided.

Fixes #104

### DIFF
--- a/spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/SqlCredentialFactory.java
+++ b/spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/SqlCredentialFactory.java
@@ -1,0 +1,60 @@
+/*
+ *  Copyright 2017 original author or authors.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package org.springframework.cloud.gcp.sql;
+
+import java.io.FileInputStream;
+import java.io.IOException;
+
+import com.google.api.client.auth.oauth2.Credential;
+import com.google.api.client.googleapis.auth.oauth2.GoogleCredential;
+import com.google.cloud.sql.CredentialFactory;
+
+import org.springframework.cloud.gcp.core.autoconfig.GcpContextAutoConfiguration;
+
+/**
+ * Returns the credentials that are written to a system property by the Cloud SQL starter.
+ *
+ * <p>
+ * Since the sockets factory creates an instance of this class by reflection and without any
+ * arguments, the credential location must be in a place that this class knows without any context.
+ *
+ * @author João André Martins
+ */
+public class SqlCredentialFactory implements CredentialFactory {
+
+	public static final String CREDENTIAL_LOCATION_PROPERTY_NAME =
+			"GOOGLE_CLOUD_SQL_CREDS_LOCATION";
+
+	@Override
+	public Credential create() {
+		// TODO(joaomartins): Consider supporting Spring Resources as credential locations. There
+		// would need to be a way to create or inject a Spring context here statically, to load
+		// the resource from there.
+		String credentialResourceLocation = System.getProperty(CREDENTIAL_LOCATION_PROPERTY_NAME);
+		if (credentialResourceLocation == null) {
+			return null;
+		}
+
+		try {
+			return GoogleCredential.fromStream(new FileInputStream(credentialResourceLocation))
+					.createScoped(GcpContextAutoConfiguration.CREDENTIALS_SCOPES_LIST);
+		}
+		catch (IOException e) {
+			return null;
+		}
+	}
+}
--- a/spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/SqlCredentialFactory.java
+++ b/spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/SqlCredentialFactory.java
@@ -22,6 +22,8 @@ import java.io.IOException;
 import com.google.api.client.auth.oauth2.Credential;
 import com.google.api.client.googleapis.auth.oauth2.GoogleCredential;
 import com.google.cloud.sql.CredentialFactory;
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
 
 import org.springframework.cloud.gcp.core.autoconfig.GcpContextAutoConfiguration;
 
@@ -39,6 +41,8 @@ public class SqlCredentialFactory implements CredentialFactory {
 	public static final String CREDENTIAL_LOCATION_PROPERTY_NAME =
 			"GOOGLE_CLOUD_SQL_CREDS_LOCATION";
 
+	private static final Log LOGGER = LogFactory.getLog(SqlCredentialFactory.class);
+
 	@Override
 	public Credential create() {
 		// TODO(joaomartins): Consider supporting Spring Resources as credential locations. There
@@ -46,6 +50,10 @@ public class SqlCredentialFactory implements CredentialFactory {
 		// the resource from there.
 		String credentialResourceLocation = System.getProperty(CREDENTIAL_LOCATION_PROPERTY_NAME);
 		if (credentialResourceLocation == null) {
+			if (LOGGER.isDebugEnabled()) {
+				LOGGER.debug(CREDENTIAL_LOCATION_PROPERTY_NAME + " property does not exist. "
+						+ "Socket factory will use application default credentials.");
+			}
 			return null;
 		}
 
@@ -53,7 +61,8 @@ public class SqlCredentialFactory implements CredentialFactory {
 			return GoogleCredential.fromStream(new FileInputStream(credentialResourceLocation))
 					.createScoped(GcpContextAutoConfiguration.CREDENTIALS_SCOPES_LIST);
 		}
-		catch (IOException e) {
+		catch (IOException ioe) {
+			LOGGER.warn("There was an error loading Cloud SQL credential.", ioe);
 			return null;
 		}
 	}
--- a/spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/autoconfig/GcpCloudSqlAutoConfiguration.java
+++ b/spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/autoconfig/GcpCloudSqlAutoConfiguration.java
@@ -16,6 +16,7 @@
 
 package org.springframework.cloud.gcp.sql.autoconfig;
 
+import java.io.File;
 import java.io.IOException;
 import java.nio.file.Files;
 import java.security.GeneralSecurityException;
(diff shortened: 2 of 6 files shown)


spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/SqlCredentialFactory.java
```diff
@@ -0,0 +1,60 @@
...
+		if (credentialResourceLocation == null) {
+			return null;
+		}
+
+		try {
+			return GoogleCredential.fromStream(new FileInputStream(credentialResourceLocation))
+					.createScoped(GcpContextAutoConfiguration.CREDENTIALS_SCOPES_LIST);
+		}
+		catch (IOException e) {
+			return null;
```
meltsufin: Can we log the exception instead of just swallowing?

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/autoconfig/GcpCloudSqlAutoConfiguration.java
```diff
@@ -127,8 +138,59 @@ public DataSource cloudSqlDataSource(CloudSqlJdbcInfoProvider cloudSqlJdbcInfoPr
...
+		hikariConfig.setInitializationFailFast(this.gcpCloudSqlProperties.isInitFailFast());
 
 		return new HikariDataSource(hikariConfig);
 	}
+
+	/**
+	 * Set credentials to be used by the Google Cloud SQL socket factory.
+	 *
+	 * <p>
+	 * The only way to pass a {@link CredentialFactory} to the socket factory is by passing a class
```
meltsufin: It's a bit strong to say that it's the *only* way because you can also use a static variable or an environment variable, or some other global store.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/autoconfig/GcpCloudSqlAutoConfiguration.java
```diff
@@ -127,8 +138,59 @@ public DataSource cloudSqlDataSource(CloudSqlJdbcInfoProvider cloudSqlJdbcInfoPr
...
+			else {
+				return;
+			}
+
+			// If there are specified credentials, tell sockets factory to use them.
+			System.setProperty(CredentialFactory.CREDENTIAL_FACTORY_PROPERTY,
+					SqlCredentialFactory.class.getName());
+		}
+		catch (IOException ioe) {
+			// Do nothing, let sockets factory use application default credentials.
```
meltsufin: Log the exception anyway as a warning maybe.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/autoconfig/GcpCloudSqlAutoConfiguration.java
```diff
@@ -71,6 +75,9 @@
 	@Autowired
 	private GcpCloudSqlProperties gcpCloudSqlProperties;
 
+	@Autowired
+	private GcpProperties gcpProperties;
```
meltsufin: What if `GcpCloudSqlProperties` contained `gcpProperties` instead?

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/autoconfig/GcpCloudSqlAutoConfiguration.java
```diff
@@ -127,8 +138,59 @@ public DataSource cloudSqlDataSource(CloudSqlJdbcInfoProvider cloudSqlJdbcInfoPr
...
+			}
+			// Then, the global credential.
+			else if (this.gcpProperties != null
+					&& this.gcpProperties.getCredentialsLocation() != null
+					&& this.gcpProperties.getCredentialsLocation().exists()) {
+				// A resource might not be in the filesystem, but the Cloud SQL credential must.
+				if (this.gcpProperties.getCredentialsLocation().getFile() != null) {
+					System.setProperty(SqlCredentialFactory.CREDENTIAL_LOCATION_PROPERTY_NAME,
+							this.gcpProperties.getCredentialsLocation()
+									.getFile().getAbsolutePath());
```
meltsufin: Why is it `getCredentialsLocation().getFile().getAbsolutePath()` here, but just `getCredentialsLocation().toString()` above?

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/autoconfig/GcpCloudSqlAutoConfiguration.java
```diff
@@ -127,8 +138,59 @@ public DataSource cloudSqlDataSource(CloudSqlJdbcInfoProvider cloudSqlJdbcInfoPr
...
+				System.setProperty(SqlCredentialFactory.CREDENTIAL_LOCATION_PROPERTY_NAME,
+						this.gcpCloudSqlProperties.getCredentialsLocation().toString());
+			}
+			// Then, the global credential.
+			else if (this.gcpProperties != null
+					&& this.gcpProperties.getCredentialsLocation() != null
+					&& this.gcpProperties.getCredentialsLocation().exists()) {
+				// A resource might not be in the filesystem, but the Cloud SQL credential must.
+				if (this.gcpProperties.getCredentialsLocation().getFile() != null) {
+					System.setProperty(SqlCredentialFactory.CREDENTIAL_LOCATION_PROPERTY_NAME,
```
meltsufin: It's a matter of style, but I would establish the location of the credentials file first in a local variable, and then set it outside of this if-else, and would also avoid a `return` statement in the middle of the method.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/autoconfig/GcpCloudSqlAutoConfiguration.java
joaoandremartins: They're different things. Here, `GcpProperties` refers to the core configuration, `GcpCloudSqlProperties` refers to the SQL configuration. It should be possible for both instances to contain different project IDs and credentials.

If you mean `GcpCloudSqlProperties` should contain a `GcpProperties` member instead of separate `projectId` and `credentialsLocation` members, then, as per http://docs.spring.io/spring-boot/docs/current-SNAPSHOT/reference/htmlsingle/#configuration-metadata-nested-properties, we'd end up with something like:

`spring.cloud.gcp.sql.gcp-properties:project-id` (similarly for `credentials-location`)

which I'm not sure I like..

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/autoconfig/GcpCloudSqlAutoConfiguration.java
joaoandremartins: This is separate from our discussion the other day. Here I'm talking about passing a credentials factory to the socket factory, it really is the only way in which the factory receives credentials factory.

I think you're referring to passing the resource URI to `SqlCredentialsFactory` via env var and then loading the resource in `SqlCredentialsFactory` using the Spring context. That probably warrants a change below where I say "... the class can read it without any context".

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/autoconfig/GcpCloudSqlAutoConfiguration.java
joaoandremartins: To avoid the return in the middle of the method, I would need to add a condition before setting `CREDENTIAL_FACTORY_PROPERTY`. Pick your poison :)

I changed the logic a bit to extract the location to its own variable (and noticed a return lacking). I don't think we can simplify the cascading conditions (e.g., by having a huge  "no creds file exists" return condition at the beginning) because of the credentials fallback. LMKWYT

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/autoconfig/GcpCloudSqlAutoConfiguration.java
joaoandremartins: `this.gcpCloudSqlProperties.getCredentialsLocation()` is a `Path`, `this.gcpProperties.getCredentialsLocation().getFile()` is a `File`.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/autoconfig/GcpCloudSqlAutoConfiguration.java
joaoandremartins: Logging as debug not to pollute logs.

joaoandremartins:
Hey guys, can you see if anything is missing? I'd like to merge this asap since this might be causing some issues/confusion with some users.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/SqlCredentialFactory.java
```diff
@@ -0,0 +1,69 @@
...
+			"GOOGLE_CLOUD_SQL_CREDS_LOCATION";
+
+	private static final Log LOGGER = LogFactory.getLog(SqlCredentialFactory.class);
+
+	@Override
+	public Credential create() {
+		// TODO(joaomartins): Consider supporting Spring Resources as credential locations. There
+		// would need to be a way to create or inject a Spring context here statically, to load
+		// the resource from there.
+		String credentialResourceLocation = System.getProperty(CREDENTIAL_LOCATION_PROPERTY_NAME);
```
artembilan: If you worry about Spring Context, you can just set this system property from some bean which would create a temporary file based on the `Resource`.
Otherwise that is always bad idea to store Spring context in the static scope. You shouldn't forget that your JVM may manage several applications, which indeed may use the same classes.

joaoandremartins: Yes, I really don't like the static way, it sounds way too hacky.

Creating a file copy could indeed make it work with Spring Resources. I'm wondering if the added complexity justifies the added flexibility? Is specifying credentials files in places other than files important? A use case is getting creds stored in GCS, though we should take into account we're copying sensitive private keys around which shouldn't be done lightly and is a possible security breach..

joaoandremartins:
Great question on how it's going to work in the Cloud. The plan is that in a managed compute platform, at least in a Google one (GAE, GCE), the user doesn't specify anything and the default creds are used, which come from the App Identity API in GAE or the GCE metadata server.

However, none of this is documented and I think it warrants more thought. I would perhaps leave reinforcing this logic for another issue/PR, since right now it's critical that we provide our credentials to the Cloud SQL sockets factory rather than letting it use application default ones.

PR #114, opened 2017-08-18
Removes curly braces in property examples to hopefully reduce confusion

joaoandremartins:


meltsufin:
We've had someone confused by this and put their values in quotes like this:
```
spring.cloud.gcp.project-id="my-project"
```

Maybe we should do validation instead?

joaoandremartins:
I think validation is too heavy here. Maybe just something to look for on user reports...

PR #121, opened 2017-08-21
Scopes SQL credentials to Cloud SQL only

### LINKED ISSUE #119: Scope SQL starter credentials to SQL only
Right now, they're being scoped to all scopes supported by our starters (SQL, Pub/Sub, Storage, etc.).

joaoandremartins:
The SslSocketFactory uses the SQL Admin API and requires this scope. The
rest of the scope should be narrowed down only to what is required.

Fixes #119

### DIFF
--- a/spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/src/main/java/org/springframework/cloud/gcp/core/autoconfig/GcpContextAutoConfiguration.java
+++ b/spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/src/main/java/org/springframework/cloud/gcp/core/autoconfig/GcpContextAutoConfiguration.java
@@ -55,15 +55,15 @@ import org.springframework.util.StringUtils;
 @EnableConfigurationProperties(GcpProperties.class)
 public class GcpContextAutoConfiguration {
 
-	private static final String PUBSUB_SCOPE = "https://www.googleapis.com/auth/pubsub";
+	public static final String PUBSUB_SCOPE = "https://www.googleapis.com/auth/pubsub";
 
-	private static final String SQLADMIN_SCOPE =
+	public static final String SQLADMIN_SCOPE =
 			"https://www.googleapis.com/auth/sqlservice.admin";
 
-	private static final String STORAGE_READ_SCOPE =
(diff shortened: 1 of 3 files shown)


spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/src/main/java/org/springframework/cloud/gcp/core/autoconfig/GcpContextAutoConfiguration.java
```diff
@@ -55,15 +55,15 @@
 @EnableConfigurationProperties(GcpProperties.class)
 public class GcpContextAutoConfiguration {
 
-	private static final String PUBSUB_SCOPE = "https://www.googleapis.com/auth/pubsub";
+	public static final String PUBSUB_SCOPE = "https://www.googleapis.com/auth/pubsub";
```
meltsufin: Perhaps create a GcpScopes enum?

joaoandremartins: Good idea. Added

joaoandremartins:
@artembilan can you provide your review here? Thanks!

joaoandremartins:
@artembilan just checking if you plan on reviewing this one and #122. Thanks!

artembilan:
Well, I will, but somewhere later the next week or so.
I just have some other commitments to do, like releasing my other projects together with upcoming SF 5.0 RC4. 

I'm so sorry to tell you that, but I'm just out of time pool to track everything you do here. Otherwise I would abandon all those projects what is out of my responsibility already.

meltsufin:
@artembilan It's totally understandable that you have other commitments and reviewing everything here is not always the top priority. We appreciate the reviews that you're doing already very much.
If you don't mind, when you know you won't be able to get to a PR for a while, just let us know here and we can think of alternatives or waiting for you to free up.
Thanks!!

@joaoandremartins in this particular PR, I don't think there's anything controversial. So, I think it's OK to merge.

PR #122, opened 2017-08-21
Allows default credential scopes to be parameterised

### LINKED ISSUE #118: Allow specifying OAuth2 scopes via a property
Right now, we use a static list of scopes for the core credentials. This should be configurable.

joaoandremartins:
Scopes list was static so far. Useful to limit credentials power in
case not every service is used.

Fixes #118

### DIFF
--- a/spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/SqlCredentialFactory.java
+++ b/spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/SqlCredentialFactory.java
@@ -59,7 +59,7 @@ public class SqlCredentialFactory implements CredentialFactory {
 
 		try {
 			return GoogleCredential.fromStream(new FileInputStream(credentialResourceLocation))
-					.createScoped(GcpContextAutoConfiguration.CREDENTIALS_SCOPES_LIST);
+					.createScoped(GcpContextAutoConfiguration.DEFAULT_CREDENTIAL_SCOPES);
 		}
 		catch (IOException ioe) {
 			LOGGER.warn("There was an error loading Cloud SQL credential.", ioe);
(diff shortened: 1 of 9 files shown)


joaoandremartins:
What do you guys think of introducing the nested `Credentials` class? I'm thinking of following a similar approach with `GcpProperties` by renaming it to `Core` and then nesting it in other starters configurations, then referring to it like

`spring.cloud.gcp.sql.core.project-id`
`spring.cloud.gcp.sql.core.credentials.location`

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/SqlCredentialFactory.java
```diff
@@ -59,7 +59,7 @@ public Credential create() {
 
 		try {
 			return GoogleCredential.fromStream(new FileInputStream(credentialResourceLocation))
-					.createScoped(GcpContextAutoConfiguration.CREDENTIALS_SCOPES_LIST);
+					.createScoped(GcpContextAutoConfiguration.DEFAULT_CREDENTIAL_SCOPES);
```
meltsufin: Why not in this instance use only the scopes needed for Cloud SQL?

joaoandremartins: #121 takes care of that, this is just from the refactor.

joaoandremartins:
Can you think of an alternative name to `core`? That's what GcpProperties will have to be renamed to, if we want to reuse it.

meltsufin:
Unfortunately, I can't think of a better name. :)
That's why I'm not sure about the whole idea of reusing it this way.

joaoandremartins:
@artembilan can you provide your review here? Thanks!

PR #123, opened 2017-08-22
Adds Cloud SQL starter support for PostgreSQL

### LINKED ISSUE #116: Google Cloud SQL w/ PostgreSQL
The `spring-cloud-gcp-starter-sql` is easily configured for MySQL databases.

I was able to get it working against PostgreSQL too:

pom.xml:
```
        <dependency>
            <artifactId>spring-cloud-gcp-starter-sql</artifactId>
            <version>1.0.0.BUILD-SNAPSHOT</version>
        </dependency>
        <dependency>
            <groupId>org.postgresql</groupId>
            <artifactId>postgresql</artifactId>
            <version>42.1.4</version>
        </dependency>
       <dependency>
            <artifactId>postgres-socket-factory</artifactId>
            <version>1.0.4</version>
        </dependency>
```
Then, application.properties:
```
spring.cloud.gcp.projectId=[[PROJECT-ID]]
spring.cloud.gcp.sql.instanceName=[[INSTANCE-ID]]
spring.cloud.gcp.sql.jdbcDriver=org.postgresql.Driver
spring.cloud.gcp.sql.jdbcUrl=jdbc:postgresql://google/[[DATABASE-NAME]]?socketFactory=com.google.cloud.sql.postgres.SocketFactory&socketFactoryArg=[[CONNECTION-NAME]]
spring.cloud.gcp.sql.databaseName=[[DATABASE-NAME]]
spring.cloud.gcp.sql.userName=[[DATABASE-USER]]
spring.cloud.gcp.sql.password=[[DATABASE-PASSWORD]]
```

//cc:@joaoandremartins

joaoandremartins:
Only MySQL is supported currently, and Cloud SQL also supports
PostgreSQL.

Fixes #116


joaoandremartins:
FYI @saturnism @DazWilkin

DazWilkin:
Nice!!

joaoandremartins:
@artembilan we'll also need your review here whenever you can. :)

artembilan:
Sorry, I don't know what's going on and it is fully unclear for me why do we need the MySQL driver at all.
Therefore I can't imagine from which side to review this for Postgre as well.

Why don't we provide provide the starter for Oracle on the other hand?
How does this Cloud magic work at all?

I thought that there is some URL in the cloud and we just provide a `DataSource` proxy for that. Nothing more from the target SQL engine impl. 

Sorry for my ignorance, but I'm really useless for review in the subject.

Maybe there is something to read in the evening?
Thanks

joaoandremartins:
I can try to explain the motivation for this starter in a few words.

We're connecting to Cloud SQL through a [TLS tunnel](https://cloud.google.com/sql/docs/mysql/connect-external-app#java) (so that we don't need to open the instance to any external IPs). We also do other conveniences for users, for example, we auto-discover an instance region if the user didn't specify it. Essentially, you can connect to Google Cloud SQL with a couple of properties and our starter.

Connecting to Cloud SQL from App Engine is also special in the sense that we need to use our driver, not the generic MySQL one.

PostgreSQL follows a similar philosophy. Since the URL is slightly different for PostgreSQL, we need this work.

Let me know if you feel at ease reviewing or if you'd rather pass.

artembilan:
Thanks, @joaoandremartins , for explanation.
Well, with the fact that it must work as it is explained of course I can review. 
I just don't understand why should we be tied to the particular driver at all. 
Or does that mean that the real SQL server in the cloud is MySQL and the work in this PR is done for the PostgreSQL server in the cloud?

joaoandremartins:
Google Cloud SQL supports both [MySQL](https://cloud.google.com/sql/docs/mysql/) and [PostgreSQL](https://cloud.google.com/sql/docs/postgres/) instances. So, we're releasing Spring support for both of those.
I hope that clarifies it! Let me know if you have any other questions.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/README.adoc
```diff
@@ -59,25 +56,63 @@ If `spring.cloud.gcp.sql.jdbc-url` is specified, it is used as the final JDBC UR
...
 https://brettwooldridge.github.io/HikariCP/[HikariCP]. If you want to use another kind of
 `DataSource`, feel free to use the provided `CloudSqlJdbcInfoProvider` to provide your own
 `DataSource` bean.
+
+== Troubleshooting tips
+
+If you're not able to connect to a database and see an endless loop of
+`Connecting to Cloud SQL instance [...] on IP [...]`, it's likely that exceptions are being thrown
+and logged at a level lower than your logger's level. This may be the case with HikariCP, the
+default DataSource use by this starter, if your logger is set to INFO or higher level.
```
artembilan: Typo: use**d**

meltsufin:
I think that one of the options we've discussed offline is to have a separate starters for Cloud SQL PostgreSQL vs MySQL. Although both services are under the Cloud SQL offering, they're two different products with different drivers. I don't remember what were the major cons of such bifurcation.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/DatabaseType.java
```diff
@@ -0,0 +1,52 @@
...
+			+ "socketFactory=com.google.cloud.sql.mysql.SocketFactory&useSSL=false"),
+	POSTGRESQL("postgresql", "org.postgresql.Driver", "jdbc:postgresql://google/%s?"
+			+ "socketFactory=com.google.cloud.sql.postgres.SocketFactory&socketFactoryArg=%s"
+			+ "&useSSL=false");
+
+	private String name;
+
+	private String jdbcDriverName;
+
+	private String jdbcUrlTemplate;
```
artembilan: All properties must be `final` and there should be a blank line between enum values.
I think we should configure some Checkstyle rule on the matter. Don't know which one though...

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/GcpCloudSqlProperties.java
```diff
@@ -48,6 +48,8 @@
 
 	private Path credentialsLocation;
 
+	private String databaseType = "mysql";
```
artembilan: I'm pretty sure this can be of `DatabaseType` and, therefore, we don't need a custom `Validator` above.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/autoconfig/CloudSqlJdbcUrlResolver.java
```diff
@@ -0,0 +1,80 @@
...
+		Assert.hasText(this.properties.getDatabaseName(), "A database name must be provided.");
+
+		String instanceConnectionName = this.properties.getInstanceConnectionName();
+		if (StringUtils.isEmpty(instanceConnectionName)) {
+			Assert.hasText(this.properties.getInstanceName(),
+					"An instance name is required, or specify an instance connection name explicitly");
+			String region = this.properties.getRegion();
+			if (StringUtils.isEmpty(region)) {
+				try {
+					region = determineRegion(this.projectId, this.properties.getInstanceName());
```
artembilan: I see the main reason of this class to determine a `region` against provided `projectId` and `instanceName`.
Why can't we be more strict and really require the `region` from end-user all the time? If `instanceConnectionName` isn't provided, of course.

joaoandremartins: It's a feature to get people started even more easily. People tend to memorize the instance name better than its region (at least, I do). It's the difference between having to go to the Cloud Console to get an instance's region or skipping that step completely.

Although we should point out in the docs that this is not recommended for production environments.

artembilan: Yeah... I would really prefer to ask end-user to provide it rather than have such a "creepy" hooks during application start up... But let it be if you're OK to support this code in the future. Only the problem I'd like to see is `static` bean for the `SqlAdmin` do not load any other beans for nothing during this initialization phase.

joaoandremartins: Apologies, I didn't understand what you want with the `SQLAdmin` bean.

artembilan: OK. Never mind. It doesn't help because that `DefaultCloudSqlJdbcInfoProvider` fully depends on the `SQLAdmin` bean. Although again: the last one is used only to determine a string for the `region`. Where it would be much easier just to request it from the user and don't care about `SQLAdmin` overhead at all.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/autoconfig/GcpCloudSqlAutoConfiguration.java
```diff
@@ -113,16 +116,40 @@ public CloudSqlJdbcInfoProvider appengineCloudSqlJdbcInfoProvider(
...
-	public CloudSqlJdbcInfoProvider cloudSqlJdbcInfoProvider(SQLAdmin sqlAdmin,
+	@ConditionalOnProperty(value = "spring.cloud.gcp.sql.database-type", havingValue = "postgresql")
+	public CloudSqlJdbcInfoProvider postgreSqlJdbcInfoProvider(SQLAdmin sqlAdmin,
 			GcpProjectIdProvider projectIdProvider) {
 		CloudSqlJdbcInfoProvider defaultProvider = new DefaultCloudSqlJdbcInfoProvider(
 				projectIdProvider.getProjectId(),
 				sqlAdmin,
-				this.gcpCloudSqlProperties);
+				this.gcpCloudSqlProperties,
+				DatabaseType.POSTGRESQL);
```
artembilan: Something smells here. We have now `public DatabaseType getDatabaseType() `, so `DefaultCloudSqlJdbcInfoProvider` can easily determine what to do or even it won't care because `DatabaseType` does the magic already.
IMO we need only one `DefaultCloudSqlJdbcInfoProvider` bean and rely on the provided properties, without any `@ConditionalOnProperty`

joaoandremartins: Yes, good catch!

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/GcpCloudSqlProperties.java
```diff
@@ -38,7 +38,7 @@
 
 	private String jdbcUrl;
 
-	private String jdbcDriver = "com.mysql.jdbc.Driver";
+	private String jdbcDriver;
```
artembilan: Do we need this property if we really stick only with what we have in the `DatabaseType` enum?

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/GcpCloudSqlProperties.java
```diff
@@ -127,4 +129,12 @@ public Path getCredentialsLocation() {
 	public void setCredentialsLocation(Path credentialsLocation) {
 		this.credentialsLocation = credentialsLocation;
 	}
+
+	public DatabaseType getDatabaseType() {
```
artembilan: `@NotNull` ?

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/GcpCloudSqlProperties.java
joaoandremartins: Yes, because users might want a driver class that's not in any DatabaseType.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/GcpCloudSqlProperties.java
joaoandremartins: I think you mean in the setter. Agreed. We should probably also add assertions to all the setters to prevent setting to `null`.

artembilan: Well, if `jdbcDriver` is an alternative, then this can be `null`.
So, never mind.
Thanks

Although I'm not sure which other driver end-user can choose, if we have only MySQL and PostgreSQL choice...

joaoandremartins: One example I like to think of is MySQL 8: the driver is now `com.mysql.cj.jdbc.Driver`. If users add that dependency on their side and set the driver class name to that, it should work with that MySQL version too.
For App Engine Standard, we also found that we had to select a different driver than the default MySQL one (`com.mysql.jdbc.GoogleDriver`).
There might be other edge cases in other environments there that we don't know about. So, it's nice to give users the flexibility to change this if they need to.

`DatabaseType` shouldn't be set to `null`. If the user provides the JDBC URL and driver class name,  we don't need to check its value, but in any other case we do. `databaseType` already has a default value, there's no good reason for a user to be setting `null` I can think of.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/autoconfig/CloudSqlJdbcUrlResolver.java
joaoandremartins: Judging from some talks, removing region auto-discovery is where we're headed.

artembilan: OK. And let me guess: this issue is exactly about that: https://github.com/spring-cloud/spring-cloud-gcp/issues/142.

joaoandremartins: You got it :)
I think I addressed all your comments, but let me know if I missed anything.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/GcpCloudSqlProperties.java
joaoandremartins: Initially I was thinking of adding null checks to every setter, but since we have plenty, I started wondering whether it was worth it and then decided not to.

joaoandremartins:
Had to fix merge changes, need a new approval.

meltsufin:
I think we do merge because (1) it's readily available from the Github UI, and (2) it doesn't rewrite the commit history on the branch. Of course we rewrite the commit history when we squash and merge into master, but that's more acceptable because we want a clean history there. On PR branches, we don't care as much about what the commit history looks like because they're temporary. That being said, there is no serious issue with doing it on a PR branch either. In the end, when we merge to master, it will look exactly the same.

artembilan:
I'm not sure what you mean about "Github UI" for the working branch, not the `master`. Well, I really do care about history of commits in branch as well. And here that is more logical that current `master` state is fully under all the commits in this branch, not somewhere in between. Your changes are essentially on top of `master`, just like they will be after merging to the `master`. So, if we care about the clean merging, that would be better to `rebase` branches to the master.

On the other hand with the rebase we don't have extra noise in the email box about `merging master to...` and sometime we even don't have any notifications if your branch rebases to master clearly.

Just IMO, of course... I can live with your process but really would like to know what is the point of extra `merge` commits in the working branch.

Thanks

markfisher:
Given that the PR branches *are* temporary, it doesn't seem like rewriting history should be a concern there IF the rebased version is easier to review and more closely resembles the final result that will be merged to master.

meltsufin:
The diff that Github will show on the PR is the same either way. We don't really have a policy for rebase vs. merge against master in PRs. So, either is fine.

joaoandremartins:
@artembilan the issues I found with rebase were that 1) it rewrites commit history **and, therefore,** the GitHub comments on previous versions disappear and 2) I always have to force push whenever master changes. I used to be a rebase fanboy but got convinced that merge is slightly more friendly, due to this couple of issues.

In any case, I'd always have to change things on this PR since it contained region discovery that was removed in another PR and I had relocated in this PR, so merge didn't get rid of it.

joaoandremartins:
Apologies, but I'll need another approval since there was another merge conflict with another PR merged in the meantime. :(

artembilan:
What am I missing? It shows that conflicts are still there.
Maybe you forgot to push the fix?

meltsufin:
I don't see any conflicts. I think you can merge this @joaoandremartins.

PR #125, opened 2017-08-23
Adds spring-jdbc dependency so that user doesn't have to

joaoandremartins:
Currently, users need to manually add this dependency.

Unclear what the value of adding spring-boot-starter-jdbc instead is, but let me know if I'm missing anything.

@jabubake fyi


meltsufin:
@joaoandremartins Let's roll this back because we want to be a drop-in replacement for "MySQL".

PR #126, opened 2017-08-24
Calls setApplicationName on SQLAdmin builder and adds useSSL to MySQL

### LINKED ISSUE #124: Application name is not set. Call Builder#setApplicationName.
I see this warning in the log when using Cloud SQL starter.
Filed issue: https://github.com/GoogleCloudPlatform/google-cloud-java/issues/2371

joaoandremartins:
...connection string

We might get usage metrics from SQLAdmin from setApplicationName and
the warning logs will now be silenced.

Fixes #124

### DIFF
--- a/spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/autoconfig/DefaultCloudSqlJdbcInfoProvider.java
+++ b/spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/autoconfig/DefaultCloudSqlJdbcInfoProvider.java
@@ -70,7 +70,7 @@ public class DefaultCloudSqlJdbcInfoProvider implements CloudSqlJdbcInfoProvider
 			}
 
 			String jdbcUrl = String.format("jdbc:mysql://google/%s?cloudSqlInstance=%s&"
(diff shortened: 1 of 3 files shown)


spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/autoconfig/DefaultCloudSqlJdbcInfoProvider.java
```diff
@@ -70,7 +70,7 @@ public DefaultCloudSqlJdbcInfoProvider(String projectId, SQLAdmin sqlAdmin, GcpC
 			}
 
 			String jdbcUrl = String.format("jdbc:mysql://google/%s?cloudSqlInstance=%s&"
-					+ "socketFactory=com.google.cloud.sql.mysql.SocketFactory",
+					+ "socketFactory=com.google.cloud.sql.mysql.SocketFactory&useSSL=true",
```
meltsufin: I think this should be a configuration property (with a default of true?).
I don't see it's documented in our public docs. Let's look into this more.

joaoandremartins:
We got a response in https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/issues/50, PTAL

joaoandremartins:
Merging this without any other approvals since it's a trivial change.

Issue #92, opened 2017-08-10
Enhance Spring Cloud Pub/Sub Template/Operation


saturnism:
Currently PubSubOperation has only 1 method, send(topic, message). It can be further enhanced with better Pub/Sub support
- [x] JavaDoc on key public interfaces, like PubSubOperation
- [ ] Additional Surface
  - [x] A way to send just a string, e.g., send(topic, String message)
  - [x] A way to do a sync send
  - [ ] A way to send an object and marshall w/ converter, e.g., send(topic, Object message) & use converter, potentially detect Jackson in classpath and automatically register a JSON converter.
  - [ ] Revisit MessageHeaders, make sure headers captures and propagates Pub/Sub metadata
  - [x] Revisit configuration, make sure all settings are configurable via a bean, e.g., RetrySetting
  - [x] A way to pull a message, e.g., pull(subscription)
  - [x] A way to subscribe to a topic, e.g., subscribe(messageListener)
  - [x] Expose underlying API, e.g., getSubscriberClient(), getPublisherClient()
  - [x] Clean up converter, currently converter is not symmetric (e.g., toMessage() & fromMessage() returns very different types
  - [ ] Add SubscriberConfiguration and PublsiherConfiguration
- [x] Sanity check - Spring Integration Pub/Sub should probably use PubSubOperation for everything

jabubake:
+1 on a way to send an object w encoder /decoder support

joaoandremartins:
We should probably also rename `send()` to `publish()` as part of this issue.

joaoandremartins:
@saturnism can you please clarify the requirements around `Add SubscriberConfiguration and PublsiherConfiguration` ?

meltsufin:
I created separated issues for all unfinished tasks here.

Issue #54, opened 2017-06-07
Use CompositeMessageConverter in PubSubTemplate


joaoandremartins:
There have been suggestions to chain MessageConverters in PubSubTemplate, e.g., through the use of a CompositeMessageConverter.

Need to look more deeply into this, including how it's done in other parts of Spring.

PR #127, opened 2017-08-25
Populates Spring Cloud GCP reference document

### LINKED ISSUE #105: Publish reference document
We should write up and publish a reference document and link it from our [project's page](http://cloud.spring.io/spring-cloud-gcp/).

@sobychacko can you help with the process to publish the document to the gh-pages branch?

@garyrussell can you help with why the page is linking to a strange link in `docs.github.io` and the URL contains `spring-new`? http://docs.spring.io/spring-new/docs/1.0.0.BUILD-SNAPSHOT/spring-cloud-gcp/htmlsingle/

joaoandremartins:
Fixes #105

### DIFF
--- a/spring-cloud-gcp-docs/src/main/asciidoc/core.adoc
+++ b/spring-cloud-gcp-docs/src/main/asciidoc/core.adoc
@@ -17,10 +17,10 @@ Maven coordinates, using Spring Cloud GCP BOM:
 
 Gradle coordinates:
 
-[source]
+[source,subs="normal"]
 ----
 dependencies {
-    compile group: 'org.springframework.cloud', name: 'spring-cloud-gcp-starter-core', version: '1.0.0.BUILD-SNAPSHOT'
+    compile group: 'org.springframework.cloud', name: 'spring-cloud-gcp-starter-core', version: '{project-version}'
 }
 ----
 
@@ -35,9 +35,9 @@ public interface GcpProjectIdProvider {
 }
 ----
 
-The Spring Boot GCP Core starter auto-configures a `GcpProjectIdProvider`. If a
-`spring.cloud.gcp.project-id` property is specified, the provided `GcpProjectIdProvider` returns
-that property value.
+The Spring Boot GCP Core starter auto-configures a `GcpProjectIdProvider`.
+If a `spring.cloud.gcp.project-id` property is specified, the provided `GcpProjectIdProvider`
+returns that property value.
 
 [source,java]
 ----
@@ -45,7 +45,7 @@ spring.cloud.gcp.project-id=my-gcp-project-id
 ----
 
 Otherwise, the project ID is discovered based on a
-https://googlecloudplatform.github.io/google-cloud-java/0.20.1/apidocs/com/google/cloud/ServiceOptions.html#getDefaultProjectId--[set of rules]:
+https://googlecloudplatform.github.io/google-cloud-java/latest/apidocs/com/google/cloud/ServiceOptions.html#getDefaultProjectId-[set of rules]:
 
 1. The project ID specified by the `GOOGLE_CLOUD_PROJECT` environment variable
 2. The Google App Engine project ID
@@ -66,12 +66,14 @@ public interface CredentialsProvider {
 }
 ----
 
-The Spring Boot GCP Core starter auto-configures a `CredentialsProvider`. It uses the
-`spring.cloud.gcp.credentials.location` property to locate the OAuth2 private key of a Google
-service account. Keep in mind this property is a Spring Resource, so the credentials file can be obtained from a number of
+The Spring Boot GCP Core starter auto-configures a `CredentialsProvider`.
+It uses the `spring.cloud.gcp.credentials.location` property to locate the OAuth2 private key of a
+Google service account.
+Keep in mind this property is a Spring Resource, so the credentials file can be obtained from a
+number of
 https://docs.spring.io/spring/docs/current/spring-framework-reference/html/resources.html#resources-implementations[different locations]
-such as the file system, classpath, URL, etc. The next example specifies the credentials location
-property in the file system.
+such as the file system, classpath, URL, etc.
+The next example specifies the credentials location property in the file system.
 
 [source,yaml]
 ----
@@ -101,8 +103,9 @@ service supported by Spring Cloud GCP.
 |===
 
 The Spring Boot GCP Core starter allows you to configure a custom scope list for the provided
-credentials. To do that, specify a comma-delimited list of scopes in the
-`spring.cloud.gcp.credentials.scopes` property.
+credentials.
+To do that, specify a comma-delimited list of scopes in the `spring.cloud.gcp.credentials.scopes`
+property.
 
 [source,yaml]
 ----
--- a/spring-cloud-gcp-docs/src/main/asciidoc/core.adoc
+++ b/spring-cloud-gcp-docs/src/main/asciidoc/core.adoc
@@ -3,6 +3,27 @@
 At the center of every Spring Cloud GCP module are the concepts of `GcpProjectIdProvider` and
 `CredentialsProvider`.
 
+Spring Cloud GCP provides a Spring Boot starter to auto-configure the core components.
+
+Maven coordinates, using Spring Cloud GCP BOM:
+
+[source,xml]
+----
+<dependency>
+    <groupId>org.springframework.cloud</groupId>
+    <artifactId>spring-cloud-gcp-starter-core</artifactId>
+</dependency>
+----
+
+Gradle coordinates:
+
+[source]
+----
+dependencies {
+    compile group: 'org.springframework.cloud', name: 'spring-cloud-gcp-starter-core', version: '1.0.0.BUILD-SNAPSHOT'
+}
+----
+
 === Project ID
 
 `GcpProjectIdProvider` is a functional interface that returns a GCP project ID string.
@@ -27,11 +48,11 @@ Otherwise, the project ID is discovered based on a
 https://googlecloudplatform.github.io/google-cloud-java/0.20.1/apidocs/com/google/cloud/ServiceOptions.html#getDefaultProjectId--[set of rules]:
 
 1. The project ID specified by the `GOOGLE_CLOUD_PROJECT` environment variable
-1. The Google App Engine project ID
-1. The project ID specified in the JSON credentials file pointed by the
+2. The Google App Engine project ID
+3. The project ID specified in the JSON credentials file pointed by the
 `GOOGLE_APPLICATION_CREDENTIALS` environment variable
-1. The Google Cloud SDK project ID
-1. The Google Compute Engine project ID, from the Google Compute Engine Metadata Server
+4. The Google Cloud SDK project ID
+5. The Google Compute Engine project ID, from the Google Compute Engine Metadata Server
 
 === Credentials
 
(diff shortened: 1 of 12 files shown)


joaoandremartins:
This is actually WIP -- I'm just seeking feedback on the format so far to know if I'm going on the right track or not.

@saturnism @jabubake PTAL

joaoandremartins:
FYI this is staged at https://storage.googleapis.com/spring-docs/htmlsingle/index.html

Please review there and comment here.

@parkjam4 it would be great if you could also take a look here. Thanks a lot!

sobychacko:
One general suggestion for the reference document over all. Might want to consider using the new ref guide themes and style. 
See here: https://docs.spring.io/spring-cloud-stream/docs/current/reference/htmlsingle/
or http://cloud.spring.io/spring-cloud-static/Edgware.M1/

possibly look at this issue for the changes added in `Spring Cloud Stream`: https://github.com/spring-cloud/spring-cloud-stream-starters/issues/11

joaoandremartins:
Thanks for the tip @sobychacko . I've applied some changes and I can see that the resulting html in `target/generated-docs/index.html` follows an identical style to the other reference guides you posted. Will the reference publishing guide pull the guide from there? I noticed no `index.html` was being generated at that location previously.

@meltsufin this is a revised version of the document, with changes from comments on the doc shared by @saturnism . I'll need another approval. If any issues are remaining, we can address them with other PRs.

spring-cloud-gcp-docs/src/main/asciidoc/core.adoc
```diff
@@ -0,0 +1,110 @@
...
+`spring.cloud.gcp.project-id` property is specified, the provided `GcpProjectIdProvider` returns
+that property value.
+
+[source,java]
+----
+spring.cloud.gcp.project-id=my-gcp-project-id
+----
+
+Otherwise, the project ID is discovered based on a
+https://googlecloudplatform.github.io/google-cloud-java/0.20.1/apidocs/com/google/cloud/ServiceOptions.html#getDefaultProjectId--[set of rules]:
```
artembilan: Isn't there some "versionless" URL to use?
I'm do not worry about this from upgrade to upgrade

spring-cloud-gcp-docs/src/main/asciidoc/core.adoc
```diff
@@ -0,0 +1,110 @@
...
+| Storage (Write/Write) | https://www.googleapis.com/auth/devstorage.read_write
+|===
+
+The Spring Boot GCP Core starter allows you to configure a custom scope list for the provided
+credentials. To do that, specify a comma-delimited list of scopes in the
+`spring.cloud.gcp.credentials.scopes` property.
+
+[source,yaml]
+----
+spring.cloud.gcp.credentials.scopes=https://www.googleapis.com/auth/pubsub,https://www.googleapis.com/auth/sqlservice.admin
```
artembilan: Can't we have some `enum` on the matter to be more friendly for users?

spring-cloud-gcp-docs/src/main/asciidoc/dependency-management.adoc
```diff
@@ -0,0 +1,26 @@
...
+versions of the used dependencies.
+
+[source,xml]
+----
+<dependencyManagement>
+    <dependencies>
+        <dependency>
+            <groupId>org.springframework.cloud</groupId>
+            <artifactId>spring-cloud-gcp-dependencies</artifactId>
+            <version>1.0.0.BUILD-SNAPSHOT</version>
```
artembilan: The same version placeholder concern

spring-cloud-gcp-docs/src/main/asciidoc/index.adoc
```diff
@@ -1,8 +1,26 @@
...
 :toc:
+:toclevels: 4
+:source-highlighter: prettify
+:numbered:
 
-include::overview.adoc[]
+== Introduction
+
+The Spring Cloud GCP project aims at making Google Cloud Platform (GCP) the platform of choice to
+run the popular Spring framework.
```
artembilan: Spring **F**ramework is a brand, not just Spring

spring-cloud-gcp-docs/src/main/asciidoc/pubsub.adoc
```diff
@@ -0,0 +1,271 @@
...
+----
+public ListenableFuture<String> send(final String topic, Message message)
+----
+
+The `send()` method takes a Spring Message and it uses the built-in `SimpleMessageConverter` to
+convert from `Message` to the `PubsubMessage` accepted by the Google Cloud Java Library for Pub/Sub.
+`SimpleMessageConverter` copies the `Message` headers into `PubsubMessage` attributes and the
+`Message` payload into the `PubsubMessage` data. `SimpleMessageConverter` calls `toString()` in both
+the `Message` headers and payload, since `PubsubMessage` only supports `String` attributes and data
+at the transmission level. We're considering JSON object encoding/decoding in the future to support
```
artembilan: I wouldn't say anything about future plans.
First: plans might be changes.
Second: we may forget to remove this sentence when the fix is there already.

OTOH do we have an appropriate GH issue if plans are announced?
😄

spring-cloud-gcp-docs/src/main/asciidoc/spring-integration.adoc
```diff
@@ -1,7 +1,111 @@
...
+[source,java]
+----
+@Bean
+@ServiceActivator(inputChannel = "pubsubInputChannel")
+public MessageHandler messageReceiver() {
+    return message -> {
+        LOGGER.info("Message arrived! Payload: "
+              + ((ByteString) message.getPayload()).toStringUtf8());
+        AckReplyConsumer consumer =
+              (AckReplyConsumer) message.getHeaders().get(GcpHeaders.ACKNOWLEDGEMENT);
```
artembilan: FYI, `MessageHeaders` has this method to consider here:
```
public <T> T get(Object key, Class<T> type)
```

spring-cloud-gcp-docs/src/main/asciidoc/core.adoc
```diff
@@ -0,0 +1,110 @@
...
+    <artifactId>spring-cloud-gcp-starter-core</artifactId>
+</dependency>
+----
+
+Gradle coordinates:
+
+[source]
+----
+dependencies {
+    compile group: 'org.springframework.cloud', name: 'spring-cloud-gcp-starter-core', version: '1.0.0.BUILD-SNAPSHOT'
```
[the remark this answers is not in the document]
joaoandremartins: Using {project-version} now.

Maven samples don't use version because they assume our BOM is being used.

spring-cloud-gcp-docs/src/main/asciidoc/core.adoc
```diff
@@ -0,0 +1,110 @@
...
+
+[source,java]
+----
+public interface GcpProjectIdProvider {
+	String getProjectId();
+}
+----
+
+The Spring Boot GCP Core starter auto-configures a `GcpProjectIdProvider`. If a
+`spring.cloud.gcp.project-id` property is specified, the provided `GcpProjectIdProvider` returns
```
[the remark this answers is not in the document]
joaoandremartins: I love it! I think I'll go for a mixed approach -- break lines at each sentence final and wrap sentences per line width, so they're more easily readable on the IDE.

spring-cloud-gcp-docs/src/main/asciidoc/core.adoc
joaoandremartins: Meant to get around this. Swapped 0.20.1 to latest.

spring-cloud-gcp-docs/src/main/asciidoc/core.adoc
joaoandremartins: I'd rather take care of this in another PR.

spring-cloud-gcp-docs/src/main/asciidoc/index.adoc
joaoandremartins: It's because Spring Framework is its own thing that I didn't want to use it.
I'm afraid that Spring Framework is spring-core, whereas we integrate with Spring as a whole.
Maybe it's still fine to use Spring Framework in that case...

spring-cloud-gcp-docs/src/main/asciidoc/pubsub.adoc
joaoandremartins: I meant to address this object conversion issue in two places: #54 and then #92.
Because this is such a relevant issue, I'm pretty positive we will change the text when we get around to implementing this!

joaoandremartins:
PTAL

spring-cloud-gcp-docs/src/main/asciidoc/core.adoc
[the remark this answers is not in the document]
artembilan: Well, with the https://github.com/spring-gradle-plugins/dependency-management-plugin we can reach the same BOM behavior in the Gradle config as well:
```
buildscript {
	repositories {
		maven { url 'http://repo.spring.io/plugins-release' }
	}
	dependencies {
		classpath 'io.spring.gradle:dependency-management-plugin:1.0.3.RELEASE'
	}
}

ext {
	springCloudGcpVersion = '1.0.0.BUILD-SNAPSHOT'
}

dependencyManagement {
	imports {
		mavenBom "org.springframework.cloud:spring-cloud-gcp-dependencies:$springCloudGcpVersion"
}
```
But right, the main my concern was about an explicit version meanwhile we can use configured property from the project.
Thanks for that!

spring-cloud-gcp-docs/src/main/asciidoc/core.adoc
artembilan: I don't mind. Just an idea to improve things a bit.

spring-cloud-gcp-docs/src/main/asciidoc/index.adoc
markfisher: might be better to say something like:

```
...aims at bringing the best possible support for Spring applications to the Google Cloud Platform (GCP)
```

(could potentially be explicit that it's actually "Spring Boot" applications, given the reliance on autoconfig, etc)

Overall, it seems that it should be focused on the fact that GCP support will be as strong as possible, without implying that it would be "the platform of choice". Even though I'm personally glad that's the aim ;) , this is a Spring project and shouldn't really include any potentially competitive stance in the doc.

spring-cloud-gcp-docs/src/main/asciidoc/core.adoc
joaoandremartins: I was thinking about this after closing #122 and I can't tell if specifying e.g., `PUBSUB` instead of `https://www.googleapis.com/auth/pubsub` is actually useful.

Using such a shortcut would require users to look up some kind of table to see what the scope key is, or what it refers to (e.g., storage has multiple scopes, read, read_write, etc.).

This way, users can also provide any scopes they want, regardless of whether they're included in the enum or not, which is useful if a new project came out and we haven't added the scope yet.

WDYT?

spring-cloud-gcp-docs/src/main/asciidoc/index.adoc
joaoandremartins: How about

`
The Spring Cloud GCP project aims at making the Spring Framework a first-class citizen on the Google Cloud Platform (GCP).
`
?

spring-cloud-gcp-docs/src/main/asciidoc/core.adoc
artembilan: Yeah... that's valid point. Never mind for now. Let's what people think about this! That's not a first lib which configs are based on some urls

spring-cloud-gcp-docs/src/main/asciidoc/core.adoc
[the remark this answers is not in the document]
joaoandremartins: That actually reminds me, I should probably also this Gradle BOM bit to the Dependency Management section.
I'll even do that on this PR.

spring-cloud-gcp-docs/src/main/asciidoc/index.adoc
markfisher: sounds good to me. thanks!

joaoandremartins:
@sobychacko I verified that the content of `target/generated-docs` isn't published to https://docs.spring.io/spring-cloud-gcp/docs/1.0.0.BUILD-SNAPSHOT/reference/. It seems that URL contains the `target/docbook` content, which doesn't contain the HTML with the latest style that can be found in `target/generated-docs/index.html`.

Is there anywhere the latter is published to, that we can link from http://cloud.spring.io/spring-cloud-gcp/ ? Right now, it's linking to a reference document with the [old style](https://docs.spring.io/spring-cloud-gcp/docs/1.0.0.BUILD-SNAPSHOT/reference/htmlsingle/).

Issue #128, opened 2017-08-28
Configurable datasource pool from properties when building the datasource


saturnism:
Currently there is no way to specify connection pool size and a number of datasource configurations.  We should inherit whatever was configured in `spring.datasource`.  Or, configure `spring.cloud.gcp.sql.datasource` instead? Preferring the former for now.

joaoandremartins:
Fixed by #129.

PR #129, opened 2017-08-28
Let DatasourceAutoconfiguration create DS

saturnism:
For #128

### DIFF
--- a/spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/GcpCloudSqlProperties.java
+++ b/spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/GcpCloudSqlProperties.java
@@ -27,20 +27,9 @@ import org.springframework.boot.context.properties.ConfigurationProperties;
  */
 @ConfigurationProperties("spring.cloud.gcp.sql")
 public class GcpCloudSqlProperties {
-
 	private String databaseName;
 
 	private String instanceConnectionName;
-
-	private String jdbcUrl;
-
-	private String jdbcDriver;
-
-	private String userName = "root";
-
-	private String password = "";
-
-	private boolean initFailFast;
 
 	private Path credentialsLocation;
 
@@ -60,46 +49,6 @@ public class GcpCloudSqlProperties {
 
 	public void setInstanceConnectionName(String instanceConnectionName) {
 		this.instanceConnectionName = instanceConnectionName;
-	}
-
-	public String getJdbcUrl() {
-		return this.jdbcUrl;
-	}
-
-	public void setJdbcUrl(String jdbcUrl) {
-		this.jdbcUrl = jdbcUrl;
-	}
-
-	public String getJdbcDriver() {
-		return this.jdbcDriver;
-	}
-
-	public void setJdbcDriver(String jdbcDriver) {
-		this.jdbcDriver = jdbcDriver;
-	}
-
-	public String getUserName() {
-		return this.userName;
-	}
-
-	public void setUserName(String userName) {
-		this.userName = userName;
-	}
-
-	public String getPassword() {
-		return this.password;
-	}
-
-	public void setPassword(String password) {
-		this.password = password;
-	}
-
-	public boolean isInitFailFast() {
-		return this.initFailFast;
-	}
-
-	public void setInitFailFast(boolean initFailFast) {
-		this.initFailFast = initFailFast;
 	}
 
 	public Path getCredentialsLocation() {
--- a/spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/autoconfig/AppEngineCloudSqlJdbcInfoProvider.java
+++ b/spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/autoconfig/AppEngineCloudSqlJdbcInfoProvider.java
@@ -46,8 +46,7 @@ public class AppEngineCloudSqlJdbcInfoProvider implements CloudSqlJdbcInfoProvid
 
 	@Override
 	public String getJdbcUrl() {
-		return String.format("jdbc:google:mysql://%s/%s?user=%s&password=%s",
-				this.properties.getInstanceConnectionName(), this.properties.getDatabaseName(),
-				this.properties.getUserName(), this.properties.getPassword());
+		return String.format("jdbc:google:mysql://%s/%s",
+				this.properties.getInstanceConnectionName(), this.properties.getDatabaseName());
 	}
 }
--- a/spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/autoconfig/DefaultCloudSqlJdbcInfoProvider.java
+++ b/spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/autoconfig/DefaultCloudSqlJdbcInfoProvider.java
@@ -19,7 +19,6 @@ package org.springframework.cloud.gcp.sql.autoconfig;
 import org.springframework.cloud.gcp.sql.CloudSqlJdbcInfoProvider;
 import org.springframework.cloud.gcp.sql.GcpCloudSqlProperties;
 import org.springframework.util.Assert;
-import org.springframework.util.StringUtils;
 
 /**
  * Provides default JDBC driver class name and constructs the JDBC URL for Cloud SQL v2
@@ -36,27 +35,21 @@ public class DefaultCloudSqlJdbcInfoProvider implements CloudSqlJdbcInfoProvider
 	public DefaultCloudSqlJdbcInfoProvider(GcpCloudSqlProperties properties) {
 		this.properties = properties;
 		Assert.hasText(this.properties.getDatabaseName(), "A database name must be provided.");
-		if (StringUtils.isEmpty(this.properties.getJdbcUrl())) {
-			Assert.hasText(this.properties.getInstanceConnectionName(),
-					"An instance connection name must be provided. Refer to "
-							+ GcpCloudSqlAutoConfiguration.INSTANCE_CONNECTION_NAME_HELP_URL
-							+ " for more information.");
-		}
+		Assert.hasText(this.properties.getInstanceConnectionName(),
+				"An instance connection name must be provided. Refer to "
+						+ GcpCloudSqlAutoConfiguration.INSTANCE_CONNECTION_NAME_HELP_URL
+						+ " for more information.");
 	}
 
 	@Override
 	public String getJdbcDriverClass() {
-		return StringUtils.isEmpty(this.properties.getJdbcDriver())
-				? this.properties.getDatabaseType().getJdbcDriverName()
-				: this.properties.getJdbcDriver();
+		return this.properties.getDatabaseType().getJdbcDriverName();
 	}
 
 	@Override
 	public String getJdbcUrl() {
-		return StringUtils.isEmpty(this.properties.getJdbcUrl())
-				? String.format(this.properties.getDatabaseType().getJdbcUrlTemplate(),
+		return String.format(this.properties.getDatabaseType().getJdbcUrlTemplate(),
 				this.properties.getDatabaseName(),
-				this.properties.getInstanceConnectionName())
-				: this.properties.getJdbcUrl();
+				this.properties.getInstanceConnectionName());
 	}
 }
--- a/spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/autoconfig/GcpCloudSqlAutoConfiguration.java
+++ b/spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/autoconfig/GcpCloudSqlAutoConfiguration.java
@@ -21,7 +21,6 @@ import java.io.IOException;
 import java.nio.file.Files;
 
 import com.google.cloud.sql.CredentialFactory;
-
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 
@@ -44,6 +43,7 @@ import org.springframework.cloud.gcp.sql.SqlCredentialFactory;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Conditional;
 import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.Primary;
 import org.springframework.util.StringUtils;
 
 /**
@@ -110,6 +110,30 @@ public class GcpCloudSqlAutoConfiguration {
 		return defaultProvider;
 	}
 
+	@Bean
+	@Primary
+	public DataSourceProperties cloudSqlDataSourceProperties(DataSourceProperties properties,
+			CloudSqlJdbcInfoProvider cloudSqlJdbcInfoProvider) {
+		if (StringUtils.isEmpty(properties.getUsername())) {
+			properties.setUsername("root");
+			LOGGER.warn("spring.datasource.username is not specified. Setting default username.");
+		}
+		if (StringUtils.isEmpty(properties.getDriverClassName())) {
+			properties.setDriverClassName(cloudSqlJdbcInfoProvider.getJdbcDriverClass());
+		}
+		else {
+			LOGGER.warn(
+					"spring.datasource.driver-class-name is specified. Not using generated Cloud SQL configuration");
+		}
+		if (StringUtils.isEmpty(properties.getUrl())) {
+			properties.setUrl(cloudSqlJdbcInfoProvider.getJdbcUrl());
+		}
+		else {
+			LOGGER.warn("spring.datasource.jdbc-url is specified. Not using generated Cloud SQL configuration");
+		}
+		return properties;
+	}
+
 	/**
 	 * Set credentials to be used by the Google Cloud SQL socket factory.
 	 *
@@ -167,31 +191,4 @@ public class GcpCloudSqlAutoConfiguration {
 			LOGGER.debug(" Error reading Cloud SQL credentials file.", ioe);
 		}
 	}
-
-	@Configuration
-	protected static class DataSourcePropertiesMutatorConfiguration {
-
-		public DataSourcePropertiesMutatorConfiguration(DataSourceProperties properties,
-				CloudSqlJdbcInfoProvider cloudSqlJdbcInfoProvider) {
-			if (StringUtils.isEmpty(properties.getUsername())) {
-				properties.setUsername("root");
-				LOGGER.warn("spring.datasource.username is not specified. Setting default username.");
-			}
-			if (StringUtils.isEmpty(properties.getDriverClassName())) {
-				properties.setDriverClassName(cloudSqlJdbcInfoProvider.getJdbcDriverClass());
-			}
-			else {
-				LOGGER.warn("spring.datasource.driver-class-name is specified. " +
-						"Not using generated Cloud SQL configuration");
-			}
-			if (StringUtils.isEmpty(properties.getUrl())) {
-				properties.setUrl(cloudSqlJdbcInfoProvider.getJdbcUrl());
-			}
-			else {
-				LOGGER.warn("spring.datasource.jdbc-url is specified. Not using generated Cloud SQL configuration");
-			}
-		}
-
-	}
-
 }
--- a/spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/autoconfig/GcpCloudSqlAutoConfiguration.java
+++ b/spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/autoconfig/GcpCloudSqlAutoConfiguration.java
@@ -19,13 +19,9 @@ package org.springframework.cloud.gcp.sql.autoconfig;
 import java.io.File;
 import java.io.IOException;
 import java.nio.file.Files;
-import java.security.GeneralSecurityException;
-
-import javax.sql.DataSource;
 
 import com.google.cloud.sql.CredentialFactory;
-import com.zaxxer.hikari.HikariConfig;
-import com.zaxxer.hikari.HikariDataSource;
+
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 
@@ -34,7 +30,9 @@ import org.springframework.boot.autoconfigure.AutoConfigureAfter;
 import org.springframework.boot.autoconfigure.AutoConfigureBefore;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.boot.autoconfigure.jdbc.DataSourceAutoConfiguration;
+import org.springframework.boot.autoconfigure.jdbc.DataSourceProperties;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.cloud.gcp.core.AppEngineCondition;
 import org.springframework.cloud.gcp.core.GcpProjectIdProvider;
@@ -46,29 +44,31 @@ import org.springframework.cloud.gcp.sql.SqlCredentialFactory;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Conditional;
 import org.springframework.context.annotation.Configuration;
+import org.springframework.util.StringUtils;
 
 /**
  * Google Cloud SQL starter.
  *
  * <p>
- * Provides Google Cloud SQL instance connectivity through Spring JDBC by providing only a database
- * and instance connection name.
+ * Provides Google Cloud SQL instance connectivity through Spring JDBC by providing only a
+ * database and instance connection name.
  *
  * @author João André Martins
  */
 @Configuration
-@EnableConfigurationProperties(GcpCloudSqlProperties.class)
+@EnableConfigurationProperties({ GcpCloudSqlProperties.class, DataSourceProperties.class })
 @ConditionalOnClass(GcpProjectIdProvider.class)
+@ConditionalOnProperty(name = "spring.cloud.gcp.sql.enabled", havingValue = "true", matchIfMissing = true)
 @AutoConfigureBefore(DataSourceAutoConfiguration.class)
 @AutoConfigureAfter(GcpContextAutoConfiguration.class)
 public class GcpCloudSqlAutoConfiguration {
 
-	private static final Log LOGGER = LogFactory.getLog(GcpCloudSqlAutoConfiguration.class);
-
 	public final static String INSTANCE_CONNECTION_NAME_HELP_URL =
 			"https://github.com/spring-cloud/spring-cloud-gcp/tree/master/"
-					+ "spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql"
-					+ "#google-cloud-sql-instance-connection-name";
+			+ "spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql"
+			+ "#google-cloud-sql-instance-connection-name";
+
+	private static final Log LOGGER = LogFactory.getLog(GcpCloudSqlAutoConfiguration.class);
 
 	@Autowired
 	private GcpCloudSqlProperties gcpCloudSqlProperties;
@@ -76,12 +76,14 @@ public class GcpCloudSqlAutoConfiguration {
 	@Autowired
 	private GcpProperties gcpProperties;
 
+	@Autowired
+	private DataSourceProperties dataSourceProperties;
+
 	@Bean
 	@ConditionalOnMissingBean(CloudSqlJdbcInfoProvider.class)
 	@Conditional(AppEngineCondition.class)
 	public CloudSqlJdbcInfoProvider appengineCloudSqlJdbcInfoProvider() {
-		CloudSqlJdbcInfoProvider appEngineProvider =
-				new AppEngineCloudSqlJdbcInfoProvider(this.gcpCloudSqlProperties);
+		CloudSqlJdbcInfoProvider appEngineProvider = new AppEngineCloudSqlJdbcInfoProvider(this.gcpCloudSqlProperties);
 
 		if (LOGGER.isInfoEnabled()) {
 			LOGGER.info("App Engine JdbcUrl provider. Connecting to "
@@ -90,6 +92,7 @@ public class GcpCloudSqlAutoConfiguration {
 		}
 
 		setCredentialsProperty();
+		mutateDatasourceProperties(appEngineProvider);
 
 		return appEngineProvider;
 	}
@@ -97,8 +100,7 @@ public class GcpCloudSqlAutoConfiguration {
 	@Bean
 	@ConditionalOnMissingBean(CloudSqlJdbcInfoProvider.class)
 	public CloudSqlJdbcInfoProvider defaultJdbcInfoProvider() {
-		CloudSqlJdbcInfoProvider defaultProvider =
-				new DefaultCloudSqlJdbcInfoProvider(this.gcpCloudSqlProperties);
+		CloudSqlJdbcInfoProvider defaultProvider = new DefaultCloudSqlJdbcInfoProvider(this.gcpCloudSqlProperties);
 
 		if (LOGGER.isInfoEnabled()) {
 			LOGGER.info("Default " + this.gcpCloudSqlProperties.getDatabaseType().name()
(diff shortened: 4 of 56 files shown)


spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/autoconfig/GcpCloudSqlAutoConfiguration.java
```diff
@@ -131,34 +132,34 @@ public CloudSqlJdbcInfoProvider cloudSqlJdbcInfoProvider(SQLAdmin sqlAdmin,
 	}
 
 	@Bean
+	@ConfigurationProperties(prefix = "spring.datasource")
```
joaoandremartins: A problem with this approach is that we still don't have access to those low level knobs, like max pool size, etc.

I was thinking we could leverage [DataSourceConfiguration](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jdbc/DataSourceConfiguration.java) somehow, but it isn't looking like we can right off the bat. e.g., the class is package protected, not integrated with our `CloudSqlJdbcInfoProvider`, etc. But maybe there's still a way to use [DataSourceProperties](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jdbc/DataSourceProperties.java) with our stuff, to support a more customized Hikari `DataSource` (and possibly others).

joaoandremartins: @artembilan might have an opinion about this

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/autoconfig/GcpCloudSqlAutoConfiguration.java
```diff
@@ -131,34 +132,34 @@ public CloudSqlJdbcInfoProvider cloudSqlJdbcInfoProvider(SQLAdmin sqlAdmin,
...
+	@Primary
```
joaoandremartins: Isn't this Primary going to mess with App Engine? If not, is it really necessary here? I thought with the `@ConditionalOnMissingBeans` there was always only one @Bean.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/test/java/org/springframework/cloud/gcp/sql/autoconfig/GcpCloudSqlAutoConfigurationMockTests.java
```diff
@@ -153,6 +153,27 @@ public void test() {
 	}
 
 	@TestPropertySource(properties = {
+			"spring.cloud.gcp.sql.instanceName=test-instance",
+			"spring.datasource.hikari.connectionTestQuery=select 1",
+			"spring.datasource.hikari.maximumPoolSize=19"
+
+	})
+	public static class GcpCloudSqlAutoConfigurationWithDatasourceProperties
```
joaoandremartins: I don't think this test is being ran by Maven because its name doesn't end in `Test` or `Tests`. I suspect it's going to fail if you run it, since those settings like max pool size shouldn't be being set IIUC.

saturnism: you are absolutely right! I was wondering if it ran before i left for the day :D i have another POC abt to be committed. But would need to discuss which options would make sense.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/test/java/org/springframework/cloud/gcp/sql/autoconfig/GcpCloudSqlAutoConfigurationMockTests.java
```diff
@@ -80,14 +82,14 @@ public void test() {
 					+ "&socketFactory=com.google.cloud.sql.mysql.SocketFactory&useSSL=false",
 					this.urlProvider.getJdbcUrl());
 			assertEquals("root", dataSource.getUsername());
-			assertEquals("", dataSource.getPassword());
+			assertEquals(null, dataSource.getPassword());
```
joaoandremartins: This should be an empty string by default.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/autoconfig/GcpCloudSqlAutoConfiguration.java
```diff
@@ -131,34 +129,31 @@ public CloudSqlJdbcInfoProvider cloudSqlJdbcInfoProvider(SQLAdmin sqlAdmin,
...
-		hikariConfig.setDriverClassName(cloudSqlJdbcInfoProvider.getJdbcDriverClass());
-		hikariConfig.setJdbcUrl(cloudSqlJdbcInfoProvider.getJdbcUrl());
-		hikariConfig.setUsername(this.gcpCloudSqlProperties.getUserName());
-		hikariConfig.setPassword(this.gcpCloudSqlProperties.getPassword());
-		// Setting this is useful to disable connection initialization. Especially in unit tests.
-		hikariConfig.setInitializationFailFast(this.gcpCloudSqlProperties.isInitFailFast());
-
-		return new HikariDataSource(hikariConfig);
+	@Primary
+	public DataSourceProperties enhancedDatasourceProperties(DataSourceProperties properties,
```
joaoandremartins: As part of this, we should probably remove `jdbcUrl`, `jdbcDriver`, `userName` and `password` from `GcpCloudSqlProperties`.
I think `DefaultCloudSqlJdbcInfoProvider` should also be modified to also take in a `DataSourceProperties` object and immediately return its `url` and `driverClassName` fields if they have a value (much like we do today if `GcpCloudSqlProperties.jdbcUrl` has a value).

Also, I wonder which `spring-cloud-gcp-starter-sql` dependency is instantiating the `DataSource` bean from `DataSourceProperties`? I wonder is it `spring-jdbc`?

joaoandremartins: `GcpCloudSqlProperties` would only contain the GCP Cloud SQL specific properties, which is actually fair game. We can leave it to the `CloudSqlJdbcInfoProviders` to enforce the defaults, e.g., user="root", password="", jdbcDriver="com.mysql.jdbc.Driver", etc.

joaoandremartins:
We'll also need to update the SQL starter README.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/test/java/org/springframework/cloud/gcp/sql/autoconfig/GcpCloudSqlAutoConfigurationMockTests.java
joaoandremartins: I think we need it to be an empty string, otherwise it won't work with the default configuration of Cloud SQL instance ("root", "").

saturnism:
@artembilan would you be able to help us get Ollie or Phil to validate this approach?

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/autoconfig/DefaultCloudSqlJdbcInfoProvider.java
```diff
@@ -39,43 +39,42 @@
...
 
-				this.properties.setInstanceConnectionName(
-						String.format("%s:%s:%s", projectId, this.properties.getRegion(),
-								this.properties.getInstanceName()));
+		if (StringUtils.isEmpty(properties.getInstanceConnectionName())) {
+			Assert.hasText(this.properties.getInstanceName(),
+					"Instance Name is required, or specify Instance Connection Name explicitly.");
+			if (StringUtils.isEmpty(properties.getRegion())) {
+				try {
+					this.properties.setRegion(
```
artembilan: I somehow think it's not good idea to *mutate* configuration properties.
What is the purpose to do this, please?

saturnism: Hence the first ping that we needed validation on this approach.

The first iteration of the starter produces a datasource, but the datasource couldn't be configured via the properties.

We wanted to make sure the datasource connection pool is configurable. We can do it in two ways, one is to have our own datasource configuration prefix, e.g., `spring.cloud.gcp.sql.datasource`, or, we can utilize the existing one.

If we use our own prefix, then it seems like we'd have to duplicate the logic to detect different datasource pool implementations (hikari, tomcat, etc).

Or, we could re-use the the existing configurations `spring.datasource.*`, and the only thing we really want to auto-configure is the driver class and JDBC URL.

This current POC is trying to auto-configure driver class and JDBC URL so that Spring Boot's DataSourceAutoconfiguration can manage the creation of the datasource as before. There is little reason to duplicate that logic.

saturnism:
Hence the first ping that we needed validation on this approach.

The first iteration of the starter produces a datasource, but the datasource couldn't be configured via the properties.

We wanted to make sure the datasource connection pool is configurable. We can do it in two ways, one is to have our own datasource configuration prefix, e.g., `spring.cloud.gcp.sql.datasource`, or, we can utilize the existing one.

If we use our own prefix, then it seems like we'd have to duplicate the logic to detect different datasource pool implementations (hikari, tomcat, etc).

Or, we could re-use the the existing configurations `spring.datasource.*`, and the only thing we really want to auto-configure is the driver class and JDBC URL.

This current POC is trying to auto-configure driver class and JDBC URL so that Spring Boot's DataSourceAutoconfiguration can manage the creation of the datasource as before. There is little reason to duplicate that logic.

saturnism:
For reference, the approach to use our own prefix is here: here: https://github.com/spring-cloud/spring-cloud-gcp/pull/137

artembilan:
There is something like `EnvironmentPostProcessor` to "hack" the properties before auto-configuration: https://docs.spring.io/spring-boot/docs/current/reference/html/howto-spring-boot-application.html#howto-customize-the-environment-or-application-context.

Don't you think that may help somehow here do not reinvent the wheel?

saturnism:
Absolutely didn't want to re-invent the wheel - Just want to make sure overriding the property values here is an OK approach. Happy to know a mechanism to do this as well.

Just to be sure, we'll need to:
1. add the post processor and register in spring.factories
1. create a new PropertySource and set the `spring.datasource.url` properties, etc.
1. add the newly created PropertySource to last, this way, user defined `spring.datasource.url` will still override our configuration.  We do need to warn the user or fail fast though if they specified `spring.cloud.gcp.sql....` and still defined their own `spring.datasource.url`

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/autoconfig/GcpCloudSqlAutoConfiguration.java
```diff
@@ -131,34 +130,36 @@ public CloudSqlJdbcInfoProvider cloudSqlJdbcInfoProvider(SQLAdmin sqlAdmin,
...
-		hikariConfig.setPassword(this.gcpCloudSqlProperties.getPassword());
-		// Setting this is useful to disable connection initialization. Especially in unit tests.
-		hikariConfig.setInitializationFailFast(this.gcpCloudSqlProperties.isInitFailFast());
-
-		return new HikariDataSource(hikariConfig);
+	@Primary
+	public DataSourceProperties dataSourceProperties(DataSourceProperties properties,
+			CloudSqlJdbcInfoProvider cloudSqlJdbcInfoProvider) {
+		if (StringUtils.isEmpty(properties.getUsername())) {
+			properties.setUsername("root");
```
joaoandremartins: Would be nice to set an empty password as well, unless `getPassword()` works well with `null`.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/GcpCloudSqlProperties.java
```diff
@@ -36,14 +36,6 @@
...
 
-	private String jdbcUrl;
-
-	private String jdbcDriver = "com.mysql.jdbc.Driver";
-
-	private String userName = "root";
-
-	private String password = "";
-
 	private boolean initFailFast;
```
joaoandremartins: `initFailFast` can also go if we're not creating the Hikari datasource ourselves.

saturnism:
@artembilan I've looked into `EnvironmentPostProcessor`. It seems to be harder to implement what we need. I.e., the post processor is triggered before any of the autoconfiguration is done and it can't inject any dependencies, such as the `CloudSqlJdbcInfoProvider`.

Is there any other way to produce the override `spring.datasource` properties so that we don't need to redo the logic in `DatasourceAutoConfiguration`?

artembilan:
> Is there any other way to produce the override spring.datasource properties so that we don't need to redo the logic in DatasourceAutoConfiguration?

Sorry for delay, @saturnism , I couldn't come back to you quickly because I just didn't have any appropriate answer for you.

So, if we can't use `EnvironmentPostProcessor` just because we really depend on some  auto-configuration (`CloudSqlJdbcInfoProvider`), let's forget this variant! 
Let's just live with the current state as is!
Too much time has already spent without results...

Or... if it is really acceptable by Spring Boot team to let us override `DataSourceProperties` from our auto-confing before `DataSourceAutoConfiguration`, let's take a look what we can do on the matter!

WDYT?

artembilan:
Right, that is the place I've seen your comment:
>  but it looks like the property overriding approach is good from spring team.

So, can we go this way trying to override `DataSourceProperties.setUrl()` from our auto-config ?

joaoandremartins:
@artembilan not sure if I'm following the reasoning here.

Currently, we're building our own `DataSource`, which mirrors a lot of what happens already in `spring-jdbc`. We could go on and manually add all the other `DataSource` configuration options, but it would be a lot of duplicate code, and we would only support Hikari `DataSources`.

Using the alternative suggested here by @saturnism , we would be able to fully leverage the existing `DataSource` construction mechanisms in `spring-jdbc` to build Hikari, but also Tomcat, etc. `DataSources`, and fully customize them.

I'm not sure what the drawbacks of this approach are, but if any, would they outweigh the benefit of not having any code duplication and providing all this support?

Either way, let's clarify which way we want to go and act fast here, since this is probably the only remaining issue before we can go M1. Hopefully by the end of next week we could have an M1 release!

artembilan:
Yes, I agree with all your comments.
Let's try to come up with the solution to override options in the `DataSourceProperties` and then let Spring Boot auto-configuration to make its work around `DataSource` population!

joaoandremartins:
@artembilan so, just to be absolutely clear, you agree with what we're doing on this PR? :smile:

artembilan:
> you agree with what we're doing

I agree with the idea, but I didn't tell that code is ideal.
Just don't overuse my reputation! :wink 

I will come back to overview the code the next week when it will be rebased and polished.

Thanks for trusting me still! 😄

joaoandremartins:
Yes, just wanted to validate the approach here of mutating the properties in a `@Configuration` file as opposed to using alternatives like `EnvironmentPostProcessor`.
There are still unresolved comments in the code and it needs a rebase on master.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/autoconfig/DefaultCloudSqlJdbcInfoProvider.java
```diff
@@ -36,27 +35,21 @@
 	public DefaultCloudSqlJdbcInfoProvider(GcpCloudSqlProperties properties) {
 		this.properties = properties;
 		Assert.hasText(this.properties.getDatabaseName(), "A database name must be provided.");
-		if (StringUtils.isEmpty(this.properties.getJdbcUrl())) {
```
joaoandremartins: Won't removing this always cause config to error out if instanceConnectionName isn't filled?

artembilan: There is `@Validate` and `@NotEmpty` annotations for the `@ConfigurationProperties` instead: https://docs.spring.io/spring-boot/docs/1.5.7.RELEASE/reference/html/boot-features-external-config.html#boot-features-external-config-validation

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/autoconfig/GcpCloudSqlAutoConfiguration.java
```diff
@@ -113,34 +111,44 @@ public CloudSqlJdbcInfoProvider defaultJdbcInfoProvider() {
...
-		hikariConfig.setDriverClassName(cloudSqlJdbcInfoProvider.getJdbcDriverClass());
-		hikariConfig.setJdbcUrl(cloudSqlJdbcInfoProvider.getJdbcUrl());
-		hikariConfig.setUsername(this.gcpCloudSqlProperties.getUserName());
-		hikariConfig.setPassword(this.gcpCloudSqlProperties.getPassword());
-		// Setting this is useful to disable connection initialization. Especially in unit tests.
-		hikariConfig.setInitializationFailFast(this.gcpCloudSqlProperties.isInitFailFast());
-
-		return new HikariDataSource(hikariConfig);
+	@Primary
+	public DataSourceProperties cloudSqlDataSourceProperties(DataSourceProperties properties,
```
artembilan: Why can't we just *mutate* injected `DataSourceProperties` instead of creating a new bean?
I understand that you are going to override that one for the `DataSourceAutoConfiguration` via `@Primary`, but this `@Bean` will make two of `DataSourceProperties` beans in the application meanwhile one of them becomes obsolete anyway.

saturnism: good point! is there a way to let spring call my method that has no return value, but still inject DataSourceProperties into the param?  E.g.,

@ Some annotation?
public void overrideCloudSqlDataSourceProperties(DataSourceProperties properties) {
 ...
}

artembilan: Well, I think need a separate inner `@Configuration` class with the `DataSourceProperties` and `CloudSqlJdbcInfoProvider` ctor injections and make an appropriate modification there.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/autoconfig/AppEngineCloudSqlJdbcInfoProvider.java
```diff
@@ -46,8 +46,7 @@ public String getJdbcDriverClass() {
 
 	@Override
 	public String getJdbcUrl() {
-		return String.format("jdbc:google:mysql://%s/%s?user=%s&password=%s",
-				this.properties.getInstanceConnectionName(), this.properties.getDatabaseName(),
-				this.properties.getUserName(), this.properties.getPassword());
+		return String.format("jdbc:google:mysql://%s/%s",
```
joaoandremartins: Why removing the username and password?
If an instance has username and password set, don't we need to specify them here?

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/autoconfig/DefaultCloudSqlJdbcInfoProvider.java
joaoandremartins: The issue with using validators here is that `jdbcUrl` can be empty is `instanceConnectionName`, and vice-versa.
I don't think validators have enough expression power to cover that?

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/autoconfig/DefaultCloudSqlJdbcInfoProvider.java
```diff
@@ -36,27 +35,21 @@
...
-		}
+		Assert.hasText(this.properties.getInstanceConnectionName(),
+				"An instance connection name must be provided. Refer to "
+						+ GcpCloudSqlAutoConfiguration.INSTANCE_CONNECTION_NAME_HELP_URL
+						+ " for more information.");
 	}
 
 	@Override
 	public String getJdbcDriverClass() {
-		return StringUtils.isEmpty(this.properties.getJdbcDriver())
```
joaoandremartins: Also, why removing this?
Users must be able to specify a custom JDBC driver through properties.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/autoconfig/DefaultCloudSqlJdbcInfoProvider.java
artembilan: Well, that can be done via validators as well. See sample here: https://github.com/spring-cloud-stream-app-starters/file/blob/master/spring-cloud-starter-stream-source-file/src/main/java/org/springframework/cloud/stream/app/file/source/FileSourceProperties.java#L90

But I'm fine to leave it as is here.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/autoconfig/AppEngineCloudSqlJdbcInfoProvider.java
artembilan: That's main point of this change: everything rest is fully up to the existing `DataSourceProperties` and `DataSourceAutoConfiguration` in Spring Boot.

Isn't it @saturnism ?

joaoandremartins: It sounds like we should be setting username and password in the `AppEngineCloudSqlJdbcInfoProvider` `@Bean` declaration in `GcpCloudSqlAutoConfiguration`.
Unless I'm missing something and the App Engine driver doesn't do anything with the SQL username and password.

saturnism: username/password is configured in the datasource, can be overwritten by spring.datasource.username and spring.datasource.password.  It does not seem necessary as part of URL construction.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/autoconfig/GcpCloudSqlAutoConfiguration.java
saturnism: i've tried a couple of inner autoconfigs - but unless DataSourceProperties is returned as a bean, the values are not kept.  Returning a bean seams to be the only way. WDYT if we clone the properties and return a new instance?

artembilan:
Oh, sorry! I somehow pushed commit to your branch: https://github.com/spring-cloud/spring-cloud-gcp/pull/129/commits/9d3c57faf526a68c979ec599204f77c575715140.

I hope you'll find it useful anyway since it resolves our concerns.

saturnism:
interesting, I tried a similar approach. although the tests passed - the actual app using wasn't able to create the datasource.  i'll have to test it out again.

saturnism:
yeah, as suspected, this didn't work for me in an real app either:

```
ava.lang.reflect.InvocationTargetException
        at org.springframework.boot.maven.AbstractRunMojo$LaunchRunner.run(AbstractRunMojo.java:527)
        at java.lang.Thread.run(Thread.java:748)
Caused by: org.springframework.beans.factory.UnsatisfiedDependencyException: Error creating bean with name 'org.springframework.boot.autoconfigure.orm.jpa.HibernateJpaAutoConfiguration': Unsatisfied dependency expressed through constructor parameter 0; nested exception is org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'dataSource' defined in class path resource [org/springframework/boot/autoconfigure/jdbc/DataSourceConfiguration$Tomcat.class]: Bean instantiation via factory method failed; nested exception is org.springframework.beans.BeanInstantiationException: Failed to instantiate [org.apache.tomcat.jdbc.pool.DataSource]: Factory method 'dataSource' threw exception; nested exception is org.springframework.boot.autoconfigure.jdbc.DataSourceProperties$DataSourceBeanCreationException: Cannot determine embedded database driver class for database type NONE. If you want an embedded database please put a supported one on the classpath. If you have database settings to be loaded from a particular profile you may need to active it (no profiles are currently active).
        at org.springframework.beans.factory.support.ConstructorResolver.createArgumentArray(ConstructorResolver.java:749)
        at org.springframework.beans.factory.support.ConstructorResolver.autowireConstructor(ConstructorResolver.java:189)
        at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.autowireConstructor(AbstractAutowireCapableBeanFactory.java:1193)
        at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.createBeanInstance(AbstractAutowireCapableBeanFactory.java:1095)
        at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.doCreateBean(AbstractAutowireCapableBeanFactory.java:513)
        at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.createBean(AbstractAutowireCapableBeanFactory.java:483)
        at org.springframework.beans.factory.support.AbstractBeanFactory$1.getObject(AbstractBeanFactory.java:306)
        at org.springframework.beans.factory.support.DefaultSingletonBeanRegistry.getSingleton(DefaultSingletonBeanRegistry.java:230)
        at org.springframework.beans.factory.support.AbstractBeanFactory.doGetBean(AbstractBeanFactory.java:302)
        at org.springframework.beans.factory.support.AbstractBeanFactory.getBean(AbstractBeanFactory.java:197)
        at org.springframework.beans.factory.support.ConstructorResolver.instantiateUsingFactoryMethod(ConstructorResolver.java:372)
        at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.instantiateUsingFactoryMethod(AbstractAutowireCapableBeanFactory.java:1173)
        at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.createBeanInstance(AbstractAutowireCapableBeanFactory.java:1067)
        at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.doCreateBean(AbstractAutowireCapableBeanFactory.java:513)
        at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.createBean(AbstractAutowireCapableBeanFactory.java:483)
        at org.springframework.beans.factory.support.AbstractBeanFactory$1.getObject(AbstractBeanFactory.java:306)
        at org.springframework.beans.factory.support.DefaultSingletonBeanRegistry.getSingleton(DefaultSingletonBeanRegistry.java:230)
        at org.springframework.beans.factory.support.AbstractBeanFactory.doGetBean(AbstractBeanFactory.java:302)
        at org.springframework.beans.factory.support.AbstractBeanFactory.getBean(AbstractBeanFactory.java:197)
        at org.springframework.context.support.AbstractApplicationContext.getBean(AbstractApplicationContext.java:1078)
        at org.springframework.context.support.AbstractApplicationContext.finishBeanFactoryInitialization(AbstractApplicationContext.java:857)
        at org.springframework.context.support.AbstractApplicationContext.refresh(AbstractApplicationContext.java:543)
        at org.springframework.boot.context.embedded.EmbeddedWebApplicationContext.refresh(EmbeddedWebApplicationContext.java:122)
        at org.springframework.boot.SpringApplication.refresh(SpringApplication.java:693)
        at org.springframework.boot.SpringApplication.refreshContext(SpringApplication.java:360)
        at org.springframework.boot.SpringApplication.run(SpringApplication.java:303)
        at org.springframework.boot.SpringApplication.run(SpringApplication.java:1118)
        at org.springframework.boot.SpringApplication.run(SpringApplication.java:1107)
        at com.example.gcpdemo.GcpDemoApplication.main(GcpDemoApplication.java:26)
        ... 6 more
Caused by: org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'dataSource' defined in class path resource [org/springframework/boot/autoconfigure/jdbc/DataSourceConfiguration$Tomcat.class]: Bean instantiation via factory method failed; nested exception is org.springframework.beans.BeanInstantiationException: Failed to instantiate [org.apache.tomcat.jdbc.pool.DataSource]: Factory method 'dataSource' threw exception; nested exception is org.springframework.boot.autoconfigure.jdbc.DataSourceProperties$DataSourceBeanCreationException: Cannot determine embedded database driver class for database type NONE. If you want an embedded database please put a supported one on the classpath. If you have database settings to be loaded from a particular profile you may need to active it (no profiles are currently active).
        at org.springframework.beans.factory.support.ConstructorResolver.instantiateUsingFactoryMethod(ConstructorResolver.java:599)
        at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.instantiateUsingFactoryMethod(AbstractAutowireCapableBeanFactory.java:1173)
        at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.createBeanInstance(AbstractAutowireCapableBeanFactory.java:1067)
        at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.doCreateBean(AbstractAutowireCapableBeanFactory.java:513)
        at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.createBean(AbstractAutowireCapableBeanFactory.java:483)
        at org.springframework.beans.factory.support.AbstractBeanFactory$1.getObject(AbstractBeanFactory.java:306)
        at org.springframework.beans.factory.support.DefaultSingletonBeanRegistry.getSingleton(DefaultSingletonBeanRegistry.java:230)
        at org.springframework.beans.factory.support.AbstractBeanFactory.doGetBean(AbstractBeanFactory.java:302)
        at org.springframework.beans.factory.support.AbstractBeanFactory.getBean(AbstractBeanFactory.java:202)
        at org.springframework.beans.factory.config.DependencyDescriptor.resolveCandidate(DependencyDescriptor.java:208)
        at org.springframework.beans.factory.support.DefaultListableBeanFactory.doResolveDependency(DefaultListableBeanFactory.java:1138)
        at org.springframework.beans.factory.support.DefaultListableBeanFactory.resolveDependency(DefaultListableBeanFactory.java:1066)
        at org.springframework.beans.factory.support.ConstructorResolver.resolveAutowiredArgument(ConstructorResolver.java:835)
        at org.springframework.beans.factory.support.ConstructorResolver.createArgumentArray(ConstructorResolver.java:741)
        ... 34 more
Caused by: org.springframework.beans.BeanInstantiationException: Failed to instantiate [org.apache.tomcat.jdbc.pool.DataSource]: Factory method 'dataSource' threw exception; nested exception is org.springframework.boot.autoconfigure.jdbc.DataSourceProperties$DataSourceBeanCreationException: Cannot determine embedded database driver class for database type NONE. If you want an embedded database please put a supported one on the classpath. If you have database settings to be loaded from a particular profile you may need to active it (no profiles are currently active).
        at org.springframework.beans.factory.support.SimpleInstantiationStrategy.instantiate(SimpleInstantiationStrategy.java:189)
        at org.springframework.beans.factory.support.ConstructorResolver.instantiateUsingFactoryMethod(ConstructorResolver.java:588)
        ... 47 more
Caused by: org.springframework.boot.autoconfigure.jdbc.DataSourceProperties$DataSourceBeanCreationException: Cannot determine embedded database driver class for database type NONE. If you want an embedded database please put a supported one on the classpath. If you have database settings to be loaded from a particular profile you may need to active it (no profiles are currently active).
        at org.springframework.boot.autoconfigure.jdbc.DataSourceProperties.determineDriverClassName(DataSourceProperties.java:246)
        at org.springframework.boot.autoconfigure.jdbc.DataSourceProperties.initializeDataSourceBuilder(DataSourceProperties.java:183)
        at org.springframework.boot.autoconfigure.jdbc.DataSourceConfiguration.createDataSource(DataSourceConfiguration.java:42)
        at org.springframework.boot.autoconfigure.jdbc.DataSourceConfiguration$Tomcat.dataSource(DataSourceConfiguration.java:56)
        at org.springframework.beans.factory.support.SimpleInstantiationStrategy.instantiate(SimpleInstantiationStrategy.java:162)
        ... 48 more
```

artembilan:
@saturnism ,

One more attempt then!
If this last commit doesn't work too, let's come back to your original solution with the `@Primary` `DataSourceProperties` bean definition!

saturnism:
@artembilan i tried both of those tricks (PostControl and afterPropertiesSet earlier too ;) it didn't work for me... this new update didn't work either :( I'm not entirely sure why.

saturnism:
@joaoandremartins can you try the PR branch and see if Artem's latest change works for you?
i.e., https://help.github.com/articles/checking-out-pull-requests-locally/

joaoandremartins:
I suggest we release Ray's changes. They will make the API set in stone and then we can iterate on the best way of mutating the properties object instead of generating a new pointer. We need to get this M1 out the door ASAP!

Thoughts?

artembilan:
>  We need to get this M1 out the door ASAP!

OK. Fine with me.

But how about to give it a last chance tomorrow together and try to figure out what's going on in debug mode? I believe that pairing on the the same machine may bring some interesting results.
If we don't have anything valuable, we'll definitely go the bean override way.

joaoandremartins:
@artembilan how does your configuration look like for the case where it works? Do you specify `driver-class-name` or `url` anywhere?

joaoandremartins:
If you're using a configuration similar to the below and nothing else, then you should be seeing the same thing as we are.

```
spring.cloud.gcp.sql.instance-connection-name=[instance-connection-name]
spring.cloud.gcp.sql.database-name=[db-name]
spring.datasource.username=[username]
spring.datasource.password=[password]
```

artembilan:
I don't have anything specific. I only run tests in this project.
Well, that might be a signal that we just don't have an appropriate test to reflect what you have in your target application since all tests pass, but your app doesn't.

joaoandremartins:
Definitely worth investigating why these changes didn't trigger off any tests.
We should also roll back @artembilan 's changes and then possibly merged this PR.

@saturnism @artembilan thoughts?

artembilan:
OK. After some debugging the  `spring-cloud-gcp-sql-example` I have figured out what's going on.

There is `WebController` which is autowired with the `JdbcTemplate`. The first one is end user bean so it is initialized as early as possible to let to override auto-configurations. Since it must be initialized the last one is initialized for injection as well. But this one indeed pulls the `DataSource` bean, which, in turn, pulls the `DataSourceProperties`. And that one comes as a default one. No body in this cycle is going to call our `CloudSqlJdbcInfoProvider` to let it to mutate `DataSourceProperties`!

So, let's really revert all my changes and merge it as is - with the `@Primary` for the `DataSourceProperties`!

I still don't like this approach though, but we can live with it for now.

joaoandremartins:
@artembilan I reverted your commits. This is ready for another review!

artembilan:
Good. Thanks, @joaoandremartins !

Would you mind to explain why now we have too much not related to the subject changes in this PR?

joaoandremartins:
Thanks for the catch, I fixed it.
We still need docs here, but I suggest we address that in a separate PR so as not to drag this any more.
There's also the issue where the password is null in the tests, but an empty password is working when I manually test it.
WDYT?

artembilan:
Docs in the separate issue/PR.

We can't assume what database vendor is going to use end-user, so we can't make an assumption about `null` instead of empty string. We can make empty string from null configuration property, but I doubt we can opposite...

PR #135, opened 2017-09-01
Fixes checkstyle configuration

### LINKED ISSUE #131: Configure `checkstyle` plugin to allow for builds from submodules
Currently, `checkstyle` plugin is configured with `src/checkstyle/` path : this approach does not allow for builds from sub-modules like `google-cloud-examples`, fails on not finding the checkstyle configuration.

[This](https://maven.apache.org/plugins/maven-checkstyle-plugin/examples/multi-module-config.html) seems like the recommended solution.

joaoandremartins:
It is now possible to build from any sub-module without getting the
annoying checkstyle warning.

Fixes #131

### DIFF
--- a/CONTRIBUTING.adoc
+++ b/CONTRIBUTING.adoc
@@ -1,25 +1,28 @@
 = Code formatting guidelines
 
-* The directory ./src/eclipse has two files for use with code formatting,
-`eclipse-code-formatter.xml` for the majority of the code formatting rules and `eclipse.importorder`
-to order the import statements.
+The directory src/eclipse has two files for use with code formatting, `eclipse-code-formatter.xml`
+for the majority of the code formatting rules and `eclipse.importorder` to order the import
+statements.
 
-* In Eclipse you import these files by navigating `Windows -> Preferences` and then the menu items
+== Eclipse
+Import these files by navigating `Windows -> Preferences` and then the menu items
 `Preferences > Java > Code Style > Formatter` and `Preferences > Java > Code Style >
 Organize Imports` respectively.
 
-* In IntelliJ IDEA, install the plugin `Eclipse Code Formatter`.  You can find it by searching in
-"Browse Repositories", under `Settings > Plugins` within IDEA (Once installed, you will need to
-reboot IDEA for it to take effect).
+== IntelliJ IDEA
+Install the plugin `Eclipse Code Formatter`. You can find it by searching in "Browse Repositories",
+under `Settings > Plugins` within IDEA (Once installed, you will need to reboot IDEA for it to take
+effect).
+
 Then navigate to `Settings > Other Settings` (this might be under `Preferences` on Mac) and select
 the Eclipse Code Formatter. Select the `eclipse-code-formatter.xml` file for the field `Eclipse Java
 Formatter config file` and the file `eclipse.importorder` for the field `Import order`.
 Enable the `Eclipse code formatter` by clicking `Use the Eclipse code formatter` radio button at the
 top of the page, then click the *OK* button.
 
-** NOTE: If you configure the `Eclipse Code Formatter` from `File > Other Settings > Default
+* NOTE: If you configure the `Eclipse Code Formatter` from `File > Other Settings > Default
 Settings`, it will set this policy across all of your IDEA projects.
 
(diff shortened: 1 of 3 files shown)


joaoandremartins:
The problem with getting it from source is that the files aren't found when building a sub-project, so you have to temporarily disable checkstyle to build sub-projects...
Although not being able to check tests isn't great either.

I'll look for alternatives.

artembilan:
I somehow think there should be something like `${rootDirectory}` in Maven to take into account for the proper path to the config files. For example: https://stackoverflow.com/questions/3084629/finding-the-root-directory-of-a-multi-module-maven-reactor-project

artembilan:
> Is there a problem with that approach?

Then it's not an issue for this project, but for that `spring-cloud-butild-tools` one...
I don't think that we can somehow effect that root project for all the Spring Cloud umbrella and don't have choice unless go ahead with some local solution or just continue to live with the fact that Checkstyle is going to work only if we run the whole project. Anyway you always can skip tests to have just only Checkstyle task results.

meltsufin:
@artembilan I wasn't suggesting that we change the `checkstyle.xml` in `spring-cloud-build-tools` that is used by other Spring projects. I was thinking of just adding a `checkstyle-strict.xml` there and making it available to be used for `spring-cloud-gcp` and other Spring projects that want to strictly adhere to Spring code style. You can control which config file is used via the checkstyle plugin configuration.

joaoandremartins:
I think @artembilan 's concern is that we'll need to ask for a review any time we want to change any checkstyle rule. I don't see adding our checkstyle to spring-cloud-build-tools as a very necessary dependency, especially if we're using our own rules.

While it's not possible to run `mvn test`, it's possible to run `mvn package/install` (which also runs tests) and run test on each module individually. Therefore, I prefer this way to what we had previously, which is also the solution for the current issue.

artembilan:
> You can control which config file is used via the checkstyle plugin configuration.

Why do we need it there though, if we can use that configuration in the project directly?

That is unfortunate that Maven doesn't provide the proper way to specify the multi-project root dir, but another solution is to specify `maven-checkstyle-plugin` in each sub-projects with proper relative path to the `/src/checkstyle/checkstyle.xml`.
Or... Just live with the Checkstyle when we run only from the root!

meltsufin:
>Why do we need it there though, if we can use that configuration in the project directly?

Since the [codestyle](https://github.com/spring-projects/spring-framework/wiki/Spring-Framework-Code-Style) it implements is universal to all Spring projects, why not benefit other Spring projects by contributing it to a common module? I'm just trying to think of the larger Spring ecosystem.

artembilan:
>  I'm just trying to think of the larger Spring ecosystem.

Yeah... That's good idea, but unfortunately the reality is more complicated then it feels and many projects doesn't follow the same code style rules. You can raise the concern though in that mentioned project. But again - that is already outside of scope this our project.

joaoandremartins:
> ...another solution is to specify maven-checkstyle-plugin in each sub-projects with proper relative path to the /src/checkstyle/checkstyle.xml.
> Or... Just live with the Checkstyle when we run only from the root!

Between these two solutions and the one suggested here, I prefer this one due to the reasons previously mentioned (only top level mvn test doesn't work).

artembilan:
I'm against this PR because we have the mentioned spring-cloud/spring-cloud-stream#1037 issue.
And I'm pretty sure all of us are going to bump the problem when we run only `mvn test` without `install`.

I think that's really the Maven problem that it isn't so flexible for us without `rootDir` and and such a vulnerable Checkstyle plugin.

I'm pretty fine with the current state and always run `mvnw clean test` (or at least `validate`) from the root.
If you want to play only with the sub-modules, we should supply the plugin in those sub-projects. Just no choice! Always rely on the `install` is not a solution. We can disable `Checkstyle`, if we are not in the root project: the Travis here in PR is going to build and validate the code changes anyway...

joaoandremartins:
@artembilan turns out there was a much simpler fix to this issue that didn't involve creating a new module.

Adding a `<headerLocation>` to the checkstyle configuration and the `${checkstyle.header.file}` variable to checkstyle.xml seems to fix all our issues. LMKWYT

PR #137, opened 2017-09-02
For #128 - Use Datasource Builder to build a Hikari datasource & apply configuration properties

saturnism:
For #128


joaoandremartins:
This PR seems to be concurrent with #129 . What do you plan to do with this one?

saturnism:
this was just an example to the alternative approach. but it looks like the property overriding approach is good from spring team. will close.

PR #138, opened 2017-09-02
Sleuth integration with Stackdriver Trace

saturnism:
For #139 - Added ability to forward Sleuth trace spans to Stackdriver Trace.

### DIFF
--- a/spring-cloud-gcp-dependencies/pom.xml
+++ b/spring-cloud-gcp-dependencies/pom.xml
@@ -18,7 +18,10 @@
 	<properties>
 		<google-cloud-core.version>1.2.3</google-cloud-core.version>
 		<google-cloud-pubsub.version>0.20.1-beta</google-cloud-pubsub.version>
+		<google-cloud-trace.version>0.20.1-alpha</google-cloud-trace.version>
 		<google-cloud-storage.version>1.2.1</google-cloud-storage.version>
+		<cloud-trace-java.version>0.4.0</cloud-trace-java.version>
+		<spring-cloud-sleuth.version>1.3.0.M1</spring-cloud-sleuth.version>
 	</properties>
 
 	<dependencyManagement>
@@ -36,6 +39,11 @@
 			<dependency>
 				<groupId>org.springframework.cloud</groupId>
 				<artifactId>spring-cloud-gcp-storage</artifactId>
+				<version>${project.version}</version>
+			</dependency>
+			<dependency>
+				<groupId>org.springframework.cloud</groupId>
+				<artifactId>spring-cloud-gcp-trace</artifactId>
 				<version>${project.version}</version>
 			</dependency>
 			<dependency>
@@ -58,11 +66,47 @@
 				<artifactId>google-cloud-storage</artifactId>
 				<version>${google-cloud-storage.version}</version>
 			</dependency>
+			<dependency>
+				<groupId>com.google.cloud</groupId>
+				<artifactId>google-cloud-trace</artifactId>
+				<version>${google-cloud-trace.version}</version>
+			</dependency>
+			<dependency>
+				<groupId>com.google.cloud.trace.v1</groupId>
+				<artifactId>sink</artifactId>
+				<version>${cloud-trace-java.version}</version>
+				<exclusions>
+					<exclusion>
+						<groupId>io.grpc</groupId>
+						<artifactId>grpc-all</artifactId>
+					</exclusion>
+					<exclusion>
+						<groupId>com.google.protobuf</groupId>
+						<artifactId>protobuf-java</artifactId>
+					</exclusion>
+					<exclusion>
+						<groupId>com.google.guava</groupId>
+						<artifactId>guava</artifactId>
+					</exclusion>
+				</exclusions>
+			</dependency>
 
 			<dependency>
 				<groupId>commons-logging</groupId>
 				<artifactId>commons-logging</artifactId>
 				<version>1.2</version>
+			</dependency>
+
+			<dependency>
+				<groupId>org.springframework.cloud</groupId>
+				<artifactId>spring-cloud-sleuth-core</artifactId>
+				<version>${spring-cloud-sleuth.version}</version>
+			</dependency>
+
+			<dependency>
+				<groupId>org.springframework.cloud</groupId>
+				<artifactId>spring-cloud-starter-sleuth</artifactId>
+				<version>${spring-cloud-sleuth.version}</version>
 			</dependency>
 
 			<!-- spring-cloud-gcp-starter-core -->
@@ -83,6 +127,12 @@
 				<groupId>com.google.apis</groupId>
 				<artifactId>google-api-services-sqladmin</artifactId>
 				<version>v1beta4-rev45-1.22.0</version>
+				<exclusions>
+					<exclusion>
+						<groupId>com.google.guava</groupId>
+						<artifactId>guava-jdk5</artifactId>
+					</exclusion>
+				</exclusions>
 			</dependency>
 
 			<dependency>
@@ -111,6 +161,11 @@
 			<dependency>
 				<groupId>org.springframework.cloud</groupId>
 				<artifactId>spring-cloud-gcp-starter-sql</artifactId>
+				<version>${project.version}</version>
+			</dependency>
+			<dependency>
+				<groupId>org.springframework.cloud</groupId>
+				<artifactId>spring-cloud-gcp-starter-trace</artifactId>
 				<version>${project.version}</version>
 			</dependency>
 		</dependencies>
--- a/spring-cloud-gcp-starters/spring-cloud-gcp-starter-trace/README.adoc
+++ b/spring-cloud-gcp-starters/spring-cloud-gcp-starter-trace/README.adoc
@@ -34,7 +34,8 @@ The following properties are optional:
 # Number of threads to use by the underlying gRPC channel to send the trace request to Stackdriver
 spring.cloud.gcp.trace.executor-threads=[NUMBER_OF_SENDER_THREADS]
 
-# Buffer size, defaults to 4096, traces will be flushed to Stackdriver when buffered trace messages exceed this size.
+# Buffer size, defaults to 1% of Runtime.totalMemory(), traces will be flushed to Stackdriver when buffered trace
+# messages exceed this size.
 spring.cloud.gcp.trace.buffer-size-bytes=[BUFFER_SIZE_BYTES]
 
 # Buffered trace messages will be flushed to Stackdriver when buffered longer than scheduled delays.
@@ -51,3 +52,9 @@ spring.sleuth.web.skipPattern=(^cleanup.*|.+favicon.*)          # Ignore some UR
 
 Stackdriver Trace requires the use of 128-bit Trace ID. This starter ignores `spring.sleuth.traceId128` property
 and it'll always uses 128-bit Trace ID.
+
+== Enable APIs
+You must enable Stackdriver Trace API from the Google Cloud Console in order to capture traces.
+Navigate to the link:https://console.cloud.google.com/apis/api/cloudtrace.googleapis.com/overview[Stackdriver Trace API]
+for your project and make sure it's enabled.
+
--- a/spring-cloud-gcp-starters/spring-cloud-gcp-starter-trace/README.adoc
+++ b/spring-cloud-gcp-starters/spring-cloud-gcp-starter-trace/README.adoc
@@ -1,0 +1,50 @@
+= Google Cloud Stackdriver Trace Spring Boot Starter
+
+Maven coordinates, using Spring Cloud GCP BOM:
+
+[source,xml]
+----
+<dependency>
+    <groupId>org.springframework.cloud</groupId>
+    <artifactId>spring-cloud-gcp-starter-trace</artifactId>
+</dependency>
+----
+
+Gradle coordinates:
+
+[source]
+----
+dependencies {
+    compile group: 'org.springframework.cloud', name: 'spring-cloud-gcp-starter-trace', version: '1.0.0.BUILD-SNAPSHOT'
+}
+----
+
+This starter uses Spring Cloud Sleuth and provides a `SpanReporter` that sends the trace to Stackdriver Trace.
+
+All configurations are optional. You can override a number of default configuration values in addition to the core
+`spring.cloud.gcp.project-id` and `spring.cloud.gcp.credentials-location`.
+
+By default, this starter will send trace asynchronously using a flushable trace consumer that auto-flushes when
+its buffered trace messages exceed its buffer size or have been buffered for longer than its scheduled delay.
+
+The following properties are optional:
+
+[source,yaml]
+----
+# Number of threads to use by the underlying gRPC channel to send the trace request to Stackdriver
+spring.cloud.gcp.trace.executor-threads=[NUMBER_OF_SENDER_THREADS]
+
+# Buffer size, defaults to 4096, traces will be flushed to Stackdriver when buffered trace messages exceed this size.
+spring.cloud.gcp.trace.buffer-size-bytes=[BUFFER_SIZE_BYTES]
+
+# Buffered trace messages will be flushed to Stackdriver when buffered longer than scheduled delays.
+spring.cloud.gcp.trace.scheduled-delay-seconds=[SCHEDULED_DELAY_SECONDS]
+----
+
+You can use all of the core Spring Cloud Sleuth properties to control Sleuth's sampling rate, etc. For example:
+
+[source,yaml]
+---
+spring.sleuth.sampler.percentage=.1                             # 20% of traffic
+spring.sleuth.web.skipPattern=(^cleanup.*|.+favicon.*)          # Ignore some URL paths
+---
--- a/spring-cloud-gcp-starters/spring-cloud-gcp-starter-trace/README.adoc
+++ b/spring-cloud-gcp-starters/spring-cloud-gcp-starter-trace/README.adoc
@@ -29,7 +29,6 @@ its buffered trace messages exceed its buffer size or have been buffered for lon
 
 The following properties are optional:
 
-[source,yaml]
 ----
 # Number of threads to use by the underlying gRPC channel to send the trace request to Stackdriver
 spring.cloud.gcp.trace.executor-threads=[NUMBER_OF_SENDER_THREADS]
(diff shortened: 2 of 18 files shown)


jabubake:
w00t :)

spring-cloud-gcp-starters/spring-cloud-gcp-starter-trace/README.adoc
```diff
@@ -0,0 +1,50 @@
...
+
+# Buffered trace messages will be flushed to Stackdriver when buffered longer than scheduled delays.
+spring.cloud.gcp.trace.scheduled-delay-seconds=[SCHEDULED_DELAY_SECONDS]
+----
+
+You can use all of the core Spring Cloud Sleuth properties to control Sleuth's sampling rate, etc. For example:
+
+[source,yaml]
+---
+spring.sleuth.sampler.percentage=.1                             # 20% of traffic
```
joaoandremartins: 20% or 10%?

spring-cloud-gcp-starters/spring-cloud-gcp-starter-trace/src/main/java/org/springframework/cloud/gcp/trace/autoconfig/StackdriverTraceAutoConfiguration.java
```diff
@@ -0,0 +1,166 @@
...
+	public SpanReporter traceSpanReporter(Environment environment, GcpProjectIdProvider gcpProjectIdProvider,
+			LabelExtractor labelExtractor,
+			List<SpanAdjuster> spanAdjusters, TraceConsumer traceConsumer) {
+		String instanceId = IdUtils.getDefaultInstanceId(environment);
+		return new StackdriverTraceSpanListener(instanceId, gcpProjectIdProvider.getProjectId(), labelExtractor,
+				spanAdjusters, traceConsumer);
+	}
+
+	@Bean
+	@ConditionalOnMissingBean(Tracer.class)
```
joaoandremartins: Could this method just return a `Tracer`? It's not immediately clear to me why it can't.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-trace/README.adoc
```diff
@@ -0,0 +1,50 @@
...
+Gradle coordinates:
+
+[source]
+----
+dependencies {
+    compile group: 'org.springframework.cloud', name: 'spring-cloud-gcp-starter-trace', version: '1.0.0.BUILD-SNAPSHOT'
+}
+----
+
+This starter uses Spring Cloud Sleuth and provides a `SpanReporter` that sends the trace to Stackdriver Trace.
```
joaoandremartins: Which trace?

spring-cloud-gcp-starters/spring-cloud-gcp-starter-trace/README.adoc
saturnism: oops, it should be 10% in the comment.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-trace/README.adoc
saturnism: Will clarify here - it's Sleuth's trace information

spring-cloud-gcp-starters/spring-cloud-gcp-starter-trace/src/main/java/org/springframework/cloud/gcp/trace/autoconfig/StackdriverTraceAutoConfiguration.java
saturnism: I was wondering that too. But decided to follow the convention here: https://github.com/spring-cloud/spring-cloud-sleuth/blob/master/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/autoconfig/TraceAutoConfiguration.java

spring-cloud-gcp-starters/spring-cloud-gcp-starter-trace/README.adoc
```diff
@@ -0,0 +1,50 @@
+= Google Cloud Stackdriver Trace Spring Boot Starter
```
joaoandremartins: Perhaps we should explain somewhere here that, in order for this to work, the Trace API needs to be enabled from the Cloud Console to work.

Ideally we'd also have a code sample that shows this in action. I learned from [here](https://cloud.spring.io/spring-cloud-sleuth/) that just logging a line works. We could also direct people to the Cloud Console Trace page.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-trace/src/main/java/org/springframework/cloud/gcp/trace/GcpTraceProperties.java
```diff
@@ -0,0 +1,68 @@
...
+	 * Number of threads to use by the underlying gRPC channel to send the trace request to
+	 * Stackdriver.
+	 */
+	private int executorThreads = 4;
+
+	/**
+	 * Buffer size in bytes. Traces will be flushed to Stackdriver when buffered trace
+	 * messages exceed this size.
+	 */
+	private int bufferSizeBytes = 4096;
```
joaoandremartins: Keeping in line with other (not all, to be fixed) starters, we could have per-service credentials?

saturnism: yes we should - is the PR for per-service credentials already in? I can replicate if there is already one done. If not, we can apply it to all the starters in one shot.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-trace/README.adoc
saturnism: added Enable APIs section.

Let's add the sample afterwards into a `spring-cloud-gcp-examples` after Jisha's code is merged in. Otherwise we'll to deal w/ conflicts :(

saturnism: Added https://github.com/spring-cloud/spring-cloud-gcp/issues/146

spring-cloud-gcp-trace/src/main/java/org/springframework/cloud/gcp/trace/sleuth/LabelExtractor.java
```diff
@@ -0,0 +1,125 @@
...
+
+	public static final String DEFAULT_TIMESTAMP_FORMAT = "yyyy-MM-dd (HH:mm:ss.SSS z)";
+
+	private static final Map<String, String> DEFAULT_RENAME_MAP;
+	static {
+		DEFAULT_RENAME_MAP = new HashMap<>();
+		DEFAULT_RENAME_MAP.put("http.host", "/http/host");
+		DEFAULT_RENAME_MAP.put("http.method", "/http/method");
+		DEFAULT_RENAME_MAP.put("http.status_code", "/http/status_code");
+		DEFAULT_RENAME_MAP.put("http.request.size", "/request/size");
```
meltsufin: It should be `http/request/size` and `http/response/size` below.
See here: https://cloud.google.com/trace/docs/viewing-details#timeline

spring-cloud-gcp-starters/spring-cloud-gcp-starter-trace/src/main/java/org/springframework/cloud/gcp/trace/GcpTraceProperties.java
```diff
@@ -0,0 +1,85 @@
...
+
+	/**
+	 * Buffer size in bytes. Traces will be flushed to Stackdriver when buffered trace
+	 * messages exceed this size.
+	 *
+	 * This value is defaulted to 1% of the {@link Runtime#totalMemory()} in bytes.
+	 * However, be careful when running inside of a containerized environment. You should either set JVM's max heap
+	 * or set this value explicitly.
+	 */
+	private int bufferSizeBytes = onePercentOfRuntimeTotalMemory();
```
meltsufin: Maybe change the method to be `percentOfRuntimeTotalMemory(0.01)`? Also, is there a default somewhere in the trace library that we can leverage? The same goes for `executorThreads`.

spring-cloud-gcp-dependencies/pom.xml
```diff
@@ -18,7 +18,10 @@
 	<properties>
 		<google-cloud-core.version>1.2.3</google-cloud-core.version>
 		<google-cloud-pubsub.version>0.20.1-beta</google-cloud-pubsub.version>
+		<google-cloud-trace.version>0.20.1-alpha</google-cloud-trace.version>
 		<google-cloud-storage.version>1.2.1</google-cloud-storage.version>
+		<cloud-trace-java.version>0.4.0</cloud-trace-java.version>
+		<spring-cloud-sleuth.version>1.3.0.M1</spring-cloud-sleuth.version>
```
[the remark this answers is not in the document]
saturnism: spring boot dependencies doesn't have spring cloud dependencies.  Doc suggests using Spring Cloud Sleuth's BOM to include the dependency: https://cloud.spring.io/spring-cloud-sleuth/

Although, Sleuth is also part of the Spring Cloud release train. I don't think we can dependency import from spring cloud release train though. Because spring-cloud-gcp would ultimately be part of it.

spring-cloud-gcp-trace/src/main/java/org/springframework/cloud/gcp/trace/sleuth/LabelExtractor.java
saturnism: good catch!

spring-cloud-gcp-trace/src/main/java/org/springframework/cloud/gcp/trace/TraceServiceClientTraceConsumer.java
```diff
@@ -0,0 +1,58 @@
...
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
+/**
+ * This {@link TraceConsumer} immediately sends trace data to Stackdriver Trace.
+ *
+ * @author Ray Tsang
+ */
+public class TraceServiceClientTraceConsumer implements TraceConsumer, Closeable {
+	private static final Log LOGGER = LogFactory.getLog(TraceServiceClientTraceConsumer.class);
```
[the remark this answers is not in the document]
saturnism: good catch. removed.

saturnism:
@meltsufin `http:/` is on par w/ what zipkin offers at the moment as well, seems on-par. Does App Engine trace show something different today?

`http://localhost:8080/gcs/read` should show up when the buffer size is reached or when the delay period timed out. The UI is tricky - you may need to turn on auto-reload, and sometimes, reload the whole page.

joaoandremartins:
Also, can you please add a little entry to the top-level README too?

saturnism:
@joaoandremartins done!

meltsufin:
@marcingrzejszczak, Mark suggested that you might be interested in reviewing this PR. WDYT? Thanks!

spring-cloud-gcp-dependencies/pom.xml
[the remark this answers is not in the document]
marcingrzejszczak: @saturnism is right. You could either reference the Spring Cloud release train BOM and then you get the Sleuth version out of the box, or you provide the Spring Cloud Sleuth's BOM but you have to provide info which version of the BOM is it.

spring-cloud-gcp-dependencies/pom.xml
```diff
@@ -58,13 +66,49 @@
...
+				</exclusions>
+			</dependency>
 
 			<dependency>
 				<groupId>commons-logging</groupId>
 				<artifactId>commons-logging</artifactId>
 				<version>1.2</version>
 			</dependency>
 
+			<dependency>
```
marcingrzejszczak: Starter is enough - https://github.com/spring-cloud/spring-cloud-sleuth/blob/master/spring-cloud-starter-sleuth/pom.xml#L26-L29

spring-cloud-gcp-starters/spring-cloud-gcp-starter-trace/README.adoc
```diff
@@ -0,0 +1,60 @@
...
+
+# Buffered trace messages will be flushed to Stackdriver when buffered longer than scheduled delays.
+spring.cloud.gcp.trace.scheduled-delay-seconds=[SCHEDULED_DELAY_SECONDS]
+----
+
+You can use core Spring Cloud Sleuth properties to control Sleuth's sampling rate, etc. For example:
+
+[source,yaml]
+---
+spring.sleuth.sampler.percentage=.1                             # Trace 10% of the requests
```
marcingrzejszczak: 10% of requests are sent to Stackdriver. 100% of requests are traced (you'll see 100% of requests traced in the logs)

spring-cloud-gcp-starters/spring-cloud-gcp-starter-trace/src/main/java/org/springframework/cloud/gcp/trace/GcpTraceProperties.java
```diff
@@ -0,0 +1,86 @@
...
+
+	/**
+	 * Scheduled delay in seconds. Buffered trace messages will be flushed to Stackdriver when
+	 * buffered longer than scheduled delays.
+	 */
+	private int scheduledDelaySeconds = 10;
+
+	/**
+	 * A utility method to determine 1% of total memory based on Zipkin's AsyncReporter.
+	 * However, need to be careful about running this inside of a container environment since only certain versions
```
marcingrzejszczak: Yeah, that's a very good remark :)

spring-cloud-gcp-starters/spring-cloud-gcp-starter-trace/src/main/java/org/springframework/cloud/gcp/trace/autoconfig/StackdriverTraceAutoConfiguration.java
marcingrzejszczak: I have no idea why we put `DefaultTracer` there ;) It should be `Tracer`.

spring-cloud-gcp-trace/src/main/java/org/springframework/cloud/gcp/trace/sleuth/LabelExtractor.java
```diff
@@ -0,0 +1,125 @@
...
+import java.util.Map;
+
+import com.google.devtools.cloudtrace.v1.TraceSpan;
+
+import org.springframework.cloud.sleuth.Log;
+import org.springframework.cloud.sleuth.Span;
+import org.springframework.util.StringUtils;
+
+/**
+ * Translated from Andrian Cole's
```
marcingrzejszczak: Adrian Cole :)

spring-cloud-gcp-trace/src/main/java/org/springframework/cloud/gcp/trace/sleuth/LabelExtractor.java
```diff
@@ -0,0 +1,125 @@
...
+ */
+public class LabelExtractor {
+	public static final String DEFAULT_AGENT_NAME = "spring-cloud-gcp-trace";
+
+	public static final String DEFAULT_PREFIX = "cloud.spring.io/";
+
+	public static final String DEFAULT_TIMESTAMP_FORMAT = "yyyy-MM-dd (HH:mm:ss.SSS z)";
+
+	private static final Map<String, String> DEFAULT_RENAME_MAP;
+	static {
```
marcingrzejszczak: AFAIR these values can be customized in Sleuth since they are taken from `ConfigurationProperties` . I guess you could inject via constructor the `TraceKeys` and check the values of keys from the `Http` part of the props.

spring-cloud-gcp-dependencies/pom.xml
saturnism: `spring-cloud-gcp-trace` depends only on the core and not the starter, our starter does depend on the spring-cloud-starter-sleuth though. Makes sense to keep both?

spring-cloud-gcp-starters/spring-cloud-gcp-starter-trace/README.adoc
saturnism: ah good call. will update to "Send 10% of the request traces to Stackdriver"

spring-cloud-gcp-trace/src/main/java/org/springframework/cloud/gcp/trace/sleuth/LabelExtractor.java
saturnism: good catch!

spring-cloud-gcp-trace/src/main/java/org/springframework/cloud/gcp/trace/sleuth/LabelExtractor.java
saturnism: changing `TraceKey` feels like there are more fundamental changes to users that might be expecting Sleuth's default trace keys, and not necessarily need to tie it w/ Stackdriver-specific keys.

saturnism: OH wait, I think you meant looking up these keys can be from TraceKeys. That's a good idea.

saturnism: I think this should prob be passed in during autoconfiguration time.

saturnism: ptal

spring-cloud-gcp-dependencies/pom.xml
marcingrzejszczak: IMO no. Either pick the starter or core. Cause the starter picks core transitively.

joaoandremartins:
I would be great if you could add something to our reference document in this PR, on after this one is merged. :)

saturnism:
@joaoandremartins i just noticed the ref doc is merged. added a bug to make sure we have ref doc.

spring-cloud-gcp-dependencies/pom.xml
saturnism: Spoke w/ @marcingrzejszczak on hangout. This is the dependency management section so this was OK. `spring-cloud-gcp-trace` depends on core, and `spring-cloud-gcp-starter-trace` depends on starter

spring-cloud-gcp-starters/spring-cloud-gcp-starter-trace/README.adoc
```diff
@@ -0,0 +1,60 @@
...
+[source,yaml]
+----
+# Number of threads to use by the underlying gRPC channel to send the trace request to Stackdriver
+spring.cloud.gcp.trace.executor-threads=[NUMBER_OF_SENDER_THREADS]
+
+# Buffer size, defaults to 1% of Runtime.totalMemory(), traces will be flushed to Stackdriver when buffered trace
+# messages exceed this size.
+spring.cloud.gcp.trace.buffer-size-bytes=[BUFFER_SIZE_BYTES]
+
+# Buffered trace messages will be flushed to Stackdriver when buffered longer than scheduled delays.
```
meltsufin: We should mention what the default is.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-trace/README.adoc
```diff
@@ -0,0 +1,60 @@
...
+spring.cloud.gcp.trace.scheduled-delay-seconds=[SCHEDULED_DELAY_SECONDS]
+----
+
+You can use core Spring Cloud Sleuth properties to control Sleuth's sampling rate, etc. For example:
+
+[source,yaml]
+---
+spring.sleuth.sampler.percentage=.1                             # Send 10% of the request traces to Stackdriver
```
meltsufin: What is the default? When testing this out, the user may be confused why the traces don't show up in Stackdriver if the default is anything but `1.0`. Can we draw a bit more attention to this in the readme?

saturnism:
@meltsufin i always test locally :) it could be your gcloud's default project ID is going somewhere else - but i'm pretty positive local works :)

spring-cloud-gcp-starters/spring-cloud-gcp-starter-trace/README.adoc
saturnism: This is already called out in Sleuth's doc.
https://github.com/spring-cloud/spring-cloud-sleuth#features

We should probably just link it.

PR #141, opened 2017-09-06
Adds credential information for GCP compute environments

### LINKED ISSUE #130: README give better information
I see in the README:
"These properties are optional and, if not specified, Spring Boot will attempt to automatically find them for you"
I would like to see why I would need to modify the default value:
Maybe when I talk to a different project for resources?
Maybe I am not on GCP?
In the case of GAE Standard, or GCE? 

If the value are good as default, then it is confusing to put in a short README doc that you can change them when you do not have to...
KISS.

joaoandremartins:
Aims to ease the friction customers with local service accounts have
when deploying to Google App Engine.

Fixes #130

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -61,6 +61,6 @@ These properties are optional and, if not specified, Spring Boot will attempt to
 them for you. For details on how Spring Boot finds these properties, refer to the desired starter
 README.
 
-If your app is running on Google App Engine or Google Compute Engine, you should omit the
-`spring.cloud.gcp.credentials-location` property and, instead, let the Spring Cloud GCP Core Starter
-get the correct credentials for those environments.
+If your app is running on Google App Engine or Google Compute Engine, in most cases, you should omit
+the `spring.cloud.gcp.credentials-location` property and, instead, let the Spring Cloud GCP Core
+Starter get the correct credentials for those environments.
--- a/README.adoc
+++ b/README.adoc
@@ -60,3 +60,7 @@ spring.cloud.gcp.credentials-location=file:[LOCAL_PRIVATE_KEY_FILE]
 These properties are optional and, if not specified, Spring Boot will attempt to automatically find
 them for you. For details on how Spring Boot finds these properties, refer to the desired starter
 README.
+
+If your app is running on Google App Engine or Google Compute Engine, you should omit the
+`spring.cloud.gcp.credentials-location` property and, instead, let the Spring Cloud GCP Core Starter
+get the correct credentials for those environments.
--- a/spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/README.adoc
+++ b/spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/README.adoc
@@ -43,3 +43,13 @@ Similarly for `spring.cloud.gcp.credentials-location`, if a property is configur
 https://docs.spring.io/spring/docs/current/spring-framework-reference/html/resources.html#resources-implementations[file system paths, HTTP URLs, etc.]
 If no property is configured, the provided `CredentialsProvider` returns the
 http://google.github.io/google-auth-library-java/releases/0.7.1/apidocs/com/google/auth/oauth2/GoogleCredentials.html#getApplicationDefault()[application default credentials].
+
+If your app is running on Google App Engine or Google Compute Engine, you should omit the
+`spring.cloud.gcp.credentials-location` property and, instead, let the Spring Cloud GCP Core Starter
+get the correct credentials for those environments. On App Engine Standard, the
+https://cloud.google.com/appengine/docs/standard/java/appidentity/[App Identity service account credentials]
+are used, on App Engine Flexible, the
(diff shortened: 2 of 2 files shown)


joaoandremartins:
@saturnism @ludoch would be great what you guys think about this.

ludoch:
lgtm

README.adoc
```diff
@@ -60,3 +60,7 @@ spring.cloud.gcp.credentials-location=file:[LOCAL_PRIVATE_KEY_FILE]
 These properties are optional and, if not specified, Spring Boot will attempt to automatically find
 them for you. For details on how Spring Boot finds these properties, refer to the desired starter
 README.
+
+If your app is running on Google App Engine or Google Compute Engine, you should omit the
+`spring.cloud.gcp.credentials-location` property and, instead, let the Spring Cloud GCP Core Starter
```
artembilan: Why should we say "you should omit" meanwhile it would be better to ignore `spring.cloud.gcp.credentials-location` property in case of those environments (some `@Conditional` ?) and even WARN end-user that it does not make sense.

WDYT?

meltsufin:
Are we absolutely sure that providing credentials file location makes no sense when running on GAE or GCE? There might be some use-cases there. I think that just clarifying that `spring.cloud.gcp.credentials-location` is optional in all scenarios and what will be used if it's missing is sufficient.

README.adoc
joaoandremartins: When would you warn users? If you try to deploy an app with a credentials path on your local file system, the deployment just fails.

We might want to add default credentials on our App Engine conditional, for example. But we would be overriding a user property, which I'd rather not. And that wouldn't clear the need to document this, anyway.

joaoandremartins:
@meltsufin I prefer to document this behaviour. For GAE, I can't even think of a way to get external service account credentials to work (I can't load the credentials file from anywhere). On GCE, you can copy the private key to the local disk. But I'm not sure if this is the recommended approach. And in that specific case, users can still specify their own service account creds if they really want.

meltsufin:
@joaoandremartins Maybe we don't need `spring.cloud.gcp.credentials-location` at all then. Users always have the option of specifying the environment variable with credentials location, right?

joaoandremartins:
We need credentials location for any environment that's not GAE/GCE/GCP, or even local dev.

A property can come from an [environment variable](https://docs.spring.io/spring-boot/docs/current/reference/html/boot-features-external-config.html) with the same name. I prefer to use the Spring property since users have the option to specify that property on an `application.properties` file, which wouldn't happen using THE `GOOGLE_APPLICATION_CREDENTIALS` environment variable.

artembilan:
> I prefer to use the Spring property since users have the option to specify that property on an application.properties file

That's absolutely OK and really consistent and friendly. What is not that you are so strict with it in your doc. IMO we should just ignore it in those environments in favor of their mechanism and WARN that the `spring.cloud.gcp.credentials-location` property is ignored because of... That's all: no reason to fail fast in this case.

joaoandremartins:
The problem is that the properties are resolved before we can assert any conditionals. And when the properties resolve, an error will be thrown because the file couldn't be found in the file system. Or am I missing anything?

I don't like it when the program overrides users choices in the properties. It creates undesirable "magic" and a lot more overhead on our side to keep up with it (e.g., GAE standard/flex conditional, GCE conditional, etc.). I'd rather the program blows up and the user fixes the issue, so they're more aware of what's going on...

README.adoc
meltsufin: I would just say "If your app is running on Google App Engine or Google Compute Engine, in most cases, you will want to omit the `spring.cloud.gcp.credentials-location` property and rely on the built-in credentials."

spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/README.adoc
```diff
@@ -43,3 +43,13 @@ Similarly for `spring.cloud.gcp.credentials-location`, if a property is configur
 https://docs.spring.io/spring/docs/current/spring-framework-reference/html/resources.html#resources-implementations[file system paths, HTTP URLs, etc.]
 If no property is configured, the provided `CredentialsProvider` returns the
 http://google.github.io/google-auth-library-java/releases/0.7.1/apidocs/com/google/auth/oauth2/GoogleCredentials.html#getApplicationDefault()[application default credentials].
+
+If your app is running on Google App Engine or Google Compute Engine, you should omit the
```
meltsufin: Again,
"If your app is running on Google App Engine or Google Compute Engine, in most cases, you will want to omit the `spring.cloud.gcp.credentials-location` property, and instead, let the Spring Cloud GCP Core Starter get the built-in credentials in those environments."

joaoandremartins:
If need arises, we should change this.
I looked into leveraging Spring profiles, so that `application-dev.properties` would contain the local creds that wouldn't be uploaded running the `prod` profile, but from what I've read, I don't think there's a way it could work like this.

PR #145, opened 2017-09-12
Adds Spring Cloud Stream GCP Pub/Sub binder

### LINKED ISSUE #85: Implement Google Cloud Pub/Sub Spring Cloud Stream binder

joaoandremartins:
Fixes #85


pom.xml
```diff
@@ -31,6 +31,7 @@
 		<module>spring-cloud-gcp-starters</module>
 		<module>spring-integration-gcp</module>
 		<module>spring-cloud-gcp-docs</module>
+		<module>spring-cloud-gcp-stream</module>
```
sobychacko: if this is a  pubsub specific binder, why not reflect that in the artifact name? such as ` spring-cloud-stream-binder-pubsub` or similar? `spring-cloud-gcp-stream` doesn't give the impression that it is a binder implementation.

spring-cloud-gcp-dependencies/pom.xml
```diff
@@ -44,6 +45,17 @@
 				<version>${project.version}</version>
 			</dependency>
 			<dependency>
+				<groupId>org.springframework.cloud</groupId>
+				<artifactId>spring-cloud-gcp-stream-core</artifactId>
```
sobychacko: In other binders, for core modules, there is a standard naming convention we follow. for ex: `spring-cloud-stream-binder-core`. You may want to consider something similar for consistency reasons as this is a binder implementation.

spring-cloud-gcp-dependencies/pom.xml
```diff
@@ -65,6 +77,12 @@
 				<version>1.2</version>
 			</dependency>
 
+			<dependency>
+				<groupId>org.springframework.cloud</groupId>
+				<artifactId>spring-cloud-stream</artifactId>
+				<version>1.2.1.RELEASE</version>
```
sobychacko: Are you planning to target against this release for any reason? There is an RC version available for `1.3.0` and the GA is scheduled for the end of the month for that.

spring-cloud-gcp-stream/spring-cloud-gcp-stream-binder/src/main/java/org/springframework/cloud/gcp/stream/binder/pubsub/PubSubMessageChannelBinder.java
```diff
@@ -0,0 +1,87 @@
...
+	private PubSubExtendedBindingProperties pubSubExtendedBindingProperties =
+			new PubSubExtendedBindingProperties();
+
+	public PubSubMessageChannelBinder(boolean supportsHeadersNatively, String[] headersToEmbed,
+			PubSubTopicProvisioner provisioningProvider) {
+		super(supportsHeadersNatively, headersToEmbed, provisioningProvider);
+	}
+
+	@Override
+	protected MessageHandler createProducerMessageHandler(ProducerDestination destination,
```
Andy2003: It should be checked, if the topic exists and if not a topic should be created

spring-cloud-gcp-stream/spring-cloud-gcp-stream-binder/src/main/java/org/springframework/cloud/gcp/stream/binder/pubsub/PubSubMessageChannelBinder.java
```diff
@@ -0,0 +1,87 @@
...
+			ExtendedProducerProperties<PubSubProducerProperties> producerProperties)
+			throws Exception {
+		PubSubMessageHandler pubSubMessageHandler = new PubSubMessageHandler(this.pubSubTemplate);
+		pubSubMessageHandler.setTopic(destination.getName());
+
+		return pubSubMessageHandler;
+	}
+
+	@Override
+	protected MessageProducer createConsumerEndpoint(ConsumerDestination destination, String group,
```
Andy2003: Same here. A subscription should be created, if it does no already exists. I think the topic could optionally also be created (subject to configuration)

Andy2003: the group name should also be taken into consideration to make this binder work

spring-cloud-gcp-stream/spring-cloud-gcp-stream-core/src/main/java/org/springframework/cloud/gcp/stream/binder/pubsub/provisioning/PubSubProducerDestination.java
```diff
@@ -0,0 +1,38 @@
...
+	}
+
+	@Override
+	public String getName() {
+		return this.name;
+	}
+
+	@Override
+	public String getNameForPartition(int partition) {
+		return this.name;
```
Andy2003: Should partitioning be supported?

    @Override
	public String getNameForPartition(int partition) {
		return this.name + "-" + partition;
	}

spring-cloud-gcp-stream/spring-cloud-gcp-stream-core/src/main/java/org/springframework/cloud/gcp/stream/binder/pubsub/properties/PubSubConsumerProperties.java
```diff
@@ -0,0 +1,20 @@
...
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package org.springframework.cloud.gcp.stream.binder.pubsub.properties;
+
+public class PubSubConsumerProperties {
+}
```
Andy2003: It would be helpful to have a prefix for consumers and producers, which gets prepended to topic names

pom.xml
joaoandremartins: The `spring-cloud-gcp-stream` module contains two sub-modules: `spring-cloud-gcp-stream-binder` and `spring-cloud-gcp-stream-core`.

I agree this warrants a rename if we're planning other binders besides Pub/Sub. If Pub/Sub is going to be our only binder, then perhaps naming is good as-is?

spring-cloud-gcp-dependencies/pom.xml
joaoandremartins: Renamed the two sub-modules to `spring-cloud-stream-binder-gcp-pubsub` and `spring-cloud-stream-binder-gcp-pussub-core` to keep up naming consistency.

spring-cloud-gcp-dependencies/pom.xml
joaoandremartins: No special reason, we can change to `1.3.0` after it goes GA. I doubt this PR will be ready before end of the month.

joaoandremartins: Noticed some issues with RC1, will wait for the GA version before fully moving to `1.3.0`.

spring-cloud-gcp-stream/spring-cloud-gcp-stream-binder/src/main/java/org/springframework/cloud/gcp/stream/binder/pubsub/PubSubMessageChannelBinder.java
joaoandremartins: Did you mean `PubSubConsumerProperties`?

Andy2003: Yes, this would be a property of PubSubConsumerProperties

joaoandremartins:
A known issue with the latest commit is that some weird padding is being added to my message at the beginning.
e.g., I send the message "hoy hoy" and receive ""
If I enable `useNativeEncoding` on the producer, it works, but it should probably work without it too?

I'm also not sure if I'm handling groups well in `PubSubMessageChannelBinder#createConsumerEndpoint` and I also plan to add unit tests (although `PartitionCapableBinderTests` might not do since I don't think this binder will support partitioning on its first iteration).

I have also only tried sending String payloads.

Apart from that, this PR is ready for review.

spring-cloud-gcp-stream/spring-cloud-stream-binder-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/stream/binder/pubsub/PubSubMessageChannelBinder.java
```diff
@@ -0,0 +1,92 @@
...
+		PubSubProducerProperties> {
+
+	private PubSubTemplate pubSubTemplate;
+
+	private PubSubExtendedBindingProperties pubSubExtendedBindingProperties =
+			new PubSubExtendedBindingProperties();
+
+	public PubSubMessageChannelBinder(boolean supportsHeadersNatively, String[] headersToEmbed,
+			PubSubChannelProvisioner provisioningProvider, PubSubTemplate pubSubTemplate) {
+		super(supportsHeadersNatively, headersToEmbed, provisioningProvider);
```
Andy2003: concerning the `supportsHeadersNatively`: this setting should be `true` otherwise the tests will fail. This setting was also set to `true` in this project https://github.com/spring-cloud/spring-cloud-stream-binder-google-pubsub/pull/5

joaoandremartins:
The problem with the test is that, for some reason, the Pub/Sub emulator apparently doesn't like this:

`MessageBuilder.withPayload(testPayload2.getBytes()).build()` (the .getBytes() part)

In any case, I didn't realise you were referring to integration tests.
I think we should take care of those in #171

Andy2003:
It seems to be a problem within this project. With the project
https://github.com/spring-cloud/spring-cloud-stream-binder-google-pubsub/pull/5
I was able to run all tests successfully against the emulator (even the
partitioned ones). Unfortunately, I didn't have time to narrow down the
problem in this project.

Am 03.11.2017 22:21 schrieb "João Martins" <notifications@github.com>:

joaoandremartins:
I realise that in spring-cloud/spring-cloud-stream-binder-google-pubsub#5, you were using google-cloud-pubsub version `0.22.0-beta`. We're up to `0.26.0-beta` in master.

I can find no other explanation for the fact that when I send strings instead of bytes, the emulator doesn't error out and the test succeeds. I debugged the issue and that was my finding.

Since spinning up the emulator is probably going to require some instrumentation, I suggest we leave integration tests for #171

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/PubSubAdmin.java
```diff
@@ -95,6 +98,28 @@ public Topic createTopic(String topicName) {
...
+	 * @return topic configuration or {@code null} if topic doesn't exist
+	 */
+	public Topic getTopic(String topicName) {
+		Assert.hasText(topicName, "No topic name was specified.");
+
+		try {
+			return this.topicAdminClient.getTopic(TopicName.create(this.projectId, topicName));
+		}
+		catch (ApiException aex) {
+			if (aex.getStatusCode() instanceof GrpcStatusCode
```
joaoandremartins: gax version 1.11.0 allows us to deal with this more gracefully, but we need to wait for a google-cloud-pubsub release that uses that gax version.

artembilan:
I'm not so good in Spring Cloud Stream especially in the latest changes.
And also I have to prepare my talk for Spring One Platform.
So, I really don't know when I will be able to overview this.

Can't it wait really until after S1P conference?

Thanks for understanding

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/PubSubAdmin.java
Andy2003: I have just updated https://github.com/kiwigrid/spring-cloud-stream-binder-google-pubsub so that `google-cloud-pub:0.26.0-beta` is also used there. Even with this version all tests are still green! So there's a mistake somewhere in this project here. I've already done some research, but I haven't been able to find the problem yet.

Andy2003:
I have just updated https://github.com/kiwigrid/spring-cloud-stream-binder-google-pubsub so that google-`cloud-pub:0.26.0-beta` is also used there. Even with this version all tests are still green! So there's a mistake somewhere in this project here. I've already done some research, but I haven't been able to find the problem yet.

Of course, you can handle the tests with #171. However, as long as these are not green, I would not recommend using this binder.

joaoandremartins:
@artembilan can you recommend anyone else in Spring who is a Stream expert and could help here?

@Andy2003 the goal of this PR is to set the base code for the binder. It will naturally require some changes, but those can be made through other PRs (including the tests) so that we don't end up with a giant, unreviewable PR. I'm OK stating that the binder isn't good for consumption until the tests pass.

Speaking of the tests, try overriding the failing test with the same code, but remove the [`.getBytes()` call on the payload at the MessageBuilder time](https://github.com/spring-cloud/spring-cloud-stream/blob/master/spring-cloud-stream-binder-test/src/main/java/org/springframework/cloud/stream/binder/AbstractBinderTests.java#L177) and see if the test passes. Maybe that points to an issue on this project, but I haven't been able to make a correlation yet.

Andy2003:
@joaoandremartins If I make the modifications to the test that you mentioned, it will also fail in the other project.

sobychacko:
If native headers are supported, the HeaderMode needs to be set to `raw`. I think that is the reason for this issue: https://github.com/spring-cloud/spring-cloud-gcp/pull/145#issuecomment-341565542

joaoandremartins:
After I set `supportsHeadersNatively` in `PubSubMessageChannelBinder`, I stopped getting the added padding.

@Andy2003 can you clarify under which circumstances the tests fail on the other project?

joaoandremartins:
@garyrussell @artembilan ping!

Now that SpringOne is wrapped up, can we revisit this PR?
Also, feel free to add other pertinent reviewers for Spring Cloud Stream, if necessary.

joaoandremartins:
Sorry for the nagging -- any chance we can get a review here before the end of the week?

I just merged with master and verified that [my toy app](https://github.com/joaoandremartins/chatroom-stream) works.

Andy2003:
@joaoandremartins My comment https://github.com/spring-cloud/spring-cloud-gcp/pull/145#issuecomment-342423208 referred to your question in comment https://github.com/spring-cloud/spring-cloud-gcp/pull/145#issuecomment-342241706
I made these adjustments. Afterwards, the tests in the other project failed as well.

joaoandremartins:
@Andy2003 I suspect this could be an issue with the emulator, since when it happens, I can see an error logged in the emulator window.

In any case, I'm able to run the simple case I mentioned. I think it's OK punting this to when I add the integration tests and actually see this error.

spring-cloud-gcp-dependencies/pom.xml
```diff
@@ -19,6 +19,7 @@
 		<cloud-trace-java.version>0.5.0</cloud-trace-java.version>
 		<spring-cloud-context.version>1.2.3.RELEASE</spring-cloud-context.version>
 		<spring-cloud-sleuth.version>1.2.5.RELEASE</spring-cloud-sleuth.version>
+        <spring-cloud-stream.version>1.3.0.RELEASE</spring-cloud-stream.version>
```
artembilan: May someone here explain me why can't we upgrade to Spring Cloud/Spring Boot 2.0 yet?
Right, they are still under development, but we are too. And we really don't have plans for real GA here in the nearest future. That is absolutely fine to depend on those brand new versions and pull the best from all the Spring portfolio.
More over that would be great feedback from our side to their since we are all still develop something.

The situation with this Binder that Spring Cloud Stream 2.0 is slightly different than 1.3 so we again will have some work in the future when we upgrade anyway. Well, that's fine too if we release SC-GCP-2.0, but when it is going to happen? I think people will just lose an interest to our project if it isn't aligned with the current Spring IO state. No body will be interested to use Spring Boot 1.5 based project when Spring Boot 2.0 is here.

Just my five fingers why I don't review this PR.

Thanks for understanding.

/CC @markfisher , @garyrussell , @olegz

joaoandremartins: Of course we want to be on par with what's out there, but I personally prefer against doing the change in this PR, since it's dragged on for way too long now (e.g., months) and it will drag on even more.

That sounds very much like the kind of change that we can easily do outside of this PR (#201). In the meantime, we would also be seeing feedback from prospective users that we otherwise wouldn't if the PR isn't merged.

artembilan: My point of view if we review/merge this now, we will be forced to review/merge one more version when we upgrade to SCSt 2.0.

 The feature is still fresh and why don't do it only once?

meltsufin: I think we should carefully consider the decision to move (or branch) to Spring Boot 2.0 because it could affect many other parts of the project. We definitely want to tap into as many users as possible, and I think most of the users are on 1.5 right now. That being said, we'll need to move to 2.0 at some point, but I would hope that it wouldn't hold back this PR. I hope whatever changes are in SCSt 2.0, most of the code would still be the same and the effort spent reviewing this wouldn't be all wasted.

Let's discuss more on the next call.

artembilan: Well, IMO those users who can use our milestone may easily use Spring Boot 2.0 and so on.
More over most Spring project are already at GA for Spring Framework 5.0 base line and developers from there really encourage people to use the latest version for new projects.
I don't know why you try to stick with an old version for this new project when our real GA is far away the next year. That is absolutely fine to stay in line with modern trends.

OK. I don't have more arguments to convince you do not do the same work twice, but if you still insist in SF-4.3 base line for our version 1.0, let it.

I would be glad to hear other points of you of my team members, but as long as I am in development for AWS Kinesis Binder I can say that there are a lot of critical changes in SCSt 2.0 so we decided to move it directly to Spring Boot 2.0 to avoid extra work with the upcoming features.

joaoandremartins: I don't mind doing the switchover to SB2.0 right after this PR.

I would just rather do this work with other PRs than on this one. This one is as good as done anyway, as far as initial implementation goes.

artembilan: Correct. I don't request to do Boot 2.0 work here in this PR.
I suggest to do that in the separate isssue/PR.

Merge it and be scared how much work we have to do here for SCSt 2.0 compatibility yet. 😄 

That's just my feeling though: there are a lot of interesting tasks to do rather than always redo not released code yet, just because we have started not from proper version...

I think I have already exhausted you with this subject, so, I will start to do what is easier for you and what you will say is the best to go. I might just don see the whole picture with the market around and insist in the moving on according my experience in other projects.

joaoandremartins:
A friendly reminder that this PR still needs an approval :)

meltsufin:
@joaoandremartins Can you please fix the build? There is a lot of stuff that is still missing here like docs and samples. I'm inclined to just merge this once the build is fixed and iterate on it.
However, please create issues for things that still need to be done to consider the binder completed.
Thanks!

joaoandremartins:
Done! #292 #293

PR #148, opened 2017-09-13
Adds subscribe() to PubSubOperations

joaoandremartins:
Replaces references to PubSubTemplate by PubSubOperations. Adds
JavaDoc to PubSubOperations.

Fixes bits of #92

### DIFF
--- a/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubOperations.java
+++ b/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubOperations.java
@@ -16,14 +16,39 @@
 
 package org.springframework.cloud.gcp.pubsub.core;
 
+import com.google.cloud.pubsub.v1.MessageReceiver;
+import com.google.cloud.pubsub.v1.Subscriber;
+
 import org.springframework.messaging.Message;
 import org.springframework.util.concurrent.ListenableFuture;
 
 /**
+ * An abstraction for Google Cloud Pub/Sub.
+ *
  * @author Vinicius Carvalho
+ * @author João André Martins
  */
 public interface PubSubOperations {
 
-	ListenableFuture<String> send(String topic, Message message);
+	/**
+	 * Sends a Spring {@link Message} to Pub/Sub.
+	 *
+	 * @param topic the name of an existing topic
+	 * @param message the Spring message
+	 * @return the listenable future of the call
+	 */
+	ListenableFuture<String> publish(String topic, Message message);
+
+	/**
+	 * Adds a callback method to an existing subscription.
+	 *
+	 * <p>
+	 * The created {@link Subscriber} is returned so it can be stopped.
+	 *
+	 * @param subscription the name of an existing subscription
+	 * @param messageHandler the callback method triggered when new messages arrive
+	 * @return subscriber listening to new messages
+	 */
+	Subscriber subscribe(String subscription, MessageReceiver messageHandler);
 
 }
--- a/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
+++ b/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
@@ -19,11 +19,14 @@ package org.springframework.cloud.gcp.pubsub.core;
 import com.google.api.core.ApiFuture;
 import com.google.api.core.ApiFutureCallback;
 import com.google.api.core.ApiFutures;
+import com.google.cloud.pubsub.v1.MessageReceiver;
+import com.google.cloud.pubsub.v1.Subscriber;
 import com.google.pubsub.v1.PubsubMessage;
 
 import org.springframework.beans.factory.InitializingBean;
 import org.springframework.cloud.gcp.pubsub.converters.SimpleMessageConverter;
 import org.springframework.cloud.gcp.pubsub.support.PublisherFactory;
+import org.springframework.cloud.gcp.pubsub.support.SubscriberFactory;
 import org.springframework.messaging.Message;
 import org.springframework.messaging.converter.MessageConversionException;
 import org.springframework.messaging.converter.MessageConverter;
@@ -41,12 +44,15 @@ public class PubSubTemplate implements PubSubOperations, InitializingBean {
 
 	private final PublisherFactory publisherFactory;
 
-	public PubSubTemplate(PublisherFactory publisherFactory) {
+	private final SubscriberFactory subscriberFactory;
+
+	public PubSubTemplate(PublisherFactory publisherFactory, SubscriberFactory subscriberFactory) {
 		this.publisherFactory = publisherFactory;
+		this.subscriberFactory = subscriberFactory;
 	}
 
 	@Override
-	public ListenableFuture<String> send(final String topic, Message message) {
+	public ListenableFuture<String> publish(final String topic, Message message) {
 		// Convert from payload into PubsubMessage.
 		Object pubsubMessageObject =
 				this.messageConverter.fromMessage(message, PubsubMessage.class);
@@ -76,6 +82,13 @@ public class PubSubTemplate implements PubSubOperations, InitializingBean {
 		});
 
 		return settableFuture;
+	}
+
+	@Override
+	public Subscriber subscribe(String subscription, MessageReceiver messageHandler) {
+		Subscriber subscriber = this.subscriberFactory.getSubscriber(subscription, messageHandler);
+		subscriber.startAsync();
+		return subscriber;
 	}
 
 	@Override
--- a/spring-cloud-gcp-pubsub/src/test/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplateTests.java
+++ b/spring-cloud-gcp-pubsub/src/test/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplateTests.java
@@ -20,8 +20,11 @@ import java.util.HashMap;
 import java.util.Map;
 import java.util.concurrent.ExecutionException;
 
+import com.google.api.core.ApiService;
 import com.google.api.core.SettableApiFuture;
+import com.google.cloud.pubsub.v1.MessageReceiver;
 import com.google.cloud.pubsub.v1.Publisher;
+import com.google.cloud.pubsub.v1.Subscriber;
 import com.google.pubsub.v1.PubsubMessage;
 import org.junit.Before;
 import org.junit.Test;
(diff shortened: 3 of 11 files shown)


spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
```diff
@@ -79,6 +85,13 @@ public void onSuccess(String result) {
 	}
 
 	@Override
+	public Subscriber poll(String subscription, MessageReceiver messageHandler) {
+		Subscriber subscriber = this.subscriberFactory.getSubscriber(subscription, messageHandler);
+		subscriber.startAsync();
+		return subscriber;
```
artembilan: I somehow don't see a reason to have this functionality here. It just doesn't bring too match value and just fully delegates to the `subscriberFactory`. 
What is the point to overhead with this, please?

saturnism: In my mind, the PubSubOperation should be the one stop surface for commonly used Pub/Sub interactions, which is to publish, pull, subscribe. SubscriberFactory is a bit of a utility that helps to do it,  the basic functionality should be in the PubSubOperation rather than having to deal w/ SubscriberFactory (or other things).

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubOperations.java
```diff
@@ -16,14 +16,39 @@
...
+	 * Polls a subscription for new messages.
+	 *
+	 * <p>
+	 * The created {@link Subscriber} is returned so it can be stopped.
+	 *
+	 * @param subscription the name of an existing subscription
+	 * @param messageHandler the callback method activated whenever a new message arrives
+	 * @return subscriber listening to new messages
+	 */
+	Subscriber poll(String subscription, MessageReceiver messageHandler);
```
saturnism: I'd recommend calling this `subscribe`.  Poll indicates that we are pulling/polling for new messages. But that may not be true.

joaoandremartins: I had `subscribe` at the beginning, but then changed.
`subscribe` isn't the right term, because you subscribe to a topic and the argument here is `subscription`.
How would polling not happen here? We start and return a subscriber, it's bound to poll. New messages may or may not arrive, but polling is happening regardless.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
joaoandremartins: I tend to agree with @saturnism .
Using `PubSubOperations`/`PubSubTemplate` as a standalone abstraction, it isn't possible to pull messages currently.
Users have to directly use `SubscriberFactory`, which makes `PubSubOperations` incomplete as a Pub/Sub surface. I consider `SubscriberFactory` to be an implementation detail.

artembilan: Yeah. Agreed. Much easier to inject only one abstraction. Thank you and never mind!

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubOperations.java
artembilan: Well, `poll` in most cases mean one time operation. See Apache Kafka and Spring Integration Polling Channel Adapter. 
`subscribe` is definitely good to determine that we are about long-life repeatable process. I can suggest a `consume` term. 
However the `MessageReceiver` argument may definitely be our saver and together with a good JavaDoc we are good to go with out `poll` variant.
WDYT?

markfisher: From the perspective of the `MessageHandler`, it's more like a subscribe (polling is an implementation detail). It's arguably the parameter that this method is most focused on since the subscription String is just a name. So I agree with @saturnism FWIW.

joaoandremartins: If the subscriber is a push subscriber, then you're right that polling doesn't happen, so this should definitely be renamed. (It's the only time polling doesn't happen, as our Pub/Sub Java API only supports polling/async pulling AFAIK)

I still have some reservations about `subscribe`, though. Conceptually, you subscribe to a topic, whereas here we're subscribing to a subscription which doesn't make much sense?

`subscribe` would be fine if the argument was `topic` instead of `subscription` and `PubSubTemplate` created a subscription for that topic. However, this is a little out of the scope of this PR. I'll reticently go for `subscribe` due to public demand.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubOperations.java
```diff
@@ -16,14 +16,39 @@
...
+	 * Adds a callback method to an existing subscription.
+	 *
+	 * <p>
+	 * The created {@link Subscriber} is returned so it can be stopped.
+	 *
+	 * @param subscription the name of an existing subscription
+	 * @param messageHandler the callback method triggered when new messages arrive
+	 * @return subscriber listening to new messages
+	 */
+	Subscriber subscribe(String subscription, MessageReceiver messageHandler);
```
joaoandremartins: I'd like to make a suggestion here: instead of a `MessageReceiver`, how about we provide our own callback format?

The reason is that `MessageReceiver` makes users handle a `PubsubMessage`, which is undesirable I think.

In `PubSubTemplate`, we could translate from a `PubsubMessage` to a `Message` and let then call our own callback on that.

Thoughts before I jump into implementing this? cc @saturnism

meltsufin: Agreed, either that, or change the `publish` method to accept a `PubsubMessage` as well, instead of `Message` to keep the API consistent.

markfisher: does it make sense to just use the `org.springframework.messaging.MessageHandler` then?

artembilan: If you would like to go this way, I would suggest to have it as variant. Don't forget that we have `PubSubInboundChannelAdapter`, which looks like use this `PubSubOperations` and that would be much better to let it to deal with the low level protocol directly: there are many angles to be implemented exactly in the Channel Adapter, like headers mapping and some additional headers indicating that we are exactly from the Channel Adapter.

I somehow last time don't have strong opinion about APIs: if people made their vote then they had had some reason and if it really makes some sense I don't mind. Like I said about this `poll()` before. Although IMO this subscribe and process is not an operation of the template: it really must be stateless. That's why `RabbitTemplate` and `JmsTemplate` don't have long-living consuming operations. That's already a `MessageListenerContainer` responsibility. But you see: I still don't mind. Just because those two pretty old (but stable) API, and here we introduce a new modern API, so I'm fully open to give it a shot and let people to deal with callback API from our template. But on the other hand I would keep in mind `PubSubInboundChannelAdapter` and wouldn't try to eliminate its possible flexibility.

So, feel free to mix Messaging API in this template (although in other places we have it as a separate one - `RabbitMessagingTemplate`), but, please, keep the raw API to be able to manipulate with it from other places.

spring-cloud-gcp-pubsub/src/test/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplateTests.java
```diff
@@ -126,4 +148,12 @@ public void testSend_onFailure() {
 	public void testSetNullMessageConverter() {
 		this.pubSubTemplate.setMessageConverter(null);
 	}
+
+	@Test
+	public void testPoll() {
```
artembilan: `testSubscribe()` then? 😄

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubOperations.java
joaoandremartins: @markfisher that `MessageHandler` seems just like what I was thinking. Thanks!

@artembilan the main reason I wanted to do this is exactly message conversion, and I came across this issue exactly on `PubSubInboundChannelAdapter`! The issue is `PubSubTemplate` contains the `MessageConverter` to turn `Messages` into payload and vice-versa. By giving a `PubsubMessage` to `PubSubInboundChannelAdapter`, we make it also have to know about message converters, which is probably avoidable.

I'll keep your suggestion in mind and keep the low-level trigger. We can revisit this later if needed be.

joaoandremartins: I realise that Messaging is only really needed for the channel adapters, so we might want to keep `PubSubOperations` Spring Messaging-agnostic, or provide utility Messaging methods like suggested.
I guess the issue I should be trying to tackle in `PubSubTemplate` is payload conversion (Object2Jackson/Gson and vice-versa). We might even be able to move the `MessageConverter` out of `PubSubTemplate` and into the adapters.

joaoandremartins:
I believe all comments here were addressed. Let me know if I missed anything.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubOperations.java
```diff
@@ -16,14 +16,39 @@
...
 
-	ListenableFuture<String> send(String topic, Message message);
+	/**
+	 * Sends a Spring {@link Message} to Pub/Sub.
+	 *
+	 * @param topic the name of an existing topic
+	 * @param message the Spring message
+	 * @return the listenable future of the call
+	 */
+	ListenableFuture<String> publish(String topic, Message message);
```
meltsufin: As I commented before, I think this method should take `PubsubMessage` and keep the messaging APIs out of `PubSubOperations`/`PubSubTemplate`. This would make it consistent with `JmsTemplate` and `RabbitTemplate`. If we think that a `PubSubMessagingTemplate` is useful for channel adapters, we can create that as well.

joaoandremartins: This PR concerns only adding subscribe() to `PubSubOperations`.
If you don't mind, I'd prefer filing a new issue to add `PubsubMessage` payload support, or even address it in an open PR like #151, although I think we need to discuss before, as I'm not sure if surfacing `PubsubMessage` is useful, or a good idea.

PR #151, opened 2017-09-15
Moves message conversion out of PubSubTemplate to SI adapters

### LINKED ISSUE #54: Use CompositeMessageConverter in PubSubTemplate

joaoandremartins:
Improves payload and message conversion logic. Channel adapter users
will no longer get a ByteString (from Google Cloud Pub/Sub API) payload
but a String one.

Fixes #54 and bits of #92


joaoandremartins:
What do you guys think about this? We can also add a `publish(String topic, Message message)` to `PubSubOperations` if there are Messaging clients other than SI.

This is a proposal to separate concerns a little bit. Next step is to add a `publish(Object)` method to `PubSubOperations` that uses Jackson to serialize/deserialize objects. I was playing around with this today and found that it isn't very straightforward.

A benefit of this PR is the separation of concerns between Messaging and core Pub/Sub abstraction, and the payload in the `Messages` from the channel adapters are a String, instead of a ByteString, so no casts or knowledge of the Google Cloud Pub/Sub API are needed by the user.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubOperations.java
```diff
@@ -16,14 +16,16 @@
...
 import org.springframework.util.concurrent.ListenableFuture;
 
 /**
  * @author Vinicius Carvalho
+ * @author João André Martins
  */
 public interface PubSubOperations {
 
-	ListenableFuture<String> send(String topic, Message message);
+	ListenableFuture<String> publish(String topic, String payload, Map<String, String> headers);
```
meltsufin: This is interesting. In the other PR you have open, I was suggesting using `PubsubMessage`, but maybe we can have both.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
```diff
@@ -16,55 +16,63 @@
...
+	@Override
+	public ListenableFuture<String> publish(final String topic, byte[] payload,
+			Map<String, String> headers) {
+		return publish(topic, buildPubsubMessage(payload, headers));
+	}
 
-		// Send message to Google Cloud Pub/Sub.
-		ApiFuture<String> publishFuture = this.publisherFactory.getPublisher(topic)
-				.publish((PubsubMessage) pubsubMessageObject);
+	public ListenableFuture<String> publish(final String topic, PubsubMessage pubsubMessage) {
```
joaoandremartins: Made visibility of this `public`, so users have access to the more low-level method, but didn't add it to `PubSubOperations`. LMKWYT

spring-integration-gcp/src/main/java/org/springframework/integration/gcp/outbound/PubSubMessageHandler.java
```diff
@@ -29,15 +37,28 @@
 
 	private final PubSubOperations pubSubTemplate;
 
+	private MessageConverter messageConverter;
```
meltsufin: I think we would want the message converter to be configurable from the outside.

spring-integration-gcp/src/main/java/org/springframework/integration/gcp/outbound/PubSubMessageHandler.java
```diff
@@ -29,15 +37,28 @@
...
+		stringMessageConverter.setSerializedPayloadClass(String.class);
+		this.messageConverter =
+				new CompositeMessageConverter(ImmutableList.of(stringMessageConverter));
 	}
 
 	@Override
-	protected void handleMessageInternal(Message<?> message) throws Exception {
-		this.pubSubTemplate.publish(this.topic, message);
+	protected void handleMessageInternal(Message<?> message) {
+		String payload = (String) this.messageConverter.fromMessage(message, String.class);
```
meltsufin: Pub/Sub represents the message payload as `ByteString`. So, I think we should be converting directly to `ByteString` here instead of `String`.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
artembilan: I don't see a big problem to let end-users to build `PubsubMessage` by themselves and publish them as is.
So, +1 to expose this API on the interface level. See `RabbitTemplate.send(Message)`. 
Although all those non-`Message` methods are called `convertAndSend`.
We may consider the same here, however I'm fine with just overloaded variants of the `publish()`

spring-integration-gcp/src/main/java/org/springframework/integration/gcp/inbound/PubSubInboundChannelAdapter.java
```diff
@@ -50,6 +53,11 @@ public PubSubInboundChannelAdapter(PubSubOperations pubSubTemplate,
 			String subscriptionName) {
 		this.pubSubTemplate = pubSubTemplate;
 		this.subscriptionName = subscriptionName;
+
+		StringMessageConverter stringMessageConverter = new StringMessageConverter();
+		stringMessageConverter.setSerializedPayloadClass(String.class);
+		getMessagingTemplate().setMessageConverter(
+				new CompositeMessageConverter(ImmutableList.of(stringMessageConverter)));
```
artembilan: We don't use Guava 😉 

On the other hand your immutability doesn't have effect. See that `CompositeMessageConverter` ctor:
```
public CompositeMessageConverter(Collection<MessageConverter> converters) {
		Assert.notEmpty(converters, "Converters must not be empty");
		this.converters = new ArrayList<>(converters);
	}
```
There is just local copy of your provided list.

OTOH why don't you allow end-user to inject `MessageConverter` at all?

spring-integration-gcp/src/main/java/org/springframework/integration/gcp/inbound/PubSubInboundChannelAdapter.java
```diff
@@ -70,8 +78,9 @@ private void receiveMessage(PubsubMessage message, AckReplyConsumer consumer) {
 		}
 
 		try {
+			String payload = message.getData().toStringUtf8();
```
artembilan: Isn't this related to the `Charset` in the `publish()` above?
What will happen here if I didn't use `UTF-8` there though?

spring-integration-gcp/src/main/java/org/springframework/integration/gcp/outbound/PubSubMessageHandler.java
artembilan: 👍

spring-integration-gcp/src/main/java/org/springframework/integration/gcp/outbound/PubSubMessageHandler.java
```diff
@@ -29,15 +37,28 @@
...
+	private MessageConverter messageConverter;
+
 	private String topic;
 
 	public PubSubMessageHandler(PubSubOperations pubSubTemplate, String topic) {
 		this.pubSubTemplate = pubSubTemplate;
 		this.topic = topic;
+
+		StringMessageConverter stringMessageConverter = new StringMessageConverter();
+		stringMessageConverter.setSerializedPayloadClass(String.class);
```
artembilan: This property isn't use during conversion from `Message`

spring-integration-gcp/src/main/java/org/springframework/integration/gcp/outbound/PubSubMessageHandler.java
artembilan: In addition I would like to allow to send messages to this `PubSubMessageHandler` with the payload as `PubsubMessage`, `ByteString`, `byte[]` bypassing any conversion at all.

spring-integration-gcp/src/main/java/org/springframework/integration/gcp/outbound/PubSubMessageHandler.java
joaoandremartins: But we don't use message converter for anything here.
What do you want to use it for?

spring-integration-gcp/src/main/java/org/springframework/integration/gcp/outbound/PubSubMessageHandler.java
joaoandremartins: @artembilan you already have `PubSubTemplate` for that. Why should that be in the `PubSubMessageHandler`? `AbstractMessageHandler` only declares `handleMessageInternal(Message)`.

spring-integration-gcp/src/main/java/org/springframework/integration/gcp/outbound/PubSubMessageHandler.java
artembilan: ??? Really? Why have you added it here then at all?
I see below the code like:
```
this.messageConverter.fromMessage()
```
So, looks like you use it here.
Gotcha?

The idea is to let end user to inject externally configured `MessageConverter`. We need setter for this property. That's all.

spring-integration-gcp/src/main/java/org/springframework/integration/gcp/inbound/PubSubInboundChannelAdapter.java
joaoandremartins: I'm not using ImmutableMap so much because of the immutability, but more for syntactic sugar (inline map declaration and construction).
Although I can always declare the map separately.

spring-integration-gcp/src/main/java/org/springframework/integration/gcp/outbound/PubSubMessageHandler.java
joaoandremartins: Woops, I'm sorry, I thought this was `PubSubTemplate`.
Yes of course, we totally need to allow users to specify a converter!

spring-integration-gcp/src/main/java/org/springframework/integration/gcp/outbound/PubSubMessageHandler.java
artembilan: Well, I have an Integration flow. All the components are connected via message channels which are agnostic of the `payload` of the message they transport. And you can't predict what upstream logic I have. So, why would someone force me to switch to the `PubSubTemplate` if I would like to send message to the channel and don't care. The target `MH` must be smart and let me to send different `payload`s.

I don't talk about `Message`, @joaoandremartins , I talk about its `payload`. 
That is one of the gain of Spring Integration - let end-user do his logic in his own domain and be as much flexible for his `payload` as possible.

I feel that you are mixing somehow purpose of Channel Adapter and Template. That's why I suggested in the beginning do not look to the Template abstraction and just concentrate on the comprehensive Channel Adapters implementations.

Sorry if I'm confusing you but right now it is somehow in a mess.

spring-integration-gcp/src/main/java/org/springframework/integration/gcp/inbound/PubSubInboundChannelAdapter.java
joaoandremartins: That's a great point, I meant to get back to this.
The only way of not doing this that I can think of is to add a header to `PubsubMessage` saying which encoding the payload is in, assuming it is a `String`.
We might even be able to add another `PubsubMessage` header with the class of the payload and that might allow us to do the Object -> JSON -> Object conversion, to support Object payloads.

Without any context about the payload, we have to make an assumption... WDYT?

spring-integration-gcp/src/main/java/org/springframework/integration/gcp/outbound/PubSubMessageHandler.java
joaoandremartins: Apologies, I didn't understand what you meant at first.
Yes, I agree we must be able to send `Messages` with payloads other than String.

spring-integration-gcp/src/main/java/org/springframework/integration/gcp/inbound/PubSubInboundChannelAdapter.java
artembilan: Well, all works for me, but we might need to follow best practice here with Google Pub/Sub.
Right, now let's stick with default `UTF-8` and reconsider content-type negotiation in the separate issue. 
Does it make sense ?

spring-integration-gcp/src/main/java/org/springframework/integration/gcp/outbound/PubSubMessageHandler.java
joaoandremartins: Added a message conversion setter.

spring-integration-gcp/src/main/java/org/springframework/integration/gcp/outbound/PubSubMessageHandler.java
```diff
@@ -53,12 +57,44 @@ public PubSubMessageHandler(PubSubOperations pubSubTemplate, String topic) {
...
+		}
+		else if (payload instanceof byte[]) {
+			pubsubPayload = ByteString.copyFrom((byte[]) payload);
+		}
+		else if (payload instanceof ByteString) {
+			pubsubPayload = (ByteString) payload;
+		}
+		else {
+			throw new PubSubException("Payload type "
+					+ payload.getClass().getName() + " is unsupported.");
```
artembilan: Would be great to say *what* is supported.
By the way why can't we try to convert from any `Object` to `String` using mentioned `messageConverter`.
Isn't that its purpose and general idea of this PR?
Thanks

spring-integration-gcp/src/main/java/org/springframework/integration/gcp/outbound/PubSubMessageHandler.java
joaoandremartins: Right, I'm using it here just so `toMessage()` adds `String` payload, not `byte[]`.

To make simple code like this possible:

```
  @Bean
  @ServiceActivator(inputChannel = "pubsubInputChannel")
  public MessageHandler messageReceiver() {
    return message -> {
      LOGGER.info("Message arrived here! Payload: " + message.getPayload());
    };
  }
```

spring-integration-gcp/src/main/java/org/springframework/integration/gcp/outbound/PubSubMessageHandler.java
```diff
@@ -29,15 +41,60 @@
...
+		Object payload = message.getPayload();
+
+		if (payload instanceof PubsubMessage) {
+			this.pubSubTemplate.publish(this.topic, (PubsubMessage) payload);
+			return;
+		}
+
+		ByteString pubsubPayload;
+
+		if (payload instanceof String) {
```
joaoandremartins: @artembilan what do you think about this?

I looked at other integration outbound adapters, but they all followed a different approach.

spring-integration-gcp/src/main/java/org/springframework/integration/gcp/outbound/PubSubMessageHandler.java
artembilan: ??? We talk here about `PubSubMessageHandler`, not `PubSubInboundChannelAdapter`

spring-integration-gcp/src/main/java/org/springframework/integration/gcp/outbound/PubSubMessageHandler.java
artembilan: See my comment on the exception: https://github.com/spring-cloud/spring-cloud-gcp/pull/151#discussion_r140080119

spring-integration-gcp/src/main/java/org/springframework/integration/gcp/outbound/PubSubMessageHandler.java
artembilan: BTW, that is not good exception here from the `MessageHandler`. There is nothing about PubSub yet.
A `MessageHandlingException` is the right candidate.

joaoandremartins: Ended up getting rid of the exception since we can transform any byte array into a string..

spring-integration-gcp/src/main/java/org/springframework/integration/gcp/inbound/PubSubInboundChannelAdapter.java
```diff
@@ -56,8 +58,10 @@ public PubSubInboundChannelAdapter(PubSubOperations pubSubTemplate,
 
 		StringMessageConverter stringMessageConverter = new StringMessageConverter();
 		stringMessageConverter.setSerializedPayloadClass(String.class);
+		List<MessageConverter> messageConverters = new ArrayList<>();
+		messageConverters.add(stringMessageConverter);
 		getMessagingTemplate().setMessageConverter(
-				new CompositeMessageConverter(ImmutableList.of(stringMessageConverter)));
+				new CompositeMessageConverter(messageConverters));
```
artembilan: Why do you wrap to `CompositeMessageConverter` at all if we have only one `StringMessageConverter` in the list?

spring-integration-gcp/src/main/java/org/springframework/integration/gcp/inbound/PubSubInboundChannelAdapter.java
```diff
@@ -112,4 +116,12 @@ public void setAckMode(AckMode ackMode) {
 		Assert.notNull(ackMode, "The acknowledgement mode can't be null.");
 		this.ackMode = ackMode;
 	}
+
+	public void setMessageConverter(MessageConverter messageConverter) {
+		getMessagingTemplate().setMessageConverter(messageConverter);
+	}
+
+	public MessageConverter getMessageConverter() {
+		return getMessagingTemplate().getMessageConverter();
```
artembilan: I think we don't need extra delegation to the `MessagingTemplate` since we don't use its goodies afterward.
Right we use its `send(Message)` lately, but definitely not conversion already.
I mean that it would be much clear to have a local `MessageConverter` property.

Just IMO and I can live if you decide to stay with this style

spring-integration-gcp/src/main/java/org/springframework/integration/gcp/outbound/PubSubMessageHandler.java
```diff
@@ -48,11 +47,8 @@
 	public PubSubMessageHandler(PubSubOperations pubSubTemplate, String topic) {
 		this.pubSubTemplate = pubSubTemplate;
 		this.topic = topic;
-
-		StringMessageConverter stringMessageConverter = new StringMessageConverter();
-		stringMessageConverter.setSerializedPayloadClass(String.class);
-		this.messageConverter =
-				new CompositeMessageConverter(ImmutableList.of(stringMessageConverter));
+		this.messageConverter = new CompositeMessageConverter(
+				Collections.singletonList(new StringMessageConverter()));
```
artembilan: One more redundant wrapping to the `CompositeMessageConverter`

spring-integration-gcp/src/main/java/org/springframework/integration/gcp/inbound/PubSubInboundChannelAdapter.java
joaoandremartins: I was thinking we might want to add more in the future, but this is the classic case of YAGNI.
If we need more converters in the future, we can add the `CompositeMessageConverter` at that time.

artembilan: I think we don't. 
If end-user needs, he can simply inject this `CompositeMessageConverter` thanks to our setter on the matter.

joaoandremartins:
I think I covered every comment, but let me know if I missed anything.
It's hard to mock our `Subscriber`, so I'm going to defer further `PubSubInboundChannelAdapter` tests to when we have some integration tests.

spring-integration-gcp/src/main/java/org/springframework/integration/gcp/inbound/PubSubInboundChannelAdapter.java
```diff
@@ -102,4 +111,12 @@ public void setAckMode(AckMode ackMode) {
 		Assert.notNull(ackMode, "The acknowledgement mode can't be null.");
 		this.ackMode = ackMode;
 	}
+
+	public void setMessageConverter(MessageConverter messageConverter) {
+		this.messageConverter = messageConverter;
```
artembilan: `Assert.notNull(messageConverter)`?

spring-integration-gcp/src/main/java/org/springframework/integration/gcp/outbound/PubSubMessageHandler.java
```diff
@@ -29,15 +40,52 @@
...
 
+	private MessageConverter messageConverter;
+
 	private String topic;
 
 	public PubSubMessageHandler(PubSubOperations pubSubTemplate, String topic) {
 		this.pubSubTemplate = pubSubTemplate;
 		this.topic = topic;
+		this.messageConverter = new CompositeMessageConverter(
+				Collections.singletonList(new StringMessageConverter()));
```
artembilan: Why do we use `CompositeMessageConverter` here yet?

spring-integration-gcp/src/main/java/org/springframework/integration/gcp/outbound/PubSubMessageHandler.java
```diff
@@ -29,15 +40,52 @@
...
+
+		this.pubSubTemplate.publish(this.topic, pubsubPayload, headers);
+	}
+
+	public MessageConverter getMessageConverter() {
+		return this.messageConverter;
+	}
+
+	public void setMessageConverter(MessageConverter messageConverter) {
+		this.messageConverter = messageConverter;
```
artembilan: `Assert.notNull(messageConverter)` ?

PR #160, opened 2017-09-20
Documents recent undocumented modifications

### LINKED ISSUE #47: Generate meta-data for the @ConfigurationProperties

joaoandremartins:
Keeps the reference document up to date with previous changes to the API.

Incidentally, fixes #47

### DIFF
--- a/spring-cloud-gcp-docs/src/main/asciidoc/pubsub.adoc
+++ b/spring-cloud-gcp-docs/src/main/asciidoc/pubsub.adoc
@@ -28,7 +28,7 @@ dependencies {
 
 `PubSubOperations` is an abstraction that allows Spring users to use Google Cloud Pub/Sub without
 depending on any Google Cloud Pub/Sub API semantics.
-It comprehends all the possible operations to be carried by a Google Cloud Pub/Sub client.
+It provides the common set of operations needed to interact with Google Cloud Pub/Sub.
 `PubSubTemplate` is the default implementation of `PubSubOperations` and it uses the
 https://github.com/GoogleCloudPlatform/google-cloud-java/tree/master/google-cloud-pubsub[Google Cloud Java Client for Pub/Sub]
 to interact with Google Cloud Pub/Sub.
@@ -59,45 +59,22 @@ The default value for both these properties is 4 threads.
 ==== Publishing to a topic
 
 `PubSubTemplate` provides asynchronous methods to publish messages to a Google Cloud Pub/Sub topic.
+It supports different types of payloads, including `Strings` with different encodings, `byte[]`,
+`ByteString` and `PubsubMessage`.
 
 [source,java]
 ----
 public ListenableFuture<String> publish(String topic, String payload, Map<String, String> headers)
-----
 
-Publishes a message containing the `payload` and `headers` to the target `topic`.
-The `payload` is converted into a `byte[]` using `Charset.defaultCharset()`.
-
-[source,java]
-----
 public ListenableFuture<String> publish(String topic, String payload, Map<String, String> headers,
 Charset charset)
-----
 
-Publishes a message containing the `payload` and `headers` to the target `topic`.
-The `payload` is converted into a `byte[]` using the specified `charset`.
-
-[source,java]
-----
 public ListenableFuture<String> publish(String topic, byte[] payload, Map<String, String> headers)
-----
 
-Publishes a message containing the `payload` and `headers` to the target `topic`.
-
-[source,java]
-----
 public ListenableFuture<String> publish(String topic, ByteString payload, Map<String, String> headers)
-----
 
-Publishes a message containing the `payload` and `headers` to the target `topic`.
-The `payload` is in Google Cloud Pub/Sub Client Library's native payload format, `ByteString`.
-
-[source,java]
-----
 public ListenableFuture<String> publish(String topic, PubsubMessage pubsubMessage)
 ----
-
-Low level method to publish a Google Cloud Pub/Sub `PubSubMessage` to the target `topic`.
 
 Here is an example of how to publish a message to a Google Cloud Pub/Sub topic:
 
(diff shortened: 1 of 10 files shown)


spring-cloud-gcp-docs/src/main/asciidoc/pubsub.adoc
```diff
@@ -51,39 +50,56 @@ In the example, we're setting the number of threads used by both the `Publisher`
...
-`Message` payload into the `PubsubMessage` data.
-`SimpleMessageConverter` calls `toString()` in both the `Message` headers and payload, since
-`PubsubMessage` only supports `String` attributes and data at the transmission level.
-We're considering JSON object encoding/decoding in the future to support more object types.
+Publishes a message containing the `payload` and `headers` to the target `topic`.
+The `payload` is converted into a `byte[]` using `Charset.defaultCharset()`.
+
+[source,java]
+----
+public ListenableFuture<String> publish(String topic, String payload, Map<String, String> headers,
```
joaoandremartins: Pending further advancements in #151

spring-cloud-gcp-docs/src/main/asciidoc/pubsub.adoc
```diff
@@ -26,13 +26,12 @@ dependencies {
 
 === Pub/Sub operations abstraction
 
-`PubSubTemplate` is an abstraction that allows Spring users to use Google Cloud Pub/Sub without
+`PubSubOperations` is an abstraction that allows Spring users to use Google Cloud Pub/Sub without
```
meltsufin: I don't mind actually calling it `PubSubTemplate`, even though interface is `PubSubOperations`. I think "template" is a more familiar term for Spring users.

spring-cloud-gcp-docs/src/main/asciidoc/pubsub.adoc
```diff
@@ -26,13 +26,12 @@ dependencies {
...
 depending on any Google Cloud Pub/Sub API semantics.
-`PubSubTemplate` uses the
+It comprehends all the possible operations to be carried by a Google Cloud Pub/Sub client.
```
meltsufin: "Comprehends" is an odd choice of words. I also don't think it's completely accurate to say that it covers all possible operations.

spring-cloud-gcp-docs/src/main/asciidoc/pubsub.adoc
```diff
@@ -51,39 +50,64 @@ In the example, we're setting the number of threads used by both the `Publisher`
...
+
+Publishes a message containing the `payload` and `headers` to the target `topic`.
+The `payload` is in Google Cloud Pub/Sub Client Library's native payload format, `ByteString`.
+
+[source,java]
+----
+public ListenableFuture<String> publish(String topic, PubsubMessage pubsubMessage)
+----
+
+Low level method to publish a Google Cloud Pub/Sub `PubSubMessage` to the target `topic`.
```
meltsufin: This whole section above just feels like a duplication of the javadoc. Can we just link to the javadoc instead of maintaining the same information in two places?

spring-cloud-gcp-docs/src/main/asciidoc/pubsub.adoc
markfisher: could say "PubSubTemplate implements PubSubOperations, which is an abstraction that allows..."

joaoandremartins: One thing to have in mind is that the Spring Boot GCP Core Starter injects a `PubSubOperations` instance, not a `PubSubTemplate`, which is why I decided to make `PubSubOperations` more widespread.
We can change the Core Starter to redundantly provide a `PubSubTemplate` bean, which is an exact copy of the `PubSubOperations` it currently provides, to mitigate this risk.
WDYT?

spring-cloud-gcp-docs/src/main/asciidoc/pubsub.adoc
joaoandremartins: It sounds natural to me, but English isn't my native language!

How about "It specifies the typical set of operations carried by a Google Cloud Pub/Sub client." ?

spring-cloud-gcp-docs/src/main/asciidoc/pubsub.adoc
joaoandremartins: We currently don't have Javadoc published anywhere (we should probably look into this, the API link from http://cloud.spring.io/spring-cloud-gcp/ should link there).

Instead of making this listing, I can just say there are `publish()` flavours that can handle different `Message` payload types such as `String`, `byte[]`, `ByteString` and `PubsubMessage`...

meltsufin: Yes, this is better, and maybe create an issue for publishing javadocs.

spring-cloud-gcp-docs/src/main/asciidoc/pubsub.adoc
meltsufin: Nor mine! :)

 "It provides the common set of operations needed for interacting with Google Cloud Pub/Sub.

spring-cloud-gcp-docs/src/main/asciidoc/pubsub.adoc
meltsufin: `PubSubTemplate` gets auto-wired just fine as is right now. So, I would take Mark's suggestion here.

joaoandremartins: It actually doesn't, which is why I'm suggesting we make the Core Starter provide it.

joaoandremartins: So as we found out last week, auto-wiring a `PubSubTemplate` without providing a `@Bean` for it has different behaviour depending on if a class is `@Configuration` or not.

If you keep reading the paragraph, for the remainder of the section, the main subject is `PubSubTemplate`, not `PubSubOperations`. Is it not good enough?

joaoandremartins:
Applied requested changes.

joaoandremartins:
Merging since this is mostly documentation.

PR #162, opened 2017-09-21
Adds project ID and credentials to every service configuration

### LINKED ISSUE #107: Also add project ID and credentials to different Boot Starters configurations
Reason being it's possible for users to have different services living in different project IDs. That alone justifies the need to specify different credentials per service, but also users might not want to have a single uber service account with all the privileges but, instead, have separate service accounts per project.

joaoandremartins:
It's possible to host different services in different GCP project IDs.
But the current spring-cloud-gcp configuration only allows for a single
project ID.

Fixes #107


spring-cloud-gcp-starters/spring-cloud-gcp-starter-config/src/main/java/org/springframework/cloud/gcp/config/GoogleConfigPropertySourceLocator.java
```diff
@@ -136,4 +144,9 @@ GoogleConfigEnvironment getRemoteEnvironment() throws IOException, HttpClientErr
 		}
 		return new MapPropertySource(PROPERTY_SOURCE_NAME, config);
 	}
+
+	@VisibleForTesting
```
artembilan: Do we really need this?
I'm pretty fine to have the getter for the property on the class.

joaoandremartins:
Docs added -- PTAL

spring-cloud-gcp-starters/spring-cloud-gcp-starter-config/src/main/java/org/springframework/cloud/gcp/config/GoogleConfigPropertySourceLocator.java
```diff
@@ -145,7 +144,6 @@ GoogleConfigEnvironment getRemoteEnvironment() throws IOException, HttpClientErr
 		return new MapPropertySource(PROPERTY_SOURCE_NAME, config);
 	}
 
-	@VisibleForTesting
 	String getProjectId() {
```
artembilan: That's minor and we can live with that, but don't think that I have missed that this method isn't `public` 😉

joaoandremartins: Good catch. PTAL

Issue #81, opened 2017-07-28
org.springframework.integration.gcp.outbound.PubsubMessageHandler swallows errors on published messages


balopat:
The following line in [org.springframework.integration.gcp.outbound.PubsubMessageHandler](https://github.com/spring-cloud/spring-cloud-gcp/blob/master/spring-integration-gcp/src/main/java/org/springframework/integration/gcp/outbound/PubsubMessageHandler.java#L41
) should be a blocking call to be able to determine whether the returned listenableFuture failed or succeeded: 
````
    protected void handleMessageInternal(Message<?> message) throws Exception {
        this.pubsubTemplate.send(this.topic, message); // this fails silently if e.g. a topic is not defined
    }
````
This could be just a simple ``.get()`` call, or we can even have a debug log which would print out the result in case of success.  

Optionally we could even take into account timeout configuration?

artembilan:
The idea how it should be implemented can be borrowed from the https://github.com/spring-projects/spring-integration-kafka/blob/master/src/main/java/org/springframework/integration/kafka/outbound/KafkaProducerMessageHandler.java

joaoandremartins:
Yep, this is in our roadmap already (#51). IIUC this issue is a duplicate.

meltsufin:
I'm not clear on why we can't even output an error to the log even if it's async?

PR #166, opened 2017-09-22
Adds sync publish mode to PubSubMessageHandler

### LINKED ISSUE #51: Add (a)sync mode to PubSubMessageHandler
There is an example on [KafkaMessageProducerHandler](https://github.com/spring-projects/spring-integration-kafka/blob/master/src/main/java/org/springframework/integration/kafka/outbound/KafkaProducerMessageHandler.java#L104).

joaoandremartins:
Fixes #51, #81


joaoandremartins:
I followed spring-integration-kafka's example, as [suggested by Artem](https://github.com/spring-cloud/spring-cloud-gcp/issues/81#issuecomment-318664280).

I figured it couldn't hurt. It's still possible to set a plain timeout in millis. There might be use cases for SpEL that I don't know about, but I'll let @artembilan comment on it. If we don't have any use for it, I can remove it.

artembilan:
> There might be use cases for SpEL 

That's correct. There was a community request to have a send timeout per message bases. Some messages are sent with one timeout, but other with different. We had a request for SI Kafka but followed the same pattern for SI AWS Kinesis.
So, I guess that won't hurt here as well.

Let me review what's going on so far!

spring-integration-gcp/src/main/java/org/springframework/integration/gcp/pubsub/outbound/PubSubMessageHandler.java
```diff
@@ -87,4 +117,43 @@ public void setMessageConverter(MessageConverter messageConverter) {
...
+	}
+
+	/**
+	 * Sets the SpEL expression to evaluate a timeout in milliseconds for a synchronous publish call
+	 * to Google Cloud Pub/Sub.
+	 *
+	 * @param publishTimeoutExpression the {@link Expression} for the publish timeout in
+	 *                                 milliseconds
+	 */
+	public void setPublishTimeoutExpression(Expression publishTimeoutExpression) {
```
artembilan: Also, please, add a variant for `String` parsing:
```
public void setPublishTimeoutExpressionString(String publishTimeoutExpression) {
		setPublishTimeoutExpression(EXPRESSION_PARSER.parseExpression(publish));
}
```
Really useful for Java config.

spring-integration-gcp/src/main/java/org/springframework/integration/gcp/pubsub/outbound/PubSubMessageHandler.java
```diff
@@ -87,4 +117,43 @@ public void setMessageConverter(MessageConverter messageConverter) {
 				"The specified message converter can't be null.");
 		this.messageConverter = messageConverter;
 	}
+
+	public boolean isSync() {
+		return this.sync;
+	}
+
+	public void setSync(boolean sync) {
```
artembilan: Need JavaDocs.

joaoandremartins:
I think I addressed all the comments except for the tests.
What tests were you hoping to see here?

artembilan:
> What tests were you hoping to see here?

At least mock tests to be sure that our setters has an effect for target logic.

spring-integration-gcp/src/test/java/org/springframework/integration/gcp/pubsub/outbound/PubSubMessageHandlerTests.java
```diff
@@ -69,4 +74,15 @@ public void testPublish() {
...
+
+	@Test
+	public void testPublishSync() {
+		this.adapter.setSync(true);
+		Expression timeout = spy(this.adapter.getPublishTimeoutExpression());
+		this.adapter.setPublishTimeoutExpression(timeout);
+
+		this.adapter.handleMessage(this.message);
+		verify(timeout, times(1)).getValue(
+				eq(null), eq(this.message), eq(Long.class));
```
joaoandremartins: @artembilan I couldn't find a way to mock the `onInit()` behaviour, so I just put `null` as the first arg for `evaluationContext`. Let me know if there is anything I could do here!

artembilan: Well, I would mock `PubsubTemplate` and verify that it is invoked with the proper args and return value .
But this is fine since we ensure that expression is called downstream.

joaoandremartins: Yeah. The previous test ensures the `PubSubTemplate` call is done. I didn't think it'd be necessary here since that happens outside of the `if (this.sync)` block.

PR #178, opened 2017-09-28
Logs errors on publish

### LINKED ISSUE #81: org.springframework.integration.gcp.outbound.PubsubMessageHandler swallows errors on published messages

joaoandremartins:
Errors on async publish are currently not logged, which can be confusing
to users.

Fixes #81

### DIFF
--- a/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
+++ b/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
@@ -26,6 +26,8 @@ import com.google.cloud.pubsub.v1.MessageReceiver;
 import com.google.cloud.pubsub.v1.Subscriber;
 import com.google.protobuf.ByteString;
 import com.google.pubsub.v1.PubsubMessage;
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
 
 import org.springframework.beans.factory.InitializingBean;
 import org.springframework.cloud.gcp.pubsub.support.PublisherFactory;
@@ -40,6 +42,8 @@ import org.springframework.util.concurrent.SettableListenableFuture;
  * @author João André Martins
  */
 public class PubSubTemplate implements PubSubOperations, InitializingBean {
+
+	private static final Log LOGGER = LogFactory.getLog(PubSubTemplate.class);
 
 	private final PublisherFactory publisherFactory;
 
@@ -90,11 +94,13 @@ public class PubSubTemplate implements PubSubOperations, InitializingBean {
 
 			@Override
 			public void onFailure(Throwable throwable) {
+				LOGGER.warn("Publishing to " + topic + " topic failed.", throwable);
 				settableFuture.setException(throwable);
 			}
 
 			@Override
 			public void onSuccess(String result) {
+				LOGGER.debug("Publishing to " + topic + " was successful. Message ID: " + result);
 				settableFuture.set(result);
 			}
 
--- a/spring-cloud-gcp-docs/src/main/asciidoc/spring-integration.adoc
+++ b/spring-cloud-gcp-docs/src/main/asciidoc/spring-integration.adoc
@@ -115,6 +115,5 @@ The provided `PubSubTemplate` contains all the necessary configuration to publis
 GCP Pub/Sub topic.
 
 `PubSubMessageHandler` publishes messages asynchronously by default.
-On asynchronous mode, publish errors are not displayed.
 A publish timeout can be configured for synchronous publishing. If none is provided, the adapter
 waits indefinitely for a response.
--- a/spring-integration-gcp/README.adoc
+++ b/spring-integration-gcp/README.adoc
@@ -25,7 +25,6 @@ Makes extensive use of the constructs provided by the Spring Cloud GCP Pub/Sub m
 `PubSubOperations`.
 
 The outbound channel adapter, `PubSubMessageHandler`, publishes messages asynchronously by default.
-On asynchronous mode, publish errors are not displayed.
 A publish timeout can be configured for synchronous publishing. If none is provided, the adapter
 waits indefinitely for a response.
 


joaoandremartins:
A consequence of this is that errors will be logged twice for SI sync publish.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
```diff
@@ -90,11 +94,13 @@ public PubSubTemplate(PublisherFactory publisherFactory, SubscriberFactory subsc
 
 			@Override
 			public void onFailure(Throwable throwable) {
+				LOGGER.warn("Publishing to " + topic + " topic failed.", throwable);
```
meltsufin: I think this might be OK, but the alternative is to just make `PubSubMessageHandler.handleMessageInternal` add an error callback to the `ListenableFuture` returned by `PubSubTemplate`, and log the error there if `sync == false`. WDYT?

joaoandremartins: That's one possibility, but then `PubSubTemplate` clients wouldn't see any errors logged..

Another alternative is to log the error as debug instead of warn, but then errors still aren't easily visible. I think we can live with errors being logged twice for SI?

meltsufin: I think the expectation for users of the template is that they should use the returned `ListenableFuture` to handle errors.

artembilan: I'm fine having logs here in the  `PubSubTemplate`, but for the `PubSubMessageHandler` I would suggest to inject `ListenableFutureCallback` to really react to the async results.
The similar behavior we have in the `KafkaProducerMessageHandler`: https://github.com/spring-projects/spring-integration-kafka/blob/master/src/main/java/org/springframework/integration/kafka/outbound/KafkaProducerMessageHandler.java#L259
Or https://github.com/spring-projects/spring-integration-aws/blob/master/src/main/java/org/springframework/integration/aws/outbound/KinesisMessageHandler.java#L170

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
```diff
@@ -90,11 +94,13 @@ public PubSubTemplate(PublisherFactory publisherFactory, SubscriberFactory subsc
...
 				settableFuture.setException(throwable);
 			}
 
 			@Override
 			public void onSuccess(String result) {
+				LOGGER.debug("Publishing to " + topic + " was successful. Message ID: " + result);
```
artembilan: This must be wrapped with the `LOGGER.ifDebugEnabled()`

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
artembilan: Well, I commented that way already: https://github.com/spring-cloud/spring-cloud-gcp/issues/81#issuecomment-318664280

meltsufin: More specifically, are you talking about an error channel? There's a lot that's going on in the Kafka message handler. :)

artembilan: Well, let's start from the simple `ListenableFutureCallback` injection!
That should be enough for now.

joaoandremartins: @artembilan but what would that `ListenableFutureCallback` achieve? Logging the error? We already do that in `PubSubTemplate`. Unless we want to add failure channels to `PubSubMessageHandler`, I'm not sure what `ListenableFutureCallback` would accomplish.

@meltsufin how would `PubSubTemplate` users intercept async errors in that case? I think adding a callback to the returned future wouldn't do because the call might have already returned by the time the future is returned.

artembilan: The injected `ListenableFutureCallback` by end-user will exactly do anything (s)he needs. That their injection and that is fully up to them what to do in the implementation.

Right, by default we may have `LoggingListenableFutureCallback` to really log error/success.

That doesn't matter if `ListenableFuture` has been finished or not before `Callback`  injection. The last one is called appropriately.

joaoandremartins: PTAL hope this is what you were referring to!

spring-integration-gcp/src/main/java/org/springframework/integration/gcp/pubsub/outbound/PubSubMessageHandler.java
```diff
@@ -170,6 +177,19 @@ public void setPublishTimeout(long timeoutMillis) {
 		setPublishTimeoutExpression(new ValueExpression<>(timeoutMillis));
 	}
 
+	public ListenableFutureCallback<String> getPublishFutureCallback() {
```
artembilan: Why do you always add these `public` getters?
When I use `PubSubMessageHandler` from IDE declaring bean, I have an extra noise for unexpected getters.
I somehow think there is just enough to let end-user to see what (s)he expects - setters.
I don't mind to have getters for properties, but let's make them `protected` for unusual extensions.

spring-integration-gcp/src/main/java/org/springframework/integration/gcp/pubsub/outbound/PubSubMessageHandler.java
```diff
@@ -61,6 +62,8 @@
 
 	private Expression publishTimeoutExpression = new ValueExpression<>(DEFAULT_PUBLISH_TIMEOUT);
 
+	private ListenableFutureCallback<String> publishFutureCallback;
```
meltsufin: Maybe just `publishCallback`?

spring-integration-gcp/src/main/java/org/springframework/integration/gcp/pubsub/outbound/PubSubMessageHandler.java
```diff
@@ -96,6 +99,10 @@ else if (payload instanceof ByteString) {
 		ListenableFuture<String> pubsubFuture =
 				this.pubSubTemplate.publish(this.topic, pubsubPayload, headers);
 
+		if (this.publishFutureCallback != null) {
+			pubsubFuture.addCallback(this.publishFutureCallback);
```
meltsufin: We may want to just add a logging one when none is specified. In that case we wouldn't want the logging done in the template. I'm not sure if that is what @artembilan was thinking of when he mentioned `LoggingListenableFutureCallback`.

artembilan: Yeah... If we have logging on the `template` level, we don't need it on the adapter.
Since `template` can even be used outside of the SI, I don't see reason to overhead with extra default callback.
Current state is fine with me.

meltsufin: sg

spring-integration-gcp/src/main/java/org/springframework/integration/gcp/pubsub/outbound/PubSubMessageHandler.java
artembilan: +1 and setter would sound better - `setPublishCallback`

PR #179, opened 2017-09-28
Resets spring-cloud-sleuth version

joaoandremartins:
1.2.4 suffers from the issue in
https://github.com/spring-cloud/spring-cloud-sleuth/issues/684.

Resetting the version to 1.3.0.M1 until it gets to GA.

### DIFF
--- a/spring-cloud-gcp-dependencies/pom.xml
+++ b/spring-cloud-gcp-dependencies/pom.xml
@@ -197,18 +197,6 @@
 		</dependencies>
 	</dependencyManagement>
 
-	<!-- Remove this when spring-cloud-sleuth 1.3.0 moves to GA -->
-	<repositories>
-		<repository>
-			<id>spring-milestones</id>
-			<name>Spring Milestones</name>
-			<url>https://repo.spring.io/libs-milestone-local</url>
-			<snapshots>
-				<enabled>false</enabled>
-			</snapshots>
-		</repository>
-	</repositories>
-
 	<profiles>
 		<profile>
 			<id>spring</id>
--- a/spring-cloud-gcp-dependencies/pom.xml
+++ b/spring-cloud-gcp-dependencies/pom.xml
@@ -22,7 +22,7 @@
 		<google-cloud-storage.version>1.2.1</google-cloud-storage.version>
 		<cloud-trace-java.version>0.4.0</cloud-trace-java.version>
 		<spring-cloud-context.version>1.2.3.RELEASE</spring-cloud-context.version>
-		<spring-cloud-sleuth.version>1.2.4.RELEASE</spring-cloud-sleuth.version>
+		<spring-cloud-sleuth.version>1.3.0.M1</spring-cloud-sleuth.version>
 	</properties>
 
 	<dependencyManagement>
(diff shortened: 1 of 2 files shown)


joaoandremartins:
I'm adding the `<repositories>` block in our BOM in the hopes that users won't have to add it themselves.
I'm not sure if this will work, though.

spring-cloud-gcp-dependencies/pom.xml
```diff
@@ -197,6 +197,18 @@
 		</dependencies>
 	</dependencyManagement>
 
+	<!-- Remove this when spring-cloud-sleuth 1.3.0 moves to GA -->
+	<repositories>
+		<repository>
+			<id>spring-milestones</id>
+			<name>Spring Milestones</name>
+			<url>https://repo.spring.io/libs-milestone-local</url>
```
artembilan: I think that's already not our responsibility from where end-user get his/her dependencies.
How about to follow Spring Boot approach?
https://github.com/spring-projects/spring-boot/blob/master/spring-boot-dependencies/pom.xml#L2839

Not sure, though, how they remove that section during release: https://search.maven.org/#artifactdetails%7Corg.springframework.boot%7Cspring-boot-dependencies%7C1.5.7.RELEASE%7Cpom

@wilkinsona ,
would you mind sharing your experience, thanks?

wilkinsona: I agree with Artem. Putting a repository definition in a pom isn't recommended as it pollutes a user's whole build with a repository that they may not want or be able to use.

We strip out the repositories like this: https://github.com/spring-projects/spring-boot/blob/4caeed2ac4cd295b78a1e6eba053cb4f88256e56/pom.xml#L336-L357. We have a reworked build on its way where we're using the Maven Flatten plugin instead.

artembilan: Thank you, Andy, for confirmation!

We definitely can get into service your solution.

On the other hand our parent is `<artifactId>spring-cloud-build</artifactId>` which brings all the repos for development. Therefore we have to just remove this entry from the BOM.

artembilan: Or I might be wrong...
That is about `distribution`:
```
<distributionManagement>
	<repository>
		<id>repo.spring.io</id>
		<name>Spring Milestone Repository</name>
		<url>https://repo.spring.io/libs-milestone-local</url>
	</repository>
</distributionManagement>
```

joaoandremartins: It seems that the best way is then just to recommend users to add the Milestones repo to their pom.xml until 1.3.0 goes GA.
I wanted to avoid this, but it seems like we don't have any better option.

joaoandremartins: My knowledge of Maven is still a bit limited, but I wonder if users run `mvn clean package -P spring` on their side, if the Milestones repo is picked up from`spring-cloud-gcp-dependencies`, or if that's even advisable due to the build pollution concerns..

artembilan: I think we should stop this discussion because it is going to be really an abuse to force end-users to rely on our `/milestone`.
We definitely can recommend to use it from the docs, but that should not be an unconditional option from our pom.

joaoandremartins: Until spring-cloud-sleuth 1.3.0 goes GA, we'll need to have users add Spring's Milestone repo to their build files.
I added that bit in the README for the Trace Starter so it's a little bit less invasive. PTAL

PR #180, opened 2017-09-28
Return PubSubTemplate rather than PubSubOperations

saturnism:
For the autoconfiguration, return PubSubTemplate. This is inline w/ other starter templates, e.g. RabbitMQ Template, MongoTemplate, etc.


joaoandremartins:
This PR will require some changes in our docs, but I can perform them later.

Issue #155, opened 2017-09-18
Move properties from starter module and autoconfig to core module


saturnism:
E.g., PubSub autoconfiguration currently lives in the starter module, but shouldn't it be in the core module?

Most of the autoconfig checks for .enabled flag in properties to offer the user to turn it off. 
Starters only manages the default dependencies

joaoandremartins:
`@ConfigurationProperties` is a Boot concept. I don't think we would want the core module to depend on Boot, but let me know if I'm missing anything @saturnism

joaoandremartins:
#262 captures this requirement in more detail. Let's follow up there.

PR #185, opened 2017-10-05
Added default option to create blob if not exists

### LINKED ISSUE #165: Auto-create blob in Storage with GoogleStorageResource.getOutputStream()
Our implementation of the `WritableResource` (`GoogleStorageResource`) doesn't currently auto-create the file when you call `getOutputStream(`), but it would be useful to automatically create it if it doesn't exist. This behavior would mirror the `FileSystemResource`.

ChengyuanZhao:
fixes #165

### DIFF
--- a/spring-cloud-gcp-starters/spring-cloud-gcp-starter-storage/README.adoc
+++ b/spring-cloud-gcp-starters/spring-cloud-gcp-starter-storage/README.adoc
@@ -23,10 +23,10 @@ dependencies {
 This starter provides the underpinnings of the Spring Cloud GCP Storage. It doesn't provide any
 bean to be used by an end-user.
 
-There is an optional property, `spring.cloud.gcp.storage.credentials-location`. Its value is the
-filesystem path for a credentials file, in case Storage is served from a different GCP project ID
-than the one configured in the link:../spring-cloud-gcp-starter-core/README.adoc[Core Starter].
-
-The optional property `spring.cloud.gcp.storage.auto-create-files` causes a new file to be created
-with a given URI if an output stream was requested from that URI but there was no file at that URI.
-If `true`, the default setting, then new files are automatically created.
+[options="header",]
+|=======================================================================
+|Setting |Optional |Default Value |Meaning
+|`spring.cloud.gcp.storage.credentials-location`|Yes|Configured in the link:../spring-cloud-gcp-starter-core/README.adoc[Core Starter]|the
+filesystem path for a credentials file
+|`spring.cloud.gcp.storage.auto-create-files`|Yes|`true`|If `true`, new file will be created at given URL if none exists when getting output stream
+|=======================================================================
--- a/spring-cloud-gcp-starters/spring-cloud-gcp-starter-storage/README.adoc
+++ b/spring-cloud-gcp-starters/spring-cloud-gcp-starter-storage/README.adoc
@@ -26,3 +26,7 @@ bean to be used by an end-user.
 There is an optional property, `spring.cloud.gcp.storage.credentials-location`. Its value is the
 filesystem path for a credentials file, in case Storage is served from a different GCP project ID
 than the one configured in the link:../spring-cloud-gcp-starter-core/README.adoc[Core Starter].
+
+The optional property `spring.cloud.gcp.storage.auto-create-files` causes a new file to be created
+with a given URI if an output stream was requested from that URI but there was no file at that URI.
+If `true`, the default setting, then new files are automatically created.
--- a/spring-cloud-gcp-starters/spring-cloud-gcp-starter-storage/src/main/java/org/springframework/cloud/gcp/storage/autoconfig/StorageAutoConfiguration.java
+++ b/spring-cloud-gcp-starters/spring-cloud-gcp-starter-storage/src/main/java/org/springframework/cloud/gcp/storage/autoconfig/StorageAutoConfiguration.java
@@ -68,8 +68,7 @@ public class StorageAutoConfiguration {
 	}
 
 	@Bean
-	@ConditionalOnMissingBean
-	public static GoogleStorageProtocolResolverSettings googleStorageProtocolResolverContext(
+	public static GoogleStorageProtocolResolverSettings googleStorageProtocolResolverSettings(
 			GcpStorageProperties gcpStorageProperties) throws IOException {
 		return new GoogleStorageProtocolResolverSettings(
 				gcpStorageProperties.isAutoCreateFiles());
--- a/spring-cloud-gcp-starters/spring-cloud-gcp-starter-storage/src/main/java/org/springframework/cloud/gcp/storage/autoconfig/StorageAutoConfiguration.java
+++ b/spring-cloud-gcp-starters/spring-cloud-gcp-starter-storage/src/main/java/org/springframework/cloud/gcp/storage/autoconfig/StorageAutoConfiguration.java
@@ -66,11 +66,4 @@ public class StorageAutoConfiguration {
 						: credentialsProvider.getCredentials())
 				.build().getService();
 	}
-
-	@Bean
-	public static GoogleStorageProtocolResolverSettings googleStorageProtocolResolverSettings(
-			GcpStorageProperties gcpStorageProperties) throws IOException {
-		return new GoogleStorageProtocolResolverSettings(
-				gcpStorageProperties.isAutoCreateFiles());
-	}
 }
--- a/spring-cloud-gcp-starters/spring-cloud-gcp-starter-storage/src/main/java/org/springframework/cloud/gcp/storage/autoconfig/StorageAutoConfiguration.java
+++ b/spring-cloud-gcp-starters/spring-cloud-gcp-starter-storage/src/main/java/org/springframework/cloud/gcp/storage/autoconfig/StorageAutoConfiguration.java
@@ -21,7 +21,6 @@ import java.util.Collections;
 
 import com.google.api.gax.core.CredentialsProvider;
 import com.google.auth.oauth2.GoogleCredentials;
-import com.google.cloud.storage.Storage;
 import com.google.cloud.storage.StorageOptions;
 
 import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
@@ -31,14 +30,15 @@ import org.springframework.cloud.gcp.core.GcpProperties;
 import org.springframework.cloud.gcp.core.GcpScope;
 import org.springframework.cloud.gcp.storage.GcpStorageProperties;
 import org.springframework.cloud.gcp.storage.GoogleStorageProtocolResolver;
+import org.springframework.cloud.gcp.storage.GoogleStorageProtocolResolverContext;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.context.annotation.Import;
 
 /**
- * An auto-configuration for Google {@link Storage} bean definition. Also it
- * {@link Import} a {@link GoogleStorageProtocolResolver} to register it
- * with the {@code DefaultResourceLoader}.
+ * An auto-configuration for Google {@link GoogleStorageProtocolResolverContext} bean
+ * definition. Also it {@link Import} a {@link GoogleStorageProtocolResolver} to register
+ * it with the {@code DefaultResourceLoader}.
  *
  * @author Vinicius Carvalho
  * @author Artem Bilan
@@ -46,17 +46,18 @@ import org.springframework.context.annotation.Import;
  * @see GoogleStorageProtocolResolver
  */
 @Configuration
-@ConditionalOnClass(Storage.class)
+@ConditionalOnClass(GoogleStorageProtocolResolverContext.class)
 @EnableConfigurationProperties({GcpProperties.class, GcpStorageProperties.class})
 @Import(GoogleStorageProtocolResolver.class)
 public class StorageAutoConfiguration {
 
 	@Bean
 	@ConditionalOnMissingBean
-	public static Storage storage(
+	public static GoogleStorageProtocolResolverContext googleStorageProtocolResolverContext(
 			CredentialsProvider credentialsProvider,
 			GcpStorageProperties gcpStorageProperties) throws IOException {
-		return StorageOptions.newBuilder()
+		return new GoogleStorageProtocolResolverContext(
+				StorageOptions.newBuilder()
 				.setCredentials(gcpStorageProperties.getCredentialsLocation() != null
 						? GoogleCredentials.fromStream(
 								gcpStorageProperties.getCredentialsLocation().getInputStream())
@@ -64,7 +65,7 @@ public class StorageAutoConfiguration {
 								GcpScope.STORAGE_READ_WRITE.getUrl()))
 						: credentialsProvider.getCredentials())
 				.build()
-				.getService();
+						.getService(),
+				gcpStorageProperties.isAutoCreateFiles());
 	}
-
 }
--- a/spring-cloud-gcp-starters/spring-cloud-gcp-starter-storage/src/main/java/org/springframework/cloud/gcp/storage/autoconfig/StorageAutoConfiguration.java
+++ b/spring-cloud-gcp-starters/spring-cloud-gcp-starter-storage/src/main/java/org/springframework/cloud/gcp/storage/autoconfig/StorageAutoConfiguration.java
@@ -21,6 +21,7 @@ import java.util.Collections;
 
 import com.google.api.gax.core.CredentialsProvider;
 import com.google.auth.oauth2.GoogleCredentials;
+import com.google.cloud.storage.Storage;
 import com.google.cloud.storage.StorageOptions;
 
 import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
@@ -30,13 +31,13 @@ import org.springframework.cloud.gcp.core.GcpProperties;
 import org.springframework.cloud.gcp.core.GcpScope;
 import org.springframework.cloud.gcp.storage.GcpStorageProperties;
 import org.springframework.cloud.gcp.storage.GoogleStorageProtocolResolver;
-import org.springframework.cloud.gcp.storage.GoogleStorageProtocolResolverContext;
+import org.springframework.cloud.gcp.storage.GoogleStorageProtocolResolverSettings;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.context.annotation.Import;
 
 /**
- * An auto-configuration for Google {@link GoogleStorageProtocolResolverContext} bean
+ * An auto-configuration for Google {@link GoogleStorageProtocolResolverSettings} bean
  * definition. Also it {@link Import} a {@link GoogleStorageProtocolResolver} to register
  * it with the {@code DefaultResourceLoader}.
  *
@@ -46,26 +47,31 @@ import org.springframework.context.annotation.Import;
  * @see GoogleStorageProtocolResolver
  */
 @Configuration
-@ConditionalOnClass(GoogleStorageProtocolResolverContext.class)
+@ConditionalOnClass({ GoogleStorageProtocolResolverSettings.class, Storage.class })
 @EnableConfigurationProperties({GcpProperties.class, GcpStorageProperties.class})
 @Import(GoogleStorageProtocolResolver.class)
 public class StorageAutoConfiguration {
 
 	@Bean
 	@ConditionalOnMissingBean
-	public static GoogleStorageProtocolResolverContext googleStorageProtocolResolverContext(
-			CredentialsProvider credentialsProvider,
+	public static Storage storage(CredentialsProvider credentialsProvider,
 			GcpStorageProperties gcpStorageProperties) throws IOException {
-		return new GoogleStorageProtocolResolverContext(
-				StorageOptions.newBuilder()
+		return StorageOptions.newBuilder()
 				.setCredentials(gcpStorageProperties.getCredentialsLocation() != null
-						? GoogleCredentials.fromStream(
-								gcpStorageProperties.getCredentialsLocation().getInputStream())
-						.createScoped(Collections.singletonList(
-								GcpScope.STORAGE_READ_WRITE.getUrl()))
+						? GoogleCredentials
+								.fromStream(gcpStorageProperties.getCredentialsLocation()
+										.getInputStream())
+								.createScoped(Collections.singletonList(
+										GcpScope.STORAGE_READ_WRITE.getUrl()))
 						: credentialsProvider.getCredentials())
-				.build()
-						.getService(),
+				.build().getService();
+	}
+
+	@Bean
+	@ConditionalOnMissingBean
+	public static GoogleStorageProtocolResolverSettings googleStorageProtocolResolverContext(
+			GcpStorageProperties gcpStorageProperties) throws IOException {
+		return new GoogleStorageProtocolResolverSettings(
 				gcpStorageProperties.isAutoCreateFiles());
 	}
 }
--- a/spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageProtocolResolver.java
+++ b/spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageProtocolResolver.java
@@ -45,8 +45,9 @@ public class GoogleStorageProtocolResolver
 
 	private ConfigurableListableBeanFactory beanFactory;
 
-	private GoogleStorageProtocolResolverSettings settings;
-	private boolean settingsAvailable = true;
+	private GoogleStorageProtocolResolverSettings settings
+			= new GoogleStorageProtocolResolverSettings();
+	private boolean checkedIfSettingsBeanAvailable;
 
 	private volatile Storage storage;
 
@@ -73,22 +74,23 @@ public class GoogleStorageProtocolResolver
 	@Override
 	public Resource resolve(String location, ResourceLoader resourceLoader) {
 		if (location.startsWith(PROTOCOL)) {
-			if (this.settingsAvailable && this.settings == null) {
+			if (!this.checkedIfSettingsBeanAvailable) {
 				try {
 					this.settings = this.beanFactory
 							.getBean(GoogleStorageProtocolResolverSettings.class);
 				}
 				catch (NoSuchBeanDefinitionException e) {
-					logger.warn("There is no bean definition for the resolver settings.",
-							e);
-					this.settingsAvailable = false;
+					logger.info("There is no bean definition for the resolver settings, "
+							+ "so defaults are used.");
+				}
+				finally {
+					this.checkedIfSettingsBeanAvailable = true;
 				}
 			}
 			if (this.storage == null) {
 				this.storage = this.beanFactory.getBean(Storage.class);
 			}
-			return new GoogleStorageResource(this.storage, location,
-					this.settingsAvailable ? this.settings.isAutoCreateFiles() : true);
+			return new GoogleStorageResource(this.storage, location, this.settings.isAutoCreateFiles());
 		}
 		return null;
 	}
--- a/spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageProtocolResolver.java
+++ b/spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageProtocolResolver.java
@@ -16,8 +16,6 @@
 
 package org.springframework.cloud.gcp.storage;
 
-import com.google.cloud.storage.Storage;
-
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 
@@ -36,7 +34,8 @@ import org.springframework.core.io.ResourceLoader;
  * @author Vinicius Carvalho
  * @author Artem Bilan
  */
-public class GoogleStorageProtocolResolver implements ProtocolResolver, BeanFactoryPostProcessor, ResourceLoaderAware {
+public class GoogleStorageProtocolResolver
+		implements ProtocolResolver, BeanFactoryPostProcessor, ResourceLoaderAware {
 
 	private static final Log logger = LogFactory.getLog(GoogleStorageProtocolResolver.class);
 
@@ -44,13 +43,14 @@ public class GoogleStorageProtocolResolver implements ProtocolResolver, BeanFact
 
 	private ConfigurableListableBeanFactory beanFactory;
 
-	private volatile Storage storage;
+	private volatile GoogleStorageProtocolResolverContext context;
 
 	GoogleStorageProtocolResolver() {
 	}
 
 	@Override
-	public void postProcessBeanFactory(ConfigurableListableBeanFactory beanFactory) throws BeansException {
+	public void postProcessBeanFactory(ConfigurableListableBeanFactory beanFactory)
+			throws BeansException {
 		this.beanFactory = beanFactory;
 	}
 
@@ -68,10 +68,12 @@ public class GoogleStorageProtocolResolver implements ProtocolResolver, BeanFact
 	@Override
 	public Resource resolve(String location, ResourceLoader resourceLoader) {
 		if (location.startsWith(PROTOCOL)) {
-			if (this.storage == null) {
-				this.storage = this.beanFactory.getBean(Storage.class);
+			if (this.context == null) {
+				this.context = this.beanFactory
+						.getBean(GoogleStorageProtocolResolverContext.class);
 			}
-			return new GoogleStorageResource(this.storage, location);
+			return new GoogleStorageResource(this.context.getStorage(), location,
+					this.context.isAutoCreateFiles());
 		}
 		return null;
 	}
--- a/spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageProtocolResolver.java
+++ b/spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageProtocolResolver.java
@@ -16,10 +16,12 @@
 
 package org.springframework.cloud.gcp.storage;
 
+import com.google.cloud.storage.Storage;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 
 import org.springframework.beans.BeansException;
+import org.springframework.beans.factory.NoSuchBeanDefinitionException;
 import org.springframework.beans.factory.config.BeanFactoryPostProcessor;
 import org.springframework.beans.factory.config.ConfigurableListableBeanFactory;
 import org.springframework.context.ResourceLoaderAware;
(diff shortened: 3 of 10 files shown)


spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResource.java
```diff
@@ -46,16 +47,32 @@
 
 	private final String location;
 
-	public GoogleStorageResource(Storage storage, String location) {
+	private boolean createBlobIfNotExists = true;
```
joaoandremartins: How about we surface this through a property rather than a private field?

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResource.java
```diff
@@ -46,16 +47,32 @@
...
+
+	public void setCreateBlobIfNotExists(boolean createBlobIfNotExists) {
+		this.createBlobIfNotExists = createBlobIfNotExists;
 	}
 
 	@Override
 	public boolean exists() {
 		try {
-			return resolve().exists();
+			return throwExceptionForNullBlob(resolve()).exists();
```
meltsufin: I think this violates the contract of the `exists` method. It shouldn't throw an exception if the file doesn't exist.

artembilan:
@ChengyuanZhao ,

Thank you for contribution.
Would you mind confirming that you have signed CLA: https://cla.pivotal.io/ ?

I think we need to add Webhook integration (https://cla.pivotal.io/github/hooks/pull_request/spring) here in project for the external contributors. Unfortunately I don't have permissions to do that. 

@markfisher, would you mind helping us?
Thanks

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResource.java
```diff
@@ -93,7 +110,18 @@ private Blob resolve() throws IOException {
 		URI uri = getURI();
 		BlobId blobId = BlobId.of(uri.getHost(),
 				uri.getPath().substring(1, uri.getPath().length()));
-		return this.storage.get(blobId);
+		Blob blob = this.storage.get(blobId);
+		if ((blob == null || !blob.exists()) && this.createBlobIfNotExists) {
+			return this.storage.create(BlobInfo.newBuilder(blobId).build());
```
meltsufin: We should only create the blob on an attempt to write to it, not on just any `resovle()`

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResource.java
```diff
@@ -104,12 +132,12 @@ public File getFile() throws IOException {
 
 	@Override
 	public long contentLength() throws IOException {
-		return resolve().getSize();
+		return throwExceptionForNullBlob(resolve()).getSize();
```
meltsufin: Let `resolve()` throw the exception.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResource.java
```diff
@@ -144,6 +172,6 @@ public boolean isWritable() {
 
 	@Override
 	public OutputStream getOutputStream() throws IOException {
-		return Channels.newOutputStream(resolve().writer());
+		return Channels.newOutputStream(throwExceptionForNullBlob(resolve()).writer());
```
meltsufin: Create file here if one doesn't exist.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResource.java
artembilan: @meltsufin ,

We don't review until the CLA confirmation. The contributor must confirm that any our comments won't insult him/her.

Thanks for understanding.

meltsufin: @ChengyuanZhao is a new member on my team. He's covered under the Google CLA. :-)

artembilan: Sure thing, @meltsufin !

Thank you for confirmation!

Anyway I think the Webhook for CLA is something to must have.

meltsufin: Maybe it's already on, but wasn't triggered because Chengyuan has a google.com email?

artembilan: I don't think so. Even my PR's to my own projects are marked with the CLA confirmations:

image


That's why I think we just don't have it in this project.

meltsufin: I see

joaoandremartins: The issue might be that @ChengyuanZhao 's membership of the GoogleCloudPlatform organization is not visible.
You need to make it visible from your user settings (and join the organization if you haven't yet).

ChengyuanZhao: @meltsufin, I don't think this changes the behavior though. Originally resolve() returned this.storage.get() directly, and the .get() function's javadoc says it can return null possibly if the blob doesn't exist, which would mean the function would get an NPE if calling resolve().exists(), right? The difference now is that we get an IOException instead of NPE.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResource.java
ChengyuanZhao: @joaoandremartins apologies for the dumb question (still building up my Java vocab): do you mean I should have a getter/setter for this or that I should be using like System.getProperty properties?

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResource.java
ChengyuanZhao: This would be a different exception reason, though. This one would get thrown because resolve() might return null, on which we then call.getSize() resulting in an NPE. resolve() does this.storage.get(), which says it can return null if blob isn't found.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResource.java
artembilan: I think the idea is to `try...catch` that exception and return `false`.
See `AbstractFileResolvingResource` implementation for example:
```
catch (IOException ex) {
	return false;
}
```
in the end.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResource.java
artembilan: +1 for this point. Right, no reason to create if no request for that.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResource.java
artembilan: I think we have to follow contract here and always throws `FileNotFoundException` if there is no resource and it isn't about `exists()` method

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResource.java
```diff
@@ -46,16 +47,32 @@
...
+	public GoogleStorageResource(Storage storage, String location) {
+		this(storage, location, true);
+	}
+
+	public boolean isCreateBlobIfNotExists() {
+		return this.createBlobIfNotExists;
+	}
+
+	public void setCreateBlobIfNotExists(boolean createBlobIfNotExists) {
+		this.createBlobIfNotExists = createBlobIfNotExists;
```
artembilan: I think we don't need this setter. The ctor arg is fully enough to avoid mutation in the future.
And, of course, needs to make `createBlobIfNotExists` property as `final`

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResource.java
ChengyuanZhao: Ah my apologies I said it wrong. Before it would throw an NPE, but now instead it will throw an IOException, which would cause it to return false.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResource.java
joaoandremartins: @ChengyuanZhao no worries!
I was thinking that instead of this being in a private field, it should be in `GcpStorageProperties`. And then you would pass a `GcpStorageProperties` object to the constructor instead of a `boolean`.
WDYT?

artembilan: No, that's not. This `GoogleStorageResource` must be really "Plain Old Java Object". No any dependencies from other unrelated resources!

I'm not sure that it's OK to have a global configuration property on the matter though.

Something like `?noAutoCreate` as a parameter in the resource URL definition should be fully enough.

joaoandremartins:
Interesting that the tests passed, but this change should have made `GoogleStorageProtocolResolver` not compile because of the addition to the `GoogleStorageResource` constructor that isn't here.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResource.java
joaoandremartins: If that's the case, then at lease make it a property and then call `setCreateBlobIfNotExists()` on `GoogleStorageProtocolResolver`.

I don't think GCS uses the URL to pass parameters around. Maybe we shouldn't start now.

What issues do you see with the global property?

artembilan: With the global property we have an issue that we can't control it per instance basis.
That's why I suggest something like:
```
@Value("gs://test-spring/images/spring.png?autoCreate=true|false")
private Resource remoteResource;
```
Well, and from here it even looks like we don't need that additional property at all.
We can resolve parameter already in the `GoogleStorageResource` via provided `location`.

Well, I don't say that we are going to pass parameter to the GCS, because I think it isn't possible according existing logic:
```
BlobId blobId = BlobId.of(uri.getHost(),
				uri.getPath().substring(1, uri.getPath().length()));
```

Here is a result of my testing:
image

Pay, attention, please, to the selected row.

artembilan: `uri.getQuery().split("=")` comes to the rescue to extract the target flag.
I tend now to have it as `autoCreate=true|false` instead of just `noAutoCreate` because we may specify it as Properties Placeholder to be resolved against external properties:
```
@Value("gs://test-spring/images/spring.png?autoCreate=${my.autoCreate:false}")
```

ChengyuanZhao:
Apologies for the checkstyle commits, but I've had an issue with "mvn package" when trying to check style locally on an untouched pull of the master. 

Failed to execute goal org.codehaus.mojo:exec-maven-plugin:1.6.0:exec (imports) on project spring-cloud-gcp: Command execution failed. Process exited with an error: 0 (Exit value: 0) -> [Help 1]

meltsufin:
@ChengyuanZhao Are you running from the root of the project? Try `./mvnw clean install` to make sure it's not a Maven version issue.

ChengyuanZhao:
Thank you! that worked.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-storage/src/main/java/org/springframework/cloud/gcp/storage/autoconfig/StorageAutoConfiguration.java
```diff
@@ -31,40 +30,42 @@
...
-@ConditionalOnClass(Storage.class)
+@ConditionalOnClass(GoogleStorageProtocolResolverContext.class)
 @EnableConfigurationProperties({GcpProperties.class, GcpStorageProperties.class})
 @Import(GoogleStorageProtocolResolver.class)
 public class StorageAutoConfiguration {
 
 	@Bean
 	@ConditionalOnMissingBean
-	public static Storage storage(
+	public static GoogleStorageProtocolResolverContext googleStorageProtocolResolverContext(
```
meltsufin: It was nice being able to inject a `Storage` bean after adding the GCS starter to the app. Now, one will have to rely on the `GoogleStorageProtocolResolverContext` instead.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageProtocolResolverContext.java
```diff
@@ -0,0 +1,43 @@
...
+
+package org.springframework.cloud.gcp.storage;
+
+import com.google.cloud.storage.Storage;
+
+/**
+ * Holds {@link Storage} and associated settings for use with the
+ * {@link GoogleStorageResource}.
+ */
+public class GoogleStorageProtocolResolverContext {
```
meltsufin: Why not just call it `GoogleStorageResourceSettings`?

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageProtocolResolverContext.java
```diff
@@ -0,0 +1,43 @@
...
+
+	private final Storage storage;
```
meltsufin: I don't think we need the `Storage` instance here.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-storage/src/main/java/org/springframework/cloud/gcp/storage/autoconfig/StorageAutoConfiguration.java
meltsufin: I would just configure both beans.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-storage/README.adoc
```diff
@@ -26,3 +26,7 @@ bean to be used by an end-user.
 There is an optional property, `spring.cloud.gcp.storage.credentials-location`. Its value is the
 filesystem path for a credentials file, in case Storage is served from a different GCP project ID
 than the one configured in the link:../spring-cloud-gcp-starter-core/README.adoc[Core Starter].
+
+The optional property `spring.cloud.gcp.storage.auto-create-files` causes a new file to be created
+with a given URI if an output stream was requested from that URI but there was no file at that URI.
+If `true`, the default setting, then new files are automatically created.
```
saturnism: pls add configuration properties into a single table.

I feel this whole sentence should be rephrased since it says the same thing twice.  Again, this is much more expressible in a table

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageProtocolResolver.java
```diff
@@ -68,10 +73,22 @@ public void setResourceLoader(ResourceLoader resourceLoader) {
 	@Override
 	public Resource resolve(String location, ResourceLoader resourceLoader) {
 		if (location.startsWith(PROTOCOL)) {
+			if (this.settingsAvailable && this.settings == null) {
+				try {
+					this.settings = this.beanFactory
```
saturnism: why do we need to get the settings from a beanFactory? can this be initialized during construction time or postProcessBeanFactory time?

spring-cloud-gcp-starters/spring-cloud-gcp-starter-storage/src/main/java/org/springframework/cloud/gcp/storage/autoconfig/StorageAutoConfiguration.java
```diff
@@ -31,40 +31,47 @@
...
+										GcpScope.STORAGE_READ_WRITE.getUrl()))
 						: credentialsProvider.getCredentials())
-				.build()
-				.getService();
+				.build().getService();
 	}
 
+	@Bean
+	@ConditionalOnMissingBean
+	public static GoogleStorageProtocolResolverSettings googleStorageProtocolResolverContext(
```
saturnism: why do we need a new Settings object? can we not just re-use the gcp storage properties?

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageProtocolResolver.java
meltsufin: I think this has to do with the place of the resolver in the Spring application lifecycle.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-storage/src/main/java/org/springframework/cloud/gcp/storage/autoconfig/StorageAutoConfiguration.java
meltsufin: The properties object is in the stater and the core storage module doesn't depend on the starter. I'm not sure if it *has* to be that way though.

ChengyuanZhao: The Gcp storage properties is in the starter project, which depends on the cloud storage project, and to use the gcp storage properties in the cloud storage project would mean a circular dependency.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageProtocolResolver.java
ChengyuanZhao: This is because the primary example of how to use the autocreatefiles setting comes from the starter package in the gcp storage properties class. Because an instance of that gcp storage properties is needed for the Storage bean in that starter project, and at the same time we also get the autocreatefiles setting, it's pretty natural to create the two beans at the same time in the same way.

artembilan:
You know I came up with something like this: https://github.com/artembilan/spring-cloud-gcp/commit/f71e634601310dd43fac389fe1c35cf016dc52a9 and we don't need any `Contenxt`/`Settings` objects any more. That is for the case you really would like to have that global property. However I would like to see an answer to my request about `?autoCreate=true/false` param in the resource URL.

In addition would be great to see some tests on the matter, like my one about `properties = "spring.cloud.gcp.storage.autoCreateFiles=false"` but with additional assertion that we really carry an option to the target `GoogleStorageResource`.

Thanks

ChengyuanZhao:
I am thinking having the auto-create setting in the URL is maybe less intuitive than the global way because a user would need to keep in mind that the getOutputStream call could result in a FileNotFoundException exception if he has the autocreate=false set in the URL string, so he would need to handle that exception only in cases when he sets autocreate=false. He could get over that concern if he  decides to have that exception handling in either case (regardless of what URL setting he uses), but then why not just also use the global setting of auto-create being false? 

I'm also in favor of keeping the autocreate setting in GcpStorageProperties so that the "spring.cloud.gcp.storage" settings stay in the same place (both in GcpStorageProperties instead of one of them being in there and one being in StorageAutoConfiguration). I do think it's a good idea that we make a bean for GoogleStorageProtocolResolver in there. 

I'll also add the settings-to-resource test. 

What do you guys think @joaoandremartins , @meltsufin

artembilan:
> why not just also use the global setting of auto-create being false

I can live with your decision. That's fine. But we may really end up someday with the request "per resource". For example end-user may have WRITE permission in one bucket but not in another. Therefore that might `autoCreate` for the first one, but (s)he would like do not let to fail with permissions error, but really with the `FileNotFoundException`.

ChengyuanZhao:
I added the test, but I noticed that GoogleStorageProtocolResolver can't be used until its postProcessBeanFactory method is called and it's provided a bean factory, so it might not be that useful to have it as a bean in StorageAutoConfiguration.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageProtocolResolverSettings.java
```diff
@@ -0,0 +1,33 @@
...
+ *  limitations under the License.
+ */
+
+package org.springframework.cloud.gcp.storage;
+
+/**
+ * Holds settings for use with the {@link GoogleStorageResource}. These settings are
+ * optional.
+ */
+public class GoogleStorageProtocolResolverSettings {
```
artembilan: Why do you still try to keep this class?
Why my suggestion in the mentioned commit doesn't work for?

OK. If you really would like to keep this `setting` intermediate class, let's consider to do `GcpStorageProperties extends GoogleStorageProtocolResolverSettings`!

artembilan:
Pay attention also, please, to the Travis report:
```
home/travis/build/spring-cloud/spring-cloud-gcp/spring-cloud-gcp-starters/spring-cloud-gcp-starter-storage/src/test/java/org/springframework/cloud/gcp/storage/GoogleStorageAutoConfigurationTests.java:46: error: Line is longer than 120 characters (found 140).
```

artembilan:
>  so it might not be that useful to have it as a bean in StorageAutoConfiguration.

It can be but it must be as `static` `@Bean` method and must not have dependencies on other beans.
Right, just because it is `BeanFactoryPostProcessor`:
https://docs.spring.io/spring/docs/4.3.12.RELEASE/spring-framework-reference/html/beans.html#beans-factorybeans-annotations

> You may declare @Bean methods as static, allowing for them to be called without creating their containing configuration class as an instance. This makes particular sense when defining post-processor beans, e.g. of type BeanFactoryPostProcessor or BeanPostProcessor, since such beans will get initialized early in the container lifecycle and should avoid triggering other parts of the configuration at that point.

And that's exactly why we get access to the `Storage` bean in the `ProtocolResolver` from the `resolve()` method - as late as possible.

And that's why I didn't do `GcpStorageProperties` injection in its definition but just `Environment` and some magic via `RelaxedPropertyResolver`.

But as you wish, just let's consider an inheritance option!

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageProtocolResolverSettings.java
ChengyuanZhao: Sorry I forgot to explain. So without this settings class I could 

1. put GoogleStorageProtocolResolver(GSPR) as bean in StorageAutoConfiguration like you suggested, but at that point the bean method can't provide the beanFactory to the GSPR.
2. use GcpStorageProperties in GSPR, but we decided not to have the storage project depend on the starter-storage project. 
3. get the "auto-create-files" setting from Spring inside GSPR, but I thought it would be better to not have that name hard-coded

Of course please let me know an option I'm missing.

ChengyuanZhao: I just saw your explanation further down and thanks much for clearing it up for me; that's definitely something new I learned today!

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageProtocolResolver.java
```diff
@@ -36,21 +36,26 @@
...
+		implements ProtocolResolver, BeanFactoryPostProcessor, ResourceLoaderAware {
 
 	private static final Log logger = LogFactory.getLog(GoogleStorageProtocolResolver.class);
 
 	private static final String PROTOCOL = "gs://";
 
 	private ConfigurableListableBeanFactory beanFactory;
 
+	private GoogleStorageProtocolResolverSettings settings;
+	private boolean settingsAvailable = true;
```
artembilan: I guess we can have something like `settings = new GoogleStorageProtocolResolverSettings();` with all the default possible options.

That way we don't need to require from end-user to specify such a bean in his/her application.
Therefore no reason to warn about that bellow.
The `info()` is OK though with the note that we are going to use all the default settings.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageProtocolResolver.java
```diff
@@ -45,8 +45,9 @@
 
 	private ConfigurableListableBeanFactory beanFactory;
 
-	private GoogleStorageProtocolResolverSettings settings;
-	private boolean settingsAvailable = true;
+	private GoogleStorageProtocolResolverSettings settings
+			= new GoogleStorageProtocolResolverSettings();
+	private boolean checkedIfSettingsBeanAvailable;
```
artembilan: To be honest, my previous comment was exactly about the fact that we don't need extra variable.
We definitely need to load `storage` and `settings` beans only once. 
Since `storage` is mandatory, there is just enough to pull the `getBean(GoogleStorageProtocolResolverSettings.class)` just after `getBean(Storage.class)` in the same `if` block.

Does it make sense for you?

Also, please, consider to add blank lines around class properties.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageProtocolResolverContext.java
ChengyuanZhao: Done

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageProtocolResolverContext.java
ChengyuanZhao: Done

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResource.java
```diff
@@ -89,11 +103,25 @@ public URI getURI() throws IOException {
...
+		return this.storage.get(getBlobId());
+	}
+
+	private Blob createBlob() throws IOException {
+		return this.storage.create(BlobInfo.newBuilder(getBlobId()).build());
+	}
+
+	private Blob throwExceptionForNullBlob(Blob blob) throws IOException {
+		if (blob == null) {
+			throw new FileNotFoundException("The blob was not found and is null.");
```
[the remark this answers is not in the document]
ChengyuanZhao: Done

spring-cloud-gcp-starters/spring-cloud-gcp-starter-storage/src/main/java/org/springframework/cloud/gcp/storage/autoconfig/StorageAutoConfiguration.java
saturnism: hmm this is where we really ought to put properties and autoconfig into the same package as most of the other spring starters do.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageProtocolResolver.java
saturnism: can't they be fetched at the postProcessBeanFactory time during the lifecycle?

artembilan: No, they can't. It's too early and `BeanFactory` isn't ready yet to produce beans:
```
/**
	 * Modify the application context's internal bean factory after its standard
	 * initialization. All bean definitions will have been loaded, but no beans
	 * will have been instantiated yet. This allows for overriding or adding
	 * properties even to eager-initializing beans.
	 * @param beanFactory the bean factory used by the application context
	 * @throws org.springframework.beans.BeansException in case of errors
	 */
	void postProcessBeanFactory(ConfigurableListableBeanFactory beanFactory) throws BeansException;
```

The start up lifecycle of `ApplicationContext` is like this:
```
public void refresh() throws BeansException, IllegalStateException {
		synchronized (this.startupShutdownMonitor) {
			// Prepare this context for refreshing.
			prepareRefresh();

			// Tell the subclass to refresh the internal bean factory.
			ConfigurableListableBeanFactory beanFactory = obtainFreshBeanFactory();

			// Prepare the bean factory for use in this context.
			prepareBeanFactory(beanFactory);

			try {
				// Allows post-processing of the bean factory in context subclasses.
				postProcessBeanFactory(beanFactory);

				// Invoke factory processors registered as beans in the context.
				invokeBeanFactoryPostProcessors(beanFactory);

				// Register bean processors that intercept bean creation.
				registerBeanPostProcessors(beanFactory);

				// Initialize message source for this context.
				initMessageSource();

				// Initialize event multicaster for this context.
				initApplicationEventMulticaster();

				// Initialize other special beans in specific context subclasses.
				onRefresh();

				// Check for listener beans and register them.
				registerListeners();

				// Instantiate all remaining (non-lazy-init) singletons.
				finishBeanFactoryInitialization(beanFactory);

				// Last step: publish corresponding event.
				finishRefresh();
			}

			catch (BeansException ex) {
				if (logger.isWarnEnabled()) {
					logger.warn("Exception encountered during context initialization - " +
							"cancelling refresh attempt: " + ex);
				}

				// Destroy already created singletons to avoid dangling resources.
				destroyBeans();

				// Reset 'active' flag.
				cancelRefresh(ex);

				// Propagate exception to caller.
				throw ex;
			}

			finally {
				// Reset common introspection caches in Spring's core, since we
				// might not ever need metadata for singleton beans anymore...
				resetCommonCaches();
			}
		}
	}
```

All the beans are instantiated in the `finishBeanFactoryInitialization()` - already after any BF customization.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageProtocolResolver.java
```diff
@@ -36,21 +36,26 @@
...
+		implements ProtocolResolver, BeanFactoryPostProcessor, ResourceLoaderAware {
 
 	private static final Log logger = LogFactory.getLog(GoogleStorageProtocolResolver.class);
 
 	private static final String PROTOCOL = "gs://";
 
 	private ConfigurableListableBeanFactory beanFactory;
 
+	private GoogleStorageProtocolResolverSettings settings
+			= new GoogleStorageProtocolResolverSettings();
```
meltsufin: Why create the setting here?

artembilan: This is a *default* state. It's absolutely OK to have it during object initialization.
If end-user decides to have its own option (s)he is always free to declare such a bean in the application context. Exactly this happens when we use Spring Boot.

Not sure what is your concern: this default is here for those users who don't use Boot and don't want to override something.

meltsufin: OK, thanks for the clarification. I missed the fact that we don't fail when the settings bean doesn't exist.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageProtocolResolverSettings.java
joaoandremartins: Did we forget to delete this class?

artembilan: I think know. See here: https://github.com/spring-cloud/spring-cloud-gcp/blob/master/spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageProtocolResolver.java#L79.

This is the point of collaboration between end-user and the target `GoogleStorageProtocolResolver` which is very specific to be instantiated as a bean in the target end-user configuration.

joaoandremartins: Seems like it's required, but why didn't we simply add this field to `GcpStorageProperties` instead of having this extra class?

joaoandremartins: @artembilan it sounds like a `GcpStorageProperties` bean could be loaded instead, if I'm not missing anything.

artembilan: Right, you are missing the fact that `GcpStorageProperties` is in the `starter` module. That one which is Spring Boot aware already. But this *bridge* we need here in pure Store module for better end-developer experience.

joaoandremartins: If that's the case, then we should probably merge `GcpStorageProperties` and `GoogleStorageProtocolResolverSettings` when moving the properties classes to the main modules as part of #155.

artembilan: Well, your concern there about Spring Boot dependency is valid.
If there is anything changed since there, let me know.
But I'd definitely stick away from Spring Boot dependencies in the core modules.
That's not their responsibility to carry Boot feature. True.

If there is an idea to merge something together, I would suggest `auto-configure` module and indeed keep starters as just dependency management. The fact is that even Spring Boot per se has all the auto-configuration just only in one module: https://github.com/spring-projects/spring-boot/tree/master/spring-boot-project/spring-boot-autoconfigure

PR #187, opened 2017-10-06
Updates property name

joaoandremartins:


meltsufin:
I thought both versions were valid??

artembilan:
No-op, @meltsufin . That is for camel case properties, but not for nested like in our case: https://docs.spring.io/spring-boot/docs/1.5.7.RELEASE/reference/htmlsingle/#boot-features-external-config-relaxed-binding

meltsufin:
Ok, that makes sense. Thanks!

joaoandremartins:
When you specify credentials on a per-service bases, a dash is used. e.g., `spring.cloud.gcp.sql.credentials-location`.
But at the core level, it's `spring.cloud.gcp.credentials.location`.
This might become a source of confusion in the future...

meltsufin:
Indeed, I'd rather have `spring.cloud.gcp.sql.credentials.location`.

joaoandremartins:
The only way I see it could be solved is with an inner `Credentials` object like we do in `core`, but it would just store a `location` field here...

meltsufin:
Why not take `Credentials` class out of `GcpProperties` and reuse it here? Besides location developers may want to specify precise scopes as well.

joaoandremartins:
Unfortunately, `Credentials` only works as a nested class. We would need to have a `Credentials` nested class in every properties class. Since I'm not convinced that users would want to configure scopes for a service (we should just configure the right scope for them), I think we'll have to live with this slight incoherence...

artembilan:
Maybe we can have some super common class with that `Credentials` as nested one and make all other to extend that already ?

PR #190, opened 2017-10-09
Change rewriteId(Long) to rewriteId(long)

saturnism:
changed Long type to long since object type is not necessary.
also, if input value is 0, it should not rewrite ID and simply return 0 as well.


saturnism:
sorry, updated test to validate :) cheers,

PR #192, opened 2017-10-16
Allows for "credentials.location" notation in property names

joaoandremartins:
Creates an abstract CredentialsProperty class that is extended by
nested classes in each properties class.

### DIFF
--- a/spring-cloud-gcp-core/pom.xml
+++ b/spring-cloud-gcp-core/pom.xml
@@ -18,5 +18,9 @@
 			<groupId>com.google.cloud</groupId>
 			<artifactId>google-cloud-core</artifactId>
 		</dependency>
+		<dependency>
+			<groupId>org.springframework</groupId>
+			<artifactId>spring-core</artifactId>
+		</dependency>
 	</dependencies>
 </project>
--- a/spring-cloud-gcp-core/src/main/java/org/springframework/cloud/gcp/core/CredentialsProperty.java
+++ b/spring-cloud-gcp-core/src/main/java/org/springframework/cloud/gcp/core/CredentialsProperty.java
@@ -14,27 +14,35 @@
  *  limitations under the License.
  */
 
-package org.springframework.cloud.gcp.storage;
+package org.springframework.cloud.gcp.core;
 
-import org.springframework.boot.context.properties.ConfigurationProperties;
-import org.springframework.cloud.gcp.core.CredentialsProperty;
+import java.util.List;
+
+import org.springframework.core.io.Resource;
 
 /**
- * @author João André Martins
+ * To be extended by property class subclasses and allow e.g., a "credentials.location" notation.
+ *
+ * @author João Andre Martins
  */
-@ConfigurationProperties("spring.cloud.gcp.storage")
-public class GcpStorageProperties extends GoogleStorageProtocolResolverSettings {
+public abstract class CredentialsProperty {
 
-	private Credentials credentials;
+	private Resource location;
+	private List<String> scopes;
 
-	public Credentials getCredentials() {
-		return this.credentials;
+	public Resource getLocation() {
+		return this.location;
 	}
 
-	public void setCredentials(Credentials credentials) {
-		this.credentials = credentials;
+	public void setLocation(Resource location) {
+		this.location = location;
 	}
 
-	public class Credentials extends CredentialsProperty {
+	public List<String> getScopes() {
+		return this.scopes;
+	}
+
+	public void setScopes(List<String> scopes) {
+		this.scopes = scopes;
 	}
 }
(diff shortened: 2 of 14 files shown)


joaoandremartins:
Seeking a design opinion before making test and docs changes.

spring-cloud-gcp-core/src/main/java/org/springframework/cloud/gcp/core/CredentialsProperty.java
```diff
@@ -0,0 +1,48 @@
...
+import java.util.List;
+
+import org.springframework.core.io.Resource;
+
+/**
+ * To be extended by property class subclasses and allow e.g., a "credentials.location" notation.
+ *
+ * @author João Andre Martins
+ */
+public abstract class CredentialsProperty {
```
meltsufin: `AbstractCredentials` maybe for the name?

spring-cloud-gcp-core/pom.xml
```diff
@@ -18,5 +18,9 @@
 			<groupId>com.google.cloud</groupId>
 			<artifactId>google-cloud-core</artifactId>
 		</dependency>
+		<dependency>
+			<groupId>org.springframework</groupId>
+			<artifactId>spring-core</artifactId>
```
artembilan: Why do we need this explicitly?

I don't see any commit comment explaining the change.

Thanks

spring-cloud-gcp-core/src/main/java/org/springframework/cloud/gcp/core/CredentialsProperty.java
artembilan: Well, my idea was about something like:
```
@ConfigurationProperties("spring.cloud.gcp")
public class GcpProperties extends BaseGcpProperties {
```
Where that `BaseGcpProperties` will contain inner public `Credentials`.

WDYT?
Doesn't it work?

spring-cloud-gcp-core/pom.xml
joaoandremartins: Apologies, this is needed because the core module now uses `Resource`.

spring-cloud-gcp-core/src/main/java/org/springframework/cloud/gcp/core/CredentialsProperty.java
joaoandremartins: Renamed to `AbstractCredentialsProperty`. Kept `Property` in the name just so it isn't confused with a usable `Credentials` object that can be passed to an API.

spring-cloud-gcp-core/pom.xml
artembilan: Oh! I see now. Sorry. I thought that was a root pom 😄

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/src/main/java/org/springframework/cloud/gcp/sql/autoconfig/GcpCloudSqlAutoConfiguration.java
```diff
@@ -146,10 +145,10 @@ private void setCredentialsProperty() {
 		try {
 			// First tries the SQL configuration credential.
 			if (this.gcpCloudSqlProperties != null
-					&& this.gcpCloudSqlProperties.getCredentialsLocation() != null
-					&& Files.exists(this.gcpCloudSqlProperties.getCredentialsLocation())) {
+					&& this.gcpCloudSqlProperties.getCredentials() != null) {
 				System.setProperty(SqlCredentialFactory.CREDENTIAL_LOCATION_PROPERTY_NAME,
-						this.gcpCloudSqlProperties.getCredentialsLocation().toString());
+						this.gcpCloudSqlProperties.getCredentials().getLocation().getFile()
+								.getAbsolutePath());
```
artembilan: How about if `Resource` not file based?
```
public File getFile() throws IOException {
	throw new UnsupportedOperationException(
			getDescription() + " cannot be resolved to absolute file path");
}
```
for example.

Maybe we should have some `if...else` or `try...catch` and don't fail the whole application with the `FileNotFoundException`. Any way this looks like just a system property population...

joaoandremartins: `FileNotFoundException` is an `IOException`, there is already a catch block for those.

artembilan: Ah! Good. I see.
Well, IMO that is bad practice to have such a big `try...catch` block. It is pretty easy to lose what is really wrong.
If we are indeed fine with the failure on `FileNotFound` would be better to have that catch on particular peace of code, around `this.gcpCloudSqlProperties.getCredentials().getLocation().getFile()`.

More over I see we have a logic a bit below like:
```
// This should happen if the Spring resource isn't in the filesystem, but a URL,
 				// classpath file, etc.
 				if (credentialsLocationFile == null) {
 					if (LOGGER.isWarnEnabled()) {
 						LOGGER.warn("The private key of the Google Cloud SQL credential must "
 								+ "be in a file on the filesystem.");
}
```
That's great message for end-developer, but it isn't reachable in case of non-file `Resource`.

joaoandremartins: Moved things around. Let me know if you like it

PR #203, opened 2017-10-23
Set BOM parent to a RELEASE version

joaoandremartins:
We can rely on Spring Cloud to select spring-cloud version dependencies
for us.

### DIFF
--- a/spring-cloud-gcp-dependencies/pom.xml
+++ b/spring-cloud-gcp-dependencies/pom.xml
@@ -5,7 +5,7 @@
 	<parent>
 		<groupId>org.springframework.cloud</groupId>
 		<artifactId>spring-cloud-dependencies-parent</artifactId>
-		<version>1.3.3.BUILD-SNAPSHOT</version>
+		<version>1.3.5.RELEASE</version>
 		<relativePath/>
 	</parent>
 


spring-cloud-gcp-dependencies/pom.xml
```diff
@@ -21,14 +21,20 @@
...
-		<spring-cloud-context.version>1.2.3.RELEASE</spring-cloud-context.version>
-		<spring-cloud-sleuth.version>1.2.5.RELEASE</spring-cloud-sleuth.version>
+		<spring-cloud-dependencies.version>Dalston.SR4</spring-cloud-dependencies.version>
 	</properties>
 
 	<dependencyManagement>
 		<dependencies>
 			<dependency>
 				<groupId>org.springframework.cloud</groupId>
+				<artifactId>spring-cloud-dependencies</artifactId>
```
artembilan: Yeah, looking into the content of https://github.com/spring-cloud/spring-cloud-release/blob/master/spring-cloud-dependencies/pom.xml#L18[this BOM], I'm retracting my idea. Eventually our project will become as part of that as well, so, we definitely are not intended to introduce some circular dependencies.
I guess we don't have choice unless follow the model we have so far.

I guess aforementioned Sleuth project may be as sample how would we proceed: https://github.com/spring-cloud/spring-cloud-sleuth/blob/master/pom.xml. See how it uses `spring-cloud-stream` as dependency and what is important there is no any entry about `spring-cloud-stream` in its BOM: https://github.com/spring-cloud/spring-cloud-sleuth/blob/master/spring-cloud-sleuth-dependencies/pom.xml. That comes as transitive dependency of the particular `spring-cloud-sleuth-stream` module: https://github.com/spring-cloud/spring-cloud-sleuth/blob/master/spring-cloud-sleuth-stream/pom.xml.

We may reconsider that after Milestone 1 already any way.

WDYT?

joaoandremartins: Are you suggesting we depend on the `spring-cloud-stream-dependencies` BOM instead? IIUC that would mean the number of BOMs we'd depend on would roughly be 1:1 with the number of Spring (Cloud) sub-projects that we use.
So, if we're not going this way, I'm almost tempted to go the way we had before.

One thing I think we do need to do is to change the version of the parent from a SNAPSHOT to a RELEASE though. We could possibly get this in for M1, it's a simple change, I just need to change this PR to revert the other BOM stuff (if you agree).

artembilan: Right, that was my point : just no snapshots and that’s all. I just mean that we can boms for our does, but not include super bom as I suggested originaly . That was my mistake. Sorry for confuse.

PR #204, opened 2017-10-24
Spring Cloud Config reference document section

joaoandremartins:
Adds the missing Spring Cloud Config section to the Spring Cloud GCP
reference document.

### DIFF
--- a/spring-cloud-gcp-docs/src/main/asciidoc/config.adoc
+++ b/spring-cloud-gcp-docs/src/main/asciidoc/config.adoc
@@ -6,9 +6,9 @@ Runtime Configuration API] as a
 https://cloud.spring.io/spring-cloud-config/[Spring Cloud Config] server to remotely store your
 application configuration data.
 
-The Spring Cloud GCP Config support is provided via a Spring Boot starter that makes notation like
-`@Value("${gcp_runtime_config_api_var_name}")` populate variables with values stored in the
-Google Runtime Configuration API.
+The Spring Cloud GCP Config support is provided via a Spring Boot starter that enables the use of
+the `@Value` annotation for injecting values stored in the Google Runtime Configuration API into
+fields of Spring-managed beans.
 
 Maven coordinates, using Spring Cloud GCP BOM:
 
@@ -39,9 +39,9 @@ The following parameters are configurable in Spring Cloud GCP Config:
 | `spring.cloud.gcp.config.name` |
 Name of your application | No |
 | `spring.cloud.gcp.config.profile` |
-Configuration's Spring profile (e.g., prod) | Yes | default
+Configuration's Spring profile (e.g., prod) | Yes | `default`
 | `spring.cloud.gcp.config.timeout` | Timeout in milliseconds for connecting to the Google Runtime
-Configuration API | Yes | 60000
+Configuration API | Yes | `60000`
 | `spring.cloud.gcp.config.project-id` | GCP project ID where the Google Runtime Configuration API
 is hosted, if different from the one in the <<spring-cloud-gcp-core,Spring Cloud GCP Core Module>>
 | Yes |
@@ -80,7 +80,7 @@ gcloud beta runtime-config configs variables set queue_size 25 --config-name mya
 spring.cloud.gcp.config.name=myapp
 spring.cloud.gcp.config.profile=prod
 ----
-3. Add a field annotated with `@Value` to a POJO:
+3. Add a field annotated with `@Value` to a Spring-managed bean:
 +
 ----
 @Component
@@ -100,7 +100,7 @@ public class SampleConfig {
 ----
 
 When your Spring application starts, the `queueSize` field value will be set to 25 for any
-auto-wired instances of the above POJO.
+auto-wired instances of the above `SampleConfig` bean.
 
 === Refreshing the configuration at runtime
 
--- a/spring-cloud-gcp-docs/src/main/asciidoc/config.adoc
+++ b/spring-cloud-gcp-docs/src/main/asciidoc/config.adoc
@@ -80,15 +80,27 @@ gcloud beta runtime-config configs variables set queue_size 25 --config-name mya
 spring.cloud.gcp.config.name=myapp
 spring.cloud.gcp.config.profile=prod
 ----
-3. Add a `@Configuration` class that contains a field annotated with `@Value` in the following way:
+3. Add a field annotated with `@Value` to a POJO:
 +
 ----
-@Value("${queue_size}")
-private int queueSize;
+@Component
+public class SampleConfig {
+
+  @Value("${queue_size}")
+  private String queueSize;
+
+  public String getQueueSize() {
+    return this.queueSize;
+  }
+
+  public void setQueueSize(String queueSize) {
+    this.queueSize = queueSize;
+  }
+}
 ----
 
-When your Spring application starts, if an instance of the class that contains the `queueSize`
-field is auto-wired, the `queueSize` field value will be set to 25.
+When your Spring application starts, the `queueSize` field value will be set to 25 for any
+auto-wired instances of the above POJO.
 
 === Refreshing the configuration at runtime
 
--- a/spring-cloud-gcp-docs/src/main/asciidoc/config.adoc
+++ b/spring-cloud-gcp-docs/src/main/asciidoc/config.adoc
@@ -34,35 +34,37 @@ dependencies {
 The following parameters are configurable in Spring Cloud GCP Config:
 
 |===
-| Name | Description | Default value
-| `spring.cloud.gcp.config.enabled` | Enables the Config client | `true`
+| Name | Description | Optional | Default value
+| `spring.cloud.gcp.config.enabled` | Enables the Config client | Yes | `true`
 | `spring.cloud.gcp.config.name` |
 http://cloud.spring.io/spring-cloud-static/spring-cloud-config/1.3.3.RELEASE/single/spring-cloud-config.html#_locating_remote_configuration_resources[Name
-of your application] |
+of your application] | No |
 | `spring.cloud.gcp.config.profile` |
 http://cloud.spring.io/spring-cloud-static/spring-cloud-config/1.3.3.RELEASE/single/spring-cloud-config.html#_locating_remote_configuration_resources[Configuration's
-Spring profile] (e.g., prod) | default
+Spring profile] (e.g., prod) | Yes | default
 | `spring.cloud.gcp.config.timeout` | Timeout in milliseconds for connecting to the Google Runtime
-Configuration API | 60000
+Configuration API | Yes | 60000
 | `spring.cloud.gcp.config.project-id` | GCP project ID where the Google Runtime Configuration API
-is hosted, if different from the one in the <<spring-cloud-gcp-core,Spring Cloud GCP Core Module>> |
+is hosted, if different from the one in the <<spring-cloud-gcp-core,Spring Cloud GCP Core Module>>
+| Yes |
 | `spring.cloud.gcp.config.credentials.location` | OAuth2 credentials for authenticating with the
 Google Runtime Configuration API, if different from the ones in the
-<<spring-cloud-gcp-core,Spring Cloud GCP Core Module>> |
+<<spring-cloud-gcp-core,Spring Cloud GCP Core Module>> | Yes |
 | `spring.cloud.gcp.config.credentials.scopes` |
 https://developers.google.com/identity/protocols/googlescopes[OAuth2 scope] for Spring Cloud GCP
-Config credentials | https://www.googleapis.com/auth/cloudruntimeconfig
+Config credentials | Yes | https://www.googleapis.com/auth/cloudruntimeconfig
 |===
 
-NOTE: These properties should be specified in a `bootstrap.yml`/`bootstrap.properties` file, rather
-than the usual `applications.yml`/`application.properties`.
+NOTE: These properties should be specified in a
+http://cloud.spring.io/spring-cloud-static/spring-cloud.html#_the_bootstrap_application_context[`bootstrap.yml`/`bootstrap.properties`]
+file, rather than the usual `applications.yml`/`application.properties`.
 
 === Quick start
 
 1. Create a configuration in the Google Runtime Configuration API that is called
-`spring.cloud.gcp.config.name`_`spring.cloud.gcp.config.profile`. In other words, if
-`spring.cloud.gcp.config.name` is `myapp` and `spring.cloud.gcp.config.profile` is prod, the
-configuration should be called `myapp_prod`.
+`${spring.cloud.gcp.config.name}_${spring.cloud.gcp.config.profile}`.
+In other words, if `spring.cloud.gcp.config.name` is `myapp` and `spring.cloud.gcp.config.profile`
+is prod, the configuration should be called `myapp_prod`.
 +
 In order to do that, you should have the
 https://cloud.google.com/sdk/[Google Cloud SDK] installed, own a Google Cloud Project and run the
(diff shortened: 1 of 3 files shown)


joaoandremartins:
@jabubake FYI

joaoandremartins:
All -- this is likely the last missing piece for M1, a timely review is appreciated so M1 can be launched asap :)

spring-cloud-gcp-docs/src/main/asciidoc/config.adoc
```diff
@@ -0,0 +1,101 @@
+== Spring Cloud Config
+
+Spring Cloud GCP makes it possible to use the
+https://cloud.google.com/deployment-manager/runtime-configurator/reference/rest/[Google
+Runtime Configuration API] as a
+https://cloud.spring.io/spring-cloud-config/[Spring Cloud Config] server to remotely store your
+application configuration data.
+
+The Spring Cloud GCP Config support is provided via a Spring Boot starter that makes notation like
```
jabubake: Config support => Config client support

spring-cloud-gcp-docs/src/main/asciidoc/config.adoc
```diff
@@ -0,0 +1,101 @@
...
+is hosted, if different from the one in the <<spring-cloud-gcp-core,Spring Cloud GCP Core Module>> |
+| `spring.cloud.gcp.config.credentials.location` | OAuth2 credentials for authenticating with the
+Google Runtime Configuration API, if different from the ones in the
+<<spring-cloud-gcp-core,Spring Cloud GCP Core Module>> |
+| `spring.cloud.gcp.config.credentials.scopes` |
+https://developers.google.com/identity/protocols/googlescopes[OAuth2 scope] for Spring Cloud GCP
+Config credentials | https://www.googleapis.com/auth/cloudruntimeconfig
+|===
+
+NOTE: These properties should be specified in a `bootstrap.yml` file, rather than the usual
```
jabubake: should be we state `boostrap.yml` / `bootstrap.properties`.

spring-cloud-gcp-docs/src/main/asciidoc/config.adoc
```diff
@@ -0,0 +1,101 @@
...
+`applications.yml`
+
+=== Quick start
+
+1. Create a configuration in the Google Runtime Configuration API that is called
```
jabubake: Thank you for  clearly explaining this :)

spring-cloud-gcp-docs/src/main/asciidoc/config.adoc
```diff
@@ -0,0 +1,101 @@
...
+`spring.cloud.gcp.config.name`_`spring.cloud.gcp.config.profile`. In other words, if
+`spring.cloud.gcp.config.name` is `myapp` and `spring.cloud.gcp.config.profile` is prod, the
+configuration should be called `myapp_prod`.
++
+In order to do that, you should have the
+https://cloud.google.com/sdk/[Google Cloud SDK] installed, own a Google Cloud Project and run the
+following command:
++
+----
+gcloud beta runtime-config configs create myapp_prod
```
jabubake: `gcloud init` as well ?

spring-cloud-gcp-docs/src/main/asciidoc/config.adoc
```diff
@@ -0,0 +1,101 @@
...
+@Value("${queue_size}")
+private int queueSize;
+----
+
+When your Spring application starts, if an instance of the class that contains the `queueSize`
+field is auto-wired, the `queueSize` field value will be set to 25.
+
+=== Refreshing the configuration at runtime
+
+Using http://cloud.spring.io/spring-cloud-static/docs/1.0.x/spring-cloud.html#_endpoints[Spring
```
jabubake: remove "Using"

spring-cloud-gcp-docs/src/main/asciidoc/config.adoc
```diff
@@ -0,0 +1,101 @@
...
+Boot Actuator] enables a `/refresh` endpoint in your application that can be used to refresh the
+values of all the Spring Config-managed variables.
+
+To achieve that, add a `@RefreshScope` annotation to your `@Configuration` class(es).
```
jabubake: s/achieve that/achieve this/

jabubake: I think @meltsufin noted that it does n't have to be a `@Configuration` annotation and can be any auto-wired bean ?

spring-cloud-gcp-docs/src/main/asciidoc/config.adoc
```diff
@@ -0,0 +1,101 @@
...
+Then, if you change the value of `myapp_prod`'s `queue_size` variable in the Google Runtime
+Configuration API and then hit the `/refresh` endpoint of your application, you can verify that the
+value of the `queueSize` field is now the changed value.
```
jabubake: s/is now the changed value/has been updated/

spring-cloud-gcp-docs/src/main/asciidoc/config.adoc
joaoandremartins: I omitted the "client" bit on purpose because I don't want readers to think they can use the generic Spring Cloud Config client to connect to the Google Runtime Config API. The Spring Cloud Config client actually can't be used to get config information from us (which is why it was a good idea in the end to not re-use the Spring Cloud Config property namespace `spring.cloud.config.*`).

spring-cloud-gcp-docs/src/main/asciidoc/config.adoc
joaoandremartins: I'm afraid that omitting "Using" will cause it to sound like Actuator is enabled by default?

spring-cloud-gcp-docs/src/main/asciidoc/config.adoc
joaoandremartins: Ah, touché. I don't know where I read that it had to be a `@Configuration`...

joaoandremartins: Actually, if it isn't an `@Configuration`, the class has to be annotated with something, otherwise it doesn't get picked up... The thing is there might be annotations other than `@Configuration` that also make it possible to auto-wire the class (and not include the `@Configuration` annotation themselves).
I'll reword things here a bit, even though above I say that `@Configuration` should be added, and that should probably stay that way.

artembilan: Some answers should be here: http://cloud.spring.io/spring-cloud-static/Dalston.SR4/multi/multi__spring_cloud_context_application_context_services.html#_refresh_scope

> A Spring `@Bean` that is marked as `@RefreshScope` will get special treatment when there is a configuration change. 

And there is definitely an IMPORTANT not to be cautious with the `@Configuration` classes

joaoandremartins: In my toy app, I'm using a class such as the following one to auto-wire later:

```
@Configuration
public class SampleConfig {

  @Value("${url}")
  private String url;

  public String getUrl() {
    return url;
  }

  public void setUrl(String url) {
    this.url = url;
  }
}
```

Can you recommend an annotation to use other than `@Configuration` that makes beans of that kind auto-wireable?

markfisher: the most generic option is `@Component`

artembilan: ... or nothing on the class and in some `@Configuration` have it like:
```
@Bean
@RefreshScope
SampleConfig myConfig() {
   return new SampleConfig();
}
```

spring-cloud-gcp-docs/src/main/asciidoc/config.adoc
```diff
@@ -0,0 +1,115 @@
...
+[source,subs="normal"]
+----
+dependencies {
+    compile group: 'org.springframework.cloud', name: 'spring-cloud-gcp-starter-config', version: '{project-version}'
+}
+----
+
+=== Configuration
+
+The following parameters are configurable in Spring Cloud GCP Config:
```
meltsufin: Which ones are optional and which required? Also, maybe state up here that they should go into `boostrap.properties` and link to what that is.

spring-cloud-gcp-docs/src/main/asciidoc/config.adoc
```diff
@@ -0,0 +1,115 @@
...
+Config credentials | https://www.googleapis.com/auth/cloudruntimeconfig
+|===
+
+NOTE: These properties should be specified in a `bootstrap.yml`/`bootstrap.properties` file, rather
+than the usual `applications.yml`/`application.properties`.
+
+=== Quick start
+
+1. Create a configuration in the Google Runtime Configuration API that is called
+`spring.cloud.gcp.config.name`_`spring.cloud.gcp.config.profile`. In other words, if
```
meltsufin: This is a bit confusing. Maybe `${spring.cloud.gcp.config.name}_${spring.coud.gcp.config.profile}` for better clarity?

spring-cloud-gcp-docs/src/main/asciidoc/config.adoc
jabubake: I think in that case, it might best to rephrase the server part as "Configuration server" and not imply that it may be Git backed.

joaoandremartins: Because we don't actually use Spring Cloud Config for anything (i.e., we basically re-implement it), I wonder if it's worth rewording to say something like "storing configuration data in a similar way to Spring Cloud Config"?
I'm also not comfortable with the Spring Cloud Config server bit.

spring-cloud-gcp-docs/src/main/asciidoc/config.adoc
joaoandremartins: Added an Optional column to the table and added a hyperlink to `bootstrap.properties`.

joaoandremartins:
I think I resolved every comment, but let me know if there is anything I missed.
Thanks!

spring-cloud-gcp-docs/src/main/asciidoc/config.adoc
```diff
@@ -0,0 +1,117 @@
...
+----
+
+=== Configuration
+
+The following parameters are configurable in Spring Cloud GCP Config:
+
+|===
+| Name | Description | Optional | Default value
+| `spring.cloud.gcp.config.enabled` | Enables the Config client | Yes | `true`
+| `spring.cloud.gcp.config.name` |
```
meltsufin: Is it config name or application name? It's confusing because you're linking to docs about `spring.application.name`. Should we just use `spring.application.name` or at least default to it?

joaoandremartins: They're not related. It's suggested that this is set to `spring.application.name`, however, that's merely a suggestion, users can name the config anything they want.

meltsufin: I think we can omit suggesting any connection between this and `spring.application.name` and also not link to http://cloud.spring.io/spring-cloud-static/spring-cloud-config/1.3.3.RELEASE/single/spring-cloud-config.html#_locating_remote_configuration_resources.

spring-cloud-gcp-docs/src/main/asciidoc/config.adoc
```diff
@@ -0,0 +1,115 @@
...
+Name of your application | No |
+| `spring.cloud.gcp.config.profile` |
+Configuration's Spring profile (e.g., prod) | Yes | default
```
meltsufin: default -> `default`

spring-cloud-gcp-docs/src/main/asciidoc/config.adoc
```diff
@@ -0,0 +1,115 @@
...
+| `spring.cloud.gcp.config.timeout` | Timeout in milliseconds for connecting to the Google Runtime
+Configuration API | Yes | 60000
```
meltsufin: 60000 -> `60000`

spring-cloud-gcp-docs/src/main/asciidoc/config.adoc
```diff
@@ -0,0 +1,115 @@
+== Spring Cloud Config
+
+Spring Cloud GCP makes it possible to use the
+https://cloud.google.com/deployment-manager/runtime-configurator/reference/rest/[Google
+Runtime Configuration API] as a
+https://cloud.spring.io/spring-cloud-config/[Spring Cloud Config] server to remotely store your
+application configuration data.
+
+The Spring Cloud GCP Config support is provided via a Spring Boot starter that makes notation like
```
meltsufin: I'm not a fan of using the word "notation" here.
Here's how I would rephrase:

The Spring Cloud GCP Config support is provided via a Spring Boot starter that enables the use of the `@Value` annotation for injecting values stored in the Google Runtime Configuration API into fields of Spring-managed beans.

spring-cloud-gcp-docs/src/main/asciidoc/config.adoc
```diff
@@ -0,0 +1,115 @@
...
+| Yes |
+| `spring.cloud.gcp.config.credentials.location` | OAuth2 credentials for authenticating with the
+Google Runtime Configuration API, if different from the ones in the
+<<spring-cloud-gcp-core,Spring Cloud GCP Core Module>> | Yes |
+| `spring.cloud.gcp.config.credentials.scopes` |
+https://developers.google.com/identity/protocols/googlescopes[OAuth2 scope] for Spring Cloud GCP
+Config credentials | Yes | https://www.googleapis.com/auth/cloudruntimeconfig
+|===
+
+NOTE: These properties should be specified in a
```
meltsufin: Link to some explanation of why or at least what bootstrap configuration is?

spring-cloud-gcp-docs/src/main/asciidoc/config.adoc
```diff
@@ -0,0 +1,115 @@
...
+file, rather than the usual `applications.yml`/`application.properties`.
+
+=== Quick start
+
+1. Create a configuration in the Google Runtime Configuration API that is called
+`${spring.cloud.gcp.config.name}_${spring.cloud.gcp.config.profile}`.
+In other words, if `spring.cloud.gcp.config.name` is `myapp` and `spring.cloud.gcp.config.profile`
+is prod, the configuration should be called `myapp_prod`.
++
+In order to do that, you should have the
```
meltsufin: In order to do *what*?

spring-cloud-gcp-docs/src/main/asciidoc/config.adoc
```diff
@@ -0,0 +1,115 @@
...
+gcloud beta runtime-config configs variables set queue_size 25 --config-name myapp_prod
+----
+
+2. Configure your bootstrap.yml file with your application's configuration data:
++
+----
+spring.cloud.gcp.config.name=myapp
+spring.cloud.gcp.config.profile=prod
+----
+3. Add a field annotated with `@Value` to a POJO:
```
meltsufin: Why POJO? I think it works with any Spring-managed bean.

spring-cloud-gcp-docs/src/main/asciidoc/config.adoc
joaoandremartins: There is already a link that explains bootstrap configuration pretty well IMO.
Do you think more explanation is due here?

spring-cloud-gcp-docs/src/main/asciidoc/config.adoc
joaoandremartins: To create the configuration.

spring-cloud-gcp-docs/src/main/asciidoc/config.adoc
meltsufin: Sorry, I think I just missed it.

PR #221, opened 2017-10-30
Adds Snapshot repository info to reference document

### LINKED ISSUE #215: Retouch Storage refdoc section
- Mention dependency on Pivotal Milestones repo.

joaoandremartins:
Fixes #215

### DIFF
--- a/spring-cloud-gcp-docs/src/main/asciidoc/config.adoc
+++ b/spring-cloud-gcp-docs/src/main/asciidoc/config.adoc
@@ -29,6 +29,9 @@ dependencies {
 }
 ----
 
+You also need to add Pivotal's Snapshot repository per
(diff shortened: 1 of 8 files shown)


artembilan:
What is the point to mention all that info in the Reference Manual?
I wonder how you have time at all to write this here for time being and then revert when we go to GA...

Seems for me this is redundant work and extra unnecessary noise.

The interesting fact when we go to the project page (https://cloud.spring.io/spring-cloud-gcp/), there is a section about "Quick Start" and when we select there BUILD-SNAPSHOT, it really adds the `https://repo.spring.io/libs-snapshot` to the snippet.

We can add some note though into our `Dependency Management` section, but that has to be some generic phrase without an intention to be removed in the future.

Thank you for understanding

PR #222, opened 2017-10-30
Polishes spring-cloud-gcp-config-example

### LINKED ISSUE #214: Retouch spring-cloud-gcp-config-example README
- Mention code should be checked out
- Link to available scopes in [here](https://github.com/spring-cloud/spring-cloud-gcp/blob/master/spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/README.adoc)
- Resolve the another set of instructions problem on step one, including removing duplicate instructions
- Give more details around what/how the created service account should be
- Explain why the variables are being set
- Instead of asking users to add `@Value`, just mention it's there and it's what does the trick
- Explain that the credentials location resource must have a "file:" before it (might already be there), otherwise Spring looks for the value in the classpath
- Maybe call out for the fact that if a 404 is being returned, the config name and properties maybe mismatch?

joaoandremartins:
Clarifies a bunch of things.

Fixes #214

### DIFF
--- a/spring-cloud-gcp-examples/spring-cloud-gcp-config-example/README.adoc
+++ b/spring-cloud-gcp-examples/spring-cloud-gcp-config-example/README.adoc
@@ -4,15 +4,17 @@ This sample application demonstrates using the
 link:../../spring-cloud-gcp-starters/spring-cloud-gcp-starter-config[Spring Cloud GCP Config Client Starter] in your code.
 
 == Setup & Configuration
-1. Follow link:../../spring-cloud-gcp-starters/spring-cloud-gcp-starter-config[these setup instructions].
-2.  Create a configuration using the
+1. Checkout out the code of this sample.
+2. Setup the Google Runtime Configuration API with
+link:../../spring-cloud-gcp-starters/spring-cloud-gcp-starter-config#setup[these instructions].
+3. Create a configuration using the
 https://cloud.google.com/sdk/[Google Cloud SDK].
 +
 ....
 gcloud beta runtime-config configs create myapp_prod
 ....
 +
-Then set the variables:
+Then set the variables that your application will load:
 +
 ....
   gcloud beta runtime-config configs variables set \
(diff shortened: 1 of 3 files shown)


spring-cloud-gcp-examples/spring-cloud-gcp-config-example/README.adoc
```diff
@@ -4,15 +4,17 @@ This sample application demonstrates using the
 link:../../spring-cloud-gcp-starters/spring-cloud-gcp-starter-config[Spring Cloud GCP Config Client Starter] in your code.
 
 == Setup & Configuration
-1. Follow link:../../spring-cloud-gcp-starters/spring-cloud-gcp-starter-config[these setup instructions].
-2.  Create a configuration using the
+1. Checkout out the code of this sample.
```
meltsufin: minus "out", just "Checkout the code"

spring-cloud-gcp-examples/spring-cloud-gcp-config-example/README.adoc
```diff
@@ -4,15 +4,17 @@ This sample application demonstrates using the
 link:../../spring-cloud-gcp-starters/spring-cloud-gcp-starter-config[Spring Cloud GCP Config Client Starter] in your code.
 
 == Setup & Configuration
-1. Follow link:../../spring-cloud-gcp-starters/spring-cloud-gcp-starter-config[these setup instructions].
-2.  Create a configuration using the
+1. Checkout the code of this sample.
+2. Setup the Google Runtime Configuration API with
+link:../../spring-cloud-gcp-starters/spring-cloud-gcp-starter-config#setup[these instructions].
```
artembilan: Why don't use *Internal links* support in ASCII Doc?

joaoandremartins: If you mean the `<<>>` links, I don't think they can be used because the links would be in another document. Let me know if I'm missing anything.

artembilan: ??? We have one big HTML or PDF in the end. What is the "another document" are you referring to?

joaoandremartins:
Let me know if there are any more questions. If not, please approve.

PR #223, opened 2017-10-30
Polishes spring-cloud-gcp-integration-pubsub-example

### LINKED ISSUE #213: Retouch spring-cloud-gcp-integration-pubsub-example
- Reword line 1) so that settings are auto-discovered if Cloud SDK is installed and initiated and mention that further configuration can be achieved via the Spring Cloud GCP Core starter.
- Add dummy application.properties files to both samples.
- s/Google Cloud Console/Google Cloud Console Pub/Sub page
- Explain where users need to click to create topic and subscriptions
- Mention that examples are runnable with `mvn spring-boot:run` and that there might have to be more than one terminal window.
- Make sync call and log something if message was sent.

joaoandremartins:
Makes instructions clearer.
Tells whether a message was delivered or not.

Fixes #213


spring-cloud-gcp-examples/spring-cloud-gcp-integration-pubsub-example/README.adoc
```diff
@@ -8,13 +8,27 @@ The other app subscribes to that topic and logs received messages.
 == Setup & Configuration
 
 1. Configure your GCP project ID and credentials by following
-link:../../spring-cloud-gcp-starters/spring-cloud-gcp-starter-core[these instructions], or just let
-Spring discover those parameters for you.
-2. Go to the https://console.cloud.google.com/cloudpubsub/topicList[Google Cloud Console] and create
-a topic called `exampleTopic` and a subscription for that topic called `exampleSubscription`.
-3. Start the `spring-cloud-gcp-spring-integration-pubsub-example-sender` and
-`spring-cloud-gcp-spring-integration-pubsub-example-receiver` applications.
+link:../../spring-cloud-gcp-starters/spring-cloud-gcp-starter-core[these instructions].
```
artembilan: Why don't use Internal links support in ASCII Doc?

joaoandremartins: If you're referring to `<<>>`, it won't work because we're linking to outside this document.
It does work in the reference document because everything is compiled to the same final document.

artembilan: OK. I give up. I don't understand you, sorry.
What is the difference in your terms between `document` and ` reference document`?

In the same Spring Kafka we have several files for the documentation and when we build the final HTML or PDF those *internal links* are compiled properly to the actual links based on the actual version of the project.

joaoandremartins: This is not included in the reference document, this is the README of the SI example, to be read from GitHub mostly.

artembilan: Ah! Got it! 
Sorry for my stupidity 😄 

I need to do something with my mind. It lets me down last time...

So, then your solution is OK.

LGTM

PR #224, opened 2017-10-30
Polishes the Stackdriver Trace example

### LINKED ISSUE #212: Retouch spring-cloud-gcp-trace-example README
- Default application credentials are sufficient. Might mention that service account credentials can also be used (even though this creates a dependency in the Cloud SDK).
- Add link to reference document.
- Either reduce the trace transmission delay to 1 second or mention that the 10 second default might cause the traces not to be immediately visible in the Cloud Console.

joaoandremartins:
Fixes #212

### DIFF
--- a/spring-cloud-gcp-examples/spring-cloud-gcp-trace-example/README.adoc
+++ b/spring-cloud-gcp-examples/spring-cloud-gcp-trace-example/README.adoc
@@ -3,23 +3,41 @@
 This sample application demonstrates using the
 link:../../spring-cloud-gcp-starters/spring-cloud-gcp-starter-trace[Spring Cloud GCP Trace Starter] in your code.
 
+You can read about the Spring Boot Starter for Stackdriver Trace in more detail at the
+https://docs.spring.io/spring-cloud-gcp/docs/1.0.0.BUILD-SNAPSHOT/reference/htmlsingle/#_spring_cloud_sleuth[Spring
+Cloud GCP Reference Document Stackdriver Trace section].
+
 == Setup & Configuration
 1. Create a Google Cloud Platform Project
 1. https://cloud.google.com/docs/authentication/getting-started#creating_the_service_account[Create a service account]
-   with Trace Append permission. Furnish a new JSON key and then set the credentials using the
-   `GOOGLE_APPLICATION_CREDENTIALS` environment variable or
-   link:../../spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/README.adoc[using GCP Starter Core properties].
+with Trace Append permission. Furnish a new JSON key and then set the credentials using the
+`GOOGLE_APPLICATION_CREDENTIALS` environment variable or
+link:../../spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/README.adoc[using GCP Starter Core properties].
++
+Alternatively, if you have the https://cloud.google.com/sdk/[Google Cloud SDK] installed and
+initialized and are logged in with
+https://developers.google.com/identity/protocols/application-default-credentials[application
+default credentials], you can skip this step since Spring Cloud GCP will auto-discover those
+settings for you.
+
 1. Enable the https://console.cloud.google.com/apis/api/cloudtrace.googleapis.com/overview[Stackdriver Trace API]
 
 == Run the Example
 Run the example from Maven:
+
 ----
 mvn spring-boot:run
 ----
 
 Browse to `http://localhost:8080`. This will trigger the `ExampleController.work()` method,
-which in turn will call other services, and also a remote RESTful call to `ExampleController.meet()` method.
+which in turn will call other services, and also a remote RESTful call to `ExampleController.meet()`
+method.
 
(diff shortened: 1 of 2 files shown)


spring-cloud-gcp-examples/spring-cloud-gcp-trace-example/README.adoc
```diff
@@ -3,23 +3,41 @@
...
+
 == Setup & Configuration
 1. Create a Google Cloud Platform Project
 1. https://cloud.google.com/docs/authentication/getting-started#creating_the_service_account[Create a service account]
-   with Trace Append permission. Furnish a new JSON key and then set the credentials using the
-   `GOOGLE_APPLICATION_CREDENTIALS` environment variable or
-   link:../../spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/README.adoc[using GCP Starter Core properties].
+with Trace Append permission. Furnish a new JSON key and then set the credentials using the
+`GOOGLE_APPLICATION_CREDENTIALS` environment variable or
+link:../../spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/README.adoc[using GCP Starter Core properties].
```
artembilan: Hm. There is some inconsistency with links.
The previous one is from "https://docs.spring.io/spring-cloud-gcp/docs/1.0.0.BUILD-SNAPSHOT/" - absolute value from the Internet. This one is relative to the current document state in the project. But both of them are related to the same SC-GCP document. 

Am I missing anything? :shrug:

OTOH in other projects we definitely use [ASCII Doc](http://asciidoctor.org/docs/asciidoc-syntax-quick-reference/#links) (section `Internal cross references`) style for linking. For example Spring Kafka: https://raw.githubusercontent.com/spring-projects/spring-kafka/master/src/reference/asciidoc/whats-new.adoc

Pay attention to the style like: `See <<message-listeners>> for more information.`.

spring-cloud-gcp-examples/spring-cloud-gcp-trace-example/README.adoc
```diff
@@ -3,23 +3,41 @@
...
+which in turn will call other services, and also a remote RESTful call to `ExampleController.meet()`
+method.
+
+To see the traces, navigate to Stackdriver Trace console's
+https://console.cloud.google.com/traces/traces[Trace List] view.
+You should see the trace information in detail.
 
-To see the traces, navigate to Stackdriver Trace console's https://console.cloud.google.com/traces/traces[Trace List]
-view. You should see the trace information in detail.
+Bear in mind the trace transmission delay default value is 10 seconds, so it can take a little
```
meltsufin: "Bear in mind" -> "Note that"

spring-cloud-gcp-examples/spring-cloud-gcp-trace-example/README.adoc
```diff
@@ -3,23 +3,41 @@
...
+bit to see the traces in the Trace List page.
```
meltsufin: I would rephrase to:
"while for the traces show up in the Trace List page"

spring-cloud-gcp-examples/spring-cloud-gcp-trace-example/README.adoc
```diff
@@ -3,23 +3,41 @@
 This sample application demonstrates using the
 link:../../spring-cloud-gcp-starters/spring-cloud-gcp-starter-trace[Spring Cloud GCP Trace Starter] in your code.
 
+You can read about the Spring Boot Starter for Stackdriver Trace in more detail at the
+https://docs.spring.io/spring-cloud-gcp/docs/1.0.0.BUILD-SNAPSHOT/reference/htmlsingle/#_spring_cloud_sleuth[Spring
```
meltsufin: This link will get outdated quickly. Is there some kind "latest" URL?

artembilan: Right, I have the same concern. See my comment.
That is fully unclear why to use external links if we can compile internals...

joaoandremartins: Unfortunately, there is no alternative.
I don't know how we publish the reference document in order to generate a separate "latest" version and it doesn't exist currently.

We can't use internal links because we're linking from a README file to the reference document.

spring-cloud-gcp-examples/spring-cloud-gcp-trace-example/README.adoc
joaoandremartins: We're linking from a README to the reference document, can't use internal links.

spring-cloud-gcp-examples/spring-cloud-gcp-trace-example/README.adoc
artembilan: How about just refer from this README to the target .adoc relatively instead?

spring-cloud-gcp-examples/spring-cloud-gcp-trace-example/README.adoc
artembilan: OK. I got it. Thanks.
Never mind

PR #228, opened 2017-11-02
Changes the root pom's parent version to non-SNAPSHOT

joaoandremartins:
Tested change and saw no issues.


joaoandremartins:
Thanks for the quick approval.
No other situations that I could find!

PR #229, opened 2017-11-02
Update to release version: 1.0.0.M1

sobychacko:


spring-cloud-gcp-examples/pom.xml
```diff
@@ -5,7 +5,7 @@
 	<parent>
 		<artifactId>spring-cloud-gcp</artifactId>
 		<groupId>org.springframework.cloud</groupId>
-		<version>1.0.0.BUILD-SNAPSHOT</version>
+		<version>1.0.0.M1</version>
```
artembilan: I think we are not going to release `examples`. They must be excluded from the release artifacts at all.

sobychacko: It might be better to pull the module completely from this repo and may be put under a new samples repo or something. It will be inconvenient from a maven angle to exclude just this module for releasing purposes.

artembilan: Well, we have to figure out the way. It is really *convenient* just to exclude this module.
I think we even can build all the jars, but don't deploy them into `/milestone` repo. 
Modify the release script somehow to exclude.

How about this
http://prystash.blogspot.com/2009/06/maven-excluding-module-from-deploy.html 
?

joaoandremartins: Is it that inconvenient to release the examples as well?

If it is, how about just removing the `spring-cloud-gcp-examples` module from the parent pom?

artembilan: > Is it that inconvenient to release the examples as well?

What is the point of those compiled artifacts?
How should end-users get again of the dependency from them?
:shrug:
Let's don't confuse community!

joaoandremartins: Makes sense.
In that case, we could just remove `<module>spring-cloud-gcp-examples</modules>` from the top pom.xml in this PR?

artembilan: Let's give it a try!

@sobychacko , would you mind fixing accordingly and run some `dryRun`  release to be sure that we don't deploy `samples` artifacts?

Thanks

PR #230, opened 2017-11-02
Next update version: 1.0.0.BUILD-SNAPSHOT

### LINKED ISSUE #170: Release 1.0.0.M1
A tracking issue to capture MVP requirements 1.0 M1 release and to pushing M1 to Maven Central.

Current list: https://github.com/spring-cloud/spring-cloud-gcp/milestone/2

sobychacko:
Resolves #170

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -11,7 +11,7 @@
 	</parent>
 
 	<artifactId>spring-cloud-gcp</artifactId>
-	<version>1.0.0.M1</version>
+	<version>1.0.0.BUILD-SNAPSHOT</version>
 	<packaging>pom</packaging>
 	<name>Spring Cloud GCP</name>
 	<description>Spring Cloud GCP</description>
--- a/pom.xml
+++ b/pom.xml
@@ -32,6 +32,7 @@
 		<module>spring-cloud-gcp-starters</module>
 		<module>spring-integration-gcp</module>
 		<module>spring-cloud-gcp-docs</module>
+		<module>spring-cloud-gcp-examples</module>
 	</modules>
 
 	<properties>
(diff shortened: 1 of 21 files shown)


pom.xml
```diff
@@ -11,7 +11,7 @@
 	</parent>
 
 	<artifactId>spring-cloud-gcp</artifactId>
-	<version>1.0.0.M1</version>
+	<version>1.0.0.BUILD-SNAPSHOT</version>
```
joaoandremartins: Can we also add back spring-cloud-gcp-examples here?

artembilan: I would prefer "No". IMHO.

I really don't like this manual tasks during release process. We need to have some automatic tool, indeed.

sobychacko: Updated the PR with the samples module. Next time around, we will make the changes in the build script to not build the samples module so that we don't have to do this manual process.

artembilan: Sounds good @sobychacko !

The same is applied for the `docs` module as well.
See my comment: https://github.com/spring-cloud/spring-cloud-gcp/issues/170#issuecomment-341553552

Also need to figure out what to do with JavaDocs. But that's already different story and has to be addressed in the separate issue/PR.

joaoandremartins:
@sobychacko it looks like the parent version also needs to be changed in each example from `1.0.0.M1` to `1.0.0.BUILD-SNAPSHOT`.

Thanks!

joaoandremartins:
@sobychacko sorry for the haste, but do you think we could merge this PR today?

Would be nice to leave the code freeze as soon as we can! Thanks

sobychacko:
@joaoandremartins updated the PR. Feel free to merge it. Code freeze can be lifted then.

PR #231, opened 2017-11-03
gh-85 add tests

Andy2003:
I added the test for the binder. There is still one Test failing


joaoandremartins:
Thanks a bunch, @Andy2003 !

PR #232, opened 2017-11-03
Bumps to version 1.1.0.BUILD-SNAPSHOT

joaoandremartins:


joaoandremartins:
Travis build is currently broken too.

artembilan:
> Bumps to version 1.1.0

??? 

We have not release `1.0.0` yet...

joaoandremartins:
We'll want to release things like the Cloud Stream Binder soon,

artembilan:
Well, we are not in GA, so that is fully OK to add more features and release one more Milestone.
What was then the point to release this M1 if we just jump to the next point version?

I'm fully confused how you are messing with our versioning politics...

I can't find the doc describing the versioning process, but that is definitely not what we all do here.

joaoandremartins:
That's what I was thinking about. For M2, I'd be thinking about adding the Spring Cloud Stream Pub/Sub binder, which should probably go through its Milestone phase before GAing.
On the other hand, I don't want to overcomplicate things by having different versions for the binder and the rest of the stuff, so I'm OK going this way.

I didn't mean to mess with anything. If this is the way you guys do, I'm totally happy going this way :)

artembilan:
No problem! We are not simple guys as well 😉 

So, yes, would be much better to release a single version for all the modules, and that is OK to add new big feature in the Milestone. Code freeze comes on RC phase. And only bug fixed and docs on GA one.

If we decide to add something new, it is already the next minor version. But at the same time we have to maintain the current version in the separate branch for some time as point versions.

Some critical changes goes already to the major version.

We just only need to have a plan of features pack to position ourselves that we are going to GA after this and that.

Something like this. I'm sure there are some other arguments why we live this versioning style....

PR #234, opened 2017-11-06
Updates README file with M1 information

joaoandremartins:


joaoandremartins:
@artembilan sorry for the haste, but @jabubake is going to give a talk at Devoxx.be about Spring Cloud GCP tomorrow or the day after.
It would be nice to merge this PR before the talk. Thanks!

PR #235, opened 2017-11-10
Bumps google-cloud-java API versions to 0.28

joaoandremartins:


artembilan:
Why so much breaking changes? 

We can't afford such a luxury in Spring with our backward compatibility policy from day first.

Yes, that's just a rhetorical question, but I'm afraid a maintenance pain is waiting for us...

Mistic92:
I think there are some breaking changes because it's alpha/beta. 

> Beta: Libraries defined at a Beta quality level are expected to be mostly stable and we're working towards their release candidate. We will address issues and requests with a higher priority.

> Alpha: Libraries defined at an Alpha quality level are still a work-in-progress and are more likely to get backwards-incompatible updates.

artembilan:
Thanks, @Mistic92 , for explanation!

joaoandremartins:
In any case, I'm not super worried about this because the set of dependencies that we cut releases with should remain stable.
This issue only appeared because @Mistic92 overrode `google-cloud-pubsub`'s version to 0.28, which uses higher gax versions than 1.9, which is what M1 uses in its BOM.

FWIW we need to specify the `gax`'s version on our side because we're using classes from it (`ExecutorProvider`, `ChannelProvider`, etc.)

PR #236, opened 2017-11-10
Astoria.RC1 release train

joaoandremartins:


joaoandremartins:
@artembilan @sobychacko is this how a release train is made?

I think we'll only want to merge this when we're ready for GA, which shouldn't be before we get some usage. Just checking how you usually do release trains.

sobychacko:
@joaoandremartins why did you version it as RC1? You can set that to `Astoria.BUILD-SNAPSHOT` and treat that as the BOM for the snapshots. Then when we are ready for the next milestone or RC1, then we can update the release train version accordingly.

artembilan:
Need to understand, first of all, why do we need that `train` at all in this project...
Typically it makes sense if we have several repositories and each of them are released in its own pace and different versions. That's how a `train` feature helps to consolidate them as a siblings or neighbors.

A good sample is Spring Data.

Since this project is a single entry for the release process and everything here is released at the same time, I don't see reason to overhead and overcomplicate. See similar `spring-cloud-aws` project.

Just curious: what is `Astoria`? 😄 

Thanks.

sobychacko:
@artembilan that's a great point. If we are not dealing with multiple repositories and releases, don't see the point of a release train based release. Simple numeric versioning will suffice in that case.

joaoandremartins:
The problem is that everything isn't released at the same time.
I thought of the release train as a way to release what we can to GA, in this case everything that doesn't use `google-cloud-pubsub` and `google-cloud-trace`.
So we'd release `Astoria` whenever we think it's ready, then when `google-cloud-pubsub` or `google-cloud-trace` go GA we can release a `Brooklyn` with `spring-cloud-gcp-pubsub`, `spring-integration-gcp`, etc.

If there are other ways to get there, let me know.

artembilan:
Well, that's still about numeric releasing yet.
We are still in the same repository.
Right, we might decide do not release some module as GA because of its dependencies or any other reasons.

We always can add them to new GA eventually. 
That isn't an argument to `train` this project.

When one of excluded modules becomes ready for GA, we just will need to release the whole project as point version. And semantically it doesn't matter if it is alphabetic or numeric - it is still about new version.
You can't add new dependency to existing train.

For example we add to Spring Integration `5.0` an new `webflux` modules not released before. And its version is still `5.0` - `spring-integration-webflux-5.0`, - not `1.0` as it would be in case we have a separate repository and and release it independently. So, then we have a separate `train` project which collects all necessary dependencies for its release.

I'm not sure why you think other way, but I don't see `train` in one simple project.

Even Spring Boot being too big project still releases as one numeric version.

joaoandremartins:
OK, sounds good.

So when the time to release to GA comes, we'll need to come up with some script trickery to only release the GAable modules, or merge a PR similar to this one to exclude the non-GA modules from the deployment (and roll back after the release is made).

PR #239, opened 2017-11-13
Gh 210 buckets resources

### LINKED ISSUE #210: .exists() throws exception for existing GCS buckets
https://github.com/spring-cloud/spring-cloud-gcp/blob/9debe0aff000396f67045c17e3b70fc056ebcd3c/spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResource.java#L109

throws ``java.lang.StringIndexOutOfBoundsException: String index out of range: -1``

when used with a bucket name that exists.

ChengyuanZhao:
fixes #210

### DIFF
--- a/spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageProtocolResolver.java
+++ b/spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageProtocolResolver.java
@@ -41,7 +41,7 @@ public class GoogleStorageProtocolResolver
 
 	private static final Log logger = LogFactory.getLog(GoogleStorageProtocolResolver.class);
 
-	private static final String PROTOCOL = "gs://";
+	public static final String PROTOCOL = "gs://";
 
 	private ConfigurableListableBeanFactory beanFactory;
 
@@ -73,6 +73,12 @@ public class GoogleStorageProtocolResolver
 	@Override
 	public Resource resolve(String location, ResourceLoader resourceLoader) {
 		if (location.startsWith(PROTOCOL)) {
+
+			int bucketSeparatorIndex = location.indexOf('/', PROTOCOL.length());
+
+			String bucketPath = location.substring(0,
+					bucketSeparatorIndex < 0 ? location.length() : bucketSeparatorIndex);
+
 			if (this.storage == null) {
 				this.storage = this.beanFactory.getBean(Storage.class);
 				try {
@@ -84,7 +90,17 @@ public class GoogleStorageProtocolResolver
 							+ "so defaults are used.");
 				}
 			}
-			return new GoogleStorageResource(this.storage, location, this.settings.isAutoCreateFiles());
+
+			GoogleStorageResourceBucket bucketResource = new GoogleStorageResourceBucket(
+					this.storage, bucketPath, this.settings.isAutoCreateFiles());
+
+			if (bucketSeparatorIndex < 0
+					|| bucketSeparatorIndex + 1 == location.length()) {
+				return bucketResource;
+			}
+
+			return new GoogleStorageResourceObject(this.storage, location,
+					this.settings.isAutoCreateFiles());
 		}
 		return null;
 	}
--- a/spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageProtocolResolver.java
+++ b/spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageProtocolResolver.java
@@ -72,7 +72,9 @@ public class GoogleStorageProtocolResolver
 
 	@Override
 	public Resource resolve(String location, ResourceLoader resourceLoader) {
-		if (location.startsWith(PROTOCOL)) {
+			if (!location.startsWith(PROTOCOL)) {
+					return null;
+			}
 
 			String pathWithoutProtocol = location.substring(PROTOCOL.length());
 
@@ -107,8 +109,6 @@ public class GoogleStorageProtocolResolver
 						this.settings.isAutoCreateFiles());
 			}
 			return resource;
-		}
-		return null;
 	}
 
 }
--- a/spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageProtocolResolver.java
+++ b/spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageProtocolResolver.java
@@ -74,7 +74,13 @@ public class GoogleStorageProtocolResolver
 	public Resource resolve(String location, ResourceLoader resourceLoader) {
 		if (location.startsWith(PROTOCOL)) {
 
-			String[] pathParts = location.substring(PROTOCOL.length()).split("/");
+			String pathWithoutProtocol = location.substring(PROTOCOL.length());
+
+			String[] pathParts = pathWithoutProtocol.split("/");
+
+			if (pathParts.length == 0 || pathWithoutProtocol.isEmpty()) {
+				return null;
+			}
 
 			String bucketName = pathParts[0];
 
@@ -90,13 +96,17 @@ public class GoogleStorageProtocolResolver
 				}
 			}
 
+			Resource resource;
+
 			if (pathParts.length < 2) {
-				return new GoogleStorageResourceBucket(this.storage, bucketName,
+				resource = new GoogleStorageResourceBucket(this.storage, bucketName,
 						this.settings.isAutoCreateFiles());
 			}
-
-			return new GoogleStorageResourceObject(this.storage, location,
-					this.settings.isAutoCreateFiles());
+			else {
+				resource = new GoogleStorageResourceObject(this.storage, location,
+						this.settings.isAutoCreateFiles());
+			}
+			return resource;
 		}
 		return null;
 	}
--- a/spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageProtocolResolver.java
+++ b/spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageProtocolResolver.java
@@ -39,14 +39,14 @@ import org.springframework.core.io.ResourceLoader;
 public class GoogleStorageProtocolResolver
 		implements ProtocolResolver, BeanFactoryPostProcessor, ResourceLoaderAware {
 
-	private static final Log logger = LogFactory.getLog(GoogleStorageProtocolResolver.class);
-
 	public static final String PROTOCOL = "gs://";
+
+	private static final Log logger = LogFactory
+			.getLog(GoogleStorageProtocolResolver.class);
 
 	private ConfigurableListableBeanFactory beanFactory;
 
-	private GoogleStorageProtocolResolverSettings settings
-			= new GoogleStorageProtocolResolverSettings();
+	private GoogleStorageProtocolResolverSettings settings = new GoogleStorageProtocolResolverSettings();
 
 	private volatile Storage storage;
 
@@ -54,8 +54,8 @@ public class GoogleStorageProtocolResolver
 	}
 
 	@Override
-	public void postProcessBeanFactory(ConfigurableListableBeanFactory beanFactory)
-			throws BeansException {
+	public void postProcessBeanFactory(
+			ConfigurableListableBeanFactory beanFactory) throws BeansException {
 		this.beanFactory = beanFactory;
 	}
 
@@ -65,50 +65,50 @@ public class GoogleStorageProtocolResolver
 			((DefaultResourceLoader) resourceLoader).addProtocolResolver(this);
 		}
 		else {
-			logger.warn("The provided delegate resource loader is not an implementation " +
-					"of DefaultResourceLoader. Custom Protocol using gs:// prefix will not be enabled.");
+			logger.warn("The provided delegate resource loader is not an implementation "
+					+ "of DefaultResourceLoader. Custom Protocol using gs:// prefix will not be enabled.");
 		}
 	}
 
 	@Override
 	public Resource resolve(String location, ResourceLoader resourceLoader) {
-			if (!location.startsWith(PROTOCOL)) {
-					return null;
+		if (!location.startsWith(PROTOCOL)) {
+			return null;
+		}
+
+		String pathWithoutProtocol = location.substring(PROTOCOL.length());
+
+		String[] pathParts = pathWithoutProtocol.split("/");
+
+		if (pathParts.length == 0 || pathWithoutProtocol.isEmpty()) {
+			return null;
+		}
+
+		String bucketName = pathParts[0];
+
+		if (this.storage == null) {
+			this.storage = this.beanFactory.getBean(Storage.class);
+			try {
+				this.settings = this.beanFactory
+						.getBean(GoogleStorageProtocolResolverSettings.class);
 			}
-
-			String pathWithoutProtocol = location.substring(PROTOCOL.length());
-
-			String[] pathParts = pathWithoutProtocol.split("/");
-
-			if (pathParts.length == 0 || pathWithoutProtocol.isEmpty()) {
-				return null;
+			catch (NoSuchBeanDefinitionException e) {
+				logger.info("There is no bean definition for the resolver settings, "
+						+ "so defaults are used.");
 			}
+		}
 
-			String bucketName = pathParts[0];
+		Resource resource;
 
-			if (this.storage == null) {
-				this.storage = this.beanFactory.getBean(Storage.class);
-				try {
-					this.settings = this.beanFactory
-							.getBean(GoogleStorageProtocolResolverSettings.class);
-				}
-				catch (NoSuchBeanDefinitionException e) {
-					logger.info("There is no bean definition for the resolver settings, "
-							+ "so defaults are used.");
-				}
-			}
-
-			Resource resource;
-
-			if (pathParts.length < 2) {
-				resource = new GoogleStorageResourceBucket(this.storage, bucketName,
-						this.settings.isAutoCreateFiles());
-			}
-			else {
-				resource = new GoogleStorageResourceObject(this.storage, location,
-						this.settings.isAutoCreateFiles());
-			}
-			return resource;
+		if (pathParts.length < 2) {
+			resource = new GoogleStorageResourceBucket(this.storage, bucketName,
+					this.settings.isAutoCreateFiles());
+		}
+		else {
+			resource = new GoogleStorageResourceObject(this.storage, location,
+					this.settings.isAutoCreateFiles());
+		}
+		return resource;
 	}
 
 }
--- a/spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageProtocolResolver.java
+++ b/spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageProtocolResolver.java
@@ -74,10 +74,9 @@ public class GoogleStorageProtocolResolver
 	public Resource resolve(String location, ResourceLoader resourceLoader) {
 		if (location.startsWith(PROTOCOL)) {
 
-			int bucketSeparatorIndex = location.indexOf('/', PROTOCOL.length());
+			String[] pathParts = location.substring(PROTOCOL.length()).split("/");
 
-			String bucketPath = location.substring(0,
-					bucketSeparatorIndex < 0 ? location.length() : bucketSeparatorIndex);
+			String bucketName = pathParts[0];
 
 			if (this.storage == null) {
 				this.storage = this.beanFactory.getBean(Storage.class);
@@ -91,12 +90,9 @@ public class GoogleStorageProtocolResolver
 				}
 			}
 
-			GoogleStorageResourceBucket bucketResource = new GoogleStorageResourceBucket(
-					this.storage, bucketPath, this.settings.isAutoCreateFiles());
-
-			if (bucketSeparatorIndex < 0
-					|| bucketSeparatorIndex + 1 == location.length()) {
-				return bucketResource;
+			if (pathParts.length < 2) {
+				return new GoogleStorageResourceBucket(this.storage, bucketName,
+						this.settings.isAutoCreateFiles());
 			}
 
 			return new GoogleStorageResourceObject(this.storage, location,
--- a/spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceBucket.java
+++ b/spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceBucket.java
@@ -17,6 +17,7 @@
 package org.springframework.cloud.gcp.storage;
 
 import java.io.File;
+import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
@@ -24,6 +25,7 @@ import java.net.URI;
 import java.net.URISyntaxException;
 import java.net.URL;
 
+import com.google.cloud.storage.Bucket;
 import com.google.cloud.storage.BucketInfo;
 import com.google.cloud.storage.Storage;
 
@@ -38,29 +40,29 @@ public class GoogleStorageResourceBucket implements WritableResource {
 
 	private final Storage storage;
 
-	private final String location;
+	private final String bucketName;
 
-	public GoogleStorageResourceBucket(Storage storage, String location,
+	public GoogleStorageResourceBucket(Storage storage, String bucketName,
 			boolean createBucketIfNotExists) {
 		Assert.notNull(storage, "Storage object can not be null");
-		Assert.notNull(storage, "Path to bucket can not be null");
+		Assert.notNull(bucketName, "Bucket name can not be null");
 
 		this.storage = storage;
-
-		// Ignore trailing slash
-		int trailingSlashIndex = location.indexOf('/',
-				GoogleStorageProtocolResolver.PROTOCOL.length());
-		this.location = location.substring(0,
-				trailingSlashIndex < 0 ? location.length() : trailingSlashIndex);
+		this.bucketName = bucketName;
 
 		if (!exists() && createBucketIfNotExists) {
-			this.storage.create(BucketInfo.newBuilder(this.location).build());
+			this.storage.create(BucketInfo.newBuilder(this.bucketName).build());
 		}
 	}
 
 	@Override
 	public boolean exists() {
-		return this.storage.get(this.location) != null;
+		try {
+			return throwExceptionForNullBucket(resolve()).exists();
+		}
+		catch (IOException e) {
+			return false;
+		}
 	}
 
 	@Override
@@ -82,7 +84,7 @@ public class GoogleStorageResourceBucket implements WritableResource {
 	public URI getURI() throws IOException {
 		URI uri = null;
 		try {
-			uri = new URI(this.location);
+			uri = new URI(GoogleStorageProtocolResolver.PROTOCOL + getFilename());
 		}
 		catch (URISyntaxException e) {
 			throw new IOException("Invalid URI syntax", e);
@@ -111,17 +113,23 @@ public class GoogleStorageResourceBucket implements WritableResource {
 	@Override
 	public Resource createRelative(String relativePath) throws IOException {
 		return new GoogleStorageResourceObject(this.storage,
-				this.location + (relativePath.startsWith("/") ? "" : "/") + relativePath);
+				getURI().toString() + (relativePath.startsWith("/") ? "" : "/")
+						+ relativePath);
 	}
 
 	@Override
 	public String getFilename() {
-		return this.location;
+		try {
+			return throwExceptionForNullBucket(resolve()).getName();
+		}
+		catch (IOException e) {
+			return null;
+		}
 	}
 
 	@Override
 	public String getDescription() {
-		return this.location;
+		return this.bucketName;
 	}
 
 	@Override
@@ -139,5 +147,17 @@ public class GoogleStorageResourceBucket implements WritableResource {
 	public OutputStream getOutputStream() throws IOException {
 		throw new UnsupportedOperationException(
 				"Cannot get output stream for a storage bucket.");
+	}
+
+	private Bucket resolve() {
+		return this.storage.get(this.bucketName);
+	}
+
+	private Bucket throwExceptionForNullBucket(Bucket bucket) throws IOException {
+		if (bucket == null) {
+			throw new FileNotFoundException(
+					"The bucket was not found: " + this.bucketName);
+		}
+		return bucket;
 	}
 }
--- a/spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceBucket.java
+++ b/spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceBucket.java
@@ -1,0 +1,143 @@
+/*
+ *  Copyright 2017 original author or authors.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package org.springframework.cloud.gcp.storage;
+
+import java.io.File;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.net.URL;
+
+import com.google.cloud.storage.BucketInfo;
+import com.google.cloud.storage.Storage;
+
+import org.springframework.core.io.Resource;
+import org.springframework.core.io.WritableResource;
+import org.springframework.util.Assert;
+
+/**
+ * Implements {@link WritableResource} for buckets in Google Cloud Storage.
+ */
+public class GoogleStorageResourceBucket implements WritableResource {
+
+	private final Storage storage;
+
+	private final String location;
+
+	public GoogleStorageResourceBucket(Storage storage, String location,
+			boolean createBucketIfNotExists) {
+		Assert.notNull(storage, "Storage object can not be null");
+		Assert.notNull(storage, "Path to bucket can not be null");
+
+		this.storage = storage;
+
+		// Ignore trailing slash
+		int trailingSlashIndex = location.indexOf('/',
+				GoogleStorageProtocolResolver.PROTOCOL.length());
+		this.location = location.substring(0,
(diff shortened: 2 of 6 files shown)


ChengyuanZhao:
@artembilan @meltsufin I can't add reviewers/ assign myself to PRs and issues for this repo until I'm added as a collaborator. In the meantime, please add take a look and add yourself as reviewers?

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceBucket.java
```diff
@@ -0,0 +1,163 @@
...
+import com.google.cloud.storage.Storage;
+
+import org.springframework.core.io.Resource;
+import org.springframework.core.io.WritableResource;
+import org.springframework.util.Assert;
+
+/**
+ * Implements {@link WritableResource} for buckets in Google Cloud Storage.
+ */
+public class GoogleStorageResourceBucket implements WritableResource {
```
joaoandremartins: WritableResource not needed here?

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceObject.java
```diff
@@ -114,6 +115,12 @@ private Blob resolve() throws IOException {
 	}
 
 	private Blob createBlob() throws IOException {
+		// Creating the bucket resource will create the bucket if necessary.
```
joaoandremartins: Unhelpful comment?

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceBucket.java
```diff
@@ -0,0 +1,163 @@
...
+	public OutputStream getOutputStream() throws IOException {
+		throw new UnsupportedOperationException(
+				"Cannot get output stream for a storage bucket.");
+	}
+
+	private Bucket resolve() {
+		return this.storage.get(this.bucketName);
+	}
+
+	private Bucket throwExceptionForNullBucket(Bucket bucket) throws IOException {
```
joaoandremartins: I know this code is copied and pasted, but should this method be called `getBucket()` instead?

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceBucket.java
```diff
@@ -0,0 +1,163 @@
...
+
+	public GoogleStorageResourceBucket(Storage storage, String bucketName,
+			boolean createBucketIfNotExists) {
+		Assert.notNull(storage, "Storage object can not be null");
+		Assert.notNull(bucketName, "Bucket name can not be null");
+
+		this.storage = storage;
+		this.bucketName = bucketName;
+
+		if (!exists() && createBucketIfNotExists) {
```
joaoandremartins: Can we do this elsewhere? I don't like this kind of advanced logic inside the constructor...

How about we create a `createBucketInGCS()` method?

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceBucket.java
```diff
@@ -0,0 +1,163 @@
...
+			this.storage.create(BucketInfo.newBuilder(this.bucketName).build());
+		}
+	}
+
+	@Override
+	public boolean exists() {
+		try {
+			return throwExceptionForNullBucket(resolve()).exists();
```
joaoandremartins: Again, I know this is copy&pasted, but can we change this to call `resolve()` and check if the result is `null`?
That would spare us the unnecessary and costly exception handling.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceBucket.java
```diff
@@ -0,0 +1,163 @@
...
+		catch (URISyntaxException e) {
+			throw new IOException("Invalid URI syntax", e);
+		}
+		return uri;
+	}
+
+	@Override
+	public File getFile() throws IOException {
+		throw new UnsupportedOperationException(
+				"Cannot get a file from a storage bucket.");
```
joaoandremartins: Change message to something like "GCS bucket is not in the local file system"?

ChengyuanZhao:
travis-ci had trouble starting up:

Downloading https://repo1.maven.org/maven2/org/apache/maven/apache-maven/3.3.9/apache-maven-3.3.9-bin.zip

Exception in thread "main" javax.net.ssl.SSLHandshakeException: java.security.cert.CertificateException: No subject alternative DNS name matching repo1.maven.org found.

I'll close and reopen the PR to trigger a re-build

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceBucket.java
```diff
@@ -0,0 +1,163 @@
...
+	@Override
+	public long lastModified() throws IOException {
+		throw new UnsupportedOperationException(
+				"Cannot get last-modified for a storage bucket.");
+	}
+
+	@Override
+	public Resource createRelative(String relativePath) throws IOException {
+		return new GoogleStorageResourceObject(this.storage,
+				getURI().toString() + (relativePath.startsWith("/") ? "" : "/")
```
joaoandremartins: What if the bucket URI already ends in a `/`?
We'd want to avoid cases like `gs://bucket//blob`, but we should also take this chance to fix the case where we can't create blobs on buckets that don't end in a `/`.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceBucket.java
```diff
@@ -0,0 +1,163 @@
...
+						+ relativePath);
+	}
+
+	@Override
+	public String getFilename() {
+		try {
+			return throwExceptionForNullBucket(resolve()).getName();
```
joaoandremartins: Ditto for unnecessary exception handling.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceBucket.java
joaoandremartins: On a second look, it looks like this method is even unnecessary due to my previous comments.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceObject.java
```diff
@@ -114,6 +115,12 @@ private Blob resolve() throws IOException {
...
+		Resource bucket = getBucket();
+		if (!bucket.exists()) {
```
joaoandremartins: I'd prefer to see another kind of logic here, like -- check if the bucket exists, if it doesn't, create it if you can (e.g. configuration allows) by calling an explicit `create()`-like method (not a constructor).

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceObject.java
```diff
@@ -114,6 +115,12 @@ private Blob resolve() throws IOException {
...
+			throw new IOException("Unable to create blob: " + getBlobId().toString()
+					+ ", because the bucket does not exist: " + getBlobId().getBucket());
```
joaoandremartins: Does `getBlobId().getBucket()` have the desired `toString()` effect?

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceBucket.java
ChengyuanZhao: I was thinking of keeping it there just so we can throw the unsupported operation exception along with the explanation string that it's a bucket resource. The reason being that the user will be using bucket path resources and object path resources in the same way, and if they inadvertently try to get an output stream or check isWriteable I would rather they get the unsupported operation exception with an explanation than a class casting exception because those two functions don't exist at all.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceBucket.java
ChengyuanZhao: Will do.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceBucket.java
ChengyuanZhao: Will do.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceBucket.java
ChengyuanZhao: Yup will do.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceBucket.java
ChengyuanZhao: We don't store the entire bucket path in this class, just the bucket name. The reason is the GCS bucket functions all work on the bucket name, not the path. As a result, the path we construct here for the bucket will always not have the /.

ChengyuanZhao: Also, i've got a unit test for this case , too.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceBucket.java
ChengyuanZhao: This one is necessary, since resolve() might return null, and .getName() would then give an NPE.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceObject.java
ChengyuanZhao: Will remove.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceObject.java
ChengyuanZhao: Yes, the toString() gets the entire path to the blob , and the getBucket() just gets the bucket name.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceBucket.java
ChengyuanZhao: Oh on second look, I had this here because resolve() might return null, and calling .exists() would then give an NPE.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceBucket.java
joaoandremartins: Actually, the way things are now, the user needs to cast a `Resource` into a `WritableResource` to be able to write to it. In which case, the cast should actually just fail, instead of allowing it and then throwing UOE.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceBucket.java
```diff
@@ -0,0 +1,158 @@
...
+
+	public GoogleStorageResourceBucket(Storage storage, String bucketName,
+			boolean createBucketIfNotExists) {
+		Assert.notNull(storage, "Storage object can not be null");
+		Assert.notNull(bucketName, "Bucket name can not be null");
+
+		this.storage = storage;
+		this.bucketName = bucketName;
+
+		createBucketIfRequired(createBucketIfNotExists);
```
joaoandremartins: I actually meant not doing this in the constructor at all. Instead, the user can call `create()` if they want to create it explicitly, or we create it for them automatically in any appropriate methods (`createRelative()` is really the only one that comes to mind) if `createBucketIfNotExists` is `true`.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceBucket.java
joaoandremartins: Yeah, I meant to ask about that -- do we really need them to work based on the name?
I'd rather have them work based on a `Bucket`, with `getName()` etc., unless there is a good reason not to.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageProtocolResolver.java
```diff
@@ -84,7 +89,14 @@ public Resource resolve(String location, ResourceLoader resourceLoader) {
...
 				}
 			}
-			return new GoogleStorageResource(this.storage, location, this.settings.isAutoCreateFiles());
+
+			if (pathParts.length < 2) {
+				return new GoogleStorageResourceBucket(this.storage, bucketName,
+						this.settings.isAutoCreateFiles());
+			}
+
+			return new GoogleStorageResourceObject(this.storage, location,
```
joaoandremartins: I'm guessing we're letting the GCS API error out for malformed GCS URIs, like `gs://bucket-name/asdasd/asdasd/asdasd`. I think I'm fine with that.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceBucket.java
```diff
@@ -0,0 +1,158 @@
...
+	public Resource createRelative(String relativePath) throws IOException {
+		return new GoogleStorageResourceObject(this.storage,
+				getURI().toString() + (relativePath.startsWith("/") ? "" : "/")
+						+ relativePath);
+	}
+
+	@Override
+	public String getFilename() {
+		Bucket bucket = resolve();
+		return bucket == null ? null : bucket.getName();
```
joaoandremartins: This could probably be improved if `bucket` were a non-nullable field.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceObject.java
```diff
@@ -114,6 +115,11 @@ private Blob resolve() throws IOException {
 	}
 
 	private Blob createBlob() throws IOException {
+		Resource bucket = getBucket();
+		if (!bucket.exists()) {
```
joaoandremartins: Previous comment on logic change here still stands.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceBucket.java
ChengyuanZhao: I actually think it's pretty clean to just have the bucket name as a string, and I would prefer to not have a Bucket that might refer to a nonexistent bucket on GCS. As it is now a GSRB is an object that refers to a potentially nonexistent bucket on GCS, and any actual Bucket we're using exists on GCS.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceBucket.java
ChengyuanZhao: I think it's just a difference of opinion. I prefer doing it in the constructor, and we control it with the auto-create boolean argument.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageProtocolResolver.java
ChengyuanZhao: Isn't `gs://bucket-name/asdasd/asdasd/asdasd` a valid path though?

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceBucket.java
ChengyuanZhao: It is gone :)

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceObject.java
ChengyuanZhao: Yea I just disagreed that comment. It feels pretty natural to me to have the constructor take care of creating the bucket or not based on the auto-create bool setting.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageProtocolResolver.java
joaoandremartins: You're right -- nevermind that!

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceBucket.java
joaoandremartins: It's more than just opinion.
The constructor should not contain extraneous logic that should belong in other methods. Later on this PR you're creating the file remotely in GCS by calling the constructor. I would not expect that as a Java developer.
There is no reason why the status of an object of this kind in memory should map to the status of the file in GCS. That's why methods like `exists()` exist.

It also makes unit testing objects of this kind a bummer.
That method should probably also throw an `IOException` in case the file couldn't be constructed, which is a code smell coming from a constructor.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceBucket.java
joaoandremartins: How is calling `resolve()` on every method that needs the bucket _cleaner_?

And again, no reason why the state of this Java object in memory should map to the state of the GCS entity it refers to IMO.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceBucket.java
```diff
@@ -0,0 +1,158 @@
...
+			uri = new URI(GoogleStorageProtocolResolver.PROTOCOL + getFilename());
+		}
+		catch (URISyntaxException e) {
+			throw new IOException("Invalid URI syntax", e);
+		}
+		return uri;
+	}
+
+	@Override
+	public File getFile() throws IOException {
```
joaoandremartins: Sorry for bothering you with changes to pasted code, but the thrown `IOException` should also probably be changed to `UnsupportedOperationException`.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceBucket.java
```diff
@@ -0,0 +1,158 @@
...
+	}
+
+	@Override
+	public long lastModified() throws IOException {
+		throw new UnsupportedOperationException(
+				"Cannot get last-modified for a storage bucket.");
+	}
+
+	@Override
+	public Resource createRelative(String relativePath) throws IOException {
```
joaoandremartins: Seems like this method actually doesn't throw an `IOException`?

joaoandremartins: NVM -- that is part of the method signature.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceBucket.java
ChengyuanZhao: Again, I think this is an opinion thing. I think it's conceptually much cleaner, and calling resolve 3 or 4 times in the class isn't messy in my opinion.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceBucket.java
ChengyuanZhao: I really think your explanation still sounds just like your opinion. I can explain my perspective too:

- a user can create a GSRB setting the auto-create bool
- the user doesn't need to know if the bucket already exists: that's the whole point of the auto-create functionality. 
- the `exists` function works as expected

As a Java developer I think this is pretty simple and intuitive.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceBucket.java
ChengyuanZhao: Ah my bad! Will fix.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceBucket.java
joaoandremartins: I'll defer to what others think. @artembilan @meltsufin

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceBucket.java
meltsufin: In the case of a GCS object resource we only create it when attempting to write to it. I don't see why it should be any different for a bucket resource. The "write" operation for a Bucket is `.createRelative()`. So, I think that's where the call to `createBucketIfRequired` belongs.

`GoogleStorageResourceBucket`, just like `java.util.File` is simply a handle to a resource that may or may not exist. So, auto-creation in the constructor doesn't make sense to me. I would also make `createBucketIfRequired` public.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceBucket.java
```diff
@@ -0,0 +1,149 @@
...
+		return new GoogleStorageResourceObject(this.storage,
+				getURI().toString() + (relativePath.startsWith("/") ? "" : "/")
+						+ relativePath);
+	}
+
+	@Override
+	public String getFilename() {
+		Bucket bucket = resolve();
```
meltsufin: You already have `this.bucketName`. Why do `resolve()`?

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceBucket.java
```diff
@@ -0,0 +1,149 @@
...
+	}
+
+	@Override
+	public boolean exists() {
+		return resolve() != null;
+	}
+
+	@Override
+	public boolean isReadable() {
+		return false;
```
meltsufin: Why `false`?

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceBucket.java
meltsufin: See my other comment on `getFilename` which doesn't require a call to `resolve` and it also guarantees that there is no trailing `/`. So, as long as `getFilename` is fixed, simply using `this.bucketName`, this code is fine.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceObject.java
meltsufin: If you're doing auto-create, it should be here, not in the constructor.

spring-cloud-gcp-storage/src/test/java/org/springframework/cloud/gcp/storage/GoogleStorageTests.java
```diff
@@ -51,13 +53,73 @@
 	@Value("gs://test-spring/images/spring.png")
 	private Resource remoteResource;
 
+	@Value("gs://test-spring")
```
meltsufin: Please also test
```
gs://
gs:///
```

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageProtocolResolver.java
```diff
@@ -84,7 +89,14 @@ public Resource resolve(String location, ResourceLoader resourceLoader) {
 							+ "so defaults are used.");
 				}
 			}
-			return new GoogleStorageResource(this.storage, location, this.settings.isAutoCreateFiles());
+
+			if (pathParts.length < 2) {
```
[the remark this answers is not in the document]
ChengyuanZhao: gs://bucket/ would work fine, gs:// would try to create a bucket with "" as the name, which would get you an error from the GCS functions. gs:/// would cause an array out of bounds exception when i try to get that bucket name after splitting the string. I'll add a check for that last case.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceBucket.java
ChengyuanZhao: The doc says if you can't get an input stream from the resource then it should be false.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceBucket.java
ChengyuanZhao: Yea I'll make the change. Originally I had wanted it to return the name only if it exists, but that's not necessary.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceBucket.java
ChengyuanZhao: Yea this makes sense. I'll make the change

spring-cloud-gcp-storage/src/test/java/org/springframework/cloud/gcp/storage/GoogleStorageTests.java
ChengyuanZhao: Changes made, please take a look :)

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageProtocolResolver.java
```diff
@@ -73,6 +73,17 @@ public void setResourceLoader(ResourceLoader resourceLoader) {
 	@Override
 	public Resource resolve(String location, ResourceLoader resourceLoader) {
 		if (location.startsWith(PROTOCOL)) {
+
+			String pathWithoutProtocol = location.substring(PROTOCOL.length());
+
+			String[] pathParts = pathWithoutProtocol.split("/");
+
+			if (pathParts.length == 0 || pathWithoutProtocol.isEmpty()) {
```
joaoandremartins: Could be nice to throw an exception here saying the path is invalid?
Spring's `Assert` library is nice for that.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageProtocolResolver.java
```diff
@@ -84,7 +95,18 @@ public Resource resolve(String location, ResourceLoader resourceLoader) {
...
 			}
-			return new GoogleStorageResource(this.storage, location, this.settings.isAutoCreateFiles());
+
+			Resource resource;
+
+			if (pathParts.length < 2) {
+				resource = new GoogleStorageResourceBucket(this.storage, bucketName,
+						this.settings.isAutoCreateFiles());
+			}
+			else {
```
joaoandremartins: The local variable `resource` is unnecessary.
You can get rid of it and this else by returning where you have `resource =`

ChengyuanZhao: Yea I had it like that before, but with the return-null for the gs:/// path above it is 4  return statements in the function and checkstyle disallows that.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceBucket.java
```diff
@@ -0,0 +1,150 @@
...
+ *
+ * @author Chengyuan Zhao
+ */
+public class GoogleStorageResourceBucket implements WritableResource {
+
+	private final Storage storage;
+
+	private final String bucketName;
+
+	private final boolean createBucketIfRequired;
```
joaoandremartins: Where do we stand regarding automatic bucket creation?
@meltsufin suggested we shouldn't be doing it, and I'm tempted to agree.
I think users are expected to create the buckets on their own, even more so because bucket names are unique across GCS.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceObject.java
```diff
@@ -180,4 +185,10 @@ public OutputStream getOutputStream() throws IOException {
 		}
 		return Channels.newOutputStream(throwExceptionForNullBlob(blob).writer());
 	}
+
+	@VisibleForTesting
```
joaoandremartins: This is actually used by `createBlob()`, so it might make sense to remove `@VisibleForTesting`.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceBucket.java
ChengyuanZhao: I thought you were on board with the idea at the beginning? Tbh I think intentionally not giving this functionality  would be paternalism/parentalism. The API for creating buckets is super simple to use, and people could go wild with that with minimal effort as it is. Also, since we primarily let people use this functionality via @Value() annotations in source code, that's not conducive to wanton bucket creation anyway.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageProtocolResolver.java
joaoandremartins: You might be able to remove the previous `return null` if you throw an exception for invalid path, which I think is actually the better way to go there.
If that doesn't help, feel free to ship like this, though that checkstyle rule has annoyed me before too...

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceBucket.java
joaoandremartins: I don't really have strong feelings on this matter.
You do have a point that there is an API for that already and we're just making it easy for users.

So, if @meltsufin is aboard, I'm OK with shipping this feature.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceObject.java
ChengyuanZhao: Ah yes, will fix.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageProtocolResolver.java
ChengyuanZhao: That sounds like the way to go. Will change.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageProtocolResolver.java
ChengyuanZhao: Actually, I think this function should return null for any situation other than success. Looking at the parent's javadoc, it says something like that, and does not throw something like IOException.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageProtocolResolver.java
joaoandremartins: I would also change the first if to `if (!location.startsWith(PROTOCOL)) { return null; }`, so that the whole method doesn't need to be inside the while loop and doesn't end with that `return null`.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageProtocolResolver.java
```diff
@@ -72,7 +72,9 @@ public void setResourceLoader(ResourceLoader resourceLoader) {
 
 	@Override
 	public Resource resolve(String location, ResourceLoader resourceLoader) {
-		if (location.startsWith(PROTOCOL)) {
```
joaoandremartins: Indendation doesn't seem right

ChengyuanZhao: Ah yea, I felt like it looked wrong. I just installed intellij again here and I found this formatting file (https://raw.githubusercontent.com/spring-cloud/spring-cloud-build/master/spring-cloud-dependencies-parent/eclipse-code-formatter.xml) is that the right one?

PR #240, opened 2017-11-14
Spring Data JPA code sample

joaoandremartins:


artembilan:
What is the point of this application?

It absolutely doesn't demonstrate any Spring GCP functionality.
This is raw JPA application based on the Spring Boot auto-configuration.

Since our GCP-SQL support provides standard `DataSource` bean configuration that is really obvious that JPA will work automatically.

Thanks for understanding.

joaoandremartins:
Many people ask about if our starter is compatible with Spring Data JPA.
Having this example here should answer that question very easily for new users, at a very low cost.

artembilan:
OK. Let it be.
Let merge after fixing Travis issues!

```
[ERROR] /home/travis/build/spring-cloud/spring-cloud-gcp/spring-cloud-gcp-examples/spring-cloud-gcp-data-jpa-example/src/main/java/com/example/DemoApplication.java:41: Undefined reference: java.util.stream.Stream java.util.stream.Stream.of(Object[])

[ERROR] /home/travis/build/spring-cloud/spring-cloud-gcp/spring-cloud-gcp-examples/spring-cloud-gcp-data-jpa-example/src/main/java/com/example/DemoApplication.java:45: Undefined reference: void java.util.stream.Stream.forEach(java.util.function.Consumer)

[ERROR] /home/travis/build/spring-cloud/spring-cloud-gcp/spring-cloud-gcp-examples/spring-cloud-gcp-data-jpa-example/src/main/java/com/example/DemoApplication.java:47: Undefined reference: void Iterable.forEach(java.util.function.Consumer)
```
I guess something wrong with Java 8 compatibility on this nasty Maven level.
Maybe there is need something like `.jdk8` in this new module to trick the Animal Sniffer Plugin?

PR #242, opened 2017-11-15
Bumps google-cloud-storage version to 1.11.0

### LINKED ISSUE #227: Dependency conflict in BOM, Trace
FYI @saturnism @joaoandremartins 

Looks like there's been a recent update, the google-cloud-java libraries for Pub/Sub, Trace got bumped to 0.26, while storage is still on an older version of

```
<google-cloud-storage.version>1.2.1</google-cloud-storage.version>
```

Trace starter seems to be using a hard-coded version of GAX in its [pom.xml](https://github.com/spring-cloud/spring-cloud-gcp/blob/master/spring-cloud-gcp-starters/spring-cloud-gcp-starter-trace/pom.xml#L31)  which also conflicts. This should not be required.

joaoandremartins:
Tested change.

Fixes #227

### DIFF
--- a/spring-cloud-gcp-dependencies/pom.xml
+++ b/spring-cloud-gcp-dependencies/pom.xml
@@ -19,7 +19,7 @@
 		<google-cloud-core.version>1.2.3</google-cloud-core.version>
 		<google-cloud-pubsub.version>0.28.0-beta</google-cloud-pubsub.version>
 		<google-cloud-trace.version>0.28.0-beta</google-cloud-trace.version>
-		<google-cloud-storage.version>1.2.1</google-cloud-storage.version>
+		<google-cloud-storage.version>1.8.0</google-cloud-storage.version>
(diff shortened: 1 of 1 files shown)


Mistic92:
Why 1.8.0? Current is 1.11.0. I think all gcloud dependencies should have one version field. Ofc separate for ga/beta/alpha

joaoandremartins:
@Mistic92 you're right!
It seems like they must update their page: https://cloud.google.com/storage/docs/reference/libraries

If I understood what you meant with one field for every gcloud dependency, currently we could do that for `google-cloud-pubsub` and `google-cloud-trace`, but only recently, since `google-cloud-trace` used to be 0.xx.0-alpha and `google-cloud-pubsub` 0.xx.0-beta.

I suggest we do that when we start using more dependencies that use the same version number.

Mistic92:
Yep, docs on cloud.google.com are not updated :( But you can track updates at GitHub
https://github.com/GoogleCloudPlatform/google-cloud-java

Issue #201, opened 2017-10-19
Move to SF 5.0 and SB 2.0 when they're GA


joaoandremartins:

artembilan:
I would say not a fact that we need their GA for our development.
That really might be just enough to rely on the same `spring-cloud-dependencies-parent:2.0.0.BUILD-SNAPSHOT` during development phase.
The Milestones can be released with appropriate Milestones from there.

If you definitely have plans to go GA for `1.0.0` before appropriate GAs for our `Spring Cloud 2.0` dependencies (including Spring Cloud Stream 2.0 for Binder), than let's post-pone this task to the next year for our version `2.0`.

But still we need keep in mind: even if Google release its `alpha` and `beta` into Maven Central, that wouldn't  be good to release our GA based on them. So, we definitely need GA for all our dependencies.

PR #243, opened 2017-12-01
Initial implementation of the GCS Spring Integration channel adapters

### LINKED ISSUE #34: Integration inbound/outbound handlers for Storage
Create inbound/outbound gateways for Storage. Inbound should be based on `AbstractRemoteFileStreamingMessageSource`

joaoandremartins:
Fixes #34


joaoandremartins:
To do: separate the `spring-integration-gcp` module into two, `spring-integration-gcp-pubsub` and `spring-integration-gcp-storage`, which I suggest to do in another PR so this one doesn't become too unreadable.

garyrussell:
Hi João,

I see you just have the streaming inbound channel adapter; for consistency with FTP etc, it should be called something like `GCSStreamingMessageSource`

Are you planning on creating the `GCSInboundFileSynchronizingMessageSource` too?

Again, for consistency, we should probably create a `GCSRemoteFileTemplate` subclass too.

joaoandremartins:
@garyrussell I renamed the adapters and added `GCSRemoteFileTemplate`.

I'm planning on adding `GCSInboundFileSynchronizingMessageSource` and `FileListFilters` too, but I'm thinking we can do so on separate PRs so this one doesn't grow too big.

Please review whenever you can. Thanks!

garyrussell:
Hi João,

I am on vacation until Thursday (took a few days off after s1p). I have a couple of things on my plate when I return, including this; it may be Friday or Monday before I can get to it.

G.

spring-integration-gcp/pom.xml
```diff
@@ -21,10 +21,18 @@
 			<groupId>org.springframework.integration</groupId>
 			<artifactId>spring-integration-core</artifactId>
 		</dependency>
+		<dependency>
+			<groupId>org.springframework.integration</groupId>
+			<artifactId>spring-integration-file</artifactId>
```
artembilan: `optional`.
Someone might not be interested in Google Cloud Storage, but just PubSub.

Similar situation we have with the SI-AWS and I don't think that it is going to be good to granulate to separate short modules. End user will be forced to add an extra dependency anyway

spring-integration-gcp/src/main/java/org/springframework/integration/gcp/storage/GCSFileInfo.java
```diff
@@ -0,0 +1,68 @@
...
+package org.springframework.integration.gcp.storage;
+
+import com.google.cloud.storage.BlobInfo;
+
+import org.springframework.integration.file.remote.AbstractFileInfo;
+
+/**
+ * @author João André Martins
+ */
+public class GCSFileInfo extends AbstractFileInfo<BlobInfo> {
```
artembilan: I think this must be like `GcsFileInfo`. See `FtpFileInfo` or `SftpFileInfo` for example.

spring-integration-gcp/src/main/java/org/springframework/integration/gcp/storage/GCSFileInfo.java
```diff
@@ -0,0 +1,68 @@
...
+
+	private BlobInfo gcsFile;
+
+	public GCSFileInfo(BlobInfo gcsFile) {
+		this.gcsFile = gcsFile;
```
artembilan: `Assert.notNull()`

spring-integration-gcp/src/main/java/org/springframework/integration/gcp/storage/GCSSession.java
```diff
@@ -0,0 +1,192 @@
...
+ * @author João André Martins
+ */
+public class GCSSession implements Session<BlobInfo> {
+
+	private Storage gcs;
+
+	private static final Log LOGGER = LogFactory.getLog(GCSSession.class);
+
+	public GCSSession(Storage gcs) {
+		this.gcs = gcs;
```
artembilan: `Assert.notNull()`

spring-integration-gcp/src/main/java/org/springframework/integration/gcp/storage/GCSSession.java
```diff
@@ -0,0 +1,192 @@
...
+	}
+
+	@Override
+	public boolean remove(String path) throws IOException {
+		String[] tokens = getBucketAndObjectFromPath(path);
+
+		return tokens.length < 3
+				? this.gcs.delete(tokens[1])
+				: this.gcs.delete(tokens[1], tokens[2]);
```
artembilan: Something wrong is here. You do `path.split("/", 2);`, so, you can't get more than 2 tokens and the second one will be like `folder/blob`.
I might miss something, so elaborate, please.

spring-integration-gcp/src/main/java/org/springframework/integration/gcp/storage/GCSSessionFactory.java
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+ * @author João André Martins
+ */
+public class GCSSessionFactory implements SessionFactory<BlobInfo>, SharedSessionCapable {
+
+	private Storage gcs;
+
+	public GCSSessionFactory(Storage gcs) {
+		this.gcs = gcs;
```
artembilan: `Assert.notNull()`

spring-integration-gcp/src/main/java/org/springframework/integration/gcp/storage/inbound/GCSInboundFileSynchronizingMessageSource.java
```diff
@@ -0,0 +1,44 @@
...
+	}
+
+	public GCSInboundFileSynchronizingMessageSource(GCSInboundFileSynchronizer synchronizer,
+			Comparator<File> comparator) {
+		super(synchronizer, comparator);
+	}
+
+	@Override
+	public String getComponentType() {
+		return "gcp:gcs-inbound-streaming-channel-adapter";
```
artembilan: It's not `streaming`. This one is simple `gcs-inbound-channel-adapter`

spring-integration-gcp/src/test/java/org/springframework/integration/gcp/storage/inbound/GCSInboundFileSynchronizerTest.java
```diff
@@ -0,0 +1,140 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Matchers.eq;
+import static org.mockito.Mockito.mock;
+
+/**
+ * @author João André Martins
+ */
+@RunWith(SpringJUnit4ClassRunner.class)
+@ContextConfiguration
+public class GCSInboundFileSynchronizerTest {
```
artembilan: We prefer the `*Tests` style  - plural: https://github.com/spring-projects/spring-framework/wiki/Code-Style#tests

spring-integration-gcp/src/test/java/org/springframework/integration/gcp/storage/inbound/GCSInboundFileSynchronizerTest.java
```diff
@@ -0,0 +1,140 @@
...
+	public static class Config {
+
+		@Bean
+		public Storage gcs() {
+			Storage gcsMock = mock(Storage.class);
+
+			Blob blob1 = mock(Blob.class);
+			Blob blob2 = mock(Blob.class);
+
+			BDDMockito.willAnswer(invocation -> "legend of heroes").given(blob1).getName();
```
artembilan: This can be statically imported.
Yes, nit-pick, but again code style 😄

spring-integration-gcp/pom.xml
joaoandremartins: Why don't you think it would be good to have separate sub-modules like `spring-integration-pubsub` and `spring-integration-storage`? The `spring-integration` project is modular itself (e.g., `spring-integration-amqp`, `spring-integration-jms`, etc.)
Asking because I meant to split the `spring-integration-gcp` module into two after this PR.

spring-integration-gcp/src/main/java/org/springframework/integration/gcp/storage/GCSFileInfo.java
joaoandremartins: Makes sense.

spring-integration-gcp/pom.xml
artembilan: Right, it is but, pay attention how all the protocol related stuff is collected under the single module, like AWS or JMX or XML. In thins case it is GCP protocol, so I tend to see all the Spring Integration GCP stuff in a single module.

Any other opinions?

spring-integration-gcp/src/main/java/org/springframework/integration/gcp/storage/GCSSession.java
joaoandremartins: The way I understand it is `remove()` can be used to delete either GCS blobs or buckets.
You're right that the `2` in `path.split("/", 2)` needs to be a `3` so that, for `/bucket/path/blob`, the result is `["", "bucket", "path/blob"]`.

spring-integration-gcp/src/main/java/org/springframework/integration/gcp/storage/inbound/GCSInboundFileSynchronizingMessageSource.java
joaoandremartins: Good catch!

spring-integration-gcp/src/test/java/org/springframework/integration/gcp/storage/inbound/GCSInboundFileSynchronizerTest.java
joaoandremartins: I'll add this new `org.mockito.BDDMockito` package to the checkstyle rules.

spring-integration-gcp/src/main/java/org/springframework/integration/gcp/storage/GCSSession.java
artembilan: ???
If you would like to remove just bucket, you have to use just bucket name and we will have `tokens.length == 1`. If you would like to remove a blob, the path should be with bucket and blob over a `/` . Therefore `tokens.length == 2`, but never `3`.

I don't understand everything rest you try to explain to me, but I would like to see a test case for this remove, for both `bucket` and `bucket/path/blob` use-cases.

Thanks

spring-integration-gcp/src/main/java/org/springframework/integration/gcp/storage/GCSSession.java
```diff
@@ -0,0 +1,192 @@
...
+	}
+
+	@Override
+	public Object getClientInstance() {
+		return this.gcs;
+	}
+
+	private String[] getBucketAndObjectFromPath(String path) throws IOException {
+		// Assumes paths of the form /bucket/folder/blob
+		String[] tokens = path.split("/", 2);
```
joaoandremartins: It would be good to have a way of setting the separator here, since `AbstractRemoteFileStreamingMessageSource` allows for different separators, but its conceptually disconnected from our `GcsSession`..

spring-integration-gcp/pom.xml
joaoandremartins: OK cool, I'll keep everything under the same module for now!

spring-integration-gcp/src/main/java/org/springframework/integration/gcp/storage/GCSSession.java
joaoandremartins: Sorry for the confusion. I thought the paths in there would be `/bucket/file`, but there is no starting `/`, so 3 doesn't make sense.

spring-integration-gcp/src/main/java/org/springframework/integration/gcp/storage/GcsSession.java
```diff
@@ -189,6 +191,10 @@ public Object getClientInstance() {
 		// Assumes paths of the form bucket/folder/blob
 		Assert.hasText(path, "Path can't be empty.");
 
-		return path.split("/", 2);
+		return path.split(this.remoteFileSeparator, 2);
```
artembilan: I wonder if this is something what is allowed by Google Storage. I mean AWS S3 allows only `/` as a virtual directory separator.

WDYT?

spring-integration-gcp/src/main/java/org/springframework/integration/gcp/storage/inbound/GcsInboundFileSynchronizingMessageSource.java
```diff
@@ -0,0 +1,44 @@
...
+	}
+
+	public GcsInboundFileSynchronizingMessageSource(GcsInboundFileSynchronizer synchronizer,
+			Comparator<File> comparator) {
+		super(synchronizer, comparator);
+	}
+
+	@Override
+	public String getComponentType() {
+		return "gcp:gcs-inbound-streaming-channel-adapter";
```
artembilan: This has not been addressed yet.

spring-cloud-gcp-docs/src/main/asciidoc/spring-integration-storage.adoc
```diff
@@ -0,0 +1,89 @@
+=== Channel Adapters for Google Cloud Storage
+
+The channel adapters for Google Cloud Storage allow you to read and write files to Google Cloud
+Storage through `MessageChannels`.
+
+Spring Cloud GCP provides two inbound adapters, `GCSInboundFileSynchronizingMessageSource` and
```
artembilan: Since you have renamed target classes the docs must be fixed as well

spring-integration-gcp/src/main/java/org/springframework/integration/gcp/storage/outbound/GcsMessageHandler.java
```diff
@@ -0,0 +1,42 @@
...
+
+import org.springframework.integration.file.remote.RemoteFileTemplate;
+import org.springframework.integration.file.remote.handler.FileTransferringMessageHandler;
+import org.springframework.integration.file.remote.session.SessionFactory;
+import org.springframework.integration.file.support.FileExistsMode;
+
+/**
+ * @author João André Martins
+ */
+public class GcsMessageHandler extends FileTransferringMessageHandler<BlobInfo> {
```
artembilan: I wonder if we can implement `doChmod()` here via that `createAcl(BlobId blob, Acl acl)`.
But that is already different story. Just do not forget that we may have a hook.

spring-integration-gcp/src/main/java/org/springframework/integration/gcp/storage/GcsSession.java
joaoandremartins: That's a good point.

Although the paths we're getting here can come from e.g., `AbstractRemoteFileStreamingMessageSource` which uses a `remoteFileSeparator` that can be set to something else.
But it's probably safe to assume that this wouldn't be set to anything else other than  `/`, so let's rollback this bit for now.

artembilan: We can override `setRemoteFileSeparator()` there with the `UnsupportedOperationException`.

Out of topic: GH inhibits for me here, especially when I comment, like there is some hard JavaScript.
Don't you know what that might be?

spring-integration-gcp/src/main/java/org/springframework/integration/gcp/storage/outbound/GcsMessageHandler.java
joaoandremartins: I'm failing to see a way to map `chmod` permissions to ACLs.
chmod allows the user, its group or everyone to do reads/writes/executes, where ACLs let specific usernames (i.e., Google user accounts), so it doesn't translate very well, but I may not be seeing the whole picture here.

On top of that, I don't really understand why `doChmod()` is a method of `FileTransferringMessageHandler` and not `Session`.

We can leave this as a TODO or open an issue to think about this afterwards?

spring-integration-gcp/src/main/java/org/springframework/integration/gcp/storage/GcsSession.java
joaoandremartins: Do you mean GH is hiding the comments?
If so, that may be because in my last commits I changed the lines being commented.

Unless you mean something else?

artembilan: Yeah.. Sure. I mean it is very slow and doesn't let me to comment properly like it is reloading page on each my key press

joaoandremartins:
@artembilan anything else on this PR?

spring-integration-gcp/src/main/java/org/springframework/integration/gcp/storage/GcsSession.java
```diff
@@ -0,0 +1,196 @@
...
+import org.apache.commons.logging.LogFactory;
+
+import org.springframework.integration.file.remote.session.Session;
+import org.springframework.util.Assert;
+import org.springframework.util.StreamUtils;
+
+/**
+ * @author João André Martins
+ */
+public class GcsSession implements Session<BlobInfo> {
```
meltsufin: I think I'm missing some context here, but having to implement the whole `Session` interface seems a bit of an overkill here. It looks like a lot of the functionality is similar to `GoogleStorageResourceObject`. For example, why is support for renaming and deleting needed to implement the channel adapters?

spring-cloud-gcp-docs/src/main/asciidoc/spring-integration-storage.adoc
```diff
@@ -0,0 +1,89 @@
...
+Spring Cloud GCP provides two inbound adapters, `GcsInboundFileSynchronizingMessageSource` and
+`GcsStreamingMessageSource`, and one outbound adapter, `GcsMessageHandler`.
+
+Maven coordinates, using Spring Cloud GCP BOM:
+
+[source,xml]
+----
+<dependency>
+    <groupId>org.springframework.cloud</groupId>
+    <artifactId>spring-integration-gcp</artifactId>
```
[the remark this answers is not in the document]
artembilan: Agreed

spring-cloud-gcp-docs/src/main/asciidoc/spring-integration-storage.adoc
```diff
@@ -0,0 +1,89 @@
...
+The Google Cloud Storage inbound channel adapter polls a Google Cloud Storage bucket for new files and sends each of
+them in a `Message` payload to the `MessageChannel` specified in the `@InboundChannelAdapter` annotation.
+The files are temporarily stored in a folder in the local file system.
+
+Here is an example of how to configure a Google Cloud Storage inbound channel adapter.
+
+[source,java]
+----
+@Bean
+@InboundChannelAdapter(channel = "new-file-channel", poller = @Poller(fixedDelay = "5000"))
```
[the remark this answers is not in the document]
artembilan: You don't need that `newFileChannel` bean definition. SI creates `DirectChannel` beans automatically by the `@ServiceActivator` parsing.

spring-integration-gcp/src/main/java/org/springframework/integration/gcp/storage/GcsSession.java
artembilan: This is how those GCS channel adapters are based. There is a base functionality in the `spring-integration-file` for remote files processing and it comes with the `Session` and `RemoteFileTemplate` abstractions.

I would say it is overkill to have that `GoogleStorageResourceObject` since all the storage functionality is covered by the `RemoteFileTemplate` and appropriate `Session` implementation.

artembilan:
>  So, if you have 100 files in the bucket, you'll only see them all after 500 seconds elapse!?

Well, first of all it is `pollable` channel adapter. It polls the target protocol for data via provided `fixedDelay` itnerval. And there is also `maxMessagesPerPoll`:
```
	/**
	 * @return The maximum number of messages to receive for each poll.
	 * Can be specified as 'property placeholder', e.g. {@code ${poller.maxMessagesPerPoll}}.
	 * Defaults to -1 (infinity) for polling consumers and 1 for polling inbound channel adapters.
	 */
	String maxMessagesPerPoll() default "";
```
So,it is `1` for your by default. You can configure it into `-1` to get all the remote files in one polling cycle.

More info is in SI Docs: https://docs.spring.io/spring-integration/docs/5.0.0.RELEASE/reference/html/messaging-channels-section.html#polling-consumer

spring-cloud-gcp-docs/src/main/asciidoc/spring-integration-storage.adoc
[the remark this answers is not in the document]
meltsufin: Ok, I didn't know that, but then the `@ServiceActivator` should be in the example.

spring-integration-gcp/src/main/java/org/springframework/integration/gcp/storage/GcsSession.java
meltsufin: Fair enough, but in either case we should find a way to share the implantation. @joaoandremartins Can you create an issue for that?

meltsufin:
> So,it is `1` for your by default. You can configure it into `-1` to get all the remote files in one polling cycle.

Thanks! That makes sense.

spring-cloud-gcp-docs/src/main/asciidoc/spring-integration-storage.adoc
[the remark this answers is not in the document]
artembilan: Not a fact, but let it be. That might be done properly in the appropriate sample module, but seems like overhead in the docs just for the channel adapter.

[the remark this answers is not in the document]
joaoandremartins: +1 for providing a more complete sample later.

spring-integration-gcp/src/main/java/org/springframework/integration/gcp/storage/GcsSession.java
joaoandremartins: @meltsufin I don't think we want to create the Spring Integration dependency in the Spring Storage module. WDYT?

spring-integration-gcp/README.adoc
```diff
@@ -34,3 +36,74 @@ These are useful to process the message ID, in case of success, or the error if
 
 Refer to the https://spring.io/guides/gs/spring-cloud-gcp/[Spring getting started guide] to learn
 how to use the adapters.
+
+== Google Cloud Storage
```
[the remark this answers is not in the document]
joaoandremartins: Tried something like `1include::/spring-cloud-gcp-docs/src/main/asciidoc/spring-integration-storage.adoc[lines=33..-1]`, but it didn't work.
I think it might be because this README is processed differently from the reference document.
I don't know of another way to make this work..

joaoandremartins:
@meltsufin I think all your points were addressed, let me know if I missed anything.
@artembilan can you re-approve please?

spring-cloud-gcp-docs/src/main/asciidoc/spring-integration-storage.adoc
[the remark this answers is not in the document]
artembilan: Need to add `spring-integration-file` dependency since it isn't going to be polled transitively.
And really explain why. 
That's not addresses. 
Thanks to @meltsufin ! 😄

spring-integration-gcp/src/main/java/org/springframework/integration/gcp/storage/GcsSession.java
artembilan: I think the idea was exactly opposite: to reuse `GoogleStorageResourceObject` here instead of `GcsSession`.
But IMO that would be much harder to support in the future. As I said not one time: I would better stay away from the overkill in the `Resourse` implementation and would keep it just for simple read-only `@Value`. Everything rest can be done by special proprietary API. But you have done that already, so let's at least keep them separately to avoid more confuses. In most cases this `Session` is used from the SI Channel Adapters and there you really don't care what API is there underneath.

joaoandremartins: Yes I'm also much more comfortable doing it the Spring Integration way here and the Resource way in the Resource land.

spring-cloud-gcp-docs/src/main/asciidoc/spring-integration-storage.adoc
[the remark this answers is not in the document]
joaoandremartins: Added a short explanation of why the dependency is required.
Let me know if it's enough or if it needs more explanation.

spring-cloud-gcp-docs/pom.xml
```diff
@@ -90,6 +90,7 @@
 								<snippets>${basedir}/target/generated-snippets</snippets>
 								<github-tag>${github-tag}</github-tag>
 								<docinfo>true</docinfo>
+								<spring-integration-version>4.3.12.RELEASE</spring-integration-version>
```
artembilan: `4.3.13` is the latest in that line. 
Although there is already `5.0.0`: https://projects.spring.io/spring-integration

However I don't understand why do we need this explicit value here since we might be able to pick up it from the parent Spring Boot. Something like:
```
<spring-integration-version>${spring-integration.version}</spring-integration-version>
```

spring-cloud-gcp-docs/src/main/asciidoc/spring-integration-pubsub.adoc
```diff
@@ -0,0 +1,120 @@
...
+This enables messaging between different processes, applications or micro-services backed up by
+Google Cloud Pub/Sub.
+
+Maven coordinates, using Spring Cloud GCP BOM:
+
+[source,xml]
+----
+<dependency>
+    <groupId>org.springframework.cloud</groupId>
+    <artifactId>spring-integration-gcp</artifactId>
```
artembilan: I think `Pub/Sub` is `optional` now as well and, therefore, deserves some additions here, too.

spring-cloud-gcp-docs/src/main/asciidoc/spring-integration-storage.adoc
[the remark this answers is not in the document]
artembilan: Fine with me. Thanks

spring-cloud-gcp-docs/pom.xml
joaoandremartins: Yeah I was thinking the same, but `${spring-integration.version}` isn't recognised in the `spring-cloud-gcp-docs` `pom.xml`.
I tried adding spring-boot's BOM but it also doesn't work...

spring-cloud-gcp-docs/src/main/asciidoc/spring-integration-storage.adoc
```diff
@@ -0,0 +1,99 @@
...
+</dependency>
+----
+
+Gradle coordinates:
+
+[source,subs="normal"]
+----
+dependencies {
+    compile group: 'org.springframework.cloud', name: 'spring-cloud-gcp-integration', version: '{project-version}'
+    compile group: 'org.springframework.integration', name: 'spring-integration-file', version: '{spring-integration-version}'
```
artembilan: Well, I think we don't need version here at all - Spring Integration is managed by Spring Boot. Only what you need is to add appropriate `spring-boot` plugin to your Gradle config. 
Anyway I would leave all of these tricks and hooks up to end-user and appropriate project references like Spring Boot.
That is not our goal here to cover all the Spring portfolio in one doc.

spring-cloud-gcp-docs/pom.xml
```diff
@@ -90,6 +90,7 @@
 								<snippets>${basedir}/target/generated-snippets</snippets>
 								<github-tag>${github-tag}</github-tag>
 								<docinfo>true</docinfo>
+								<spring-integration-version>5.0.0.RELEASE</spring-integration-version>
```
artembilan: BTW. Doesn't it work only in IDE or at runtime when you build the project?
Maybe we can have a project property with the reference to that SI version property in Boot and use it from here as a link.
Or... just don't use SI version in Docs at all since it is managed by Boot anyway.

More over you can't use here SI-5.0 because you don't use Boot 2.0 and, therefore, SF-5.0.
And I guess you can't use until Spring Cloud 2.0 anyway. Although that raises the question if we should consider to switch to Spring Cloud 2.0/Boot 2.0 already now and develop Binder on a fresh Spring Cloud Stream foundation...

spring-cloud-gcp-docs/src/main/asciidoc/spring-integration-storage.adoc
joaoandremartins: Indeed, which is why I prefer to add the version straight up.

One thing is that while we don't upgrade to SI 5 on our side, we should probably recommend users to use 4.3.13.
I might settle for 4.3.13.

spring-integration-gcp/src/main/java/org/springframework/integration/gcp/storage/GcsSession.java
```diff
@@ -0,0 +1,196 @@
...
+		for (Blob blob : this.gcs.list(path).iterateAll()) {
+			blobs.add(blob);
+		}
+
+		return blobs.toArray(new BlobInfo[blobs.size()]);
+	}
+
+	@Override
+	public void read(String source, OutputStream outputStream) throws IOException {
+		String[] tokens = getBucketAndObjectFromPath(source);
```
ChengyuanZhao: Is a check needed here for bucket-only paths? A few lines down you use use `tokens[1]` but that wouldn't be there if the path is just a bucket.

spring-cloud-gcp-docs/src/main/asciidoc/spring-integration-storage.adoc
artembilan: ??? That's not good. You eliminate the purpose of Spring Boot at all.

And again: that is the part which you would have to support in the future - fix the version from upgrade to upgrade.

spring-integration-gcp/src/main/java/org/springframework/integration/gcp/storage/GcsSession.java
```diff
@@ -0,0 +1,196 @@
...
+
+		try (OutputStream os = outputStream) {
+			os.write(this.gcs.readAllBytes(tokens[0], tokens[1]));
+		}
+	}
+
+	@Override
+	public void write(InputStream inputStream, String destination) throws IOException {
+		String[] tokens = getBucketAndObjectFromPath(destination);
+		BlobInfo gcsBlobInfo = BlobInfo.newBuilder(BlobId.of(tokens[0], tokens[1])).build();
```
ChengyuanZhao: I might be missing the assumption that the `destination` path must be an object and not a folder/bucket, but if so perhaps throw an exception if the path given is  bucket-only?

spring-cloud-gcp-docs/src/main/asciidoc/spring-integration-storage.adoc
joaoandremartins: I thought you were suggesting against advising for tricks like using the gradle bom plugin here.
Which only leaves us with the option of leaving the version, if I understand correctly?

If there is a way to get rid of the version here other than adding gradle plugins, I would like to pursue it!

spring-cloud-gcp-docs/src/main/asciidoc/spring-integration-storage.adoc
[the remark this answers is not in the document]
joaoandremartins: Since we're being exhaustive here, I guess we should also add `google-cloud-storage` to the dependencies.

spring-integration-gcp/src/main/java/org/springframework/integration/gcp/storage/GcsSession.java
```diff
@@ -0,0 +1,196 @@
...
+				.map(BlobInfo::getName)
+				.collect(Collectors.toList());
+
+		return names.toArray(new String[names.size()]);
+	}
+
+	@Override
+	public InputStream readRaw(String source) throws IOException {
+		String[] tokens = getBucketAndObjectFromPath(source);
+		return Channels.newInputStream(this.gcs.reader(tokens[0], tokens[1]));
```
ChengyuanZhao: This one too would need to check that the path is not bucket-only right? 

Would it be worthwhile to make a private function that checks if a path is an object path for those functions that only work on objects and not buckets?

spring-cloud-gcp-docs/src/main/asciidoc/spring-integration-storage.adoc
artembilan: 😕 . I said:
> we don't need version here at all

AFAIK there is no other way for Gradle. However you still stick with something for Maven as well. 
And don't forget that there are some other building and dependency management tools and you don't cover them here. I would stick away not only from versions (see my comment about future support), but also from the dependency management tool. Any way it is not our project responsibility to demonstrate how to manage deps and their versions. Too much noise and no concentration on the purpose of the feature we are describing. As a middle ground we may send people to Spring Boot docs on the matter - in most cases people these days start from there for their projects.

spring-integration-gcp/src/main/java/org/springframework/integration/gcp/storage/GcsSession.java
```diff
@@ -0,0 +1,196 @@
...
+	@Override
+	public void append(InputStream inputStream, String destination) throws IOException {
+		// TODO(joaomartins): We could do compose here, but it assumes that InputStream is first copied to a
+		// GCS object that we can then "compose" to the original object.
+		throw new UnsupportedOperationException("Appending isn't supported by Google Cloud Storage.");
+	}
+
+	@Override
+	public boolean mkdir(String directory) throws IOException {
+		try {
```
ChengyuanZhao: For this and rmdir, should it be checked that the directory is actually just a bucket name and not a folder with slashes? I know folders aren't like a real thing in GCS, but if some other implementations of Session allow for actual folders with slashes then people could be annoyed by GCS not supporting it and they would need to change their code. Perhaps we could make an empty nameless file in the desired folder?

artembilan:
@ChengyuanZhao ,

I think you need to re-add your comments on the fresh code lines. Right now they are in the `outdated` section.

Thanks

spring-cloud-gcp-docs/src/main/asciidoc/spring-integration-storage.adoc
joaoandremartins: The goal has always been for users to be able to copy and paste snippets from the refdoc to their apps. That wouldn't work with Gradle if we were to omit versions.

The reason the Maven snippets don't contain versions and Gradle ones do is because [we say we will omit them](https://docs.spring.io/spring-cloud-gcp/docs/1.0.0.M1/reference/htmlsingle/#_dependency_management) earlier on.

BOM isn't a Gradle concept, so I chose to show the versions for those.

We can change that and hide versions for Gradle as well if we want to, but I'd prefer to do that in another PR.

joaoandremartins: As a matter of fact, now that I'm having to also manually add the `google-cloud-storage` version, I think I'll want to say in the dependency management chapter that we will hide versions for Gradle dependencies too.
I'll open an issue for that.

spring-cloud-gcp-docs/pom.xml
joaoandremartins: I'll get rid of the versions after this PR.
https://github.com/spring-cloud/spring-cloud-gcp/issues/279

spring-cloud-gcp-docs/src/main/asciidoc/spring-integration-storage.adoc
artembilan: Well, since you show there how to use BOM in Maven that's not a big deal to show the same for Gradle I guess, since you already mention `dependency-management-plugin`. 
And I don't think that RM is a copy/paste guide. You definitely still miss a lot of pieces in these samples and end-user still has to do something, for example you don't explain where to place all those dependencies etc. etc. We have samples applications in the end! Why just don't point there rather than pursue this pain how to resolve current SI version for our Gradle snippet in the manual?

That's my last comment on the matter and my general idea here is to avoid routine support as much as possible. 

And that's somehow too much confronting points of view how project should be developed...

artembilan: Oh! Sorry, forgot to mention. If you still insist, let it be: I really spent already too much time to convince you in different vision. Gone to do some other useful work...

Sorry for the noise

joaoandremartins: @artembilan you convinced me!
I hated having to keep those versions around and manually maintaining in the first place. I'll take care of it after this PR, as I said.

spring-integration-gcp/src/main/java/org/springframework/integration/gcp/storage/GcsSession.java
joaoandremartins: The reason I didn't add the check is because the SI infrastructure never called `read()` on buckets in my tests.
But I guess it can't hurt to add one.

spring-integration-gcp/src/main/java/org/springframework/integration/gcp/storage/GcsSession.java
joaoandremartins: The AWS implementation assumes that a remote directory is a bucket, I think it's safe to assume the same thing on our side.
We can always change it later if we need to.

artembilan: That's correct. The story is exactly about top level bucket. Because there is no a notion like sub-dir. 
We can just mention that in the JavaDocs here to protect ourselves from similar questions in the future. In addition we may mention that in the Reference Manual if we document this `GcsSession` there at all...

joaoandremartins:
FYI I'm aiming for a GA release mid-March.

Before going GA, we need user adoption, which is really hard if the code isn't committed to GH :)

It's already great that this will be released for M2, which we're targeting to early-mid January.

joaoandremartins:
Dismissing @meltsufin since I believe all points were covered. We can fix anything else with further PRs.

artembilan:
>  I'm aiming for a GA release mid-March.

Hm. Is it confirmed that all our dependencies are going to be GA that time already?

OTOH: I see that you move Google dependencies to the latest versions without any doubts.
I wonder what makes you against Spring then...

joaoandremartins:
We'll go GA with whatever we have that is GAed by then. :)

The problem with our dependencies is that we don't have any GA ones that we can rely on. Fortunately, the same can't be said about Spring. However, #201 aims at making that change, we can get to it earlier, on a side branch or so.

Issue #245, opened 2017-12-02
Spring Cloud GCP Config - don't validate properties when disabled


saturnism:
https://github.com/spring-cloud/spring-cloud-gcp/blob/master/spring-cloud-gcp-starters/spring-cloud-gcp-starter-config/src/main/java/org/springframework/cloud/gcp/config/GoogleConfigPropertySourceLocator.java

It's currently checking name/profile values to be set regardless of whether the starter has been disabled.  If disabled, it shouldn't force configure these values.

PR #252, opened 2017-12-05
Avoid checks when spring-cloud-gcp-config is disabled

saturnism:
For #245

### DIFF
--- a/spring-cloud-gcp-starters/spring-cloud-gcp-starter-config/src/test/java/org/springframework/cloud/gcp/config/GcpConfigAutoConfigurationTest.java
+++ b/spring-cloud-gcp-starters/spring-cloud-gcp-starter-config/src/test/java/org/springframework/cloud/gcp/config/GcpConfigAutoConfigurationTest.java
@@ -24,7 +24,7 @@ import org.springframework.cloud.gcp.config.autoconfig.GcpConfigAutoConfiguratio
 import org.springframework.context.annotation.AnnotationConfigApplicationContext;
 
 import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
 
 /**
  * @author Jisha Abubaker
@@ -54,13 +54,13 @@ public class GcpConfigAutoConfigurationTest {
 		loadEnvironment("spring.cloud.gcp.config.name=myapp",
 				"spring.cloud.gcp.config.profile=prod",
 				"spring.cloud.gcp.config.timeout=120000",
-				"spring.cloud.gcp.config.enabled=false",
+				"spring.cloud.gcp.config.enabled=true",
 				"spring.cloud.gcp.config.project-id=pariah");
 		GcpConfigProperties config = this.context.getBean(GcpConfigProperties.class);
 		assertEquals(config.getName(), "myapp");
 		assertEquals(config.getProfile(), "prod");
 		assertEquals(config.getTimeout(), 120000);
-		assertFalse(config.isEnabled());
+		assertTrue(config.isEnabled());
 		assertEquals(config.getProjectId(), "pariah");
 	}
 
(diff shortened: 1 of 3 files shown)


spring-cloud-gcp-starters/spring-cloud-gcp-starter-config/src/test/java/org/springframework/cloud/gcp/config/GcpConfigAutoConfigurationTest.java
```diff
@@ -54,13 +54,13 @@ public void testConfigurationValuesAreCorrectlyLoaded() {
...
 				"spring.cloud.gcp.config.timeout=120000",
-				"spring.cloud.gcp.config.enabled=false",
+				"spring.cloud.gcp.config.enabled=true",
 				"spring.cloud.gcp.config.project-id=pariah");
 		GcpConfigProperties config = this.context.getBean(GcpConfigProperties.class);
 		assertEquals(config.getName(), "myapp");
 		assertEquals(config.getProfile(), "prod");
 		assertEquals(config.getTimeout(), 120000);
-		assertFalse(config.isEnabled());
+		assertTrue(config.isEnabled());
```
joaoandremartins: Why this? `enabled` is `true` by default, testing if setting it to `false` works could make more sense.

meltsufin: He set it to `true` above. So, the corresponding change to the assertion makes sense. If we want to test if setting to `false` works, that could be done in `disabledPropertySourceAvoidChecks` below.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-config/src/test/java/org/springframework/cloud/gcp/config/GoogleConfigPropertySourceLocatorTest.java
```diff
@@ -84,6 +84,15 @@ public void disabledPropertySourceReturnsNull() throws Exception {
 		verify(this.googleConfigPropertySourceLocator, never()).getRemoteEnvironment();
 	}
 
+	@Test
+	public void disabledPropertySourceAvoidChecks() throws Exception {
+		this.gcpConfigProperties.setEnabled(false);
```
meltsufin: Why not use `loadEnvironment("spring.cloud.gcp.config.enabled=false")` instead?

artembilan: This is different test class and there is no such a method yet.
I think this is kinda unit test, so that's fine.
However I agree that `GcpConfigAutoConfigurationTest` should have one more test about the fully blown Spring Boot environment for the `spring.cloud.gcp.config.enabled=false` case.

joaoandremartins: I'm not too sure what you mean with that additional test case.
Can you explain?

artembilan: OK. This test class is about unit-tests, just for the `GoogleConfigPropertySourceLocator` functionality verifications. There is no `ApplicationContext` involved, so @meltsufin 's `loadEnvironment()` cannon be here. 
We can add a requested test-case, fully blown Boot test, into the `GcpConfigAutoConfigurationTest`. There is one for the `spring.cloud.gcp.config.enabled=true`, but there is not already one for the `spring.cloud.gcp.config.enabled=false`.
That's what I mean. Does it make sense now?

artembilan:
Closed in favor of #260

I hope now it is right 😄 

I won't come back to you today since it looks like I'm not in a good shape for proper review 😉

PR #256, opened 2017-12-11
allows user to access the Cloud Storage object if he or she wants to set attributes specific to GCS. 

### LINKED ISSUE #250: GoogleStorageResource - add ability to set content-type and other GCS attributes
GCS objects can have different attributes such as content-type, allow this to be set if user casts the resource to GoogleStorageResource.

ChengyuanZhao:
fixes #250


spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceObject.java
```diff
@@ -109,7 +110,12 @@ private BlobId getBlobId() throws IOException {
 				uri.getPath().substring(1, uri.getPath().length()));
 	}
 
-	private Blob resolve() throws IOException {
+	/**
+	 * Gets the underlying storage object in Cloud Storage.
```
joaoandremartins: nit: Google Cloud Storage or GCS

joaoandremartins:
Can we add some documentation around what the user can do with this?

Considering #250, I'm thinking documenting e.g., a way to set content type to JPEG for an image.

ChengyuanZhao:
How about referring/linking to the GCS devsite in the documentation? I would prefer to not teach GCS-specific things that don't particularly have anything to do with Spring in documentation here if they're already in GCS docs and also might change in future versions.

joaoandremartins:
My suggestion is to document a simple example of what you can do in the refdoc and GitHub README. Then, if you want, you can point to the devsite documentation. Although we don't want to depend to much on GCP APIs, so I would actually be OK if we didn't even link to devsite.

ChengyuanZhao:
Ah I see, will do.

artembilan:
> . Although we don't want to depend to much on GCP APIs

That's not correct. We develop here just only some Spring abstractions over existing API and we do that just because that one isn't so convenient, but Spring never relegate the API it its based on. So, that's always good to link to documentation from Spring and even mention how to get access to the target API from our Spring abstraction. Otherwise it sounds like we develop something different and encourage end-users do not look into target API at all.

joaoandremartins:
My fear is that APIs change between versions and users get locked in to our APIs, which I thought Spring wanted to prevent.

artembilan:
Well, in that case we don't have choice unless release our new version based on the new target API. That's already not Spring responsibility how target API works. And to be honest: I was against this "fancy" `@Value` API from the beginning: just let end-user to get read access is fully enough for this level of abstraction. If he/she wants to write or change the target resource, there is some other API. And one of them might be like `GcsRemoteFileTemplate` - similar to `(S3)((S)Ftp)RemoteFileTemplate`. However again: we don't restrict users there in access to low-level client:
```
/**
 * Get the underlying client library's client instance for this session.
 * Returns an {@code Object} to avoid significant changes to -file, -ftp, -sftp
 * modules, which would be required
 * if we added another generic parameter. Implementations should narrow the
 * return type.
 * @return The client instance.
 * @since 4.1
 */
Object getClientInstance();
```

joaoandremartins:
The only other Spring API that I know of is Spring Integration, and the learning curve might be too steep just to allow file writes.
I'm fine with delegating to the original API for now.

ChengyuanZhao:
This just got very philosophical. It could be an easy fix regardless of what we do in the future as far as this PR goes. PTAL at the README additions.

spring-cloud-gcp-examples/spring-cloud-gcp-storage-resource-example/src/main/java/com/example/WebController.java
```diff
@@ -38,7 +38,7 @@
 @RestController
 public class WebController {
 
-	@Value("gs://REPLACE_WITH_YOUR_BUCKET/my-file.txt")
+	@Value("gs://cccz-test-7/")
```
joaoandremartins: I guess the changes on this file were committed by accident? :laughing: 
Saying because of the real bucket and the commented code.

ChengyuanZhao: Woops!

spring-cloud-gcp-examples/spring-cloud-gcp-storage-resource-example/src/main/java/com/example/WebController.java
```diff
@@ -48,10 +48,15 @@ public String readGcsFile() throws IOException {
...
+	@RequestMapping(value = "/write", method = RequestMethod.GET)
+	String writeGcs() throws IOException {
+		/*
 		try (OutputStream os = ((WritableResource) this.gcsFile).getOutputStream()) {
-			os.write(data.getBytes());
+			os.write("AHAHAHA".getBytes());
+		}
+		*/
+		try (OutputStream os = ((WritableResource) this.gcsFile.createRelative("/test2.txt")).getOutputStream()) {
+			os.write("AHAHAHA".getBytes());
```
joaoandremartins: nit:s/AHAHAHA/BFG9000 ?

spring-cloud-gcp-examples/spring-cloud-gcp-storage-resource-example/src/main/java/com/example/WebController.java
```diff
@@ -48,16 +48,11 @@ public String readGcsFile() throws IOException {
...
-		}
-		*/
-		try (OutputStream os = ((WritableResource) this.gcsFile.createRelative("/test2.txt")).getOutputStream()) {
-			os.write("AHAHAHA".getBytes());
+			os.write(data.getBytes());
 		}
 		return "file was updated\n";
 	}
-}
+}
```
joaoandremartins: Add newline

PR #260, opened 2017-12-14
 Avoid checks when spring-cloud-gcp-config is disabled

### LINKED ISSUE #245: Spring Cloud GCP Config - don't validate properties when disabled

joaoandremartins:
Fixes #245


artembilan:
??? What is this about?
Looks like a duplication of the #252

joaoandremartins:
Indeed, but @saturnism went on vacations and the tests were failing on his PR, so we couldn't merge.

artembilan:
Closed in favor of #260

artembilan:
Oh! Sorry, I've mixed this one with the just merged fix for SCOPES. Shame on me 😢

joaoandremartins:
@artembilan test added. PTAL

PR #261, opened 2017-12-14
get signed URLs for GCS objects

### LINKED ISSUE #251: GoogleStorageResource - Signed URL
Produce signed URL for GoogleStorageResource. This is useful when sending the GCS object to end user w/o having to proxy the request in the server.  Can send signed URL to user easily.

ChengyuanZhao:
fixes #251

### DIFF
--- a/spring-cloud-gcp-docs/src/main/asciidoc/storage.adoc
+++ b/spring-cloud-gcp-docs/src/main/asciidoc/storage.adoc
@@ -54,7 +54,7 @@ If the resource path refers to an object on Google Cloud Storage (as opposed to 
 can be cast as a `GoogleStorageResourceObject` and the `getGoogleStorageObject` method can be called
 to obtain a https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java[`Blob`].
 This type represents a GCS file, which has associated https://cloud.google.com/storage/docs/gsutil/addlhelp/WorkingWithObjectMetadata[metadata], such as content-type, that can be set.
-The `getSignedUrl` method can also be used to obtain signed URLs for GCS objects.
+The `createSignedUrl` method can also be used to obtain https://cloud.google.com/storage/docs/access-control/signed-urls[signed URLs] for GCS objects.
 
 
 The Spring Boot Starter for Google Cloud Storage auto-configures the `Storage` bean required by the
--- a/spring-cloud-gcp-docs/src/main/asciidoc/storage.adoc
+++ b/spring-cloud-gcp-docs/src/main/asciidoc/storage.adoc
@@ -54,7 +54,7 @@ If the resource path refers to an object on Google Cloud Storage (as opposed to 
 can be cast as a `GoogleStorageResourceObject` and the `getGoogleStorageObject` method can be called
 to obtain a https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java[`Blob`].
 This type represents a GCS file, which has associated https://cloud.google.com/storage/docs/gsutil/addlhelp/WorkingWithObjectMetadata[metadata], such as content-type, that can be set.
-
+The `getSignedUrl` method can also be used to obtain signed URLs for GCS objects.
 
 
 The Spring Boot Starter for Google Cloud Storage auto-configures the `Storage` bean required by the
--- a/spring-cloud-gcp-docs/src/main/asciidoc/storage.adoc
+++ b/spring-cloud-gcp-docs/src/main/asciidoc/storage.adoc
@@ -55,7 +55,7 @@ can be cast as a `GoogleStorageResourceObject` and the `getGoogleStorageObject` 
 to obtain a https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java[`Blob`].
 This type represents a GCS file, which has associated https://cloud.google.com/storage/docs/gsutil/addlhelp/WorkingWithObjectMetadata[metadata], such as content-type, that can be set.
 The `createSignedUrl` method can also be used to obtain https://cloud.google.com/storage/docs/access-control/signed-urls[signed URLs] for GCS objects.
-
+However, creating signed URLs requires that the resource was created using service account credentials.
 
 The Spring Boot Starter for Google Cloud Storage auto-configures the `Storage` bean required by the
 `spring-cloud-gcp-storage` module, based on the `CredentialsProvider` provided by the Spring Boot
--- a/spring-cloud-gcp-storage/README.adoc
+++ b/spring-cloud-gcp-storage/README.adoc
@@ -55,3 +55,4 @@ can be cast as a `GoogleStorageResourceObject` and the `getGoogleStorageObject` 
 to obtain a https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java[`Blob`].
 This type represents a GCS file, which has associated https://cloud.google.com/storage/docs/gsutil/addlhelp/WorkingWithObjectMetadata[metadata], such as content-type, that can be set.
 The `createSignedUrl` method can also be used to obtain https://cloud.google.com/storage/docs/access-control/signed-urls[signed URLs] for GCS objects.
+However, creating signed URLs requires that the resource was created using service account credentials.
--- a/spring-cloud-gcp-storage/README.adoc
+++ b/spring-cloud-gcp-storage/README.adoc
@@ -54,4 +54,4 @@ If the path refers to an object on Google Cloud Storage (as opposed to a bucket)
 can be cast as a `GoogleStorageResourceObject` and the `getGoogleStorageObject` method can be called
 to obtain a https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java[`Blob`].
 This type represents a GCS file, which has associated https://cloud.google.com/storage/docs/gsutil/addlhelp/WorkingWithObjectMetadata[metadata], such as content-type, that can be set.
-The `getSignedUrl` method can also be used to obtain signed URLs for GCS objects.
+The `createSignedUrl` method can also be used to obtain https://cloud.google.com/storage/docs/access-control/signed-urls[signed URLs] for GCS objects.
--- a/spring-cloud-gcp-storage/README.adoc
+++ b/spring-cloud-gcp-storage/README.adoc
@@ -54,3 +54,4 @@ If the path refers to an object on Google Cloud Storage (as opposed to a bucket)
 can be cast as a `GoogleStorageResourceObject` and the `getGoogleStorageObject` method can be called
 to obtain a https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java[`Blob`].
 This type represents a GCS file, which has associated https://cloud.google.com/storage/docs/gsutil/addlhelp/WorkingWithObjectMetadata[metadata], such as content-type, that can be set.
+The `getSignedUrl` method can also be used to obtain signed URLs for GCS objects.
--- a/spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceObject.java
+++ b/spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceObject.java
@@ -17,6 +17,7 @@
 package org.springframework.cloud.gcp.storage;
 
 import java.io.File;
+import java.io.FileInputStream;
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
@@ -25,15 +26,19 @@ import java.net.URI;
 import java.net.URISyntaxException;
 import java.net.URL;
 import java.nio.channels.Channels;
+import java.util.concurrent.TimeUnit;
 
+import com.google.auth.oauth2.ServiceAccountCredentials;
 import com.google.cloud.storage.Blob;
 import com.google.cloud.storage.BlobId;
 import com.google.cloud.storage.BlobInfo;
 import com.google.cloud.storage.Storage;
+import com.google.cloud.storage.Storage.SignUrlOption;
 
 import org.springframework.core.io.Resource;
 import org.springframework.core.io.WritableResource;
 import org.springframework.util.Assert;
+import org.springframework.util.StringUtils;
 
 /**
  * Implements {@link WritableResource} for reading and writing objects in Google Cloud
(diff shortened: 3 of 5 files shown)


spring-cloud-gcp-storage/src/test/java/org/springframework/cloud/gcp/storage/GoogleStorageTests.java
```diff
@@ -260,6 +261,16 @@ public void testCreateRelativeSubdirs() throws IOException {
 		Assert.assertEquals("gs://test-spring/t1/r1/relative.png", relative.getURI().toString());
 	}
 
+	@Test
+	public void nullSignedUrlForNullBlob() throws IOException {
```
meltsufin: I think we need more tests to fully cover the method.
1) `IOException` when `privateKeyPath` is empty.
2) Verify that `this.storage.signUrl` is called in the happy path.

spring-cloud-gcp-docs/src/main/asciidoc/storage.adoc
```diff
@@ -54,7 +54,7 @@ If the resource path refers to an object on Google Cloud Storage (as opposed to
 can be cast as a `GoogleStorageResourceObject` and the `getGoogleStorageObject` method can be called
 to obtain a https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java[`Blob`].
 This type represents a GCS file, which has associated https://cloud.google.com/storage/docs/gsutil/addlhelp/WorkingWithObjectMetadata[metadata], such as content-type, that can be set.
-
+The `getSignedUrl` method can also be used to obtain signed URLs for GCS objects.
```
meltsufin: `getSignedUrl` -> `createSignedUrl`
Also, link to the documentation about signed URLs: https://cloud.google.com/storage/docs/access-control/signed-urls

spring-cloud-gcp-storage/README.adoc
```diff
@@ -54,3 +54,4 @@ If the path refers to an object on Google Cloud Storage (as opposed to a bucket)
 can be cast as a `GoogleStorageResourceObject` and the `getGoogleStorageObject` method can be called
 to obtain a https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java[`Blob`].
 This type represents a GCS file, which has associated https://cloud.google.com/storage/docs/gsutil/addlhelp/WorkingWithObjectMetadata[metadata], such as content-type, that can be set.
+The `getSignedUrl` method can also be used to obtain signed URLs for GCS objects.
```
meltsufin: Same comment as above.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceObject.java
```diff
@@ -119,6 +124,31 @@ public Blob getGoogleStorageObject() throws IOException {
 		return this.storage.get(getBlobId());
 	}
 
+	/**
+	 * Gets a signed URL to an object if it exists.
```
meltsufin: Gets -> Creates

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceObject.java
```diff
@@ -119,6 +124,31 @@ public Blob getGoogleStorageObject() throws IOException {
...
+	 * @param timeUnit the time unit used to determine how long the URL is valid.
+	 * @param timePeriods the number of periods to determine how long the URL is valid.
+	 * @return the URL if the object exists, and null if it does not.
+	 * @throws IOException
+	 */
+	public URL getSignedUrl(TimeUnit timeUnit, int timePeriods) throws IOException {
```
meltsufin: createSignedUrl

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceObject.java
```diff
@@ -119,6 +124,31 @@ public Blob getGoogleStorageObject() throws IOException {
...
+		Blob blob = this.getGoogleStorageObject();
+		if (blob == null) {
+			return null;
+		}
+		String privateKeyPath = System.getenv(
```
meltsufin: Can we use the auto-configured `CredentialsProvider` to get the private key instead? @joaoandremartins

ChengyuanZhao: Doesn't look like there's a way to get a private key from com.google.auth.Credentials. I'm not a security guy, but perhaps it's intentional that the only copy of the private key you're supposed to have is from the JSON file you download just once when creating the service account in the UI.

spring-cloud-gcp-storage/src/test/java/org/springframework/cloud/gcp/storage/GoogleStorageTests.java
ChengyuanZhao: 1. Will be not be nice to test, because env vars are not great to test with unit tests. Remember a few months ago we had issues with the CI stuff and tests that altered env vars. The GOOGLE_APPLICATION_CREDENTIALS env var is very widespread and set in stone, so I think this is the right way to access that private key JSON file. I think maybe we can just skip testing it. Even if this test would have failed, it's not a catastrophic or mysterious failure anyway. 

2. Will do.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceObject.java
meltsufin: What about the `GcpStorageProperties.credentials.location`? We already have a mechanism for specifying the private key file. Why use the environment variable instead?

spring-cloud-gcp-storage/src/test/java/org/springframework/cloud/gcp/storage/GoogleStorageTests.java
meltsufin: 1. Why can't we rely on `spring.cloud.gcp.storage.credentials.location`, which is easy to mock in the test?

spring-cloud-gcp-docs/src/main/asciidoc/storage.adoc
ChengyuanZhao: Done

spring-cloud-gcp-storage/README.adoc
ChengyuanZhao: Done

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceObject.java
ChengyuanZhao: Done

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceObject.java
ChengyuanZhao: Done

spring-cloud-gcp-storage/src/test/java/org/springframework/cloud/gcp/storage/GoogleStorageTests.java
ChengyuanZhao: Isn't that one only in the starter project? Also, other than a unit test of minimal (if any) value, what's the purpose? GOOGLE_APPLICATION_CREDENTIALS is much more widely known, and is mentioned directly in the documentation in this repo (not just the starter projects as the spring.cloud.gcp.xxxxx.credentials.location props).

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceObject.java
ChengyuanZhao: See my other comment. The prop you're thinking of is actually only in the starter projects, and infact the larger repo-level docs here specify using the env var. The env var is also standard across any GCP code, so I don't feel it's out of place (it's also mentioned to be set in this repo's docs anyway).

ChengyuanZhao: Actually I could make this path a member that we set in the constructor (like the auto-create setting last time), and that it would just default to GOOGLE_APPLICATION_CREDENTIALS. Still, we would have this env var involved, but we could at least use the path from the starter instead if it was provided.

meltsufin:
@ChengyuanZhao Also, I noticed that this PR does not add support for creating signed URLs for buckets. One way to fix it would be to just go back to having a single `GoogleStorageResource` class and represent a GCS resource and deal with bucket vs. object internally, as I suggested previously.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceObject.java
```diff
@@ -119,6 +125,42 @@ public Blob getGoogleStorageObject() throws IOException {
...
+	 */
+	public URL createSignedUrl(TimeUnit timeUnit, long timePeriods) throws IOException {
+		Blob blob = this.getGoogleStorageObject();
+		if (blob == null) {
+			return null;
+		}
+		String privateKeyPath = getGcpPrivateKeyPath();
+		return this.storage.signUrl(
+				BlobInfo.newBuilder(blob.getBucket(), blob.getName()).build(),
+				timePeriods, timeUnit, getSignUrlOptionForPrivateKey(privateKeyPath));
```
joaoandremartins: Why are you passing credentials here at all?
`this.storage` already contains the correct credentials, no need to pass them here again.

Plus, if for whatever reason they needed to be passed here again, you would depend on `GoogleCredentials`, not get the credentials on your own.

joaoandremartins:
My understanding of #251 was that when a blob is created, we should return (e.g., by logging) the GCS URL to users. Although now that I'm re-reading #251 I'm not totally sure what it's about.

Maybe I'd just add a log statement to `GoogleStorageResourceObject#createBlob()` logging the URL of the newly created object (same for buckets).

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceObject.java
ChengyuanZhao: What's needed is a ServiceAccountCredentials, which is a child of GoogleCredentials, and I'll try looking for if that is available at all, but it might not be, since none of the existing code requires specifically a service account credentials to work.

joaoandremartins:
On a second thought, I may have misinterpreted the meaning of signed URL.
So maybe disregard my comment about logging the signed URL on blob creation.

ChengyuanZhao:
@meltsufin 

I must say I feel completely the opposite. We had a previous PR that exposes setting metadata, which only makes sense for an object ( not a bucket), and that was elegantly handled by having the two separate types.

Also ,the function for getting signed url is from Storage, and only works for objects, not buckets, so again I feel our current setup with 2 objects is very fitting. I definitely don't think we should be implementing our own url building functionn for buckets especially since Storage only provides one for objects. not buckets.

ChengyuanZhao:
@meltsufin 

Ultimately it's because while a gs://bucket and gs://bucket/object path look very similar to a user, they represent very different things and have different functionality even from the user perspective, so I really don't think we should be hiding both behind a single type and instead warning the user at runtime that one operation is supported or not. For example, setting meta data or even getting a write stream.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceObject.java
joaoandremartins: I see what you mean.

I think the best way forward here is to add a `GoogleCredentials` argument to this method and then checking if it's an instance of `ServiceAccountCredentials` and, if so, cast it.

If it isn't, we should fail with clear instructions for the user (e.g., `createSignedUrl` only works if you're using service account creds).

joaoandremartins:
I'm also OK with having the two types vs single type and internal `ifs` (and I've read something on this recently in Effective Java where multiple types are encouraged vs internal ifs, Item 20 - Prefer class hierarchies to tagged classes).

In this case, this comes at the price of having a duplicated `createSignedUrl()`, which is OK for now. If duplication grows, we might think of doing something else.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceObject.java
ChengyuanZhao: Makes sense, will do.

meltsufin:
If it's true that you can't create a signed URL for a bucket, then this makes total sense and two classes are good.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceObject.java
```diff
@@ -119,6 +120,24 @@ public Blob getGoogleStorageObject() throws IOException {
...
+	 * @param timePeriods the number of periods to determine how long the URL is valid.
+	 * @return the URL if the object exists, and null if it does not.
+	 * @throws IOException
+	 */
+	public URL createSignedUrl(TimeUnit timeUnit, long timePeriods) throws IOException {
+		Blob blob = this.getGoogleStorageObject();
+		if (blob == null) {
+			return null;
+		}
+		return this.storage.signUrl(
```
joaoandremartins: Have you tested this change without using application default credentials?

From reading the `Storage#signUrl()` documentation, I think we do need to pass it a `SignUrlOptions` object containing the desired service account credentials.
In other words, I don't think the `Storage` object is smart enough to use the `GoogleCredentials` that it was passed in its constructor, but I'd be curious to test if this is true.

ChengyuanZhao: I did try it just now and it did make a working URL for me. This part of the doc makes it sound like it is aware of the credentials with which it was made:

 If the credentials passed to
   * {@link StorageOptions} do not implement {@link ServiceAccountSigner} (this is the case, for
   * instance, for Compute Engine credentials and Google Cloud SDK credentials) then {@code signUrl}
   * will throw an {@link IllegalStateException} unless an implementation of
   * {@link ServiceAccountSigner} is passed using the
   * {@link SignUrlOption#signWith(ServiceAccountSigner)} option.

spring-cloud-gcp-storage/src/test/java/org/springframework/cloud/gcp/storage/GoogleStorageTests.java
```diff
@@ -260,6 +261,30 @@ public void testCreateRelativeSubdirs() throws IOException {
...
+	public void signedUrlFunctionCalled() throws IOException {
+		String location = "gs://test-spring/t1/test.png";
+		Storage storage = Mockito.mock(Storage.class);
+		Blob blob = Mockito.mock(Blob.class);
+		Mockito.when(blob.getBucket()).thenReturn("fakeBucket");
+		Mockito.when(blob.getName()).thenReturn("fakeObject");
+		GoogleStorageResourceObject resource = new GoogleStorageResourceObject(storage, location);
+		Mockito.when(storage.get(Mockito.any(BlobId.class))).thenReturn(blob);
+		resource.createSignedUrl(TimeUnit.DAYS, 1L);
+		Mockito.verify(storage, Mockito.times(1))
```
joaoandremartins: nit: we've been using static imports in tests, you might want to do the same here for consistency.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceObject.java
joaoandremartins: Beautiful!

PR #264, opened 2017-12-15
Polish credentials configuration

snicoll:
This commit renames AbstractCredentialsProperty to Credentials and offer
a way to initialize it with default scopes. Rather than creating a class
just for the purpose of setting them, all configuration properties
simply reuse the class now.

Also, this commit fixes a potential bug as we were previously relying on
the fact that the Credentials object was initialized when a property was
set. Arguably, we could initialize it by setting a scope (and not a
location, leading to a NPE).

Finally, this commit generates proper metadata for the credentials
sub-namespaces.

### DIFF
--- a/spring-cloud-gcp-core/src/main/java/org/springframework/cloud/gcp/core/Credentials.java
+++ b/spring-cloud-gcp-core/src/main/java/org/springframework/cloud/gcp/core/Credentials.java
@@ -16,27 +16,26 @@
 
 package org.springframework.cloud.gcp.core;
 
+import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 
 import org.springframework.core.io.Resource;
 
 /**
- * To be extended by property class subclasses and allow e.g., a "credentials.location" property
- * notation.
+ * Credentials configuration
  *
  * @author João Andre Martins
+ * @author Stephane Nicoll
  */
-public abstract class AbstractCredentialsProperty {
+public class Credentials {
+
+	private List<String> scopes = new ArrayList<>();
 
 	private Resource location;
-	private List<String> scopes;
 
-	public Resource getLocation() {
-		return this.location;
-	}
-
-	public void setLocation(Resource location) {
-		this.location = location;
+	public Credentials(String... defaultScopes) {
+		this.scopes.addAll(Arrays.asList(defaultScopes));
 	}
 
 	public List<String> getScopes() {
@@ -46,4 +45,13 @@ public abstract class AbstractCredentialsProperty {
 	public void setScopes(List<String> scopes) {
 		this.scopes = scopes;
 	}
+
+	public Resource getLocation() {
+		return this.location;
+	}
+
+	public void setLocation(Resource location) {
+		this.location = location;
+	}
+
 }
--- a/spring-cloud-gcp-starters/spring-cloud-gcp-starter-config/src/main/java/org/springframework/cloud/gcp/config/GcpConfigProperties.java
+++ b/spring-cloud-gcp-starters/spring-cloud-gcp-starter-config/src/main/java/org/springframework/cloud/gcp/config/GcpConfigProperties.java
@@ -16,10 +16,9 @@
 
 package org.springframework.cloud.gcp.config;
 
-import java.util.Collections;
-
 import org.springframework.boot.context.properties.ConfigurationProperties;
-import org.springframework.cloud.gcp.core.AbstractCredentialsProperty;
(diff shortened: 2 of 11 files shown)


spring-cloud-gcp-starters/spring-cloud-gcp-starter-config/src/main/java/org/springframework/cloud/gcp/config/GcpConfigProperties.java
```diff
@@ -41,7 +40,8 @@
 
 	private String projectId;
 
-	private Credentials credentials;
+	@NestedConfigurationProperty
```
meltsufin: @joaoandremartins I thought this didn't work when you tried it before??

spring-cloud-gcp-core/src/main/java/org/springframework/cloud/gcp/core/Credentials.java
```diff
@@ -16,27 +16,26 @@
...
 
-	public Resource getLocation() {
-		return this.location;
-	}
+	private Resource location;
 
-	public void setLocation(Resource location) {
-		this.location = location;
+	public Credentials(String... defaultScopes) {
+		this.scopes.addAll(Arrays.asList(defaultScopes));
```
artembilan:
```
ArrayList(E[] array) {
 a = Objects.requireNonNull(array);
}
```
So, we need assertions do not allow end-users to specify `null` for the `defaultScopes`.

Or is it OK in the current Spring Boot to rely just on the `NPE`, @snicoll ?

Thanks

spring-cloud-gcp-starters/spring-cloud-gcp-starter-config/src/main/java/org/springframework/cloud/gcp/config/GcpConfigProperties.java
artembilan: That's cool! That's exactly what we have missed in the past when we tried to expose `Credentials` as top-level class.
Thank you @snicoll !

spring-cloud-gcp-core/src/main/java/org/springframework/cloud/gcp/core/Credentials.java
snicoll: We can add an assertion yes. It's not really a public API so I didn't bother

spring-cloud-gcp-starters/spring-cloud-gcp-starter-config/src/main/java/org/springframework/cloud/gcp/config/GoogleConfigPropertySourceLocator.java
```diff
@@ -76,7 +76,7 @@ public GoogleConfigPropertySourceLocator(GcpProjectIdProvider projectIdProvider,
 		if (gcpConfigProperties.isEnabled()) {
 			Assert.notNull(credentialsProvider, "Credentials provider cannot be null");
 			Assert.notNull(projectIdProvider, "Project ID provider cannot be null");
-			this.credentials = gcpConfigProperties.getCredentials() != null
+			this.credentials = gcpConfigProperties.getCredentials().getLocation() != null
```
joaoandremartins: Sounds like we should also add an assertion to check for a null `gcpConfigProperties.getCredentials()`?

spring-cloud-gcp-starters/spring-cloud-gcp-starter-config/src/main/java/org/springframework/cloud/gcp/config/GcpConfigProperties.java
joaoandremartins: Yes, I thought I had looked into `@NestedConfigurationProperty` before and it didn't work.
But after thinking about this for a while, I can't remember what the problem was.
It might have been because the name of the object (`Credentials`) is mapped to the name of the property.
I didn't want to call it that because end users might try to use it, but it's basically useless outside of the context of config properties. But if you're happy with this, I guess I won't oppose.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-config/src/main/java/org/springframework/cloud/gcp/config/GoogleConfigPropertySourceLocator.java
joaoandremartins: I guess if credentials are `final` and we're setting them right away, they can't be null.

artembilan:
All concerns are addressed.
Merging...

PR #268, opened 2017-12-15
Polish

snicoll:
This commit changes the config example to use @ConfigurationProperties
rather than @Value. Spring Boot users are more familiar with the former
and it is the recommended way to bind properties from the Environment.

### DIFF
--- a/spring-cloud-gcp-examples/spring-cloud-gcp-config-example/src/main/java/com/example/MyAppProperties.java
+++ b/spring-cloud-gcp-examples/spring-cloud-gcp-config-example/src/main/java/com/example/MyAppProperties.java
@@ -16,22 +16,21 @@
 
 package com.example;
 
-import org.springframework.beans.factory.annotation.Value;
+import org.springframework.boot.context.properties.ConfigurationProperties;
 import org.springframework.cloud.context.config.annotation.RefreshScope;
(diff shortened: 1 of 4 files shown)


spring-cloud-gcp-examples/spring-cloud-gcp-config-example/src/main/java/com/example/MyAppProperties.java
```diff
@@ -16,22 +16,21 @@
...
 /**
  * Sample Configuration class with property values loaded using the Runtime Configurator API.
+ *
  * @author Jisha Abubaker
+ * @author Stephane Nicoll
  */
 @RefreshScope
-@Configuration
-public class SampleConfig {
+@ConfigurationProperties("myapp")
```
joaoandremartins: Just wondering if we need the prefix?

The configuration name already contains `myapp`, feels a bit extraneous here.

snicoll: You have to have a prefix with `@ConfigurationProperties`. We (the Spring Boot team) don't really want to promote the use of  `@Value` so I'd appreciate if the example here would show what we expect users to do in practice. 

If you set a value in the environment, you'll very much likely to group them under a given prefix.

PR #273, opened 2017-12-15
Moves internal dependency management to parent pom, instead of bom

### LINKED ISSUE #265: Invalid dependency management in spring-cloud-gcp-dependencies
The  `spring-cloud-gcp-dependencies` is a BOM and should not have dependency management for the modules that are not strictly managed by GCP. 

This includes:

* Any other Spring Cloud project (they have a BOM themselves so this potentially overrides a choice a user would make by importing a certain release train)
* 3rd party provided by Spring Boot (mysql & postgres driver, `commons-logging`)

joaoandremartins:
Fixes #265


joaoandremartins:
@snicoll hopefully I'm on the right track here.

I noticed that `spring-cloud-netflix` specifies `commons-logging` in `spring-cloud-netflix-dependencies`. But the way I read it, we don't want to mandate its version to our users, so I left it out.

pom.xml
```diff
@@ -49,6 +53,103 @@
 				<type>pom</type>
 				<scope>import</scope>
 			</dependency>
+			<dependency>
+				<groupId>com.google.cloud.trace.v1</groupId>
+				<artifactId>sink</artifactId>
+				<version>${cloud-trace-java.version}</version>
```
snicoll: The google ones are fine IMO. Nobody else but this project is suited to have an opinion on those and it is actually better that you have an opinion so that things are consistent.

The exclusion is key as well in case a user adds the dependency manually in a project generated with the bom.

pom.xml
```diff
@@ -49,6 +53,103 @@
...
+			<dependency>
+				<groupId>org.springframework.cloud</groupId>
+				<artifactId>spring-cloud-starter-sleuth</artifactId>
+				<version>${spring-cloud-sleuth.version}</version>
+			</dependency>
+
+			<!-- spring-cloud-gcp-starter-core -->
+			<dependency>
+				<groupId>com.google.auth</groupId>
+				<artifactId>google-auth-library-oauth2-http</artifactId>
```
snicoll: This one can go back too IMO

pom.xml
```diff
@@ -49,6 +53,103 @@
...
+						<groupId>com.google.guava</groupId>
+						<artifactId>guava</artifactId>
+					</exclusion>
+				</exclusions>
+			</dependency>
+
+			<!-- spring-cloud-gcp-starter-sql -->
+			<dependency>
+				<groupId>com.google.cloud.sql</groupId>
+				<artifactId>mysql-socket-factory</artifactId>
```
snicoll: This one can go back too IMO

pom.xml
```diff
@@ -49,6 +53,103 @@
...
+			<dependency>
+				<groupId>org.springframework.cloud</groupId>
+				<artifactId>spring-cloud-context</artifactId>
+				<version>${spring-cloud-context.version}</version>
+			</dependency>
+
+			<!--google-cloud-java BOM-->
+			<dependency>
+				<groupId>com.google.cloud</groupId>
+				<artifactId>google-cloud-bom</artifactId>
```
snicoll: This one can go back too IMO.

artembilan:
I don't have a good Spring Boot-based projects dependency management ground, so I will deffer a review to @snicoll and will study how to be in the future 😄 .

Thanks

pom.xml
```diff
@@ -49,6 +51,43 @@
...
+			<dependency>
+				<groupId>org.springframework.cloud</groupId>
+				<artifactId>spring-cloud-starter-sleuth</artifactId>
+				<version>${spring-cloud-sleuth.version}</version>
+			</dependency>
+
+			<dependency>
+				<groupId>mysql</groupId>
+				<artifactId>mysql-connector-java</artifactId>
+				<version>5.1.42</version>
```
snicoll: Spring  Boot already has [dependency management on 5.1.44](https://github.com/spring-projects/spring-boot/blob/98f577e62c7bd9d95663b0c9c69cc0a4c2d8e788/spring-boot-dependencies/pom.xml#L137) for this one. I strongly recommend you not to deviate from the version that Spring Boot provides. If you have a good reason to do that, I am happy to hear what it is

pom.xml
```diff
@@ -49,6 +51,43 @@
...
+			</dependency>
+
+			<dependency>
+				<groupId>org.postgresql</groupId>
+				<artifactId>postgresql</artifactId>
+				<version>42.1.1</version>
```
snicoll: In 1.5.x, [we couldn't upgrade](https://github.com/spring-projects/spring-boot/blob/98f577e62c7bd9d95663b0c9c69cc0a4c2d8e788/spring-boot-dependencies/pom.xml#L141) to the next major version of the driver. Is there a reason why you need to stick with one particular version of the driver?

joaoandremartins: No special reason. It was the latest version at the time, but if you're already tracking it, no need for us to.

Issue #262, opened 2017-12-15
Consider creating a spring-cloud-gcp-autoconfigure module


snicoll:
I'd like you to consider moving the auto-configuration code in a central place (`spring-cloud-gcp-autoconfigure` comes to mind with a root `org.springframework.cloud.gcp.autoconfigure` package). There are 28 classes currently so it really isn't a lot of code.

There are several advantages to this:

* Getting the code in a central place allows you to easily provide support for GCP as a whole (basically, users can get a single dependency and trigger auto-config by adding depedencies to the stuff they want to use). This can lead to a generic "GCP Support" on start.spring.io
* Separating the code from the starter allows users to deviate from that opinion. If they don't like the dependencies the starter brings, they can craft their own or add them manually
* It is easier to reflect on conditions if the code and the dependencies are separated (`@ConditionalOnClass` is super cheap and a great way to backoff early)
* You don't need to add a starter to get all the available metadata integration in your IDE
* It is more consistent with what Spring Boot does (not really an argument, except it's been working great for us so far)

This would allow you to have a `spring-cloud-gcp-starter` that  brings this module (replacing what `spring-cloud-gcp-starter-core` does with basically all features ready to be enabled by adding a dependency).

One major drawback though is that you can't use the assumption an auto-configuration or an `EnvironmentPostProcessor` has to kick in based on the fact that it's available. But I think relying on that is a bit broken as well.

IMO, the documentation should also be updated to make less reference to "starter". A starter is a group of dependencies that provides a feature for something. It is not the feature itself.

I am happy to provide guidance  or review if you want to pursue this.

meltsufin:
Hi Stéphane, and thank you for taking such a thoughtful look at the project!

I believe we've considered a separate auto-config module before, but decided that keeping auto-config in the corresponding starters was simpler. One downside I see is that all these autoconfig classes will be on the classpath even if someone is only using a single module from the project. However, I understand your arguments for aligning with how Spring Boot organizes auto-config and I think we should do this.

joaoandremartins:
This was already in our roadmap, #155.

We can prioritise this if getting to Initializr depends on it.

artembilan:
Good point, @joaoandremartins !

I think that one should be closed in favor of this, since we have Stephane here already with his all Spring Boot wisdom 😄

snicoll:
> One downside I see is that all these autoconfig classes will be on the classpath even if someone is only using a single module from the project. 

[...]

PR #276, opened 2017-12-18
Creates spring-cloud-gcp-autoconfigure module

joaoandremartins:
Only contains the config module for now, remaining ones to come
separately so that the PR doesn't grow too big.
Also creates a spring-cloud-gcp-config module.

Helps fixing #262


joaoandremartins:
Will need to repeat for all the other starters.
One thing that's happening is that, by importing only `spring-cloud-gcp-autoconfigure` (and not `spring-cloud-gcp-starter-config`), `GcpConfigProperties` is somehow being populated, with only `GcpConfigAutoConfiguration` in the classpath (i.e., no `GoogleConfigPropertySourceLocator`).

@snicoll FYI to make sure I'm on the right track.

artembilan:
Travis doesn't like some your changes:
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.1:compile (default-compile) on project spring-cloud-gcp-config-example: Compilation failure: Compilation failure:

[ERROR] /home/travis/build/spring-cloud/spring-cloud-gcp/spring-cloud-gcp-examples/spring-cloud-gcp-config-example/src/main/java/com/example/MyAppProperties.java:[20,59] package org.springframework.cloud.context.config.annotation does not exist

[ERROR] /home/travis/build/spring-cloud/spring-cloud-gcp/spring-cloud-gcp-examples/spring-cloud-gcp-config-example/src/main/java/com/example/MyAppProperties.java:[28,2] cannot find symbol

[ERROR] symbol: class RefreshScope
```

joaoandremartins:
Nevermind the point about config always working. Turns out I didn't maven reinstall everything, working as intended now.

spring-cloud-gcp-autoconfigure/pom.xml
```diff
@@ -0,0 +1,35 @@
...
+        </dependency>
+        <!-- Configuration processor -->
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-configuration-processor</artifactId>
+            <optional>true</optional>
+        </dependency>
+        <dependency>
+            <groupId>org.springframework.cloud</groupId>
+            <artifactId>spring-cloud-gcp-starter-core</artifactId>
```
artembilan: I think this must go. The new ` spring-cloud-gcp-autoconfigure` should absorb its code fully.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/config/GcpConfigAutoConfiguration.java
```diff
@@ -14,22 +14,29 @@
...
  * Bootstrap auto configuration for Google Cloud Runtime Configurator Starter.
+ *
  * @author Jisha Abubaker
+ * @author João André Martins
  */
 @Configuration
 @AutoConfigureAfter(GcpContextAutoConfiguration.class)
 @EnableConfigurationProperties(GcpConfigProperties.class)
+@ConditionalOnClass(GoogleConfigPropertySourceLocator.class)
+@ConditionalOnMissingBean(GcpConfigProperties.class)
```
artembilan: I think we don't need this condition. The `@EnableConfigurationProperties` does the trick anyway.

spring-cloud-gcp-autoconfigure/src/main/resources/META-INF/spring.factories
```diff
@@ -1,4 +1,5 @@
+
 org.springframework.cloud.bootstrap.BootstrapConfiguration=\
-org.springframework.cloud.gcp.config.autoconfig.GcpConfigAutoConfiguration,\
+org.springframework.cloud.gcp.autoconfigure.config.GcpConfigAutoConfiguration,\
```
artembilan: That's not `BootstrapConfiguration` part. This must be for the `org.springframework.boot.autoconfigure.EnableAutoConfiguration` property

spring-cloud-gcp-config/pom.xml
```diff
@@ -0,0 +1,28 @@
...
+    <artifactId>spring-cloud-gcp-config</artifactId>
+
+    <dependencies>
+        <dependency>
+            <groupId>org.springframework.cloud</groupId>
+            <artifactId>spring-cloud-context</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.springframework.cloud</groupId>
+            <artifactId>spring-cloud-gcp-starter-core</artifactId>
```
artembilan: Non-Boot module can't depend on Boot. I see there `GcpContextAutoConfiguration` which should be as part of the `spring-boot-autoconfigure` as well and this one (`spring-cloud-gcp-config`) must be revisited respectivelly do not depend on Boot. 
If we need this separate module at all though...

spring-cloud-gcp-autoconfigure/src/main/resources/META-INF/spring.factories
joaoandremartins: I think this is correct, because it's for `BootstrapConfiguration`, as required by Spring Cloud Config.

joaoandremartins:
For some reason that I don't understand, the trace starter tests are stalling.

artembilan:
Any ideas regarding current Travis failure?

joaoandremartins:
It seems something is adding `BootstrapApplicationListener` to the test application's listeners.
That's causing the application to try to load `GoogleConfigPropertySourceLocator`, I suspect because it's in `spring-cloud-gcp-autoconfigure`'s `spring.factories`.

There's a class load error and the Spring app gets into an endless loop through `handleRunFailure()`.
Tomorrow, I'll need to understand why this test is trying to load config components, or why the similar SQL test doesn't try to load them at all, with a similar set of dependencies.

spring-cloud-gcp-autoconfigure/pom.xml
```diff
@@ -0,0 +1,36 @@
...
+        </dependency>
+        <!-- Configuration processor -->
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-configuration-processor</artifactId>
+            <optional>true</optional>
+        </dependency>
+        <dependency>
+            <groupId>org.springframework.cloud</groupId>
+            <artifactId>spring-cloud-gcp-core</artifactId>
```
meltsufin: Is this dependency needed currently?

meltsufin:
For me, a simple `mvn clean install` at the root fails.

```
testResolveScopesDefaultScopes(org.springframework.cloud.gcp.autoconfigure.core.GcpContextAutoConfigurationTests)  Time elapsed: 0.616 sec  <<< ERROR!
org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'gcpProjectIdProvider' defined in org.springframework.cloud.gcp.autoconfigure.core.GcpContextAutoConfiguration: Bean instantiation via factory method failed; nested exception is org.springframework.beans.BeanInstantiationException: Failed to instantiate [org.springframework.cloud.gcp.core.GcpProjectIdProvider]: Factory method 'gcpProjectIdProvider' threw exception; nested exception is java.lang.NoSuchMethodError: org.json.JSONTokener.<init>(Ljava/io/InputStream;)V
        at org.springframework.beans.factory.support.ConstructorResolver.instantiateUsingFactoryMethod(ConstructorResolver.java:599)
        at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.instantiateUsingFactoryMethod(AbstractAutowireCapableBeanFactory.java:1173)
        at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.createBeanInstance(AbstractAutowireCapableBeanFactory.java:1067)
        at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.doCreateBean(AbstractAutowireCapableBeanFactory.java:513)
        at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.createBean(AbstractAutowireCapableBeanFactory.java:483)
        at org.springframework.beans.factory.support.AbstractBeanFactory$1.getObject(AbstractBeanFactory.java:306)
        at org.springframework.beans.factory.support.DefaultSingletonBeanRegistry.getSingleton(DefaultSingletonBeanRegistry.java:230)
        at org.springframework.beans.factory.support.AbstractBeanFactory.doGetBean(AbstractBeanFactory.java:302)
        at org.springframework.beans.factory.support.AbstractBeanFactory.getBean(AbstractBeanFactory.java:197)
        at org.springframework.beans.factory.support.DefaultListableBeanFactory.preInstantiateSingletons(DefaultListableBeanFactory.java:761)
        at org.springframework.context.support.AbstractApplicationContext.finishBeanFactoryInitialization(AbstractApplicationContext.java:867)
        at org.springframework.context.support.AbstractApplicationContext.refresh(AbstractApplicationContext.java:543)
        at org.springframework.cloud.gcp.autoconfigure.core.GcpContextAutoConfigurationTests.loadEnvironment(GcpContextAutoConfigurationTests.java:103)
        at org.springframework.cloud.gcp.autoconfigure.core.GcpContextAutoConfigurationTests.testResolveScopesDefaultScopes(GcpContextAutoConfigurationTests.java:71)
        at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
        at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
        at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
        at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
        at org.junit.internal.runners.statements.RunAfters.evaluate(RunAfters.java:27)
        at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
        at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
        at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:57)
        at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
        at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
        at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
        at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
        at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)
        at org.junit.runners.ParentRunner.run(ParentRunner.java:363)
        at org.apache.maven.surefire.junit4.JUnit4Provider.execute(JUnit4Provider.java:252)
        at org.apache.maven.surefire.junit4.JUnit4Provider.executeTestSet(JUnit4Provider.java:141)
        at org.apache.maven.surefire.junit4.JUnit4Provider.invoke(JUnit4Provider.java:112)
        at org.apache.maven.surefire.util.ReflectionUtils.invokeMethodWithArray(ReflectionUtils.java:189)
        at org.apache.maven.surefire.booter.ProviderFactory$ProviderProxy.invoke(ProviderFactory.java:165)
        at org.apache.maven.surefire.booter.ProviderFactory.invokeProvider(ProviderFactory.java:85)
        at org.apache.maven.surefire.booter.ForkedBooter.runSuitesInProcess(ForkedBooter.java:115)
        at org.apache.maven.surefire.booter.ForkedBooter.main(ForkedBooter.java:75)
Caused by: org.springframework.beans.BeanInstantiationException: Failed to instantiate [org.springframework.cloud.gcp.core.GcpProjectIdProvider]: Factory method 'gcpProjectIdProvider' threw exception; nested exception is java.lang.NoSuchMethodError: org.json.JSONTokener.<init>(Ljava/io/InputStream;)V
        at org.springframework.beans.factory.support.SimpleInstantiationStrategy.instantiate(SimpleInstantiationStrategy.java:189)
        at org.springframework.beans.factory.support.ConstructorResolver.instantiateUsingFactoryMethod(ConstructorResolver.java:588)
        ... 43 more
Caused by: java.lang.NoSuchMethodError: org.json.JSONTokener.<init>(Ljava/io/InputStream;)V
        at com.google.cloud.ServiceOptions.getServiceAccountProjectId(ServiceOptions.java:442)
        at com.google.cloud.ServiceOptions.getDefaultProjectId(ServiceOptions.java:315)
        at org.springframework.cloud.gcp.core.DefaultGcpProjectIdProvider.getProjectId(DefaultGcpProjectIdProvider.java:36)
        at org.springframework.cloud.gcp.autoconfigure.core.GcpContextAutoConfiguration.gcpProjectIdProvider(GcpContextAutoConfiguration.java:157)
        at org.springframework.cloud.gcp.autoconfigure.core.GcpContextAutoConfiguration$$EnhancerBySpringCGLIB$$9e94071a.CGLIB$gcpProjectIdProvider$2(<generated>)
        at org.springframework.cloud.gcp.autoconfigure.core.GcpContextAutoConfiguration$$EnhancerBySpringCGLIB$$9e94071a$$FastClassBySpringCGLIB$$76e931a9.invoke(<generated>)
        at org.springframework.cglib.proxy.MethodProxy.invokeSuper(MethodProxy.java:228)
        at org.springframework.context.annotation.ConfigurationClassEnhancer$BeanMethodInterceptor.intercept(ConfigurationClassEnhancer.java:358)
        at org.springframework.cloud.gcp.autoconfigure.core.GcpContextAutoConfiguration$$EnhancerBySpringCGLIB$$9e94071a.gcpProjectIdProvider(<generated>)
        at org.springframework.beans.factory.support.SimpleInstantiationStrategy.instantiate(SimpleInstantiationStrategy.java:162)
        ... 44 more

```

spring-cloud-gcp-autoconfigure/pom.xml
joaoandremartins: Yeah, we need to import `GcpProperties`, `GcpProjectIdProvider`, etc.

joaoandremartins:
@meltsufin not sure why that's happening... Sounds like you have the wrong version of `org.json`.

meltsufin:
I found the conflict. `spring-boot-configuration-processor` brings in `android-json` that conflicts with the regular `json` jar. Adding the exclusion fixes the problem for me.

```xml
        <dependency>
            <optional>true</optional>
            <exclusions>
                <exclusion>
                    <groupId>com.vaadin.external.google</groupId>
                    <artifactId>android-json</artifactId>
                </exclusion>
            </exclusions>
        </dependency>
```

However, now the trace tests are stalling on:
```
10:30:39.815 [main] DEBUG org.springframework.core.env.StandardEnvironment - Adding PropertySource 'Inlined Test Properties' with highest search precedence
```

joaoandremartins:
Sorry, I'll merge the fix shortly.

joaoandremartins:
@meltsufin it should be working now

joaoandremartins:
This change disables auto-completion for both `core` and `config` properties.
I haven't been able to understand why, yet.

meltsufin:
The build still fails for me unless I exclude `android-json`, but then it succeeds.
I think we need to resolve the `json` dependency conflict.
`google-cloud-core` brings in ` org.json:json` and `spring-boot-configuration-processor` brings in `com.vaadin.external.google:android-json`. We need to exclude one or the other.

joaoandremartins:
I wonder if this something with your local environment, since neither I nor Travis are seeing the same thing.
In any case, it would be very helpful not to block the review on this.

spring-cloud-gcp-core/pom.xml
```diff
@@ -22,5 +22,9 @@
 			<groupId>org.springframework</groupId>
 			<artifactId>spring-core</artifactId>
 		</dependency>
+		<dependency>
+			<groupId>org.springframework.boot</groupId>
+			<artifactId>spring-boot-autoconfigure</artifactId>
```
artembilan: Cyclic dependency?
Why should we worry about autoconfig here if it is just `core`?
Not sure what I am missing, but would like to take your attention on this.

Thanks

spring-cloud-gcp-starters/spring-cloud-gcp-starter-config/src/main/resources/META-INF/spring.factories
```diff
@@ -1,4 +1,2 @@
 org.springframework.cloud.bootstrap.BootstrapConfiguration=\
-org.springframework.cloud.gcp.config.autoconfig.GcpConfigAutoConfiguration,\
-org.springframework.cloud.gcp.core.autoconfig.GcpContextAutoConfiguration,\
-org.springframework.cloud.gcp.config.GoogleConfigPropertySourceLocator
+org.springframework.cloud.gcp.config.GoogleConfigPropertySourceLocator
```
artembilan: New line in the end of file.
The IDEA can be configured to do that automatically.

spring-cloud-gcp-core/pom.xml
joaoandremartins: `spring-boot-autoconfigure`, not `spring-cloud-gcp-autoconfigure`.

We need `spring-boot-autoconfigure` to get `GcpProperties`'s `@ConfigurationProperties` annotation. LMK if I should be using another dependency.

artembilan: Oh! My bad. I see what is that.
Well, I think the problem that all the `@ConfigurationProperties` must be as a part of our `autoconfigure` not more.

Not sure how have you arranged it with Stephane, but I feel like something in architecture is still missing. Maybe there is still something in the target modules what is Boot-based? So, that should be moved to the `autoconfigure` module as well...

joaoandremartins: [Other spring-boot projects](https://github.com/spring-projects/spring-boot/tree/master/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/health) seem to be keeping their `@ConfigurationProperties` in the autoconfigure module as well.

I reviewed our project and all `@ConfigurationProperties` are used only within autoconfigure, except for the config and SQL ones.

For the config one, we might be able to make `GoogleConfigPropertySourceLocator` take in individual arguments instead of `GcpConfigProperties` (e.g., `config name`, `enabled`, etc.).
The SQL properties are used by the `JdbcUrlProviders`, but I think we can add those to the autoconfigure module as well..

artembilan: > For the config one, we might be able to make 

Are you going to do those changes to avoid this `spring-boot-autoconfigure` dependency in the `core`?
Or do you mean to do that in the separate issue?

Thanks

spring-cloud-gcp-autoconfigure/pom.xml
snicoll: Consider the case someone decides to add the auto-configuration of GCP to their stack but there are scenarios where they might not use it yet. 

IMO, this should be `<optional>true</optional>` and you should have an extra condition to guard the fact this dependency is not available.

spring-cloud-gcp-autoconfigure/src/main/resources/META-INF/spring.factories
```diff
@@ -0,0 +1,7 @@
+org.springframework.boot.autoconfigure.EnableAutoConfiguration=\
+org.springframework.cloud.gcp.autoconfigure.core.GcpContextAutoConfiguration,\
+org.springframework.cloud.gcp.autoconfigure.config.GcpConfigAutoConfiguration
+
+org.springframework.cloud.bootstrap.BootstrapConfiguration=\
+org.springframework.cloud.gcp.autoconfigure.config.GcpConfigAutoConfiguration,\
+org.springframework.cloud.gcp.autoconfigure.core.GcpContextAutoConfiguration
```
snicoll: I  don't know enough about that stuff to be positive but the fact those classes are both auto-config and bootstrap config is a smell IMO.

(nothing to do with this issue per se but I thought I'd mention it).

spring-cloud-gcp-samples/spring-cloud-gcp-config-sample/pom.xml
```diff
@@ -24,6 +24,10 @@
             <groupId>org.springframework.cloud</groupId>
             <artifactId>spring-cloud-gcp-starter-config</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.springframework.cloud</groupId>
+            <artifactId>spring-cloud-context</artifactId>
+        </dependency>
```
snicoll: If you need to add that manually in the sample, something is broken in a starter IMO

snicoll:
I gave some thoughts after [my previous review](https://github.com/spring-cloud/spring-cloud-gcp/pull/276#pullrequestreview-85479115) and I'd like to discuss this PR from another angle

@joaoandremartins what is the rationale of `spring-cloud-gcp-config`? Right now it has three classes and its code is highly tied to auto-configuration (configuration options). Why is it a first  class concept?

I am guessing the reason why a module was created is to avoid putting code in `spring-cloud-gcp-autoconfigure`. If the code was moved because of its dependencies, it really isn't fixing anything IMO (compared with having that dependencing in the autoconfigure module with the `optional` flag.

If I am right, can you expand on that? IMO, the current situation is not desirable: having a first-class concept that is not going to  be used in its own right does not seem legit to me. 

My vote really is to move that code to the `autoconfigure` module as described initially. I am happy to discuss the pros/cons of that approach again.

joaoandremartins:
@snicoll thanks a lot for your all your comments.

1) The high-level goal of the split was to follow the model of other SB starters. The starter itself doesn't contain any code and imports the core and autoconfiguration parts of the service.
2) `GoogleConfigPropertySourceLocator` can be used directly if the `autoconfigure` module isn't used. I think that warrants its own module (LMK if you think I'm wrong). I don't really expect `GoogleConfigEnvironment` to be used directly, but if not in this module, where would it live? You're recommending `autoconfigure`, but it doesn't do any auto configuration, would it be the best place for it to live in?
3) Enabled in the constructor: Unfortunately, Spring Cloud Config picks up any `PropertySourceLocator` from the classpath, therefore your assumption that `we only create it if we want it` isn't totally accurate.
Truth be told, `gcpConfigProperties.isEnabled()`'s usage is more important in the `locate()` method. It was added to the constructor so that the lack of name/profile doesn't generate warnings if config is disabled. (#245)

Overall, I'm for moving the `@ConfigurationProperties` classes to the `autoconfigure` module but leaving the others in `spring-cloud-gcp-config` in the same place.

snicoll:
> GoogleConfigPropertySourceLocator can be used directly if the autoconfigure module isn't used. I think that warrants its own module.

Not the way the API is designed IMO. Having to pass a `@ConfigurationProperties` object to its constructor and a check on `enabled` doesn't feel right to me. You have to do this for good reasons which links the code as currently designed to the auto-configuration.

Also, this looks very artificial to me. Spring Cloud requires Spring Boot  to run and while I totally agree that having a clean dependency tree is important, I don't think that the current modules are all warranted. If we have a `spring-cloud-gcp-config`, then it means it's a public feature, that a user may have a direct dependency to it to do something outside of the Spring Boot auto-configuration use case. What is it?

The reason why my vote goes to move things in one place is that you could create those first level concept later should you need to. Before, they were in the starter (totally dependant on Spring Boot) so I don't think this work should change anything to that. And if you don't expect the user to extend that stuff, it should probably be package private.

joaoandremartins:
I think the comments were all addressed.

Last thing is for me to understand why the GcpProperties injected in `GcpContextAutoConfiguration` doesn't have the settings specified in `application.properties`..

spring-cloud-gcp-samples/spring-cloud-gcp-config-sample/pom.xml
joaoandremartins: `@RefreshScope` requires this dependency. LMK if I can use a more minimal one.

joaoandremartins:
Ignored properties already appears to be the case in master, so I'll investigate in parallel to this issue.

spring-cloud-gcp-autoconfigure/pom.xml
```diff
@@ -0,0 +1,45 @@
...
+
+    <dependencies>
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-autoconfigure</artifactId>
+        </dependency>
+        <!-- Configuration processor -->
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-configuration-processor</artifactId>
```
meltsufin: This exclusion for `android-json` should be added because it indeed conflicts with `com.google.cloud.ServiceOptions.getServiceAccountProjectId(ServiceOptions.java:442)` use of the `org.json.JSONTokener.<init>(Ljava/io/InputStream;)V` constructor, which doesn't exist in the stripped down `android-json` library.

joaoandremartins: Then should we exclude `android-json` from the other sources and use the one from `spring-boot-configuration-processor`?
It's hard to fix this issue because I'm not seeing it, so I'm not sure to what extent this is happening because of your (or mine) local configuration.

In the meantime, my toy project is again picking up project ID and creds and I didn't do any change, not sure what happened but things work now...

meltsufin: Maven loads dependencies in the order that they are declared. Since `spring-boot-configuration-processor` comes before `spring-cloud-gcp-core` in this POM, `JSONTokenizer` from `android-json` should get loaded. I'm not sure why you're not seeing it.

For me, actually just switching the order of `spring-boot-configuration-processor` and `spring-cloud-gcp-core` fixes the build.

Can you try running `mvn dependency:tree` on this POM to see in what order dependencies are loaded.

snicoll: I think at this point we need to configure the annotation processor [in the compiler plugin](http://maven.apache.org/plugins/maven-compiler-plugin/compile-mojo.html#annotationProcessorPaths). You may lose AP support in the IDE but it's better than the current arrangement IMO.

snicoll:
Is the plan is to do this step by step? Are we focused on core and config here and another PR is going to migrate sql and storage?

I find an interesting pattern with the new arrangement already. The difference between `spring-cloud-gcp-starter-config` and `spring-cloud-gcp-starter-core` is that the latter brings `spring-cloud-gcp-core`. 

I find this odd. I would expect that something named "core" is always brought. 

We can discuss about naming things later but this new arrangement calls for a `spring-boot-gcp-starter` IMO (what the config one currently is?). And all starters can have a reference to this one (with the pattern basic starter + this specific feature). This is how we do things in Spring Boot.

Also, please make sure that the starters bring [`spring-boot-starter` somehow](https://docs.spring.io/spring-boot/docs/current-SNAPSHOT/reference/htmlsingle/#boot-features-custom-starter-module-starter).

joaoandremartins:
The starter you're talking about is currently `spring-cloud-gcp-starter-core`. `spring-cloud-gcp-starter-config` is only for Spring Cloud Config/Google Runtime Configuration API, at the same level as the Storage, SQL, etc. starters.

Indeed, I didn't want to end up with a huge PR, so we're doing this migration step by step, starting with core and config, and then doing the rest (I did the SQL one locally, just waiting to merge this to send another PR).

Regarding bringing core along with config, are you referring to the fact that `spring-cloud-gcp-starter-config`'s `spring.factories` doesn't specify `spring-cloud-gcp-core`?
I'm not too sure what that file is or how it works, so didn't add any transitive dependencies to it. (`spring-cloud-gcp-core` is a transitive dependency of `spring-cloud-gcp-starter-config` through `spring-cloud-gcp-autoconfigure`).

joaoandremartins:
Also, we should be able to specify the `spring-boot-starter` dependency right at the `spring-cloud-gcp-starters` level, so it is a dependency of every starter. LMK if there is any issue with that, I'll do it in a separate PR.

spring-cloud-gcp-autoconfigure/pom.xml
joaoandremartins: Only `spring-boot-configuration-processor` is importing `android-json` for me.

[INFO] --- maven-dependency-plugin:2.8:tree (default-cli) @ spring-cloud-gcp-autoconfigure ---
[INFO] org.springframework.cloud:spring-cloud-gcp-autoconfigure:jar:1.0.0.BUILD-SNAPSHOT
[INFO] +- org.springframework.boot:spring-boot-autoconfigure:jar:1.5.7.RELEASE:compile
[INFO] |  \- org.springframework.boot:spring-boot:jar:1.5.7.RELEASE:compile
[INFO] +- org.springframework.boot:spring-boot-configuration-processor:jar:1.5.7.RELEASE:compile
[INFO] |  \- com.vaadin.external.google:android-json:jar:0.0.20131108.vaadin1:compile
[INFO] +- org.springframework.cloud:spring-cloud-gcp-core:jar:1.0.0.BUILD-SNAPSHOT:compile
[INFO] |  +- com.google.cloud:google-cloud-core:jar:1.14.0:compile
[INFO] |  |  +- com.google.guava:guava:jar:20.0:compile
[INFO] |  |  +- joda-time:joda-time:jar:2.9.2:compile
[INFO] |  |  +- org.json:json:jar:20160810:compile
[INFO] |  |  +- com.google.http-client:google-http-client:jar:1.23.0:compile
[INFO] |  |  |  \- org.apache.httpcomponents:httpclient:jar:4.5.3:compile
[INFO] |  |  |     +- org.apache.httpcomponents:httpcore:jar:4.4.6:compile
[INFO] |  |  |     \- commons-codec:commons-codec:jar:1.10:compile
[INFO] |  |  +- com.google.code.findbugs:jsr305:jar:3.0.0:compile
[INFO] |  |  +- com.google.api:api-common:jar:1.2.0:compile
[INFO] |  |  +- com.google.api:gax:jar:1.15.0:compile
[INFO] |  |  |  +- com.google.auto.value:auto-value:jar:1.2:compile
[INFO] |  |  |  +- org.threeten:threetenbp:jar:1.3.3:compile
[INFO] |  |  |  \- com.google.auth:google-auth-library-oauth2-http:jar:0.7.1:compile
[INFO] |  |  |     +- com.google.auth:google-auth-library-credentials:jar:0.9.0:compile
[INFO] |  |  |     \- com.google.http-client:google-http-client-jackson2:jar:1.23.0:compile
[INFO] |  |  |        \- com.fasterxml.jackson.core:jackson-core:jar:2.8.10:compile
[INFO] |  |  +- com.google.protobuf:protobuf-java-util:jar:3.4.0:compile
[INFO] |  |  |  +- com.google.protobuf:protobuf-java:jar:3.4.0:compile
[INFO] |  |  |  \- com.google.code.gson:gson:jar:2.8.1:compile
[INFO] |  |  +- com.google.api.grpc:proto-google-common-protos:jar:1.0.2:compile
[INFO] |  |  \- com.google.api.grpc:proto-google-iam-v1:jar:0.1.26:compile
[INFO] |  \- org.springframework:spring-core:jar:4.3.11.RELEASE:compile
[INFO] +- org.springframework:spring-web:jar:4.3.11.RELEASE:compile
[INFO] |  +- org.springframework:spring-aop:jar:4.3.11.RELEASE:compile
[INFO] |  +- org.springframework:spring-beans:jar:4.3.11.RELEASE:compile
[INFO] |  \- org.springframework:spring-context:jar:4.3.11.RELEASE:compile
[INFO] |     \- org.springframework:spring-expression:jar:4.3.11.RELEASE:compile
[INFO] +- commons-logging:commons-logging:jar:1.2:compile
[INFO] +- org.springframework.cloud:spring-cloud-context:jar:1.2.3.RELEASE:compile
[INFO] |  \- org.springframework.security:spring-security-crypto:jar:4.2.3.RELEASE:compile
[INFO] +- junit:junit:jar:4.12:test
[INFO] |  \- org.hamcrest:hamcrest-core:jar:1.3:test
[INFO] +- org.slf4j:slf4j-api:jar:1.7.25:compile
[INFO] +- org.mockito:mockito-core:jar:1.10.19:test
[INFO] |  \- org.objenesis:objenesis:jar:2.1:test
[INFO] \- org.springframework.boot:spring-boot-starter-test:jar:1.5.7.RELEASE:test
[INFO]    +- org.springframework.boot:spring-boot-test:jar:1.5.7.RELEASE:test
[INFO]    +- org.springframework.boot:spring-boot-test-autoconfigure:jar:1.5.7.RELEASE:test
[INFO]    +- com.jayway.jsonpath:json-path:jar:2.2.0:test
[INFO]    |  \- net.minidev:json-smart:jar:2.2.1:test
[INFO]    |     \- net.minidev:accessors-smart:jar:1.1:test
[INFO]    |        \- org.ow2.asm:asm:jar:5.0.3:test
[INFO]    +- org.assertj:assertj-core:jar:2.6.0:test
[INFO]    +- org.hamcrest:hamcrest-library:jar:1.3:test
[INFO]    +- org.skyscreamer:jsonassert:jar:1.4.0:test
[INFO]    \- org.springframework:spring-test:jar:4.3.11.RELEASE:test
[INFO] ------------------------------------------------------------------------

snicoll:
> The starter you're talking about is currently spring-cloud-gcp-starter-core. spring-cloud-gcp-starter-config is only for Spring Cloud Config/Google Runtime Configuration API, at the same level as the Storage, SQL, etc. starters.

Yes, but now that we have improved the structure, that difference is going away. If you look at the current config starter, it only brings the `autoconfigure` module. If you provide a feature with only that, it qualifies as the "core" starter IMO. But, as I've indicated, I find it weird that the base starter doesn't bring a module that's called "core". Perhaps "core" should be renamed then.

Said differently, adding things in one place will let you provide several feature with a single starter potentially so reviewing the starter structure is a good thing (since they won't have code in them anymore). It is your call to decide if the base starter should bring your "core" module (and any dependency that it requires) but my opinion is that if it's optional it shouldn't be named core.

snicoll:
FYI, I intend to shade the JSON api in the annotation processor so that you don't have to play with excludes as of `1.5.10`.

joaoandremartins:
@snicoll are you suggesting we merge the `config` starter with the `core` one? Conceptually, it doesn't make sense to me -- the core module shouldn't bring anything Spring Cloud Config-like. Only importing `spring-cloud-gcp-autoconfigure` is also exclusive to the config starter, because we merged the `spring-cloud-gcp-config` module with `autoconfigure`, which I'm still not 100% sure of, but won't be replicated with the other modules since none aims at the use case of "configuring an app".

Also regarding the `android-json` issue, given the PR you linked, I'm not going to address the exclusion here. @meltsufin we can add that exclusion on a separate PR that can be rolled back when SB 1.5.10 is released.

snicoll:
I am not suggesting anything besides what I wrote in my last comment. Let's look at the state of your branch:

* `spring-cloud-gcp-starter-config` brings `spring-cloud-gcp-autoconfigure`
* `spring-cloud-gco-starter-core` brings `spring-cloud-gcp-autoconfigure` and `spring-cloud-gcp-core`

> the core module shouldn't bring anything Spring Cloud Config-like

We need to discuss based on something we both could resonate with then. The core starter right now is essentially the config starter + something more. So the config starter doesn't bring anything extra as far as I know.

That's why, based on that state, I am suggesting that those two _could be_ merged. Now if that's a mistake, then something else needs to be fixed and we can resume this discussion.

If that's not a mistake, I see zero value into having a _dedicated_ starter for something that does "less" than something you call "core".

spring-cloud-gcp-autoconfigure/pom.xml
meltsufin: @joaoandremartins Can you just switch the order of `spring-boot-configuration-processor` and `spring-cloud-gcp-core` to fix the build for me?

meltsufin:
@snicoll @joaoandremartins 

I think we should follow the patterns established in the spring-boot project itself as closely as possible. Perhaps, one source of the confusion in the PR is the choice of the first module to apply it to -- `config`. Storage might've been a better one to start with, but let's continue with config anyway.

Another source of confusion IMO is the decision earlier in the PR to move `GoogleConfigPropertySourceLocator` to the `gcp-autoconfigure` module. The reason it's a problem is because `GcpConfigAutoConfiguration` itself is conditional on the presence of the `GoogleConfigPropertySourceLocator` and of course it will always be present because it's in the same module now. Therefore, it would be simpler to just keep the `gcp-config` module and add it as an optional dependency to `gcp-autoconfigure` and required to `gcp-starter-config`.

Finally, I think there is confusion about `gcp-starter` and `gcp-starter-core`. I think what we're calling `gcp-starter-core` in this project is equivalent to `spring-boot-starter`. So, we might consider renaming our `gcp-starter-core` to just `gcp-starter`. Indeed, it should depend on `gcp-autoconfigure` and be a dependency of all the other starters, just like the pattern in the official Spring Boot starters.

Let me know if I'm missing something here. Here's a diagram of what I imagine the structure should be:
image

joaoandremartins:
@snicoll aaah, now I think I understand better what you meant with core bringing more than config.

The decision to have `spring-cloud-gcp-core` as a standalone module is because it makes it possible for people to use our libraries without Spring Boot.
The motivation was the same for having `spring-cloud-gcp-config`, but you raised the point that `config` does configure an app and, therefore, should be part of `autoconfigure`. However, it is possible to use Spring Cloud Config without Spring Boot, so on a second thought, it should really be a separate module. It would also be more consistent with your theory -- both core and config would bring two modules each.

There is the issue of `GcpConfigProperties` being an input of `GoogleConfigPropertySourceLocator`. We followed [the same pattern as Spring Cloud Config](https://github.com/spring-cloud/spring-cloud-config/blob/master/spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigServicePropertySourceLocator.java#L68) taking in a `@ConfigurationProperties` annotated [`ConfigClientProperties`](https://github.com/spring-cloud/spring-cloud-config/blob/master/spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigClientProperties.java). However, it seems there is a strict dependency on Spring Boot from [`spring-cloud-config-client`](https://github.com/spring-cloud/spring-cloud-config/blob/master/spring-cloud-config-client/pom.xml).

I'm pretty undecided as to what path to follow, but I wouldn't want to block progress (and momentum) on this detail.

joaoandremartins:
If it helps, we can have a quick 30m hangout to discuss and unblock this issue.

artembilan:
To be honest Spring Cloud Config is fully based on Spring Boot. There no separation and so on.
I even heard that everything what is "Spring Cloud" not necessarily should strictly be separated from Spring Boot. So, we really may consider some compromise here as well, at least for the `config` module.

snicoll:
> The decision to have spring-cloud-gcp-core as a standalone module is because it makes it possible for people to use our libraries without Spring Boot.

OK but there's not much there, is it? I am wondering if we couldn't move `AppEngineCondition` (see https://github.com/spring-projects/spring-boot/issues/11531) which leaves you with the `GcpProjectIdProvider` stuff. It is perfectly fine to have such a module and I am not complaining (except I'd call that `spring-cloud-gcp`).

>  it is possible to use Spring Cloud Config without Spring Boot

IMO, I haven't seen a single use case for that and, as I've indicated previously, I don't think it makes sense to have a first class citizen concept in your project (that module) for something that doesn't address a first class level use case. Having that code with the auto-configuration in package private level gives you the maximum flexibility to change the implementation whenever you feel like.

Having said that, both should work I suppose (but I am still very confused about the starters and what they do). A clean commit that summarizes the proposal + an hangout should help.

snicoll:
@joaoandremartins we are willing to move `AppEngineCondition` to Spring Boot 2. As the code is in here, would you be willing to submit a PR to Spring Boot please? 

The change is about [adding a value in `CloudPlatform`](https://github.com/spring-projects/spring-boot/blob/7dc8ada1353e2e5ebeb41ff55c47cf78dfafe184/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/cloud/CloudPlatform.java) and update the tests to validate the sceario. If you're ok me taking the code, I am happy to apply that myself directly, just let me know what you prefer.

joaoandremartins:
@snicoll PTAL

I've added back the `spring-cloud-gcp-config` module. It doesn't have a use besides activating the `@EnableConfigurationProperties` annotation in `GcpConfigAutoConfiguration`. This is so that Spring Cloud Config isn't always enabled for any project that uses core. It's the best compromise we found here, but let us know what you think.

I'd prefer to keep `spring-cloud-gcp-core`. `spring-cloud-gcp` is not descriptive enough, but I can change this if you feel very strongly about it.

dsyer:
I'm not sure I follow all of the discussion above (possibly refers to previous drafts), so at the risk of misunderstanding something, I have a couple of comments.

We should look at it from the point of view of the user. As a user I want to use GCP as a config properties source (whatever that means). How do I achieve that? I expect it is: I add something to the classpath (one thing, or two?), and configure an env var (or equivalent). I believe @snicoll is offering to add a feature to start.spring.io so that even "two things on the classpath" can be specified as "one checkbox". So that would be nice. IMO we should make that happen, and reduce the number of tiny modules here as much as possible.

> This is so that Spring Cloud Config isn't always enabled for any project that uses core

I don't understand that comment, and I don't understand why `spring-cloud-gcp-config` is needed as a separate module (it's basically trivial to merge with the autoconfig one and is only ever going to be used with that).

spring-cloud-gcp-autoconfigure/pom.xml
dsyer: I don't think you can exclude the android dependency and expect the APT to work. I believe it's no longer a dependency in Spring Boot 1.5.10 though, so upgrading to snapshots will fix it.

meltsufin:
@dsyer Thank you for looking at this PR and your comments.
In the interest of moving this PR forward, I would suggest limiting its scope just to the introduction of the `autoconfigure` module and making the starters have no code -- basically, just aligning the module structure to what is done in `spring-boot`.

We've gone back and forth on whether to have the `config` module or merge into the `autoconfigure` module. I don't think it makes much of a difference to the user because enabling it is the same -- by adding the `starter-config` module. In any case, let's continue this discussion in #299 so as not to block this PR. 

On the topic of merging starters or exposing combination of them on start.spring.io, let's have that discussion in #300.

spring-cloud-gcp-config/pom.xml
```diff
@@ -0,0 +1,33 @@
...
+    <name>Spring Cloud GCP Runtime Configuration Module</name>
+
+    <dependencies>
+        <dependency>
+            <groupId>org.springframework</groupId>
+            <artifactId>spring-web</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>commons-logging</groupId>
+            <artifactId>commons-logging</artifactId>
```
meltsufin: `<scope>test</scope>`

dsyer: +1. Can't have commons-logging as a transitive for user apps. OTOH, if we just merge this tiny module and make it depend on Spring Boot (like it feels it should), then it's probably moot.

joaoandremartins: Merger of the `config` module is being discussed in https://github.com/spring-cloud/spring-cloud-gcp/issues/299#issuecomment-356641220

As per my comment, I don't think it's doable, but your input is welcome!

meltsufin:
@artembilan Do you still have concerns on this PR? Can you do another quick pass please?

artembilan:
Sorry, I'm not so good in everything we have in this PR, but if @snicoll or @dsyer says it's OK, I'll approve.

Will take a look one more time otherwise somewhere later in the afternoon.

Investigating `NPE` in some other place - blocker though 😄

joaoandremartins:
@artembilan an NPE in our project? If so, I'd like to know about it. Maybe I can help.

Also, provided @snicoll is OK with this PR, hope you don't mind if I dismiss your review. Comments were are resolved, anyway.

artembilan:
No, that's not related to this project. I have many others in my plate.

OK. Feel free to dismiss and merge. I can review afterwards.

snicoll:
> Also, provided @snicoll is OK with this PR

I really would like to move forward with this and get a better idea on the bigger picture but unfortunately I can't. I feel we've been back and forth on this so I don't know what to do to unblock the situation.

Here are some feedback on the current state, I hope this will help

`spring-cloud-gcp-config` shouldn't exist IMO:
* The `GcpConfigPropertiesProvider` doesn't look like an API you'd have designed without the constraints that have been expressed here. There is only one implementation in the auto-configuration module and I doubt anyone would bother implementing this interface to use this feature without Spring Boot
* The only reason it exists as far as I understand is to make sure the feature only kicks in when it is added to the classpath. But it doesn't have to be that way. If we want to be good citizen, we shouldn't enable this feature when running the code in our IDE. Perhaps we need a key, or we should check we are running on GCP.
* This module that, supposingly, isn't related to Spring Boot/Spring Cloud as a `spring.factories` in it. That doesn't sound right.

`GcpConfigAutoConfiguration` is empty. I can't help but think that something is missing there. What does this auto-configuration do exactly besides processing `GcpConfigProperties`? Why is it an auto-configuration if the thing that it auto-configures should run in the bootstrap context? I don't know how `BootstrapConfiguration` is supposed to work but I assume these are components that run in the parent context that processes the `Environment`. If I am right, that auto-configuration seems useless.

Perhaps it makes sense to have a `spring-cloud-gcp-config` module but I fail to see why considering that this feature should only be enabled (?) when the app run on GCP.

Having said that, I don't think I am giving any argument that I'ven't given already so feel free to move forward if you believe those objections aren't warranted.

Thank you!

meltsufin:
@snicoll Since the last comment primarily relates to #299. I moved it there. If you don't mind, let's continue this discussion there, and merge this PR for the time being.

snicoll:
@meltsufin I don't understand why you would want to merge something that _adds_ something that is discussed for _removal_ in another issue (I find that super confusing). But if merging this PR helps moving things forward, sure thing.

PR #277, opened 2017-12-20
 Moves spring-integration-gcp to spring-cloud-gcp-integration

### LINKED ISSUE #269: Consider renaming spring-integration-gcp
`spring-integration-gcp` is a bit weird. I understand why it was named that way but all modules should ideally start with the same prefix. 

Regardless and unless this is meant to be contributed back to Spring Integration, the package should not start with `org.springframework.integration`.

joaoandremartins:
Fixes #269


artembilan:
Hm. Plus, there is already some conflict against `master` already.

PR #280, opened 2017-12-21
Ensures all publisher's and subscriber's settings are configurable

joaoandremartins:
Addresses #92's point:

> Revisit configuration, make sure all settings are configurable via a bean, e.g., RetrySetting


meltsufin:
Why not just make the `SubscriberBuilder` a constructor argument here?

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/DefaultPublisherFactory.java
```diff
@@ -72,17 +93,36 @@ public DefaultPublisherFactory(GcpProjectIdProvider projectIdProvider,
...
-						.setChannelProvider(this.channelProvider)
-						.setCredentialsProvider(this.credentialsProvider)
-						.build();
+				Publisher.Builder publisherBuilder =
+						Publisher.newBuilder(TopicName.of(this.projectId, key))
+								.setExecutorProvider(this.executorProvider)
+								.setChannelProvider(this.channelProvider)
+								.setCredentialsProvider(this.credentialsProvider);
+
+				if (this.retrySettings != null) {
```
artembilan: If they are not mandatory why they are not setters?

That is really a rule we follow in many Spring projects: if property is mandatory, it must be a part of ctor args. Otherwise setter.

I don't see any reason in a builder pattern implementation here. We need that one for cases when we may go different branches downstream chain according the current selection. More over the setters way is good for XML configuration.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/DefaultSubscriberFactory.java
```diff
@@ -42,10 +45,41 @@
...
 	public DefaultSubscriberFactory(GcpProjectIdProvider projectIdProvider,
 			ExecutorProvider executorProvider,
 			TransportChannelProvider channelProvider,
-			CredentialsProvider credentialsProvider) {
+			CredentialsProvider credentialsProvider,
+			HeaderProvider headerProvider,
+			ExecutorProvider systemExecutorProvider,
+			FlowControlSettings flowControlSettings,
+			Duration maxAckDurationPeriod,
+			Integer parallelPullCount) {
```
artembilan: No, that's not what I'd like to see for end-user API. That is really confusing. Would be better to move non-mandatory properties to setters. If you worry about autowiring, we have a Spring Boot auto-configuration on the matter.

More over no one is going to provide `Duration` and `Integer` beans.

Out of topic: `Duration` is a Java 8 class. So, end-users won't see benefits to stick with SF-4.3 when there is already SF-5.0 for better Java 8 support. I mean end-users might start complain why we require Java 8 if we don't support SF-5.0? Or: why we are not based on SF-5.0 if we require Java 8?

:shrugged:

joaoandremartins: We're using `org.threeten.bp.Duration`, a backport for Java 5, 6 and 7. Although I have no issues requiring Java 8.

On the topic of SI 5.0, the way we're currently controlling our SI version is by setting `spring-cloud-gcp` `pom.xml`'s parent to `spring-cloud-build`.
Changing its version to the latest `2.0.0.M6` would always change SI to 5 and SB to 2.

In other words, is there a way to change SI to 5 while keeping SB < 2? Is that even recommended?
We could always explicitly set an SI version in our bom or parent pom to 5...

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/DefaultPublisherFactory.java
joaoandremartins: I'll change this to setters because they play along well with XML.
@meltsufin wdyt?

joaoandremartins:
@meltsufin because the subscription builder takes a subscription name that we only pass in `getSubscriber()`.
Do you think there's value in taking in a `Subscriber.Builder` in the constructor and then just complement that builder with the subscription name at `getSubscriber()` time?

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/DefaultPublisherFactory.java
```diff
@@ -52,37 +50,92 @@
...
 		this.projectId = projectIdProvider.getProjectId();
 		Assert.hasText(this.projectId, "The project ID can't be null or empty.");
+	}
+
+	/**
+	 * @param executorProvider provides the executor to be used by the publisher, useful to specify
+	 *                         the number of threads to be used by each executor
+	 */
+	public void setExecutorProvider(ExecutorProvider executorProvider) {
 		this.executorProvider = executorProvider;
```
meltsufin: Does it make sense to just create the empty builder in the constructor, and set directly on the builder here?
```java
this.publisherBuilder.setExecutorProvider(executorProvider);
```
This way you can remove all of those instance variables and also simplify the `getPublisher` to a simple `this.publisherBuilder.build()`.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/DefaultSubscriberFactory.java
```diff
@@ -34,37 +37,130 @@
...
-				SubscriptionName.of(this.projectId, subscriptionName), receiver)
-				.setChannelProvider(this.channelProvider)
-				.setExecutorProvider(this.executorProvider)
-				.setCredentialsProvider(this.credentialsProvider)
-				.build();
+		Subscriber.Builder subscriberBuilder = Subscriber.newBuilder(
+				SubscriptionName.of(this.projectId, subscriptionName), receiver);
+
+		if (this.channelProvider != null) {
+			subscriberBuilder.setChannelProvider(this.channelProvider);
```
meltsufin: Again, all of this can be simplified if you do the builder configuration in the setters.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/DefaultPublisherFactory.java
joaoandremartins: I love the idea, but don't think it's feasible because `Subscriber.newBuilder()` takes a `SubscriptionName`, which we don't have at constructor time.

We can't really add it to the constructor because we'll go from 1 factory -> n subscriptions to 1 factory -> 1 subscription, at which point `SubscriberFactory` would just be a proxy for `Subscriber.Builder` (which is also not easily mockable).

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/DefaultSubscriberFactory.java
artembilan: The story is not about SI, but SF. Anyway I agree that it's not good to pull those version if other aren't bumped. I still think that we have to rely on the Spring Platform management dependencies as much as possible (Spring Cloud build should do that for us) and only use explicit versions for Spring Projects if they have some critical fix.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/DefaultPublisherFactory.java
```diff
@@ -52,37 +50,92 @@
...
 
-	private final CredentialsProvider credentialsProvider;
+	private BatchingSettings batchingSettings;
 
-	public DefaultPublisherFactory(GcpProjectIdProvider projectIdProvider,
-			ExecutorProvider executorProvider,
-			TransportChannelProvider channelProvider,
-			CredentialsProvider credentialsProvider) {
+	/**
+	 * @param projectIdProvider provides the GCP project ID
```
artembilan: FYI: just explain `@params` isn't not enough. There must a general description for methods.
When you list class method in JavaDocs page, that general description is present there in the table.
You see `@params` only when you jump into the method.

Anyway that is not Spring way to make end-user wrinkle from our not beautiful code 😄

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/DefaultPublisherFactory.java
meltsufin: Yeah, I see, it's a bummer. :(

joaoandremartins:
@artembilan tried to improve Javadoc. LMKWYT

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/DefaultPublisherFactory.java
```diff
@@ -71,38 +73,41 @@ public DefaultPublisherFactory(GcpProjectIdProvider projectIdProvider) {
 	}
 
 	/**
-	 * @param executorProvider provides the executor to be used by the publisher, useful to specify
-	 *                         the number of threads to be used by each executor
+	 * Sets the provider for the executor that will be used by the publisher.
+	 *
```
artembilan: This fails our Spring code style:

> Use an imperative style (i.e. Return and not Returns) for the first sentence.
No blank lines between the description and the parameter descriptions.

And no `<p>`, please, in method JavaDocs!

https://github.com/spring-projects/spring-framework/wiki/Code-Style#javadoc-formatting

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/DefaultPublisherFactory.java
```diff
@@ -61,6 +61,8 @@
 	private BatchingSettings batchingSettings;
 
 	/**
+	 * Default {@link DefaultPublisherFactory} constructor.
```
artembilan: This is somehow really non-informative JavaDoc.

We develop Open Source API and that would be great to respect end-users and let them to know what is this about. 
Just several words in the JavaDocs and a couple of your minutes and there is nothing to nit pick in the perfect Spring code!

I don't say that we should polish brilliantly our code right now, but that is exactly what we should work out in time.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/DefaultPublisherFactory.java
joaoandremartins: The link you pasted says `<p>` should be used for multiple paragraphs on methods.

```
If the description is defined with multiple paragraphs, start each of them with <p>.
```

Am I missing anything?

artembilan: Well, that "please" was really from me: I don't see reason to make separate paragraphs just from two sentences.

You are failing to follow other rules though 😄

joaoandremartins: You'd be surprised by the amount of different opinions on that "one-liner method description" rule. :P

In any case, I can change that if you want.

artembilan: Yes, please. I see reason in a new paragraph if we really start to talk about a new matter, like we do writing the book. But if the second sentence is  a continuation of the first one, that isn't read clearly.

Let's just follow what SF dictates and don't pay attention to other opinions which just would waste our time.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/DefaultPublisherFactory.java
joaoandremartins: Not sure what action you'd like to see here.

Should I just remove this Javadoc bit entirely?

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/DefaultPublisherFactory.java
```diff
@@ -52,37 +50,93 @@
...
 
-	private final CredentialsProvider credentialsProvider;
+	private BatchingSettings batchingSettings;
 
-	public DefaultPublisherFactory(GcpProjectIdProvider projectIdProvider,
-			ExecutorProvider executorProvider,
-			TransportChannelProvider channelProvider,
-			CredentialsProvider credentialsProvider) {
+	/**
+	 * Default {@link DefaultPublisherFactory} constructor.
```
artembilan: I would say like:
```
Create {@link DefaultPublisherFactory} instance based on the provided {@link GcpProjectIdProvider}.
The {@link GcpProjectIdProvider} must not be null, neither provide an empty {@code projectId}.
```

meltsufin:
@joaoandremartins I think this needs to be revisited

PR #281, opened 2017-12-21
Exposes underlying Pub/Sub API

joaoandremartins:
Addresses #92's point:

> Expose underlying API


joaoandremartins:
@artembilan I don't usually add Javadoc when the methods are self explanatory.
Do you think these methods need Javadoc?

artembilan:
Well, I would say that these methods are exactly  not self explanatory. 
We injected those factories to the `PubSubTemplate` externally and that isn't clear to me why I would use these new getters. Therefore some JavaDocs there would enlighten me.

More over that isn't good to say for Spring "I don't usually add Javadoc". Any pubic API must be with JavaDocs. That's one of the factors why people love Spring 😄

PR #282, opened 2017-12-21
Polish Docs

artembilan:


artembilan:
@meltsufin ,

Thank you for coming back to me 😄 
To be honest I haven't checked the doc locally.
Any advises how to build it and read from browser?
Thanks

meltsufin:
@artembilan I just use the "view" feature on Github, which is an approximation of just the snippet.
@joaoandremartins Might have a better way to preview. We should wait for his review anyway. He's in a different timezone right now though.

artembilan:
To build docs locally we need command:
```
./mvnw package -Pfull -DskipTests=true
```
(just to avoid tests 😄 )

With that I've figured out that we have some problem with the first page.
That still unclear why our HTML is not styled with CSS, but at least I see it properly on-line.
Will hope it will be fixed automatically when we merge this and CI build passes.

joaoandremartins:
Yeah, I use a command similar to that one - build with profile `full`.

I think you're not seeing the CSS because you might be opening the `target/docbook/html/index.html` or `htmlsingle/index.html`.

I usually open `target/generated-docs/index.html` or `target/contents/reference/htmlsingle/index.html` and CSS is correctly processed.

PR #283, opened 2017-12-22
Adds synchronous pull to PubSubTemplate

joaoandremartins:
Addresses #92's point:

> A way to pull a message

### DIFF
--- a/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubOperations.java
+++ b/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubOperations.java
@@ -1,5 +1,5 @@
 /*
- *  Copyright 2017 original author or authors.
+ *  Copyright 2017-2018 original author or authors.
  *
  *  Licensed under the Apache License, Version 2.0 (the "License");
  *  you may not use this file except in compliance with the License.
--- a/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubOperations.java
+++ b/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubOperations.java
@@ -20,6 +20,7 @@ import java.nio.charset.Charset;
 import java.util.List;
 import java.util.Map;
 
+import com.google.api.gax.retrying.RetrySettings;
 import com.google.cloud.pubsub.v1.MessageReceiver;
 import com.google.cloud.pubsub.v1.Subscriber;
 import com.google.protobuf.ByteString;
@@ -99,19 +100,20 @@ public interface PubSubOperations {
 	Subscriber subscribe(String subscription, MessageReceiver messageHandler);
 
 	/**
-	 * Pulls a number of messages from a Google Cloud Pub/Sub subscription.
+	 * Pulls and auto-acknowledges a number of messages from a Google Cloud Pub/Sub subscription.
 	 *
 	 * @param subscription the subscription name
 	 * @param maxMessages the maximum number of pulled messages
 	 * @param returnImmediately returns immediately even if subscription doesn't contain enough
 	 *                         messages to satisfy {@code maxMessages}
+	 * @param retrySettings the timeout and retry setting for the pull request
 	 * @return the list of received messages
 	 */
 	List<PubsubMessage> pull(String subscription, Integer maxMessages,
-			Boolean returnImmediately);
+			Boolean returnImmediately, RetrySettings retrySettings);
 
 	/**
-	 * Pulls a message from a Google Cloud Pub/Sub subscription.
+	 * Pulls and auto-acknowledges a message from a Google Cloud Pub/Sub subscription.
 	 *
 	 * @param subscription the subscription name
 	 * @return a received message, or {@code null} if none exists in the subscription
--- a/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubOperations.java
+++ b/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubOperations.java
@@ -17,12 +17,14 @@
 package org.springframework.cloud.gcp.pubsub.core;
 
 import java.nio.charset.Charset;
+import java.util.List;
 import java.util.Map;
 
 import com.google.cloud.pubsub.v1.MessageReceiver;
 import com.google.cloud.pubsub.v1.Subscriber;
 import com.google.protobuf.ByteString;
 import com.google.pubsub.v1.PubsubMessage;
+import com.google.pubsub.v1.PullRequest;
 
 import org.springframework.util.concurrent.ListenableFuture;
 
@@ -97,4 +99,34 @@ public interface PubSubOperations {
 	 */
 	Subscriber subscribe(String subscription, MessageReceiver messageHandler);
 
+	/**
+	 * Pulls a number of messages from a Google Cloud Pub/Sub subscription.
+	 *
+	 * <p>
+	 * Automatically acknowledges received messages.
+	 *
+	 * @param pullRequest request object containing request settings
+	 * @return the list of received messages
+	 */
+	List<PubsubMessage> pull(PullRequest pullRequest);
+
+	/**
+	 * Pulls a number of messages from a Google Cloud Pub/Sub subscription.
+	 *
+	 * @param subscription the subscription name
+	 * @param maxMessages the maximum number of pulled messages
+	 * @param returnImmediately returns immediately even if subscription doesn't contain enough
+	 *                         messages to satisfy {@code maxMessages}
+	 * @return the list of received messages
+	 */
+	List<PubsubMessage> pull(String subscription, Integer maxMessages,
+			Boolean returnImmediately);
+
+	/**
+	 * Pulls a message from a Google Cloud Pub/Sub subscription.
+	 *
+	 * @param subscription the subscription name
+	 * @return a received message, or {@code null} if none exists in the subscription
+	 */
+	PubsubMessage pullNext(String subscription);
 }
--- a/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubOperations.java
+++ b/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubOperations.java
@@ -33,12 +33,12 @@ import org.springframework.util.concurrent.ListenableFuture;
  *
  * @author Vinicius Carvalho
  * @author João André Martins
+ * @author Mike Eltsufin
  */
 public interface PubSubOperations {
 
 	/**
-	 * Sends a message to Pub/Sub.
-	 *
+	 * Send a message to Pub/Sub.
 	 * @param topic the name of an existing topic
 	 * @param payload the message String payload
 	 * @param headers map of String to String headers
@@ -47,8 +47,7 @@ public interface PubSubOperations {
 	ListenableFuture<String> publish(String topic, String payload, Map<String, String> headers);
 
 	/**
-	 * Sends a message to Pub/Sub.
-	 *
+	 * Send a message to Pub/Sub.
 	 * @param topic the name of an existing topic
 	 * @param payload the message String payload
 	 * @param headers map of String to String headers
@@ -59,8 +58,7 @@ public interface PubSubOperations {
 			Charset charset);
 
 	/**
-	 * Sends a message to Pub/Sub.
-	 *
+	 * Send a message to Pub/Sub.
 	 * @param topic the name of an existing topic
 	 * @param payload the message payload in bytes
 	 * @param headers map of String to String headers
@@ -69,8 +67,7 @@ public interface PubSubOperations {
 	ListenableFuture<String> publish(String topic, byte[] payload, Map<String, String> headers);
 
 	/**
-	 * Sends a message to Pub/Sub.
-	 *
+	 * Send a message to Pub/Sub.
 	 * @param topic the name of an existing topic
 	 * @param payload the message payload on the {@link PubsubMessage} payload format
 	 * @param headers map of String to String headers
@@ -79,8 +76,7 @@ public interface PubSubOperations {
 	ListenableFuture<String> publish(String topic, ByteString payload, Map<String, String> headers);
 
 	/**
-	 * Sends a message to Pub/Sub.
-	 *
+	 * Send a message to Pub/Sub.
 	 * @param topic the name of an existing topic
 	 * @param pubsubMessage a Google Cloud Pub/Sub API message
 	 * @return the listenable future of the call
@@ -88,11 +84,8 @@ public interface PubSubOperations {
 	ListenableFuture<String> publish(String topic, PubsubMessage pubsubMessage);
 
 	/**
-	 * Adds a callback method to an existing subscription.
-	 *
-	 * <p>
-	 * The created {@link Subscriber} is returned so it can be stopped.
-	 *
+	 * Add a callback method to an existing subscription.
+	 * <p>The created {@link Subscriber} is returned so it can be stopped.
 	 * @param subscription the name of an existing subscription
 	 * @param messageHandler the callback method triggered when new messages arrive
 	 * @return subscriber listening to new messages
@@ -100,12 +93,11 @@ public interface PubSubOperations {
 	Subscriber subscribe(String subscription, MessageReceiver messageHandler);
 
 	/**
-	 * Pulls and auto-acknowledges a number of messages from a Google Cloud Pub/Sub subscription.
-	 *
+	 * Pull and auto-acknowledge a number of messages from a Google Cloud Pub/Sub subscription.
 	 * @param subscription the subscription name
 	 * @param maxMessages the maximum number of pulled messages
 	 * @param returnImmediately returns immediately even if subscription doesn't contain enough
-	 *                         messages to satisfy {@code maxMessages}
+	 * messages to satisfy {@code maxMessages}
 	 * @param retrySettings the timeout and retry setting for the pull request
 	 * @return the list of received messages
 	 */
@@ -113,8 +105,7 @@ public interface PubSubOperations {
 			Boolean returnImmediately, RetrySettings retrySettings);
 
 	/**
-	 * Pulls and auto-acknowledges a message from a Google Cloud Pub/Sub subscription.
-	 *
+	 * Pull and auto-acknowledge a message from a Google Cloud Pub/Sub subscription.
 	 * @param subscription the subscription name
 	 * @return a received message, or {@code null} if none exists in the subscription
 	 */
--- a/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubOperations.java
+++ b/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubOperations.java
@@ -24,7 +24,6 @@ import com.google.cloud.pubsub.v1.MessageReceiver;
 import com.google.cloud.pubsub.v1.Subscriber;
 import com.google.protobuf.ByteString;
 import com.google.pubsub.v1.PubsubMessage;
-import com.google.pubsub.v1.PullRequest;
 
 import org.springframework.util.concurrent.ListenableFuture;
 
--- a/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubOperations.java
+++ b/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubOperations.java
@@ -102,17 +102,6 @@ public interface PubSubOperations {
 	/**
 	 * Pulls a number of messages from a Google Cloud Pub/Sub subscription.
 	 *
-	 * <p>
-	 * Automatically acknowledges received messages.
-	 *
-	 * @param pullRequest request object containing request settings
-	 * @return the list of received messages
-	 */
-	List<PubsubMessage> pull(PullRequest pullRequest);
-
-	/**
-	 * Pulls a number of messages from a Google Cloud Pub/Sub subscription.
-	 *
 	 * @param subscription the subscription name
 	 * @param maxMessages the maximum number of pulled messages
 	 * @param returnImmediately returns immediately even if subscription doesn't contain enough
--- a/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
+++ b/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
@@ -16,7 +16,6 @@
 
 package org.springframework.cloud.gcp.pubsub.core;
 
-import java.io.IOException;
 import java.nio.charset.Charset;
 import java.util.List;
 import java.util.Map;
@@ -36,10 +35,8 @@ import com.google.pubsub.v1.PubsubMessage;
 import com.google.pubsub.v1.PullRequest;
 import com.google.pubsub.v1.PullResponse;
 import com.google.pubsub.v1.ReceivedMessage;
-import com.google.pubsub.v1.SubscriptionName;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
-import org.threeten.bp.Duration;
 
 import org.springframework.beans.factory.InitializingBean;
 import org.springframework.cloud.gcp.core.GcpProjectIdProvider;
@@ -72,6 +69,7 @@ import org.springframework.util.concurrent.SettableListenableFuture;
  *
  * @author Vinicius Carvalho
  * @author João André Martins
+ * @author Mike Eltsufin
  */
 public class PubSubTemplate implements PubSubOperations, InitializingBean {
 
@@ -80,11 +78,6 @@ public class PubSubTemplate implements PubSubOperations, InitializingBean {
 	private final PublisherFactory publisherFactory;
 
 	private final SubscriberFactory subscriberFactory;
-
-	/**
-	 * Contains the retry settings of the synchronous pull client.
-	 */
-	private SubscriptionAdminSettings subscriptionAdminSettings;
 
 	/**
 	 * Default {@link PubSubTemplate} constructor.
@@ -174,9 +167,12 @@ public class PubSubTemplate implements PubSubOperations, InitializingBean {
 	 */
 	private List<PubsubMessage> pull(PullRequest pullRequest) {
 		Assert.notNull(pullRequest, "The pull request cannot be null.");
+		Assert.notNull(subscriberFactory.getSubscriptionAdminSettings(),
+				"The SubscriptionAdminSettings cannot be null.");
 
 		try {
-			try (SubscriberStub subscriber = GrpcSubscriberStub.create(this.subscriptionAdminSettings)) {
+			try (SubscriberStub subscriber = GrpcSubscriberStub.create(
+					subscriberFactory.getSubscriptionAdminSettings())) {
 				PullResponse pullResponse =	subscriber.pullCallable().call(pullRequest);
 
 				// Ack received messages.
@@ -219,13 +215,6 @@ public class PubSubTemplate implements PubSubOperations, InitializingBean {
 
 	@Override
 	public void afterPropertiesSet() throws Exception {
-	}
-
-	public void setPullTimeoutMillis(long pullTimeoutMillis) throws IOException {
-		SubscriptionAdminSettings.Builder subscriptionAdminSettingsBuilder = SubscriptionAdminSettings.newBuilder();
-		SubscriptionAdminSettings.newBuilder().pullSettings()
-				.setSimpleTimeoutNoRetries(Duration.ofSeconds(pullTimeoutMillis));
-		this.subscriptionAdminSettings = subscriptionAdminSettingsBuilder.build();
 	}
 
 	public PublisherFactory getPublisherFactory() {
--- a/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
+++ b/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
@@ -154,11 +154,11 @@ public class PubSubTemplate implements PubSubOperations, InitializingBean {
 	 */
 	private List<PubsubMessage> pull(PullRequest pullRequest) {
 		Assert.notNull(pullRequest, "The pull request cannot be null.");
-		Assert.notNull(this.subscriberFactory.getSubscriberStub(),
+		Assert.notNull(this.subscriberFactory.createSubscriberStub(),
 				"The SubscriberStub cannot be null.");
 
 		try {
-			SubscriberStub subscriber = this.subscriberFactory.getSubscriberStub();
+			SubscriberStub subscriber = this.subscriberFactory.createSubscriberStub();
 			PullResponse pullResponse =	subscriber.pullCallable().call(pullRequest);
 
 			// Ack received messages.
--- a/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
+++ b/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
@@ -26,8 +26,6 @@ import com.google.api.core.ApiFutureCallback;
 import com.google.api.core.ApiFutures;
 import com.google.cloud.pubsub.v1.MessageReceiver;
 import com.google.cloud.pubsub.v1.Subscriber;
-import com.google.cloud.pubsub.v1.SubscriptionAdminSettings;
-import com.google.cloud.pubsub.v1.stub.GrpcSubscriberStub;
 import com.google.cloud.pubsub.v1.stub.SubscriberStub;
 import com.google.protobuf.ByteString;
 import com.google.pubsub.v1.AcknowledgeRequest;
@@ -39,7 +37,6 @@ import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 
 import org.springframework.beans.factory.InitializingBean;
-import org.springframework.cloud.gcp.core.GcpProjectIdProvider;
 import org.springframework.cloud.gcp.pubsub.support.PublisherFactory;
 import org.springframework.cloud.gcp.pubsub.support.SubscriberFactory;
 import org.springframework.util.Assert;
@@ -55,17 +52,7 @@ import org.springframework.util.concurrent.SettableListenableFuture;
  *
  * <p>
  * This class also allows for synchronous message pulling from a subscription through the {@code pull()} and
- * {@code pullNext()} methods. In order for synchronous pulling to work, a {@link GcpProjectIdProvider} object must be
- * provided through the available setter method. For example:
- *
- * <pre>
- * <code>
- *     @Autowired PubSubTemplate pubSubTemplate;
- *
- *     pubSubTemplate.setProjectIdProvider(() -> "myGcpProjectId");
- *     pubSubTemplate.pull(PullRequest.newBuilder().setSubscription("my subscription").build());
- * </code>
- * </pre>
+ * {@code pullNext()} methods.
  *
  * @author Vinicius Carvalho
  * @author João André Martins
@@ -152,7 +139,7 @@ public class PubSubTemplate implements PubSubOperations, InitializingBean {
 
 	@Override
 	public Subscriber subscribe(String subscription, MessageReceiver messageHandler) {
-		Subscriber subscriber = this.subscriberFactory.getSubscriber(subscription, messageHandler);
+		Subscriber subscriber = this.subscriberFactory.createSubscriber(subscription, messageHandler);
 		subscriber.startAsync();
 		return subscriber;
 	}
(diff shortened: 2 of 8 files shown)


joaoandremartins:
Two things I'm not sure of in this PR:
1) Passing a `GcpProjectIdProvider` into `PubSubTemplate` just for sync pulling, besides the `SubscriberFactory` and `PublisherFactory`, who won't use that `GcpProjectIdProvider`.
We need the project ID to form the pull and acknowledgement requests, and it can't be found anywhere else in `PubSubTemplate`, so it doesn't look like we have much of a choice.
2) We do all the acking for users in sync pull. Allowing users to ack sync pull would be opening a can of worms. We would probably need to introduce as many methods as we have, for explicit acking, as well as a new concept, a `ReceivedMessage` which contains an ack ID and a `PubsubMessage`.

WDYT?

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
```diff
@@ -49,9 +60,29 @@
...
+	 * @param subscriberFactory the {@link com.google.cloud.pubsub.v1.Subscriber} factory to
+	 *                          subscribe to subscriptions
+	 * @param pullSettings a {@link com.google.cloud.pubsub.v1.SubscriptionAdminSettings} object to
+	 *                     create a {@link com.google.cloud.pubsub.v1.stub.SubscriberStub} to pull
+	 *                     messages on demand
+	 * @param projectIdProvider provides the project ID for the on demand pull requests. Only used
+	 *                          in the pull methods
+	 */
+	public PubSubTemplate(PublisherFactory publisherFactory, SubscriberFactory subscriberFactory,
+			SubscriptionAdminSettings pullSettings, GcpProjectIdProvider projectIdProvider) {
```
artembilan: I think these two new options are not mandatory. So, they should go to setters.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
```diff
@@ -119,6 +150,70 @@ public Subscriber subscribe(String subscription, MessageReceiver messageHandler)
...
+	 * Pulls messages on demand using the pull request in argument.
+	 *
+	 * <p>
+	 * After messages are received, this method sends acknowledgements for the received messages if
+	 * any was received.
+	 */
+	@Override
+	public List<PubsubMessage> pull(PullRequest pullRequest) {
+		try {
+			try (SubscriberStub subscriber = GrpcSubscriberStub.create(this.pullSettings)) {
```
artembilan: I wonder what will happen if `this.pullSettings == null`...

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
```diff
@@ -119,6 +150,70 @@ public Subscriber subscribe(String subscription, MessageReceiver messageHandler)
...
+					+ pullRequest.getSubscription() + ".", ioe);
+		}
+	}
+
+	@Override
+	public List<PubsubMessage> pull(String subscription, Integer maxMessages,
+			Boolean returnImmediately) {
+		PullRequest.Builder pullRequestBuilder =
+				PullRequest.newBuilder().setSubscriptionWithSubscriptionName(
+						SubscriptionName.of(this.projectIdProvider.getProjectId(), subscription));
```
artembilan: I think we have to check `projectIdProvider` for null before performing its methods.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/autoconfig/GcpPubSubAutoConfiguration.java
```diff
@@ -178,4 +181,16 @@ public SubscriptionAdminClient subscriptionAdminClient() {
 					ioe);
 		}
 	}
+
+	@Bean
+	@ConditionalOnMissingBean
+	public SubscriptionAdminSettings subscriptionAdminSettings() {
+		try {
+			return SubscriptionAdminSettings.newBuilder().build();
```
artembilan: Looks like this is useless bean, without any configuration properties based customization.

So, maybe we should consider to use this expression in the `PubsubTemplate` directly:
```
private SubscriptionAdminSettings pullSettings = SubscriptionAdminSettings.newBuilder().build();
```
as the default value?

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
```diff
@@ -49,9 +60,29 @@
 
 	private final SubscriberFactory subscriberFactory;
 
-	public PubSubTemplate(PublisherFactory publisherFactory, SubscriberFactory subscriberFactory) {
+	private final SubscriptionAdminSettings pullSettings;
```
meltsufin: The type doesn't seem right. Should it be `UnaryCallSettings<PullRequest, PullResponse>`? `SubscriptionAdminSettings` has a lot of other configuration that is not relevant to pull settings.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
```diff
@@ -119,6 +150,70 @@ public Subscriber subscribe(String subscription, MessageReceiver messageHandler)
...
 
+	/**
+	 * Pulls messages on demand using the pull request in argument.
+	 *
+	 * <p>
+	 * After messages are received, this method sends acknowledgements for the received messages if
+	 * any was received.
+	 */
+	@Override
+	public List<PubsubMessage> pull(PullRequest pullRequest) {
```
meltsufin: As I understand, there are two ways to do a pull: (1) with a simple timeout and (2) with complicated retry settings. I would argue that (1) is the most common/useful one, and we should make it easy for the developer to specify the timeout without going through the convoluted builders.

One possible way of doing it is to get rid of the whole pull settings option, and accept a timeout in pull methods directly: `pull(PullRequest pullRequest, long timeoutMs)`.

Another alternative is to introduce a configuration property like `spring.cloud.gcp.pubsub.pullTimeoutMs` and inject it into the `SubscriptionAdminSettings` in the auto-config.

WDYT?

artembilan: There indeed might be a default `pullTimeout` option for the method like `pull(PullRequest pullRequest)`, but I'd really consider auto-config on the matter in the separate issue/pull request.

Looks like we don't have consensus for the original API yet, so no reason to overcomplicate.

garyrussell:
Please note that over the holidays, I issued a [PR for a general abstraction](https://github.com/spring-projects/spring-integration/pull/2310) for a `DeferredAcknowlegmentMessageSource` - where a general ack/nack abstraction [(`AcknowledgmentCallback`)](https://github.com/spring-projects/spring-integration/pull/2310/files#diff-dfacd578246e3ea86fda898e052994d8) is added as a header and the framework will auto ack/nack when the poll completes, if the application code has not already done so.

I think we talked before about the need for this; comments on that PR are welcome.

I have implementations for RabbitMQ (in that PR) and [Kafka here](https://github.com/garyrussell/spring-integration-kafka/commit/5862f2406ee5c6e1740751a85a8dc2e48808e6a2).

artembilan:
@garyrussell ,

Looks like your comment isn't related to this `PubSubTemplate` PR. However your idea might be considered in the future, but already in the Spring Integration module.

Plus I still fight here to convince to move to Spring Cloud 2.0 base line. So, your PR unfortunately is far away from plans here... 

But anyway I'm switching to it for review since that is something what people request in SCSt.

Thanks

garyrussell:
I just saw the words "synchronous pull" which is what that PR does for rabbit and kafka.

Sorry for the noise.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubOperations.java
```diff
@@ -97,4 +99,34 @@
...
+	/**
+	 * Pulls a number of messages from a Google Cloud Pub/Sub subscription.
+	 *
+	 * <p>
+	 * Automatically acknowledges received messages.
+	 *
+	 * @param pullRequest request object containing request settings
+	 * @return the list of received messages
+	 */
+	List<PubsubMessage> pull(PullRequest pullRequest);
```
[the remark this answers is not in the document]
joaoandremartins: Hmmm why does the method signature have to symmetric?

I wanted to avoid exposing a new kind of response like `PullResponse`. Users only knew `PubsubMessage` before.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
joaoandremartins: +1 for the `spring.cloud.gcp.pubsub.pullTimeoutMillis` property.

@artembilan wdyt?

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
joaoandremartins: Why do you say so?

`GrpcSusbcriberStub.create()` takes in a `SubscriptionAdminSettings`.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/autoconfig/GcpPubSubAutoConfiguration.java
joaoandremartins: The idea is that the user can provide their own `SusbcriptionAdminSettings` bean with their desired configuration, overriding the default one we provide.

Thoughts?

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
artembilan: As I said: I don't want to hijack original API PR for auto-configuration story. 
We have to come back to it in the separate issue/PR.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
```diff
@@ -43,12 +43,31 @@
...
+ * {@link GcpProjectIdProvider} objects must be provided through the available setter methods. For example:
+ *
+ * <pre>
+ *     @Autowired PubSubTemplate pubSubTemplate;
+ *
+ *     pubSubTemplate.setPullSettings(SubscriptionAdminSettings.newBuilder().build());
+ *     pubSubTemplate.setProjectIdProvider(() -> "myGcpProjectId");
+ *     pubSubTemplate.pull(PullRequest.newBuilder().setSubscription("my subscription").build());
+ * </pre>
+ * @Autowired
```
artembilan: This looks like some leftover after copying from some other place

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
```diff
@@ -151,14 +162,19 @@ public Subscriber subscribe(String subscription, MessageReceiver messageHandler)
 	}
 
 	/**
-	 * Pulls messages on demand using the pull request in argument.
+	 * Pulls messages synchronously, on demand, using the pull request in argument.
```
artembilan: It still `pulls`, not `pull` like we discussed in other PR about JavaDocs.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
```diff
@@ -218,6 +249,20 @@ public PubsubMessage pullNext(String subscription) {
 	public void afterPropertiesSet() throws Exception {
 	}
 
+	/**
+	 * Sets the synchronous pull settings.
```
artembilan: Missed `@param pullSettings` description.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/autoconfig/GcpPubSubAutoConfiguration.java
artembilan: Well, I agree that we should be able to eat user-provided `SusbcriptionAdminSettings` (optional one) in the `pubSubTemplate`, but since this bean is really void, I would prefer do not have it at all since it is going to be used in most cases. So, would be better to move it into the `PubSubTemplate` and create it on demand on first `pull` call if it isn't provided.

meltsufin:
@joaoandremartins @artembilan I refactored the `PubSubTemplate` a little by moving the responsibility of creating a synchronous `PullRequest` and the associated `SubscriptionAdminSettings` to the `SubscriberFactory`. Ultimately, something like the pull timeout and retry settings should be configurable using `GcpPubSubProperties`, which will inform the `DefaultSubscriptinFactory`.

artembilan:
Please, address Travis failure report.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/DefaultSubscriberFactory.java
```diff
@@ -21,16 +21,22 @@
 import com.google.api.gax.rpc.TransportChannelProvider;
 import com.google.cloud.pubsub.v1.MessageReceiver;
 import com.google.cloud.pubsub.v1.Subscriber;
+import com.google.cloud.pubsub.v1.SubscriptionAdminSettings;
+import com.google.pubsub.v1.PullRequest;
 import com.google.pubsub.v1.SubscriptionName;
 
 import org.springframework.cloud.gcp.core.GcpProjectIdProvider;
 import org.springframework.util.Assert;
+import org.threeten.bp.Duration;
```
artembilan: I think this is wrong import. I guess you thought about `java.time.Duration`.

Anyway this looks like unused one...

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/DefaultSubscriberFactory.java
```diff
@@ -67,4 +73,32 @@ public Subscriber getSubscriber(String subscriptionName, MessageReceiver receive
 				.setCredentialsProvider(this.credentialsProvider)
 				.build();
 	}
+
+	@Override
+	public PullRequest getPullRequest(String subscriptionName, Integer maxMessages,
```
artembilan: That's not `get` buy its logic. This is really `create` or `make`.
The `get` assumes to be idempotent and should return the same object for each call.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
```diff
@@ -119,6 +157,62 @@ public Subscriber subscribe(String subscription, MessageReceiver messageHandler)
...
+	 * @param pullRequest pull request containing the subscription name
+	 * @return the list of {@link PubsubMessage} containing the headers and payload
+	 */
+	private List<PubsubMessage> pull(PullRequest pullRequest) {
+		Assert.notNull(pullRequest, "The pull request cannot be null.");
+		Assert.notNull(subscriberFactory.getSubscriptionAdminSettings(),
+				"The SubscriptionAdminSettings cannot be null.");
+
+		try {
+			try (SubscriberStub subscriber = GrpcSubscriberStub.create(
```
artembilan: I wonder if this can go to the `SubscriberFactory` as well.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/DefaultSubscriberFactory.java
```diff
@@ -67,4 +73,32 @@ public Subscriber getSubscriber(String subscriptionName, MessageReceiver receive
...
+			pullRequestBuilder.setReturnImmediately(returnImmediately);
+		}
+
+		return pullRequestBuilder.build();
+	}
+
+	@Override
+	public SubscriptionAdminSettings getSubscriptionAdminSettings() {
+		try {
+			return SubscriptionAdminSettings.newBuilder().build();
```
artembilan: Looks like this can go to the interface as a `default` method.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
meltsufin: Yeah, I guess.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/DefaultSubscriberFactory.java
meltsufin: Unfortunately, the import package is right, but it's not needed here.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/DefaultSubscriberFactory.java
meltsufin: In that case the `getSubscriber` above should also be renamed.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/DefaultSubscriberFactory.java
meltsufin: Not really, because I'll be adding some timeout logic here.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/DefaultSubscriberFactory.java
```diff
@@ -93,9 +100,15 @@ public PullRequest getPullRequest(String subscriptionName, Integer maxMessages,
 	}
 
 	@Override
-	public SubscriptionAdminSettings getSubscriptionAdminSettings() {
+	public SubscriberStub getSubscriberStub() {
```
artembilan: Why `get` again?
Looks like we have just agreed for `create`...

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/DefaultSubscriberFactory.java
```diff
@@ -48,10 +51,13 @@
...
 	private final CredentialsProvider credentialsProvider;
 
+	private final Long pullSimpleTimeoutNoRetriesMillis;
+
 	public DefaultSubscriberFactory(GcpProjectIdProvider projectIdProvider,
 			ExecutorProvider executorProvider,
 			TransportChannelProvider channelProvider,
-			CredentialsProvider credentialsProvider) {
+			CredentialsProvider credentialsProvider,
+			Long pullSimpleTimeoutNoRetriesMillis) {
```
artembilan: In Spring we follow rule like: if options isn't mandatory, it has to be exposed via setter.
We had such a discussion in other PRs. Why are we here again?
Thanks

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/DefaultSubscriberFactory.java
```diff
@@ -48,10 +51,13 @@
 
 	private final CredentialsProvider credentialsProvider;
 
+	private final Long pullSimpleTimeoutNoRetriesMillis;
```
artembilan: May we have not so long name? 
Good JavaDoc should nail the confuse.

Thanks

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/DefaultSubscriberFactory.java
meltsufin: Oops. :)

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
```diff
@@ -114,11 +139,65 @@ public void onSuccess(String result) {
...
+	 * @param pullRequest pull request containing the subscription name
+	 * @return the list of {@link PubsubMessage} containing the headers and payload
+	 */
+	private List<PubsubMessage> pull(PullRequest pullRequest) {
+		Assert.notNull(pullRequest, "The pull request cannot be null.");
+		Assert.notNull(this.subscriberFactory.createSubscriberStub(),
+				"The SubscriberStub cannot be null.");
+
+		try {
+			SubscriberStub subscriber = this.subscriberFactory.createSubscriberStub();
```
meltsufin: TODO: I think we actually need to `close()` this subscriber after we're done.

artembilan: Isn't that what we can do with the `try-with-resources`: https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html?
```
try(SubscriberStub subscriber = this.subscriberFactory.createSubscriberStub()) {
   ...
}
```

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/DefaultSubscriberFactory.java
```diff
@@ -42,10 +51,13 @@
...
 	private final CredentialsProvider credentialsProvider;
 
+	private Long pullTimeoutMillis;
+
 	public DefaultSubscriberFactory(GcpProjectIdProvider projectIdProvider,
 			ExecutorProvider executorProvider,
 			TransportChannelProvider channelProvider,
-			CredentialsProvider credentialsProvider) {
+			CredentialsProvider credentialsProvider,
+			Long pullTimeoutMillis) {
```
artembilan: Why this is still here?
Thanks

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
meltsufin: Yep. Thanks!

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/DefaultSubscriberFactory.java
```diff
@@ -56,15 +68,61 @@ public DefaultSubscriberFactory(GcpProjectIdProvider projectIdProvider,
...
 		return Subscriber.newBuilder(
 				SubscriptionName.of(this.projectId, subscriptionName), receiver)
 				.setChannelProvider(this.channelProvider)
 				.setExecutorProvider(this.executorProvider)
 				.setCredentialsProvider(this.credentialsProvider)
 				.build();
 	}
+
+	@Override
+	public PullRequest createPullRequest(String subscriptionName, Integer maxMessages,
```
joaoandremartins: It's weird for a `SubscriberFactory` to be generating a `PullRequest`, but I guess it's the best option we have because they take in a project ID?
How about creating a `PullRequestFactory`? Too overkill?

spring-cloud-gcp-starters/spring-cloud-gcp-starter-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/autoconfig/GcpPubSubAutoConfiguration.java
```diff
@@ -127,7 +128,8 @@ public SubscriberFactory defaultSubscriberFactory(
 				this.finalProjectIdProvider,
 				executorProvider,
 				channelProvider,
-				this.finalCredentialsProvider);
+				this.finalCredentialsProvider,
+				this.gcpPubSubProperties.getPullSimpleTimeoutNoRetriesMillis());
```
joaoandremartins: Should we add some very simple unit test to make sure this gets populated?

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/DefaultSubscriberFactory.java
joaoandremartins: What happened to the idea of having a `createPullRequest()` method that takes in a timeout? :)

meltsufin: I think it's OK to just have on factory for publishing and one for consuming/subscribing.

meltsufin: I still think it makes sense to do that, but the Pub/Sub client library makes it awkward to implement because the `PullRequest` object does not have a timeout property. Instead timeout and retry configuration has to be configured on the `SubscriberStub`. Let me experiment with this more though.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
```diff
@@ -17,29 +17,47 @@
...
 import org.springframework.cloud.gcp.pubsub.support.SubscriberFactory;
+import org.springframework.util.Assert;
 import org.springframework.util.concurrent.ListenableFuture;
 import org.springframework.util.concurrent.SettableListenableFuture;
 
 /**
  * Default implementation of {@link PubSubOperations}.
  *
+ * <p>
+ * This class's main task is to asynchronously publish to topics and subscribe to subscriptions through the
```
artembilan: I don't know English well, but I remember the rule do not add `s` after apostrophe if the word ends with `s` already. 😄 

Well, we could rephrase the JavaDoc the way to do not say `this class` at all:
```
The main Google Cloud PubSub integration component for asynchronous publishing to topics...
```

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
```diff
@@ -17,29 +17,47 @@
...
+ * {@code publish()} and {@code subscribe()} methods.
+ *
+ * <p>
+ * This class also allows for synchronous message pulling from a subscription through the {@code pull()} and
+ * {@code pullNext()} methods.
```
artembilan: You can use `@link` instead of `@code` to let end-user to jump to the methods during reading JavaDocs.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
```diff
@@ -114,11 +140,64 @@ public void onSuccess(String result) {
...
+		}
+		catch (Exception ioe) {
+			throw new PubSubException("Error pulling messages from subscription "
+					+ pullRequest.getSubscription() + ".", ioe);
+		}
+	}
+
+	@Override
+	public List<PubsubMessage> pull(String subscription, Integer maxMessages, Boolean returnImmediately,
+			RetrySettings retrySettings) {
```
artembilan: This param is out of use.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
meltsufin: You're just being modest! Your English, at least written, is better than that of most native speakers. :smile:

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
meltsufin: I think I'll update without linking to any methods instead.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
meltsufin: ooh good catch!

meltsufin:
@artembilan Do you think your concerns are addressed after the last commit? Thanks!

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubOperations.java
```diff
@@ -97,4 +99,24 @@
 	 */
 	Subscriber subscribe(String subscription, MessageReceiver messageHandler);
 
+	/**
+	 * Pulls and auto-acknowledges a number of messages from a Google Cloud Pub/Sub subscription.
+	 *
```
artembilan: These two lines of the JavaDoc on method level violate our Spring convention about JavaDocs: https://github.com/spring-projects/spring-framework/wiki/Code-Style#javadoc-formatting.

But let it be: we can address it in the separate issue.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/DefaultSubscriberFactory.java
```diff
@@ -59,12 +67,49 @@ public DefaultSubscriberFactory(GcpProjectIdProvider projectIdProvider,
...
+			Boolean returnImmediately) {
+		PullRequest.Builder pullRequestBuilder =
+				PullRequest.newBuilder().setSubscriptionWithSubscriptionName(
+						SubscriptionName.of(this.projectId, subscriptionName));
```
artembilan: I guess `subscriptionName` cannot be empty. Let's ensure that with an appropriate `Assert.notEmpy()` in the begging of the method! Unless you can confirm that this `SubscriptionName.of()` does something on the matter to warn and end-user about  the mistake.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/DefaultSubscriberFactory.java
```diff
@@ -59,12 +67,49 @@ public DefaultSubscriberFactory(GcpProjectIdProvider projectIdProvider,
...
+					.setTransportChannelProvider(this.channelProvider);
+
+			if (retrySettings != null) {
+				subscriptionAdminSettingsBuilder.pullSettings().setRetrySettings(retrySettings);
+			}
+
+			return GrpcSubscriberStub.create(subscriptionAdminSettingsBuilder.build());
+		}
+		catch (IOException e) {
+			throw new RuntimeException("Error creating default SubscriptionAdminSettings", e);
```
artembilan: This message for exception is misleading somehow. We expect `SubscriberStub` as an output but the message says something about some unrelated `SubscriptionAdminSettings`. Moreover *default*... :shrugged:

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/SubscriberFactory.java
```diff
@@ -16,15 +16,23 @@
...
 package org.springframework.cloud.gcp.pubsub.support;
 
+import com.google.api.gax.retrying.RetrySettings;
 import com.google.cloud.pubsub.v1.MessageReceiver;
 import com.google.cloud.pubsub.v1.Subscriber;
+import com.google.cloud.pubsub.v1.stub.SubscriberStub;
+import com.google.pubsub.v1.PullRequest;
 
 /**
  * Generates {@link Subscriber} objects.
```
artembilan: This JavaDoc isn't actual any more. Moreover would be great to have JavaDocs on the methods as well.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/GcpPubSubProperties.java
```diff
@@ -59,5 +59,4 @@ public void setProjectId(String projectId) {
 	public Credentials getCredentials() {
 		return this.credentials;
 	}
-
```
artembilan: Blank like before the end of class.
Well, actually it sounds better like: blank like after each method and property. Or if you want: surround class members with blank lines.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
```diff
@@ -152,13 +153,11 @@ public Subscriber subscribe(String subscription, MessageReceiver messageHandler)
 	 * @param pullRequest pull request containing the subscription name
 	 * @return the list of {@link PubsubMessage} containing the headers and payload
 	 */
-	private List<PubsubMessage> pull(PullRequest pullRequest) {
+	private List<PubsubMessage> pull(PullRequest pullRequest, RetrySettings retrySettings) {
 		Assert.notNull(pullRequest, "The pull request cannot be null.");
-		Assert.notNull(this.subscriberFactory.createSubscriberStub(),
```
joaoandremartins: Why getting rid of this assertion?

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/SubscriberFactory.java
```diff
@@ -33,5 +34,5 @@
 
 	PullRequest createPullRequest(String subscriptionName, Integer maxMessages, Boolean returnImmediately);
 
-	SubscriberStub createSubscriberStub();
+	SubscriberStub createSubscriberStub(RetrySettings retrySettings);
```
joaoandremartins: Can we also add some Javadoc here explaining what `retrySettings` is for?

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubOperations.java
meltsufin: Thanks for pointing this out. I wasn't aware. Can we enforce these rules using checkstyle though?

meltsufin:
@artembilan Regarding the copyright notice, as I understand only the first year of publication needs to be indicated. "The general rule is that the year to include in a copyright notice is the year of first publication of the work." (https://www.copyrightlaws.com/copyright-notice-year/). So, giving copyright year ranges seems redundant. Is this something you do in all Spring projects?

artembilan:
Exactly, @meltsufin ! That's what we do: https://github.com/spring-projects/spring-framework/wiki/Code-Style#license. 

And we had already a discussion about JavaDoc style in other PRs...

spring-cloud-gcp-starters/spring-cloud-gcp-starter-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/GcpPubSubProperties.java
meltsufin: Sounds like something that can be enforced with checkstyle.

artembilan: That's good idea, but I'm not aware of such a checker there: http://checkstyle.sourceforge.net/checks.html. As well as there is no the way to validate JavaDoc style we have talked before...

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubOperations.java
```diff
@@ -97,4 +99,24 @@
...
+	List<PubsubMessage> pull(String subscription, Integer maxMessages,
+			Boolean returnImmediately, RetrySettings retrySettings);
+
+	/**
+	 * Pulls and auto-acknowledges a message from a Google Cloud Pub/Sub subscription.
+	 *
+	 * @param subscription the subscription name
+	 * @return a received message, or {@code null} if none exists in the subscription
+	 */
+	PubsubMessage pullNext(String subscription);
```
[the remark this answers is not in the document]
meltsufin: Yeah, I was considering it too, but then figured if someone wants to specify more settings, they can use the `pull(String subscription, Integer maxMessages, Boolean returnImmediately, RetrySettings retrySettings)`.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/GcpPubSubProperties.java
meltsufin: Maybe this one? http://checkstyle.sourceforge.net/config_whitespace.html#EmptyLineSeparator

artembilan: Oh! Great! Let's give it a shot! 😄

meltsufin: We can! :)
However, looking back at the [Spring code style guide](https://github.com/spring-projects/spring-framework/wiki/Code-Style#blank-lines), I'm not seeing the requirement to put blank lines after methods. Where are you getting this style requirement from?

artembilan: We follow it in our projects. That's really so simple: just surround class members with blank lines.
The SF Docs just might be a bit out dated.

Do you really always need some official doc for confirmation though?..

meltsufin: An official doc is handy and automatic enforcement is even better!
That is if you care about these rules and want them followed when many developers are involved and new ones get on-boarded.

PR #286, opened 2017-12-28
Stackdriver Logging Support

### LINKED ISSUE #154: Add spring-cloud-gcp-logging
Log correlation w/ Spring Cloud Sleuth and GCP Trace IDs
Automatic log module configuration depending on the environment (GCE, GAE Flex, GAE Standard, & Kubernetes)

meltsufin:
This initial implementation includes a Web MVC interceptor that
extracts the request trace ID from the X-Cloud-Trace-Context HTTP
header and makes it available to Stackdriver logging appenders.

This includes spring-cloud-gcp-logging and
spring-cloud-gcp-starter-logging modules with the corresponding unit
tests and readmes.

Reference documentation and a sample app are coming seprately.

Fixes #154.


spring-cloud-gcp-logging/src/main/java/org/springframework/cloud/gcp/logging/TraceIdLoggingWebMvcInterceptor.java
```diff
@@ -0,0 +1,71 @@
...
+ * @author Mike Eltsufin
+ */
+public class TraceIdLoggingWebMvcInterceptor extends HandlerInterceptorAdapter {
+
+	private static final String X_CLOUD_TRACE = "x-cloud-trace-context";
+
+	@Override
+	public boolean preHandle(HttpServletRequest req,
+			HttpServletResponse resp, Object handler) throws Exception {
+		TraceLoggingEnhancer.setCurrentTraceId(extractTraceIdFromRequest(req));
```
artembilan: Why would one set to the `ThreadLocal` null value?

Take a look into the `ThreadLocal` logic, please:
```
public void set(T value) {
        Thread t = Thread.currentThread();
        ThreadLocalMap map = getMap(t);
        if (map != null)
            map.set(this, value);
        else
            createMap(t, value);
    }
```

So, if there is no a `Map` associated with the current thread, it is created independently of the `value`. Therefore when it is `null` we create extra object for nothing.

spring-cloud-gcp-logging/src/main/java/org/springframework/cloud/gcp/logging/TraceIdLoggingWebMvcInterceptor.java
```diff
@@ -0,0 +1,71 @@
...
+		return true;
+	}
+
+	@Override
+	public void afterCompletion(HttpServletRequest httpServletRequest,
+			HttpServletResponse httpServletResponse, Object handler, Exception e) throws Exception {
+		TraceLoggingEnhancer.setCurrentTraceId(null);
```
artembilan: Yeah... That's not good. There is a `ThreadLocal.remove()` on the matter.
I mean that thread might be reused somewhere else, outside of the servlet container, but its state is polluted with a map with a `null` value.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-logging/README.adoc
```diff
@@ -0,0 +1,53 @@
...
+
+  <root level="INFO">
+    <appender-ref ref="STACKDRIVER" />
+  </root>
+</configuration>
+----
+
+Note that, when the application is run locally, the logs will go into the "spring.log"
+channel of the "Global" resource. However, when running on GCP on App Engine or GKE, the log channel
+will under the appropriate monitored resource.
```
joaoandremartins: will "run" under...?

spring-cloud-gcp-starters/spring-cloud-gcp-starter-logging/src/test/java/org/springframework/cloud/gcp/logging/autoconfig/StackdriverLoggingAutoConfigurationTests.java
```diff
@@ -0,0 +1,73 @@
...
+		int count = 0;
+		for (int i = 0; i < this.interceptors.length; i++) {
+			if (this.interceptors[i] instanceof TraceIdLoggingWebMvcInterceptor) {
+				count++;
+			}
+		}
+		return count;
+	}
+
+	public static class StackdriverLoggingAutoConfigurationTest_default
```
joaoandremartins: Just curious, are these tests run by Maven?
I thought the classes needed to end in "Test" or "Tests".

joaoandremartins:
Also, it sounds like this PR enables matching logs to trace entries.
Would it make more sense to add this to the trace module?

spring-cloud-gcp-logging/src/main/java/org/springframework/cloud/gcp/logging/TraceIdLoggingWebMvcInterceptor.java
```diff
@@ -0,0 +1,71 @@
...
+	 *
+	 * @param req The HTTP servlet request.
+	 * @return The trace ID or null, if none found.
+	 */
+	public String extractTraceIdFromRequest(HttpServletRequest req) {
+		String traceId = (String) req.getHeader(X_CLOUD_TRACE);
+
+		if (traceId != null) {
+			int slash = traceId.indexOf('/');
+			if (slash >= 0) {
```
ChengyuanZhao: Is it possible for the X_CLOUD_TRACE value to start with the slash? If that's the case should this return null or just the empty string?

spring-cloud-gcp-logging/src/main/java/org/springframework/cloud/gcp/logging/LoggingWebMvcConfigurer.java
```diff
@@ -0,0 +1,43 @@
...
+ * MVC Adapter that adds the {@link TraceIdLoggingWebMvcInterceptor}
+ *
+ * @author Mike Eltsufin
+ */
+@Configuration
+public class LoggingWebMvcConfigurer extends WebMvcConfigurerAdapter {
+
+	private TraceIdLoggingWebMvcInterceptor interceptor;
+
+	public LoggingWebMvcConfigurer(TraceIdLoggingWebMvcInterceptor interceptor) {
```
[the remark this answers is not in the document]
meltsufin: The idea is that the `TraceIdLoggingWebMvcInterceptor` bean will be created in the autoconfig class `StackdriverLoggingAutoConfiguration` or provided by the app.

`@EnableStackdriverLogging` could be an interesting trick. I'll look into that.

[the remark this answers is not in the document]
artembilan: OK! But if it isn't auto-config and the app doesn't provide such a bean?

Do we have a test on the matter?

So, just consider to create `TraceIdLoggingWebMvcInterceptor` here if one isn't provided.
Something `@Autowired(required = false)` should help here.

spring-cloud-gcp-logging/src/main/java/org/springframework/cloud/gcp/logging/TraceIdLoggingWebMvcInterceptor.java
meltsufin: As I understand, the map entry will be created anyway on the first call to `ThreadLocal.get()`. So, we might as well pay the price upfront.
I could put a guard for `null` if you still think it's worth it.

spring-cloud-gcp-logging/src/main/java/org/springframework/cloud/gcp/logging/TraceIdLoggingWebMvcInterceptor.java
meltsufin: Agreed, but we're restricted by the `TraceLoggingEnhancer` API, which only has `setCurrentTraceId` and `getCurrentTraceId`.
This should be fixed upstream first.

spring-cloud-gcp-logging/src/main/java/org/springframework/cloud/gcp/logging/TraceIdLoggingWebMvcInterceptor.java
artembilan: Right, and that is really painful for Java per se.

But from here, in this code, you can't guarantee that someone always is going to call `ThreadLocal.get()` downstream 😄

spring-cloud-gcp-starters/spring-cloud-gcp-starter-logging/pom.xml
```diff
@@ -0,0 +1,41 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  ~  Copyright 2017 original author or authors.
```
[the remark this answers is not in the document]
meltsufin: Good catch! I think the IDE did this for me. :smile:

spring-cloud-gcp-logging/src/main/java/org/springframework/cloud/gcp/logging/TraceIdLoggingWebMvcInterceptor.java
ChengyuanZhao: In any case I don't think it's appropriate to address the issue of `null` trace IDs here because TraceLoggingEnhancer already expects the possibility of null trace ID:
```  @Override
  public void enhanceLogEntry(com.google.cloud.logging.LogEntry.Builder builder) {
    String traceId = getCurrentTraceId();
    if (traceId != null) {
      builder.addLabel(traceIdLabel, traceId);
    }
  }```
Though we might have a different opinion on the efficiency impact of that it doesn't feel right to circumvent the logging team's implementation.

File: spring-cloud-gcp-starters/spring-cloud-gcp-starter-logging/pom.xml (hunk 7 above)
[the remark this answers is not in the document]
Comment (artembilan) [reply spring-cloud-gcp-starters/spring-cloud-gcp-starter-logging/pom.xml]: Ha-ha! And now we are in the 2018 anyway 😄

File: spring-cloud-gcp-starters/spring-cloud-gcp-starter-logging/src/main/java/org/springframework/cloud/gcp/logging/autoconfig/StackdriverLoggingAutoConfiguration.java [hunk 8]
```diff
@@ -0,0 +1,43 @@
...
+import org.springframework.context.annotation.Import;
+
+/**
+ * @author Mike Eltsufin
+ */
+@Configuration
+@ConditionalOnClass({ TraceIdLoggingWebMvcInterceptor.class, TraceLoggingEnhancer.class })
+@ConditionalOnProperty(value = "spring.cloud.gcp.logging.enabled", matchIfMissing = true)
+@Import(LoggingWebMvcConfigurer.class)
+public class StackdriverLoggingAutoConfiguration {
```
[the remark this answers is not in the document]
meltsufin: Naturally. :)

spring-cloud-gcp-starters/spring-cloud-gcp-starter-logging/src/test/java/org/springframework/cloud/gcp/logging/autoconfig/StackdriverLoggingAutoConfigurationTests.java
meltsufin: Yep they run. Maven/JUnit only really care about the `@Test` annotation on the methods.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-logging/pom.xml
```diff
@@ -0,0 +1,41 @@
...
+		 xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+	<parent>
+		<artifactId>spring-cloud-gcp-starters</artifactId>
+		<groupId>org.springframework.cloud</groupId>
+		<version>1.0.0.BUILD-SNAPSHOT</version>
+	</parent>
+	<modelVersion>4.0.0</modelVersion>
+
+	<artifactId>spring-cloud-gcp-starter-logging</artifactId>
+	<name>Spring Cloud GCP Cloud Logging Starter</name>
```
joaoandremartins: Should we commit to only using "Stackdriver Logging" as the product name?

spring-cloud-gcp-logging/src/main/java/org/springframework/cloud/gcp/logging/TraceIdLoggingWebMvcInterceptor.java
meltsufin: It should never start with the slash. The format is described [here](https://cloud.google.com/trace/docs/support).

spring-cloud-gcp-starters/spring-cloud-gcp-starter-logging/pom.xml
meltsufin: Yeah, I guess.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-logging/src/test/java/org/springframework/cloud/gcp/logging/autoconfig/StackdriverLoggingAutoConfigurationTests.java
joaoandremartins: They didn't run when I just mvn installed the starter locally.

-------------------------------------------------------
 T E S T S
-------------------------------------------------------

Results :

Tests run: 0, Failures: 0, Errors: 0, Skipped: 0

meltsufin:
> Also, it sounds like this PR enables matching logs to trace entries.
> Would it make more sense to add this to the trace module?

It doesn't exactly interoperate with Trace module yet because the Trace module doesn't use the `x-cloud-trace-context` HTTP header. Instead, it uses the Zipkin `X-B3-TraceId` header. So we should either modify Logging to also check `X-B3-TraceId` and somehow decide which one to use if both are present, or have our Trace module prefer `x-cloud-trace-context` when it's present.
cc: @saturnism

spring-cloud-gcp-logging/src/main/java/org/springframework/cloud/gcp/logging/TraceIdLoggingWebMvcInterceptor.java
meltsufin: Created issue: https://github.com/GoogleCloudPlatform/google-cloud-java/issues/2746

spring-cloud-gcp-logging/src/main/java/org/springframework/cloud/gcp/logging/TraceIdLoggingWebMvcInterceptor.java
meltsufin: https://github.com/GoogleCloudPlatform/google-cloud-java/issues/2746

spring-cloud-gcp-starters/spring-cloud-gcp-starter-logging/pom.xml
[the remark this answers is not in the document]
meltsufin: PR opened in 2017 :rofl:

spring-cloud-gcp-logging/pom.xml
```diff
@@ -0,0 +1,43 @@
...
+            <artifactId>spring-webmvc</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>javax.servlet</groupId>
+            <artifactId>javax.servlet-api</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>com.google.cloud</groupId>
+            <artifactId>google-cloud-logging-logback</artifactId>
+            <version>0.32.0-alpha</version>
```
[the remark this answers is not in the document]
meltsufin: Not yet, https://github.com/GoogleCloudPlatform/google-cloud-java/issues/2745.
However, I'll move it to our own BOM.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-logging/pom.xml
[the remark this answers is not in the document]
artembilan: That's not related. An important part when the change appears in the `master`.

You can do in the PR anything you want and can, but that doesn't mean that your custom code will go to upstream for broader audience.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-logging/src/test/java/org/springframework/cloud/gcp/logging/autoconfig/StackdriverLoggingAutoConfigurationTests.java
meltsufin: You're right! Good catch!

joaoandremartins:
Tinkered with JUL logging, but wasn't able to see anything in Cloud Logging.
We should probably explain better what this module/starter does. Or maybe what we need is a good sample that shows what users should see. But from the READMEs description, I wasn't able to understand very well what this does.

meltsufin:
@joaoandremartins Can you please try it with Logback because that is the only one I really documented, and it is recommended by Spring Boot over JUL.

@ChengyuanZhao Have you tried using the starter?

I totally agree that screenshots and a better explanation of why one would use the starter is in order. I plan to add all that to the sample app.

Otherwise, I believe I've addressed all the review comments. Please take another look.

ChengyuanZhao:
@meltsufin I haven't tried the starter yet. Would it be worthwhile to add a sample for this under the spring-cloud-gcp-samples for logging?

spring-cloud-gcp-logging/src/main/java/org/springframework/cloud/gcp/logging/LoggingWebMvcConfigurer.java
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * MVC Adapter that adds the {@link TraceIdLoggingWebMvcInterceptor}
+ *
+ * @author Mike Eltsufin
+ */
+@Configuration
+public class LoggingWebMvcConfigurer extends WebMvcConfigurerAdapter {
+
+	private TraceIdLoggingWebMvcInterceptor interceptor;
```
artembilan: Must be `final`

spring-cloud-gcp-logging/src/main/java/org/springframework/cloud/gcp/logging/LoggingWebMvcConfigurer.java
```diff
@@ -0,0 +1,47 @@
+/*
+ *  Copyright 2017 original author or authors.
```
artembilan: I still insist that new classes have to make it into the `master` in the current year.

meltsufin:
> Would it be worthwhile to add a sample for this under the spring-cloud-gcp-samples for logging?

Yes. In fact, I have an issue for that #287.

ChengyuanZhao:
@meltsufin I'll branch from this branch and add a sample and at the same time it'll allow me to try out the starter

meltsufin:
@ChengyuanZhao Sounds good. Thanks!

meltsufin:
@artembilan Last comments addressed as well now.

spring-cloud-gcp-logging/pom.xml
```diff
@@ -0,0 +1,42 @@
...
+    <description>Spring Cloud GCP Stackdriver Logging Module</description>
+
+    <dependencies>
+        <dependency>
+            <groupId>org.springframework</groupId>
+            <artifactId>spring-webmvc</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>javax.servlet</groupId>
+            <artifactId>javax.servlet-api</artifactId>
```
artembilan: This must be `provided`. That is the target Servlet platform responsibility to bring this API.
See SF dependency management: https://github.com/spring-projects/spring-framework/blob/master/spring-webmvc/spring-webmvc.gradle#L16

spring-cloud-gcp-logging/pom.xml
```diff
@@ -0,0 +1,42 @@
...
+        </dependency>
+
+        <dependency>
+            <groupId>org.springframework</groupId>
+            <artifactId>spring-test</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>commons-logging</groupId>
+            <artifactId>commons-logging</artifactId>
```
artembilan: I think we don't need this dependency at all.
What am I missing?
Thanks

meltsufin: `TraceIdLoggingWebMvcInterceptorTests` fails without it because `MockHttpServletRequest` needs it. It's just a test dependency anyway.

spring-cloud-gcp-logging/pom.xml
meltsufin: Good catch.

artembilan:
Travis somehow doesn't like your latest changes. The test has just failed

meltsufin:
Dependencies with `provided` scope are not transitive. So, I have to add `servlet-api` to the starter module as well.

artembilan:
> Dependencies with provided scope are not transitive.

That's correct. That's why we have to specify it explicitly.
But the concern that we should specify this `provided` only for the compile time. I don't think that we should push such a dependency to the end-user level.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-logging/pom.xml
```diff
@@ -0,0 +1,30 @@
...
+			<artifactId>spring-cloud-gcp-starter-core</artifactId>
+		</dependency>
+		<dependency>
+			<groupId>org.springframework.cloud</groupId>
+			<artifactId>spring-cloud-gcp-logging</artifactId>
+		</dependency>
+		<dependency>
+			<groupId>javax.servlet</groupId>
+			<artifactId>javax.servlet-api</artifactId>
+			<scope>provided</scope>
```
artembilan: That's what I mean. We don't have to push end-users to use this dependency.

You may take a look into the similar `web` starter in the Spring Boot: https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-starters/spring-boot-starter-web/pom.xml

meltsufin:
> I don't think that we should push such a dependency to the end-user level.

@artembilan What do you mean by that? With the `provided` scope, how is it pushed to the end-users?

My original omission of the `provided` scope was just a mistake, not intentional.

artembilan:
I mean that this `javax.servlet-api` should not be present in the `starter`. See similar Spring Boot stuff.
We need it really only for the compile time of our own code.

meltsufin:
It's needed for the tests in the starter. Hence the previous Travis failure you pointed out. I guess the best thing to do is to set it as a `test` scope dependency there.

artembilan:
> It's needed for the tests in the starter

Yeah... That's why `starter` should not have any code. Again: see Spring Boot and Stephane's concerns in the other PRs... 

Let it be for now, but we should keep it in mind when we refactor it to `autoconfigure` module and proper `starter` state.

meltsufin:
@artembilan Completely agree. The code from this starter will move to the `autoconfigure` module after #276 is merged.

spring-cloud-gcp-logging/README.adoc
```diff
@@ -0,0 +1,63 @@
...
+  <include resource="org/springframework/cloud/gcp/logging/logback-appender.xml" />
+
+  <root level="INFO">
+    <appender-ref ref="STACKDRIVER" />
+  </root>
+</configuration>
+----
+
+`STACKDRIVER_LOG_NAME` and `STACKDRIVER_LOG_FLUSH_LEVEL` environment variables can be used to customize the
+`STACKDIRVER` appender.
```
artembilan: Typo. 😄

meltsufin: haha :blush:

PR #290, opened 2018-01-03
Adds spring-boot-starter to all Spring Cloud GCP Boot starters

joaoandremartins:
As recommended in https://docs.spring.io/spring-boot/docs/current-SNAPSHOT/reference/htmlsingle/#boot-features-custom-starter-module-starter

### DIFF
--- a/spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/pom.xml
+++ b/spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/pom.xml
@@ -44,5 +44,10 @@
 			<groupId>org.springframework.boot</groupId>
 			<artifactId>spring-boot-autoconfigure</artifactId>
 		</dependency>
+
+		<dependency>
+			<groupId>org.springframework.boot</groupId>
+			<artifactId>spring-boot-starter</artifactId>
+		</dependency>
 	</dependencies>
 </project>
--- a/spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/pom.xml
+++ b/spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/pom.xml
@@ -42,11 +42,6 @@
 
 		<dependency>
 			<groupId>org.springframework.boot</groupId>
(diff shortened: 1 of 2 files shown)


joaoandremartins:
@snicoll FYI

snicoll:
@joaoandremartins that's not how I would do this. I prefer things to be explicit and not redudant if possible. 

Besides, if we follow our chain of thoughts of having a `spring-cloud-gcp-starter`(that itself brings `spring-cloud-starter`), this PR will be irrelevant anyway: all GCP starters (besides the main one of course) would bring `spring-cloud-gcp-starter` and that one itself will bring the base stuff. This is the same pattern that we use [in Spring  Boot](https://github.com/spring-projects/spring-boot/blob/bee5fa7fc625e3878523834240c4e85746b2cd0c/spring-boot-project/spring-boot-starters/spring-boot-starter-web/pom.xml#L19)

spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/pom.xml
```diff
@@ -44,5 +44,10 @@
 			<groupId>org.springframework.boot</groupId>
 			<artifactId>spring-boot-autoconfigure</artifactId>
```
snicoll: this dependency is no longer necessary now.

joaoandremartins:
Currently, all our other starters already import `spring-cloud-gcp-starter-core`, so that should be sorted out.
`spring-cloud-gcp-starter-core` should be regarded as our core starter. What is required for it to be in a proper state?

snicoll:
> Currently, all our other starters already import spring-cloud-gcp-starter-core

That's not the case in the branch you are working on for the autoconfigure stuff. Perhaps rebasing would be in order?

joaoandremartins:
My goal is to apply this in master first (to become effective immediately) and then rebase/merge with the `autoconfigure` branch.

snicoll:
@joaoandremartins OK!

PR #291, opened 2018-01-05
Added Stackdriver logging sample

### LINKED ISSUE #287: Create sample app to demonstrate Stackdriver Logging

ChengyuanZhao:
fixes #287


spring-cloud-gcp-samples/spring-cloud-gcp-logging-sample/README.adoc
```diff
@@ -0,0 +1,33 @@
+= Spring Cloud GCP Logging Starter Example
+
+This sample application demonstrates using the
+link:../../spring-cloud-gcp-starters/spring-cloud-gcp-starter-logging[Spring Cloud GCP Logging Starter] in your code.
+
+== Setup & Configuration
+1. Create a Google Cloud Platform Project
+1. https://cloud.google.com/docs/authentication/getting-started#creating_the_service_account[Create a service account]
```
meltsufin: I don't think creating a service account is necessary. Application Default credentials work fine for logging.

spring-cloud-gcp-samples/spring-cloud-gcp-logging-sample/README.adoc
```diff
@@ -0,0 +1,33 @@
...
+with Trace Append permission. Furnish a new JSON key and then set the credentials using the
```
meltsufin: Why the need for Trace Append permission? We're just using the trace ID as a "request id".

spring-cloud-gcp-samples/spring-cloud-gcp-logging-sample/README.adoc
```diff
@@ -0,0 +1,33 @@
...
+`GOOGLE_APPLICATION_CREDENTIALS` environment variable or
+link:../../spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/README.adoc[using GCP Starter Core properties].
++
+Alternatively, if you have the https://cloud.google.com/sdk/[Google Cloud SDK] installed and
```
meltsufin: I think I would offer this as the first option because it's easier to get started with.

spring-cloud-gcp-samples/spring-cloud-gcp-logging-sample/README.adoc
```diff
@@ -0,0 +1,33 @@
...
+initialized and are logged in with
+https://developers.google.com/identity/protocols/application-default-credentials[application
+default credentials], you can skip this step since Spring Cloud GCP will auto-discover those
+settings for you.
+
+1. Enable the Stackdriver Logging API.
```
meltsufin: Is it not on by default? Not sure.

spring-cloud-gcp-samples/spring-cloud-gcp-logging-sample/README.adoc
```diff
@@ -0,0 +1,33 @@
...
+
+----
+mvn appengine:deploy
+----
+
+Browse to `http://[Your-Project-ID].appspot.com/log`. This will trigger the `ExampleController.log()` method,
+which uses a typical logger supporting Logback to write a line of text to the Stackdriver log.
+
+To see the logged line, navigate to the Logging section of the Cloud Console.
+Select your App Engine deployment as the resource and look under the `spring.log` channel to see the
```
meltsufin: We should also direct the user to expand the log entry metadata and verify that the trace_id label is set.

Also, one of the major benefits of attaching a trace id to log messages is that you can expand the request log line and see all of the associated log messages created as part of processing that request.

spring-cloud-gcp-samples/spring-cloud-gcp-logging-sample/src/main/java/com/example/ExampleController.java
```diff
@@ -0,0 +1,40 @@
...
+ * @author Chengyuan Zhao
+ */
+@RestController
+public class ExampleController {
+	private static final Log LOGGER = LogFactory.getLog(ExampleController.class);
+
+	@RequestMapping("/log")
+	public String log() {
+		String message = "This line was written to the log.";
+		LOGGER.info(message);
```
meltsufin: Let's write more than one message to demonstrate that the same request trace ID is added to both.

spring-cloud-gcp-samples/spring-cloud-gcp-logging-sample/README.adoc
```diff
@@ -0,0 +1,33 @@
...
++
+Alternatively, if you have the https://cloud.google.com/sdk/[Google Cloud SDK] installed and
+initialized and are logged in with
+https://developers.google.com/identity/protocols/application-default-credentials[application
+default credentials], you can skip this step since Spring Cloud GCP will auto-discover those
+settings for you.
+
+1. Enable the Stackdriver Logging API.
+
+== Run the Example
```
meltsufin: It might be worthwhile to also talk about running the example locally. In that case, the "Resource" to which the `spring.log` stream will be assigned will be "Global" because it's not running on App Engine.

meltsufin:
@ChengyuanZhao You can rebase on `master` now.

ChengyuanZhao:
@meltsufin I made changes according to the review, PTAL

spring-cloud-gcp-samples/spring-cloud-gcp-logging-sample/README.adoc
```diff
@@ -0,0 +1,37 @@
...
+link:../../spring-cloud-gcp-starters/spring-cloud-gcp-starter-logging[Spring Cloud GCP Logging Starter] in your code.
+
+== Setup & Configuration
+1. Create a Google Cloud Platform Project
+1. Have the https://cloud.google.com/sdk/[Google Cloud SDK] installed and
+initialized and logged in with
+https://developers.google.com/identity/protocols/application-default-credentials[application
+default credentials].
+
+1. Check that the Stackdriver Logging API is enabled.
```
meltsufin: Link?

spring-cloud-gcp-samples/spring-cloud-gcp-logging-sample/README.adoc
```diff
@@ -0,0 +1,37 @@
...
+
+== Run the Example
+Deploy the sample application from Maven:
```
meltsufin: Maybe clarify that you're deploying to "App Engine flexible environment"

spring-cloud-gcp-samples/spring-cloud-gcp-logging-sample/README.adoc
```diff
@@ -0,0 +1,37 @@
...
+
+----
+mvn appengine:deploy
+----
+
+Browse to `http://[Your-Project-ID].appspot.com/log`. This will trigger the `ExampleController.log()` method,
+which uses a typical logger supporting Logback to write a line of text to the Stackdriver log.
+
+To see the logged lines, navigate to the Logging section of the Cloud Console.
```
meltsufin: Link to the Stackdriver Logging view?

spring-cloud-gcp-samples/spring-cloud-gcp-logging-sample/README.adoc
```diff
@@ -0,0 +1,37 @@
...
+Select your App Engine deployment as the resource and look under the `spring.log` channel to see the
+logged text. You can further narrow down the log by request by using the following filtering text
```
meltsufin: Maybe also add an instruction to "Expand the log entry labels and verify that `appengine.googleapis.com/trace_id` label is set to the same value for both log messages."

spring-cloud-gcp-samples/spring-cloud-gcp-logging-sample/README.adoc
```diff
@@ -0,0 +1,37 @@
...
+that takes advantage of the trace ID label in each logged line:
+
+----
+resource.type="gae_app"
+logName="projects/[Your-Project-ID]/logs/spring.log"
+labels."appengine.googleapis.com/trace_id"="[Trace-ID-Of-Request]"
+----
+
+If you run this application locally, then these logs will appear in Stackdriver under the `Global`
```
meltsufin: "locally, by using `mvn spring-boot:run`...."

ChengyuanZhao:
@meltsufin changes made

spring-cloud-gcp-samples/spring-cloud-gcp-logging-sample/README.adoc
```diff
@@ -0,0 +1,40 @@
...
+1. Create a Google Cloud Platform Project
+1. Have the https://cloud.google.com/sdk/[Google Cloud SDK] installed and
+initialized and logged in with
+https://developers.google.com/identity/protocols/application-default-credentials[application
+default credentials].
+
+1. Check that the https://cloud.google.com/logging/docs/access-control[Stackdriver Logging API is enabled].
+
+== Run the Example
+Deploy the sample application from Maven to App Engine Flexible Environment:
```
joaoandremartins: Instead of using App Engine to test, can we use curl locally and adding the header?
Would prefer to avoid any connections to App Engine unless they're strictly required.

spring-cloud-gcp-samples/spring-cloud-gcp-logging-sample/pom.xml
```diff
@@ -0,0 +1,43 @@
...
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>com.google.cloud.tools</groupId>
+                <artifactId>appengine-maven-plugin</artifactId>
+                <version>1.3.2</version>
+            </plugin>
+            <plugin>
+                <groupId>org.eclipse.jetty</groupId>
+                <artifactId>jetty-maven-plugin</artifactId>
```
joaoandremartins: Why is this plugin needed?

spring-cloud-gcp-samples/spring-cloud-gcp-logging-sample/src/main/appengine/app.yaml
```diff
@@ -0,0 +1,9 @@
+runtime: java
```
joaoandremartins: I'd erase this file and not use App Engine in our sample.

spring-cloud-gcp-samples/spring-cloud-gcp-logging-sample/src/main/java/com/example/ExampleController.java
```diff
@@ -0,0 +1,42 @@
...
+/**
+ * Sample REST Controller to demonstrate Stackdriver Logging
+ *
+ * @author Chengyuan Zhao
+ */
+@RestController
+public class ExampleController {
+	private static final Log LOGGER = LogFactory.getLog(ExampleController.class);
+
+	@RequestMapping("/log")
```
joaoandremartins: I think `@GetMapping` is the new `@RequestMapping`

spring-cloud-gcp-samples/spring-cloud-gcp-logging-sample/README.adoc
ChengyuanZhao: This is a sample though, not a test. It's nice that we can show how this would work with a GAE deployment since we're doing this for GCP.

spring-cloud-gcp-samples/spring-cloud-gcp-logging-sample/pom.xml
ChengyuanZhao: The instructions for setting up `mvn appengine:deploy` to work mentions it: https://cloud.google.com/appengine/docs/flexible/java/using-maven

spring-cloud-gcp-samples/spring-cloud-gcp-logging-sample/src/main/appengine/app.yaml
ChengyuanZhao: I think there's some value to get people to try this out on GAE. It could introduce them to using GCP and also some other examples for other GCP products require you to setup a GCP account anyway, and it would be good to use this here. 

I mean really i think it's much more likely somebody is using stackdriver GCP logging if they're also using GCP to host their application...

spring-cloud-gcp-samples/spring-cloud-gcp-logging-sample/README.adoc
joaoandremartins: I'm sorry, but I can't approve this PR if it contains this GAE bit.
Can talk more offline if you want.

joaoandremartins: Actually, using Spring Cloud GCP logging with GAE Flex (or Standard) sounds great for a codelab at g.co/codelabs/spring.

ChengyuanZhao: Could you explain why you feel strongly against having it in?

ChengyuanZhao:
@joaoandremartins I emphasized running locally as the primary way to use this sample, but I really want to keep the GAE deployment in because otherwise the sample does not have that trace ID header, and this sample really loses its value. This sample is simple enough that really all the user has to do is `mvn appengine:deploy`, and it should work. @meltsufin what do you think?

joaoandremartins:
@ChengyuanZhao you only need to spin up a local app with our starter enabled and shoot a request using e.g., `curl --header "X-Cloud-Trace-Context: trace-id-123" localhost:8080/log`

ChengyuanZhao:
@joaoandremartins But why bother doing that when the point of alot of this is to make it feel seamless and "magical" ? I think it's really worthwhile to use GAE here and show that GCP adds this header and gives it a unique value all by itself.

spring-cloud-gcp-samples/spring-cloud-gcp-logging-sample/README.adoc
ChengyuanZhao: done

spring-cloud-gcp-samples/spring-cloud-gcp-logging-sample/README.adoc
ChengyuanZhao: done

spring-cloud-gcp-samples/spring-cloud-gcp-logging-sample/README.adoc
ChengyuanZhao: done

spring-cloud-gcp-samples/spring-cloud-gcp-logging-sample/README.adoc
ChengyuanZhao: done

spring-cloud-gcp-samples/spring-cloud-gcp-logging-sample/README.adoc
ChengyuanZhao: done

spring-cloud-gcp-samples/spring-cloud-gcp-logging-sample/pom.xml
meltsufin: It's for running `WAR` App Engine Flex apps locally. For Spring Boot `JAR`, we don't need it because we can use `mvn spring-boot:run`.

meltsufin:
I would prefer both local and cloud execution of the sample to be documented. For local, mention the `curl` trick. I think having both will also clarify how the starter works.

ChengyuanZhao:
@meltsufin done, PTAL

meltsufin:
@ChengyuanZhao Can we have two titled sections "Running Locally" and "Running on App Engine Flexible Environment"? This way people can skip more easily to how they want to test it.

ChengyuanZhao:
@meltsufin I don't think that's a good idea, because the text about looking on the stackdriver UI and looking at the trace IDs is common between the two. Also, I think people should read the entirety of the text to know the differences that running on GAE provides vs locally, and in this case the more succinct single section makes more sense in my humble opinion.

spring-cloud-gcp-samples/spring-cloud-gcp-logging-sample/README.adoc
```diff
@@ -0,0 +1,47 @@
...
+label appear in Stackdriver:
+----
+curl --header "x-cloud-trace-context: [Some-Trace-ID]" localhost:8080/log
+----
+
+To see the logged lines, navigate to the https://pantheon.corp.google.com/logs/viewer[Logging section of the Cloud Console].
+Select the `Global` resource and look under the `spring.log` channel to see the
+logged text.
+
+If your application is deployed to Google App Engine or Google Container Engine,
```
meltsufin: It's "Kubernetes Engine" now.

spring-cloud-gcp-samples/spring-cloud-gcp-logging-sample/README.adoc
```diff
@@ -0,0 +1,47 @@
...
+you will instead find the logs under the deployments resource in the Stackdriver UI under the `spring.log` channel.
```
meltsufin: s/deployment/deployment's?

spring-cloud-gcp-samples/spring-cloud-gcp-logging-sample/README.adoc
```diff
@@ -0,0 +1,47 @@
...
+(by using `mvn appengine:deploy` in this sample) you can use the following to filter by the trace ID:
+
+----
+resource.type="gae_app"
+logName="projects/[Your-Project-ID]/logs/spring.log"
+labels."appengine.googleapis.com/trace_id"="[Trace-ID-Of-Request]"
+----
+
+Expand the log entry labels and note that `appengine.googleapis.com/trace_id` label is set to the
+same value for both log messages. The trace ID header when deployed to App Engine is set automatically.
```
meltsufin: Maybe call out the trace ID header `x-cloud-trace-context` for clarity?

spring-cloud-gcp-samples/spring-cloud-gcp-logging-sample/pom.xml
meltsufin: `jetty-maven-plugin` is not needed here because we're building a `jar`-packaged app.

ChengyuanZhao: Ah I see. latest changes made.

PR #295, opened 2018-01-09
Added secondary header for logging trace id

### LINKED ISSUE #289: Log correlation with Spring Cloud Sleuth and GCP Trace IDs
Currently, Logging and Trace support don't interoperate to provide trace and log message correlation because the Trace module doesn't use the `x-cloud-trace-context` HTTP header. Instead, it uses the Zipkin `X-B3-TraceId` header. So we should either modify Logging to also check `X-B3-TraceId` and somehow decide which one to use if both are present, or have our Trace module prefer `x-cloud-trace-context` when it's present.

Related to #154 and #286

ChengyuanZhao:
fixes #289

### DIFF
--- a/spring-cloud-gcp-logging/src/main/java/org/springframework/cloud/gcp/logging/TraceIdLoggingWebMvcInterceptor.java
+++ b/spring-cloud-gcp-logging/src/main/java/org/springframework/cloud/gcp/logging/TraceIdLoggingWebMvcInterceptor.java
@@ -16,10 +16,13 @@
 
 package org.springframework.cloud.gcp.logging;
 
+import java.util.List;
+
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
 import com.google.cloud.logging.TraceLoggingEnhancer;
+import com.google.common.collect.ImmutableList;
 
 import org.springframework.web.servlet.handler.HandlerInterceptorAdapter;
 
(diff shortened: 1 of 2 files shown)


spring-cloud-gcp-logging/src/main/java/org/springframework/cloud/gcp/logging/TraceIdLoggingWebMvcInterceptor.java
```diff
@@ -37,6 +40,11 @@
 
 	private static final String X_CLOUD_TRACE = "x-cloud-trace-context";
 
+	private static final String X_B3_TRACE = "X-B3-TraceId";
+
+	private static final List<String> TRACE_HEADERS = ImmutableList.of(X_CLOUD_TRACE,
```
meltsufin: It's possible that both headers are set. What do we do in that case? Allow the precedence order to be configurable? Error out?

Also, I was thinking of an alternative design where for Zipkin we would have another implementation of `HandlerInterceptorAdapter`, and the user can choose which one (or both) to configure.

spring-cloud-gcp-logging/src/test/java/org/springframework/cloud/gcp/logging/TraceIdLoggingWebMvcInterceptorTests.java
```diff
@@ -33,10 +33,14 @@
 
 	public static final String TEST_TRACE_ID = "105445aa7843bc8bf206b120001000";
 
+	public static final String TEST_TRACE_ID_2 = "205445aa7843bc8bf206b120001000";
```
meltsufin: We should probably call this `ZIPKIN_TRACE_ID_HEADER_VALUE` and the one below `GCP_TRACE_ID_HEADER_VALUE` for clarity.

Issue #296, opened 2018-01-09
Add config choices for stackdriver logging trace id header


ChengyuanZhao:
@meltsufin mentioned potentially adding choices for the user to choose their own prioritizing `x-cloud-trace-context` over `x-b3-traceid` or vice versa if both are present. Also a setting that would throw an exception if both are present and user only wants one to be present.

meltsufin:
See: https://github.com/spring-cloud/spring-cloud-gcp/pull/295#pullrequestreview-87583173

Issue #288, opened 2017-12-28
Add reference documentation for Stackdriver Logging support


meltsufin:

joaoandremartins:
When did we intend to add this?

meltsufin:
Before M2. :)

joaoandremartins:
#346 added Logging content to the refdoc.
@meltsufin would you mind going over it and applying any necessary fixes, or closing this issue?

meltsufin:
Will do.

meltsufin:
Since the documentation is already there, I think we can close this PR. However, I don't think logging documentation should be lumped together with trace docs.

joaoandremartins:
It might be me who (still) doesn't understand what the logging module does, but I think in its current form, the logging module only really makes sense if coupled to the trace module?

That might change if we add any other functionality to the logging module in the future.

Feel free to suggest docs changes.

meltsufin:
They are definitely related, especially since the main feature we're adding is passing around the Trace ID. However, just like Stackdriver Logging can be used independently from Stackdriver Trace, our support for logging is not strictly dependent on Trace. One use-case of what we're providing is to just link log messages by HTTP request.

PR #297, opened 2018-01-09
Constructor to add more headers for Stackdriver logging trace ID

ChengyuanZhao:
Please see #296 for details


spring-cloud-gcp-logging/src/main/java/org/springframework/cloud/gcp/logging/TraceIdLoggingWebMvcInterceptor.java
```diff
@@ -38,12 +38,41 @@
...
+	/**
+	 * Constructor that specifies the headers to check in what order, and if to only
+	 * expect a single header to be in the request.
+	 * @param headers the headers to check in order.
+	 * @param allowOnlyOneHeader true to allow multiple headers with trace IDs, false to
+	 * throw an exception if more than one header is found to have trace ID.
+	 */
+	public TraceIdLoggingWebMvcInterceptor(List<String> headers,
+			boolean allowOnlyOneHeader) {
+		if (headers == null) {
```
joaoandremartins: Nit: could use `Assert.notNull()`

ChengyuanZhao:
linking to issue that inspired this https://github.com/spring-cloud/spring-cloud-gcp/issues/296.

meltsufin:
@ChengyuanZhao Can you please provide a more descriptive title and fill in the description of the PR with the reference to the issue you are addressing? The convention is that the title and the PR description become the commit message when we squash & merge. The commit message should follow the [Git commit guidance](https://git-scm.com/book/en/v2/Distributed-Git-Contributing-to-a-Project#Commit-Guidelines).

spring-cloud-gcp-logging/src/main/java/org/springframework/cloud/gcp/logging/TraceIdLoggingWebMvcInterceptor.java
```diff
@@ -38,12 +38,41 @@
...
+	private final boolean allowOnlyOneHeader;
+
+	/**
+	 * Constructor that specifies the headers to check in what order, and if to only
+	 * expect a single header to be in the request.
+	 * @param headers the headers to check in order.
+	 * @param allowOnlyOneHeader true to allow multiple headers with trace IDs, false to
+	 * throw an exception if more than one header is found to have trace ID.
+	 */
+	public TraceIdLoggingWebMvcInterceptor(List<String> headers,
```
meltsufin: I'm not sure how autowire-friendly this constructor is.
It seems like it would be more appropriate in Spring to just turn headers into a bean and replace `DEFAULT_HEADERS` with a bean that is created by auto-configuration.

ChengyuanZhao: the `getLoggingWebMvcInterceptor` function in `StackdriverLoggingAutoConfiguration` class just returns an instance of this right? Couldn't we check if a setting like `spring.cloud.gcp.logging.onlyoneheader` and use that in the constructor?

meltsufin: I think the list of headers itself can be a bean, and in `StackdriverLoggingAutoConfiguration` we would create it with the two default headers, if one is not defined by the user. In that case, we probably don't even need the `onlyoneheader` option because, the user can just define a bean with one header to be used in the `LoggingWebMvcInterceptor`.

ChengyuanZhao: " the user can just define a bean with one header to be used "
But wouldn't this be a loss of functionality? You mentioned originally throwing an exception, which implies the user should know about an unexpected situation of having both headers when he only expected one. If the user only ever provides one header name, then we wouldn't be able to check for this.

In fact I kind of question the need for the ability to specify headers in spring settings files at all. Isn't the selling point of this project that GCP services specifically use the `x-cloud-content-trace` and `x-b3-trace` headers?

meltsufin: One might argue that "the selling point" of Spring is openness to customization and configuration. :) Auto-configuration in Spring Boot can provide the two default headers, but still keep the class open for configuration. If more than one search header has a value, we need to decide what to do -- fail or use first value silently. I don't have a strong opinion on that.
Let's discuss offline.

ChengyuanZhao: PTAL, I changed it up as discussed. I will make the springboot config changes in the following PR ,so right now even though all the classes and such are added it's still using the GCP header like before.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-logging/src/main/java/org/springframework/cloud/gcp/logging/autoconfig/StackdriverLoggingAutoConfiguration.java
```diff
@@ -38,6 +39,6 @@
 	@Bean
 	@ConditionalOnMissingBean
 	public TraceIdLoggingWebMvcInterceptor getLoggingWebMvcInterceptor() {
-		return new TraceIdLoggingWebMvcInterceptor();
+		return new TraceIdLoggingWebMvcInterceptor(new XCloudTraceIdFromRequestExtractor());
```
meltsufin: Since we're using dependency injection, we shouldn't be explicitly creating an instance of ` XCloudTraceIdFromRequestExtractor` like that. Instead, I would modify this method to:

```java
public TraceIdLoggingWebMvcInterceptor getLoggingWebMvcInterceptor(TraceIdFromRequestExtractor extractor)
```
and add:
```java
@Bean
@ConditionalOnMissingBean
public TraceIdFromRequestExtractor traceIdFromRequestExtractor() {
    return new XCloudTraceIdFromRequestExtractor();
}
```
This would enable the users to override just the `TraceIdFromRequestExtractor`, rather than the whole `TraceIdLoggingWebMvcInterceptor`.

spring-cloud-gcp-logging/src/test/java/org/springframework/cloud/gcp/logging/XCloudTraceIdFromRequestExtractorTest.java
```diff
@@ -0,0 +1,71 @@
...
+import static org.hamcrest.Matchers.is;
+import static org.hamcrest.Matchers.nullValue;
+import static org.junit.Assert.assertThat;
+
+/**
+ * @author Mike Eltsufin
+ * @author Chengyuan Zhao
+ */
+
+public class XCloudTraceIdFromRequestExtractorTest {
```
artembilan: Would be better to follow Spring Framework code style: https://github.com/spring-projects/spring-framework/wiki/Code-Style#tests

spring-cloud-gcp-logging/src/main/java/org/springframework/cloud/gcp/logging/PrioritizedMultipleHeaderTraceIdFromRequestExtractor.java
```diff
@@ -0,0 +1,62 @@
...
+
+import org.springframework.util.Assert;
+
+/**
+ * Checks HTTP request for multiple headers that might contain a trace ID, and provides
+ * the first one.
+ *
+ * @author Chengyuan Zhao
+ */
+public class PrioritizedMultipleHeaderTraceIdFromRequestExtractor
```
meltsufin: Do you want to add the option of failing if more than one header in the list is actually present in the request?
In that case, maybe just call it `MultiHeaderTraceIdFromRequestExtractor`.

artembilan: Well, in Spring we usually follow the naming style like `Composite`. For example `CompositeMessageConverter`.
The good JavaDoc should explain the behavior.

spring-cloud-gcp-logging/src/main/java/org/springframework/cloud/gcp/logging/PrioritizedMultipleHeaderTraceIdFromRequestExtractor.java
```diff
@@ -0,0 +1,62 @@
...
+		implements TraceIdFromRequestExtractor {
+
+	private List<TraceIdFromRequestExtractor> subExtractors;
```
artembilan: Well, usually we call them `target` or `delegate`, but they definitely are not `sub`. Because I don't see that they are doing some substruction from the result.

spring-cloud-gcp-logging/src/main/java/org/springframework/cloud/gcp/logging/PrioritizedMultipleHeaderTraceIdFromRequestExtractor.java
```diff
@@ -0,0 +1,62 @@
...
+
+		this.subExtractors = subExtractors;
+	}
+
+	public List<TraceIdFromRequestExtractor> getSubExtractors() {
+		return this.subExtractors;
+	}
+
+	public void setSubExtractors(List<TraceIdFromRequestExtractor> subExtractors) {
+		this.subExtractors = subExtractors;
```
artembilan: I would stay away from the setter to avoid mutation at runtime. More over you allow me here to set `null`.

spring-cloud-gcp-logging/src/main/java/org/springframework/cloud/gcp/logging/PrioritizedMultipleHeaderTraceIdFromRequestExtractor.java
```diff
@@ -0,0 +1,62 @@
...
+ *
+ * @author Chengyuan Zhao
+ */
+public class PrioritizedMultipleHeaderTraceIdFromRequestExtractor
+		implements TraceIdFromRequestExtractor {
+
+	private List<TraceIdFromRequestExtractor> subExtractors;
+
+	public PrioritizedMultipleHeaderTraceIdFromRequestExtractor(
+			List<TraceIdFromRequestExtractor> subExtractors) {
```
artembilan: Consider vararg instead.

spring-cloud-gcp-logging/src/main/java/org/springframework/cloud/gcp/logging/PrioritizedMultipleHeaderTraceIdFromRequestExtractor.java
```diff
@@ -0,0 +1,62 @@
...
+		Assert.notNull(subExtractors, "A valid list of extractors is required.");
+
+		this.subExtractors = subExtractors;
```
artembilan: Need to consider to do a copy to avoid effect after external container mutation.

spring-cloud-gcp-logging/src/main/java/org/springframework/cloud/gcp/logging/TraceIdFromRequestExtractor.java
```diff
@@ -0,0 +1,35 @@
...
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package org.springframework.cloud.gcp.logging;
+
+import javax.servlet.http.HttpServletRequest;
+
+/**
+ * A type which can extract a trace ID from an HTTP request.
```
artembilan: Type does nothing.
The strategy, really...

spring-cloud-gcp-logging/src/main/java/org/springframework/cloud/gcp/logging/TraceIdFromRequestExtractor.java
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * @author Chengyuan Zhao
+ */
+public interface TraceIdFromRequestExtractor {
```
artembilan: I'm really fine just with the `TraceIdExtractor` name...

spring-cloud-gcp-logging/src/main/java/org/springframework/cloud/gcp/logging/XCloudTraceIdFromRequestExtractor.java
```diff
@@ -0,0 +1,42 @@
...
+	public static final String X_CLOUD_TRACE_HEADER = "x-cloud-trace-context";
+
+	@Override
+	public String extractTraceIdFromRequest(HttpServletRequest req) {
+		String traceId = req.getHeader(X_CLOUD_TRACE_HEADER);
+
+		if (traceId != null) {
+			int slash = traceId.indexOf('/');
+			if (slash >= 0) {
+				traceId = traceId.substring(0, slash);
```
artembilan: Originally I was going to suggest just a request header customization for the `TraceIdLoggingWebMvcInterceptor` but this strategy approach is much better, indeed!

spring-cloud-gcp-logging/src/main/java/org/springframework/cloud/gcp/logging/PrioritizedMultipleHeaderTraceIdFromRequestExtractor.java
artembilan: I don't see reason in that option. We tell clearly in the JavaDocs: first matched wins.
So, that is your application responsibility to configure the order properly.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-logging/src/main/java/org/springframework/cloud/gcp/logging/autoconfig/StackdriverLoggingAutoConfiguration.java
artembilan: We may also provide a configuration option `trace.extractor=XCloud/Zipkin/XCloud-Zipkin/XCloud-Zipkin` to allow to reconfigure existing extractors just by the configuration property.

spring-cloud-gcp-logging/src/main/java/org/springframework/cloud/gcp/logging/TraceIdFromRequestExtractor.java
meltsufin: I was thinking the same actually.

ChengyuanZhao:
@meltsufin @artembilan I tried to address all the points I could from your previous round of comments. I added the changes for configuration, and I'm not familiar with the system so please take a look !

spring-cloud-gcp-logging/src/test/java/org/springframework/cloud/gcp/logging/CompositeHeaderTraceIdExtractorTest.java
```diff
@@ -0,0 +1,69 @@
...
+
+import static org.hamcrest.Matchers.is;
+import static org.hamcrest.Matchers.nullValue;
+import static org.junit.Assert.assertThat;
+
+/**
+ * @author Chengyuan Zhao
+ */
+
+public class CompositeHeaderTraceIdExtractorTest {
```
artembilan: Why is this `Test`, but not `Tests` ?

spring-cloud-gcp-logging/src/main/java/org/springframework/cloud/gcp/logging/LoggingWebMvcConfigurerSettings.java
```diff
@@ -0,0 +1,51 @@
...
+ *  limitations under the License.
+ */
+
+package org.springframework.cloud.gcp.logging;
+
+/**
+ * Holds settings for @{@link LoggingWebMvcConfigurer}
+ * @author Chengyuan Zhao
+ */
+public class LoggingWebMvcConfigurerSettings {
```
artembilan: That's a bit awkward solution. I would stay away of such an `enum` simulation.
We have `TraceIdExtractor` strategy and some out-of-the box implementations. 
We should just allow end-user to configure his/her own `TraceIdExtractor` bean, meanwhile we should be ready for its injection.

Moreover right now you fully don't allow to inject any custom `TraceIdExtractor`.
That is really much cleaner to just let to configure beans and provide some default when nothing is provided.

spring-cloud-gcp-samples/spring-cloud-gcp-logging-sample/src/main/resources/application.properties
```diff
@@ -1 +1,4 @@
 spring.cloud.gcp.logging.enabled=true
+spring.cloud.gcp.logging.headers.xCloudTrace=true
+spring.cloud.gcp.logging.headers.zipkinTrace=true
+spring.cloud.gcp.logging.headers.prioritizeXCloudTrace=true
```
artembilan: Well, my idea was like an `enum`-based single property and indeed exactly only in the auto-configuration module.
And even when we have that property for convenience during external configuration we should leave the room for custom `TraceIdExtractor` bean. Hence `@ConditionalOnMissingBean` before parsing such a property. Although there should be nothing to parse - just `switch` by the `enum` values.

spring-cloud-gcp-logging/src/test/java/org/springframework/cloud/gcp/logging/CompositeHeaderTraceIdExtractorTest.java
ChengyuanZhao: I missed that one, will fix.

spring-cloud-gcp-logging/src/main/java/org/springframework/cloud/gcp/logging/LoggingWebMvcConfigurerSettings.java
ChengyuanZhao: Yes I agree that sounds better. I will make changes to make it more like that.

spring-cloud-gcp-logging/src/main/java/org/springframework/cloud/gcp/logging/LoggingWebMvcConfigurerSettings.java
```diff
@@ -0,0 +1,35 @@
...
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package org.springframework.cloud.gcp.logging;
+
+/**
+ * Holds settings for @{@link LoggingWebMvcConfigurer}
```
meltsufin: Why `@` before the `{`?

spring-cloud-gcp-logging/src/main/java/org/springframework/cloud/gcp/logging/LoggingWebMvcConfigurerSettings.java
```diff
@@ -0,0 +1,35 @@
...
+	public TraceIdExtractorCombination getExtractorCombination() {
+		return this.extractorCombination;
+	}
+
+	public void setExtractorCombination(
+			TraceIdExtractorCombination extractorCombination) {
+		this.extractorCombination = extractorCombination;
+	}
+
+	private TraceIdExtractorCombination extractorCombination;
```
meltsufin: Why do we need this class at all to just hold a single instance variable?

ChengyuanZhao: It's a convenient place to put other settings in the future. Also, it allows the starter project to have a `ConfigurationSettings` annotated type inherit from this. We did this exact same thing in the storage project to avoid having the project depend on its starter.

spring-cloud-gcp-logging/src/main/java/org/springframework/cloud/gcp/logging/LoggingWebMvcConfigurerSettings.java
```diff
@@ -21,31 +21,15 @@
...
-	}
-
-	public boolean isPrioritizeXCloudTrace() {
-		return this.prioritizeXCloudTrace;
-	}
-
-	public void setPrioritizeXCloudTrace(boolean prioritizeXCloudTrace) {
-		this.prioritizeXCloudTrace = prioritizeXCloudTrace;
-	}
+	private TraceIdExtractorCombination extractorCombination;
```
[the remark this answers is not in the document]
ChengyuanZhao: Yea, I always thought that too but intellij made me doubt myself. It auto generates them and puts them at the bottom :/ 

Changes in this round have been made, PTAL.

spring-cloud-gcp-logging/src/main/java/org/springframework/cloud/gcp/logging/LoggingWebMvcConfigurerSettings.java
artembilan: But this is not `Settings`, this is just an `enum`. And again: the configuration properties is a feature of the Boot. Here in the raw level we should allow just configure beans and inject them. That is exactly the purpose of the Dependency Injection. The "Configuration by Convention" is a part of other effort - Spring Boot.

spring-cloud-gcp-logging/src/main/java/org/springframework/cloud/gcp/logging/TraceIdExtractorCombination.java
```diff
@@ -0,0 +1,27 @@
...
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package org.springframework.cloud.gcp.logging;
+
+/**
+ * Enum values represent the order and combination of trace ID extractors that can be
```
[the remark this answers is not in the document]
ChengyuanZhao: This enum is used in the LoggingWebMvcConfigurer, and if it's moved to the auto-config/starter project then this project would depend on its starter, which I'm trying to avoid. 

Renaming changes and javadoc changes made.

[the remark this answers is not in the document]
artembilan: Correct. The `LoggingWebMvcConfigurer` must just don't do that logic. It is really an auto-config responsibility to do configuration by convention. The `LoggingWebMvcConfigurer` should just rely on the injection `Extractor` bean or creating the default one otherwise. No any conventions by properties in the raw module. Just follow the classical Inversion of Control and that's all!

spring-cloud-gcp-logging/src/main/java/org/springframework/cloud/gcp/logging/CompositeTraceIdExtractor.java
```diff
@@ -0,0 +1,57 @@
...
+import java.util.List;
+
+import javax.servlet.http.HttpServletRequest;
+
+import com.google.common.collect.ImmutableList;
+
+import org.springframework.util.Assert;
+
+/**
+ * Checks HTTP request for multiple headers that might contain a trace ID, and provides
```
artembilan: That's not true. It just delegates to the provided set of `TraceIdExtractor`s and that is already their responsibility to check HTTP headers or get the value from the session or some other place.

Please, fix the JavaDoc according the current state.

spring-cloud-gcp-logging/src/main/java/org/springframework/cloud/gcp/logging/CompositeTraceIdExtractor.java
```diff
@@ -0,0 +1,57 @@
...
+ * the first one.
+ *
+ * @author Chengyuan Zhao
+ */
+public class CompositeTraceIdExtractor implements TraceIdExtractor {
+
+	private final List<TraceIdExtractor> extractors;
+
+	public CompositeTraceIdExtractor(List<TraceIdExtractor> extractors) {
```
artembilan: But we talked before about `TraceIdExtractor... extractors`. What happened with that approach?
Thanks

spring-cloud-gcp-logging/src/main/java/org/springframework/cloud/gcp/logging/LoggingWebMvcConfigurer.java
```diff
@@ -25,18 +27,22 @@
...
-	public LoggingWebMvcConfigurer(@Autowired(required = false) TraceIdLoggingWebMvcInterceptor interceptor) {
+	public LoggingWebMvcConfigurer(
+			@Autowired(required = false) TraceIdLoggingWebMvcInterceptor interceptor) {
 		if (interceptor != null) {
 			this.interceptor = interceptor;
 		}
 		else {
-			this.interceptor = new TraceIdLoggingWebMvcInterceptor();
+			this.interceptor = new TraceIdLoggingWebMvcInterceptor(
+					new CompositeTraceIdExtractor(ImmutableList.of(
```
artembilan: Somehow we try to avoid in Spring ecosystem Google Guava: too much breaking changes in the point releases and too much API copying names from other places what is misleading to the wrong imports during development. So, I would insist to come back to the vararg in the `CompositeTraceIdExtractor` and don't make end-users to import Guava code at all.

spring-cloud-gcp-logging/src/main/java/org/springframework/cloud/gcp/logging/LoggingWebMvcConfigurer.java
```diff
@@ -25,18 +27,22 @@
...
+							new XCloudTraceIdExtractor(), new ZipkinTraceIdExtractor())));
```
artembilan: This default order should be documented in the JavaDocs and Docs as well.

spring-cloud-gcp-logging/src/main/java/org/springframework/cloud/gcp/logging/TraceIdExtractor.java
```diff
@@ -0,0 +1,35 @@
...
+/**
+ * An extractor that can provide a trace ID from an HTTP request.
+ *
+ * @author Chengyuan Zhao
+ */
+public interface TraceIdExtractor {
+
+	/**
+	 * Extracts trace ID from the HTTP request by checking request headers.
+	 *
```
artembilan: Spring Code style JavaDocs rules violation: https://github.com/spring-projects/spring-framework/wiki/Code-Style#javadoc-formatting

spring-cloud-gcp-logging/src/main/java/org/springframework/cloud/gcp/logging/TraceIdLoggingWebMvcInterceptor.java
```diff
@@ -35,22 +34,33 @@
...
+	}
+
+	public TraceIdExtractor getTraceIdExtractor() {
+		return this.traceIdExtractor;
+	}
 
-	private static final List<String> TRACE_HEADERS = ImmutableList.of(X_CLOUD_TRACE,
-			X_B3_TRACE);
+	public void setTraceIdExtractor(
+			TraceIdExtractor traceIdExtractor) {
```
artembilan: I would stay away of the possible mutation at runtime. So, only ctor should be available.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-logging/src/main/java/org/springframework/cloud/gcp/logging/autoconfig/StackdriverLoggingAutoConfiguration.java
```diff
@@ -16,28 +16,71 @@
...
+@EnableConfigurationProperties({ StackdriverLoggingProperties.class })
 @ConditionalOnProperty(value = "spring.cloud.gcp.logging.enabled", matchIfMissing = true)
 @Import(LoggingWebMvcConfigurer.class)
 public class StackdriverLoggingAutoConfiguration {
 
 	@Bean
 	@ConditionalOnMissingBean
-	public TraceIdLoggingWebMvcInterceptor getLoggingWebMvcInterceptor() {
-		return new TraceIdLoggingWebMvcInterceptor();
+	public TraceIdLoggingWebMvcInterceptor getLoggingWebMvcInterceptor(
```
artembilan: It's not *getter*. It's a bean. So, would be better to name the factory method according the target bean names expectations: `loggingWebMvcInterceptor` is good one.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-logging/src/main/java/org/springframework/cloud/gcp/logging/autoconfig/StackdriverLoggingAutoConfiguration.java
```diff
@@ -16,28 +16,71 @@
...
+			StackdriverLoggingProperties loggingProperties) {
+		TraceIdExtractorType type = loggingProperties.getExtractorType();
+		TraceIdExtractorType checkedCombination = type == null
+				? TraceIdExtractorType.XCLOUD_ZIPKIN
+				: type;
+		CompositeTraceIdExtractor extractor;
+		switch (checkedCombination) {
+		case XCLOUD:
+			extractor = new CompositeTraceIdExtractor(
+					ImmutableList.of(new XCloudTraceIdExtractor()));
```
artembilan: Why it is wrapped to `CompositeTraceIdExtractor` if there is only one delegate? Redundant

spring-cloud-gcp-starters/spring-cloud-gcp-starter-logging/src/main/java/org/springframework/cloud/gcp/logging/autoconfig/StackdriverLoggingAutoConfiguration.java
```diff
@@ -16,28 +16,71 @@
...
+			break;
+		case ZIPKIN:
+			extractor = new CompositeTraceIdExtractor(
+					ImmutableList.of(new ZipkinTraceIdExtractor()));
```
artembilan: DITTO

spring-cloud-gcp-starters/spring-cloud-gcp-starter-logging/src/main/java/org/springframework/cloud/gcp/logging/autoconfig/StackdriverLoggingProperties.java
```diff
@@ -0,0 +1,35 @@
...
+ */
+
+package org.springframework.cloud.gcp.logging.autoconfig;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+/**
+ * @author Chengyuan Zhao
+ */
+@ConfigurationProperties("spring.cloud.gcp.logging.headers")
```
artembilan: I think this should not be restricted just to the `headers`. Plus see my comment on the composite class. We may extract trace id from some other place, not only headers.

spring-cloud-gcp-logging/src/main/java/org/springframework/cloud/gcp/logging/LoggingWebMvcConfigurer.java
```diff
@@ -25,18 +27,22 @@
...
 
-	public LoggingWebMvcConfigurer(@Autowired(required = false) TraceIdLoggingWebMvcInterceptor interceptor) {
+	public LoggingWebMvcConfigurer(
+			@Autowired(required = false) TraceIdLoggingWebMvcInterceptor interceptor) {
 		if (interceptor != null) {
 			this.interceptor = interceptor;
 		}
 		else {
-			this.interceptor = new TraceIdLoggingWebMvcInterceptor();
+			this.interceptor = new TraceIdLoggingWebMvcInterceptor(
```
joaoandremartins: How about having two different constructors?
One with and another one without the argument.

spring-cloud-gcp-logging/src/main/java/org/springframework/cloud/gcp/logging/TraceIdLoggingWebMvcInterceptor.java
joaoandremartins: Nit: no new line needed?

spring-cloud-gcp-logging/src/main/java/org/springframework/cloud/gcp/logging/CompositeTraceIdExtractor.java
```diff
@@ -0,0 +1,57 @@
...
+
+import org.springframework.util.Assert;
+
+/**
+ * Checks HTTP request for multiple headers that might contain a trace ID, and provides
+ * the first one.
+ *
+ * @author Chengyuan Zhao
+ */
+public class CompositeTraceIdExtractor implements TraceIdExtractor {
```
joaoandremartins: It seems to me like this whole trace ID extraction mechanism can be resumed to one or two private methods.
When will we need to add more headers that justify this extensibility overhead?
Even if we do, it should be as simple as adding a new string to a Set constant or so...

spring-cloud-gcp-starters/spring-cloud-gcp-starter-logging/src/main/java/org/springframework/cloud/gcp/logging/autoconfig/StackdriverLoggingAutoConfiguration.java
```diff
@@ -16,28 +16,71 @@
...
+	@ConditionalOnMissingBean
+	public TraceIdExtractor getTraceIdExtractor(
+			StackdriverLoggingProperties loggingProperties) {
+		TraceIdExtractorType type = loggingProperties.getExtractorType();
+		TraceIdExtractorType checkedCombination = type == null
+				? TraceIdExtractorType.XCLOUD_ZIPKIN
+				: type;
+		CompositeTraceIdExtractor extractor;
+		switch (checkedCombination) {
+		case XCLOUD:
```
joaoandremartins: I might be missing something, but it seems to me this is more complex than required.
Just feed a set of headers to `TraceIdLoggingWebMvcInterceptor`?

spring-cloud-gcp-starters/spring-cloud-gcp-starter-logging/src/main/java/org/springframework/cloud/gcp/logging/autoconfig/StackdriverLoggingAutoConfiguration.java
```diff
@@ -16,28 +16,71 @@
...
+			TraceIdExtractor extractor) {
+		return new TraceIdLoggingWebMvcInterceptor(extractor);
+	}
+
+	@Bean
+	@ConditionalOnMissingBean
+	public TraceIdExtractor getTraceIdExtractor(
+			StackdriverLoggingProperties loggingProperties) {
+		TraceIdExtractorType type = loggingProperties.getExtractorType();
+		TraceIdExtractorType checkedCombination = type == null
```
joaoandremartins: You can combine these two assignments into a single one.

ChengyuanZhao:
@artembilan @joaoandremartins 
We have a separate issue for the documentation 
https://github.com/spring-cloud/spring-cloud-gcp/issues/288

spring-cloud-gcp-logging/src/main/java/org/springframework/cloud/gcp/logging/LoggingWebMvcConfigurer.java
artembilan: Sounds good

spring-cloud-gcp-starters/spring-cloud-gcp-starter-logging/src/main/java/org/springframework/cloud/gcp/logging/autoconfig/StackdriverLoggingAutoConfiguration.java
artembilan: Well, as I pointed above, on the JavaDocs comment we might end up with the requirements to extract trace id from some other place, not just HTTP headers.

I might be to abstract, of course, but that feels like the purpose, @joaoandremartins .
Also pay attention to the difference in the implementations of the `TraceIdExtractor`. We can't do that custom logic having just headers configured.

joaoandremartins: If that's the end goal, sure.
Just for self reference, what other sources could we check?

spring-cloud-gcp-logging/src/main/java/org/springframework/cloud/gcp/logging/CompositeTraceIdExtractor.java
ChengyuanZhao: You had a comment just last round about using the List instead.... I'm gonna get whiplash

artembilan: Right: the internal property of the `CompositeTraceIdExtractor` must be list. The external parameter is vararg for the coding convenience. But again: we have do to a copy of it to avoid state mutation.

spring-cloud-gcp-logging/src/main/java/org/springframework/cloud/gcp/logging/CompositeTraceIdExtractor.java
ChengyuanZhao: This isn't actually outdated. What I have there is true, but just less detailed than your description. I'll make the change.

artembilan: ??? The `TraceIdExtractor` strategy doesn't state that an implementation should always look only into the headers. How about if I'm going to resolve it from the request params, or from body, or just from session attribute? Or just from static value from testing purpose. Can I? So, why them to restrict composite purpose just to headers by its JavaDocs or just mislead that way?

Sorry for the confuse but the feature is really confusing...

spring-cloud-gcp-logging/src/main/java/org/springframework/cloud/gcp/logging/CompositeTraceIdExtractor.java
ChengyuanZhao: It was very simple to start, and in fact the existing code is like that, but we had alot of offline discussion and we settled on this approach.

ChengyuanZhao:
closing and re-opening to re-trigger CI build

joaoandremartins:
`git commit --amend` should also do it

spring-cloud-gcp-starters/spring-cloud-gcp-starter-logging/src/main/java/org/springframework/cloud/gcp/logging/autoconfig/StackdriverLoggingAutoConfiguration.java
```diff
@@ -20,24 +20,64 @@
...
+		case ZIPKIN_XCLOUD:
+			extractor = new CompositeTraceIdExtractor(
+					new ZipkinTraceIdExtractor(), new XCloudTraceIdExtractor());
+			break;
+		case XCLOUD_ZIPKIN:
+			extractor = new CompositeTraceIdExtractor(
+					new XCloudTraceIdExtractor(), new ZipkinTraceIdExtractor());
+			break;
+		default:
+			extractor = new CompositeTraceIdExtractor(
```
meltsufin: Since the default case is impossible when all cases are covered by the `switch`, I would just throw an `IllegalArgumentException` here to catch mistakes early.

spring-cloud-gcp-logging/src/main/java/org/springframework/cloud/gcp/logging/TraceIdLoggingWebMvcInterceptor.java
```diff
@@ -35,22 +34,28 @@
...
 
 	@Override
 	public boolean preHandle(HttpServletRequest req,
 			HttpServletResponse resp, Object handler) throws Exception {
-		String traceId = extractTraceIdFromRequest(req);
+		String traceId = this.traceIdExtractor.extractTraceIdFromRequest(req);
 		if (traceId != null) {
-			TraceLoggingEnhancer.setCurrentTraceId(extractTraceIdFromRequest(req));
+			TraceLoggingEnhancer.setCurrentTraceId(
+					this.traceIdExtractor.extractTraceIdFromRequest(req));
```
artembilan: You have just extracted `traceId` local variable. A fresh `extractTraceIdFromRequest()` call may return a different value.
I'm sure you've been going to use that `traceId` local variable here as well

spring-cloud-gcp-starters/spring-cloud-gcp-starter-logging/src/main/java/org/springframework/cloud/gcp/logging/autoconfig/StackdriverLoggingAutoConfiguration.java
```diff
@@ -20,24 +20,64 @@
...
+			TraceIdExtractor extractor) {
+		return new TraceIdLoggingWebMvcInterceptor(extractor);
+	}
+
+	@Bean
+	@ConditionalOnMissingBean
+	public TraceIdExtractor traceIdExtractor(
+			StackdriverLoggingProperties loggingProperties) {
+		TraceIdExtractorType type = loggingProperties.getExtractorType();
+		TraceIdExtractorType checkedCombination = type == null
```
artembilan: You definitely don't need two variables. You can't override the previous value via ternary operator.
But nit-pick: it's not always a *combination*. Just *type* feels enough.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-logging/src/main/java/org/springframework/cloud/gcp/logging/autoconfig/StackdriverLoggingAutoConfiguration.java
ChengyuanZhao: It won't compile without the default statement since Java isn't a dependently-typed language and the compiler doesn't know that all cases are covered  (and also cuz I don't assign `extractor` outside of the switch statement)

spring-cloud-gcp-starters/spring-cloud-gcp-starter-logging/src/main/java/org/springframework/cloud/gcp/logging/autoconfig/StackdriverLoggingAutoConfiguration.java
ChengyuanZhao: Will change. I just didn't want to write `loggingProperties.getExtractorType()` twice.

artembilan: You don’t need call twice - just reuse the variable!

spring-cloud-gcp-starters/spring-cloud-gcp-starter-logging/src/main/java/org/springframework/cloud/gcp/logging/autoconfig/StackdriverLoggingAutoConfiguration.java
meltsufin: I'm not suggesting to remove the default case, I'm just suggesting throwing an exception there.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-logging/src/main/java/org/springframework/cloud/gcp/logging/autoconfig/TraceIdExtractorType.java
```diff
@@ -0,0 +1,32 @@
...
+ *  limitations under the License.
+ */
+
+package org.springframework.cloud.gcp.logging.autoconfig;
+
+/**
+ * Enum values represent the order and combination of predefined trace ID extractors that
+ * can be used.
+ *
+ * @see XCloudTraceIdExtractor
```
meltsufin: All of these class references are invalid because they're not fully qualified and are not imported in this file.
Use something like `@see org.springframework.cloud.gcp.logging.XCloudTraceIdExtractor`.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-logging/src/main/java/org/springframework/cloud/gcp/logging/autoconfig/StackdriverLoggingProperties.java
```diff
@@ -0,0 +1,35 @@
...
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+/**
+ * @author Chengyuan Zhao
+ */
+@ConfigurationProperties("spring.cloud.gcp.logging.traceId")
+public class StackdriverLoggingProperties {
+	private TraceIdExtractorType extractorType;
+
+	public TraceIdExtractorType getExtractorType() {
```
meltsufin: I think we wanted to call the property just "extractor" instead of "extractorType".

spring-cloud-gcp-starters/spring-cloud-gcp-starter-logging/src/main/java/org/springframework/cloud/gcp/logging/autoconfig/TraceIdExtractorType.java
```diff
@@ -0,0 +1,32 @@
...
+ * @see ZipkinTraceIdExtractor
+ * @see CompositeTraceIdExtractor
+ * @see TraceIdExtractor
+ *
+ * @author Chengyuan Zhao
+ */
+public enum TraceIdExtractorType {
+	XCLOUD, ZIPKIN, XCLOUD_ZIPKIN, ZIPKIN_XCLOUD
```
meltsufin: You can include more documentation for each of the enum values.
See: https://stackoverflow.com/questions/6549821/how-to-javadoc-a-classs-individual-enums.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-logging/src/main/java/org/springframework/cloud/gcp/logging/autoconfig/StackdriverLoggingProperties.java
ChengyuanZhao: I think extractorType is more appropriate though. In the settings class we have the getter and that getter's name makes more sense as `getExtractorType` vs `getExtractor`, which implies getting the actual extractor instead of the type.

spring-cloud-gcp-logging/src/main/java/org/springframework/cloud/gcp/logging/CompositeTraceIdExtractor.java
```diff
@@ -0,0 +1,57 @@
...
+
+import javax.servlet.http.HttpServletRequest;
+
+import com.google.common.collect.ImmutableList;
+
+import org.springframework.util.Assert;
+
+/**
+ * Delegates checking HTTP request to a list of provided {@link TraceIdExtractor}
+ * for headers that might contain a trace ID, and provides the first one that is found.
```
artembilan: Still: there is nothing about `headers` in this class.
Just imaging you are me and you don't know what is Stackdriver. So, I'm very confused why are we talking about headers here when `TraceIdExtractor` contract doesn't imply that and doesn't make me to stay only with the headers. 

I could fix during merge, but we don't have permissions to push to the `master` directly.
This way I would like to see the fix or good argument why do we stay just only with headers.

Thanks

spring-cloud-gcp-starters/spring-cloud-gcp-starter-logging/src/main/java/org/springframework/cloud/gcp/logging/autoconfig/TraceIdExtractorType.java
ChengyuanZhao: We have this other issue for all the documentation and I was going to make all those  changes then: #288 . I didn't want to make the readme changes in this PR because as you see folks are switching it up and I didn't want to rewrite the readme every round thinking it might be final.

spring-cloud-gcp-logging/src/main/java/org/springframework/cloud/gcp/logging/LoggingWebMvcConfigurer.java
```diff
@@ -25,18 +25,29 @@
...
 
 	private final TraceIdLoggingWebMvcInterceptor interceptor;
 
-	public LoggingWebMvcConfigurer(@Autowired(required = false) TraceIdLoggingWebMvcInterceptor interceptor) {
+	/**
+	 * Constructor that accepts an {@link TraceIdLoggingWebMvcInterceptor}. If the given
+	 * interceptor is null, then a default {@link CompositeTraceIdExtractor} is used that
+	 * checks using {@link XCloudTraceIdExtractor} followed by the {@link ZipkinTraceIdExtractor}.
+	 *
+	 * @param interceptor
```
artembilan: The param must be explained.

spring-cloud-gcp-logging/src/main/java/org/springframework/cloud/gcp/logging/TraceIdExtractor.java
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * An extractor that can provide a trace ID from an HTTP request.
+ *
+ * @author Chengyuan Zhao
+ */
+public interface TraceIdExtractor {
+
+	/**
+	 * Extract trace ID from the HTTP request by checking request headers.
```
artembilan: Again: why only headers?

Why nobody explains me why I can't do anything I'm able in the `extractTraceIdFromRequest()` implementation?
Thanks

spring-cloud-gcp-starters/spring-cloud-gcp-starter-logging/src/main/java/org/springframework/cloud/gcp/logging/autoconfig/StackdriverLoggingProperties.java
meltsufin: The properties bean is a special beast. It determines the configuration surface.
Basically we're choosing between:

`spring.cloud.gcp.logging.traceId.extractor=XCLOUD`
vs
`spring.cloud.gcp.logging.traceId.extractorType=XCLOUD`

If you both think `...extactorType` is better, I'm fine with that.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-logging/src/main/java/org/springframework/cloud/gcp/logging/autoconfig/TraceIdExtractorType.java
```diff
@@ -28,5 +28,22 @@
...
+	/**
+	 * Uses only {@link org.springframework.cloud.gcp.logging.XCloudTraceIdExtractor}
+	 */
+	XCLOUD,
+	/**
+	 * Uses only {@link org.springframework.cloud.gcp.logging.ZipkinTraceIdExtractor}
+	 */
+	ZIPKIN,
+	/**
+	 * Uses the XCloud extractor followed by the Zipkiung extractor if no trace ID was found
```
artembilan: Typo: s/Zipkiung/Zipkin

spring-cloud-gcp-starters/spring-cloud-gcp-starter-logging/src/main/java/org/springframework/cloud/gcp/logging/autoconfig/TraceIdExtractorType.java
meltsufin: #288 is specific to [Reference Documentation](https://docs.spring.io/spring-cloud-gcp/docs/1.0.0.BUILD-SNAPSHOT/reference/htmlsingle/) in which we have no mention of Logging yet at all.

I'm talking about README doc updates just for this particular change of introducing the extractor option.

spring-cloud-gcp-logging/src/main/java/org/springframework/cloud/gcp/logging/CompositeTraceIdExtractor.java
ChengyuanZhao: Ah I see. My apologies I get what you mean,  I will remove the mention of headers.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-logging/src/main/java/org/springframework/cloud/gcp/logging/autoconfig/StackdriverLoggingProperties.java
artembilan: Yeah... the `getExtractor()` is really confusing on the code level, but `extractor` is shorter for the configuration properties. The JavaDocs and return type does the trick. So, I'm really more up to the `extractor`, but can live with the `extactorType` since we have similar experience in Spring Boot already: https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/KafkaProperties.java#L511

spring-cloud-gcp-starters/spring-cloud-gcp-starter-logging/src/main/java/org/springframework/cloud/gcp/logging/autoconfig/TraceIdExtractorType.java
ChengyuanZhao: Ah, ok I will do that in a following PR.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-logging/src/main/java/org/springframework/cloud/gcp/logging/autoconfig/StackdriverLoggingAutoConfiguration.java
```diff
@@ -20,24 +20,62 @@
...
+	}
+
+	@Bean
+	@ConditionalOnMissingBean
+	public TraceIdExtractor traceIdExtractor(
+			StackdriverLoggingProperties loggingProperties) {
+		TraceIdExtractorType checkedType = loggingProperties.getExtractorType();
+		if (checkedType == null) {
+			checkedType = TraceIdExtractorType.XCLOUD_ZIPKIN;
+		}
```
joaoandremartins: How about

```
TraceIdExtractorType checkedType = loggingProperties.getExtractorType() != null
  ? loggingProperties.getExtractorType()
  : TraceIdExtractorType.XCLOUD_ZIPKIN;
```

`getExtractorType()` is O(1), so it should be OK to repeat.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-logging/src/main/java/org/springframework/cloud/gcp/logging/autoconfig/TraceIdExtractorType.java
ChengyuanZhao: https://github.com/spring-cloud/spring-cloud-gcp/issues/310

ChengyuanZhao: https://github.com/spring-cloud/spring-cloud-gcp/issues/310

spring-cloud-gcp-logging/src/main/java/org/springframework/cloud/gcp/logging/LoggingWebMvcConfigurer.java
joaoandremartins: *ping*

spring-cloud-gcp-starters/spring-cloud-gcp-starter-logging/src/main/java/org/springframework/cloud/gcp/logging/autoconfig/StackdriverLoggingProperties.java
meltsufin: Sounds like it's an established pattern in Spring Boot.
@ChengyuanZhao let's change to `extractor`, and then we can merge.

spring-cloud-gcp-logging/src/main/java/org/springframework/cloud/gcp/logging/LoggingWebMvcConfigurer.java
joaoandremartins:
```
public LoggingWebMvcConfigurer() {
	this.interceptor = new TraceIdLoggingWebMvcInterceptor(
					new CompositeTraceIdExtractor(ImmutableList.of(
							new XCloudTraceIdExtractor(), new ZipkinTraceIdExtractor())));
}

public LoggingWebMvcConfigurer(TraceIdLoggingWebMvcInterceptor interceptor) {
	this.interceptor = interceptor;
}
```

spring-cloud-gcp-starters/spring-cloud-gcp-starter-logging/src/main/java/org/springframework/cloud/gcp/logging/autoconfig/StackdriverLoggingAutoConfiguration.java
meltsufin: Let's not get carried away here. :)
We've already spent a considerable amount of engineering time on this.

spring-cloud-gcp-logging/src/main/java/org/springframework/cloud/gcp/logging/LoggingWebMvcConfigurer.java
artembilan: The `LoggingWebMvcConfigurer` is intended to be used from the `@Import` on the `@Configuration`, so those different ctors won't have any affects. More over I guess the default one will be selected by Spring Container for instantiation.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-logging/src/main/java/org/springframework/cloud/gcp/logging/autoconfig/StackdriverLoggingProperties.java
```diff
@@ -0,0 +1,35 @@
...
+ */
+
+package org.springframework.cloud.gcp.logging.autoconfig;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+/**
+ * @author Chengyuan Zhao
+ */
+@ConfigurationProperties("spring.cloud.gcp.logging.traceId")
```
artembilan: The last concern: why do we drill so deep?
Why can't we just live with the `spring.cloud.gcp.logging` namespace and use `spring.cloud.gcp.logging.extractor` for the property? I guess we are not going to have some other configuration properties for the `traceId`

ChengyuanZhao: I was thinking "extractor" is a pretty vague word and we might want to extract other things in the future (that aren't the trace ID).

spring-cloud-gcp-starters/spring-cloud-gcp-starter-logging/src/main/java/org/springframework/cloud/gcp/logging/autoconfig/StackdriverLoggingAutoConfiguration.java
joaoandremartins: Just copy and paste it.

artembilan: No, I prefer without ternary operator here. Much cleaner to read and there won't be a [Speculative execution](https://en.wikipedia.org/wiki/Speculative_execution) leading to [Spectre](https://spectreattack.com/) vulnerability. 😄 

Redundant joke here, of course, but no reason to argue any more. The code is good and doesn't repeat method call for nothing.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-logging/src/main/java/org/springframework/cloud/gcp/logging/autoconfig/StackdriverLoggingProperties.java
joaoandremartins: Good catch -- this should be just `spring.cloud.gcp.logging`.

If we want a `trace-id` element, we should add a nested property of type `TraceId`.

ChengyuanZhao: Oh wow yea good point! Is there a way to specify just spring.cloud.gcp.logging for this whole class but specify traceId.extractor for just that class member?

artembilan: Yes, there is. Just call the property as `traceIdExtractor`. That's the best I can suggest.

PR #301, opened 2018-01-11
add password

javaage:


meltsufin:
@javaage It looks like you pushed your commits to the wrong remote.
I would suggest you change the passwords on the instances you were testing.
Cheers!

PR #302, opened 2018-01-11
Moves to Spring Boot 2/Spring Framework 5

### LINKED ISSUE #201: Move to SF 5.0 and SB 2.0 when they're GA

joaoandremartins:
Fixes #201

### DIFF
--- a/spring-cloud-gcp-dependencies/pom.xml
+++ b/spring-cloud-gcp-dependencies/pom.xml
@@ -193,6 +193,15 @@
 				<artifactId>spring-integration-core</artifactId>
 				<version>${spring-integration.version}</version>
 			</dependency>
+
+			<!-- google-cloud-bom is setting the version of netty-tcnative-boringssl-static which is incompatible
+			with SB2.
+			Remove when https://github.com/GoogleCloudPlatform/google-cloud-java/issues/2781 is resolved. -->
+			<dependency>
+				<groupId>io.netty</groupId>
+				<artifactId>netty-tcnative-boringssl-static</artifactId>
+				<version>2.0.7.Final</version>
+			</dependency>
 		</dependencies>
 	</dependencyManagement>
 
--- a/spring-cloud-gcp-dependencies/pom.xml
+++ b/spring-cloud-gcp-dependencies/pom.xml
@@ -18,6 +18,10 @@
 	<properties>
 		<google-cloud-bom.version>0.32.0-alpha</google-cloud-bom.version>
 		<cloud-trace-java.version>0.5.0</cloud-trace-java.version>
+		<!-- Setting SI version until
+		 https://github.com/spring-cloud/spring-cloud-gcp/pull/302#discussion_r161083818
+		 is resolved. -->
+		<spring-integration.version>5.0.1.BUILD-SNAPSHOT</spring-integration.version>
 	</properties>
 
 	<dependencyManagement>
@@ -182,6 +186,12 @@
 				<groupId>com.google.cloud</groupId>
 				<artifactId>google-cloud-logging-logback</artifactId>
 				<version>0.32.0-alpha</version>
+			</dependency>
+
+			<dependency>
+				<groupId>org.springframework.integration</groupId>
+				<artifactId>spring-integration-core</artifactId>
+				<version>${spring-integration.version}</version>
 			</dependency>
 		</dependencies>
 	</dependencyManagement>
--- a/spring-cloud-gcp-dependencies/pom.xml
+++ b/spring-cloud-gcp-dependencies/pom.xml
@@ -5,7 +5,7 @@
 	<parent>
 		<groupId>org.springframework.cloud</groupId>
 		<artifactId>spring-cloud-dependencies-parent</artifactId>
-		<version>1.3.5.RELEASE</version>
+		<version>2.0.0.M6</version>
 		<relativePath/>
 	</parent>
 
--- a/spring-cloud-gcp-integration/src/main/java/org/springframework/integration/gcp/storage/GcsFileInfo.java
+++ b/spring-cloud-gcp-integration/src/main/java/org/springframework/integration/gcp/storage/GcsFileInfo.java
@@ -16,6 +16,8 @@
 
 package org.springframework.integration.gcp.storage;
 
+import java.util.Date;
+
 import com.google.cloud.storage.BlobInfo;
 
 import org.springframework.integration.file.remote.AbstractFileInfo;
@@ -66,5 +68,13 @@ public class GcsFileInfo extends AbstractFileInfo<BlobInfo> {
 	@Override
 	public BlobInfo getFileInfo() {
 		return this.gcsFile;
+	}
+
+	@Override
+	public String toString() {
+		return "FileInfo [isDirectory=" + isDirectory() + ", isLink=" + isLink()
+				+ ", Size=" + getSize() + ", ModifiedTime="
+				+ new Date(getModified()) + ", Filename=" + getFilename()
+				+ ", RemoteDirectory=" + getRemoteDirectory() + "]";
 	}
 }
--- a/spring-cloud-gcp-integration/src/main/java/org/springframework/integration/gcp/storage/GcsFileInfo.java
+++ b/spring-cloud-gcp-integration/src/main/java/org/springframework/integration/gcp/storage/GcsFileInfo.java
@@ -16,8 +16,6 @@
 
 package org.springframework.integration.gcp.storage;
 
-import java.util.Date;
-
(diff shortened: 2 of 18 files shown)


spring-cloud-gcp-integration/src/main/java/org/springframework/integration/gcp/storage/GcsFileInfo.java
```diff
@@ -60,7 +60,8 @@ public String getFilename() {
 
 	@Override
 	public String getPermissions() {
-		throw new UnsupportedOperationException("Use [BlobInfo.getAcl()] to obtain permissions.");
+		// throw new UnsupportedOperationException("Use [BlobInfo.getAcl()] to obtain permissions.");
+		return "xyz";
```
artembilan: What is this, please?

joaoandremartins:
This PR is only a proof of concept for now.

I wonder if we should make the move to Spring Boot 2 on `spring-cloud-gcp` M2?
If it's guaranteed SB2 goes GA soon, e.g., until end of February, then I say we do M2 already with SB2.

spring-cloud-gcp-integration/src/main/java/org/springframework/integration/gcp/storage/GcsFileInfo.java
joaoandremartins: @artembilan I meant to ask you about this.

`getPermissions()` is somehow being called by our tests with SF5. Do you know what's going on? I'm not sure what the best way forward here is.

artembilan:
??? You are going to do Milestone, so that is fully fine to have dependencies as Milestone as well.

More over we have a lot of Google dependencies in Milestone and you don't have similar concern there like you have just expressed for Spring Boot...

Anyway here is a calendar of all our planned releases: http://spring-calendar.cfapps.io/

spring-cloud-gcp-integration/src/main/java/org/springframework/integration/gcp/storage/GcsFileInfo.java
artembilan: OK. I see now. The super class has a `toString()` with the ` getPermissions()` call as well.
In SI-AWS I have this override:
```
public String toString() {
	return "FileInfo [isDirectory=" + isDirectory() + ", isLink=" + isLink()
			+ ", Size=" + getSize() + ", ModifiedTime="
			+ new Date(getModified()) + ", Filename=" + getFilename()
			+ ", RemoteDirectory=" + getRemoteDirectory() + "]";
}
```
And that's all!

joaoandremartins: @artembilan I tried that suggestion, but it isn't working for me.

The issue is that on the storage inbound streaming adapter test, the payload is being transformed to JSON and the `permissions` descriptor is invoking the `getPermissions()` file directly.

Need to dig on this a bit more. But apart from this issue, and the fact that payloads are now being sent on the Stream binder as `application/json` instead of `text/plain`, I think we're pretty much worry free with this migration.

joaoandremartins:
Thanks for the link. It sounds like we should ship M2 with SB2.
@meltsufin wdyt?

spring-cloud-gcp-integration/src/main/java/org/springframework/integration/gcp/storage/GcsFileInfo.java
joaoandremartins: For posterity, there is another issue with the `netty-tcnative-boringssl-static` dependency.

```Caused by: java.lang.NoClassDefFoundError: Could not initialize class io.netty.handler.ssl.OpenSslEngine```

`google-cloud-java` mandates version `2.0.6.Final` which is incompatible with netty version `4.1.17.Final`.
We might need to add `netty-tcnative-boringssl-static`'s version to our bom.

```
		<dependency>
			<groupId>io.netty</groupId>
			<artifactId>netty-tcnative-boringssl-static</artifactId>
			<version>2.0.7.Final</version>
		</dependency>
```

Comment (artembilan) [reply spring-cloud-gcp-integration/src/main/java/org/springframework/integration/gcp/storage/GcsFileInfo.java]: Ah! I think this one: https://github.com/spring-projects/spring-integration/pull/2308.
This is in the SI-5.0.1.BUILD-SNAPSHOT, so, we might need to add `<spring-integration.version>5.0.1.BUILD-SNAPSHOT</spring-integration.version>` to our pom for time being.

> google-cloud-java mandates version 2.0.6.Final

That's bad. That's not a modern world. It must be upgraded there for the actual version otherwise it really makes it useless to use `google-cloud-java` with Spring...

**Comment by meltsufin** [comment]
@joaoandremartins Yes, I think we can ship SB2 with M2.

@artembilan You mentioned that the Cloud Stream Binder APIs have significantly changed in SB2, but it doesn't appear so in this PR. Are we missing something?

**Comment by artembilan** [comment]
@meltsufin ,

Well, I don't know what you did in the Binder, so I can't tell what is wrong or not.
On the other hand I know how we lack in this project for tests, so the merged (unreviewed) Binder implementation might just don't follow the current SCSt 2.0 state.

I will be happy if everything is really OK, but I guess we'll see a feedback from the community eventually...

Also if you are happy with merging this Spring Boot/ Spring Cloud 2.0 upgrade, I'll definitely come back one day to review the Binder implementation if it is already based on SCSt 2.0. But first of all we need to arrange the problem with auto-configuration and config because that looks like Dave and Stephane have an different  visions for the project structure.

File: spring-cloud-gcp-integration/src/main/java/org/springframework/integration/gcp/storage/GcsFileInfo.java (hunk 1 above)
Comment (joaoandremartins) [reply spring-cloud-gcp-integration/src/main/java/org/springframework/integration/gcp/storage/GcsFileInfo.java]: Set the SI version to `5.0.1.BUILD-SNAPSHOT` and it solved our issue.

google-cloud-java are definitely setting the version for that dependency. https://github.com/GoogleCloudPlatform/google-cloud-java/blob/184c219c7a7d38bd8146e3a4d35f0972a1d24a10/google-cloud-bom/pom.xml#L136

I'll ask them if they can remove it from their BOM.
For the moment, it seems like we'll have to live with setting the version on our side?

Comment (joaoandremartins) [reply spring-cloud-gcp-integration/src/main/java/org/springframework/integration/gcp/storage/GcsFileInfo.java]: https://github.com/GoogleCloudPlatform/google-cloud-java/issues/2781

In the meantime, for the `netty-tcnative-boringssl-static` issue, sounds like we either set the version on our side, or ask users to set it on their side.

**Comment by joaoandremartins** [comment]
@artembilan PTAL

File: spring-cloud-gcp-dependencies/pom.xml [hunk 2]
```diff
@@ -183,6 +187,21 @@
...
+				<version>${spring-integration.version}</version>
+			</dependency>
+
+			<!-- google-cloud-bom is setting the version of netty-tcnative-boringssl-static which is incompatible
+			with SB2.
+			Remove when https://github.com/GoogleCloudPlatform/google-cloud-java/issues/2781 is resolved. -->
+			<dependency>
+				<groupId>io.netty</groupId>
+				<artifactId>netty-tcnative-boringssl-static</artifactId>
+				<version>2.0.7.Final</version>
```
garrettjonesgoogle: This is likely incompatible with grpc 1.7 & 1.8

https://github.com/grpc/grpc-java/blob/master/SECURITY.md#netty

joaoandremartins: I'm able to run my test app (which uses Tomcat) and publish/subscribe from Google Cloud Pub/Sub.

joaoandremartins:
@artembilan @meltsufin this should be ready for merge. PTAL

artembilan:
I'm OK for merging if you bring the `toString()` we have talked about to the `GcsFileInfo`:
```
public String toString() {
	return "FileInfo [isDirectory=" + isDirectory() + ", isLink=" + isLink()
			+ ", Size=" + getSize() + ", ModifiedTime="
			+ new Date(getModified()) + ", Filename=" + getFilename()
			+ ", RemoteDirectory=" + getRemoteDirectory() + "]";
}
```

This is indeed the fixed we have missed when developed this class. We shouldn't throw an exception from the `toString()` as it is right now by the super implementation.

Thanks

meltsufin:
No concerns from my side for the merge.

joaoandremartins:
@artembilan the reason I never added it is because the tests pass without it and I don't see anything else it's necessary for.

If you are sure it's needed, I'll add it.

artembilan:
> the tests pass without it

Well, maybe we just don't have enough tests then? 😄

Issue #299, opened 2018-01-10
Merge `gcp-config` module into `gcp-autoconfigure`


meltsufin:
Spin-off from the discussion in #276.

joaoandremartins:
Currently, the `spring-cloud-gcp-config` module exists so that we can enable Spring Cloud GCP Config conditionally, at the presence of `GoogleConfigPropertySourceLocator`.

The alternatives are:

1. Merge `spring-cloud-gcp-config` with `spring-cloud-gcp-autoconfigure`. We had it like this for a while, but it means that Spring Cloud GCP Config will always be enabled through `spring-cloud-gcp-starter-core`, which does not make sense since many `spring-cloud-gcp` users won't be interested in Spring Cloud GCP Config.

2. Merge `spring-cloud-gcp-config` with `spring-cloud-gcp-starter-config`. This would solve the issue of reducing the number of modules and enabling Spring Cloud GCP Config conditionally, at the cost of having code in the starter, which is an anti-pattern comparing to the Spring Boot standard.

We felt the separate `spring-cloud-gcp-config` module was the best compromise here. It follows the Spring Boot standard where the starter doesn't have any code and imports the generic `autoconfigure` module plus the service specific `config` module, even though the `config` by itself doesn't have any use.

If there are any other alternatives besides these that are better than what we currently have, feel free to suggest them. If not, it's probably safe to close this issue as working as intended.

artembilan:
I think the main reason to have this `spring-cloud-gcp-config` as a separate artifact just because there is no external dependency which would allow us to auto-configure feature by the presence in classpath. In this case we may go the `enable` configuration property or just stay with the situation as it is right now. At least there are many samples outside of Spring Boot per se, where we have separate artifacts which are available by the `starter` afterwards.

joaoandremartins:
> In this case we may go the enable configuration property

We thought about this, but wouldn't it end up as a circular dependency?

i.e., `GcpConfigAutoConfiguration` enables `GcpConfigProperties`, but we'd need `GcpConfigProperties` to enable/disable `GcpConfigAutoConfiguration` (e.g., by using the `enabled` property).

artembilan:
Not necessary that `enabled`  property must be a part of the `GcpConfigProperties`. 

[...]

PR #305, opened 2018-01-12
Merge autoconfig module into core and config

dsyer:
There seem to be rather a lot of modules in this project
and not all of them make sense to me as standalone features.
This is a start along the path of merging some of them 
together.

See #299


joaoandremartins:
@dsyer first of all, thanks so much for the PR and for the enthusiasm around this project! :)

This PR, specifically the removal of the `autoconfigure` module, feels like going on the opposite direction of #262, which was proposed in preparation for adding our starters to Spring Initializr.

Please note that we still intended to move the rest of the starters' code to the `autoconfigure` module, in order to be aligned with the rest of Spring Boot. Maybe that would make the motivation for the current module arrangement clearer?

dsyer:
> Please note that we still intended to move the rest of the starters' code to the autoconfigure module, in order to be aligned with the rest of Spring Boot.

That's kind of the opposite of the direction I was moving in. I tried it the other way round (merging everything in to autoconfig), but it worked better this way because there are definitely optional features in (for instance) the config module. I think having fewer modules is a really valid goal, and one module per optional feature (or set of features) seems achievable. If we get to that point and a global aggregation of all autoconfig still looks like a good idea, I think it will be easier from there.

The discussion in #262 got a bit bogged down in other things, but I didn't see any conclusive arguments there. My reading of that issue is "let's just close this one and move on". This PR is moving on from there.

joaoandremartins:
I just submitted #306 so we can appreciate how things look with a fleshed out `autoconfigure` module. I'm a big fan of that approach and it aligns well with all other Boot starters I've seen.

If there is a way of merging the `config` module with `autoconfigure` while being able to enable/disable the `config` bits conditionally, I'm up for it.
I don't think the same issue of "standalone modules not being a public API" applies to our other standalone modules. In that regard, we might have been unlucky by starting `autoconfigure` with the `config` module, since `config` is kind of a special flower.

dsyer:
> `spring-cloud-starter-gcp` could then bring `spring-cloud-gcp-core`

I haven't seen any value in the extra layer yet. You might as well just have one jar per feature.

snicoll:
I am not following. There is no extra layer here. It is building on the base feature, nothing more.

snicoll:
@dsyer and I have been discussing this as I've misunderstood his last comment. Now that `spring-cloud-gcp-core` is self contained and a signal to enable that feature, there is no need for a starter. Is there  a case where one would use GCP with only that feature (and nothing else)?

joaoandremartins:
> So spring-cloud-gcp-core is contributing something to the bootstrap context and spring-cloud-gcp-config is also contributing something to the bootstrap context.

Just want to clarify that `spring-cloud-gcp-core` doesn't contribute to anything Cloud Config.

> Now that spring-cloud-gcp-core is self contained and a signal to enable that feature, there is no need for a starter.

I think you meant `spring-cloud-gcp-config`?
I can see people using `spring-cloud-gcp-core` on its own, mainly if they're not using the Core starter.

Also, this PR currently moves towards removing the `autoconfigure` module and enabling Cloud Config through the `enabled` property rather than `spring-cloud-gcp-config` on the classpath which, from your comments, I understand we want the opposite of?

snicoll:
> Just want to clarify that spring-cloud-gcp-core doesn't contribute to anything Cloud Config.

That's not what I wrote. I wrote "bootstrap context". In the state of this PR, `spring-cloud-gcp-core` [does contribute to the bootstrap context](https://github.com/dsyer/spring-cloud-gcp/blob/16da5f946b6b40fff454bdd9d32c5afd02b8bb45/spring-cloud-gcp-core/src/main/resources/META-INF/spring.factories#L2)

> I think you meant spring-cloud-gcp-config?

Yes, sorry about that.

> Also, this PR currently moves towards removing the autoconfigure module and enabling Cloud Config through the enabled property rather than spring-cloud-gcp-config on the classpath 

IMO, None of this is correct. The `autoconfigure` module is removed because, in the state of this PR, there is nothing left in it. I haven't changed my mind about this and would like that all auto-configurations to be in a `spring-cloud-gcp-autoconfigure` module. As I've explained above, this PR helped me understand what the config module does (and that's not auto-configuration).

As for the second part of the sentence, the `enabled` flag is just a feature that we properly implement (use a condition rather than doing the check ourselves manually). It would [still be enabled by default](https://github.com/dsyer/spring-cloud-gcp/blob/16da5f946b6b40fff454bdd9d32c5afd02b8bb45/spring-cloud-gcp-config/src/main/java/org/springframework/cloud/gcp/config/GoogleConfigBootstrapConfiguration.java#L39) which indicates that the presence of the classpath is what triggers the feature.

Now, if  you decided to flip this property to false, then I'd be in favour of always bringing this feature with the base starter (in other words, the base starter could bring this module or, better yet, the feature could be added to the `autoconfigure` module as I don't see a compelling reason for having those 4 classes in an isolated module). @dsyer has expressed that enabling a feature by adding a jar is a good pattern. I agree but I think it also depends on the default behaviour of `GoogleConfigPropertySourceLocator`. If it can do something sensible by default (e.g. not crash the app, require a specific environment or a specific property), then I think we're good.

joaoandremartins:
If the `autoconfigure` module is for keeping, then I'd suggest we take a look at #306 before this one.
I think the merge becomes easier that way.

snicoll:
> in other words, the base starter could bring this module or, better yet, the feature could be added to the autoconfigure module as I don't see a compelling reason for having those 4 classes in an isolated module

Let  me take that back and propose that to be merged with `spring-cloud-gcp-core` (again, if you decide to go the route of making that flag `false` by default). It looks like a better  candidate than the `autoconfigure` module.

spring-cloud-gcp-core/src/main/java/org/springframework/cloud/gcp/core/GcpContextConfiguration.java
```diff
@@ -95,7 +89,6 @@ public GcpContextAutoConfiguration(GcpProperties gcpProperties) {
 	}
 
 	@Bean
-	@ConditionalOnMissingBean
```
joaoandremartins: Why removing this? It's actually come in useful on unit tests to provide mock beans to replace these ones.

snicoll: That's exactly why this PR should be handled first IMO. This is not auto-configuration so using `@ConditionalOnMissingBean` does not make any sense.

spring-cloud-gcp-config/src/main/java/org/springframework/cloud/gcp/config/GoogleConfigPropertySourceLocator.java
```diff
@@ -45,11 +45,11 @@
  *
  * @author Jisha Abubaker
  */
-public class GoogleConfigPropertySourceLocator implements PropertySourceLocator {
+class GoogleConfigPropertySourceLocator implements PropertySourceLocator {
 
 	private static final String RUNTIMECONFIG_API_ROOT = "https://runtimeconfig.googleapis.com/v1beta1/";
 
-	private static final String ALL_VARIABLES_PATH =
+	private static final String ALL_VARIABLES_PATH = //
```
joaoandremartins: Remove `//`

spring-cloud-gcp-core/pom.xml
```diff
@@ -19,12 +19,12 @@
 			<artifactId>google-cloud-core</artifactId>
 		</dependency>
 		<dependency>
-			<groupId>org.springframework</groupId>
-			<artifactId>spring-core</artifactId>
+			<groupId>org.springframework.boot</groupId>
+			<artifactId>spring-boot-starter</artifactId>
```
joaoandremartins: This doesn't feel right...
`spring-cloud-gcp-core` should not depend on Spring Boot. It should be usable standalone.

Also, since every module depends on `spring-cloud-gcp-core`, Spring Boot would automatically spread to the whole project?

spring-cloud-gcp-starters/spring-cloud-gcp-starter-config/README.adoc
```diff
@@ -1,141 +0,0 @@
-= Spring Cloud GCP Config Client Starter
```
joaoandremartins: This file can't simply disappear.
We should move it to `spring-cloud-gcp-config`.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/README.adoc
```diff
@@ -1,62 +0,0 @@
-= Google Cloud Core Spring Boot Starter
```
joaoandremartins: This file can't also just disappear.
If `spring-cloud-gcp-core` is becoming a starter on its own, then this should probably go there.

spring-cloud-gcp-core/pom.xml
snicoll: I  agree that the dependency on the starter is wrong but I still haven't seen a single reason why this module can't depend on Spring  Boot considering the fact the code is related to a usage of Spring Cloud and the latter requires the former.

spring-cloud-gcp-core/src/main/java/org/springframework/cloud/gcp/core/GcpContextConfiguration.java
joaoandremartins: Interesting, I think of this as auto-configuration, therefore I must be missing something.
It enables `GcpProperties` and uses them to configure the provided beans. Or, if no properties are provided, it resorts to coherent default values.

Why do you say it isn't auto-configuration?

dsyer: In `spring.factories` you have 

```
org.springframework.cloud.gcp.core.GcpContextConfiguration
```

so it's explicitly not autoconfiguration (it's spring cloud boot strap configuration)

joaoandremartins: I think you are focusing too much on the Cloud Config side of things. I'm almost tempted to remove the feature from `spring-cloud-gcp` altogether so we don't get too caught up (just kidding!!).

The main use of `GcpContext(Auto)Configuration` should _definitely_ be in `org.springframework.boot.autoconfigure.EnableAutoConfiguration`.
Using it in `org.springframework.cloud.bootstrap.BootstrapConfiguration` only makes sense when using Spring Cloud GCP Config. I think a minority of our users will be using it, though.

dsyer: It doesn't make sense in bootstrap configuration to use `@ConditionalOnMissingBean` (that was Stephane's point).

joaoandremartins:
This PR leaves us in an intriguing state, where we cut with the Spring Boot convention and our Core and Config starters would be called e.g., `spring-cloud-gcp-core` and not `spring-cloud-gcp-starter-core`, like the norm I've seen in all other SB projects I've seen.
Every module would also immediately become an SB starter, given that `core` depends on `spring-boot-starter`.

In this PR, I really like what you did with `GoogleConfigBootstrapConfiguration` and having `org.springframework.cloud.bootstrap.BootstrapConfiguration` use it instead of `GoogleConfigPropertySourceLocator`. I think this will let us turn off Config much more reliably.

But the whole module removal seems to be going a bit too far IMO. Especially if `autoconfigure` will be introduced back in.

It would be really nice to split the PR in two so that we can move forward more quickly. I see #306 as a crucial dependency for M2 whereas reducing the module count is an optimization that we can discuss with the `autoconfigure` module in place.

We're still sprinting towards M2 by Friday next week, so time is of the essence here! :)

spring-cloud-gcp-core/pom.xml
dsyer: I don't get it either. I don't see any reason to make this available to apps that don't use Spring Boot.

dsyer:
I really don't care too much about the specific bytes in this PR. If you want to merge #306 first then just do it and this one can be reworked on top of that. We need to come to some agreement about the general principle of using the correct idioms and lining the dependencies up sensibly. This PR was an example of the direction I think that should take. It can be used for discussion purposes and thrown away as far as I am concerned (probably better to do that sooner rather than later).

meltsufin:
I agree, we should complete work on #306 and then come back to this PR and the discussions that it has sparked.

For now, I just want to note that one of the design goals that we've been following from the start of this project is to support non-Boot users as well. I believe @markfisher, @artembilan, and @saturnism can confirm that. Even though the project is under the Spring Cloud umbrella, things like support for Pub/Sub Spring Integration adapters is useful for non-Boot users.

artembilan:
Right, @meltsufin , the mentioned Pub/Sub and Spring Integration Channel Adapters features can definitely live without Spring Boot. And that was exactly from where we started. But when we have moved forward to the Cloud Config that indeed can't live without Spring Boot foundation and doesn't make sense outside of Spring Cloud environment. That's why I always pointed to Spring Cloud AWS as a source to borrow "modularization" ideas from.

I'm so sorry that I'm not so good in Spring Boot and Spring Cloud design to notice this problem earlier, but I think we still have time until GA to refactor to the proper route. 

And big thanks to Stephane and Dave for jumping to our jungles and helping us to search the way out!

snicoll:
> things like support for Pub/Sub Spring Integration adapters is useful for non-Boot users.

Agreed but I'd argue that this module shouldn't be in this repository then.

dsyer:
Maybe. Or we could make it a special case (like we did with the AWS libraries). It needs a place to rest, and it might as well be here, until such time as the downstream projects actually want to pick up the maintenance (i.e. Spring Integration).

joaoandremartins:
@snicoll in https://github.com/spring-io/initializr/pull/539 you said this issue is a blocker for getting our starters on Initializr.
What do you want this PR to accomplish?

I mentioned in https://github.com/spring-cloud/spring-cloud-gcp/pull/305#issuecomment-358990703 that the work on `GoogleConfigBootstrapConfiguration` is pretty cool.
I'm guessing we no longer want to go on with removing the `autoconfigure` module.
I'm not sure what else you wanted to address here, so it would be great to hear from you! Maybe I could help you guys move things more quickly here.

We're looking to release M2 by the end of this week and then announcing it the next week.
I'd love to include Initializr in the announcement but if our starters aren't there yet, we'll have to move on with the release/announcement without Initializr.

snicoll:
> you said this issue is a blocker for getting our starters on Initializr.

What I meant by that is a resolution of this PR. That way we can start looking at the big picture once the structure of the base bits has been fixed.

> the work on GoogleConfigBootstrapConfiguration is pretty cool.

So let's merge that. Do you need us to rebase and redo a proposal based on the current state?

> I'm guessing we no longer want to go on with removing the autoconfigure module.

I've mentioned several times that the only reason the module was removed was because, in the context of this PR, the module was empty when we moved things elsewhere. I don't think you should focus on that too much.

joaoandremartins:
Yes, a rebase would be grand, thanks! I'd do it myself if the source were a branch in this repo. :)

Feel free to let us know about other improvements to the project structure so we can get to them sooner rather than later!

dsyer:
I don't think it's worth rebasing. Better to move on with #319 and friends, and then come back to this.

PR #306, opened 2018-01-12
Completes autoconfigure module

### LINKED ISSUE #262: Consider creating a spring-cloud-gcp-autoconfigure module

joaoandremartins:
Fixes #262


joaoandremartins:
This PR is WIP

The build is broken because the tests in `autoconfigure` expect the `config` autoconfiguration.

Need to look more into this, but the PR should give a good idea of what it'd be like for our starters to have no code and be split into code + autoconfig.

artembilan:
I can't review this sorry. The auto-config architecture is out of level of knowledge.
Would be better to ask @snicoll and @dsyer since they are Gods on the Spring Boot and Cloud Olympus and they have some vision already in other PRs and issues.

Thanks

P.S. But I will take a peak, indeed, just for improving myself 😄

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/storage/GcpStorageAutoConfigurationTests.java
```diff
@@ -43,10 +44,18 @@
 /**
  * @author Artem Bilan
  */
-@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT,
-		properties = "spring.cloud.gcp.storage.auto-create-files=false")
 @RunWith(SpringRunner.class)
-public class GoogleStorageAutoConfigurationTests {
+@SpringBootTest(
+		classes = {
```
joaoandremartins: @artembilan @snicoll if I don't add these classes, SQL auto configuration is kicking in.

However, when I add them, I'm getting a different error:

`org.springframework.context.ApplicationContextException: Unable to start EmbeddedWebApplicationContext due to missing EmbeddedServletContainerFactory bean.`

It seems that, since this is a web application, I should add another auto-configuration class to provide the missing `EmbeddedServletContainerFactory` bean. But which one?

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/storage/GcpStorageAutoConfigurationTests.java
```diff
@@ -43,10 +44,18 @@
...
+				GcpStorageAutoConfiguration.class,
+				GcpContextAutoConfiguration.class
+		},
+		webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT,
+		properties = { "spring.cloud.gcp.storage.auto-create-files=false",
+				"spring.cloud.gcp.config.enabled=false",
+				"spring.cloud.gcp.sql.enabled=false"
```
joaoandremartins: Having everything in a single `autoconfigure` module is proving challenging at testing time.
For example, because the Cloud Config stuff is being pulled into the classpath for every test, I have to disable config everywhere otherwise `GoogleConfigPropertySourceLocator` complains that "Config name must not be null".

Not specifying `GcpConfigAutoConfiguration` in `@SpringBootTest(classes)` doesn't help, as `GoogleConfigPropertySourceLocator` is still in the classpath and when the test application is booted, it tries to instantiate it.

Is there a better alternative to specifying `spring.cloud.gcp.config.enabled=false` in every test?

spring-cloud-gcp-autoconfigure/pom.xml
```diff
@@ -22,20 +22,112 @@
...
+
+        <!-- Global -->
         <dependency>
             <groupId>org.springframework.boot</groupId>
             <artifactId>spring-boot-configuration-processor</artifactId>
             <optional>true</optional>
         </dependency>
+        <dependency>
+            <groupId>commons-logging</groupId>
+            <artifactId>commons-logging</artifactId>
```
meltsufin: Where's this used?

spring-cloud-gcp-autoconfigure/pom.xml
```diff
@@ -22,20 +22,112 @@
...
+            <optional>true</optional>
+        </dependency>
+
+        <!-- Cloud SQL -->
+
+        <!-- HikariCP used as default DataSource -->
+        <dependency>
+            <groupId>com.zaxxer</groupId>
+            <artifactId>HikariCP</artifactId>
+            <scope>compile</scope>
```
meltsufin: Should it be `optional` like everything else?

meltsufin:
@joaoandremartins In `GcpCloudSqlAutoConfiguration` you have `@ConditionalOnClass(GcpProjectIdProvider.class)` which can't be right because `GcpProjectIdProvider` is now in the same `autoconfigure` module.

Also, why did you decide to merge the whole Cloud SQL support into `autoconfigure`?

spring-cloud-gcp-autoconfigure/pom.xml
joaoandremartins: `GcpContextAutoConfiguration` uses `Log`, for example.

spring-cloud-gcp-autoconfigure/pom.xml
joaoandremartins: Good catch.

joaoandremartins:
The whole Cloud SQL support has always been fully contained in the starter, I just moved it to `autoconfigure`.

spring-cloud-gcp-autoconfigure/pom.xml
meltsufin: I see, OK.

meltsufin:
Maybe it should be conditional on `com.google.cloud.sql.mysql.SocketFactory.class`?

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/storage/GcpStorageAutoConfigurationTests.java
joaoandremartins: FFR the class I was looking for is `EmbeddedServletContainerAutoConfiguration`, but after adding it there is another configuration class that provides a required `HttpMessageConverter` and is missing.
Since I don't want to manually add every required class, I did the opposite and filtered out the conflicting classes below.

snicoll: It is unusual to test auto-configuration using `@SpringBootTest`. You should create the  context yourself. Check the auto-configuration tests in Spring Boot for more details (check the 1.5.x branch as we have a test utility on master that you won't be able to use).

artembilan: Yeah... But this one looks like an integration test. It uses `TestRestTemplate` to verify the `GoogleStorageResourceObject`. So, this one needs a Web environment. 
We might reconsider this in the future though, because I agree with Stephane it indeed will become more and more cumbersome with the when we merge more and more feature to the `autoconfigure` module.

joaoandremartins: FWIW we're on SF5/SB2 since yesterday, so it should be possible to use that class you're mentioning.

joaoandremartins: @snicoll since this PR aims exclusively at migrating the rest of the starters to the `autoconfigure` model, I created #313 to improve the tests separate from this PR.

artembilan:
@joaoandremartins ,

tell me, please, what is the decision with the work Dave did in the #305.
Looks like you do here absolutely opposite what he suggests.

I think it would be better to arrange an architecture first of all and then start to develop.

Thanks

joaoandremartins:
The `autoconfigure` module should remain as per https://github.com/spring-cloud/spring-cloud-gcp/pull/305#issuecomment-358063503:

> The autoconfigure module is removed because, in the state of this PR, there is nothing left in it. I haven't changed my mind about this and would like that all auto-configurations to be in a spring-cloud-gcp-autoconfigure module.

In https://github.com/spring-cloud/spring-cloud-gcp/pull/305#issuecomment-358134467 I suggest that this PR be looked at before, so that the merge is easier (i.e., the `autoconfigure` isn't removed).

meltsufin:
@artembilan The idea is to complete this refactoring and then consider Dave's suggestions. This basically completes Stephane's original idea to align with how Spring Boot project is structured with starters having no code and all autoconfig classes in one module.

artembilan:
OK. Let it be. Works for me either way! I just worry that we don't waste time for work which will be weeded out...

Then I deffer the final decision to @dsyer and @snicoll !

My position is close to Dave's. Since we develop here Spring Cloud project there is no reason to fine-grain to very small peaces since it is unlikely someone is going to use our project outside of Spring Cloud environment. See the same [Spring Cloud Sleuth](https://github.com/spring-cloud/spring-cloud-sleuth) for example.

But at the same time there is a [Spring Cloud AWS](https://github.com/spring-cloud/spring-cloud-aws) where we definitely have raw modules and for Spring Boot...

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
```diff
@@ -53,6 +54,8 @@
  */
 @Configuration
 @AutoConfigureAfter(GcpContextAutoConfiguration.class)
+@ConditionalOnProperty(value = "spring.cloud.gcp.trace.enabled", matchIfMissing = true)
```
joaoandremartins: Does this check if `spring.cloud.gcp.trace.enabled` is `true`? Or do we need a `havingValue` attribute?
From `@ConditionalOnProperty`, it isn't clear.

snicoll:
> so that the merge is easier (i.e., the autoconfigure isn't removed).

I don't understand this argument. The PR removes the module because there is no longer any code on it. I am sure @dsyer can revert that part and left the module with no code in it. 

I must share that having the final target in several PRs is very confusing to me. As you've noticed that the config property source locator kicks in and it is annoying, it has a direct effect on that PR so addressing that at the right place would be so much easier IMO

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
snicoll: image

I find the javadoc pretty explicit. What wasn't clear?

joaoandremartins: I meant 

```
	/**
	 * The string representation of the expected value for the properties. If not
	 * specified, the property must <strong>not</strong> be equals to {@code false}.
	 * @return the expected value
	 */
	String havingValue() default "";
```

The Javadoc you pasted is pretty clear. It seems things are OK as they are. Thanks!

snicoll: it's the class javadoc. If you think the `havingValue` javadoc could be polished, please feel free to open an issue with a suggestion.

PR #308, opened 2018-01-16
Adds usage tracking to the Google Cloud Client libraries

joaoandremartins:
We're now adding to the User-Agent header on every API request to Google
a token to mark the requests as coming from a Spring Integration. This
will allow us to measure the amount of users our integrations have.


spring-cloud-gcp-dependencies/pom.xml
```diff
@@ -16,7 +16,7 @@
 	<packaging>pom</packaging>
 
 	<properties>
-		<google-cloud-bom.version>0.32.0-alpha</google-cloud-bom.version>
+		<google-cloud-bom.version>0.32.1-alpha-SNAPSHOT</google-cloud-bom.version>
```
joaoandremartins: This is pending a new google-cloud-java release.
It will have to be change once the release is made.
This PR can't be merged until the new release is available.

joaoandremartins:
@vam-google can you please make sure we're adding the headers to the clients properly?

spring-cloud-gcp-dependencies/pom.xml
artembilan: Therefore this PR can't be review until the finalized target API.
Or we really will have a spare time...

joaoandremartins: I don't expect the API to change at all, but you can wait until we have an actual release if you prefer!

meltsufin:
For those looking at this PR, see [related PR in google-cloud-java](https://github.com/GoogleCloudPlatform/google-cloud-java/pull/2690) for some documentation.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/DefaultSubscriberFactory.java
```diff
@@ -102,7 +102,7 @@ public SubscriberStub createSubscriberStub(RetrySettings retrySettings) {
 			SubscriptionAdminSettings.Builder subscriptionAdminSettingsBuilder = SubscriptionAdminSettings.newBuilder()
 					.setExecutorProvider(this.executorProvider)
 					.setCredentialsProvider(this.credentialsProvider)
-					.setTransportChannelProvider(this.channelProvider);
```
meltsufin: Has `setTransportChannelProvider` been removed??

spring-cloud-gcp-core/src/main/java/org/springframework/cloud/gcp/core/UsageTrackingHeaderProvider.java
```diff
@@ -0,0 +1,51 @@
...
+ *
+ * @author João André Martins
+ */
+public class UsageTrackingHeaderProvider implements HeaderProvider {
+
+	private final String springLibrary;
+
+	private final String springLibraryVersion;
+
+	public UsageTrackingHeaderProvider(String springLibrary, String springLibraryVersion) {
```
meltsufin: I think you can replace `String springLibraryVersion` with just a `Class clazz`, and then use `this.clazz.getPackage().getImplementationVersion()` in `getHeaders`. This would make the users of this class a bit less verbose.

Furthermore, you can remove `springLibrary` argument altogether, and replace it with:
```java
		String[] pkg = getClass().getPackage().getName().split("\\.");
		String springLibrary = "spring-cloud-gcp-" + pkg[pkg.length-1];
```
You would then just end up with a constructor that looks like this:
`UsageTrackingHeaderProvider(Class clazz)`

Convention over configuration FTW!

joaoandremartins: 1) I want to be able to make it possible for the library version to be different from the project version. It might always be the same, but there are times we might need it to be different.

2) `getClass()` will always return `UsageTrackingHeaderProvider`, therefore `springLibrary` would always be `spring-cloud-gcp-core` IIUC?

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/DefaultSubscriberFactory.java
joaoandremartins: The only thing we were using it for was to set the similar header the way we were doing before (which was useless).
Not to worry because #280 is bringing all that stuff back.

spring-cloud-gcp-core/src/main/java/org/springframework/cloud/gcp/core/UsageTrackingHeaderProvider.java
meltsufin: 1. `clazz.getPackage().getImplementationVersion()` would return the version of the module in which the passed in class resides, right? So, it can be different from the top-level project version. Am I missing something? Currently, you're passing the same thing everywhere. If this changes some day, we can add another constructor then.

2. I meant to say `clazz.getClass()...` which would return the package name of the passed in class that doesn't have to be `UseTrackingHeaderProvider`.

joaoandremartins: Aah, so you're suggesting passing in a class and then getting the data we need from it!
Yeah sure.

joaoandremartins: One problem with this approach is that the classes calling the constructor are all in the `autoconfigure` package, so we would need to pass in another class in the feature's main module.
But some features might not even have their own modules (e.g., there were talks of getting rid of the config module).

I'm more inclined towards abandoning this idea, but wdyt?

meltsufin: They don't have to be in their own modules, just in separate packages.
You're using `this.getClass()` everywhere currently. In what instance do you need to pass in a different class? In any case, it shouldn't pose an issue.

joaoandremartins: @meltsufin I was passing classes in the `autoconfigure` module.
In the end, `springLibrary` ended up always being `spring-cloud-gcp-autoconfigure` which isn't helpful.

I don't think that suggestion is very feasible. I prefer being able to pass any value I want to report, instead of relying on a convention where the classes passed must be in a package whose last section is the service (e.g., `storage`, etc.).

spring-cloud-gcp-starters/spring-cloud-gcp-starter-storage/src/main/java/org/springframework/cloud/gcp/storage/autoconfig/GcpStorageAutoConfiguration.java
```diff
@@ -62,8 +62,7 @@ public static Storage storage(CredentialsProvider credentialsProvider,
 										.getLocation().getInputStream())
 								.createScoped(gcpStorageProperties.getCredentials().getScopes())
 						: credentialsProvider.getCredentials())
-				.setHeaderProvider(new UsageTrackingHeaderProvider("spring-cloud-gcp-storage",
-						StorageAutoConfiguration.class.getPackage().getImplementationVersion()))
+				.setHeaderProvider(new UsageTrackingHeaderProvider(GcpStorageAutoConfiguration.class))
```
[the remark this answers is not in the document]
joaoandremartins: Because this method is `static`, for some reason.

spring-cloud-gcp-core/src/main/java/org/springframework/cloud/gcp/core/UsageTrackingHeaderProvider.java
```diff
@@ -0,0 +1,56 @@
...
+	public UsageTrackingHeaderProvider(Class clazz) {
+		this.clazz = clazz;
+	}
+
+	/**
+	 * Returns the "User-Agent" header whose value should be added to the google-cloud-java REST API calls.
+	 * e.g., {@code User-Agent: Spring/1.0.0.RELEASE spring-cloud-gcp-pubsub/1.0.0.RELEASE}.
+	 */
+	@Override
+	public Map<String, String> getHeaders() {
```
meltsufin: This method can use a simple unit test.

PR #319, opened 2018-01-22
Remove redundant starters

dsyer:
The "core" and "config" starters are really just passthrus for the
jars underneath them, and as such they are redundant. This change just
removes the empty layer.

This is the first of a series of potential changes to replace #305.


spring-cloud-gcp-starters/spring-cloud-gcp-starter-core/README.adoc
```diff
@@ -1,62 +0,0 @@
-= Google Cloud Core Spring Boot Starter
```
meltsufin: Same thing here. We should move the docs somewhere else.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql/README.adoc
```diff
@@ -1,141 +0,0 @@
-= Google Cloud SQL Spring Boot Starter
```
meltsufin: Ditto on not losing the docs.

spring-cloud-gcp-samples/spring-cloud-gcp-data-jpa-sample/pom.xml
```diff
@@ -20,7 +20,15 @@
 
 		<dependency>
 			<groupId>org.springframework.cloud</groupId>
-			<artifactId>spring-cloud-gcp-starter-sql</artifactId>
+			<artifactId>spring-cloud-gcp-autoconfigure</artifactId>
+		</dependency>
+		<dependency>
+			<groupId>com.google.cloud.sql</groupId>
```
joaoandremartins: As I mentioned in https://github.com/spring-cloud/spring-cloud-gcp/issues/321, I do think there's value in having a starter for bringing in the right dependencies (along with their correct versions). Wouldn't you call that an opinionated starter?

spring-cloud-gcp-starters/pom.xml
```diff
@@ -21,10 +21,7 @@
 	</properties>
 	<modules>
 		<module>spring-cloud-gcp-starter-pubsub</module>
-		<module>spring-cloud-gcp-starter-config</module>
-		<module>spring-cloud-gcp-starter-core</module>
 		<module>spring-cloud-gcp-starter-storage</module>
-		<module>spring-cloud-gcp-starter-sql</module>
```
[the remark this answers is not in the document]
joaoandremartins: Removing this would break our integration because `DefaultCloudSqlJdbcInfoProvider` depends on `GcpCloudSqlProperties.getDatabaseType()` to return the correct driver class name.

Spring Boot users might find it odd to have to add the MySQL libraries to the classpath _and_ having to specify which kind of SQL database vendor they're using.

meltsufin:
Some of these readme's have content that we don't have elsewhere. So we may want to merge the content into `docs`.

meltsufin:
@dsyer Can we close this without merging? It has significantly diverged from master now and the things it touches are being addressed in other PRs.

dsyer:
I think so. It was all structural anyway, no code. So easy to redo.

PR #323, opened 2018-01-23
Merge integration module into pubsub and storage

dsyer:
This one is a bit more speculative, but I think I like it. The net 
effect for the user is that they only have to include one
spring-cloud-gcp jar (instead of 2) to get the same feature, as well
as spring-integration core jars.

Apply on top of #319 and #324


garyrussell:
This is fine with me.

joaoandremartins:
Closed by #360

PR #324, opened 2018-01-23
Merge stream binder into a single module

dsyer:
The spring cloud stream dependencies can be collapsed into a single jar
(the binder). There are no tests or samples for this as far as I can tell.

Merge on top of #319


joaoandremartins:
Closed by #359

PR #334, opened 2018-01-24
Add test config for logger

dsyer:


spring-cloud-gcp-pubsub/src/test/resources/logback.xml
```diff
@@ -0,0 +1,20 @@
...
+
+	<logger name="io.netty.handler" level="debug"/>
+
+	<logger name="org.springframework.cloud" level="info"/>
+
+	<root level="info">
+		<appender-ref ref="STDOUT" />
+	</root>
+
+</configuration>
```
artembilan: I think we need to add ` <module name="NewlineAtEndOfFile"/>` to our `checkstyle.xml` config: http://checkstyle.sourceforge.net/config_misc.html#NewlineAtEndOfFile

meltsufin: created separate issue for this: https://github.com/spring-cloud/spring-cloud-gcp/issues/338

artembilan:
It is really good to have logging config so any subsequent issues in the future will be easier to investigate.

We have similar practice in many our other projects. See for example Spring Integration for `log4j2-test.xml`.

PR #317, opened 2018-01-19
Adds Javadoc to @ConfigurationProperties' fields

joaoandremartins:
Fixes #263

PR #335, opened 2018-01-24
Slightly improve Javadoc

joaoandremartins:


spring-cloud-gcp-core/src/main/java/org/springframework/cloud/gcp/core/Credentials.java
```diff
@@ -42,8 +42,6 @@
 	 *
 	 * <p>Since this is a {@link Resource}, the private key can be in a multitude of locations, such as a local file
```
artembilan: I think it still doesn't follow the Configuration Properties documentation requirements: https://github.com/spring-cloud/spring-cloud-gcp/pull/317#discussion_r163013489

PR #341, opened 2018-01-24
Consolidates spring-cloud-gcp(-starter)-config with autoconfigure

joaoandremartins:
Removes two modules from the project. Enabling the Config feature will
now be done using a property.


joaoandremartins:
@snicoll @dsyer FYI

Also, I noticed that the `/refresh` endpoint doesn't seem to be active on Spring Boot Actuator 2. There is no mention to that endpoint from the SB refdoc.
Is there another endpoint for that functionality or was it removed completely?

dsyer:
Re /refresh endpoint, did you expose it ( `management.endpoints.web.expose=*` for instance)?

spring-cloud-gcp-autoconfigure/pom.xml
```diff
@@ -126,5 +119,11 @@
             <optional>true</optional>
         </dependency>
 
+        <!-- Config -->
+        <dependency>
+            <groupId>org.springframework.cloud</groupId>
+            <artifactId>spring-cloud-context</artifactId>
```
joaoandremartins: I set this dependency to optional=false, so users don't have to.
Rationale is that I see `spring-cloud-context` as something very broad, and the link to Spring Cloud GCP Config is not straightforward, but I might be wrong.
Thoughts?

dsyer: Makes sense to me as long as nothing else (except config) depends on it. But I think that might not quite be true (sleuth and stream depend on it as well). That's why I think we need the `@EnableGcpConfiguration` annotation. But we can continue that discussion elsewhere.

joaoandremartins: We've already got `@ConditionalOnProperty` for Config, that should cover it for now?
We never had any `@ConditionalOnClass` for a `spring-cloud-context` class, anyway.

spring-cloud-gcp-samples/spring-cloud-gcp-config-sample/src/main/resources/bootstrap.properties
```diff
@@ -11,7 +8,7 @@ spring.cloud.gcp.config.name=myapp
 spring.cloud.gcp.config.profile=prod
 
 # Config server API timeout in milliseconds, default : 60000 (1 minute)
```
ChengyuanZhao: This comment here says 60 seconds, but the setting below is 120 seconds it seems

joaoandremartins: Right, the point is to override the default config with the setting here.

joaoandremartins:
@meltsufin ptal

spring-cloud-gcp-starters/spring-cloud-gcp-starter-config/README.adoc
```diff
@@ -1,4 +1,4 @@
-= Spring Cloud GCP Config Client Starter
+== Spring Cloud GCP Config Client Starter
```
artembilan: Why this README file is still in the `starter` module?
Nobody is going to read it since everybody knows that any Spring Boot Starter is just empty jar with conventional pom.

Why just don't go the way to have docs only in `docs` ?

snicoll: +1 to that. I was asking me the same question (about the duplicated doc, that is).

joaoandremartins:
Should be ready for a final pass

PR #344, opened 2018-01-25
Moves the Spring Integration dependency from BOM to root pom

### LINKED ISSUE #201: Move to SF 5.0 and SB 2.0 when they're GA

joaoandremartins:
Fixes #201

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -45,6 +45,10 @@
 		<spring-cloud-context.version>2.0.0.M5</spring-cloud-context.version>
 		<spring-cloud-sleuth.version>2.0.0.M5</spring-cloud-sleuth.version>
 		<spring-cloud-stream.version>2.0.0.M3</spring-cloud-stream.version>
+		<!-- Setting SI version until
+		 https://github.com/spring-cloud/spring-cloud-gcp/pull/302#discussion_r161083818
+		 is resolved. -->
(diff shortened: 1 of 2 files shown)


snicoll:
Can we reopen please? Rather than selectively provide a version for Spring Integration,  you should just rely on the one that Spring Boot provides. Is there any problem with that?

artembilan:
H-m. Sure! If we go Spring Boot `BUILD-SNAPSHOT`, we are good to get Spring Integration `5.0.1.BUILD-SNAPSHOT` for free: https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-dependencies/pom.xml#L150

snicoll:
@artembilan it's more than that. If you're compliant with a certain Spring Boot version and you are overriding spring integration here, there is no guarantee that it will work in practice (since this override  is, as it should, local to the project)

artembilan:
Ah! Indeed! Now I remember that overriding versions for compilation our project doesn't guarantee that the target user of our lib will get the same version. He/she must override the version in their own POM/Gradle.

joaoandremartins:
I think the next SB2 RC is coming out next Wednesday, possibly with the latest SI version.
Can that change wait until then?

snicoll:
If you need the latest you should run with our snapshot.

PR #353, opened 2018-01-26
Added missing dep for DataSource

### LINKED ISSUE #337: Cloud SQL sample is broken
Running the SQL code sample results in the following error:

```
Caused by: org.springframework.beans.factory.UnsatisfiedDependencyException: Error creating bean with name 'webController' defined in file [/usr/local/google/home/joaomartins/workspaces/spring-cloud-gcp/spring-cloud-gcp-samples/spring-cloud-gcp-sql-sample/target/classes/com/example/WebController.class]: Unsatisfied dependency expressed through constructor parameter 0; nested exception is org.springframework.beans.factory.NoSuchBeanDefinitionException: No qualifying bean of type 'org.springframework.jdbc.core.JdbcTemplate' available: expected at least 1 bean which qualifies as autowire candidate. Dependency annotations: {}
	at org.springframework.beans.factory.support.ConstructorResolver.createArgumentArray(ConstructorResolver.java:729)
```
If all references to `JdbcTemplate` are removed from the code, the auto configuration kicks in. However, when the user code autowires a `JdbcTemplate` object, Spring Boot tries to instantiate it before `GcpCloudSqlAutoConfiguration` runs. Since at that time, there is no `DataSourceProperties` bean in the context, there also isn't any `DataSource` and therefore, no `JdbcTemplate`.

ChengyuanZhao:
Fixes #337 
There was a missing dependency for DataSource.

### DIFF
--- a/spring-cloud-gcp-samples/spring-cloud-gcp-sql-sample/pom.xml
+++ b/spring-cloud-gcp-samples/spring-cloud-gcp-sql-sample/pom.xml
@@ -25,5 +25,10 @@
 			<groupId>org.springframework</groupId>
 			<artifactId>spring-jdbc</artifactId>
 		</dependency>
+		<dependency>
+			<groupId>com.zaxxer</groupId>
+			<artifactId>HikariCP</artifactId>
+			<scope>compile</scope>
+		</dependency>
 	</dependencies>
 </project>


joaoandremartins:
Good job!

PR #355, opened 2018-01-26
Polish logging docs

### LINKED ISSUE #315: Fix issues with logging code sample
- [ ] Something on `logging` is not reading the project ID from the context. You can reproduce that by clearing Cloud SDK's project ID (and credentials for that matter) with `gcloud config unset core/project`. We should either make it so or, if impossible, document in the README/refdoc that the project ID from Cloud SDK/some envvar is used instead.
+
```
ERROR in ch.qos.logback.core.joran.spi.Interpreter@15:14 - RuntimeException in Action for tag [appender] java.lang.IllegalArgumentException: A project ID is required for this service but could not be determined from the builder or the environment.  Please set a project ID using the builder.
	at org.springframework.boot.logging.logback.LogbackLoggingSystem.loadConfiguration(LogbackLoggingSystem.java:163)
	at org.springframework.boot.logging.AbstractLoggingSystem.initializeWithConventions(AbstractLoggingSystem.java:81)
```

- [x] With the project ID in, running the sample gets stuck loading the Boot app. It never finishes loading. Removing the `logback-spring.xml` makes the app start. **NON-ISSUE**

- [ ] Documentation reads `Browse to "localhost:8080/log".` but this steps seems useless without a trace ID. Maybe remove this bit and just ask users to make the curl call with the trace ID?

- [ ] Like the project ID, credentials are also not being read from the `spring.cloud.gcp.credentials.location` property.

joaoandremartins:
Fixes #315

### DIFF
--- a/spring-cloud-gcp-logging/README.adoc
+++ b/spring-cloud-gcp-logging/README.adoc
@@ -26,6 +26,11 @@ Platform.
 This module provides support for associating a web request trace ID with the corresponding log entries.
 This allows grouping of log messages by request.
 
+NOTE: Due to the way logging is set up, the GCP project ID and credentials defined in
+`application.properties` are ignored.
+Instead, you should set the `GOOGLE_CLOUD_PROJECT` and `GOOGLE_APPLICATION_CREDENTIALS` environment
+variables to the project ID and credentials private key location, respectively.
+
 `TraceIdLoggingWebMvcInterceptor` extracts the request trace ID from an HTTP request using a
 `TraceIdExtractor` and stores it in a thread-local of the https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/google-cloud-logging/src/main/java/com/google/cloud/logging/TraceLoggingEnhancer.java[`TraceLoggingEnhancer`]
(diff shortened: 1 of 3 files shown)


spring-cloud-gcp-logging/README.adoc
```diff
@@ -26,6 +26,11 @@ Platform.
 This module provides support for associating a web request trace ID with the corresponding log entries.
 This allows grouping of log messages by request.
 
+NOTE: Due to the way logging is set up, the GCP project ID and credentials defined in
+`application.properties` are ignored.
+Instead, you should set the `GOOGLE_CLOUD_PROJECT` and `GOOGLE_APPLICATION_CREDENTIALS` environment
```
ChengyuanZhao: Doesn't hurt to set both of course, but the credentials file also specifies the project, so using only GOOGLE_APPLICATION_CREDENTIALS works too.

joaoandremartins: Good point, from what I observed, only application-default credentials contain the project ID. Service account credentials don't, which is why I added the `GOOGLE_CLOUD_PROJECT` safeguard here.

PR #359, opened 2018-01-29
Binder merger

joaoandremartins:


joaoandremartins:
@snicoll @dsyer FYI. This is essentially a rebase of the commit from #324

PR #360, opened 2018-01-29
Merges Integration with Pub/Sub and Storage modules

### LINKED ISSUE #348: Merge Spring Integration Storage channel adapters with spring-cloud-gcp-storage module

### LINKED ISSUE #349: Merge Spring Integration Pub/Sub channel adapters with spring-cloud-gcp-pubsub

joaoandremartins:
Fixes #348 
Fixes #349


spring-cloud-gcp-storage/pom.xml
```diff
@@ -23,5 +23,14 @@
 			<groupId>com.google.cloud</groupId>
 			<artifactId>google-cloud-storage</artifactId>
 		</dependency>
+		<dependency>
+			<groupId>org.springframework.integration</groupId>
+			<artifactId>spring-integration-core</artifactId>
```
meltsufin: Should this be `optional` as well?

spring-cloud-gcp-pubsub/pom.xml
```diff
@@ -25,7 +25,10 @@
 			<groupId>org.springframework.cloud</groupId>
 			<artifactId>spring-cloud-gcp-core</artifactId>
 		</dependency>
-
+		<dependency>
+			<groupId>org.springframework.integration</groupId>
+			<artifactId>spring-integration-core</artifactId>
```
meltsufin: Should it be `optional`, since it's possible to use the module with Spring Integration?

spring-cloud-gcp-storage/pom.xml
artembilan: We don't this dependency at all.
It is transitive with the `spring-integration-file`

joaoandremartins:
I actually quite liked the separation to denote separate projects. e.g.,
`org.springframework.cloud.gcp` for Spring Cloud
`org.springframework.integration.gcp` for Spring Integration.

Although I guess it would also look nice to only have one package and have integration belong to it.

I'm kinda leaning more towards the first. But @artembilan if you feel strongly about it, I can change it to your suggestion.

artembilan:
I can't find the recommendations and bottlenecks about jar structure right now, but seems for me it is wrong to have two branches in one jar. The Maven search must answer to the package uniquely as well.

We may ask @snicoll to share his vision and experience.

Thanks

joaoandremartins:
The way I see it, the branches would always be there, they would just assume different shapes.
Let's wait to hear from @snicoll

snicoll:
I agree with @artembilan. If there isn’t a GCP outside of Spring Cloud GCP, then it  should be under the `org.springframework.cloud.gcp` package. As such `org.springframework.gcp.integration` sounds a good candidate to me.

If there is a GCP concept outside of Spring Cloud GCP, then that code should move there.

artembilan:
I even think that all the stuff in the same Maven group should go in the same root package exactly as that group. 

Seems for me Stephane did some typo in his comment and he definitely meant `org.springframework.cloud.gcp.integration` 😉

snicoll:
Indeed! I meant `org.springframework.cloud.gcp.integration`

joaoandremartins:
PTAL

PR #364, opened 2018-01-29
Splits the SQL starter into a MySQL and a PostgreSQL one

joaoandremartins:
Instead of users specifying their database vendor in the properties, we infer what vendor they're using with `@ConditionalOnClass` and generate JDBC driver class name and URL based on that.


joaoandremartins:
Still work needed to update the refdoc.

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/sql/GcpCloudSqlAutoConfigurationMockTests.java
```diff
@@ -19,160 +19,141 @@
...
-		@Override
-		public void test() {
-			assertEquals("jdbc:postgresql://google/test-database?socketFactory="
-					+ "com.google.cloud.sql.postgres.SocketFactory&"
-					+ "socketFactoryArg=tubular-bells:singapore:test-instance&useSSL=false",
-					this.urlProvider.getJdbcUrl());
-			assertEquals("org.postgresql.Driver", this.urlProvider.getJdbcDriverClass());
-		}
-	}
+//	@TestPropertySource(properties = {
```
joaoandremartins: @snicoll is there a way to also control what's in the classpath with `ApplicationContextRunner`? I suspect not, just wanted your confirmation.

One solution to make this test work is to create two separate auto configuration classes that are `@ConditionalOnClass` and add each one separately to each context, but I'd really like to avoid doing that...

snicoll: Yes. You can tune the `ClassLoader` with a `FilteredClassLoader`. See [this example](https://github.com/spring-projects/spring-boot/blob/3f5adfbccc9a624176e17933060003e624bdfaaf/spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/jdbc/DataSourceAutoConfigurationTests.java#L166) that we use to test connection pool detection. You can also filter an individual class.

You can also overwrite the classpath with a more recent version of a jar but that is a rule that applies to the complete test.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/sql/GcpCloudSqlAutoConfiguration.java
```diff
@@ -92,11 +91,32 @@ public CloudSqlJdbcInfoProvider appengineCloudSqlJdbcInfoProvider() {
 
 	@Bean
 	@ConditionalOnMissingBean(CloudSqlJdbcInfoProvider.class)
-	public CloudSqlJdbcInfoProvider defaultJdbcInfoProvider() {
-		CloudSqlJdbcInfoProvider defaultProvider = new DefaultCloudSqlJdbcInfoProvider(this.gcpCloudSqlProperties);
+	@ConditionalOnClass(com.google.cloud.sql.mysql.SocketFactory.class)
```
snicoll: `ConditionalOnClass`  on a method is an anti-pattern. It should be added to a `class` (it can be inner` so that the JVM doesn't load it at all if the target class isn't present.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/sql/GcpCloudSqlAutoConfiguration.java
```diff
@@ -92,11 +91,32 @@ public CloudSqlJdbcInfoProvider appengineCloudSqlJdbcInfoProvider() {
...
+		}
+
+		setCredentialsProperty();
+
+		return defaultProvider;
+	}
+
+	@Bean
+	@ConditionalOnMissingBean(CloudSqlJdbcInfoProvider.class)
+	@ConditionalOnClass(com.google.cloud.sql.mysql.SocketFactory.class)
```
snicoll: Ditto

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/sql/GcpCloudSqlAutoConfigurationMockTests.java
snicoll: It is [also available in the latest documentation](https://docs.spring.io/spring-boot/docs/current-SNAPSHOT/reference/htmlsingle/#boot-features-test-autoconfig)

spring-cloud-gcp-starters/pom.xml
```diff
@@ -24,8 +24,9 @@
 		<module>spring-cloud-gcp-starter-pubsub</module>
 		<module>spring-cloud-gcp-starter-config</module>
 		<module>spring-cloud-gcp-starter-storage</module>
-		<module>spring-cloud-gcp-starter-sql</module>
 		<module>spring-cloud-gcp-starter-trace</module>
 		<module>spring-cloud-gcp-starter-logging</module>
-	</modules>
+		<module>spring-cloud-gcp-starter-sql-mysql</module>
+        <module>spring-cloud-gcp-starter-sql-postgresql</module>
+    </modules>
```
snicoll: As I've already indicated, I am -1 on a starter per database.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/sql/GcpCloudSqlAutoConfiguration.java
joaoandremartins: Then the only alternative I see is to create two SQL auto configuration classes: `GcpCloudSqlMySqlAutoConfiguration` and `GcpCloudSqlPostgreSqlAutoConfiguration`. I'm not sure I like this approach, there would be a lot of code duplication...

WDYT?

snicoll: Having auto-configuration for MySQL and PostgreSQL in the same class is a smell anyway. I am sure there is a way to restructure the code to avoid the duplication. Nothing Spring Boot related there IMO.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/sql/MySqlJdbcInfoProviderAutoConfiguration.java
```diff
@@ -0,0 +1,59 @@
...
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+/**
+ * Provides the {@link CloudSqlJdbcInfoProvider} for MySQL.
+ *
+ * @author João André Martins
+ */
+@Configuration
+@ConditionalOnClass({com.google.cloud.sql.mysql.SocketFactory.class, GcpCloudSqlProperties.class})
```
artembilan: Shound't we be sure in that we have MySQL on CP as well, e.g. `@ConditionalOnClass` for the `com.mysql.jdbc.Driver` class?

Also: shound't we have something like `@AutoConfigureBefore(GcpCloudSqlAutoConfiguration.class)`  ?

spring-cloud-gcp-autoconfigure/pom.xml
```diff
@@ -83,11 +77,10 @@
             <artifactId>postgres-socket-factory</artifactId>
             <optional>true</optional>
         </dependency>
-
         <dependency>
             <groupId>org.postgresql</groupId>
             <artifactId>postgresql</artifactId>
-            <optional>true</optional>
+            <scope>test</scope>
```
meltsufin: Why `test` vs `optional`. Also, if you're adding the `postgresql` driver, it seems that `mysql-connector-java` should also be in this POM. I think the reason it works without it is because we have the PostgreSQL test, but not a MySQL test in this module.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/sql/PostgreSqlJdbcInfoProviderAutoConfiguration.java
```diff
@@ -0,0 +1,58 @@
...
+@ConditionalOnClass({com.google.cloud.sql.postgres.SocketFactory.class, GcpCloudSqlProperties.class})
+@ConditionalOnProperty(
+		name = "spring.cloud.gcp.sql.enabled", havingValue = "true", matchIfMissing = true)
+public class PostgreSqlJdbcInfoProviderAutoConfiguration {
+
+	private static final Log LOGGER =
+			LogFactory.getLog(PostgreSqlJdbcInfoProviderAutoConfiguration.class);
+
+	@Bean
+	@ConditionalOnMissingBean(CloudSqlJdbcInfoProvider.class)
```
meltsufin: `@ConditionalOnClass(com.google.cloud.sql.postgres.SocketFactory.class)`?

spring-cloud-gcp-starters/pom.xml
meltsufin: We don't expect to have more in the near future. It's just a convenience for users and a better alternative to the combined `spring-cloud-gcp-starter-sql`. It doesn't have to translate into a direct entry in the Initializr.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/sql/MySqlJdbcInfoProviderAutoConfiguration.java
joaoandremartins: > Also: shound't we have something like `@AutoConfigureBefore(GcpCloudSqlAutoConfiguration.class)` ?

I also thought about that, but then I'm afraid that `appengineCloudSqlJdbcInfoProvider()` would not be correctly generated, for when an app runs on App Engine.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/sql/PostgreSqlJdbcInfoProviderAutoConfiguration.java
joaoandremartins: Stéphane mentioned that `@ConditionalOnClass` is an anti-pattern on methods.
That same conditional annotates this class, so the net result should be the same.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/sql/MySqlJdbcInfoProviderAutoConfiguration.java
```diff
@@ -0,0 +1,59 @@
...
+@ConditionalOnProperty(
+		name = "spring.cloud.gcp.sql.enabled", havingValue = "true", matchIfMissing = true)
+public class MySqlJdbcInfoProviderAutoConfiguration {
+
+	private static final Log LOGGER =
+			LogFactory.getLog(MySqlJdbcInfoProviderAutoConfiguration.class);
+
+	@Bean
+	@ConditionalOnMissingBean(CloudSqlJdbcInfoProvider.class)
+	@ConditionalOnClass(com.google.cloud.sql.mysql.SocketFactory.class)
```
artembilan: Why this `@ConditionalOnClass` is still here?

@snicoll , 

if that is not good to use  `@ConditionalOnClass` on method why then Spring Boot still allows to do that?

Thanks

spring-cloud-gcp-autoconfigure/pom.xml
joaoandremartins: This is only needed for the Postgre test.
The reason we don't need the MySQL one is because `com.google.cloud.sql.mysql-socket-factory` pulls it in transitively.
Same thing strangely doesn't happen with `com.google.cloud.sql.postgres-socket-factory`...

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/sql/MySqlJdbcInfoProviderAutoConfiguration.java
artembilan: OK. So who has a precedence? `appengineCloudSqlJdbcInfoProvider`? So, this one should be extracted to its own `AutoConfig` class and mustbe marked with the `@AutoConfigureBefore(GcpCloudSqlAutoConfiguration.class, MySqlJdbcInfoProviderAutoConfiguration.class, PostgreSqlJdbcInfoProviderAutoConfiguration.class)`

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/sql/MySqlJdbcInfoProviderAutoConfiguration.java
joaoandremartins: Oops, my bad! Good catch.

spring-cloud-gcp-autoconfigure/pom.xml
artembilan: That's not surprise. See its' POM:
```
   <dependencies>
        <dependency>
            <artifactId>jdbc-socket-factory-core</artifactId>
            <version>1.0.4</version>
        </dependency>
    </dependencies>
```
unlike that one for the `mysql-socket-factory`:
```
  <dependencies>
        <dependency>
            <groupId>mysql</groupId>
            <artifactId>mysql-connector-java</artifactId>
            <version>5.1.38</version>
        </dependency>
        <dependency>
            <artifactId>jdbc-socket-factory-core</artifactId>
            <version>1.0.4</version>
        </dependency>
    </dependencies>
```

Shouldn't it be fixed on the GCP level? Or let's just try to figure out what was the reason do not include `postgres` transitively.

Thanks

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/sql/MySqlJdbcInfoProviderAutoConfiguration.java
joaoandremartins: Would adding `@Primary` to `appengineCloudSqlJdbcInfoProvider` also do?
I'd like to prevent adding yet another auto configuration class if I can help it...

spring-cloud-gcp-autoconfigure/pom.xml
joaoandremartins: I kind of assumed the Cloud SQL team had a good reason for having the MySQL dependency there..

In the end, I had to mark the PostgreSQL dependency as `<scope>compile</scope>` in order to adds its driver to the other `@ConditionalOnClass`.

What we can do is to exclude the MySQL dependency from the `mysql-socket-factory` import and then add it ourselves with `<scope>compile</scope>`..

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/sql/MySqlJdbcInfoProviderAutoConfiguration.java
artembilan: No, that's not good. The `@Primary` is for use-cases when you have several beans of the same type. It's not related to what we would like to achieve here - single unique `CloudSqlJdbcInfoProvider` bean.

With the current state you don't have guarantee which class will be loaded first. So, you may have `MySqlJdbcInfoProviderAutoConfiguration` configured already, but we do not even go to the `appengineCloudSqlJdbcInfoProvider` to have its `@Primary` on board because we don't overcome  `@ConditionalOnMissingBean(CloudSqlJdbcInfoProvider.class)`.

That's not a big deal to have `AutoConfig` class per feature. What is the obstacle do you see that way?

joaoandremartins: For me, less classes makes the code easier to understand.
But you're right, we could run into that situation. Will fix.

spring-cloud-gcp-autoconfigure/pom.xml
```diff
@@ -63,12 +63,18 @@
...
+                    <artifactId>mysql-connector-java</artifactId>
+                </exclusion>
+            </exclusions>
         </dependency>
 
         <dependency>
             <groupId>mysql</groupId>
             <artifactId>mysql-connector-java</artifactId>
-            <optional>true</optional>
+            <scope>compile</scope>
```
artembilan: Why `compile`? Seems for me it is going to be included transitively when we add this ` spring-cloud-gcp-autoconfigure` dependency: https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html

> compile
This is the default scope, used if none is specified. Compile dependencies are available in all classpaths of a project. Furthermore, those dependencies are propagated to dependent projects.

Definitely must be `optional`. That's how BIG autocofigure module works.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/sql/AppEngineJdbcInfoProviderAutoConfiguration.java
```diff
@@ -0,0 +1,60 @@
...
+import org.springframework.context.annotation.Conditional;
+import org.springframework.context.annotation.Configuration;
+
+/**
+ * Provides the {@link CloudSqlJdbcInfoProvider} for apps running on Google App Engine.
+ *
+ * @author João André Martins
+ */
+@Configuration
+@ConditionalOnClass({com.google.cloud.sql.mysql.SocketFactory.class, GcpCloudSqlProperties.class})
```
artembilan: I think we don't need condition for the `GcpCloudSqlProperties` class since it is exactly here in this package.
So, it is always on CP.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/sql/MySqlJdbcInfoProviderAutoConfiguration.java
```diff
@@ -35,6 +36,7 @@
 		com.mysql.jdbc.Driver.class})
 @ConditionalOnProperty(
 		name = "spring.cloud.gcp.sql.enabled", havingValue = "true", matchIfMissing = true)
+@AutoConfigureAfter(AppEngineJdbcInfoProviderAutoConfiguration.class)
```
artembilan: Why still no `@AutoConfigureBefore(GcpCloudSqlAutoConfiguration.class)` ?

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/sql/AppEngineJdbcInfoProviderAutoConfiguration.java
```diff
@@ -0,0 +1,60 @@
...
+@ConditionalOnProperty(
+		name = "spring.cloud.gcp.sql.enabled", havingValue = "true", matchIfMissing = true)
+public class AppEngineJdbcInfoProviderAutoConfiguration {
```
artembilan: I still think that this one should have `@AutoConfigureBefore(GcpCloudSqlAutoConfiguration.class)` as well.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/sql/MySqlJdbcInfoProviderAutoConfiguration.java
joaoandremartins: Isn't that taken care of naturally by the dependency graph?

i.e., `GcpCloudSqlAutoConfiguration` requires a `CloudSqlJdbcInfoProvider`, so the other auto configs will be tried first?

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/sql/AppEngineJdbcInfoProviderAutoConfiguration.java
meltsufin: What about PostgreSQL support on App Engine standard?

joaoandremartins: There's nothing specifically about PostgreSQL in GAE's documentation.
If we want to do anything there, then I'd do it in another PR.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/sql/MySqlJdbcInfoProviderAutoConfiguration.java
artembilan: Oh! I don't know... There is no any hints on the `@AutoConfigureAfter/@AutoConfigureBefore` to indicate such a situation. But, yeah. You might be right, that import for the `CloudSqlJdbcInfoProvider` should trigger those auto-configs.

We may ask @snicoll to enlighten this subject.
Thanks

joaoandremartins: Tests are passing and I'm fairly certain that the dependency graph is solving the issue for us.
So I would suggest that if @snicoll detects any issue when he comes back tomorrow, we issue another small PR to fix?

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/sql/AppEngineJdbcInfoProviderAutoConfiguration.java
meltsufin: There is [documentation for Java 8](https://cloud.google.com/appengine/docs/standard/java/cloud-sql/using-cloud-sql-postgres), but I guess it's no different than connecting from App Engine flexible.

Does this mean that CloudSQL-MySQL on App Engine standard Java 8, could work without this special configuration? @saturnism

joaoandremartins: AFAICT MySQL on App Engine has that special `<use-google-connector-j>true</use-google-connector-j>` element, which I think is us telling App Engine that the special driver should be used.

https://cloud.google.com/appengine/docs/standard/java/cloud-sql/using-cloud-sql-mysql

joaoandremartins:
Thanks for the reminder.
Updated my local copyright template for 2018.

snicoll:
@meltsufin can't answer your comment there for some reason.

> We don't expect to have more in the near future. It's just a convenience for users and a better alternative to the combined spring-cloud-gcp-starter-sql. It doesn't have to translate into a direct entry in the Initializr.

Duly noted. For the record, my remark has nothing to do with Spring Initializr.

spring-cloud-gcp-starters/pom.xml
snicoll: Duly noted. For the record, my remark has nothing to do with Spring Initializr.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/sql/MySqlJdbcInfoProviderAutoConfiguration.java
snicoll: > GcpCloudSqlAutoConfiguration requires a CloudSqlJdbcInfoProvider, so the other auto configs will be tried first?

As long as you are injecting a bean by type (and only that) what @joaoandremartins said. But as soon as you are playing with conditions, it will break.

My advice would be to order auto-configurations properly regardless.

joaoandremartins: OK, added `@AutoConfigureBefore(GcpCloudSqlAutoConfiguration.class)` to `AppEngineJdbcInfoProviderAutoConfiguration`.
Together with the `@AutoConfigureAfter(AppEngineJdbcInfoProviderAutoConfiguration.class)`'s in `MySqlJdbcInfoProviderAutoConfiguration` and `PostgreSqlJdbcInfoProviderAutoConfiguration`, that's hopefully enough ordering.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/sql/AppEngineJdbcInfoProviderAutoConfiguration.java
```diff
@@ -0,0 +1,62 @@
...
+import org.springframework.context.annotation.Conditional;
+import org.springframework.context.annotation.Configuration;
+
+/**
+ * Provides the {@link CloudSqlJdbcInfoProvider} for apps running on Google App Engine.
+ *
+ * @author João André Martins
+ */
+@Configuration
+@ConditionalOnClass({com.google.cloud.sql.mysql.SocketFactory.class})
```
snicoll: Why is there such a condition there? Looks like an oversight. Or is MySQL specific to GAE?

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/sql/AppEngineJdbcInfoProviderAutoConfiguration.java
```diff
@@ -0,0 +1,62 @@
...
+		name = "spring.cloud.gcp.sql.enabled", havingValue = "true", matchIfMissing = true)
+@AutoConfigureBefore(GcpCloudSqlAutoConfiguration.class)
+public class AppEngineJdbcInfoProviderAutoConfiguration {
+
+	private static final Log LOGGER =
+			LogFactory.getLog(AppEngineJdbcInfoProviderAutoConfiguration.class);
+
+	@Bean
+	@ConditionalOnMissingBean(CloudSqlJdbcInfoProvider.class)
+	@Conditional(AppEngineCondition.class)
```
snicoll: Given the name of this auto-configuration, I'd move this condition at class level.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/sql/AppEngineJdbcInfoProviderAutoConfiguration.java
```diff
@@ -0,0 +1,62 @@
...
+	public CloudSqlJdbcInfoProvider appengineCloudSqlJdbcInfoProvider(
+			GcpCloudSqlProperties gcpCloudSqlProperties) {
+		CloudSqlJdbcInfoProvider appEngineProvider =
+				new AppEngineCloudSqlJdbcInfoProvider(gcpCloudSqlProperties);
+
+		if (LOGGER.isInfoEnabled()) {
+			LOGGER.info("App Engine JdbcUrl provider. Connecting to "
+					+ appEngineProvider.getJdbcUrl() + " with driver "
```
snicoll: It probably has nothing to do with this PR per se but I find this a bit confusing. The only condition is on the enabled property (and classpath check).

And then, this claim that AppEngine is "configured". But if the user did not, you'll get `null` for those, isn't it? Or is this filled in by GAE?

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/sql/GcpCloudSqlAutoConfiguration.java
```diff
@@ -25,33 +25,27 @@
...
+ * database and instance connection name.
  *
  * @author João André Martins
  */
 @Configuration
 @EnableConfigurationProperties({ GcpCloudSqlProperties.class, DataSourceProperties.class })
-@ConditionalOnClass(com.google.cloud.sql.mysql.SocketFactory.class)
-@ConditionalOnProperty(name = "spring.cloud.gcp.sql.enabled", havingValue = "true", matchIfMissing = true)
+@ConditionalOnProperty(
+		name = "spring.cloud.gcp.sql.enabled", havingValue = "true", matchIfMissing = true)
```
snicoll: Don't we want to check that the `spring.datasource.url` has _not_ been set? It feels to me that you should backoff if the user has expressed such an opinion.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/sql/MySqlJdbcInfoProviderAutoConfiguration.java
```diff
@@ -0,0 +1,60 @@
...
+ * Provides the {@link CloudSqlJdbcInfoProvider} for MySQL.
+ *
+ * @author João André Martins
+ */
+@Configuration
+@ConditionalOnClass({com.google.cloud.sql.mysql.SocketFactory.class, com.mysql.jdbc.Driver.class})
+@ConditionalOnProperty(
+		name = "spring.cloud.gcp.sql.enabled", havingValue = "true", matchIfMissing = true)
+@AutoConfigureAfter(AppEngineJdbcInfoProviderAutoConfiguration.class)
+public class MySqlJdbcInfoProviderAutoConfiguration {
```
snicoll: That shouldn't be an auto-configuration. If you've moved this from `GcpCloudSqlAutoConfiguration`, it  should be a package  protected simple `Configuration` class that you import using `@Import` on the main auto-configuration.

That way you don't have to repeat the conditions (repeating the property check is a smell IMO).

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/sql/PostgreSqlJdbcInfoProviderAutoConfiguration.java
```diff
@@ -0,0 +1,61 @@
...
+ *
+ * @author João André Martins
+ */
+@Configuration
+@ConditionalOnClass({com.google.cloud.sql.postgres.SocketFactory.class,
+		org.postgresql.Driver.class})
+@ConditionalOnProperty(
+		name = "spring.cloud.gcp.sql.enabled", havingValue = "true", matchIfMissing = true)
+@AutoConfigureAfter(AppEngineJdbcInfoProviderAutoConfiguration.class)
+public class PostgreSqlJdbcInfoProviderAutoConfiguration {
```
snicoll: Idem

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/sql/GcpCloudSqlAutoConfiguration.java
```diff
@@ -73,40 +67,6 @@ public GcpCloudSqlAutoConfiguration(GcpCloudSqlProperties gcpCloudSqlProperties,
...
-		}
-
-		setCredentialsProperty();
-
-		return defaultProvider;
-	}
-
 	@Bean
 	@Primary
 	public DataSourceProperties cloudSqlDataSourceProperties(DataSourceProperties properties,
```
snicoll: What happens if a `CloudSqlJdbcInfoProvider` could not  be created? I think there is a case missing here. If neither postgres nor mysql socket factory is available, then no instance  should be created and this part will blow up unless I am missing something.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/sql/AppEngineJdbcInfoProviderAutoConfiguration.java
joaoandremartins: [App Engine also requires the socket factory to be in the dependencies](https://cloud.google.com/appengine/docs/standard/java/cloud-sql/using-cloud-sql-mysql).

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/sql/AppEngineJdbcInfoProviderAutoConfiguration.java
joaoandremartins: I don't think I understand what you mean.

The condition is simply for logging. The user actually doesn't have to configure almost anything, we already provide the JDBC URL and driver class name so they don't have to.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/sql/GcpCloudSqlAutoConfiguration.java
joaoandremartins: We don't want to do that here.
The `cloudSqlDataSourceProperties` bean checks for the unlikely case where a user specifies the JDBC URL, then overrides our choice with theirs if they did.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/sql/GcpCloudSqlAutoConfiguration.java
joaoandremartins: That is a corner case.
If there is no MySQL or PostgreSQL drivers on the classpath, then Spring Cloud GCP is not what our user is looking for.
Making debugging for that case a little easier could be useful. However, adding log lines to this method won't help since we won't get to the method itself due to the lack of dependency...

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/sql/MySqlJdbcInfoProviderAutoConfiguration.java
joaoandremartins: If that way, the `@ConditionalOnClass` annotations work as well, I think we can do it.
Will take it for a spin.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/sql/AppEngineJdbcInfoProviderAutoConfiguration.java
meltsufin: This is the exact signal from the classpath that CloudSQL-MySQL is intended to be used rather than the regular MySQL.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/sql/AppEngineJdbcInfoProviderAutoConfiguration.java
meltsufin: +1

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/sql/GcpCloudSqlAutoConfiguration.java
```diff
@@ -25,35 +25,33 @@
...
- * <p>Provides Google Cloud SQL instance connectivity through Spring JDBC by providing only a database and instance
- * connection name.
+ * Provides Google Cloud SQL instance connectivity through Spring JDBC by providing only a
+ * database and instance connection name.
  *
  * @author João André Martins
  */
 @Configuration
 @EnableConfigurationProperties({ GcpCloudSqlProperties.class, DataSourceProperties.class })
-@ConditionalOnClass(com.google.cloud.sql.mysql.SocketFactory.class)
```
meltsufin: Since it's no longer conditional on `com.google.cloud.sql.mysql.SocketFactory`, will it always be turned on regardless of the classpath?

artembilan: Ha! Good catch! I think should be `@ConditionalOnBean(CloudSqlJdbcInfoProvider.class)`

joaoandremartins: That's actually a really good point.
I think we will now need the SQL flag to always be turned on for this one as well.
I don't think we can rely on the classpath anymore...

joaoandremartins: How cool it would be to have an OR `@ConditionalOnClass` here :smile:

joaoandremartins: @artembilan I'm wondering at the time when `@ConditionalOnClass` is evaluated, would `@Import` have already kicked in?

meltsufin: Can we just split this class (`GcpCloudSqlAutoConfiguration`) into the `MySqlJdbcInfoProviderAutoConfiguration` and `PostgreSqlJdbcInfoProviderAutoConfiguration`?

artembilan: Right, `@Import` are loaded first: 
```
// Process any @Import annotations
processImports(configClass, sourceClass, getImports(sourceClass), true);

// Process any @ImportResource annotations
AnnotationAttributes importResource =
	AnnotationConfigUtils.attributesFor(sourceClass.getMetadata(), ImportResource.class);
...
// Process individual @Bean methods
Set<MethodMetadata> beanMethods = retrieveBeanMethodMetadata(sourceClass);
for (MethodMetadata methodMetadata : beanMethods) {
	configClass.addBeanMethod(new BeanMethod(methodMetadata, configClass));
}
```

joaoandremartins: What do you mean? That's what this PR already does.
It doesn't make sense to duplicate the `DataSourceProperties` bean creation, or the credentials setting IMO.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/sql/GcpCloudSqlAutoConfiguration.java
snicoll: What I mean is that, if no `CloudSqlJdbcInfoProvider` could be created this code _must not_ run at all, otherwise you'll have an exception on startup.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/sql/GcpCloudSqlAutoConfiguration.java
snicoll: That isn't right. Why would you do that?

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/sql/AppEngineJdbcInfoProviderAutoConfiguration.java
snicoll: I think we need to take a step back. Is there a description of this feature somewhere?

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/sql/AppEngineJdbcInfoProviderAutoConfiguration.java
snicoll: I don't think that answers my question. Why is it MySQL specific? 

(please don't assume I know those features, I don't. Perhaps a document that describes the features will help me but something called `AppEngineJdbcInfoProviderAutoConfiguration` that is MySQL specific isn't right unless AppEngine only supports MySQL.

meltsufin: I'm also a bit confused by the Cloud SQL story on App Engine Standard. One thing to keep in mind is that historically App Engine had its own special network path to many of the GCP services, including Cloud SQL. However, I believe that's changing now with the latest version of App Engine Standard for Java 8. I'll try to get clarity on this.

In any case, the way this class is set up is assuming that App Engine only supports the MySQL variant of Cloud SQL. So, the class should probably be renamed to `AppEngineCloudSqlMySqlJdbcInfoProviderAutoConfiguration`. :smile: 

cc: @saturnism @jabubake

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/sql/AppEngineJdbcInfoProviderAutoConfiguration.java
meltsufin: We have some explanation of the feature in the [README](https://github.com/spring-cloud/spring-cloud-gcp/tree/master/spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql). I'm not sure if it's sufficient though.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/sql/GcpCloudSqlAutoConfiguration.java
meltsufin: We basically allow users to specify just the instance name and database name, and we build the JDBC URL for them, which is a bit convoluted, or they can specify the full JDBC URL themselves. What's wrong with that?

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/sql/GcpCloudSqlAutoConfiguration.java
meltsufin: I agree with @snicoll. `CloudSqlJdbcInfoProvider` should probably be made ` @Autowired(required = false)` and if it's null, just return the passed in `DataSourceProperties`. Or, maybe there is another mechanism to achieve the same thing.

artembilan: Why `@ConditionalOnBean(CloudSqlJdbcInfoProvider.class)` doesn't work here for you?

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/sql/AppEngineJdbcInfoProviderAutoConfiguration.java
joaoandremartins: From https://cloud.google.com/appengine/docs/standard/java/cloud-sql/using-cloud-sql-postgres, it seems that connecting to PostgreSQL from inside App Engine is exactly the same as outside, in terms of the JDBC URL and driver class name.

Same thing doesn't happen with MySQL, where both URL and driver class names are special: https://cloud.google.com/appengine/docs/standard/java/cloud-sql/using-cloud-sql-mysql

Why is MySQL specific, I also don't know. :(

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/sql/AppEngineJdbcInfoProviderAutoConfiguration.java
joaoandremartins: This feature has been working for us pretty well, so far.
I'm not sure if anyone is using it, but we haven't heard any issues from users either.
The [refdoc](https://docs.spring.io/spring-cloud-gcp/docs/1.0.0.M1/reference/htmlsingle/#_spring_jdbc) also covers the feature in some detail.

I suggest if there are any issues with the way the feature works, we file it as a separate issue.
This PR is intended to split the starters into two, not fix retroactive issues.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/sql/GcpCloudSqlAutoConfiguration.java
joaoandremartins: +1 to what @meltsufin said.
Think of it as a simple JDBC URL override, for when a user doesn't like the JDBC URL we provide for whatever reason.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/sql/GcpCloudSqlAutoConfiguration.java
joaoandremartins: I really like `@ConditionalOnBean`. Will apply.

meltsufin: Always thinking in terms of ConditionalOn*Missing*Bean. No one said, it doesn't work. :smile:

joaoandremartins: After trying with `@ConditionalOnBean`, I don't think it does what we need.

From its Javadoc:

```
 * The condition can only match the bean definitions that have been processed by the
 * application context so far and, as such, it is strongly recommended to use this
 * condition on auto-configuration classes only. If a candidate bean may be created by
 * another auto-configuration, make sure that the one using this condition runs after.
```

It looks like the classes in `@Import` are "processed", but that process doesn't generate any beans.
The Javadoc also seems to imply that the beans should be generated by another auto-configuration, and the `@Import` might make it as if they are generated by *this* auto-configuration.

On top of that, the error message is very clear that a certain bean requires another bean of type `CloudSqlJdbcInfoProvider`.
Which is better than when `GcpCloudSqlAutoConfiguration` doesn't run at all:

```
Caused by: org.springframework.boot.autoconfigure.jdbc.DataSourceProperties$DataSourceBeanCreationException: Cannot determine embedded database driver class for database type NONE. If you want an embedded database please put a supported one on the classpath. If you have database settings to be loaded from a particular profile you may need to active it (no profiles are currently active).
```

All things considered, I would like to suggest that if we want to make any changes here, we do so in another issue so as not to drag the fork of the SQL starter.

artembilan: Well, when I add `@AutoConfigureBefore(GcpCloudSqlAutoConfiguration.class)` into the `MySqlJdbcInfoProviderAutoConfiguration` and `PostgreSqlJdbcInfoProviderAutoConfiguration` and I also have that `@ConditionalOnBean(CloudSqlJdbcInfoProvider.class)` on this `cloudSqlDataSourceProperties` all the tests in the `autoconfigure` pass for me.

:shrugging:

joaoandremartins: Interesting... it wasn't working with the `@AutoConfigureBefore` in the other classes and it probably should?
With that change, it's also working for me.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/sql/GcpCloudSqlAutoConfiguration.java
snicoll: I have no problem that you build the URL for them but if the user has provided one already you should not override it.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/sql/AppEngineJdbcInfoProviderAutoConfiguration.java
snicoll: > This PR is intended to split the starters into two, not fix retroactive issues.

OK, I'll stop commenting  here. The problem is that this PR is so large that it is very hard to track what is off topic. I'll review the code once the structure has been stabilized then.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/sql/GcpCloudSqlAutoConfiguration.java
snicoll: That's why auto-configurations should be ordered if they have dependencies. `@ConditionalOnBean` or `@ConditionalOnMissingBean` will not work if the bean definition is processed after the evaluation of the condition (it's not the same thing as an injection point that forces the `ApplicationContext`  to resolve the dependency.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/sql/GcpCloudSqlAutoConfiguration.java
joaoandremartins: Sorry if I wasn't clear before.
I meant we override the URL we created with the one the user provided, so it seems we're both on the same page!

PR #365, opened 2018-01-29
Added spanner template and unit tests

ChengyuanZhao:
Added classes for spanner template and operations with javadoc and unit tests


ChengyuanZhao:
Looks like reusing the spanner-template branch shows these commits that we covered in a previous PR and already merged. I'll make a new branch and PR.

PR #366, opened 2018-01-29
Spanner template

ChengyuanZhao:
added classes for template, operations, and tests

### DIFF
--- a/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerMutationFactory.java
+++ b/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerMutationFactory.java
@@ -50,7 +50,7 @@ public interface SpannerMutationFactory {
 	 * @param <T> The object's type.
 	 * @return The mutation operation to perform the action.
 	 */
-	<T> Mutation update(T object, String... properties);
+	<T> Mutation update(T object);
 
 	/**
 	 * Deletes several objects from Spanner.
--- a/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerMutationFactory.java
+++ b/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerMutationFactory.java
@@ -44,15 +44,6 @@ public interface SpannerMutationFactory {
 	<T> Mutation upsert(T object);
 
 	/**
-	 * Replaces a single object in Spanner. The columns' values corresponding to the
-	 * object's fields are treated according to Mutation.Op.REPLACE.
-	 * @param object The object to store.
-	 * @param <T> The object's type.
-	 * @return The mutation operation to perform this action.
-	 */
-	<T> Mutation replace(T object);
-
-	/**
 	 * Updates a single object in Spanner. The columns' values corresponding to the
 	 * object's fields are treated according to Mutation.Op.UPDATE.
 	 * @param object The object to update.
--- a/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerMutationFactory.java
+++ b/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerMutationFactory.java
@@ -1,0 +1,81 @@
+/*
+ *  Copyright 2018 original author or authors.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package org.springframework.cloud.gcp.data.spanner.core;
+
+import com.google.cloud.spanner.Mutation;
+
+/**
+ * Interface for a factory that creates Spanner mutation operation objects.
+ *
+ * @author Ray Tsang
+ * @author Chengyuan Zhao
+ */
+public interface SpannerMutationFactory {
+
+	/**
+	 * Stores a single object in Spanner.
+	 * @param object The object to store.
+	 * @param <T> The object's type.
+	 * @return The mutation operation which will store the object.
+	 */
+	<T> Mutation insert(T object);
+
+	/**
+	 * Updates or inserts a single object in Spanner. The columns' values corresponding to
+	 * the object's fields are treated according to Mutation.Op.INSERT_OR_UPDATE.
+	 * @param object The object to update or newly insert.
+	 * @param <T> The object's type.
+	 * @return The mutation operation to perform the action.
+	 */
+	<T> Mutation upsert(T object);
+
+	/**
+	 * Replaces a single object in Spanner. The columns' values corresponding to the
+	 * object's fields are treated according to Mutation.Op.REPLACE.
+	 * @param object The object to store.
+	 * @param <T> The object's type.
+	 * @return The mutation operation to perform this action.
+	 */
+	<T> Mutation replace(T object);
+
+	/**
+	 * Updates a single object in Spanner. The columns' values corresponding to the
+	 * object's fields are treated according to Mutation.Op.UPDATE.
+	 * @param object The object to update.
+	 * @param <T> The object's type.
+	 * @return The mutation operation to perform the action.
+	 */
+	<T> Mutation update(T object, String... properties);
+
+	/**
+	 * Deletes several objects from Spanner.
+	 * @param entityClass The type of the objects to delete.
+	 * @param entities A list of objects to delete. Each object can be a subtype of
+	 * entityClass.
+	 * @param <T> The type of object to delete.
+	 * @return The delete mutation.
+	 */
+	<T> Mutation delete(Class<T> entityClass, Iterable<? extends T> entities);
+
+	/**
+	 * Deletes a single object from Spanner.
+	 * @param object The object to delete.
+	 * @param <T> The type of the object to delete.
+	 * @return The delete mutation.
+	 */
+	<T> Mutation delete(T object);
+}
--- a/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerOperations.java
+++ b/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerOperations.java
@@ -24,8 +24,6 @@ import com.google.cloud.spanner.KeySet;
 import com.google.cloud.spanner.Options;
 import com.google.cloud.spanner.Statement;
 
-import org.springframework.cloud.gcp.data.spanner.core.mapping.SpannerMappingContext;
-
 /**
  * Defines operations available to use with Spanner.
  *
@@ -33,12 +31,6 @@ import org.springframework.cloud.gcp.data.spanner.core.mapping.SpannerMappingCon
  * @author Chengyuan Zhao
  */
 public interface SpannerOperations {
-
-	/**
-	 * Gets the mapping context used to facilitate the available options.
-	 * @return the mapping context.
-	 */
-	SpannerMappingContext getMappingContext();
 
 	/**
 	 * Gets the Spanner database client used to execute the operations.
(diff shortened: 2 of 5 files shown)


spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/mapping/SpannerObjectMapper.java
```diff
@@ -0,0 +1,46 @@
...
+import org.springframework.data.convert.EntityReader;
+import org.springframework.data.convert.EntityWriter;
+
+/**
+ * Interface for mappers that can populate fields from Spanner results.
+ *
+ * @author Ray Tsang
+ * @author Chengyuan Zhao
+ */
+public interface SpannerObjectMapper
```
meltsufin: Again, where's the implementation?

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
```diff
@@ -0,0 +1,173 @@
...
+ * @author Ray Tsang
+ * @author Chengyuan Zhao
+ */
+public class SpannerTemplate implements SpannerOperations {
+
+	private final DatabaseClient databaseClient;
+
+	private final SpannerMappingContext mappingContext;
+
+	private final SpannerObjectMapper objectMapper;
```
meltsufin: Is the assumption that the template will always be used with `@Entity` annotated beans? It then implies that the `SpannerTemplate` cannot be used with JPA.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerMutationFactory.java
```diff
@@ -0,0 +1,81 @@
...
+
+import com.google.cloud.spanner.Mutation;
+
+/**
+ * Interface for a factory that creates Spanner mutation operation objects.
+ *
+ * @author Ray Tsang
+ * @author Chengyuan Zhao
+ */
+public interface SpannerMutationFactory {
```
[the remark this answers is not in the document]
ChengyuanZhao: Was hoping to nail down these interfaces and template here in this PR first.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerOperations.java
```diff
@@ -0,0 +1,144 @@
...
+
+import org.springframework.cloud.gcp.data.spanner.core.mapping.SpannerMappingContext;
+
+/**
+ * Defines operations available to use with Spanner.
+ *
+ * @author Ray Tsang
+ * @author Chengyuan Zhao
+ */
+public interface SpannerOperations {
```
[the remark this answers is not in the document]
ChengyuanZhao: There is one that takes a `Statement`, which an SQL query, but I can add a straight string method too.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerOperations.java
```diff
@@ -0,0 +1,144 @@
...
+	 * Insert an object into storage.
+	 * @param object the object to insert.
+	 */
+	void insert(Object object);
+
+	/**
+	 * Update an object already in storage.
+	 * @param object the object to update.
+	 */
+	void update(Object object);
```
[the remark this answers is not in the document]
ChengyuanZhao: Will add.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
ChengyuanZhao: No, the beans don't need to be annotated with that.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/mapping/SpannerObjectMapper.java
ChengyuanZhao: I'm hoping to nail down the interfaces in this PR first. I don't think an implementation is necessary to evaluate these interfaces. This also goes along with our initial goal of just getting the template in first.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerOperations.java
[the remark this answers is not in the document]
meltsufin: Yeah, I think it's nice to have an API in the template that doesn't force users into Spanner-specific classes.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
meltsufin: Hmmm, how do you map columns to bean properties then?

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/mapping/SpannerObjectMapper.java
meltsufin: OK, fair enough.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerOperations.java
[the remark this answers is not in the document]
ChengyuanZhao: Actually, there is an update from what I see.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
ChengyuanZhao: Spring Data will map every single property in the given bean type, but it's up to us in SpannerObjectMapper and the mutation factory classes to decide which properties have types that we can handle.

meltsufin: What happens when a Spanner column is called "name", and the field in the class is "firstName". How will the user specify the mapping without JPA annotations?

ChengyuanZhao: For that we have the SpannerColumn annotation that lets you give a different name.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerOperations.java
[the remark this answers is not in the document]
meltsufin: `SpannerOperations.update(T object, String... properties)`? I don't see it.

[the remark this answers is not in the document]
ChengyuanZhao: I don't follow...

there's this in SpannerMutationFactory
```	
<T> Mutation update(T object, String... properties);
```

then there's this in SpannerOptions
```
void update(Object object);
```

then there's the usage of the factory method in SpannerTemplate
```
	@Override
	public void update(Object object) {
		Mutation mutation = this.mutationFactory.update(object);
		this.databaseClient.write(Arrays.asList(mutation));
	}
```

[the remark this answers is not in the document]
meltsufin: Where are you passing the `properties` to update in this example? The user might want to update only certain properties of the object that they are passing to `SpannerTemplate.update()`.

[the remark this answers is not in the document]
ChengyuanZhao: Ooh yes, I had missed the Properties part. Will add.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
meltsufin: I see. We might want to provide other options, for users who don't want to be coupled to Spanner-specific APIs.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerOperations.java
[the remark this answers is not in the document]
ChengyuanZhao: I decided to remove being able to specify this in the update command actually. I presume the purpose of letting people specify the specific columns they want to update is so they can set those values without having to populate the values they don't want to worry about. However I think they should just be doing that kind of thing outside of framework, like via an actual SQL query. it would lead to objects getting out of sync with what's stored in Spanner, too.

ChengyuanZhao:
@meltsufin @balopat PTAL

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerOperations.java
```diff
@@ -0,0 +1,156 @@
...
+ * @author Ray Tsang
+ * @author Chengyuan Zhao
+ */
+public interface SpannerOperations {
+
+	/**
+	 * Gets the mapping context used to facilitate the available options.
+	 * @return the mapping context.
+	 */
+	SpannerMappingContext getMappingContext();
```
balopat: Why is this method required? Can we introduce it when there are other users then the tests?

ChengyuanZhao: Yes I can remove this one. 

I'm fine with doing the "thin slice" approach, and I think the simple methods mentioned in this PR constitute thin enough of a slice. They're all pretty similar and will share lots of implementation logic. I don't think we should be slicing this particular PR any thinner.

balopat: Okay. I guess this "slice" can be borderline accepted size-wise, and I don't want you to revert all your changes artificially. The rationale behind going even more thinner slices, i.e. babysteps because that creates even smaller PRs (<<<700 lines, <<13 methods), so that people can review them quickly, also gives a chance to think about every single detail. 

Can you please provide the implementation for the other interfaces as well then, so that we have an end-to-end working SpannerTemplate?

ChengyuanZhao: We initially discussed getting the template up here and that is this interface here. I will add the an implementation in the very next PR, because that and its unit tests together will make this PR swell even more.

ChengyuanZhao:
@meltsufin @balopat PTAL. changes have been addressed. I'd really like to get these very-basic, uncontroversial functions in so implementations can be added.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerOperations.java
```diff
@@ -0,0 +1,148 @@
...
+ * @author Ray Tsang
+ * @author Chengyuan Zhao
+ */
+public interface SpannerOperations {
+
+	/**
+	 * Gets the Spanner database client used to execute the operations.
+	 * @return the Spanner database client.
+	 */
+	DatabaseClient getDatabaseClient();
```
balopat: I would like to remove this from the operations interface, because the idiomatic way to expose the DatabaseClient should be providing a FactoryBean for it, hence users can inject it easily if they need it. Putting this responsibility on the Template seems like mixing concerns, it's an implementation detail.

balopat: An example to expose the DatabaseClient could be the way [MongoClient is exposed in the MongoDb module](https://github.com/spring-projects/spring-data-mongodb/blob/master/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/MongoClientFactoryBean.java).

ChengyuanZhao: @balopat done. PTAL

balopat:
LGTM. @meltsufin @artembilan ?

ChengyuanZhao:
@meltsufin PTAL, this is according to plan laid out in meeting yesterday. Next will come an implementation for these operations for the repository types.

PR #368, opened 2018-01-31
Make GcpConfig configuration explicitly bootstrap

dsyer:
It isn't AutoConfiguration so it shouldn't use those features.
There is a remaining potential issue to do with
GcpProperties (there might be 2 instances, one in bootstrap
and one in the main context, or maybe that's not an issue). Also
possibly an issue: the bootstrap configuration requires
GoogleCredentials, but isn't conditional on that class. It isn't
an issue if that class is always available, but then if it is,
why is GcpContextAutoConfiguration conditional on it?

### DIFF
--- a/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/config/GcpConfigBootstrapConfiguration.java
+++ b/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/config/GcpConfigBootstrapConfiguration.java
@@ -1,5 +1,5 @@
 /*
- *  Copyright 2017 original author or authors.
+ *  Copyright 2017-2018 original author or authors.
  *
  *  Licensed under the Apache License, Version 2.0 (the "License");
  *  you may not use this file except in compliance with the License.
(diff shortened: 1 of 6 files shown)


spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/config/GcpConfigBootstrapConfiguration.java
```diff
@@ -35,9 +35,9 @@
...
  */
 @Configuration
-@AutoConfigureAfter(GcpContextAutoConfiguration.class)
+@Import(GcpContextAutoConfiguration.class)
 @EnableConfigurationProperties(GcpConfigProperties.class)
-public class GcpConfigAutoConfiguration {
+public class GcpConfigBootstrapConfiguration {
 
 	@Bean
 	@ConditionalOnProperty(prefix = "spring.cloud.gcp.config", name = "enabled", havingValue = "true")
```
meltsufin: Isn't `@ConditionalOnProperty` also an autoconfiguration feature? Will it still work for bootstrap configuration?

spring-cloud-gcp-autoconfigure/src/main/resources/META-INF/spring.factories
```diff
@@ -7,6 +7,4 @@ org.springframework.cloud.gcp.autoconfigure.storage.GcpStorageAutoConfiguration,
 org.springframework.cloud.gcp.autoconfigure.trace.StackdriverTraceAutoConfiguration
 
 org.springframework.cloud.bootstrap.BootstrapConfiguration=\
-org.springframework.cloud.gcp.autoconfigure.config.GcpConfigAutoConfiguration,\
-org.springframework.cloud.gcp.autoconfigure.core.GcpContextAutoConfiguration,\
-org.springframework.cloud.gcp.autoconfigure.config.GoogleConfigPropertySourceLocator
```
meltsufin: Why did we ever even have `GoogleConfigPropertySourceLocator` here, since it's not a Configuration class?

joaoandremartins: I'm not sure this works.
I spent a lot of time debugging an issue due to the lack of `GoogleConfigPropertySourceLocator` here.
Meant to take this for a spin, will do it now.

joaoandremartins:
This is working in the sample app, so that's good.
It's not working in my toy app, but I suspect of an issue with the app itself.

2 `GcpProperties` instances in bootstrap and app context: not a problem as far as I'm concerned, but since we can't reuse the one from `application.properties`, I actually think we can get rid of the one in the bootstrap context since we've already got `GcpConfigProperties.projectId` and `GcpConfigProperties.credentials`.

`spring-cloud-gcp-autoconfigure` pulls in `spring-cloud-gcp-core`, so `GoogleCredentials` will always be there. `GcpContextAutoConfiguration` conditional on `GoogleCredentials` is a very old artifact from the beginnings of this project. Understanding Spring better now, I think we can actually get rid of it.

@dsyer do you want to make these two changes in this PR? I can commit them tomorrow too.

meltsufin:
My commit 2cd8315 fixes #375

PR #369, opened 2018-01-31
Bumps SB2 version to RC1

joaoandremartins:
Seems like the Milestones repo doesn't have RC1 yet, so we'll have to wait a little bit to merge.


pom.xml
```diff
@@ -43,10 +43,6 @@
 		<spring-cloud-context.version>2.0.0.M5</spring-cloud-context.version>
```
joaoandremartins: Neither `spring-cloud-build` nor `spring-cloud-dependencies` specify versions for these `spring-cloud` dependencies.
I'd really like to point to some BOM to get rid of the versions here. Any other BOM I'm missing that contains these versions?

pom.xml
```diff
@@ -6,7 +6,7 @@
 	<parent>
 		<groupId>org.springframework.cloud</groupId>
 		<artifactId>spring-cloud-build</artifactId>
-		<version>2.0.0.M6</version>
+		<version>2.0.0.RC1</version>
```
artembilan: This is wrong. There is no `spring-cloud-build:2.0.0.RC1` yet: https://github.com/spring-cloud/spring-cloud-build/releases

pom.xml
artembilan: Take a look here: https://github.com/spring-cloud/spring-cloud-release/blob/master/spring-cloud-dependencies/pom.xml#L18.

Seems for me they are there. What ma I missing ?

artembilan: The `spring-cloud-context` is a part of the `spring-cloud-commons-dependencies` BOM: https://github.com/spring-cloud/spring-cloud-commons/blob/6e9745b7a431ac2673ab5ee182f32a7e57310e87/spring-cloud-commons-dependencies/pom.xml#L42

pom.xml
joaoandremartins: Oh! Then I guess we can't really move to SB2 RC1 until there is an actual `spring-cloud-build:2.0.0.RC1`? :(

Is there any rough estimate for that?

artembilan: I think you talk about this: https://github.com/spring-cloud/spring-cloud-release/milestone/37

artembilan: I still think this is wrong. See the link I've mentioned first. Plus pay attention to the Travis report.

spring-cloud-gcp-dependencies/pom.xml
```diff
@@ -5,7 +5,7 @@
 	<parent>
 		<groupId>org.springframework.cloud</groupId>
 		<artifactId>spring-cloud-dependencies-parent</artifactId>
-		<version>2.0.0.M6</version>
+		<version>2.0.0.RC1</version>
```
artembilan: This is still wrong. There is no such a version of this `spring-cloud-dependencies-parent`. Isn't it?

pom.xml
joaoandremartins: Will there be a `spring-cloud-build` RC1 release?

I was assuming there would be, because there was an M6 one.

> Seems like the Milestones repo doesn't have RC1 yet, so we'll have to wait a little bit to merge.

I added `spring-cloud-dependencies` as a BOM, but we still need a parent for these files.

artembilan: > I was assuming there would be, because there was an M6 one.

No, Spring Cloud and Spring Boot are different projects, although the first one depends on the second. But still it doesn't mean that they have to be released at the same day. I've pointed you to the `Finchley.M6` which is now planned for Feb 13 (disappointing 😢 ): https://github.com/spring-cloud/spring-cloud-release/milestone/37.

If we don't have chances to wait until there, we don't have choice unless rollback all the versions to the proper milestones and release as is, without any Spring Boot 2.0 RC1 goals.

joaoandremartins:
I'm wondering if cf0a6cc7f1f2f3cc5d89e59f6ebae5a83362cd06 is acceptable at all as a stop-gap solution?

pom.xml
```diff
@@ -40,7 +40,8 @@
 	<properties>
 		<maven.compiler.source>1.8</maven.compiler.source>
 		<maven.compiler.target>1.8</maven.compiler.target>
-		<spring-cloud-release.version>Finchley.M5</spring-cloud-release.version>
+        <spring-cloud-release.version>Finchley.M5</spring-cloud-release.version>
```
meltsufin: Tabs?

joaoandremartins: Fixed in https://github.com/spring-cloud/spring-cloud-gcp/pull/369/commits/5963a44ccee18d120866bdc7df39769f7448f455

artembilan: Late. See  what we have merged into the `master`: https://github.com/spring-cloud/spring-cloud-gcp/blob/master/pom.xml#L42 😄

meltsufin: Sorry, saw outdated PR somehow.

PR #370, opened 2018-01-31
Reverts @ConditionalOnBean in SQL autoconfig

joaoandremartins:
Even though tests are passing, my test app is erroring out at runtime.

### DIFF
--- a/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/sql/GcpCloudSqlAutoConfiguration.java
+++ b/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/sql/GcpCloudSqlAutoConfiguration.java
@@ -25,6 +25,7 @@ import org.apache.commons.logging.LogFactory;
 
 import org.springframework.boot.autoconfigure.AutoConfigureAfter;
 import org.springframework.boot.autoconfigure.AutoConfigureBefore;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnBean;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.boot.autoconfigure.jdbc.DataSourceAutoConfiguration;
 import org.springframework.boot.autoconfigure.jdbc.DataSourceProperties;
@@ -76,6 +77,7 @@ public class GcpCloudSqlAutoConfiguration {
 
 	@Bean
 	@Primary
+	@ConditionalOnBean(CloudSqlJdbcInfoProvider.class)
 	public DataSourceProperties cloudSqlDataSourceProperties(DataSourceProperties properties,
 			CloudSqlJdbcInfoProvider cloudSqlJdbcInfoProvider) {
 		if (StringUtils.isEmpty(properties.getUsername())) {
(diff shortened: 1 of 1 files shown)


joaoandremartins:
We might want to file a new issue after this PR is merged so we get to fix the issue with the lack of a `CloudSqlJdbcInfoProvider` bean.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/sql/GcpCloudSqlAutoConfiguration.java
```diff
@@ -56,7 +55,6 @@
 @Import({AppEngineJdbcInfoProviderAutoConfiguration.class,
 		MySqlJdbcInfoProviderAutoConfiguration.class,
 		PostgreSqlJdbcInfoProviderAutoConfiguration.class})
-@ConditionalOnBean(CloudSqlJdbcInfoProvider.class)
```
artembilan: How about if move this to the `cloudSqlDataSourceProperties` `@Bean` level?

joaoandremartins: That works! I should've thought of that...

joaoandremartins:
This is still not good enough, as now `GcpCloudSqlAutoConfiguration` is always kicking in and enabling `DataSourceProperties`...

If nothing else, we can always control this with a flag, but I wish there was a better way...

artembilan:
I'm not sure in your concern.

The `cloudSqlDataSourceProperties` is marked now with the `@ConditionalOnBean(CloudSqlJdbcInfoProvider.class)` which can be there only in these case:

- the `AppEngineJdbcInfoProviderAutoConfiguration`, but only if we meet a couple conditions on that class;
- the `MySqlJdbcInfoProviderAutoConfiguration` if meet its conditions;
- the `PostgreSqlJdbcInfoProviderAutoConfiguration` if meet its conditions again;
- and finally if end-user provides his/her own `CloudSqlJdbcInfoProvider` bean.

What else are we missing?

joaoandremartins:
The issue is that if I don't have `spring.cloud.gcp.sql.enabled=false` in `application.properties`, `GcpCloudSqlAutoConfiguration` is enabled, enables `DataSourceProperties` and then

`Caused by: org.springframework.beans.BeanInstantiationException: Failed to instantiate [org.springframework.boot.autoconfigure.jdbc.DataSourceProperties]: Constructor threw exception; nested exception is java.lang.NoClassDefFoundError: org/springframework/jdbc/datasource/embedded/EmbeddedDatabaseType`

This happens for a project that is intended to use e.g., Config only.

Since the bean conditional is now in the method, this autoconfig is always enabled. Unless `enabled=false` is provided, since if omitted, `matchIfMissing=true` brings it on.

artembilan:
??? 
```
java.lang.NoClassDefFoundError: org/springframework/jdbc/datasource/embedded/EmbeddedDatabaseType
```
How is it possible at all to live without `spring-jdbc` on classpath?

That's why I say: #371

artembilan:
FYI: #377

Issue #377, opened 2018-02-01
Add @ConditionalOnClass({ DataSource.class, EmbeddedDatabaseType.class }) to the GcpCloudSqlAutoConfiguration


artembilan:
Since we definitely produce in the `GcpCloudSqlAutoConfiguration` some JDBC infrastructure we have to be sure that all required classpath entries are present before going forward.

Not a big deal to fix, but I would like to have a test-case on the matter. Seems for me a `FilteredClassLoader` can help hear to cover this situation.

We also need to take a look to the way to have a much tests as possible. 
I've heard that even if our tests pass, there are some applications failing. 
That should be a signal to bring the use-case to the test coverage in the project.

Issue #371, opened 2018-01-31
SQL Starters have to depend on the spring-boot-starter-jdbc


artembilan:
The Cloud SQL auto-configuration relies on (or contribute to) the standard Spring Boot JDBC auto-configuration. It feels natural to add to the SQL starters `spring-boot-starter-jdbc` dependency and remove `HikariCP` one. And also remove `spring-jdbc` from the `spring-cloud-gcp-sql-sample`

PR #378, opened 2018-02-01
GH-377: Improve Cloud JDBC auto-configuration

artembilan:
Fixes spring-cloud/spring-cloud-gcp#377

* Since any SQL can't leave without `DataSource`, make sure we have it
on classpath
* Since `GcpCloudSqlAutoConfiguration` triggers `DataSourceProperties`
creating, make sure we have `EmbeddedDatabaseType.class`
on the classpath as well
* Move all the `*JdbcInfoProviderAutoConfiguration` classes to the
inner classes of the `GcpCloudSqlAutoConfiguration`.
They depend on the same classpath and properties rules.
Therefore it would be easier to avoid `@Conditional` copy/paste
* Add `spring-boot-starter-jdbc` to the `spring-cloud-gcp-autoconfigure`
POM to avoid extra dependencies.
Plus we will have required `spring-jdbc` classes we need to compile
`GcpCloudSqlAutoConfiguration`
* Add `logback-test.xml` to the `spring-cloud-gcp-autoconfigure`
to avoid DEBUG noise during testing
* Add `testNoJdbc()` to ensure that we don't trigger Cloud JDBC
auto-configuration when we don't have particular JDBC classes on the
classpath


spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/sql/GcpCloudSqlAutoConfiguration.java
```diff
@@ -19,144 +19,233 @@
...
 
 	private static final Log LOGGER = LogFactory.getLog(GcpCloudSqlAutoConfiguration.class);
 
-	private final GcpCloudSqlProperties gcpCloudSqlProperties;
 
-	private final GcpProperties gcpProperties;
+	/**
+	 * The AppEngine Configuration for the {@link AppEngineCloudSqlJdbcInfoProvider}.
+	 */
+	@ConditionalOnClass(com.google.cloud.sql.mysql.SocketFactory.class)
```
joaoandremartins: @snicoll mentioned using `@ConditionalOnClass` in methods to be an anti-pattern, which is why we created all the other autoconfig classes.

meltsufin:
Is this fixing #371 as well?

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/sql/GcpCloudSqlAutoConfiguration.java
```diff
@@ -19,144 +19,233 @@
...
-			// First tries the SQL configuration credential.
-			if (this.gcpCloudSqlProperties != null
-					&& this.gcpCloudSqlProperties.getCredentials() != null) {
-				credentialsLocationFile =
-						this.gcpCloudSqlProperties.getCredentials().getLocation().getFile();
+	@Configuration
+	@Import({ GcpCloudSqlAutoConfiguration.AppEngineJdbcInfoProviderConfiguration.class,
+			GcpCloudSqlAutoConfiguration.MySqlJdbcInfoProviderConfiguration.class,
+			GcpCloudSqlAutoConfiguration.PostgreSqlJdbcInfoProviderConfiguration.class })
+	static class CloudSqlDataSourcePropertiesConfiguration {
```
joaoandremartins: Not sure if I like this more than the three classes before...

artembilan:
No, that's not.
The `starter` subject is not related.

artembilan:
> Users don't need an EmbeddedDatabaseType to use our starter, it doesn't make much sense to require it.

??? With the `GcpCloudSqlAutoConfiguration` you trigger `@EnableConfigurationProperties({ GcpCloudSqlProperties.class, DataSourceProperties.class })`, where the last one depends on the `EmbeddedDatabaseConnection` which, in turn, on the `EmbeddedDatabaseType`. So, you can't trigger `new DataSourceProperties()` without those classes on the classpath. That's exactly what you see in your logs - no `spring-jdbc` on CP, wrong attempt to instantiate the `DataSourceProperties`.

And again: we fully blow up the JDBC stuff here, so we really should follow the approach we have in the Spring Boot for the mentioned `DataSourceAutoConfiguration`.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/sql/GcpCloudSqlAutoConfiguration.java
artembilan: ???
This is not method.
Let me repeat it again with a simplified code:
```
@ConditionalOnClass(com.google.cloud.sql.mysql.SocketFactory.class)
static class AppEngineJdbcInfoProviderConfiguration {
```
Where do you see method here, please?

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/sql/GcpCloudSqlAutoConfiguration.java
artembilan: Please, read the commit comments.
Otherwise we have to copy/paste all the conditions on the `GcpCloudSqlAutoConfiguration` to all those external classes.

We can't trigger, for example, `MySqlJdbcInfoProviderAutoConfiguration` if we don't have `DataSource.class` on CP. We can't do that when we don't meet `spring.cloud.gcp.sql.enabled` condition and also we have to trigger `@EnableConfigurationProperties(GcpCloudSqlProperties.class)` because this `MySqlJdbcInfoProviderAutoConfiguration` has to be performed before `cloudSqlDataSourceProperties`.

That's not about "like - don't like". That's more about the reasonable ceiling of copy/paste and how often we are going to miss add extra condition to those external classes.

OTOH, please, take a look into Spring Boot auto-configuration classes to understand how design can be and what would be better to avoid.

Does it make sense?

joaoandremartins:
We want to avoid enabling SQL support when we don't want it. That includes not enabling `DataSourceProperties`.

We can already do that using `spring.cloud.gcp.sql.enabled=false`, but I agree we want to avoid relying on the flag only (and having to ask users to set `enabled` to `true` if they want the feature).

That's why I'm suggesting depending on `DataSource`. If it's in the classpath, it's because the user probably wants it autoconfigured.

I've always thought the SQL support would end with providing a customised `DataSourceProperties`. And we don't really need JDBC for that (we weren't depending on it), `DataSourceProperties` is in `spring-boot-autoconfigure`.

artembilan:
Again: you have `@EnableConfigurationProperties(DataSourceProperties.class)`. That triggers a bean instantiation, but if you don't have `spring-jdbc`, you can't do that, therefore your exception. 
And that's already doesn't matter what is the choice from end-user to the `spring.cloud.gcp.sql.enabled`. That's his/her own trap to enable it, but don't have `spring-jdbc` in the classpath. So, the way `DataSourceAutoConfiguration` works we are not going to have even `DataSourceProperties` bean when you don't have `spring-jdbc` on the classpath. The `EmbeddedDatabaseType` is a part of `spring-jdbc` and we need this class to instantiate `DataSourceProperties` class.

I'm not sure what you else to tell you... 
Just try to call `new DataSourceProperties()` when you don't have `spring-jdbc` on classpath.

spring-cloud-gcp-autoconfigure/pom.xml
```diff
@@ -50,11 +50,9 @@
 
         <!-- Cloud SQL -->
 
-        <!-- HikariCP used as default DataSource -->
         <dependency>
-            <groupId>com.zaxxer</groupId>
-            <artifactId>HikariCP</artifactId>
-            <scope>compile</scope>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-jdbc</artifactId>
```
meltsufin: It seems odd to have a dependency on a starter in the autoconfigure module, albeit an optional one.
I don't see any examples of this in the Spring Boot project, and for starters in this project we can't do it because it would create a circular dependency.

artembilan: Well, agree. That really was a speculation on my side.
But I did it because I see how you use here already `spring-cloud-starter-sleuth`.

OTOH I don't think that it is a big problem since the `starter` brings for us just dependencies so we don't need to copy/paste them.

Not sure what you mean about "circular dependency" because this is `optional` here, plus `spring-boot-starter-jdbc` doesn't use our `spring-cloud-gcp-autoconfigure`. 

But again: since it is `optional` here and we definitely copy/paste some dependencies to the SQL starter, that sounds like it is much better to use this `spring-boot-starter-jdbc` as a strong dependency from the starters. But again: this is a story of the issue #371 .

If you feel strong, we can reconsider this line when we do that issue.

meltsufin: OK, sgtm.

snicoll:
I agree that a starter in that module feels wrong. Can you please reconsider it?

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/sql/GcpCloudSqlAutoConfiguration.java
joaoandremartins: After looking at this with fresh eyes, it makes sense to me.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/sql/GcpCloudSqlAutoConfiguration.java
joaoandremartins: Sorry, didn't realise this was a class, my bad!

artembilan:
@snicoll , @meltsufin, 

WDYT now about dependencies in our `autoconfigure` ?

Thanks

PR #382, opened 2018-02-01
Make the logging sample talk again

joaoandremartins:


spring-cloud-gcp-samples/spring-cloud-gcp-logging-sample/src/main/resources/logback-spring.xml
```diff
@@ -1,8 +1,11 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <configuration>
     <include resource="org/springframework/cloud/gcp/logging/logback-appender.xml" />
+    <include resource="org/springframework/boot/logging/logback/defaults.xml"/>
+    <include resource="org/springframework/boot/logging/logback/console-appender.xml" />
 
     <root level="INFO">
+        <appender-ref ref="CONSOLE"/>
```
meltsufin: I think it was intentional to only send logs to `STACKDRIVER` because otherwise logs will appear twice in Cloud Logging.

joaoandremartins: Why would they appear twice in Cloud Logging if you also ask the console to print them?
I'm double checking and the logs are only being printed once in my Cloud Logging.

meltsufin: If you run the application on GCP, stdout logs are also captured and sent to Cloud Logging.

joaoandremartins: They're sent to the service they're running in, not to `Global`, right? That sounds good to me. We should document that if a user doesn't want logs in both places, they can disable either the `STACKDRIVER` or the `CONSOLE` appender.

Having no logs in the console makes it very hard to debug the app in case anything goes wrong with the Stackdriver setup (e.g., no project ID, bad credentials, etc.).

meltsufin: When deployed to GCP, both log streams will be associated with the deployed application resource, not `Global`. Maybe writing the log to a file is better for debugging?

joaoandremartins:
PTAL

spring-cloud-gcp-samples/spring-cloud-gcp-logging-sample/README.adoc
```diff
@@ -49,3 +49,6 @@ labels."appengine.googleapis.com/trace_id"="[Trace-ID-Of-Request]"
 
 Expand the log entry labels and note that `appengine.googleapis.com/trace_id` label is set to the
 same value for both log messages. The trace ID header `x-cloud-trace-context` is set automatically when deployed to App Engine.
+
+When running on GAE or GKE, also remember to remove the `CONSOLE` appender, otherwise logs will
+appear on Cloud Logging in duplicate.
```
meltsufin: ...in duplicate, once in the `spring.log` stream and once in the `stdout` stream.

PR #390, opened 2018-02-01
Force InnoDB storage engine for JPA sample

### LINKED ISSUE #380: JPA sample can't create tables
`Caused by: java.sql.SQLException: Storage engine MyISAM is disabled (Table creation is disallowed).`

It appears that something maybe in SB2 changed JPAs dialect to something that Cloud SQL doesn't like.

https://cloud.google.com/sql/docs/mysql/diagnose-issues (search for MyISAM).

I've tried a few dialects, but haven't found one that works yet.

`spring.jpa.properties.hibernate.dialect=org.hibernate.dialect.MySQL5Dialect`
`spring.jpa.properties.hibernate.dialect=org.hibernate.dialect.MySQLInnoDBDialect`

It should be a matter of finding one that Cloud SQL likes.
https://docs.jboss.org/hibernate/orm/3.5/api/org/hibernate/dialect/package-summary.html

meltsufin:
Fixes #380.

### DIFF
--- a/spring-cloud-gcp-samples/spring-cloud-gcp-data-jpa-sample/src/main/resources/application.properties
+++ b/spring-cloud-gcp-samples/spring-cloud-gcp-data-jpa-sample/src/main/resources/application.properties
@@ -4,6 +4,6 @@ spring.cloud.gcp.sql.database-name=
 spring.jpa.hibernate.ddl-auto=create-drop
 
 # Cloud SQL (MySQL) only supports InnoDB, not MyISAM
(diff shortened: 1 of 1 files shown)


meltsufin:
Ideally, we should autoconfigure this, or if it's a problem in SB 2.0, it should be fixed there.

joaoandremartins:
I suspect this is only an issue because we're using `ddl-auto` in the sample.
But if we could auto-configure this, it would create less user friction.

Issue #397, opened 2018-02-02
Should Spring Cloud Config be extracted from Autoconfigure module?


saturnism:
It was weird to see Spring Cloud Config related doc in the Autoconfigure README.

https://github.com/spring-cloud/spring-cloud-gcp/blob/master/spring-cloud-gcp-autoconfigure/README.adoc

It took me a while to realize that Spring Cloud Config for GCP is now folded into Autoconfigure. Even thought there is few classes involved, organizationally it feels a little weird.

meltsufin:
We went back and fourth on this quiet a bit. Everyone seemed to be happy in the end to have it in the autoconfigure module and use a flag to turn it on.

saturnism:
it looks like we have the config starter, but the starter has no README like other starters.

saturnism:
Oh interesting, there are READMEs for both the implementing module and the starter module. Not quite sure where the line is. I see moving the README straight out of Autoconfigure in #409 is probably incomplete.

meltsufin:
I think, ultimately, we should try to cut down on the module READMEs and link to the refdoc wherever possible instead. Currently, there is too much documentation duplication.

saturnism:
+1. That would make sense if we can extract all READMEs to the ref doc.

joaoandremartins:
I think we can live with config in the `autoconfigure` module, let's close this one out.

PR #409, opened 2018-02-02
Moved Spring Cloud Config README from Autoconfigure to the Config Starter

saturnism:
For #397.

Spring Cloud GCP Config sample also references to this file.


joaoandremartins:
I think @artembilan is of the opinion that Spring users aren't used to looking for documentation in the starters, since they're only a collection of dependencies.
However, since we have READMEs in other starters, it might make sense to have the Config one there too.
Would you guys be opposed to gathering all of the READMEs in a single one, in the `autoconfigure` module?

meltsufin:
+1 for not duplicating docs.

Regarding this PR, however, config is the only starter for which the README was moved into autoconfigure. I think for the consistency with other starters, it should move back as this PR proposes.

Ultimately, we might want to move all starter docs into a single README or the refdoc, but that's a separate conversation.

artembilan:
Well, for consistency we don't have to have READMEs in starters at all. 

And if we have such a conversion already now why should we push it into the future, but don't just fix as a part of this PR? You definitely do logical things, but the final picture doesn't look like a set of reasonable actions. I mean what is the point to do this task if we agreed to redo it in the future?

I always tried to keep you away from these READMEs, but looks like I don't have enough authority in conviction... If we wouldn't have them in the beginning, there wouldn't be this discussion and such a PR at all.

saturnism:
It's currently linked from another place.

It would make sense if all READMEs are removed from starters and folded into ref doc one way or another. Should that be the next step?

PR #412, opened 2018-02-05
Makes every sample use mvn instead of mvnw

joaoandremartins:
Also fixes their READMEs.


joaoandremartins:
I think this will be resolved by just calling `mvn` instead of `mvnw` everywhere in the samples.

I wouldn't want to be running my samples from a directory that's not the sample's. It's adding convention where none is needed.

joaoandremartins:
PTAL, the samples are now all using `mvn` instead of `mvnw`.

PR #414, opened 2018-02-05
Adds the snapshots repo to every sample

### LINKED ISSUE #399: Add spring milestone repositories to sample pom.xml-s
With a clean start (no previous spring-cloud-gcp artifacts built), the starter can't be built in isolation, without running `mvn clean install` first on the whole big spring-cloud-gcp project.

We can either add the mvn clean install step as a first step everywhere, or include 

```
<repositories>
  <repository>
   <id>spring-milestones</id> 
   <name>Spring Milestones</name>
   <url>https://repo.spring.io/milestone</url>
  </repository>
</repositories>
```
to all the poms.

Error for the spring-cloud-gcp-data-jpa-sample

````
[ERROR] 'dependencies.dependency.version' for org.springframework.cloud:spring-cloud-gcp-starter-sql-mysql:jar is missing. @ org.springframework.cloud:spring-cloud-gcp-data-jpa-sample:[unknown-version], /Users/balintp/dev/proj/spring-cloud-gcp/spring-cloud-gcp-samples/spring-cloud-gcp-data-jpa-sample/pom.xml, line 21, column 15
 ````

joaoandremartins:
For users who don't have Spring Cloud GCP locally installed.

Fixes #399

### DIFF
--- a/spring-cloud-gcp-samples/spring-cloud-gcp-config-sample/pom.xml
+++ b/spring-cloud-gcp-samples/spring-cloud-gcp-config-sample/pom.xml
@@ -30,4 +30,18 @@
             <artifactId>spring-boot-starter-actuator</artifactId>
         </dependency>
     </dependencies>
+
+    <repositories>
+        <repository>
+            <id>spring-snapshots</id>
+            <name>Spring Snapshots</name>
+            <url>https://repo.spring.io/libs-snapshot-local</url>
+            <snapshots>
+                <enabled>true</enabled>
+            </snapshots>
+            <releases>
+                <enabled>false</enabled>
+            </releases>
+        </repository>
+    </repositories>
 </project>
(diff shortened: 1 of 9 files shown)


meltsufin:
Is this still an issue when running samples from project root as suggested in the other PR?

artembilan:
Why can't we go Spring Boot way here as well: https://github.com/spring-projects/spring-boot/tree/master/spring-boot-samples?

I like idea to make a sample project as independent as possible, but looks like that isn't possible without copy/pasting as we have a concern in the #412

meltsufin:
@artembilan It wouldn't be so bad because we would only copy-paste the wrapper to the parent of all samples (`spring-cloud-gcp-samples`) in that case.

Currently, I like the samples in the main project hierarchy, because finding compilation errors and refactoring seems easier.

joaoandremartins:
We could indeed copy the repo to the parent `samples` pom.
But while trying to reproduce this issue, I see that the snapshots/milestones repo seems to be obtained from the parent.

So I'm actually not sure if this change is required but, at the same time, I don't know how @balopat managed to trigger this issue...

joaoandremartins:
Turns out this is not an actual issue.

PR #416, opened 2018-02-05
Allow createSignedUrl() calls on non-existing files

### LINKED ISSUE #411: pre signed Cloud Storage URL
I'm trying to create a pre-signed URL for an upload for cloud storage. However I can't seem to get a value other than null from spring cloud gcp. Sample code is below.

```  
    /**
     * Generates a file with for a media type
     * @param extension The media type
     * @return The file you can upload to
     */
    private fun fileForExtension(extension: String) : File {

        val objectID = generateCloudObjectID(extension)
        val resource = (uploadBucket as GoogleStorageResourceBucket).createRelative(objectID) as GoogleStorageResourceObject
        val url = resource.createSignedUrl(TimeUnit.DAYS, 7L)
        return File(put = url, objectID = objectID)

    }
```

joaoandremartins:
Fixes #411

### DIFF
--- a/spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceBucket.java
+++ b/spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceBucket.java
@@ -69,7 +69,7 @@ public class GoogleStorageResourceBucket implements WritableResource {
 			return getGcsBucket();
 		}
 
-		if (!exists() && !this.createBucketIfNotExists) {
+		if (!this.createBucketIfNotExists) {
 			return null;
 		}
 
--- a/spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceBucket.java
+++ b/spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceBucket.java
@@ -66,16 +66,11 @@ public class GoogleStorageResourceBucket implements WritableResource {
 	/**
 	 * Creates the bucket that this {@link GoogleStorageResourceBucket} represents in Google Cloud
 	 * Storage.
-	 *
-	 * <p>If the bucket already exists in Google Cloud Storage, the
-	 * {@link com.google.cloud.storage.Bucket} object for that bucket is returned.
 	 * @return the {@link com.google.cloud.storage.Bucket} object for the bucket
+	 * @throws com.google.cloud.storage.StorageException if any errors during bucket creation arise,
+	 * such as if the bucket already exists
 	 */
 	public Bucket create() {
-		if (exists()) {
-			throw new IllegalStateException("The bucket " + this.bucketName + " already exists.");
-		}
-
 		return this.storage.create(BucketInfo.newBuilder(this.bucketName).build());
 	}
 
--- a/spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceBucket.java
+++ b/spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceBucket.java
@@ -59,11 +59,17 @@ public class GoogleStorageResourceBucket implements WritableResource {
 	/**
 	 * Creates the bucket that this {@link GoogleStorageResourceBucket} represents in Google Cloud
 	 * Storage.
-	 * @return the {@link com.google.cloud.storage.Blob} of the created object or {@code null} if
-	 * the bucket already exists or cannot be created
+	 * <p>If the bucket already exists in Google Cloud Storage, the
+	 * {@link com.google.cloud.storage.Bucket} object for that bucket is returned.
+	 * @return the {@link com.google.cloud.storage.Bucket} object for the bucket or {@code null} if
+	 * the bucket doesn't exist and cannot be created
 	 */
 	public Bucket create() {
-		if (exists() || !this.createBucketIfNotExists) {
+		if (exists()) {
+			return getGcsBucket();
+		}
+
+		if (!exists() && !this.createBucketIfNotExists) {
 			return null;
 		}
 
@@ -121,7 +127,7 @@ public class GoogleStorageResourceBucket implements WritableResource {
 	}
 
 	@Override
-	public Resource createRelative(String relativePath) throws IOException {
+	public GoogleStorageResourceObject createRelative(String relativePath) throws IOException {
 		GoogleStorageResourceObject resource = new GoogleStorageResourceObject(this.storage,
 				getURI().toString() + (relativePath.startsWith("/") ? "" : "/")
 						+ relativePath,
--- a/spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceBucket.java
+++ b/spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceBucket.java
@@ -124,6 +124,15 @@ public class GoogleStorageResourceBucket implements WritableResource {
 				"Cannot get last-modified for a storage bucket.");
 	}
 
+	/**
+	 * Creates a file within a given bucket. The file resource inherits
+	 * {@code createBlobIfNotExists} from this bucket's {@code createBucketIfNotExists}.
+	 *
+	 * <p>If the file did not exist in Google Cloud Storage, it is created by this method.
+	 * @param relativePath the URL for a Google Cloud Storage bucket
+	 * @return the {@link GoogleStorageResourceObject} object for the created file
+	 * @throws IOException
+	 */
 	@Override
 	public GoogleStorageResourceObject createRelative(String relativePath) throws IOException {
 		GoogleStorageResourceObject resource = new GoogleStorageResourceObject(this.storage,
--- a/spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceBucket.java
+++ b/spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceBucket.java
@@ -36,6 +36,7 @@ import org.springframework.util.Assert;
  * Implements {@link WritableResource} for buckets in Google Cloud Storage.
  *
  * @author Chengyuan Zhao
+ * @author João André Martins
  */
 public class GoogleStorageResourceBucket implements WritableResource {
 
@@ -43,7 +44,7 @@ public class GoogleStorageResourceBucket implements WritableResource {
 
 	private final String bucketName;
 
-	private final boolean createBucketIfRequired;
+	private final boolean createBucketIfNotExists;
 
 	public GoogleStorageResourceBucket(Storage storage, String bucketName,
 			boolean createBucketIfNotExists) {
@@ -52,16 +53,26 @@ public class GoogleStorageResourceBucket implements WritableResource {
 
 		this.storage = storage;
 		this.bucketName = bucketName;
-		this.createBucketIfRequired = createBucketIfNotExists;
+		this.createBucketIfNotExists = createBucketIfNotExists;
 	}
 
-	public void createBucket() {
-		this.storage.create(BucketInfo.newBuilder(this.bucketName).build());
+	/**
+	 * Creates the bucket that this {@link GoogleStorageResourceBucket} represents in Google Cloud
+	 * Storage.
+	 * @return the {@link com.google.cloud.storage.Blob} of the created object or {@code null} if
+	 * the bucket already exists or cannot be created
+	 */
+	public Bucket create() {
+		if (exists() || !this.createBucketIfNotExists) {
+			return null;
+		}
+
+		return this.storage.create(BucketInfo.newBuilder(this.bucketName).build());
 	}
 
 	@Override
 	public boolean exists() {
-		return resolve() != null;
+		return getGcsBucket() != null;
 	}
 
 	@Override
@@ -111,10 +122,15 @@ public class GoogleStorageResourceBucket implements WritableResource {
 
 	@Override
 	public Resource createRelative(String relativePath) throws IOException {
-		return new GoogleStorageResourceObject(this.storage,
+		GoogleStorageResourceObject resource = new GoogleStorageResourceObject(this.storage,
 				getURI().toString() + (relativePath.startsWith("/") ? "" : "/")
 						+ relativePath,
-				this.createBucketIfRequired);
+				this.createBucketIfNotExists);
+		if (!resource.exists()) {
+			resource.create();
+		}
+
+		return resource;
 	}
 
 	@Override
@@ -144,7 +160,11 @@ public class GoogleStorageResourceBucket implements WritableResource {
 				"Cannot get output stream for a storage bucket.");
 	}
 
-	private Bucket resolve() {
+	private Bucket getGcsBucket() {
 		return this.storage.get(this.bucketName);
+	}
+
+	public boolean isCreateBucketIfNotExists() {
+		return this.createBucketIfNotExists;
 	}
 }
--- a/spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceBucket.java
+++ b/spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceBucket.java
@@ -45,6 +45,14 @@ public class GoogleStorageResourceBucket implements WritableResource {
 
 	private final boolean createBucketIfNotExists;
 
+	/**
+	 * Constructs the resource representation of a GCS bucket.
+	 * @param storage the Google Cloud Storage client
+	 * @param bucketName the name of the bucket
+	 * @param createBucketIfNotExists determines bucket auto-creation when another operation that
+	 *                                depends on the existence of this bucket is executed (e.g.,
+	 *                                creating a file in the bucket)
+	 */
 	public GoogleStorageResourceBucket(Storage storage, String bucketName,
 			boolean createBucketIfNotExists) {
 		Assert.notNull(storage, "Storage object can not be null");
--- a/spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceBucket.java
+++ b/spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceBucket.java
@@ -28,7 +28,6 @@ import com.google.cloud.storage.Bucket;
 import com.google.cloud.storage.BucketInfo;
 import com.google.cloud.storage.Storage;
 
-import org.springframework.core.io.Resource;
 import org.springframework.core.io.WritableResource;
 import org.springframework.util.Assert;
 
@@ -59,18 +58,14 @@ public class GoogleStorageResourceBucket implements WritableResource {
 	/**
 	 * Creates the bucket that this {@link GoogleStorageResourceBucket} represents in Google Cloud
 	 * Storage.
+	 *
 	 * <p>If the bucket already exists in Google Cloud Storage, the
 	 * {@link com.google.cloud.storage.Bucket} object for that bucket is returned.
-	 * @return the {@link com.google.cloud.storage.Bucket} object for the bucket or {@code null} if
-	 * the bucket doesn't exist and cannot be created
+	 * @return the {@link com.google.cloud.storage.Bucket} object for the bucket
 	 */
 	public Bucket create() {
 		if (exists()) {
-			return getGcsBucket();
-		}
-
-		if (!this.createBucketIfNotExists) {
-			return null;
+			throw new IllegalStateException("The bucket " + this.bucketName + " already exists.");
 		}
 
 		return this.storage.create(BucketInfo.newBuilder(this.bucketName).build());
--- a/spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceObject.java
+++ b/spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceObject.java
@@ -34,7 +34,6 @@ import com.google.cloud.storage.Storage;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 
-import org.springframework.core.io.Resource;
 import org.springframework.core.io.WritableResource;
 import org.springframework.util.Assert;
 
@@ -152,20 +151,18 @@ public class GoogleStorageResourceObject implements WritableResource {
 	 * @throws IOException
 	 */
 	public Blob create() throws IOException {
-		BlobInfo objectInfo = BlobInfo.newBuilder(getBlobId()).build();
-		if (!exists() && !this.createBlobIfNotExists) {
-			return null;
+		GoogleStorageResourceBucket bucket = getResourceBucket();
+
+		if (!bucket.exists()) {
+			if (!bucket.isCreateBucketIfNotExists()) {
+				throw new IllegalStateException("The bucket " + bucket.getFilename()
+						+ " doesn't exist and couldn't be created");
+			}
+
+			bucket.create();
 		}
 
-		GoogleStorageResourceBucket bucket = getBucket();
-
-		if (!bucket.exists() && !bucket.isCreateBucketIfNotExists()) {
-			return null;
-		}
-
-		bucket.create();
-
-		return this.storage.create(objectInfo);
+		return this.storage.create(BlobInfo.newBuilder(getBlobId()).build());
 	}
 
 	private Blob throwExceptionForNullBlob(Blob blob) throws IOException {
@@ -252,8 +249,8 @@ public class GoogleStorageResourceObject implements WritableResource {
 		return Channels.newOutputStream(blob.writer());
 	}
 
-	GoogleStorageResourceBucket getBucket() throws IOException {
-		return new GoogleStorageResourceBucket(this.storage, getBlobId().getBucket(),
-				this.createBlobIfNotExists);
+	GoogleStorageResourceBucket getResourceBucket() throws IOException {
+		return new GoogleStorageResourceBucket(
+				this.storage, getBlobId().getBucket(), this.createBlobIfNotExists);
 	}
 }
--- a/spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceObject.java
+++ b/spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceObject.java
@@ -121,8 +121,11 @@ public class GoogleStorageResourceObject implements WritableResource {
 	}
 
 	/**
-	 * Creates a signed URL to an object if it exists. This method will fail if this storage resource
-	 * was not created using service account credentials.
+	 * Creates a signed URL to an object.
+	 * <p>If an object doesn't exist and {@code createBlobIfNotExists} is {@code true}, then the
+	 * object is created in GCS.
+	 * <p>This method will fail if this storage resource was not created using service account
+	 * credentials.
 	 * @param timeUnit the time unit used to determine how long the URL is valid.
 	 * @param timePeriods the number of periods to determine how long the URL is valid.
 	 * @return the URL if the object exists, and null if it does not.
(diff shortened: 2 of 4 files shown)


spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceObject.java
```diff
@@ -130,11 +130,18 @@ public Blob getGoogleStorageObject() throws IOException {
...
 	public URL createSignedUrl(TimeUnit timeUnit, long timePeriods) throws IOException {
 		Blob blob = this.getGoogleStorageObject();
+
 		if (blob == null) {
-			return null;
+			if (!this.createBlobIfNotExists) {
+				return null;
+			}
+
+			createBlob();
```
ChengyuanZhao: createBlob() returns the Blob, so I think you could just reuse  the blob reference from above.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceObject.java
```diff
@@ -121,20 +121,30 @@ public Blob getGoogleStorageObject() throws IOException {
...
```
joaoandremartins: A question here is whether this method should create the object if it doesn't exist, or if `GoogleStorageResourceBucket.createRelative()` should create the object explicitly.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceObject.java
joaoandremartins: I'd have to call `blob = this.getGoogleStorageObject()` so that the object metadata is updated.

I thought that was a little unsightly and went for this one instead, but let me know if you have other better alternatives.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceObject.java
ChengyuanZhao: I think we should let the functions inside the resource do the creating (as it is done now). 

The reason is the createBlob() function is private. If we were to make createBlob() public then that kinda sabotages the auto-create setting.

joaoandremartins: Currently, we only create files on `getOutputStream()`. This is equivalent to saying that we only create a file if you want to write to it. But a file doesn't have to be written to in order to be created.

So I think that this follows your rule: have the API methods create the file in GCS so we don't need to expose `createBlob()`?

gavinwilliams:
It would be quite nice to be able to inject my own blob, https://github.com/GoogleCloudPlatform/google-cloud-java/blob/f40fffce3235d8093751bfb9254211bfbcccadb2/google-cloud-storage/src/main/java/com/google/cloud/storage/StorageImpl.java#L501

joaoandremartins:
@gavinwilliams you could create a new `GoogleStorageResourceObject` object, specifying the location of your blob. Or use the `@Value` annotation instead of the constructor.

The location would be `gs://your-bucket//your-blob-name`.

If you already have a blob object and just want to get the signed URL for it, I would think it's maybe easier to just use the Storage API directly?

gavinwilliams:
@joaoandremartins is there a way to create a new GoogleStorageResourceObject? I can't see to find a way to create one without having to create a new `com.google.cloud.storage.Storage` object. I can't seem to find a way to access it within a `GoogleStorageResourceBucket` object created from using the `@Value` annotation.

joaoandremartins:
@gavinwilliams you can use `spring-cloud-gcp-starter-pubsub` along with the `@Value()` annotation, as the [Storage code sample](https://github.com/spring-cloud/spring-cloud-gcp/blob/master/spring-cloud-gcp-samples/spring-cloud-gcp-storage-resource-sample/src/main/java/com/example/WebController.java#L41) depicts.
You're able to control which blob `GoogleStorageResourceObject` will actually represent by passing `@Value` the URL of the blob you're interested in.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceObject.java
```diff
@@ -121,20 +121,32 @@ public Blob getGoogleStorageObject() throws IOException {
...
 	 * @param timePeriods the number of periods to determine how long the URL is valid.
 	 * @return the URL if the object exists, and null if it does not.
 	 * @throws IOException
 	 */
 	public URL createSignedUrl(TimeUnit timeUnit, long timePeriods) throws IOException {
 		Blob blob = this.getGoogleStorageObject();
+
 		if (blob == null) {
-			return null;
+			if (!this.createBlobIfNotExists) {
```
meltsufin: Why not just create `BlobInfo` without the blob using:
```java
BlobInfo.newBuilder(getBlobId()).build()
```

Does the creation of the signed URL actually required the `Blob` to exist in GCS?

I don't think `createSignedUrl` should create the blob. It seems like an unexpected side effect.

ChengyuanZhao: @meltsufin I was thinking originally that made sense, because how would you have a link to download a nonexistent file right? But the signed URLs allow you to write to the object, too, so an empty blob needs to be created if the signed URL is going to be written to.

ChengyuanZhao: I think you can send PUT requests to that URL, so it wouldn't be via Spring and our methods here.

meltsufin:
@gavinwilliams Note that the `Storage` object is in the Spring application context, and you can easily get a reference to it using `@Autowired Storage storage`.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceObject.java
joaoandremartins: The issue here is when you create a blob programatically, don't write to it and then try to get the signed URL for that empty blob. This operation should succeed but is currently failing.
I share your concerns that creating a blob is an unexpected side effect of a method called `createSignedUrl()`, so we'll probably need to expose the blob creation mechanism and call it from `createRelative`..

meltsufin: I just tested this. You can create a signed URL to a non-existent object, but when you try to do a PUT to it, you'll get an error like "Anonymous caller does not have storage.objects.create access to [file]".

I think we should allow creation of the signed URL for non-existent objects, but output a warning from the method when that happens.

ChengyuanZhao: Why not just let the auto-create setting govern the behavior? It's very reasonable for a user to assume a writeable signed URL is given if he has the auto-create setting on.

meltsufin: Imagine a sequence of calls like this:
```
file1 = new GoogleStorageResourceObject(storage, "file1", true)
file1.exists() // (false)
file1.createSignedUrl()
file1.exists() // (true)
```

You don't think this a bit surprising?

ChengyuanZhao: @meltsufin I would only find that surprising if I also found this surprising:

```
file1 = new GoogleStorageResourceObject(storage, "file1", true)
file1.exists() // (false)
file1.getOutputStream()
file1.exists() // (true)
```

If you keep in mind that a signed URL is as writeable as an output stream then neither should be surprising.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceBucket.java
```diff
@@ -52,16 +53,26 @@ public GoogleStorageResourceBucket(Storage storage, String bucketName,
...
+
+	/**
+	 * Creates the bucket that this {@link GoogleStorageResourceBucket} represents in Google Cloud
+	 * Storage.
+	 * @return the {@link com.google.cloud.storage.Blob} of the created object or {@code null} if
+	 * the bucket already exists or cannot be created
+	 */
+	public Bucket create() {
+		if (exists() || !this.createBucketIfNotExists) {
+			return null;
```
meltsufin: I'm not a fan of returning `null` here.
One alternative is to throw an `IllegalStateException`.
Another alternative is to silently return the existing bucket, and clarify in the javadoc that the method only creates the bucket if it doesn't exist.
I'm leaning more towards the latter.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceBucket.java
```diff
@@ -111,10 +122,15 @@ public long lastModified() {
 
 	@Override
 	public Resource createRelative(String relativePath) throws IOException {
```
meltsufin: Please change the return type to `GoogleStorageResourceObject`.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceObject.java
```diff
@@ -116,46 +121,51 @@ private BlobId getBlobId() throws IOException {
...
-
-		if (blob == null) {
-			if (!this.createBlobIfNotExists) {
-				return null;
-			}
+		Blob blob = this.getGcsBlob();
 
-			createBlob();
+		if (blob == null && LOGGER.isDebugEnabled()) {
+			LOGGER.debug("Creating signed URL for non-existing GCS object " + this.location);
```
meltsufin: Maybe higher logging level? I was thinking WARN.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceObject.java
```diff
@@ -173,25 +183,33 @@ public File getFile() throws IOException {
...
 	}
 
 	@Override
 	public long lastModified() throws IOException {
-		return throwExceptionForNullBlob(getGoogleStorageObject()).getUpdateTime();
+		return throwExceptionForNullBlob(getGcsBlob()).getUpdateTime();
 	}
 
 	@Override
 	public Resource createRelative(String relativePath) throws IOException {
```
meltsufin: Return `GoogleStorageResourceObject` please.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceObject.java
```diff
@@ -205,20 +223,32 @@ public String getDescription() {
...
 	@Override
 	public OutputStream getOutputStream() throws IOException {
-		Blob blob = getGoogleStorageObject();
-		if ((blob == null || !blob.exists()) && this.createBlobIfNotExists) {
-			blob = createBlob();
+		Blob blob = getGcsBlob();
+
+		if (blob == null || !blob.exists()) {
+			if (!this.createBlobIfNotExists) {
+				return null;
```
meltsufin: I would skip this whole if statement. Just let the final return statement throw an exception for non-existing blob.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceBucket.java
joaoandremartins: I don't think they're mutually exclusive. I actually think they're complementary.
If bucket already exists, return it, if not and it can't be created, do what? Return `null` or throw exception?
I'm more inclined towards returning `null` since the reason the bucket can't be created is within the user control (e.g., change `createBucketIfNotExists` to `true` and it will start working).

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceBucket.java
joaoandremartins: Regarding `createRelative()`, the behaviour is slightly different between `GoogleStorageResourceBucket` and `GoogleStorageResourceObject`.

Whereas the latter creates a sibling blob, the former doesn't create a sibling bucket but rather a blob within that bucket.

Not sure what to do, if anything, about this...

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceObject.java
joaoandremartins: I don't think we should do that because of #165. Unless I'm missing anything?

Thoughts?

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceBucket.java
meltsufin: OK, I kind of agree with you that maybe 'getOrCreate()` is not the behavior this method implies.
However, I think that returning `null` is a poor way of communicating a user error. Throwing `IllegalStateException` is more direct and easier to debug for the user.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceObject.java
meltsufin: The issue I'm pointing out here is the same as above. Communicating an error condition by returning `null`. It's better to throw an exception IMO, just like it was before your change.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceBucket.java
```diff
@@ -59,11 +59,17 @@ public GoogleStorageResourceBucket(Storage storage, String bucketName,
...
+	 * @return the {@link com.google.cloud.storage.Bucket} object for the bucket or {@code null} if
+	 * the bucket doesn't exist and cannot be created
 	 */
 	public Bucket create() {
-		if (exists() || !this.createBucketIfNotExists) {
+		if (exists()) {
+			return getGcsBucket();
+		}
+
+		if (!exists() && !this.createBucketIfNotExists) {
```
meltsufin: At this point you already know that `!exists()` is true.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceBucket.java
joaoandremartins: I think what I was trying to say is that I don't consider this case to be an "error" -- rather a choice by a user who specified auto-create as `false`.
Because it's not an error, but rather the desired behaviour, I don't think an exception should be raised.
In other places, e.g., if we're trying to write to a file and it doesn't exist, an exception should possibly be raised.

meltsufin: I was focusing on the first part of the condition `exists()`. In fact, this method should create the bucket whether `createBucketIfNotExists` is true or not. It shouldn't matter because it's the user's explicit intent to create the bucket.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceBucket.java
```diff
@@ -28,40 +28,60 @@
...
-		this.createBucketIfRequired = createBucketIfNotExists;
-	}
+		this.createBucketIfNotExists = createBucketIfNotExists;
+	}
+
+	/**
+	 * Creates the bucket that this {@link GoogleStorageResourceBucket} represents in Google Cloud
+	 * Storage.
+	 *
+	 * <p>If the bucket already exists in Google Cloud Storage, the
```
meltsufin: The javadoc is not accurate.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceBucket.java
```diff
@@ -28,40 +28,60 @@
...
+	 * @return the {@link com.google.cloud.storage.Bucket} object for the bucket
+	 */
+	public Bucket create() {
+		if (exists()) {
+			throw new IllegalStateException("The bucket " + this.bucketName + " already exists.");
+		}
 
-	public void createBucket() {
-		this.storage.create(BucketInfo.newBuilder(this.bucketName).build());
+		return this.storage.create(BucketInfo.newBuilder(this.bucketName).build());
```
meltsufin: If the bucket already exists, wouldn't this throw some kind of an exception? If so, we can probably remove the if statement above and avoid an extra existence check.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceObject.java
```diff
@@ -161,25 +188,33 @@ public File getFile() throws IOException {
...
+	public GoogleStorageResourceObject createRelative(String relativePath) throws IOException {
 		int lastSlashIndex = this.location.lastIndexOf("/");
 		String absolutePath = this.location.substring(0, lastSlashIndex + 1) + relativePath;
-		return new GoogleStorageResourceObject(this.storage, absolutePath, this.createBlobIfNotExists);
+
+		GoogleStorageResourceObject blobResource = new GoogleStorageResourceObject(
+				this.storage, absolutePath, this.createBlobIfNotExists);
+
+		if (!blobResource.exists()) {
+			blobResource.create();
```
ChengyuanZhao: You're giving `this.createBlobIfNotExists`, which may be false, to the relative blob, but you always create it in this line, so it's like that setting no longer matters. Maybe you should just pass `true` in that constructor on the previous line?

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceBucket.java
```diff
@@ -110,11 +125,16 @@ public long lastModified() {
...
-	public Resource createRelative(String relativePath) throws IOException {
-		return new GoogleStorageResourceObject(this.storage,
+	public GoogleStorageResourceObject createRelative(String relativePath) throws IOException {
+		GoogleStorageResourceObject resource = new GoogleStorageResourceObject(this.storage,
 				getURI().toString() + (relativePath.startsWith("/") ? "" : "/")
 						+ relativePath,
-				this.createBucketIfRequired);
+				this.createBucketIfNotExists);
+		if (!resource.exists()) {
+			resource.create();
```
ChengyuanZhao: Basically the same comment I had on the other class: perhaps just pass `true` in the constructor for the relative blob? It looks like `resource.create()` will either result in a blob that is created and exists, or will throw an exception, so passing `this.createBucketIfNotExists` in the constructor doesn't have any meaning anymore.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceObject.java
joaoandremartins: `createBlobIfNotExists` determines if an object should be **auto**created, not if it should be created.

In other words, `createBlobIfNotExists` should only be taken into account for the auto-creation in `getOutputStream()`.

When users call `createRelative()`, they're saying they want to create an object, so that's what we do.

joaoandremartins:
PTAL

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceBucket.java
```diff
@@ -110,11 +125,16 @@ public long lastModified() {
 	}
 
 	@Override
-	public Resource createRelative(String relativePath) throws IOException {
-		return new GoogleStorageResourceObject(this.storage,
+	public GoogleStorageResourceObject createRelative(String relativePath) throws IOException {
```
meltsufin: This method definitely needs a javadoc even though it overrides because the behavior is pretty nuanced.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceBucket.java
```diff
@@ -110,11 +125,16 @@ public long lastModified() {
...
+		GoogleStorageResourceObject resource = new GoogleStorageResourceObject(this.storage,
 				getURI().toString() + (relativePath.startsWith("/") ? "" : "/")
 						+ relativePath,
-				this.createBucketIfRequired);
+				this.createBucketIfNotExists);
+		if (!resource.exists()) {
```
meltsufin: I'm not sure that we should be creating the file if it doesn't exist. We don't do anything like this in the `GoogleStorageResourceObject` constructor. So, why would this be any different? I would just return the `resource` handle and let the user decide if they want to attempt creating it.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceBucket.java
```diff
@@ -110,11 +125,16 @@ public long lastModified() {
...
 
 	@Override
-	public Resource createRelative(String relativePath) throws IOException {
-		return new GoogleStorageResourceObject(this.storage,
+	public GoogleStorageResourceObject createRelative(String relativePath) throws IOException {
+		GoogleStorageResourceObject resource = new GoogleStorageResourceObject(this.storage,
 				getURI().toString() + (relativePath.startsWith("/") ? "" : "/")
 						+ relativePath,
-				this.createBucketIfRequired);
+				this.createBucketIfNotExists);
```
meltsufin: This doesn't quiet make sense on the surface. You might want to let the user pass `createBlobIfNotExists` to `GoogleStorageResourceBucket`, or at least clearly document that auto-creation setting is inherited by the `GoogleStorageResourceObject`.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceObject.java
```diff
@@ -161,25 +188,33 @@ public File getFile() throws IOException {
...
 
 	@Override
 	public long lastModified() throws IOException {
-		return throwExceptionForNullBlob(getGoogleStorageObject()).getUpdateTime();
+		return throwExceptionForNullBlob(getGcsBlob()).getUpdateTime();
 	}
 
 	@Override
-	public Resource createRelative(String relativePath) throws IOException {
+	public GoogleStorageResourceObject createRelative(String relativePath) throws IOException {
```
meltsufin: Please add a javadoc.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceObject.java
```diff
@@ -161,25 +188,33 @@ public File getFile() throws IOException {
...
 		int lastSlashIndex = this.location.lastIndexOf("/");
 		String absolutePath = this.location.substring(0, lastSlashIndex + 1) + relativePath;
-		return new GoogleStorageResourceObject(this.storage, absolutePath, this.createBlobIfNotExists);
+
+		GoogleStorageResourceObject blobResource = new GoogleStorageResourceObject(
+				this.storage, absolutePath, this.createBlobIfNotExists);
```
meltsufin: The javadoc for the method should mention that `createBlobIfNotExists` is passed up the bucket.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceBucket.java
joaoandremartins: That's not in the scope of this PR.
Please file a separate issue to fix that, if you feel it's necessary.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceBucket.java
joaoandremartins: It's the one thing this method must do and we're not doing currently: creating a file.
We've seen use cases that are broken if we don't create the file here. e.g., the one reported in #411

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceObject.java
joaoandremartins: That would be adding complexity.
The property is called `autoCreateFiles`. I think it conveys that files and buckets are automatically created when needed.
I'm good with the naming as-is, but if you want to rename to something that's more general and encompasses files and buckets, sure.

joaoandremartins:
Which complexity are you referring to? I'm afraid I don't see it.
We need file auto-creation so that the API is easier on users. e.g., we don't need to have things like `gcsFile.create().getOutputStream().write()`

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceBucket.java
meltsufin: Can we at least document the current behavior in the javadoc?

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceBucket.java
meltsufin: #411 would be addressed by simply making `create()` public.

I'm not saying it's wrong, I'm just saying that it's inconsistent with the `GoogleStorageResourceObject` constructor. Why not create the file in the constructor, if doesn't exist? How is this different from `createRelative` from the bucket?

meltsufin:
The original intent was to support auto-creation on `getOutputStream()` only, but now it has spread to many more places in the code.

joaoandremartins:
The only other place it spread to is bucket auto-creation. The motivation for doing that was similar to `getOutputStream()`

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceBucket.java
joaoandremartins: `createSignedUrl()` would still not allow to create signed URLs for files that don't exist yet, so making `create()` public isn't enough.

Constructors shouldn't contain such complex logic as creating a file in a file system. The constructor Javadoc points out that the object is a "representation" of a file on GCS.
When I call `createRelative()`, I do expect the file to actually be created in GCS, not just in memory.

It isn't very much different from the bucket's `createRelative()`, except that you only need to have the bucket to call the bucket's, and if you have the file, you can call from it too.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceBucket.java
joaoandremartins: Done.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceObject.java
meltsufin: OK fair enough. My opinion was always that buckets should never be auto-created. I'll create an issue for that.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceObject.java
```diff
@@ -116,33 +128,48 @@ private BlobId getBlobId() throws IOException {
...
 	 */
-	public Blob getGoogleStorageObject() throws IOException {
+	public Blob getGcsBlob() throws IOException {
 		return this.storage.get(getBlobId());
 	}
 
 	/**
-	 * Creates a signed URL to an object if it exists. This method will fail if this storage resource
-	 * was not created using service account credentials.
+	 * Creates a signed URL to an object, if it exists. This method will fail if this storage
```
meltsufin: It will create the signed URL regardless of whether it exists.

PR #419, opened 2018-02-06
Fixes issue with SI dependencies in refdoc

joaoandremartins:


joaoandremartins:
Also @sobychacko, any chance we can re-publish the M2 docs, and only the docs, after this change goes in?
Thanks!

Issue #292, opened 2018-01-05
Add remaining unit and integrations tests to Stream binder


joaoandremartins:
#145 was submitted without tests because we need to get the Pub/Sub emulator running on Travis.

joaoandremartins:
[These tests](https://github.com/spring-cloud/spring-cloud-gcp/pull/145#pullrequestreview-64869492) are a good starting point.

PR #423, opened 2018-02-07
Evaluate consumer group settings in the Pub/Sub binder

### LINKED ISSUE #422: evaluate the consumer group settings in the pub-sub binder
For the spring binder you can configure a group so you can have multiple subscriptions for one topic.

This feature does not work currently, because the group name is not taken into consideration.

Andy2003:
Fixes #422.

### DIFF
--- a/spring-cloud-gcp-pubsub-stream-binder/src/main/java/org/springframework/cloud/gcp/stream/binder/pubsub/PubSubMessageChannelBinder.java
+++ b/spring-cloud-gcp-pubsub-stream-binder/src/main/java/org/springframework/cloud/gcp/stream/binder/pubsub/PubSubMessageChannelBinder.java
@@ -58,9 +58,6 @@ public class PubSubMessageChannelBinder
 	protected MessageHandler createProducerMessageHandler(ProducerDestination destination,
 			ExtendedProducerProperties<PubSubProducerProperties> producerProperties,
 			MessageChannel errorChannel) {
-
-		this.provisioningProvider.provisionProducerDestination(destination.getName(), producerProperties);
-
 		return new PubSubMessageHandler(this.pubSubTemplate, destination.getName());
 	}
 
--- a/spring-cloud-gcp-pubsub-stream-binder/src/main/java/org/springframework/cloud/gcp/stream/binder/pubsub/PubSubMessageChannelBinder.java
+++ b/spring-cloud-gcp-pubsub-stream-binder/src/main/java/org/springframework/cloud/gcp/stream/binder/pubsub/PubSubMessageChannelBinder.java
@@ -67,10 +67,7 @@ public class PubSubMessageChannelBinder
 	@Override
 	protected MessageProducer createConsumerEndpoint(ConsumerDestination destination, String group,
 			ExtendedConsumerProperties<PubSubConsumerProperties> properties) {
-
-		String subscription = this.provisioningProvider.createSubscription(destination.getName(), group, properties);
-
-		return new PubSubInboundChannelAdapter(this.pubSubTemplate, subscription);
+		return new PubSubInboundChannelAdapter(this.pubSubTemplate, destination.getName());
 	}
 
 	@Override
--- a/spring-cloud-gcp-pubsub-stream-binder/src/main/java/org/springframework/cloud/gcp/stream/binder/pubsub/PubSubMessageChannelBinder.java
+++ b/spring-cloud-gcp-pubsub-stream-binder/src/main/java/org/springframework/cloud/gcp/stream/binder/pubsub/PubSubMessageChannelBinder.java
@@ -57,25 +57,20 @@ public class PubSubMessageChannelBinder
 	@Override
 	protected MessageHandler createProducerMessageHandler(ProducerDestination destination,
 			ExtendedProducerProperties<PubSubProducerProperties> producerProperties,
-			MessageChannel errorChannel) throws Exception {
+			MessageChannel errorChannel) {
 
-		this.provisioningProvider.provisionProducerDestination(
-				destination.getName(), producerProperties);
+		this.provisioningProvider.provisionProducerDestination(destination.getName(), producerProperties);
 
 		return new PubSubMessageHandler(this.pubSubTemplate, destination.getName());
 	}
 
 	@Override
 	protected MessageProducer createConsumerEndpoint(ConsumerDestination destination, String group,
-			ExtendedConsumerProperties<PubSubConsumerProperties> properties) throws Exception {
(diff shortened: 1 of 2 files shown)


spring-cloud-gcp-pubsub-stream-binder/src/main/java/org/springframework/cloud/gcp/stream/binder/pubsub/provisioning/PubSubChannelProvisioner.java
```diff
@@ -54,19 +54,25 @@ public ProducerDestination provisionProducerDestination(String name,
 	public ConsumerDestination provisionConsumerDestination(String name, String group,
 			ExtendedConsumerProperties<PubSubConsumerProperties> properties)
 			throws ProvisioningException {
-		if (this.pubSubAdmin.getSubscription(name) == null) {
+		return new PubSubConsumerDestination(name);
+	}
+
+	public String createSubscription(String name, String group,
```
joaoandremartins: Why the extra method?
Shouldn't this work be done in `provisionProducerDestination` like before, but considering the group name instead of just `name`?

Andy2003: before this change `provisionConsumerDestination` was called 2 times:

1. `org.springframework.cloud.stream.binder.AbstractMessageChannelBinder#doBindConsumer` 
2. `org.springframework.cloud.gcp.stream.binder.pubsub.PubSubMessageChannelBinder#createConsumerEndpoint`

This resulted in wrong destination names.

Andy2003: I Removed the second invocation so that the PubSubConsumerDestination now directly identifies the subscription to be used

spring-cloud-gcp-pubsub-stream-binder/src/main/java/org/springframework/cloud/gcp/stream/binder/pubsub/PubSubMessageChannelBinder.java
```diff
@@ -57,24 +57,16 @@ public PubSubMessageChannelBinder(String[] headersToEmbed,
 	@Override
 	protected MessageHandler createProducerMessageHandler(ProducerDestination destination,
 			ExtendedProducerProperties<PubSubProducerProperties> producerProperties,
-			MessageChannel errorChannel) throws Exception {
```
joaoandremartins: I think you can revert changes to this file since they're just cosmetic?

spring-cloud-gcp-pubsub-stream-binder/src/main/java/org/springframework/cloud/gcp/stream/binder/pubsub/PubSubMessageChannelBinder.java
```diff
@@ -57,24 +57,16 @@ public PubSubMessageChannelBinder(String[] headersToEmbed,
...
 	}
 
 	@Override
 	protected MessageProducer createConsumerEndpoint(ConsumerDestination destination, String group,
-			ExtendedConsumerProperties<PubSubConsumerProperties> properties) throws Exception {
-
-		String consumerName = group == null
-				? destination.getName() : destination.getName() + "-" + group;
-
-		this.provisioningProvider.provisionConsumerDestination(
```
joaoandremartins: Why are we removing this?
We still want to auto-create subscriptions, right?

spring-cloud-gcp-pubsub-stream-binder/src/main/java/org/springframework/cloud/gcp/stream/binder/pubsub/provisioning/PubSubChannelProvisioner.java
joaoandremartins: Not sure if I understand?
We were calling `provisionProducerDestination` once, to automatically create the topic, and `provisionConsumerDestination` once, to create the subscription.

It sounds to me we still need both of those calls.

joaoandremartins: I realised that `provisionProducerDestination` is also called by `AbstractMessageChannelBinder.doBindProducer()`.

Should we also remove that call?

spring-cloud-gcp-pubsub-stream-binder/src/main/java/org/springframework/cloud/gcp/stream/binder/pubsub/provisioning/PubSubChannelProvisioner.java
```diff
@@ -54,19 +54,20 @@ public ProducerDestination provisionProducerDestination(String name,
...
+				this.pubSubAdmin.createSubscription(subscription, name);
 			}
 			else {
-				throw new ProvisioningException("Unexisting '" + name + "' subscription.");
+				throw new ProvisioningException("Unexisting '" + subscription + "' subscription.");
 			}
 		}
-
-		return new PubSubConsumerDestination(name);
+		return  new PubSubConsumerDestination(subscription);
```
joaoandremartins: Nit: 2 spaces between `return` and `new`.

spring-cloud-gcp-pubsub-stream-binder/src/main/java/org/springframework/cloud/gcp/stream/binder/pubsub/PubSubMessageChannelBinder.java
```diff
@@ -57,24 +57,16 @@ public PubSubMessageChannelBinder(String[] headersToEmbed,
 	@Override
 	protected MessageHandler createProducerMessageHandler(ProducerDestination destination,
 			ExtendedProducerProperties<PubSubProducerProperties> producerProperties,
-			MessageChannel errorChannel) throws Exception {
+			MessageChannel errorChannel) {
 
-		this.provisioningProvider.provisionProducerDestination(
-				destination.getName(), producerProperties);
+		this.provisioningProvider.provisionProducerDestination(destination.getName(), producerProperties);
```
joaoandremartins: This is also being called twice.
Should we remove it as part of this PR?

joaoandremartins:
Adding tests to the binder is going to be a high priority, as part of #292.

PR #427, opened 2018-02-08
Spanner mutation implementations

ChengyuanZhao:
Added implementations for mutation factory , mapper, and unit tests


ChengyuanZhao:
@meltsufin @balopat PTAL

spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/core/SpannerObjectMapperImplTests.java
```diff
@@ -0,0 +1,226 @@
...
+
+	private SpannerObjectMapper objectMapper;
+
+	@Before
+	public void setUp() {
+		this.objectMapper = new SpannerObjectMapperImpl(new SpannerMappingContext());
+	}
+
+	@Test
+	public void writeTest() throws ClassNotFoundException {
```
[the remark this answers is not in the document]
ChengyuanZhao: This test is ensuring that the correctly typed "to" methods are called. This is vital, because that's where our responsibility leaves and it's up to the Spanner client lib. The key point of the write-mapping is to make sure the correctly typed functions are mapped.  I will rename the fields.

spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/core/SpannerMutationFactoryImplTests.java
```diff
@@ -0,0 +1,155 @@
...
+		assertEquals(Op.UPDATE, mutation.getOperation());
+	}
+
+	@Test
+	public void upsertTest() {
+		Mutation mutation = this.spannerMutationFactory.upsert(new TestEntity());
+		assertEquals("custom_test_table", mutation.getTable());
+		assertEquals(Op.INSERT_OR_UPDATE, mutation.getOperation());
+	}
+
```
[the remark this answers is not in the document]
ChengyuanZhao: We don't have a public Replace method as part of the interface, so I removed the REPLACE option from the private writeBuilder function.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/mapping/SpannerObjectMapperImpl.java
```diff
@@ -0,0 +1,183 @@
...
+		ArrayList<T> result = new ArrayList<>();
+		while (resultSet.next()) {
+			result.add(read(entityClass, resultSet.getCurrentRowAsStruct()));
+		}
+		return Collections.unmodifiableList(result);
+	}
+
+	@Override
+	public <R> R read(Class<R> type, Struct source) {
+		R object = BeanUtils.instantiateClass(type);
```
balopat: why are we using a BeanUtils instantiation? A simple no-arg constructor can be assumed for entities.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/mapping/SpannerObjectMapperImpl.java
```diff
@@ -0,0 +1,183 @@
...
+				}
+
+				Class retType = method.getReturnType();
+				if (retType.isAssignableFrom(property.getType())) {
+					try {
+						accessor.setProperty(property, method.invoke(source, name));
+						valueSet = true;
+						break;
+					}
+					catch (IllegalAccessException e) {
```
balopat: nit: collapse catch blocks

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/mapping/SpannerObjectMapperImpl.java
```diff
@@ -0,0 +1,183 @@
...
+					catch (InvocationTargetException e) {
+						throw new UnsupportedOperationException(
+								"Could not set value for read.", e);
+					}
+				}
+			}
+
+			if (!valueSet) {
+				throw new UnsupportedOperationException(
+						"No compatible types in object for reading column: " + name);
```
balopat: nit: "object" is confusing. I'd say: "None of the fields of class " + type + " can be used for reading column " + name

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/mapping/SpannerObjectMapperImpl.java
```diff
@@ -0,0 +1,183 @@
...
+						accessor.setProperty(property, method.invoke(source, name));
+						valueSet = true;
+						break;
+					}
+					catch (IllegalAccessException e) {
+						throw new UnsupportedOperationException(
+								"Could not set value for read.", e);
+					}
+					catch (InvocationTargetException e) {
+						throw new UnsupportedOperationException(
```
balopat: nit: include the type and the current method of the source in the error message

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerMutationFactoryImpl.java
```diff
@@ -0,0 +1,143 @@
...
+	}
+
+	@Override
+	public Mutation update(Object object) {
+		return saveObject(Op.UPDATE, object);
+	}
+
+	@Override
+	public <T> Mutation delete(Class<T> entityClass, Iterable<? extends T> entities) {
+		return deleteKeySetFromTable(entityClass, (persistentEntity) -> {
```
ChengyuanZhao: make a private method out of the lambda function here

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerMutationFactoryImpl.java
```diff
@@ -0,0 +1,143 @@
...
+		return deleteKeySetFromTable(entityClass, (unused) -> keys);
+	}
+
+	@Override
+	public Mutation delete(Class entityClass, Key key) {
+		return delete(entityClass, KeySet.singleKey(key));
+	}
+
+	private Mutation deleteKeySetFromTable(Class entityClass,
+			Function<SpannerPersistentEntity, KeySet> keyFunction) {
```
ChengyuanZhao: get rid of this function and the lambda function actually

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
```diff
@@ -171,4 +160,13 @@ public long count(Class entityClass) {
 		resultSet.next();
 		return resultSet.getLong(0);
 	}
+
+	private <T, U> void applyMutationOfTwoArg(BiFunction<T, U, Mutation> function, T arg1,
+			U arg2) {
+		this.databaseClient.write(Arrays.asList(function.apply(arg1, arg2)));
```
ChengyuanZhao: rename to applyMutationWithClass

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/mapping/SpannerObjectMapper.java
```diff
@@ -42,5 +42,4 @@
 	 * @return A list of objects.
 	 */
 	<T> List<T> mapToUnmodifiableList(ResultSet resultSet, Class<T> entityClass);
```
ChengyuanZhao: not necessarily unmodifiable ? check other examples

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/mapping/SpannerObjectMapperImpl.java
ChengyuanZhao: will change to use standard Java object creation

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/mapping/SpannerObjectMapperImpl.java
ChengyuanZhao: perhaps need to reverse the isAssignableFrom here

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/mapping/SpannerObjectMapperImpl.java
ChengyuanZhao: Check what happens if the columns in struct don't match what's in entity and vice versa for write.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/mapping/SpannerObjectMapper.java
ChengyuanZhao: @balopat Hey good morning, I made changes we discussed in latest commit. PTAL !

ChengyuanZhao:
@balopat I added the annotation for those inner types in Iterables. I also added caching of the get-methods for reads, but not the writes. The reason is that we have a findAll method in our template , but the insert/update/upsert methods all just take one object at a time.

ChengyuanZhao:
@balopat caching for read and write methods and handling the Iterable added! PTAL

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/mapping/SpannerObjectMapperImpl.java
```diff
@@ -0,0 +1,336 @@
...
+	// methods for
+	// properties with inner types, such as Iterable.
+	private final Map<Class, Method> propertyWriteMethodMapping = new HashMap<>();
+
+	private static final Map<Class, BiConsumer<ValueBinder<WriteBuilder>, Iterable>>
+			writeBuilderIterableMapping =
+			// The casting of each biconsumer below is needed for Java 8 compilation, but not Java 9.
+			new ImmutableMap.Builder<Class, BiConsumer<ValueBinder<WriteBuilder>, Iterable>>()
+			.put(Boolean.class,
+					(BiConsumer<ValueBinder<WriteBuilder>, Iterable>) (binder,
```
balopat: This compilation error is so strange...
works for me when I add the elements one by one using the builder reference:  

```
	private static final Map<Class, BiConsumer<ValueBinder<WriteBuilder>, Iterable>>
			writeBuilderIterableMapping =
			// The casting of each biconsumer below is needed for Java 8 compilation, but not Java 9.
			createWriteBuilderIterableMapping();

	private static ImmutableMap<Class, BiConsumer<ValueBinder<WriteBuilder>, Iterable>> createWriteBuilderIterableMapping() {
		Builder<Class, BiConsumer<ValueBinder<WriteBuilder>, Iterable>> builder = new Builder<>();

		builder.put(Date.class, ValueBinder::toDateArray);
		builder.put(Boolean.class, ValueBinder::toBoolArray);
		builder.put(Long.class, ValueBinder::toInt64Array);
		builder.put(String.class, ValueBinder::toStringArray);
		builder.put(Double.class, ValueBinder::toFloat64Array);
		builder.put(Timestamp.class,ValueBinder::toTimestampArray);
		builder.put(ByteArray.class, ValueBinder::toBytesArray);

		return builder.build();
	}
```

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/mapping/SpannerObjectMapperImpl.java
```diff
@@ -0,0 +1,336 @@
...
+	private static final Map<Class, BiFunction<Struct, String, List>>
+			readIterableMapping =
+			new ImmutableMap.Builder<Class, BiFunction<Struct, String, List>>()
+			.put(Boolean.class, (struct, colName) -> struct.getBooleanList(colName))
+			.put(Long.class, (struct, colName) -> struct.getLongList(colName))
+			.put(String.class, (struct, colName) -> struct.getStringList(colName))
+			.put(Double.class, (struct, colName) -> struct.getDoubleList(colName))
+			.put(Timestamp.class, (struct, colName) -> struct.getTimestampList(colName))
+			.put(Date.class, (struct, colName) -> struct.getDateList(colName))
+			.put(ByteArray.class, (struct, colName) -> struct.getBytesList(colName))
```
balopat: we can just use method reference here too

ChengyuanZhao: Changes made!

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/mapping/SpannerObjectMapperImpl.java
```diff
@@ -76,7 +74,7 @@
 		builder.put(Timestamp.class, ValueBinder::toTimestampArray);
 		builder.put(ByteArray.class, ValueBinder::toBytesArray);
 
-		writeBuilderIterableMapping = builder.build();
+		iterablePropertyType2ToMethodMap = builder.build();
```
ChengyuanZhao: Is that "2To" redundant ? 

Also, it looks like this commit broke the build !

balopat: well, I'm mapping to a "to method"...it's awkward. Without the 2 it reads like iterablePopertyType to method map. I'll check on the build.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/mapping/SpannerObjectMapperImpl.java
```diff
@@ -0,0 +1,360 @@
...
+				valueSet = attemptReadIterableValue(property, source, columnName, accessor);
+			}
+			else {
+				Method getMethod = this.propertyReadMethodMapping.get(propType);
+				if (getMethod == null) {
+					getMethod = findReadMethod(source, propType);
+				}
+				if (getMethod != null) {
+					try {
+						accessor.setProperty(property, getMethod.invoke(source, columnName));
```
balopat: We'll need to add checking for reading null columns (on this `getMethod.invoke`), as we have the [StructReader interface behave like  this](https://googlecloudplatform.github.io/google-cloud-java/0.25.0/apidocs/com/google/cloud/spanner/StructReader.html): 
"The getTypeName() methods should only be called for non-NULL values, otherwise a NullPointerException is raised; isNull(int)/isNull(String) can be used to test for NULL-ness if necessary."

ChengyuanZhao: We have that right ? 

```
if (source.isNull(columnName)) {
			continue;
		}
```

balopat:
@ChengyuanZhao I'm wondering if we can remove bd1a7767ce32135553608b75fd993d80b7c7242a for now and raise a separate PR for it? Otherwise it's looking good, and we could merge it.

PR #428, opened 2018-02-12
Allows options to be passed to signUrl()

joaoandremartins:


joaoandremartins:
@gavinwilliams FYI

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResourceObject.java
```diff
@@ -137,18 +137,20 @@ public Blob getGcsBlob() throws IOException {
 	 * resource was not created using service account credentials.
 	 * @param timeUnit the time unit used to determine how long the URL is valid.
 	 * @param timePeriods the number of periods to determine how long the URL is valid.
+	 * @param options specifies additional options for signing URLs
```
meltsufin: Could be `null`? Maybe create two overloaded methods?

joaoandremartins: The advantage of varargs is that we don't need to overload the method to have both flavours.
I added an assertion for `null` options.

PR #431, opened 2018-02-12
Integration tests for the Pub/Sub Stream binder

### LINKED ISSUE #292: Add remaining unit and integrations tests to Stream binder

joaoandremartins:
Fixes #292


joaoandremartins:
@Andy2003 FYI

joaoandremartins:
Tests will only be enabled if the `GCP_PUBSUB_EMULATOR` envvar is set. Feel free to suggest an alternative.
For some reason, tests are failing running through Maven, but successful if I run them through JUnit.

spring-cloud-gcp-pubsub-stream-binder/src/test/java/org/springframework/cloud/gcp/stream/binder/pubsub/PubSubEmulator.java
```diff
@@ -0,0 +1,69 @@
...
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.junit.rules.ExternalResource;
+
+/**
+ * JUnit rule that starts a Pub/Sub emulator.
+ *
+ * @author João André Martins
+ */
+public class PubSubEmulator extends ExternalResource {
```
artembilan: There is an `AbstractExternalResourceTestSupport` to consider here for extension.

spring-cloud-gcp-pubsub-stream-binder/src/test/java/org/springframework/cloud/gcp/stream/binder/pubsub/PubSubMessageChannelBinderTest.java
```diff
@@ -0,0 +1,71 @@
...
+import org.springframework.cloud.stream.binder.ExtendedConsumerProperties;
+import org.springframework.cloud.stream.binder.ExtendedProducerProperties;
+import org.springframework.cloud.stream.binder.Spy;
+
+import static org.assertj.core.api.Assumptions.assumeThat;
+
+/**
+ * @author João André Martins
+ */
+public class PubSubMessageChannelBinderTest extends AbstractBinderTests<PubSubTestBinder,
```
artembilan: I would be happy to have a Checkstyle rule on the matter, but that isn't possible.

Please, don't forget that test classes names must finish with `Tests` suffix.

spring-cloud-gcp-pubsub-stream-binder/src/test/java/org/springframework/cloud/gcp/stream/binder/pubsub/PubSubTestBinder.java
```diff
@@ -0,0 +1,109 @@
...
+public class PubSubTestBinder extends AbstractTestBinder<PubSubMessageChannelBinder,
+		ExtendedConsumerProperties<PubSubConsumerProperties>,
+		ExtendedProducerProperties<PubSubProducerProperties>> {
+
+	public PubSubTestBinder() {
+		GcpProjectIdProvider projectIdProvider = () -> "sodium-gateway-790";
+
+		// Transport channel provider so that test binder talks to emulator.
+		ManagedChannel channel = ManagedChannelBuilder
+				.forTarget("localhost:8085")
```
artembilan: Looks fishy. I'm pretty sure that `AbstractExternalResourceTestSupport` will bring you some clues and hooks how to avoid magic strings, especially for ports. Yes, I would like to see port as random value. Otherwise it is definitely going to be a vulnerability on target environments and developer machines.

BTW, does that emulator work on Windows?

Thanks

spring-cloud-gcp-pubsub/pom.xml
```diff
@@ -25,6 +25,10 @@
 			<groupId>org.springframework.cloud</groupId>
 			<artifactId>spring-cloud-gcp-core</artifactId>
 		</dependency>
+		<dependency>
+			<groupId>com.fasterxml.jackson.core</groupId>
+			<artifactId>jackson-databind</artifactId>
```
artembilan: Is it really so mandatory dependency ?

Why do we need it at all?
Is there some comment in the commit message explaining such a change?
Thanks

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapter.java
```diff
@@ -48,15 +46,9 @@
 
 	private AckMode ackMode = AckMode.AUTO;
 
-	private MessageConverter messageConverter;
```
artembilan: That's not good change.
The `PubSubInboundChannelAdapter` hasn't been designed to only support Binder.

Many people use Channel Adapters as is, without any Spring Cloud Stream magic.
So, such a convert was a good functionality in this component and you just remove it without any explanation.

If you would like support `byte[]` as a `payload` to emit, would be better to improve the logic in this component like `PathThroughConverter` or even better to let to support `null` for the converter. 

But I'm against such a critical change that narrows a component functionality for nothing.

Thanks for understanding.

(Again no any commit comments on the matter)

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/outbound/PubSubMessageHandler.java
```diff
@@ -46,7 +47,7 @@
  *
  * @author João André Martins
  */
-public class PubSubMessageHandler extends AbstractMessageHandler {
+public class PubSubMessageHandler extends AbstractMessageHandler implements Lifecycle {
```
artembilan: What is this new functionality here about?

I don't see any logic in the `start()`, neither `stop()`.

So, what is the point of void implementation?

joaoandremartins: [This test](https://github.com/spring-cloud/spring-cloud-stream/blob/master/spring-cloud-stream-binder-test/src/main/java/org/springframework/cloud/stream/binder/AbstractBinderTests.java#L129) seems to require this change.

spring-cloud-gcp-pubsub-stream-binder/src/test/java/org/springframework/cloud/gcp/stream/binder/pubsub/PubSubEmulator.java
joaoandremartins: Looking at the [source code](https://github.com/spring-cloud/spring-cloud-stream/blob/master/spring-cloud-stream-test-support-internal/src/main/java/org/springframework/cloud/stream/test/junit/AbstractExternalResourceTestSupport.java), it's not obvious what `AbstractExternalResourceTestSupport` can do for us that we're not already doing here.
What am I missing?

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/outbound/PubSubMessageHandler.java
artembilan: Then that test is wrong. That is not our code responsibility to support void functionality.
Or raise an issue against SCSt to improve that test or just leave it empty like I did in the Kinesis Binder:
```
@Test
@Override
public void testClean() {
}
```

If we don't provide such a functionality what is the point to support it?

spring-cloud-gcp-pubsub-stream-binder/src/test/java/org/springframework/cloud/gcp/stream/binder/pubsub/PubSubEmulator.java
artembilan: Well, at least that helped me to develop similar tests for Kinesis Binder: https://github.com/spring-cloud/spring-cloud-stream-binder-aws-kinesis/blob/master/spring-cloud-stream-binder-kinesis/src/test/java/org/springframework/cloud/stream/binder/kinesis/LocalKinesisResource.java.

The same way we go in the Kafka Binder: https://github.com/spring-cloud/spring-cloud-stream-binder-rabbit/blob/master/spring-cloud-stream-binder-rabbit-test-support/src/main/java/org/springframework/cloud/stream/binder/test/junit/rabbit/RabbitTestSupport.java.

Your problem here that you don't expose resource but rely on the `localhost:8085`.
:shrug:

Also we should somehow ensure my environment on Windows without any Cloud SDK.

Thanks

joaoandremartins: These tests don't run by default.
The `GCP_PUBSUB_EMULATOR` envvar needs to be passed to the process so they are triggered.

artembilan: OK! So, why then they fail in Maven? And I guess they are going to fail for me on Windows

joaoandremartins: I need to look into the Maven issue.
Getting the tests to run on Windows is part of this PR scope. I'll add that in a separate commit.

spring-cloud-gcp-pubsub-stream-binder/src/test/java/org/springframework/cloud/gcp/stream/binder/pubsub/PubSubTestBinder.java
joaoandremartins: Changed the tests to run on a random port from 10000 to 20000.

There is a Windows emulator, but this test isn't using it yet.

spring-cloud-gcp-pubsub-stream-binder/src/test/java/org/springframework/cloud/gcp/stream/binder/pubsub/PubSubEmulator.java
artembilan: Let's never mind about Windows for now!
That may be addressed in the separate issue.
My concern right now is to run project either way and don't fail.
Just to ignore tests on Windows is enough for me.

Thanks

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapter.java
joaoandremartins: Indeed, the motivation for this was that the previous converter couldn't convert `application/json` messages coming from the binder.
I tried to also add a `MappingJackson2MessageConverter` into a `CompositeMessageConverter`, but apparently SCSt re-applies internal message conversion after `sendMessage()`, which it couldn't do after the message had been re-converted by our converter here.

That's when I saw it made sense to just pass through the `byte[]` that we receive and just send it as-is to the output channel.

The user could then parse the `byte[]` on their end to whatever they want. I agree this is far from ideal, but the current functionality is also not very far from this... We basically consider every payload to be a string, which is not right.

I couldn't find anything on `PathThroughConverter`.

artembilan: That's why I suggest and idea to let to path the `converter` into the `PubSubInboundChannelAdapter` as `null` to allow to return the body as a `byte[]`. 
This is much flexible and functional.
The end-user then has a choice to decide what the conversion model to select or jut deal with `byte[]`. The Binder then just will passes the `null` to the converter property.

That's exactly the way I've chosen for the Kinesis Binder: https://github.com/spring-cloud/spring-cloud-stream-binder-aws-kinesis/blob/master/spring-cloud-stream-binder-kinesis/src/main/java/org/springframework/cloud/stream/binder/kinesis/KinesisMessageChannelBinder.java#L203

A `PathThroughConverter` is just an idea you can implement, however the `null` is much performant 😄 

Nevertheless you didn't explain the change in the commit message...

joaoandremartins: Ideally, I would like to have a `CompositeMessageConverter` here similarly to SCSt that contains a bunch of pre-configured `MessageConverters` that convert payloads according to their content types.
That way we could support a multitude of payloads. Although for SCSt, this could possibly be overkill or even now work.

What do you think of having `messageConverter` always null, and if the users are interested in any specific kind of conversion/payload, they can specify the converter on demand?

artembilan: Well, for consistency with other Channel Adapters in the Spring Integration I would insist on the default convenient converter, but allow to configure it to `null` to let to deal with the raw `byte[]` as it is in case of SCSt. 

We can have any possible converter complexity, but that's really not productive to convert on the adapter and then on the SCSt side.

joaoandremartins:
PTAL

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapter.java
```diff
@@ -112,13 +120,11 @@ public void setAckMode(AckMode ackMode) {
...
-		Assert.notNull(messageConverter,
-				"The specified message converter can't be null.");
-		this.messageConverter = messageConverter;
-	}
-
 	public MessageConverter getMessageConverter() {
 		return this.messageConverter;
 	}
+
+	public void setMessageConverter(MessageConverter messageConverter) {
```
artembilan: Need JavaDocs. And explain the `null` behavior.

spring-cloud-gcp-pubsub-stream-binder/src/test/java/org/springframework/cloud/gcp/stream/binder/pubsub/PubSubEmulator.java
```diff
@@ -0,0 +1,80 @@
...
+
+	public PubSubEmulator() {
+		this.cloudSdkPath = Paths.get(System.getProperty("user.home"))
+				.resolve("google-cloud-sdk")
+				.resolve("bin")
+				.resolve(System.getProperty("os.name").toLowerCase().contains("windows")
+						? "gcloud.cmd"
+						: "gcloud");
+		// A port number between 10000 and 20000.
+		this.port = new Double((Math.random() + 1) * 10000).intValue();
```
artembilan: Consider to use `org.springframework.util.SocketUtils` instead.

spring-cloud-gcp-pubsub-stream-binder/src/test/java/org/springframework/cloud/gcp/stream/binder/pubsub/PubSubMessageChannelBinderTests.java
```diff
@@ -0,0 +1,80 @@
...
+
+	private PubSubTestBinder binder;
+
+	public PubSubMessageChannelBinderTests() {
+		this.binder = new PubSubTestBinder(this.emulator.getPort());
+	}
+
+	@BeforeClass
+	public static void enableTests() {
+		assumeThat(System.getenv("GCP_PUBSUB_EMULATOR")).isNotNull();
```
artembilan: Why do I need to set the ENV var if I start emulator manually from the rule?
Can't we just rely on the fact that emulator started or not?
For example:
```
logger.warn("Tests not running because no broker on " + url + ":", e);
assumeNoException(e);
```
in that `PubSubEmulator` instead.

joaoandremartins: This is what triggers the emulator running or not.
If `GCP_PUBSUB_EMULATOR` isn't an envvar, the rule isn't triggered.

spring-cloud-gcp-pubsub-stream-binder/src/test/java/org/springframework/cloud/gcp/stream/binder/pubsub/PubSubMessageChannelBinderTests.java
```diff
@@ -0,0 +1,80 @@
...
+ * Integration tests that require the Pub/Sub emulator to be installed and only run if the
+ * GCP_PUBSUB_EMULATOR envvar is present.
+ *
+ * @author João André Martins
+ */
+public class PubSubMessageChannelBinderTests extends AbstractBinderTests<PubSubTestBinder,
+		ExtendedConsumerProperties<PubSubConsumerProperties>,
+		ExtendedProducerProperties<PubSubProducerProperties>> {
+
+	@Rule
```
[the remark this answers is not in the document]
joaoandremartins: Nope, because each test provisions consumers/subscribers and we want a fresh environment for each test.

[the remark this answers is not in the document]
artembilan: ??? Are you going to do that in the real life as well?
Fresh environment for each  consumer/subscriber?

Then tests are wrong and definitely has to clean up resources after themselves.

I use `binding.unbind();` to do that though.

It really would be much better and more closed to the real world if we have an external resource once and reuse it from each test.

Otherwise something is bad by design...

spring-cloud-gcp-pubsub-stream-binder/src/test/java/org/springframework/cloud/gcp/stream/binder/pubsub/PubSubMessageChannelBinderTests.java
artembilan: That's wrong by design again. 
What happens if I have `GCP_PUBSUB_EMULATOR` env var, but don't have emulator?

More over it would be much better to be as friendly for external contributors as possible.
IMO just a single rule state is enough to determine if we need to run the tests or not.

I don't have any experience why I would prefer do not run tests at all during project build (unless it is just about docs building). So, if I have emulator, I definitely would like to run tests in the project.

Like I have all the Redis, MongoDB, Kafka, RabbitMQ etc. in my machine and I'm pretty happy to have all the test passes. If I don't have some resource (e.g. Oracle DB), the single rule makes a decision do not run tests. That's all.

I don't get it why do we need one more level of the management...

spring-cloud-gcp-pubsub-stream-binder/src/test/java/org/springframework/cloud/gcp/stream/binder/pubsub/PubSubMessageChannelBinderTests.java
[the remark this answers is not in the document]
joaoandremartins: It seems like all the tests [already clean after themselves](https://github.com/spring-cloud/spring-cloud-stream/blob/master/spring-cloud-stream-binder-test/src/main/java/org/springframework/cloud/stream/binder/AbstractBinderTests.java#L208).

Indeed, in real world, these tests should also be isolated IMO.

artembilan:
A general comment: when commit messages are going to be useful?
`Artem's review` is not a valuable information. And even I as a "guilty man" might not remember tomorrow what we have discussed here today.

Thanks for understanding

joaoandremartins:
Sure, I can try to fix the commit messages in the future.
But how does that help? The commits will all be squashed when the PR is merged and you can see what the PR does from the GitHub files view (provided it's on the smallish side)

spring-cloud-gcp-pubsub-stream-binder/src/test/java/org/springframework/cloud/gcp/stream/binder/pubsub/PubSubMessageChannelBinderTests.java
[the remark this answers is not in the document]
artembilan: No, my point that in real world you connect many applications with their own interests to the same Pub/Sub service and don't worry about cross-talking. Otherwise something is wrong from the applications perspective and has to be fixed.

artembilan:
We don't have GitHub locally...
And anyway the commit has to have not only `What?` but also `Why?`.

How may someone understand why did you remove `converter` support from the Channel Adapter if I haven't asked about that?

People may depend on our library and if they got some breaking changes they really would like to know "why?'. In this case good commit message helps a lot. That's not very productive to make people to read all the comments in the PR since the result of such a discussion is in the commit. 

I'm not sure how else to convince you, but I just share an experience we have here in Spring.

Sorry for noise otherwise

spring-cloud-gcp-pubsub-stream-binder/src/test/java/org/springframework/cloud/gcp/stream/binder/pubsub/PubSubMessageChannelBinderTests.java
joaoandremartins: According to https://cloud.google.com/pubsub/docs/emulator#pubsub-emulator-java, the Pub/Sub emulator is started if `PUBSUB_EMULATOR_HOST` is populated.
So I'll change this to the new envvar and delete `PubSubEmulator`. How does that sound?

joaoandremartins:
@artembilan thanks a lot for your advice, it's really appreciated!
I'll try to apply it more in the future.

spring-cloud-gcp-pubsub-stream-binder/src/test/java/org/springframework/cloud/gcp/stream/binder/pubsub/PubSubMessageChannelBinderTests.java
artembilan: Sounds good, but wait. Aren't you going to run emulator from the test suit any more then?
Well, actually that's fine with me, too. At least it is similar to what I do with `kinesalite` for testing AWS Kinesis. However I think I could start it from the `/npm` the same way you do here in the rule 😄

joaoandremartins: If the test starts the emulator, I think there might be some issues with the `@BeforeClass` not catching the populated envvar, because the emulator rule hasn't run yet.
However, I think this deserves some more thought, so I'll file an issue to consider the test running the emulator on its own.

joaoandremartins:
I checked all combinations except Windows and it seems to work everywhere, provided the emulator has been separately started, which the user or Travis can do.

Can you take this for a spin on Windows and let me know if it works? I can give you a list of steps to take if you don't know what to do.

artembilan:
Works for me locally without any GCP 😄 

I will come back to you, when I decide to test it against an Emulator and some GCP stuff.

Thanks

Issue #430, opened 2018-02-12
java.net.MalformedURLException: unknown protocol: gs


meltsufin:
The `getURL` method on `GoogleStorageResourceObject` throws a `MalformedURLException` because the `gs` protocol is unknown.

PR #433, opened 2018-02-13
Remove bucket resource

### LINKED ISSUE #426: Drop support for auto-creating GCS buckets

meltsufin:
This change merges the functionality of `GoogleStorageResourceBucket` and `GoogleStorageResourceObject` into one `GoogleStorageResource`.
It also drops the support for bucket auto-creation, which doesn't seem to provide much value anyway.

The tests have been updated, but doc updates are TODO.

Fixes #426, #430.


spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResource.java
```diff
@@ -0,0 +1,332 @@
...
+			if (this.location.getPath() != null && this.location.getPath().length() > 1) {
+				this.blobName = this.location.getPath().substring(1);
+			}
+			else {
+				this.blobName = null;
+			}
+
+			// verify that if it's a bucket handle, location ends with a '/'
+			if (this.blobName == null && !location.endsWith("/")) {
+				throw new IllegalArgumentException("Invalid location for a bucket: " + location
```
ChengyuanZhao: Could we handle these? I have a feeling it would be annoying as a user to have to keep this bit about the trailing slashes in mind.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResource.java
```diff
@@ -0,0 +1,332 @@
...
+		return this.storage.create(BlobInfo.newBuilder(getBlobId()).build());
+	}
+
+	/**
+	 * Creates the bucket that this resource references in Google Cloud Storage.
+	 * @return the {@link Bucket} object for the bucket
+	 * @throws StorageException if any errors during bucket creation arise,
+	 * such as if the bucket already exists
+	 */
+	public Bucket createBucket() {
```
ChengyuanZhao: It doesn't look like this method is used. Didn't you mention not wanting to auto-create buckets?

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResource.java
```diff
@@ -0,0 +1,332 @@
...
+	 * @return the {@link GoogleStorageResource} object for the created file
+	 * @throws IOException
+	 */
+	@Override
+	public GoogleStorageResource createRelative(String relativePath) throws IOException {
+		GoogleStorageResource blobResource = new GoogleStorageResource(
+				this.storage, this.location.resolve(relativePath).toString(), this.autoCreateFiles);
+
+		if (!blobResource.exists()) {
+			blobResource.createBlob();
```
ChengyuanZhao: Why not just like that relative blob determine its own creation in its own getOutputStream method like the original resource does?

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResource.java
```diff
@@ -0,0 +1,332 @@
...
+	public Bucket getBucket() {
+		return this.storage.get(this.bucketName);
+	}
+
+	/**
+	 * Checks for the existence of the {@link Bucket} associated with the resource.
+	 * @return true if the bucket exists
+	 */
+	public boolean bucketExists() {
+		return getBucket() != null;
```
joaoandremartins: Why do we need this extra method and why is it not proxying `exists()`?

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResource.java
```diff
@@ -0,0 +1,332 @@
...
+
+	/**
+	 * Creates a signed URL to an object, if it exists. This method will fail if this storage
+	 * resource was not created using service account credentials.
+	 * @param timeUnit the time unit used to determine how long the URL is valid.
+	 * @param timePeriods the number of periods to determine how long the URL is valid.
+	 * @return the URL if the object exists, and null if it does not.
+	 * @throws IllegalStateException if the resource reference is to a bucket, and not a blob.
+	 */
+	public URL createSignedUrl(TimeUnit timeUnit, long timePeriods)  {
```
joaoandremartins: Needs rebasing so #428 is included

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResource.java
joaoandremartins: As mentioned before in another PR, a user might want to create a blob without writing to it.
When I call `createRelative()`, I'm expecting a blob to be created right away.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResource.java
```diff
@@ -0,0 +1,332 @@
...
+	}
+
+	@Override
+	public String getDescription() {
+		return this.location.toString();
+	}
+
+	@Override
+	public InputStream getInputStream() throws IOException {
+		return Channels.newInputStream(throwExceptionForNullBlob(getBlob()).reader());
```
joaoandremartins: Should we explicitly check if this resource is a bucket rather than relying on the vague exception thrown by `throwExceptionForNullBlob()`?

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResource.java
```diff
@@ -0,0 +1,332 @@
...
+				throw new FileNotFoundException("The blob was not found: " + this.location);
+			}
+
+			blob = createBlob();
+		}
+
+		return Channels.newOutputStream(blob.writer());
+	}
+
+	public String getBlobName() {
```
joaoandremartins: Why not just using `getFilename()`?

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResource.java
```diff
@@ -0,0 +1,332 @@
...
+		return this.blobName;
+	}
+
+	public String getBucketName() {
```
joaoandremartins: Ditto on `getFilename()`

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResource.java
meltsufin: If the `/` is not there, relative URIs don't work right.

Also, I think it's better for the user to be explicit in whether they're declaring a bucket or a file path. It yields fewer surprises down the road.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResource.java
meltsufin: It a public API to allow users to explicitly create the bucket.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResource.java
meltsufin: It's possible for the bucket to exist, but the file not to exists. See tests.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResource.java
meltsufin: Implementation of the `getFilename()` might return the bucket name, if the resource represents a bucket. So, they're not exactly the same.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResource.java
ChengyuanZhao: But couldn't you add it for them if it isn't there? 

Conceptually the two approaches are identical, but I think it's not too hard to have an if-it's-there-then-do-nothing/if-it's-not-there-then-add-it section.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResource.java
meltsufin: `getFilename()` will return the blobName, if it's a valid path to a file -- not the bucket name.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResource.java
ChengyuanZhao: But you would getOutputStream on that other object, which woudl create it, before you write to it.

And if you were to say you would want to write to it via a signedURL, then I would say the signed URL function has the warning about that exact situation already. I don't see a reason in this asymmetry in behavior between the original resource and relative resources you get from it.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResource.java
joaoandremartins: This bucket can either represent a bucket or a blob.
If it represents a blob, there's no point in supporting a `bucketExists()` method.
This is the issue of having this object represent two different things.

Perhaps we shouldn't consider a bucket as a `Resource`, since you can't read it and, arguably, can't write to it (creating files isn't really writing to a bucket..)?

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResource.java
joaoandremartins: So again, if this resource represents a bucket, this would return null?
Another pointer why this probably shouldn't represent both a bucket and blob at the same time...

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResource.java
joaoandremartins: +1 for adding slashes for users.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResource.java
joaoandremartins: Why would you call `getOutputStream()` on the other object? There are no guarantees that will happen.
For all we're concerned, another completely different application which doesn't even use Spring or Java might write to that file.

joaoandremartins:
The more I think about this PR, the more I think we shouldn't be in the business of considering buckets as resources, since you can't read or write to them.
The only case where it could make sense to have buckets as resources is to create files inside them. But the blob abstraction with (auto)creation already covers that. So, if we're not interested in bucket auto-creation, might as well do away with buckets as resources altogether?

meltsufin:
I think the appropriate analogy to consider here is `java.util.File`. It represents a path to a file or a directory. It auto-creates files when you try to write to them, but it doesn't auto-create directories. So, I think these are the expectations that Java developers have, and we should try to match them.

There is a very legitimate use for bucket resources IMO.
For example, you declare the bucket:
```java
  @Value("gs://spring-demo/")
  private Resource bucket;
```
And then you use it to create files in it.

joaoandremartins:
I don't think that analogy is appropriate. We're working with `Resource`s, not `File`s.
GCS is not a file system, there's no obligation to treat it as one.
Furthermore, as I mentioned, a `Resource` is an object that gives you an input stream, with `WritableResource` giving you an output stream. We're not using buckets here for any of that, which is very suspicious.
As I mentioned as well, you can already "mimic" a file system using blobs as resources alone (especially considering the `createRelative()` method).

I feel very strongly that we shouldn't be doing this...

meltsufin:
Resources can be files. Take a look at `org.springframework.core.io.FileSystemResource`. It also supports both files and directories. I don't understand the perspective of why GCS cannot be viewed as a remote file system. I assume that most users think of it that way. What is so fundamentally different about `GoogleStorageResource` that it is not analogous to `FileSystemResource` in terms of functionality?

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResource.java
meltsufin: OK :)

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResource.java
meltsufin: I'm removing this code because judging by the `FileSystemResource` implementation, `createRelative` should not create any resources, but simply provide a handle to the relative resource.

meltsufin:
@ChengyuanZhao @joaoandremartins I think this is ready for the second pass.

PR #436, opened 2018-02-14
Adds integration tests for SI Pub/Sub channel adapters

joaoandremartins:
Fixes issue with string payloads on inbound adapter being considered as
byte arrays and making conversion error.
Emulator autoconfiguration for TransportChannelProvider isn't kicking
in, removing @ConditionalOnMissingBean worked.

An issue with this PR is that the `spring.cloud.gcp.pubsub.emulatorHost` property needs to be manually populated. Ideally, we would auto-populate that property with the value of the `PUBSUB_EMULATOR_HOST` envvar.


artembilan:
> An issue with this PR is that the spring.cloud.gcp.pubsub.emulatorHost property needs to be manually populated

But that's exactly the purpose of the `GcpPubSubEmulatorConfiguration`. It's condition is only based on a property. If we always fallback to the env var, our normal auto-config will never be triggered.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubEmulatorConfiguration.java
```diff
@@ -45,7 +45,6 @@
 public class GcpPubSubEmulatorConfiguration {
 
 	@Bean
-	@ConditionalOnMissingBean
```
artembilan: I think you have to reinstate this condition otherwise you don't give end-user any chances to provide some other instance.

Fortunately you did a good fix to the `GcpPubSubAutoConfiguration` already.

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/pubsub/it/PubSubChannelAdaptersIT.java
```diff
@@ -0,0 +1,231 @@
...
+import static org.assertj.core.api.Assumptions.assumeThat;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+
+/**
+ * @author João André Martins
+ */
+@SpringBootTest(
+		properties = {"spring.cloud.gcp.pubsub.emulatorHost=localhost:8085"},
```
artembilan: I think you can just fallback here to the `${PUBSUB_EMULATOR_HOST}`: https://docs.spring.io/spring-boot/docs/2.0.0.RC1/reference/htmlsingle/#boot-features-external-config-placeholders-in-properties

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/pubsub/it/PubSubChannelAdaptersIT.java
```diff
@@ -0,0 +1,231 @@
...
+		classes = {GcpPubSubAutoConfiguration.class,
+				GcpPubSubEmulatorConfiguration.class,
+				GcpContextAutoConfiguration.class,
+				PubSubChannelAdaptersIT.IntegrationConfiguration.class
+		}
+)
+@RunWith(SpringRunner.class)
+public class PubSubChannelAdaptersIT {
```
artembilan: `PubSubChannelAdapterIntegrationTests`, please.
`IT` is not what we use in Spring projects.

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/pubsub/it/PubSubChannelAdaptersIT.java
```diff
@@ -0,0 +1,231 @@
...
+		assertThat(message).isNotNull();
+		verify(callbackSpy, times(1)).onSuccess(any());
+	}
+
+	@Configuration
+	@EnableIntegration
+	static class IntegrationConfiguration {
+
+		@Autowired
+		@Lazy
```
artembilan: Why `@Lazy` ?

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/pubsub/it/PubSubChannelAdaptersIT.java
```diff
@@ -0,0 +1,231 @@
...
+		assertThat(message).isNotNull();
+		acker = (AckReplyConsumer) message.getHeaders().get(GcpHeaders.ACKNOWLEDGEMENT);
+		assertThat(acker).isNotNull();
+		acker.ack();
+		message = this.channel.receive(5000);
+		assertThat(message).isNull();
+	}
+
+	@Test
+	@DirtiesContext
```
artembilan: Can we consider to use `start()/stop()` on the `this.inboundChannelAdapter` instead?

I fully don't like when tests spend too much time. There are many other interesting things rather than wait for tests to pass, when we can just make them faster.

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/pubsub/it/PubSubChannelAdaptersIT.java
```diff
@@ -0,0 +1,231 @@
...
+				(AckReplyConsumer) message.getHeaders().get(GcpHeaders.ACKNOWLEDGEMENT);
+		assertThat(acker).isNotNull();
+		acker.nack();
+		message = this.channel.receive(5000);
+		assertThat(message).isNotNull();
+		acker = (AckReplyConsumer) message.getHeaders().get(GcpHeaders.ACKNOWLEDGEMENT);
+		assertThat(acker).isNotNull();
+		acker.ack();
+		message = this.channel.receive(5000);
+		assertThat(message).isNull();
```
artembilan: Why do we wait here for nothing 5 seconds?

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/pubsub/it/PubSubChannelAdaptersIT.java
```diff
@@ -0,0 +1,231 @@
...
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+
+/**
+ * @author João André Martins
+ */
+@SpringBootTest(
+		properties = {"spring.cloud.gcp.pubsub.emulatorHost=localhost:8085"},
+		classes = {GcpPubSubAutoConfiguration.class,
+				GcpPubSubEmulatorConfiguration.class,
```
artembilan: I think `GcpPubSubEmulatorConfiguration` must go first in the configuration order.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubEmulatorConfiguration.java
joaoandremartins: This `@ConditionalOnMissingBean` isn't behaving as expected.

`GcpPubSubEmulatorConfiguration` is indeed called before `GcpPubSubAutoConfiguration`, but only for the `credentialsProvider` bean. The `transportChannelProvider` bean is obtained from `GcpPubSubAutoConfiguration`.

It's almost as if Spring Boot prioritises beans with less dependencies (`transportChannelProvider` from `GcpPubSubEmulatorConfiguration` takes in `GcpPubSubProperties` whereas `GcpPubSubAutoConfiguration` doesn't take any), and `@ConditionalOnMissingBean` is a second priority criteria...

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/pubsub/it/PubSubChannelAdaptersIT.java
joaoandremartins: Not sure if I understand the question.

In my tests, I had to wait for a little while (e.g., more than one second) for the message to arrive.

In order to make sure that no message arrives, I thought we should wait for as long as the previous calls waited?

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/pubsub/it/PubSubChannelAdaptersIT.java
joaoandremartins: Because there is a theoretical circular dependency.

`IntegrationConfiguration` requires a `PubSubTemplate` bean that is created by `GcpPubSubAutoConfiguration`.
`GcpPubSubAutoConfiguration` requires a `GcpProjectIdProvider` bean that is provided by `IntegrationConfiguration` in this case.

The best way I found of resolving this issue was to make `PubSubTemplate` lazy, so that it's only instantiated when we need it. When that happens, the `GcpProjectIdProvider` is already in the context, so we're all good.

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/pubsub/it/PubSubChannelAdaptersIT.java
joaoandremartins: Not sure if I understand the comment.

`start()/stop()` don't restore the default values of the fields in the object. For each test, we want a fresh instance of a bean.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubEmulatorConfiguration.java
artembilan: Then it's not right to make your test working breaking the `main` code.
Something is not right indeed. What we have here with the `@ConditionalOnMissingBean` is the way to go. Need more investigation, but I wouldn't merge this change to `master`.

Just IMO, but you are feel free whatever you think is right, even if I think it's not right...

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/pubsub/it/PubSubChannelAdaptersIT.java
joaoandremartins: So the order here matters?
It seems to override the `@AutoConfigureBefore/After` order.

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/pubsub/it/PubSubChannelAdaptersIT.java
artembilan: Well, why should we wait for one more message at all?
Can we just remove these last couple lines at all, otherwise we definitely wait for whole 5 seconds?

As I said: that's very bad to have slow tests and even Stephane has pointed you how to increase performance for Spring Boot tests using slices approach.

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/pubsub/it/PubSubChannelAdaptersIT.java
artembilan: Right, they don't, but you can change adapters settings in between stop and start.
That's just doesn't look right to have `@DirtiesContext` on each test method.
Again: performance concern.

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/pubsub/it/PubSubChannelAdaptersIT.java
artembilan: Yes, good point!

The easy fix is to make that `gcpProjectIdProvider()` method as `static`. 
Another way is to use method argument injection instead of this field.

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/pubsub/it/PubSubChannelAdaptersIT.java
joaoandremartins: This method tests that a message which is nacked comes back and when we ack it, it goes away and doesn't come back.
I think it's important to verify that when the message is acked, it doesn't come back.

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/pubsub/it/PubSubChannelAdaptersIT.java
artembilan: I think the  `@AutoConfigureBefore/After` works when we pick up these classes from the auto-configuration phase, but in this case we just say to application context which classes to load. Nothing about auto-config.

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/pubsub/it/PubSubChannelAdaptersIT.java
joaoandremartins: It's only on each method that modifies the bean being integration tested (i.e., `PubSubInboundChannelAdapter`).

If I understand your suggestion correctly, it's that `start()/stop()` do restore the bean to its original default config? I wonder if that's a legitimate expectation? If a user configures a bean and then stops and starts, they're probably expecting the bean to have the same configuration?

Restoring the configuration on start()/stop() just because of tests doesn't sound right, let me know if I'm missing anything.

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/pubsub/it/PubSubChannelAdaptersIT.java
joaoandremartins: Actually, getting rid of `GcpContextAutoConfiguration` resolved the circular dependency for us.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubEmulatorConfiguration.java
joaoandremartins: It seems like you were right, and the order of the configuration classes in `@SpringBootTest.classes` overrides `@AutoConfigureBefore/After`, which is surprising since `@SpringBootTest`'s Javadoc doesn't mention any config order...

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/pubsub/it/PubSubChannelAdaptersIT.java
artembilan: The problem with the `@DirtiesContext` on each method that it fully restart the whole application context. I found it as very costly and would be better to just change properties of the bean in between stop and start. Right, you have to restore properties manually before starting the bean again. 

That's just doesn't feel reasonable to waste time just because you are lazy to restore bean state manually for each test-case.

Just IMO again. Although you should count that CI server is on our side, so respect its resources is friendly way to go as well :smile:

joaoandremartins: Aah, you mean the test's `@Before (start())/@After (stop())`?

That could be a better place to reset the properties. But it would require that we duplicate code that is currently on `PubSubInboundChannelAdapter`'s constructor, because we don't call the constructor, the Spring context does.

I could add something like

```
		StringMessageConverter stringMessageConverter = new StringMessageConverter();
		stringMessageConverter.setSerializedPayloadClass(String.class);
		this.inboundChannelAdapter.setMessageConverter(stringMessageConverter);
```

to the test's `@Before`, but that's an ugly duplication of code...

artembilan: Another way to go may be reached with only single test method in this class.

artembilan:
The Travis has failed somehow.

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/pubsub/it/PubSubChannelAdaptersIntegrationTests.java
```diff
@@ -160,18 +160,18 @@ public void sendAndReceiveMessageManualAck() {
...
 		acker.nack();
-		message = this.channel.receive(1000);
+		message = this.channel.receive(5000);
 		assertThat(message).isNotNull();
 		acker = (AckReplyConsumer) message.getHeaders().get(GcpHeaders.ACKNOWLEDGEMENT);
 		assertThat(acker).isNotNull();
 		acker.ack();
-		message = this.channel.receive(1000);
+		message = this.channel.receive(5000);
 		assertThat(message).isNull();
```
artembilan: I still think that wait here for nothing for 5 seconds is too expensive. Exactly for this `isNull()` use-case.
I think what you are waiting there is just for first message - when we warm up. All subsequent messages should be pulled from the topic pretty quickly. Otherwise something wrong with the configuration and maybe some `timeout` option should be applied.

joaoandremartins:
I don't understand why the Travis tests are failing. I suspect it's because of Pub/Sub emulator flakiness -- when I test locally, random tests fail randomly, eventually succeeding 100% every now and then.

artembilan:
Travis says:

> This log is too long to be displayed. Please reduce the verbosity of your build or download the raw log. 

And when I go to the link for full logs I see:
```
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 04:17 min
[INFO] Finished at: 2018-02-20T22:07:22+00:00
[INFO] Final Memory: 169M/963M
```
But a bit lower:
```
[1;31mERROR:[0m (gcloud.beta.emulators.pubsub.env-init) Unable to find env.yaml in the data_dir [/home/travis/.config/gcloud/emulators/pubsub]. Please ensure you have started the appropriate emulator.
Executing: /home/travis/build/spring-cloud/spring-cloud-gcp/google-cloud-sdk/platform/pubsub-emulator/bin/cloud-pubsub-emulator --host=localhost --port=8085
```

Not sure if that helps somehow, but we definitely need to do something to decrease logs verbosity with any reason.

artembilan:
Something is still wrong here, @joaoandremartins 

See Travis report

joaoandremartins:
@artembilan sorry, I'm doing a few experiments with Travis, since the tests are mostly passing locally.

Please disregard this PR for now. I'll nudge you when I think I found the culprit. Thanks!

joaoandremartins:
@artembilan @balopat @ChengyuanZhao this is now ready for full review.

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/pubsub/it/PubSubChannelAdaptersIntegrationTests.java
```diff
@@ -0,0 +1,289 @@
...
+		StringMessageConverter stringMessageConverter = new StringMessageConverter();
+		stringMessageConverter.setSerializedPayloadClass(String.class);
+		this.inboundChannelAdapter.setMessageConverter(stringMessageConverter);
+		this.inboundChannelAdapter.setAckMode(AckMode.AUTO);
+		this.outboundChannelAdapter.setPublishCallback(null);
+	}
+
+	@After
+	public void tearDown() {
+		if (testsRan == TOTAL_TESTS) {
```
artembilan: Can't we really have a `PubSubAdmin` as a statically configured and really rely on the `@AfterClass` instead?

I don't tell that we should somehow worry about subsequent application context we use (although we can use `static` properties for beans as well), but that way we won't forget what's going on with this hack when we add more tests.

Another way to go is just only have one `@Test` in the class.

joaoandremartins:
@artembilan should be ready for another pass. Thanks!

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/pubsub/it/PubSubChannelAdaptersIntegrationTests.java
```diff
@@ -0,0 +1,259 @@
...
+			.withPropertyValues("spring.cloud.gcp.pubsub.emulatorHost=${PUBSUB_EMULATOR_HOST}");
+
+	@BeforeClass
+	public static void checkEmulatorIsRunning() {
+		assumeThat(System.getenv(EMULATOR_HOST_ENVVAR_NAME)).isNotNull();
+	}
+
+	private void tearDown(PubSubAdmin pubSubAdmin) {
+		pubSubAdmin.deleteTopic("desafinado");
+		pubSubAdmin.deleteSubscription("doralice");
```
artembilan: But you still call this method in the end of each test.
Is it expected? What is the difference with the previous `@SpringBootTest` version ?

Thank

PR #437, opened 2018-02-15
Turns on the Pub/Sub emulator from Travis

joaoandremartins:
This PR installs the Google Cloud SDK along with the Pub/Sub emulator,
starts the emulator in a separate process and then sets the required
environment variables so that the integration tests pick it up and run.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,8 +10,9 @@ install:
   - tar -xvzf google-cloud-sdk-189.0.0-linux-x86_64.tar.gz
   - export PATH=$PWD/google-cloud-sdk/bin:$PATH
   - gcloud components update --quiet
-  - gcloud components install beta --quiet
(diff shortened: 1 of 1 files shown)


artembilan:
Travis somehow doesn't like what you propose to it. 
It even has decided do not log at all. 😄

artembilan:
I think Travis is dead right now.
I have some other PR which is building already for 43 mins: https://travis-ci.org/spring-projects/spring-integration

joaoandremartins:
I still tried to build it locally with the Travis local image, but it really doesn't mirror the remote well.
I guess I'll get back to this tomorrow!

joaoandremartins:
@artembilan @meltsufin PTAL

Twice already, the first build fails inexplicably in Travis and a second build (manually triggered by me) succeeds.
Hopefully these were just hiccups from the Travis outage (?) yesterday.

PR #438, opened 2018-02-15
Updated to work with Sleuth 2.0.0.M6+

saturnism:
First draft of the changes to work w/ Sleuth.

After talking to Adrian, it's advised that this use Sender. It's a bit more work than I expected. Will update once we migrate to [Zipkin GCP](https://github.com/GoogleCloudPlatform/stackdriver-zipkin/issues/45).

@marcingrzejszczak a few questions:
- HTTP URL and HTTP Method no longer seem to be populated by Sleuth, is that right?
- A method w/ Span annotation generates a Span, but has no Kind. Is there a way to determine whether it's made a RPC client call? It used to be doable by inspecting the log.


artembilan:
@marcingrzejszczak ,

FYI, please, review this stuff when you have a spare time.

Thanks

docs/src/main/asciidoc/trace.adoc
```diff
@@ -85,9 +85,6 @@ to 100%.
 
 [source]
 ----
-spring.sleuth.sampler.percentage=1                      # Send 100% of the request traces to Stackdriver.
+spring.sleuth.sampler.proability=1                      # Send 100% of the request traces to Stackdriver.
```
joaoandremartins: probability?

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/trace/StackdriverTraceAutoConfigurationTests.java
```diff
@@ -80,14 +84,20 @@ public void init() {
...
-			this.tracer.close(span);
-
-			// Test that we are using 128bit Trace ID
-			Assert.assertNotEquals(0, span.getTraceIdHigh());
+			Span span = this.tracer.newTrace()
+					.kind(Span.Kind.CLIENT)
+					.name("test")
+					.start();
+			span.finish();
+			span.flush();
```
marcingrzejszczak: I don't think you need to `flush` after `finish` but maybe I missed sth.

saturnism:
@adriancole @marcingrzejszczak I'll need a bit of help regarding:

- HTTP URL and HTTP Method no longer seem to be populated by Sleuth, is that right?  HTTP URL gets displayed in Stackdriver Trace. Without the URL, Span Name will be used and it's always "get" in a HTTP request, for example.
- A method w/ Span annotation generates a Span, but has no Kind. Is there a way to determine whether it's made a RPC client call? It used to be doable by inspecting the log.  Without a Client Span corresponding to a Server Span, it doesn't display properly in Stackdriver Trace.

This is a temporary measure as soon as we have proper sender and zipkin-gcp project up :D But I'm concerned that even then, Zipkin2 Spans may not be showing up in Stackdriver Trace properly.

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/trace/StackdriverTraceAutoConfigurationTests.java
saturnism: ah right, i was testing to see if there were some differences. finish should be enough.

marcingrzejszczak:
@saturnism you can - check this out https://github.com/spring-cloud/spring-cloud-sleuth/blob/master/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/TraceHttpAutoConfiguration.java#L51-L59

so just set the `spring.sleuth.http.legacy.enabled` flag to `true`.

> A method w/ Span annotation generates a Span, but has no Kind. Is there a way to determine whether it's made a RPC client call? It used to be doable by inspecting the log. Without a Client Span corresponding to a Server Span, it doesn't display properly in Stackdriver Trace.

Can you show an example? All RPC calls should have the `KIND` set

saturnism:
@marcingrzejszczak I only see a SERVER side span; i.e.,  I only see the server side span on `/meet`, but i don't see a client side span, except for `busyWork` w/o a kind.

marcingrzejszczak:
Hmm, can you confirm the same in a standalone application? I remember that @adriancole said that it might be only the problem in tests. That sometimes, due to how Brave works, you can have only spans from one side of the communication.

saturnism:
@marcingrzejszczak this is actually in the standalone example, not from tests :(

saturnism:
Moved the Span discussion to https://github.com/spring-cloud/spring-cloud-sleuth/issues/875

spring-cloud-gcp-samples/spring-cloud-gcp-trace-sample/src/main/java/com/example/WorkService.java
```diff
@@ -33,11 +33,11 @@ public WorkService(RestTemplate restTemplate) {
 		this.restTemplate = restTemplate;
 	}
 
-	@NewSpan("busyWork")
+	@NewSpan
 	public void busyWork() {
 		LOGGER.info("starting busy work");
-		for (int i = 0; i < 3; i++) {
-			this.restTemplate.getForObject("http://localhost:8080/meet", String.class);
+		for (int i = 0; i < 1; i++) {
```
meltsufin: Isn't it kind of pointless now to have the `for` loop?

spring-cloud-gcp-samples/spring-cloud-gcp-trace-sample/src/main/java/com/example/WorkService.java
```diff
@@ -33,11 +33,11 @@ public WorkService(RestTemplate restTemplate) {
...
+			this.restTemplate.getForObject("http://localhost:8081/meet", String.class);
```
meltsufin: Why is the port `8081` now?

saturnism:
I feel we should turn on legacy mode by default. to do this, I feel we'll provide a HttpTracing bean that always provide the right information. Will update.

saturnism:
@marcingrzejszczak i'm trying to return a pre-configured `HttpTracing` bean, but `SkipPatternProvider`, `SleuthHttpClientParser`, `SleuthHttpServerParser`, and `SleuthHttpSampler` are all package protected. Can those be made public accessible so that others can configure the bean outside of Sleuth's auto configuration?

saturnism:
Discussed w/ @marcingrzejszczak on hangout. Primary reason to expose these classes is to be able to configure Sleuth in Legacy Mode w/o depending on the flag, to provide an opinionated `HttpTracing` object. However, it may be better to determine the fields that we need for Stackdriver Trace, because the missing information will be a problem for users who simply use Brave and Zipkin and still may not have these fields set.

With Sleuth 2, non-legacy, we have the following labels:

"traceId":"5a9ed4a0f5b62b02c5241e30a93a615f","id":"c5241e30a93a615f","kind":"SERVER","name":"get","timestamp":1520358560499872,"duration":1803801,"localEndpoint":{"serviceName":"default"},"remoteEndpoint":{"ipv6":"::1","port":63166},"tags":{"http.method":"GET","http.path":"/","mvc.controller.class":"ExampleController","mvc.controller.method":"work"}}

With Sleuth 2, legacy, we have the following tags:
```
{"traceId":"5a9ed3e9fc80939f1d66ed2bc302d0c7","id":"1d66ed2bc302d0c7","kind":"SERVER","name":"http:/","timestamp":1520358377489193,"duration":1522756,"localEndpoint":{"serviceName":"default"},"remoteEndpoint":{"ipv6":"::1","port":63107},"tags":{"http.host":"localhost","http.method":"GET","http.path":"/","http.url":"http://localhost:8080/","mvc.controller.class":"ExampleController","mvc.controller.method":"work"}}
```

Sleuth non-legacy mode is missing a few things that we need:
- http.host
- http.url

Full list of what Stackdriver Trace can have are here:
https://cloud.google.com/trace/docs/reference/v1/rest/v1/projects.traces#tracespan

adriancole:
fyi we are pretty good match on census tags. only thing is user agent, but
that can be pulled with the existing HttpHandler.requestHeader function

https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/HTTPAttributes.md

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/trace/StackdriverTraceAutoConfiguration.java
```diff
@@ -97,44 +108,100 @@ public StackdriverTraceAutoConfiguration(GcpProjectIdProvider gcpProjectIdProvid
...
-	public Sampler defaultTraceSampler(SamplerProperties config) {
-		return new PercentageBasedSampler(config);
+	Tracing tracing(@Value("${spring.zipkin.service.name:${spring.application.name:default}}") String serviceName,
+			Propagation.Factory factory,
+			CurrentTraceContext currentTraceContext,
+			Reporter<zipkin2.Span> reporter,
+			Sampler sampler) {
+		return Tracing.newBuilder()
+				.sampler(sampler)
+				.traceId128Bit(true)
```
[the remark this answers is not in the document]
saturnism: That's pretty much it. We had to do it in the past to force 128-bit trace ID. However, I noticed the 128-bit trace ID flag is gone now, and 128-bit trace ID is defaulting to false.

docs/src/main/asciidoc/trace.adoc
```diff
@@ -85,9 +85,12 @@ to 100%.
...
-spring.sleuth.sampler.percentage=1                      # Send 100% of the request traces to Stackdriver.
+spring.sleuth.sampler.probability=1                     # Send 100% of the request traces to Stackdriver.
 spring.sleuth.web.skipPattern=(^cleanup.*|.+favicon.*)  # Ignore some URL paths.
 ----
 
-Stackdriver Trace requires the use of 128-bit Trace ID. This starter ignores
-`spring.sleuth.traceId128` property and always uses 128-bit Trace ID.
+Spring Cloud GCP Trace does override some Sleuth configurations:
+
+- Spring Cloud GCP Trace always uses 128-bit Trace IDs. This is required by Stackdriver Trace.
```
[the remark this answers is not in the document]
saturnism: if we set this to false, does it mean that i don't need to rewrite the parent IDs etc, anymore? :D

marcingrzejszczak:
I've added the flag back

docs/src/main/asciidoc/trace.adoc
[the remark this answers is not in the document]
saturnism: i think longer term, we may want to move the `StackdriverHttpClientParser` etc, to `zipkin-gcp` so that it'll be useable even w/o the starter. wdyt?

adriancole:
Yes i think that is also the plan in zipkin-aws for xray

adriancole:
Yes if join is disabled no need as the only thing that shares id is server
code. You are controlling config for server code.

adriancole:
On collector side it can look at json and only rewrite if shared flag is
set. In a uniform deployment this might never be set. Ex zipkin go also can
control this

On 9 Mar 2018 8:06 am, "Adrian Cole" <adrian.f.cole@gmail.com> wrote:

Yes if join is disabled no need as the only thing that shares id is server
code. You are controlling config for server code.

saturnism:
Done! I feel once https://github.com/spring-cloud/spring-cloud-sleuth/issues/890 and https://github.com/spring-cloud/spring-cloud-sleuth/issues/892 are resolved, we are good to go w/ this PR :)

saturnism:
@marcingrzejszczak @adriancole sleuth snapshot worked much better :) When is the next Milestone release?

marcingrzejszczak:
https://github.com/spring-cloud/spring-cloud-release/milestones

adriancole:
here's some last holes punched
https://github.com/spring-cloud/spring-cloud-sleuth/pull/906

tested with https://github.com/openzipkin/zipkin-gcp/pull/62

adriancole:
fyi stackdriver sender going out in zipkin-gcp 0.2.0 https://github.com/openzipkin/zipkin-gcp/tree/master/sender-stackdriver

Hopefully this can eliminate a lot of code here, especially things custom about reporting.

saturnism:
That is fantastic! Thank you Adrian!

I'm awaiting for M9 to update and merge this in. I've created #485 to test and update to use Zipkin Sender. I feel we'll do that in a separate PR and catch the next Milestone or RC release, but prior to GA.

PR #439, opened 2018-02-15
Start using DefaultCredentialsProvider everywhere

### LINKED ISSUE #403: Use DefaultCredentialsProvider in all product integrations
Currently, it's only used in core and config.

meltsufin:
Fixes #403.

### DIFF
--- a/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/config/GcpConfigProperties.java
+++ b/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/config/GcpConfigProperties.java
@@ -1,5 +1,5 @@
 /*
- *  Copyright 2017 original author or authors.
+ *  Copyright 2017-2018 original author or authors.
  *
  *  Licensed under the Apache License, Version 2.0 (the "License");
  *  you may not use this file except in compliance with the License.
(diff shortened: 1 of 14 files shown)


spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/config/GcpConfigProperties.java
```diff
@@ -28,7 +29,7 @@
  * @author João André Martins
  */
 @ConfigurationProperties("spring.cloud.gcp.config")
-public class GcpConfigProperties {
+public class GcpConfigProperties implements CredentialsPropertiesProvider {
```
artembilan: I would call `CredentialsSupplier`. We don't provide any properties here.

Although that is just IMO

spring-cloud-gcp-core/src/main/java/org/springframework/cloud/gcp/core/CredentialsPropertiesProvider.java
```diff
@@ -0,0 +1,24 @@
...
+ *  limitations under the License.
+ */
+
+package org.springframework.cloud.gcp.core;
+
+/**
+ * @author Mike Eltsufin
+ */
+public interface CredentialsPropertiesProvider {
+	Credentials getCredentials();
```
artembilan: JavaDocs and blank lines around class member.

meltsufin:
@artembilan Another pass? Thanks!

PR #449, opened 2018-02-21
Gets Credentials from a Pivotal Cloud Foundry environment

### LINKED ISSUE #255: Add Cloud Foundry support
Our core starter should grab `GoogleCredentials` from the right spot on a PCF cluster.

joaoandremartins:
Fixes #255


joaoandremartins:
This is a rough proof of concept for getting credentials from a PCF environment that is yet completely untested.
I just wanted to validate the initial design that I think the other services will follow.
Someone might also pick this up while I'm away on vacations.

@mkjelland @boparp-pivotal FYI

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/core/GcpContextAutoConfiguration.java
```diff
@@ -72,4 +75,11 @@ public GcpProjectIdProvider gcpProjectIdProvider() {
 
 		return projectIdProvider;
 	}
+
+	@Bean
+	@ConditionalOnMissingBean
+	@ConditionalOnProperty("VCAP_SERVICES")
+	public PcfConfiguration pcfConfiguration() {
+		return new DefaultPcfConfiguration(System.getenv("VCAP_SERVICES"));
```
artembilan: I would suggest you to pay attention to the `CloudFoundryVcapEnvironmentPostProcessor` in Spring Boot and don't worry about this ENV var at all.

I think there should be something in the [Reference Manual](https://docs.spring.io/spring-boot/docs/2.0.0.RC2/reference/htmlsingle/#production-ready-cloudfoundry) as well.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/core/GcpContextAutoConfiguration.java
```diff
@@ -72,4 +75,11 @@ public GcpProjectIdProvider gcpProjectIdProvider() {
 
 		return projectIdProvider;
 	}
+
+	@Bean
+	@ConditionalOnMissingBean
+	@ConditionalOnProperty("VCAP_SERVICES")
+	public PcfConfiguration pcfConfiguration() {
```
mkjelland: I would recommend calling this CfConfiguration instead. PCF is the Pivotal commercial version of the open source cloud foundry project. It shouldn't matter if you are using PCF or CF (open source), this would work the same way.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/core/GcpContextAutoConfiguration.java
joaoandremartins: After looking at `CloudFoundryVcapEnvironmentPostProcessor`, I don't think we can use it because it contains user-specified values as key.

In other words, say I create a Pub/Sub service (with `cf create-service google-pubsub standard cactuar`). `cactuar` is the service name that I chose. The resulting JSON will be:

```
"google-pubsub": [
   {
    "credentials": {
     "Email": "pcf-binding-3352ec74@graphite-test-spring-cloud-gcp.iam.gserviceaccount.com",
     "Name": "pcf-binding-3352ec74",
     "PrivateKeyData": "**ellided**",
     "ProjectId": "graphite-test-spring-cloud-gcp",
     "UniqueId": "118335157430758507928",
     "subscription_name": "",
     "topic_name": "pcf_sb_4_1521138145589148714"
    },
    "label": "google-pubsub",
    "name": "mypubsub",
    "plan": "default",
    "provider": null,
    "syslog_drain_url": null,
    "tags": [
     "gcp",
     "pubsub"
    ],
    "volume_mounts": []
   }
  ]
```

`CloudFoundryVcapEnvironmentPostProcessor` then transforms this into properties with the prefix `vcap.services.cactuar`. For example, `vcap.services.cactuar.credentials.PrivateKeyData`.

Alternatives to this are welcome!

joaoandremartins:
Added Cloud SQL credentials with a little bit of hacking, due to the existing restrictions.

Let me know what you think from https://github.com/spring-cloud/spring-cloud-gcp/pull/449/commits/ab3d9475315dae0fdbf5dc12ee3ec0d6d07f768c

artembilan:
I'm not sure why nobody familiar with Cloud Foundry review this yet, but I would say all those System ENV and properties manipulations are not Spring Boot friendly...

joaoandremartins:
@artembilan I agree. That being said, I explained in https://github.com/spring-cloud/spring-cloud-gcp/pull/449#discussion_r174920421 why the Spring Boot way is unfeasible.

If anyone suggests a working alternative, I'm happy to follow. Otherwise, this is the only approach I know works...

joaoandremartins:
@boparp-pivotal it would be great if you could review this PR!
If you know anyone else who has a strong knowledge of CF and could help with the review, that would be very helpful too!

joaoandremartins:
This PR is still lacking some tests, but is ready for a good pass.
@snicoll @balopat @ChengyuanZhao FYI

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/sql/SqlCredentialFactory.java
```diff
@@ -41,13 +43,52 @@
...
+	public static final String CF_SQL_SERVICE_ACCOUNT = "GOOGLE_CLOUD_SQL_CF_SERVICE_ACCOUNT";
+
 	private static final Log LOGGER = LogFactory.getLog(SqlCredentialFactory.class);
 
 	@Override
 	public Credential create() {
 		// TODO(joaomartins): Consider supporting Spring Resources as credential locations. There
 		// would need to be a way to create or inject a Spring context here statically, to load
 		// the resource from there.
+		Credential finalCredential = getCloudFoundryCredential();
```
balopat: just `credential` should suffice.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/sql/SqlCredentialFactory.java
```diff
@@ -41,13 +43,52 @@
...
+
+		if (finalCredential == null) {
+			finalCredential = getFileCredential();
+		}
+
+		return finalCredential;
+	}
+
+	private Credential getCloudFoundryCredential() {
+		String vcapServicesEnvVar = System.getenv("VCAP_SERVICES");
```
balopat: nit: extract String literal to constant

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/sql/GcpCloudSqlAutoConfiguration.java
```diff
@@ -121,6 +121,15 @@ public CloudSqlJdbcInfoProvider defaultMySqlJdbcInfoProvider(GcpCloudSqlProperti
 						+ defaultProvider.getJdbcDriverClass());
 			}
 
+			if (System.getenv("VCAP_SERVICES") != null) {
```
balopat: did you consider using [Environment](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/core/env/Environment.html) instead? That should make it easier to test too! 
+1 for extracting string literal to constant here as well

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/storage/GcpStorageAutoConfiguration.java
```diff
@@ -56,11 +59,27 @@
...
 	public static Storage storage(CredentialsProvider credentialsProvider,
 			GcpStorageProperties gcpStorageProperties,
-			GcpProjectIdProvider projectIdProvider) throws IOException {
+			GcpProjectIdProvider projectIdProvider,
+			@Autowired(required = false) CfConfiguration cfConfiguration) throws IOException {
+		Credentials credentials;
+
+		CredentialsProvider cfCredentialsProvider = null;
+		if (cfConfiguration != null) {
+			cfCredentialsProvider = cfConfiguration.getStorageCredentialsProvider();
```
balopat: I'm wondering if we could inject `@Qualifier` beans for the credentials with names like "CfStorageCredentials", that would resolve the multiple `Credentials` class of beans and we could just inject them directly and then we could get rid of this if statement

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubAutoConfigurationTests.java
```diff
@@ -0,0 +1,60 @@
...
+				new String(Files.readAllBytes(vcapServicesFile.getFile().toPath()));
+		this.contextRunner.withSystemProperties("VCAP_SERVICES=" + vcapServicesString)
+				.run(context -> {
+					GcpPubSubAutoConfiguration pubSubAutoConfiguration =
+							context.getBean(GcpPubSubAutoConfiguration.class);
+					ServiceAccountCredentials credentials =
+							(ServiceAccountCredentials) pubSubAutoConfiguration
+									.getCredentialsProvider().getCredentials();
+					assertThat(credentials.getClientEmail()).isEqualTo(
+							"pcf-binding-3352ec74@graphite-test-spring-cloud-gcp.iam."
```
balopat: one thing I usually do when dealing with file based tests (json, xml) is that I use a template engine to fill out parts that I want to control from my tests. This might be overkill for now, but overall makes it easier to maintain the tests.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/core/GcpContextAutoConfiguration.java
dsyer: If I were you I'd look at binding `vcap.services.*` to a `Map` and pull the data out of there. Did you try that? Either that or use a naming convention for the service instances (and make it configurable of course). Also see here for another approach: https://github.com/spring-cloud/spring-cloud-cloudfoundry/blob/master/spring-cloud-cloudfoundry-web/src/main/java/org/springframework/cloud/cloudfoundry/environment/VcapServiceCredentialsEnvironmentPostProcessor.java

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/storage/GcpStorageAutoConfiguration.java
joaoandremartins: I don't think it's worth the trouble, since the storage-specific credentials are only really used by the `Storage` object, so no point exposing another qualified `CredentialsProvider` just to be used in one place.

Also, we would add more complexity around `CredentialsProvider` management where it doesn't exist, nor it needs to. Currently, we know that the `CredentialsProvider` in the context is for the Core credentials. Adding more `CredentialsProvider`s to the context would require qualified usage anywhere we use `CredentialsProvider`, and require that any user knows about how they're managed.

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubAutoConfigurationTests.java
joaoandremartins: Ack

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/sql/GcpCloudSqlAutoConfigurationTests.java
```diff
@@ -174,4 +184,75 @@ public void testNoJdbc() {
 				});
 	}
 
+	@Test
+	public void testNoInstanceConnectionName() {
```
joaoandremartins: Unrelated to this change, though it'd be a good idea to add this regardless.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/core/GcpContextAutoConfiguration.java
joaoandremartins: @dsyer not sure if I'm missing anything. As I mentioned, `vcap.services.*` are indexed by user-specified keys. There is no way we can know what the user will pick. In the example I gave, I called my Pub/Sub service `cactuar` and that resulted in `vcap.services.cactuar`.

I understand that needs to be because the same service can be bound several times to the same application. In that scenario, we're actually failing, because we don't have any criteria to choose from any of the service accounts.

Also, as I understand, post processors kick in after then context is built, whereas we need the information from `VCAP_SERVICES` to build the context.

Let me know if I'm missing anything or if there are other alternatives. For now, this seems to be the only one that works for our requirements.

joaoandremartins:
@artembilan knowing you're not fully familiar with CF, would be nice to get another pass from you here.
I think @snicoll is providing the CF expertise here later, when he's got the free cycles.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/spanner/GcpSpannerAutoConfiguration.java
```diff
@@ -121,4 +138,9 @@ public SpannerMutationFactory spannerMutationFactory(
 			SpannerMappingContext spannerMappingContext) {
 		return new SpannerMutationFactoryImpl(spannerConverter, spannerMappingContext);
 	}
+
+	@VisibleForTesting
```
artembilan: We prefer do not provide any test-related stuff in the `main` code.
For the purpose to get access to the class properties there is a `DirectFieldAccessor` to be user from the tests.

joaoandremartins:
@artembilan as mentioned earlier, I believe that the post processors kick in after the context is built, whereas we need this envvar during context construction, to provide our own beans. Let me know if I'm wrong.
The problem with the existing VCAP parsing is that it uses user-provided keys that we can't guess.

Since this has been brought up so many times before, let me know if it's worth having a VC to discuss the issue.

artembilan:
Right, I think we need to have a call with appropriate people.
I still believe that custom `EnvironmentPostProcessor` does the trick here when we parse some env vars and build others. We do that in many other places and it works well. 

>  user-provided keys that we can't guess.

Well, that's not how Spring Boot works. One of its purpose is `Configuration on Convention`.
So, we require some particular properties end-user must provide to kick off auto-configuration.
If end-user doens't follow that convention we back up and he/she is on their own!

joaoandremartins:
Thanks a lot, @dsyer.

Another option would be to merge this PR as is and file a new issue to consider swapping the explicit parsing by an EnvPP, assuming there is nothing else wrong with this PR. @artembilan wdyt?

We would really like to ship CF support with our next M3 release, which we expect should happen within a week or two.

artembilan:
My position do not merge and do not ship the feature in M3 if we don't agree on the solution.
This PR brings a lot of breaking changes, which I somehow feel should gone when we come up with the `EnvironmentPostProcessor` solution.

I won't mind if someone else will express opposite position and this will be merged and `EnvironmentPostProcessor` way will be forgotten. 

Unfortunately my CF knowledge is too low so I can't hit the table to show how things should work...

joaoandremartins:
What do you mean with "breaking changes"?

I might be missing something, but I don't think anything will break moving from manual parsing to an EnvPP.
It should be invisible to a user -- they won't need to change anything to get CF credentials support.
It's just an implementation detail, which is why I'm suggesting this feature gets merged, and we optimize later.

artembilan:
Well, I admit that I didn't reviewed it properly because I'm not familiar with the feature, but I see 23 files changes and most of them are about existing classes. That's how I make a conclusion that is about breaking changes. 

On the other hand if you agree about `EnvPP`, why don't do that already now? You said we still have one-two weeks - sounds like a plenty of time to experiment.

That's not about optimization, that's about how properly to design API once and mix concerns to break your head in the future. 

Just remember how we reworked everything recently because of wrong direction for design from the Spring Boot perspective...

I guess I'm not a right person to help you here, since we don't have consensus.

snicoll:
@joaoandremartins and I had a meeting about this PR that helped a lot.

Here's a brain dump on my side, none of that is blocker IMO and we can refine later.


## Credentials provider
It would be nice if `CredentialsProvider` handling was such that the CF piece didn't show up in the regular auto-config at all. Currently, the logic works as follows:

. If the service is bound in CF, we use the information from VCAP
. If the user has set a property, we use that
. Otherwise, we use the global `CredentialsProvider` from the context

If the mapping between 1. and 2. is one-to-one, one idea is to create a special `PropertySource` that sets the property that is expected in 2. with the information from VCAP. This has several advantages:

* The code isn't aware of the CF specific bits
* The `Environment` clearly states why credentials are configured this way (you can see the property source using `/actuator/env`)

If that's not possible, then we need some sort of `Provider<CredentialsProvider>` where the implementation would apply the logic above. In this scenario, it would be beneficial that the CF pieces was registered independantly so that the provider doesn't have to know about it. An advantage of that is that if the application doesn't run in CF, then that piece isn't registered at all and we don't have to play with `@Autowired(required=false)`. Unrelated but `ObjectProvider<T>.getIfAvailable()` is what we use nowadays for this.

## CF checks
Rather that having a `@ConditionalOnProperty` we could use `@ConditinalOnCloudPlatform(CloudPlatform.CLOUD_FOUNDRY)`

## SqlCredentialFactory
The factory needs to be configured using system properties at the moment. We have a piece of code that does that for the non CF use case and this PR adds somewhat the same code when reading stuff from VCAP.

I'd like us to consider applying the same strategy as for the credentials provider. Setting that information using the regular keys seem quite appealing to me. If that's not possible, that code should be extracted from the main auto-config.

joaoandremartins:
Filed #501, #502 and #503 to address the uncovered issues.

@artembilan do you feel comfortable approving here now?

I plan to start working on #503 right away.

artembilan:
I wasn't involved in the discussion and not aware what you have worked out with Stephane.
So, that seems for me not fully correct to ask me to do something blindly.

Maximum what I can do is to dismiss my review and you can do whatever you agreed with other people.

Thanks for understanding.

joaoandremartins:
Well, I need a review from someone. :)
I don't think @snicoll can approve due to his permissions in this repo and you've been the one mostly involved with this PR.
I can ask @balopat or @ChengyuanZhao to approve, but that doesn't feel right given your insight here, and in Spring in general...

artembilan:
Why just not close this PR and start working on those issue, which looks like an agreed way to go?

joaoandremartins:
This PR implements the feature on master. The other issues are improvements to this initial implementation. That's how we've worked so far, iterating big changes on master, incrementally.
If you notice, #503 reads "Consider using". That's because I'm not sure yet if the map can be used. Or we might find that that way is actually uglier than parsing the JSON on our own. Too large of a scope for a single PR.

snicoll:
Perhaps this PR (and branch) can be used as a starting point and we should try again? IMO there are a lot of changes in this PR already without the review we had today so it doesn't really go in the direction I had anticipated.

I am for whatever option that helps us improve the project but my personal choice is to use the experiment in this PR to craft an improved more modular solution which, IMO, can perfectly happen in several, smaller PRs. If we don't think this is a good option, then merging this and then iterating on master works as well provided that we act on it short term.

(Kinda related, I'd like an M3 to add initializr support and we can't really do that now if you merge this and we have to "polish" improve the solution based on our conversation today).

Fair?

artembilan:
@joaoandremartins ,

Right, we did this way so far, but as you see an experience not so good: we reworked too much and introduced a lot of breaking changes in between milestones. Meanwhile it's good from the education and evolving perspective, that doesn't matter that all our solutions should go such a way. 

I agree with Stephane that this branch can be as a foundation for the possible agreed solution which may not require a radical rework in the future what happened to entire our previous `starter` solution. 

From here, please, explain why should we go with this feature the same route, but not start from the white list and merge that small part which work for everybody?

joaoandremartins:
Superseded by #509

PR #451, opened 2018-02-22
Spanner autoconfig

ChengyuanZhao:
Added autoconfig for spanner.


spring-cloud-gcp-autoconfigure/pom.xml
```diff
@@ -145,6 +145,11 @@
             <artifactId>spring-cloud-context</artifactId>
             <optional>true</optional>
         </dependency>
+      <dependency>
+        <groupId>org.springframework.data</groupId>
+        <artifactId>spring-cloud-gcp-data-spanner</artifactId>
+        <version>1.0.0.BUILD-SNAPSHOT</version>
+      </dependency>
```
balopat: please move this up after `<-- Trace -->` and remove version

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/spanner/GcpSpannerAutoConfiguration.java
```diff
@@ -0,0 +1,121 @@
...
+
+/**
+ * Provides Spring Data classes to use with Google Spanner.
+ *
+ * @author Chengyuan Zhao
+ */
+@Configuration
+@AutoConfigureAfter(GcpContextAutoConfiguration.class)
+@ConditionalOnProperty(value = "spring.cloud.gcp.spanner.enabled", matchIfMissing = true)
+@EnableConfigurationProperties(GcpSpannerProperties.class)
```
balopat: can you check whether it's logical AND or OR if we'd add here `@ConditionalOnClass(SpannerTemplate.class)` (similar to PubsubTemplate)?

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/Spanner/GcpSpannerAutoConfigurationTests.java
```diff
@@ -0,0 +1,59 @@
...
+/**
+ * @author Chengyuan Zhao
+ */
+@RunWith(SpringRunner.class)
+@SpringBootTest(classes = { GcpSpannerAutoConfiguration.class,
+		GcpContextAutoConfiguration.class }, properties = {
+				"spring.cloud.gcp.spanner.projectId=testProject",
+				"spring.cloud.gcp.spanner.instanceId=testInstance",
+				"spring.cloud.gcp.spanner.database=testDatabase" })
+@EnableSpannerRepositories
```
balopat: can we check if we can do this automatically for Spring Boot apps in the `GcpSpannerAutoConfiguration`?

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/config/EnableSpannerRepositories.java
```diff
@@ -0,0 +1,61 @@
...
+import org.springframework.data.repository.config.DefaultRepositoryBaseClass;
+
+/**
+ * @author Chengyuan Zhao
+ * @author Ray Tsang
+ */
+@Target(ElementType.TYPE)
+@Retention(RetentionPolicy.RUNTIME)
+@Import(SpannerRepositoriesRegistrar.class)
+public @interface EnableSpannerRepositories {
```
balopat: please review the [aerospike](https://github.com/spring-projects/spring-data-aerospike/blob/master/src/main/java/org/springframework/data/aerospike/repository/config/EnableAerospikeRepositories.java) and the mongodb version of the same and apply the right documentation + annotations (e.g. `@Inherited`, `@Documented`?)

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/spanner/GcpSpannerAutoConfiguration.java
ChengyuanZhao: I'm reading around and it seems to be AND within the list of a single annotation and also AND among all the annotations.

balopat: Thank you for looking into it! In that case I think it's best to put in the class condition too, otherwise users can enable it without having the actual classes on the classpath.

balopat:
LGTM

Issue #453, opened 2018-02-23
missing labels for logs of container running in gcp / gke


Andy2003:
using:

	<dependency>
		<groupId>org.springframework.cloud</groupId>
		<artifactId>spring-cloud-gcp-logging</artifactId>
	</dependency>

in my spring boot app with all the settings mentioned in https://github.com/spring-cloud/spring-cloud-gcp/tree/master/spring-cloud-gcp-logging the logs of the app have almost no labels for filter them out correcltly. See the following log entry:

	{
		insertId: "abc"
		labels: {
			levelName: "INFO"
			levelValue: "20000"
		}
		logName: "projects/<project-id>/logs/<app-name>"
		receiveTimestamp: "2018-02-23T11:34:20.947913720Z"
		resource: {
			labels: {
				cluster_name: ""
				container_name: ""
				instance_id: ""
				namespace_id: ""
				pod_id: ""
				project_id: "<project-id>"
				zone: "europe-west3-b"
			}
			type: "container"
		}
		severity: "INFO"
		textPayload: "message"
		timestamp: "2018-02-23T11:34:20.865Z"
	}

The app ist deployed to gcp / gke as a docker image. Additionally the trace Id of `spring-cloud-gcp-trace`is also not present in the log  entry.

I also tried using

	<dependency>
		<groupId>com.google.cloud</groupId>
		<artifactId>google-cloud-logging</artifactId>
		<version>1.17.0</version>
	</dependency>

without any improvement in the logs.

meltsufin:
What's your logging configuration? Which logging library and appender are you using?

Andy2003:
I used logback with the settings from the readme:

	<configuration scan="true" scanPeriod="1 minutes">
		<jmxConfigurator/>
		<include resource="org/springframework/cloud/gcp/logging/logback-appender.xml"/>
	
		<logger name="com.kiwigrid" level="DEBUG"/>
	
		<root level="INFO">
			<appender-ref ref="STACKDRIVER"/>
		</root>
	
	</configuration>

so `com.google.cloud.logging.logback.LoggingAppender` is used

meltsufin:
[...]

Issue #367, opened 2018-01-30
Support JSON-based logging


meltsufin:
Now that https://github.com/GoogleCloudPlatform/appengine-sidecars-docker/pull/81 has been merged, JSON-based logging is supported on App Engine flexible and GKE. So, there is an alternative to the direct API calls via the [`LoggingAppender`](https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/google-cloud-contrib/google-cloud-logging-logback/src/main/java/com/google/cloud/logging/logback/LoggingAppender.java) used in #286.

PR #454, opened 2018-02-26
add JsonLayout for logging

Andy2003:
will solve #367
add a workaround for #453


spring-cloud-gcp-logging/pom.xml
```diff
@@ -27,6 +27,10 @@
             <groupId>com.google.cloud</groupId>
             <artifactId>google-cloud-logging-logback</artifactId>
         </dependency>
+        <dependency>
+            <groupId>ch.qos.logback.contrib</groupId>
+            <artifactId>logback-json-classic</artifactId>
```
artembilan: I think this should be `optional`. Nobody says that I am always going to use JSON format for logging.

spring-cloud-gcp-logging/src/main/java/org/springframework/cloud/gcp/logging/StackdriverJsonLayout.java
```diff
@@ -0,0 +1,83 @@
...
+	private final String projectId;
+
+	public StackdriverJsonLayout() {
+		setAppendLineSeparator(true);
+		setIncludeMessage(false);
+		setIncludeException(false);
+		setIncludeFormattedMessage(false);
+		Gson formatter = new Gson();
+		setJsonFormatter(formatter::toJson);
+		this.projectId = System.getenv("GOOGLE_CLOUD_PROJECT");
```
artembilan: I think this has to be configurable: https://docs.spring.io/spring-boot/docs/2.0.0.RC2/reference/htmlsingle/#_environment_properties

As we as all other properties, too. IMHO...

spring-cloud-gcp-logging/src/main/java/org/springframework/cloud/gcp/logging/StackdriverJsonLayout.java
```diff
@@ -0,0 +1,83 @@
...
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+
+import ch.qos.logback.classic.spi.ILoggingEvent;
+import ch.qos.logback.classic.spi.IThrowableProxy;
+import com.google.cloud.logging.TraceLoggingEnhancer;
+import com.google.gson.Gson;
+
+/**
+ * @author Andreas Berger
```
artembilan: JavaDocs ?

spring-cloud-gcp-logging/src/main/java/org/springframework/cloud/gcp/logging/StackdriverJsonLayout.java
```diff
@@ -0,0 +1,83 @@
...
+ */
+public class StackdriverJsonLayout extends ch.qos.logback.contrib.json.classic.JsonLayout {
```
artembilan: I don't see reason in fully-qualified name here.
Am I missing anything?

spring-cloud-gcp-logging/src/main/java/org/springframework/cloud/gcp/logging/StackdriverJsonLayout.java
```diff
@@ -0,0 +1,83 @@
...
+
+	public static final String SEVERITY_ATTR_NAME = "severity";
```
artembilan: I have never liked abbreviation, especially public API.

We develop here Spring library, so would be great to be as friendly for users as possible.
`SEVERITY_ATTRIBUTE` or even just `SEVERITY` with good JavaDocs is fully enough.
But, please, no abbreviations.

artembilan: BTW, would be great to have JavaDocs on all the public API, including constructors and constants.

spring-cloud-gcp-logging/pom.xml
Andy2003: If I set the dependency to optional, everyone has to include both `spring-cloud-gcp-logging` and `logback-json-classic`. I don't think that's very user-friendly. Is it necessary to create an additional project `spring-cloud-gcp-json-logging` to have this dependency not optional?

spring-cloud-gcp-logging/src/main/java/org/springframework/cloud/gcp/logging/StackdriverJsonLayout.java
artembilan: Not sure what you mean about `no configuration scanning`, but what you have now is really just hard code and in Spring Boot no body is going to always set ENV var. You may have it as a fallback though...

spring-cloud-gcp-logging/src/main/java/org/springframework/cloud/gcp/logging/StackdriverJsonLayout.java
```diff
@@ -0,0 +1,229 @@
...
+	public void setIncludeExceptionInMessage(boolean includeExceptionInMessage) {
+		this.includeExceptionInMessage = includeExceptionInMessage;
+	}
+
+	/**
+	 * @param event the logging event
+	 * @return the map which should get rendered as json
+	 */
+	@Override
+	protected Map toJsonMap(ILoggingEvent event) {
```
artembilan: Raw types ?

spring-cloud-gcp-logging/src/test/java/org/springframework/cloud/gcp/logging/StackdriverJsonLayoutLoggerTests.java
```diff
@@ -0,0 +1,85 @@
...
+import org.junit.runner.RunWith;
+import org.mockito.junit.MockitoJUnitRunner;
+
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+/**
+ * @author Andreas Berger
+ */
+@RunWith(MockitoJUnitRunner.class)
```
artembilan: Why do we need this since I don't see any Mockito usage in the test ?

spring-cloud-gcp-logging/src/main/resources/org/springframework/cloud/gcp/logging/logback-json-appender.xml
```diff
@@ -0,0 +1,15 @@
...
+<!--
+Stackdriver json-log format provided for import.
+-->
+<included>
+    <property name="projectId" value="${projectId:-${GOOGLE_CLOUD_PROJECT}}"/>
+
+    <appender name="CONSOLE_JSON" class="ch.qos.logback.core.ConsoleAppender">
+        <encoder class="ch.qos.logback.core.encoder.LayoutWrappingEncoder">
+            <layout class="org.springframework.cloud.gcp.logging.StackdriverJsonLayout">
+                <projectId>${projectId}</projectId>
```
artembilan: Don't we need to cover here all other properties we can configure ?

spring-cloud-gcp-logging/README.adoc
```diff
@@ -72,3 +76,33 @@ Your configuration may then look something like this:
 `STACKDRIVER` appender.
 
 Also see the link:../spring-cloud-gcp-starters/spring-cloud-gcp-starter-logging[spring-cloud-gcp-starter-logging] module.
+
+=== Log via Console
```
artembilan: Why in the README? 
Why isn't this reflected in the Reference Manual?

Thanks

Andy2003: How to update the Reference Manual in this PR, it is not part of this project isn't it?

spring-cloud-gcp-logging/src/main/resources/org/springframework/cloud/gcp/logging/logback-json-appender.xml
Andy2003: The `projectId` is the only property which needs to get set up from external. All other properties of this class should not really be changed  if use in the google cloud.

spring-cloud-gcp-logging/README.adoc
artembilan: There is a `docs` module in this project. That's where all the `.adoc` s are hanging out. 

That's still unclear for me why support READMEs in each module where it is easy to lose the track and have inconsistency...

spring-cloud-gcp-logging/src/main/java/org/springframework/cloud/gcp/logging/StackdriverJsonLayout.java
```diff
@@ -0,0 +1,229 @@
...
+					}
+				}
+			}
+			map.put(FORMATTED_MESSAGE_ATTR_NAME, message);
+		}
+		add(MESSAGE_ATTR_NAME, this.includeMessage, event.getMessage(), map);
+		add(CONTEXT_ATTR_NAME, this.includeContextName, event.getLoggerContextVO().getName(), map);
+		addThrowableInfo(EXCEPTION_ATTR_NAME, this.includeException, event, map);
+		addTraceId(event, map);
+		add(SPAN_ID_ATTRIBUTE, this.includeSpanId, event.getMDCPropertyMap().get(MDC_FIELD_SPAN_ID), map);
```
draffensperger: How is the MDC span ID field set?

For Stackdriver Trace, the LogEntry `span_id` field should be hexadecimal encoded for the Stackdriver Trace viewer to know where to display the log entries as you look at a trace, per the [Trace Viewer > Log Entries doc](https://cloud.google.com/trace/docs/viewing-details#log_entries). This is because span IDs are specified as hex-encoded strings in the Stackdriver Trace API v2, see the [Span doc](https://cloud.google.com/trace/docs/reference/v2/rpc/google.devtools.cloudtrace.v2#span).

My thinking is that this should affect how the field is added to the MDC so this could would be fine as-is.

Andy2003: @draffensperger
Only the `traceId` is currently being processed. The connection of GCP-Traces and GCP-Logs works perfectly here. I only added the `spanid` because it's also [parsed by the fluentd sidecar](https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/blob/master/lib/fluent/plugin/out_google_cloud.rb#L123). Here I could not find any usage in the stackdriver at all.

draffensperger: OK, makes sense. Yes, the support for `span_id` in the Stackdriver Trace viewer is quite new. I think it makes sense to keep it here as yes, it's being parsed by the fluentd sidecar, but we can get it into the MDC late on.

Andy2003: Just for clarification: the `spanId` is already set by spring sleuth in the MDC and evaluated correctly by this `StackdriverJsonLayout`.

docs/src/main/asciidoc/logging.adoc
```diff
@@ -0,0 +1,107 @@
...
+----
+
+`STACKDRIVER_LOG_NAME` and `STACKDRIVER_LOG_FLUSH_LEVEL` environment variables can be used to customize the
+`STACKDRIVER` appender.
+
+Also see the link:../spring-cloud-gcp-starters/spring-cloud-gcp-starter-logging[spring-cloud-gcp-starter-logging] module.
+
+==== Log via Console
+
+If you run your spring boot application in a kubernetes cluster in the google cloud (gke), you can log json directly
```
meltsufin: Capitalization:
Spring Boot, Kubernetes, Google Cloud, GKE, JSON.

spring-cloud-gcp-dependencies/pom.xml
```diff
@@ -192,6 +192,12 @@
 				<artifactId>netty-tcnative-boringssl-static</artifactId>
 				<version>2.0.7.Final</version>
 			</dependency>
+
+			<dependency>
+				<groupId>ch.qos.logback.contrib</groupId>
+				<artifactId>logback-json-classic</artifactId>
+				<version>0.1.5</version>
```
meltsufin: I don't think we should be managing the version this dependency for everyone in the BOM. It should just go into the dependency management section of the parent POM.

spring-cloud-gcp-logging/README.adoc
```diff
@@ -1,74 +1 @@
...
-    <appender-ref ref="STACKDRIVER" />
-  </root>
-</configuration>
-----
-
-`STACKDRIVER_LOG_NAME` and `STACKDRIVER_LOG_FLUSH_LEVEL` environment variables can be used to customize the
-`STACKDRIVER` appender.
-
-Also see the link:../spring-cloud-gcp-starters/spring-cloud-gcp-starter-logging[spring-cloud-gcp-starter-logging] module.
+include::../docs/src/main/asciidoc/logging.adoc[]
```
meltsufin: We already have an issue tracking the creation of the reference documentation for Logging. See #288.
I would rather handle that separately, and limit this PR to extending this README instead.

spring-cloud-gcp-logging/src/main/java/org/springframework/cloud/gcp/logging/StackdriverJsonLayout.java
```diff
@@ -0,0 +1,229 @@
...
+import java.util.concurrent.TimeUnit;
+
+import ch.qos.logback.classic.spi.ILoggingEvent;
+import ch.qos.logback.classic.spi.IThrowableProxy;
+import ch.qos.logback.contrib.json.classic.JsonLayout;
+import com.google.cloud.logging.TraceLoggingEnhancer;
+import com.google.gson.Gson;
+
+/**
+ * This class provides a json layout for a logback appender compatible to the stackdriver
```
meltsufin: work capitalization: Stackdriver, JSON, Logback.

spring-cloud-gcp-logging/src/main/java/org/springframework/cloud/gcp/logging/StackdriverJsonLayout.java
```diff
@@ -0,0 +1,229 @@
...
+	 */
+	public static final String SPAN_ID_ATTRIBUTE = "logging.googleapis.com/spanId";
+
+	/**
+	 * The json field name for the trace-id
+	 */
+	public static final String TRACE_ID_ATTRIBUTE = "logging.googleapis.com/trace";
+
+	/**
+	 * The name of the MDC parameter, spring sleuth is storing the trace id at
```
meltsufin: Spring Sleuth

spring-cloud-gcp-logging/src/main/java/org/springframework/cloud/gcp/logging/StackdriverJsonLayout.java
```diff
@@ -0,0 +1,229 @@
...
+	private String projectId;
+
+	private boolean includeTraceId;
+
+	private boolean includeSpanId;
+
+	private boolean includeExceptionInMessage;
+
+	/**
+	 * creates a layout for a logback appender compatible to the stackdriver log format
```
meltsufin: Logback, Stackdriver

spring-cloud-gcp-logging/src/main/java/org/springframework/cloud/gcp/logging/StackdriverJsonLayout.java
```diff
@@ -0,0 +1,229 @@
...
+		this.includeExceptionInMessage = true;
+		this.includeException = false;
+		this.includeTraceId = true;
+		this.includeSpanId = true;
+		Gson formatter = new Gson();
+		setJsonFormatter(formatter::toJson);
+	}
+
+	/**
+	 * @return the google cloud project id relevant for logging the traceId
```
meltsufin: uppercase Google Cloud

spring-cloud-gcp-logging/README.adoc
artembilan: I would prefer to stay away of any work related to READMEs at all...

meltsufin: Ultimately, yes, but for now, I think it's OK.

Andy2003: So what to do now? @artembilan told me to add the doc to the reference manual and @meltsufin  tell me to move it back to the README. As the readme will link to the chapter of the reference manual, I think we are fine.

artembilan: Well, since we have that issue for Docs.
Let's just don't anything on the matter in this PR!

I think you are free to pick that up and PR the fix for Docs separately.

spring-cloud-gcp-logging/src/main/java/org/springframework/cloud/gcp/logging/StackdriverJsonLayout.java
```diff
@@ -0,0 +1,229 @@
...
+
+	private boolean includeSpanId;
+
+	private boolean includeExceptionInMessage;
+
+	/**
+	 * creates a layout for a Logback appender compatible to the Stackdriver log format
+	 */
+	public StackdriverJsonLayout() {
+		this.appendLineSeparator = true;
```
meltsufin: All of these properties should probably be described in the READNE, and maybe even added to the `logback-json-appender.xml` in a commented out form.

spring-cloud-gcp-logging/src/main/java/org/springframework/cloud/gcp/logging/StackdriverJsonLayout.java
meltsufin: The logging module may be used without Sleuth, and the `traceId` should be extracted from `TraceLoggingEnhancer.getCurrentTraceId()`. We may be able to put it into the MDC, but currently it's not happening.
See `TraceIdLoggingWebMvcInterceptor`.

Andy2003:
@meltsufin 
> The logging module may be used without Sleuth, and the traceId should be extracted from TraceLoggingEnhancer.getCurrentTraceId(). We may be able to put it into the MDC, but currently it's not happening.

I've already used the TraceId from the `TraceLoggingEnhancer` see method [`addTraceId`](https://github.com/spring-cloud/spring-cloud-gcp/pull/454/files#diff-6fe6f54054186dcbc8234f1eb8fb6ce5R217)

spring-cloud-gcp-logging/src/main/java/org/springframework/cloud/gcp/logging/StackdriverJsonLayout.java
```diff
@@ -0,0 +1,229 @@
...
+		if (!this.includeTraceId) {
+			return;
+		}
+
+		String traceId = TraceLoggingEnhancer.getCurrentTraceId();
+		if (traceId == null) {
+			traceId = event.getMDCPropertyMap().get(MDC_FIELD_TRACE_ID);
+		}
+		if (traceId != null) {
+			if (this.projectId != null) {
```
meltsufin: When I try to run this locally without defining `GOOGLE_CLOUD_PROJECT`, I still get the trace id attributed included in the JSON:
```json
{"timestampSeconds":1519845371,"timestampNanos":441000000,"severity":"INFO","thread":"qtp555635897-62","logger":"com.example.HelloController","message":"Test Logging","context":"default","logging.googleapis.com/trace":"projects/GOOGLE_CLOUD_PROJECT_IS_UNDEFINED/traces/5a96fffbfb40efeef4d7a9807df6eafc","logging.googleapis.com/spanId":"f4d7a9807df6eafc"}
```

Andy2003: Yeah, good point. I wasn't sure myself whether the `traceId` should appear in JSON at all then. Does it makes sense?

meltsufin: I think it can appear, but the condition on `projectId` being null doesn't seem to be working.

spring-cloud-gcp-logging/src/main/java/org/springframework/cloud/gcp/logging/StackdriverJsonLayout.java
```diff
@@ -0,0 +1,229 @@
...
+	private void addTraceId(ILoggingEvent event, Map<String, Object> map) {
+		if (!this.includeTraceId) {
+			return;
+		}
+
+		String traceId = TraceLoggingEnhancer.getCurrentTraceId();
+		if (traceId == null) {
+			traceId = event.getMDCPropertyMap().get(MDC_FIELD_TRACE_ID);
+		}
+		if (!StringUtils.isEmpty(traceId) && !StringUtils.isEmpty(this.projectId)) {
```
meltsufin: @Andy2003 The value of `this.projectId` seems to be `GOOGLE_CLOUD_PROJECT_IS_UNDEFINED`, when empty though. I think Logback might be translating from null to '...IS_UNDEFINED'? So, I don't think the condition is correct.

Andy2003: You are right, it can also happen that the projectId will end with `_IS_UNDEFINED`. I added an additional check for this.

Andy2003:
I also noticed that the log lines from the docker container are wrapped at a length of about 16 kB. Thus, fluentd no longer parses the JSON correctly. This happens with long stacktraces. Does anybody have a good solution?

meltsufin:
It might be better to just output stack traces directly without wrapping into JSON. The fluentd plugin parses stack traces automatically and creates error reports from them, even if they are multi-line.

@igorpeshansky Do you know regarding the log message limit?

Andy2003:
> It might be better to just output stack traces directly without wrapping into JSON. The fluentd plugin parses stack traces automatically and creates error reports from them, even if they are multi-line.

I don't think that's a solution. Hence, the TraceId for the stacktraces would be missing. But that's exactly where this connection is extremely helpful for the analysis.

meltsufin:
@Andy2003 I see your point, but can you log the short version of the exception without the full stack trace with JSON, and the full stack trace separately? This is not perfect, but might be a good solution given the log message length constraints.

igorpeshansky:
The Stackdriver limit on the log entry length is 100k. Docker has decided to go with 16k, breaking many log parsers (not just Stackdriver).
FWIW, GKE is running on a custom build of Docker that increases the wrap limit to 100k. Might be worth doing that with GAE Flex as well. I don't think there's a good solution for other systems.

Andy2003:
Our application runs within the GKE and we have reached the 16k limit. Do we have to configure anything else to achieve 100k?

Nevertheless, the PR should not be blocked by this, right?

igorpeshansky:
@Andy2003 This may depend on the master/node versions in your cluster. @crassirostris may know more.

crassirostris:
> Do we have to configure anything else to achieve 100k?

The patch was only introduced for COS images with version m64 and higher

meltsufin:
@Andy2003 I'm happy to merge this if you're not planning any more commits to this PR.

Andy2003:
Feel free to merge!

Am 05.03.2018 15:21 schrieb "Mike Eltsufin" <notifications@github.com>:

PR #458, opened 2018-03-05
name-based query methods

ChengyuanZhao:
added unit tests and support for name-based query methods

### DIFF
--- a/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerQueryLookupStrategy.java
+++ b/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerQueryLookupStrategy.java
@@ -1,0 +1,65 @@
+/*
+ *  Copyright 2018 original author or authors.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package org.springframework.cloud.gcp.data.spanner.repository.query;
+
+import java.lang.reflect.Method;
+
+import org.springframework.cloud.gcp.data.spanner.core.SpannerOperations;
+import org.springframework.cloud.gcp.data.spanner.core.mapping.SpannerMappingContext;
+import org.springframework.data.projection.ProjectionFactory;
+import org.springframework.data.repository.core.NamedQueries;
+import org.springframework.data.repository.core.RepositoryMetadata;
+import org.springframework.data.repository.query.QueryLookupStrategy;
+import org.springframework.data.repository.query.QueryMethod;
+import org.springframework.data.repository.query.RepositoryQuery;
+import org.springframework.util.Assert;
+
+/**
+ * @author Balint Pato
+ * @author Chengyuan Zhao
+ */
+public class SpannerQueryLookupStrategy implements QueryLookupStrategy {
+
+	private final SpannerOperations spannerOperations;
+
+	private final SpannerMappingContext spannerMappingContext;
+
+	public SpannerQueryLookupStrategy(SpannerMappingContext spannerMappingContext,
+			SpannerOperations spannerOperations) {
+		Assert.notNull(spannerMappingContext,
+				"A valid SpannerMappingContext is required.");
+		Assert.notNull(spannerOperations, "A valid SpannerOperations is required.");
+		this.spannerMappingContext = spannerMappingContext;
+		this.spannerOperations = spannerOperations;
+	}
+
+	@Override
+	public RepositoryQuery resolveQuery(Method method, RepositoryMetadata metadata,
+			ProjectionFactory factory, NamedQueries namedQueries) {
+		QueryMethod queryMethod = new QueryMethod(method, metadata, factory);
+
+		if (namedQueries.hasQuery(queryMethod.getNamedQueryName())) {
+			String sql = namedQueries.getQuery(queryMethod.getNamedQueryName());
+			throw new UnsupportedOperationException("Could not use SQL: " + sql
+					+ "Only method-name based methods are supported currently.");
+		}
+
+		return new PartTreeSpannerQuery(
+				queryMethod.getResultProcessor().getReturnedType().getDomainType(),
+				queryMethod, this.spannerOperations, this.spannerMappingContext);
+	}
+}
--- a/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryExecutor.java
+++ b/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryExecutor.java
@@ -1,0 +1,205 @@
+/*
+ *  Copyright 2018 original author or authors.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package org.springframework.cloud.gcp.data.spanner.repository.query;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.StringJoiner;
+import java.util.function.BiFunction;
+
+import com.google.cloud.spanner.Statement;
+
+import org.springframework.cloud.gcp.data.spanner.core.SpannerOperations;
+import org.springframework.cloud.gcp.data.spanner.core.convert.MappingSpannerWriteConverter;
+import org.springframework.cloud.gcp.data.spanner.core.mapping.SpannerMappingContext;
+import org.springframework.cloud.gcp.data.spanner.core.mapping.SpannerPersistentEntity;
+import org.springframework.data.domain.Sort;
+import org.springframework.data.repository.query.parser.PartTree;
+import org.springframework.data.util.Pair;
+
+/**
+ * Executes Google Spanner query statements using
+ * {@link org.springframework.data.repository.query.parser.PartTree}\ parsed method
+ * definitions.
+ *
+ * @author Chengyuan Zhao
+ */
+public class SpannerStatementQueryExecutor {
+
+	/**
+	 * Creates a Spanner statement.
+	 * @param sql the SQL string with tags.
+	 * @param tags the tags that appear in the SQL string.
+	 * @param params the parameters to substitute the tags. The ordering must be the same
+	 * as the tags.
+	 * @return an SQL statement ready to use with Spanner.
+	 * @throws IllegalArgumentException if the number of tags does not match the number of
+	 * params, or if a param of an unsupported type is given.
+	 */
+	public static Statement buildStatementFromSqlWithArgs(String sql, List<String> tags,
+			Object[] params) {
+		if (tags.size() != params.length) {
+			throw new IllegalArgumentException(
+					"The number of tags does match the number of params.");
+		}
+		Statement.Builder builder = Statement.newBuilder(sql);
+		for (int i = 0; i < tags.size(); i++) {
+			Object param = params[i];
+			BiFunction toMethod = MappingSpannerWriteConverter.singleItemType2ToMethodMap
+					.get(param.getClass());
+			if (toMethod == null) {
+				throw new IllegalArgumentException("Param: " + param.toString()
+						+ " is not a supported type: " + param.getClass());
+			}
+			builder = (Statement.Builder) toMethod.apply(builder.bind(tags.get(i)),
+					param);
+		}
+		return builder.build();
+	}
+
+	private static Pair<String, List<String>> buildPartTreeSqlString(PartTree tree,
+			SpannerMappingContext spannerMappingContext, Class type) {
+		SpannerPersistentEntity<?> persistentEntity = spannerMappingContext
+				.getPersistentEntity(type);
+		List<String> tags = new ArrayList<>();
+		StringBuilder stringBuilder = new StringBuilder();
+		stringBuilder.append("SELECT ");
+
+		if (tree.isDistinct()) {
+			stringBuilder.append("DISTINCT ");
+		}
+
+		stringBuilder.append("* FROM " + persistentEntity.tableName() + " ");
+
+		if (tree.hasPredicate()) {
+			stringBuilder.append("WHERE ");
+
+			StringJoiner orStrings = new StringJoiner(" OR ");
+
+			tree.iterator().forEachRemaining(orPart -> {
+				String orString = "( ";
+
+				StringJoiner andStrings = new StringJoiner(" AND ");
+
+				orPart.forEach(part -> {
+
+					String segment = part.getProperty().getSegment();
+					String tag = "tag" + tags.size();
+					tags.add(tag);
+					String andString = persistentEntity.getPersistentProperty(segment)
+							.getColumnName();
+
+					switch (part.getType()) {
+					case LIKE:
+						andString += " LIKE %@" + tag;
+						break;
+					case SIMPLE_PROPERTY:
+						andString += "=@" + tag;
+						break;
+					case TRUE:
+						andString += "=TRUE";
+						break;
+					case FALSE:
+						andString += "=FALSE";
+						break;
+					case IS_NULL:
+						andString += "=NULL";
+						break;
+					case LESS_THAN:
+						andString += "<@" + tag;
+						break;
+					case IS_NOT_NULL:
+						andString += "<>NULL";
+						break;
+					case LESS_THAN_EQUAL:
+						andString += "<=@" + tag;
+						break;
+					case GREATER_THAN:
+						andString += ">@" + tag;
+						break;
+					case GREATER_THAN_EQUAL:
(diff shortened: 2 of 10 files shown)


spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryTests.java
```diff
@@ -0,0 +1,268 @@
...
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+/**
+ * @author Chengyuan Zhao
+ */
+public class SpannerStatementQueryTests {
+
+	@Test
+	public void compoundNameConventionTest() {
```
balopat: Please break the tests up instead of compounding them.

spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryTests.java
```diff
@@ -0,0 +1,268 @@
...
+		PartTreeSpannerQuery partTreeSpannerQuery = new PartTreeSpannerQuery(Trade.class,
+				queryMethod, spannerOperations, new SpannerMappingContext());
+
+		Object[] params = new Object[] {
+				"BUY",
+				"abcd",
+				"abc123",
+		};
+
+		partTreeSpannerQuery.execute(params);
```
balopat: readability/semantics: it is unclear what the unspecified parameter is, please restructure

spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryTests.java
```diff
@@ -0,0 +1,268 @@
...
+
+		Object[] params = new Object[] {
+				"BUY",
+				"abcd",
+				"abc123",
+				8.88,
+				3.33,
+				new Trade(),
+				"ignored",
+		};
```
balopat: readability/semantics: it is unclear what the unsupported paramtype is, please restructure

spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryTests.java
```diff
@@ -0,0 +1,268 @@
...
+
+		partTreeSpannerQuery.execute(params);
+	}
+
+	@Test
+	public void countTest() {
+
+		QueryMethod queryMethod = mock(QueryMethod.class);
+
+		when(queryMethod.getName()).thenReturn("countByAction");
```
balopat: just for my sake, what does the "ByAction" part mean?

spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryTests.java
```diff
@@ -0,0 +1,268 @@
...
+
+		assertEquals(1, partTreeSpannerQuery.execute(params));
+	}
+
+	@Test
+	public void existsTest() {
+
+		QueryMethod queryMethod = mock(QueryMethod.class);
+
+		when(queryMethod.getName()).thenReturn("existsByAction");
```
balopat: same, confused by "byAction"

spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryTests.java
```diff
@@ -0,0 +1,268 @@
...
+		results.add(new Trade());
+
+		when(spannerOperations.find(any(), (Statement) any(), any()))
+				.thenReturn((List) results);
+
+		assertTrue((boolean) partTreeSpannerQuery.execute(params));
+	}
+
+	@Test
+	public void notExistsTest() {
```
balopat: lot of duplication between this and the `existsTest`, can you please refactor?

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerQueryLookupStrategy.java
```diff
@@ -0,0 +1,65 @@
...
+			SpannerOperations spannerOperations) {
+		Assert.notNull(spannerMappingContext,
+				"A valid SpannerMappingContext is required.");
+		Assert.notNull(spannerOperations, "A valid SpannerOperations is required.");
+		this.spannerMappingContext = spannerMappingContext;
+		this.spannerOperations = spannerOperations;
+	}
+
+	@Override
+	public RepositoryQuery resolveQuery(Method method, RepositoryMetadata metadata,
```
balopat: not covered by tests, please cover it!

spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryTests.java
ChengyuanZhao: The idea here was that we need a complex sequence of AND and ORs and we needed to see that all the parts come out in the correct order because that is important in SQL.

spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryTests.java
ChengyuanZhao: Could you elaborate? We can only give params as mysterious array of objects.

spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryTests.java
ChengyuanZhao: Could you elaborate? We can only give params as mysterious array of objects.

spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryTests.java
ChengyuanZhao: the two words here are countBy and Action. countBy means to do a count, and Action means the Action property of the Trade. Because "operation" words (like IsNull or GreaterThan) are specified, it means the sole arg for the query method is the Action that you want to count.

spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryTests.java
ChengyuanZhao: please see explanation for countByAction

spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryTests.java
ChengyuanZhao: I'm confused between combining and having tests separately... Please change as you see fit.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerQueryLookupStrategy.java
ChengyuanZhao: I'm not so sure this one should be tested. Testing it would just mean ensuring that we're getting the same query method back as we pass in, which is testing based on definition instead of results. I think that the method's return type is the only non-re-writing-definition test that is possible.

spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryTests.java
balopat: well, those are two really separate things: 
1.) a single testcase should have a single reason to fail (Single responsibility principle) - so when you have a compound test, the same testcase will be broken for a long time during a major refactoring for example. Typically it is easy to write this way the test (especially if you write the tests after the code is written!), but for maintenance it's suboptimal (which naturally comes out if you write code in a test driven way). 
2.) code duplication is most of the time bad - in unit test cases that's why we end up with parameterized test cases. The two methods here do the same thing with a single difference: if the returned array is empty, the result should be `false` otherwise `true`. We can create a helper method that takes a given array, and the expected return value, exercises runs the logic and assertion. We'll still have two separate test cases though at the end!

spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryTests.java
balopat: sorry, I wasn't clear: what are we testing?

spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryTests.java
balopat: what are we testing here?

spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryTests.java
ChengyuanZhao: There are too few params for the method name that was specified. 
"unspecifiedParameters" means there were some left ungiven.

spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryTests.java
ChengyuanZhao: the "new Trade()" is not a param type that Spanner can use. 
the name "unsupportedParamTest" is talking about that Trade param.

spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryTests.java
balopat: thanks for the explanation. I still don't fully understand, how would this translate to SQL? `SELECT COUNT(action) FROM trader` ? 
In that case, shouldn't we return 0 when a single `Trader()` is returned with a `null` action? 
[spanner docs for COUNT function with NULL values](https://cloud.google.com/spanner/docs/functions-and-operators#aggregate-functions)

spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryTests.java
ChengyuanZhao: I'm not quite sure where the threshold for "readability"/repetitiveness is. This method here is just a few lines,

ChengyuanZhao:
@balopat unit test changes added, PTAL

spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryTests.java
balopat: 1.) Tests should be a living documentation. My point is that by simply reading the test it is hard to infer that that parameter is the cornerstone of the test. An explanatory comment would be helpful, or extracting it to a variable like `Object unsupportedParam = new Trade()`. 
2.) I tried it deleting it, 3 and 3.0 - we get IllegalArgumentException in the first two cases, so the test passes, while it fails only with the 3.0 (double). Maybe we should match on a specific message instead just the class?

spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryTests.java
balopat: Ah, my misunderstanding: `By` is the separator for the `WHERE` clause. So this would translate to `SELECT COUNT(*) FROM trader where Action=<action_param>`.

spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryTests.java
balopat: My threshold is max 4-5 lines. This is 8.

spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryTests.java
ChengyuanZhao: I'll add a comment indicating what is causing the expected exception. But of course unit tests are by no means mathematical proofs. If one were to get hacker-y minded I'm sure the vast majority of unit tests in existence could be triggered in unexpected ways. 

The intention here is to trigger a specific condition at a specific line and that is what this test does.  I think it would be obtuse to make a new exception type just for this one case. I also don't like the idea of matching against the exact text of the exception, since the punctuation, spacing, or actual words of the exception message is not important to the information of the message.

spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryTests.java
ChengyuanZhao: fixed. (just so happens my threshold was 9!)

ChengyuanZhao: @balopat PTAL, i'd like to get this commit in so I can work on that SQL statement annotation and also the paging and sorting repo implementations !

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerQueryLookupStrategy.java
balopat: The resolution logic should be tested. As agreed verbally, we can address it when we have more query types to handle in the next PRs.

balopat:
LGTM, please review my last changes.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryExecutor.java
```diff
@@ -83,7 +84,7 @@ else if (tree.isExistsProjection()) {
 	 * @throws IllegalArgumentException if the number of tags does not match the number of
 	 * params, or if a param of an unsupported type is given.
 	 */
-	public static Statement buildStatementFromSqlWithArgs(String sql, List<String> tags,
+	private static Statement buildStatementFromSqlWithArgs(String sql, List<String> tags,
```
ChengyuanZhao: @balopat I had this as public because it will be used by the SQL annotated query method directly

spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryTests.java
```diff
@@ -33,48 +35,63 @@
...
+	@Before
+	public void initMocks() {
+		this.queryMethod = mock(QueryMethod.class);
+		this.spannerOperations = mock(SpannerOperations.class);
+		this.spannerMappingContext = new SpannerMappingContext();
+	}
+
+	private PartTreeSpannerQuery createQuery() {
+		return new PartTreeSpannerQuery(Trade.class,
+				this.queryMethod, this.spannerOperations, this.spannerMappingContext);
```
ChengyuanZhao: just wondering, why not put this inside initMocks too?

spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryTests.java
balopat: Thank you for the comment, clarity is key in saving time for future developers and ourselves. 
I agree with new exception type being too much, and on exact matching as well, that's why I usually match on partial patterns only. Error messages _are_ important - they are one of the first things you see as a developer who starts to work on a project, and they influence usability - hence having tests around them can be useful. I'll work out these examples in the next couple of days.

spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryTests.java
balopat: I can't :/ because the query creation fails in the constructor - it depends on the `queryMethod.getName()` to be not `null`.

ChengyuanZhao: LGTM on these refactorings though!

PR #466, opened 2018-03-13
Moves to google-cloud-java version 0.39

joaoandremartins:
We no longer need to set netty tcnative version on our side since
they removed its version from their bom.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -76,6 +76,12 @@
 				<artifactId>logback-json-classic</artifactId>
 				<version>0.1.5</version>
 			</dependency>
+
+			<dependency>
+				<groupId>com.google.guava</groupId>
+				<artifactId>guava</artifactId>
+				<version>20.0</version>
+			</dependency>
 		</dependencies>
 	</dependencyManagement>
 
--- a/spring-cloud-gcp-dependencies/pom.xml
+++ b/spring-cloud-gcp-dependencies/pom.xml
@@ -16,7 +16,7 @@
 	<packaging>pom</packaging>
 
 	<properties>
-		<google-cloud-bom.version>0.33.0-alpha</google-cloud-bom.version>
+		<google-cloud-bom.version>0.38.0-alpha</google-cloud-bom.version>
 		<cloud-trace-java.version>0.5.0</cloud-trace-java.version>
 	</properties>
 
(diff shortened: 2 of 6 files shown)


pom.xml
```diff
@@ -76,6 +76,12 @@
 				<artifactId>logback-json-classic</artifactId>
 				<version>0.1.5</version>
 			</dependency>
+
+			<dependency>
+				<groupId>com.google.guava</groupId>
+				<artifactId>guava</artifactId>
+				<version>20.0</version>
```
joaoandremartins: I think the guava version is being set somewhere inside Spring to 18.0 and google-cloud-java requires 20.0.

artembilan: Looks like there are much newer Guava versions: https://search.maven.org/#search%7Cgav%7C1%7Cg%3A%22com.google.guava%22%20AND%20a%3A%22guava%22

What makes you to stick with this one?

Also bear in mind: we somehow try to stay away of using Guava at all:
https://github.com/spring-projects/spring-boot/commit/654fe9a31c0588cae0cdb5d8301717c3af605fad#diff-7050cd73a8895c61627d3891a0a14bc7R76

We can talk about that tomorrow with Stephane

joaoandremartins: google-cloud-java uses it. https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/pom.xml#L145

I'm still trying to figure out what's forcing guava's version to 18.0..

artembilan: Doesn't `mvnw dependency:tree` help here to find out ?

joaoandremartins: @artembilan in #471 @snicoll seems to agree with this approach. PTAL

joaoandremartins: I think I found the culprit:

https://github.com/spring-cloud/spring-cloud-netflix/blob/master/spring-cloud-netflix-dependencies/pom.xml#L423

It's pulled in via `spring-cloud-dependencies`.

artembilan: Fine with me. Now I understand that the point is like do not use Guava from our code.
And that rule might be applied really only for the Spring Boot project.

Glad to see that you have figured out the culprit!
I wonder why do we need this Netflix dependency? 
Can't we just exclude it for free?

joaoandremartins:
Adding an exclusion block in this way doesn't prevent guava to be set to 18.

```

			<dependency>
				<artifactId>spring-cloud-dependencies</artifactId>
				<version>${spring-cloud-release.version}</version>
				<type>pom</type>
				<scope>import</scope>
				<exclusions>
					<exclusion>
						<artifactId>spring-cloud-netflix-dependencies</artifactId>
					</exclusion>
				</exclusions>
			</dependency>
```

artembilan:
I think just because the `spring-cloud-netflix-dependencies` isn't target artifact per se, but BOM.

We need to exclude `spring-cloud-starter-netflix-archaius`. That's from where Guava is pulled: https://github.com/spring-cloud/spring-cloud-netflix/blob/master/spring-cloud-starter-netflix/spring-cloud-starter-netflix-archaius/pom.xml#L54

snicoll:
Okay bit that's very fragile isn't it? What process do you have in place to prevent that from happening again?

PR #469, opened 2018-03-14
Close the ResultSet

snicoll:
This commit makes sure that the `ResultSet` internally created by `count` is closed.


snicoll:
I've noticed another case in `SpannerConverter#mapToList` but I am not sure where to fix it.

PR #471, opened 2018-03-14
Upgrade to Google Cloud bom 0.38

snicoll:
This commit upgrades to the latest Google Cloud bom. As that bom does
not provide dependency management for third party librarires, overrides
can be removed.

Also, as Guava isn't managed anymore, this project chooses a Guava
version for internal use.


joaoandremartins:
Closing as dupe of #466.

PR #474, opened 2018-03-19
removing unnecessary JUnit Runners

balopat:


balopat:
@artembilan would you mind reapproving? I repushed the branch because I reverted an extra explicit `Mockito.spy` that accidentally got in the first time.

PR #475, opened 2018-03-19
removed unnecessary column name map

ChengyuanZhao:
removed unnecessary column name map

### DIFF
--- a/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/mapping/SpannerPersistentEntity.java
+++ b/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/mapping/SpannerPersistentEntity.java
@@ -35,13 +35,6 @@ public interface SpannerPersistentEntity<T>
 	String tableName();
 
 	/**
-	 * Gets the property corresponding to the given column name.
-	 * @param columnName the name of the column corresponding to a stored property.
-	 * @return the property.
-	 */
-	SpannerPersistentProperty getPersistentPropertyByColumnName(String columnName);
-
-	/**
 	 * Gets the column names stored for this entity.
 	 * @return the column names.
 	 */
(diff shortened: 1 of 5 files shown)


spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/mapping/SpannerPersistentEntity.java
```diff
@@ -34,13 +34,6 @@
 	 */
 	String tableName();
 
-	/**
-	 * Gets the property corresponding to the given column name.
-	 * @param columnName the name of the column corresponding to a stored property.
-	 * @return the property.
-	 */
-	SpannerPersistentProperty getPersistentPropertyByColumnName(String columnName);
-
```
balopat: yay!

spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/core/SpannerObjectMapperImplTests.java
```diff
@@ -356,6 +358,34 @@ public void writeIncompatibleTypeTest() {
 		ByteArray bytes;
 	}
 
+	@SpannerTable(name = "custom_test_table_reduced")
+	private static class TestEntityReduced {
```
balopat: why is it reduced? compared to what?

ChengyuanZhao: Well previously it was ok that the entity had many more properties than there were columns, but now the opposite is true. to avoid writing out all those extra lists ( boolean, date, timestamp, long , string, etc...) I cut those out and made this new entity type.

balopat: Hmm...but then we'll have asymmetry between the read and write side, won't we? we're not covering reading those type of lists anywhere else, or do we? Is there anything bad about writing out those lists?

ChengyuanZhao: no technically it is not worse to write out those other cases, but I felt it would be obtuse since we're just getting these mapping methods from a Map we hard coded.

balopat: I think it's valuable to cover our map, and that avoids having to create a separate class just to avoid writing more too.

PR #477, opened 2018-03-19
Spanner read some col

ChengyuanZhao:
provides options to only read some columns

### DIFF
--- a/spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/spanner/GcpSpannerAutoConfigurationTests.java
+++ b/spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/spanner/GcpSpannerAutoConfigurationTests.java
@@ -14,7 +14,7 @@
  *  limitations under the License.
  */
 
-package org.springframework.cloud.gcp.autoconfigure.Spanner;
+package org.springframework.cloud.gcp.autoconfigure.spanner;
 
 import com.google.api.gax.core.CredentialsProvider;
 import com.google.auth.Credentials;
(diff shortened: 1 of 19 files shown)


balopat:
I'm not fully sold that we would want to implement partial reads this way. The version in this PR creates new, partially filled out entities.

I see two directions that are a bit more useful in my opinion: 
- (Projections)[https://github.com/spring-projects/spring-data-commons/blob/master/src/main/asciidoc/repository-projections.adoc) 
- Partial-update-in-memory: something similar to what you have here but merging the results with an existing domain object, e.g. I have a 10K objects with 250 columns/fields in memory and interested to update the timestamp field on them, it would be great if I could just "refresh" them automagically from spring data. 

@saturnism, @mp911de what do you think?

ChengyuanZhao:
@balopat Those sound like just other situations, and in fact going by your argument those two can be played off of each other and invalidate one or the other, too. 

I can see situations where this PR is useful
- user wants to run a query as he does with our existing get-everything methods, but he isn't interested in getting all the data in every row from the table. 
- he wants to get this partial data when he doesn't already have the objects in memory either because he hasn't done a previous query or he doesn't know the keys for those objects

- and projects definitely doesn't cut it because that is a compile-time thing where the user needs to create whole new interfaces! this can be used at runtime and determined at runtime.

In fact, for that reason alone I think projections , while fancy and cool, are definitely inadequate. 

For the reason about not having the objects in memory beforehand, i also find your second proposed idea inadequate by itself. 

I think the full picture requires all three.

mp911de:
Reading a subset of columns (fields, …) that are returned by a query is a common use case. MongoDB allows specifying which columns to read via `@Query` on repositories and its `Query` object (`new Query().fields().include("foo").exclude("bar")`). Similar for JPA (`@Query("SELECT user, firstname FROM USER")`.

We generally derive column availability from the resulting row/document instead of passing an enumeration of column names.

balopat:
Sounds good, I'm convinced, thank you, both!

ChengyuanZhao:
@balopat please LGTM again, I had to resolve conflicts with the write-some-cols work haha

PR #481, opened 2018-03-21
Makes autoconfig tests use ApplicationContextRunner

### LINKED ISSUE #313: Make auto configuration tests not use @SpringBootTest
As per https://github.com/spring-cloud/spring-cloud-gcp/pull/306#discussion_r162402664, the context should be built manually.

joaoandremartins:
Fixes #313

### DIFF
--- a/spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/logging/StackdriverLoggingAutoConfigurationTests.java
+++ b/spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/logging/StackdriverLoggingAutoConfigurationTests.java
@@ -27,7 +27,6 @@ import org.springframework.cloud.gcp.logging.TraceIdExtractor;
 import org.springframework.cloud.gcp.logging.TraceIdLoggingWebMvcInterceptor;
 import org.springframework.cloud.gcp.logging.XCloudTraceIdExtractor;
 import org.springframework.cloud.gcp.logging.ZipkinTraceIdExtractor;
-import org.springframework.test.web.servlet.setup.MockMvcBuilders;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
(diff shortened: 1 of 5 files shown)


spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/logging/StackdriverLoggingAutoConfigurationTests.java
```diff
@@ -32,107 +34,100 @@
...
 			}
 		}
 		return count;
 	}
 
+	@Test
+	public void testDisabledConfiguration() {
+		this.contextRunner.withPropertyValues("spring.cloud.gcp.logging.enabled=false")
+				.run(context -> {
+					Object[] interceptors = context.getBean(Object[].class);
```
joaoandremartins: @artembilan there are no interceptors (i.e., `Object[]` object) in the context. I guess this is because we don't have a Spring Boot test with `webEnvironment = SpringBootTest.WebEnvironment.MOCK` to create the web server and add interceptors to it?

In that case, I guess this is an integration test and we can't make it use `ApplicationContextRunner`?

artembilan: I'm not sure what `interceptors` you are talking about, but there is `MockMVC` if you don't want to use Spring Boot: https://docs.spring.io/spring/docs/5.0.4.RELEASE/spring-framework-reference/testing.html#spring-mvc-test-framework

However I think that `SpringBootTest.WebEnvironment.MOCK` is fully based on the mentioned testing framework. 

I guess your point is do not blow up the whole Spring Boot auto-config for this test, so yes, consider to use `MockMvc` to emulate a web environment for this test.

joaoandremartins: See my latest commit. Unfortunately, couldn't get it to work.

I'm a little out of my depth here, so I might just revert the test to its old format (and maybe rename it to something like `StackdriverLoggingIntegrationTests`).

artembilan: What do you mean with the "See my latest commit"? 
I don't see any explanation of the problem in the commit message.
Sorry

artembilan: Why do you try to get this weird `Object[]` bean? 
Looks like you would like to assert for the `TraceIdLoggingWebMvcInterceptor`.
Why don't call then `context.getBeansOfType(TraceIdLoggingWebMvcInterceptor.class)` ?

joaoandremartins: Sorry, just realised I asked you to look at the latest commit without having pushed it. :(
In any case, thanks for your suggestion. It makes total sense.

joaoandremartins:
Everyone please PTAL. This should be reused in #449, so would be nice to get it in soon.

joaoandremartins:
@artembilan do you mind having another pass at this PR whenever you have the time? Thanks!

PR #483, opened 2018-03-22
Spanner key

ChengyuanZhao:
adds support for composite primary keys


ChengyuanZhao:
@mp911de Hi Mark, I added an implementation for support for composite primary keys for Spanner. Please take a look :)

docs/src/main/asciidoc/spanner.adoc
```diff
@@ -34,7 +34,7 @@ Spring Data Spanner allows you to define entity classes and corresponding reposi
 ----
 @SpannerTable(name = "students")
 public class SimpleStudent {
-	@Id
+	@SpannerPrimaryKeyColumn(keyOrder = 1)
```
mp911de: How about `@PrimaryKeyColumn`?

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerOperations.java
```diff
@@ -38,6 +38,13 @@
  */
 public interface SpannerOperations {
 
+	/**
+	 * Gets the key for the given object.
+	 * @param object
+	 * @return
+	 */
+	Key getId(Object object);
```
mp911de: Typically you would not expose a `getKey(…)` method but rather handle this aspect internally. It's up to the Template API to figure out a key. That said, `findById(…)` methods accept typically `Object` (`findById(Object, Class)`).

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/mapping/SpannerCompositeKeyProperty.java
```diff
@@ -0,0 +1,217 @@
...
+import org.springframework.lang.Nullable;
+import org.springframework.util.Assert;
+
+/**
+ * Represents an persistent property just to represent Spanner primary keys, and does not
+ * correspond to actual properties of POJOs, as it might even be a composite, multi-column key.
+ *
+ * @author Chengyuan Zhao
+ */
+public class SpannerCompositeKeyProperty implements SpannerPersistentProperty {
```
mp911de: What's the difference to a regular `SpannerPersistentProperty`? Using Spanner's `Key` as composite primary key seems a good fit.

We represented for quite a time multiple primary key columns in Spring Data Cassandra with a single property but at some point, this required more conditional code than following the simple approach of representing a composite key as regular persistent entity distinguished by the property of [`isCompositePrimaryKey()`](https://github.com/spring-projects/spring-data-cassandra/blob/1d61f3c56d12fb707663d49fb470036ac12eddfa/spring-data-cassandra/src/main/java/org/springframework/data/cassandra/core/mapping/BasicCassandraPersistentProperty.java#L297). We reverted the multiple-columns-in-property pattern with Spring Data Cassandra 2.0

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/mapping/SpannerPersistentEntity.java
```diff
@@ -40,4 +42,11 @@
 	 * @return the column names.
 	 */
 	Iterable<String> columns();
+
+	/**
+	 * Gets the Key for an entity instance.
+	 * @param entity the entity whose Key ID will be returned.
+	 * @return
+	 */
+	Key getId(T entity);
```
mp911de: Creating a `Key` object from an actual instance is a converter concern, not one of the meta-model.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/mapping/SpannerPersistentEntityImpl.java
```diff
@@ -94,12 +103,49 @@ private Expression detectExpression() {
 	public void addPersistentProperty(SpannerPersistentProperty property) {
 		addPersistentPropertyToPersistentEntity(property);
 		this.columnNames.add(property.getColumnName());
+
```
mp911de: Have you seen [`BasicPersistentEntity.verify(…)`](https://github.com/spring-projects/spring-data-commons/blob/master/src/main/java/org/springframework/data/mapping/model/BasicPersistentEntity.java#L422) as an overridable hook to validate the entity state? [Cassandra](https://github.com/spring-projects/spring-data-cassandra/blob/1d61f3c56d12fb707663d49fb470036ac12eddfa/spring-data-cassandra/src/main/java/org/springframework/data/cassandra/core/mapping/BasicCassandraPersistentEntity.java#L161) uses this hook to perform various validations allowing to move the actual verification into a verifier object.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/mapping/SpannerPersistentEntityImpl.java
```diff
@@ -94,12 +103,49 @@ private Expression detectExpression() {
...
+		}
 	}
 
 	private void addPersistentPropertyToPersistentEntity(SpannerPersistentProperty property) {
 		super.addPersistentProperty(property);
 	}
 
+	@Override
+	public SpannerPersistentProperty getIdProperty() {
+		return new SpannerCompositeKeyProperty(this, getPrimaryKeyProperties());
```
mp911de: Staying in the same property implementation realm allows removing specializations such as this one.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/mapping/SpannerPrimaryKeyColumn.java
```diff
@@ -0,0 +1,41 @@
...
+/**
+ * Annotation for a {@link SpannerPersistentEntity} that allows specifying the primary key
+ * columns and their order.
+ *
+ * @author Chengyuan Zhao
+ */
+@Documented
+@Target(ElementType.FIELD)
+@Retention(RetentionPolicy.RUNTIME)
+public @interface SpannerPrimaryKeyColumn {
```
mp911de: You could consider adding a `value()` alias so the annotation can be expressed as `@PrimaryKeyColumn(1)`.

Example from Cassandra:

```java
	/**
	 * The name of the column in the table.
	 */
	@AliasFor(attribute = "name")
	String value() default "";

	/**
	 * The name of the column in the table.
	 */
	@AliasFor(attribute = "value")
	String name() default "";
```

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/SpannerRepository.java
```diff
@@ -16,14 +16,16 @@
...
+
 import org.springframework.cloud.gcp.data.spanner.core.SpannerOperations;
 import org.springframework.data.repository.PagingAndSortingRepository;
 
 /**
  * @author Ray Tsang
  * @author Chengyuan Zhao
  */
-public interface SpannerRepository extends PagingAndSortingRepository {
+public interface SpannerRepository<T> extends PagingAndSortingRepository<T, Key> {
```
mp911de: Limiting to `Key` as Id takes you the possibility to introduce user-defined primary key classes.

docs/src/main/asciidoc/spanner.adoc
ChengyuanZhao: Thank you so much for providing all this feedback! 

For this new annotation I wanted to follow the way we've been doing the other annotations like SpannerTable , SpannerColumn, etc...

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/mapping/SpannerCompositeKeyProperty.java
ChengyuanZhao: Regarding the difference between this and a regular SpannerPersistentPropertyImpl (SPPI), this one does not represent a real POJO property, and is never meant to be created or used anywhere other than as an ID property. It's also why SPPI's isIdProperty() is hard coded to give false, and this one true. Also as a result  ,because it's not a real property, has hardcoded simple implementations to return null or false to most of the PersistentProperty methods.

We also considered using the multi-columns-embedded-in-single ID class property approach, but felt it was less fitting for Spanner than a noSQL storage that has a more hierarchical data/table structure than Spanner. If were to use this "embedded" multiple-cols-in-single-property approach, we would ultimately need to flatten it into just a single table anyway, so I don't think we would be providing any real extra flexibility. 

Furthermore, I think there's real value in having the multiple primary key annotations in the POJO because it more closely mirrors what Spanner's table structure and syntax is.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/mapping/SpannerPersistentEntityImpl.java
ChengyuanZhao: Thank you! I did not know about this and will definitely take advantage.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/mapping/SpannerPrimaryKeyColumn.java
ChengyuanZhao: Good idea! Will do.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/SpannerRepository.java
ChengyuanZhao: I thought about this as well, but ultimately I don't think this takes away any generality or flexibility from the user. In addition to the points I made in the previous long comment reply, I want to add:
Even when we had the ID type as a type parameter, we ultimately relied on that ID type being acceptable into the Spanner Key type. In other words, we would do `Key.of(userGivenIdType)` and hope that it is a valid type for Key. By locking down that ID type to Key we are just being more upfront with this requirement, but not actually being more restrictive.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/mapping/SpannerPersistentEntityImpl.java
ChengyuanZhao: Perhaps I'm doing something wrong, but in some of my unit tests where the POJO has about 15 properties I'm seeing that even though I call getPersistentEntity (from AbstractMappingContext) once that several persistent entities are attempted to be created  sometimes on subsets of the properties, and each has their verify() called. Despite this , I do always get a full persistentEntity back at the end that has all the properties. The problem is that the verify() for those subsets-of-properties persistententities will fail. Does AbstractMappingContext do this to cache or optimize for performance?

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerOperations.java
ChengyuanZhao: Our other find() methods do take the Key type, and a Class, so I thought it made sense to provide the the user a way to get a Key. Since the Key type was already exposed to them I think this isn't showing them something they otherwise wouldn't need to know.

ChengyuanZhao:
@balopat PTAL

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerOperations.java
mp911de: The consequence of not exposing `Key` would be to accept `Object` instead of `Key` in `find(…)`.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/mapping/SpannerPersistentEntityImpl.java
mp911de: Not exactly sure what you mean. `MappingContext` adds typically an object graph of objects for a `PersistentEntity` that you request. Objects are derived from property types (i.e. having a `Person` class that has an `Address` property). Each entity is added through `addPersistentEntity(…)` and `.verify()` is only called after all properties were added.

Happy to look into your code if you can point me to a test that is able to reproduce the issue.

ChengyuanZhao: Here is a test that has this behavior:

in class SpannerObjectMapperImpltTests, the test writeSomeColumnsTest. I then placed a breakpoint on BasicPersistentEntity's  verify() method (in this current version I have not overridden this method). When running the test in debug mode I hit that breakpoint in verify() multiple times and each time the BasicPersistentEntity has a different subset of TestEntity's properties (except the final time when all 16 appear).

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/mapping/SpannerPersistentEntity.java
ChengyuanZhao: Done, removed from this interface

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerOperations.java
ChengyuanZhao: We have all these find() methods that use the ID type (Key in our case) because the CrudRepository's find methods are based on the ID type. I suppose we could also add find methods that take the entity.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/mapping/SpannerPersistentEntityImpl.java
mp911de: `verify()` is called on different entities (`Date`, `Timestamp`, `ByteArray`, `TestEntity`) only once.

One thing to note: `Date`, `Timestamp`, `ByteArray` look like data types that are natively supported by Spanner. Your mapping context should not represent these as persistent entities. We have an abstraction (`SimpleTypes`, see [`MongoSimpleTypes`](https://github.com/spring-projects/spring-data-mongodb/blob/148c4f9e249abc415c263f2e2d51ca5057763b48/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/mapping/MongoSimpleTypes.java) for implementation, [`MongoMappingContext`](https://github.com/spring-projects/spring-data-mongodb/blob/0be4d1345e80694bef0b8f6dc243570075b3ebd9/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/mapping/MongoMappingContext.java#L51) for usage) to control which types are considered simple ones. These types usually don't require a conversion but are understood natively by the driver.

ChengyuanZhao:
@balopat I think perhaps this one coming before my other PRs about read methods (stale reads, transactions, convenience methods, flow control) would make merging much easier, so if possible let's get this one in before the other ones  :)

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/mapping/SpannerPersistentEntityImpl.java
ChengyuanZhao: Ah thank you! I did not know that the AbstractMappingContext would be trying to create persistent entities for strange types automatically! 

Since we're not quite at the stage to support entities-within-entities, I'll override the `shouldCreatePersistentEntityFor` method and hardcode it to return false for the time being until we add that support.

docs/src/main/asciidoc/spanner.adoc
balopat: I agree with `@PrimaryKeyColumn`, see #489

balopat:
@ChengyuanZhao  I pushed an experimental commit to remove the extra composite key property. The thing is the code wasn't even using the properties stored in it. It seems like that we can safely introduce a `getKey(entity)` method on the `SpannerPersistentEntity` interface, making it even more explicit that with Spanner, this is the expected way things to work: an entity has to have a composite key that will be represented by `Key`. Let me know what you think.

ChengyuanZhao:
@balopat the `getKey` method was something I had previous but Mark mentioned was not fitting on a persistent entity, so I removed it (it was a converter concern not a meta-data model concern? ) Anyways, I'm fine with it back.

But I don't think you removed something unnecessary; you just removed functionality because now it seems like the persistent entity accessor can no longer be used to get the ID property... Furthermore, now there's just no ID property at all.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerOperations.java
ChengyuanZhao: Just wanted to follow up that this `getId` is added as a convenience for the user. I imagined that the user would likely have a SpannerOperations on hand autowired and he could get the IDs of objects easily. For example, he might retrieve objects, then deleteById some of them. 

I see what you're saying about our find/findall methods taking in Object instead of Key. The reasoning at the moment is that we haven't planned to support any type other than Key, but I can imagine in the future that we check what type that Object actually is and either use it straight if it is a Key or call Key.of() if it's a primitive type etc...

PR #484, opened 2018-03-22
Set NoCredentialsProvider on tests configuration

joaoandremartins:
Previously, if an app used Pub/Sub and another service that requires
credentials, using the emulator would prevent credentials to be normally
passed to that service.

### DIFF
--- a/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubEmulatorConfiguration.java
+++ b/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubEmulatorConfiguration.java
@@ -53,10 +53,4 @@ public class GcpPubSubEmulatorConfiguration {
 				.build();
 		return FixedTransportChannelProvider.create(GrpcTransportChannel.create(channel));
 	}
-
-	@Bean
-	@ConditionalOnMissingBean
-	public CredentialsProvider credentialsProvider() {
-		return NoCredentialsProvider.create();
-	}
 }
--- a/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubEmulatorConfiguration.java
+++ b/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubEmulatorConfiguration.java
@@ -16,8 +16,6 @@
 
 package org.springframework.cloud.gcp.autoconfigure.pubsub;
 
-import com.google.api.gax.core.CredentialsProvider;
-import com.google.api.gax.core.NoCredentialsProvider;
 import com.google.api.gax.grpc.GrpcTransportChannel;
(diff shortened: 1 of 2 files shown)


artembilan:
Travised failed with the:
```
/home/travis/build/spring-cloud/spring-cloud-gcp/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubEmulatorConfiguration.java:19:8: error: Unused import - com.google.api.gax.core.CredentialsProvider.
/home/travis/build/spring-cloud/spring-cloud-gcp/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubEmulatorConfiguration.java:20:8: error: Unused import - 
```

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubEmulatorConfiguration.java
```diff
@@ -53,10 +53,4 @@ public TransportChannelProvider transportChannelProvider(GcpPubSubProperties gcp
 				.build();
 		return FixedTransportChannelProvider.create(GrpcTransportChannel.create(channel));
 	}
-
-	@Bean
-	@ConditionalOnMissingBean
-	public CredentialsProvider credentialsProvider() {
```
artembilan: M-m-m. That's was exactly the point of this `GcpPubSubEmulatorConfiguration`. 

I would say that it would be better for all other non-involved tests to *exclude* this `GcpPubSubEmulatorConfiguration`  from their auto-configuartions.

joaoandremartins:
Oops.. ran mvn clean install but forgot the `-Dcheckstyle.skip=true` from previous runs :sweat_smile:

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubEmulatorConfiguration.java
joaoandremartins: I disagree. In my test app, I want to be able to use the emulator with other services. I might even want to test sending credentials to the emulator. In other words, emulator and credentials are unrelated.

artembilan: But we have here `@ConditionalOnMissingBean`! Why is that not enough if you create your own `CredentialsProvider` bean?

joaoandremartins: The problem is the granularity of the autoconfig. Consider the `@AutoConfigureBefore(GcpContextAutoConfiguration.class)` in this class. If the emulator is used, the _whole_ context gets "polluted" with this bogus credentials provider, and then `GcpContextAutoConfiguration` can't create the good one anymore.

PR #488, opened 2018-03-26
Added stale reads, encapsulated spanner options

ChengyuanZhao:
Added stale reads, encapsulated spanner options


ChengyuanZhao:
@balopat PTAL, this includes both the stale reads and also encapsulating the read/query options!

ChengyuanZhao:
@balopat  will do those changes ! However in the spanner client lib there is a distinction between "query" and "read" and I wanted to preserve that. for "Read" they mean single table reads where you specify columns. for "query" it's more general where you provide only a statement and it can join as many tables and calculate and create new columns in the result. They've even included 2 different option types, too.

ChengyuanZhao:
@balopat 

I don't think we can put the Sort and Pageable or Key into the read options, because those are required in the method signatures by the repository interface. But I can move the limit and offset in there.

ChengyuanZhao:
@balopat PTAL, read using secondary index has been added and it fit nicely

ChengyuanZhao:
@balopat LGTM again please? The merging of the composite key work caused previous LGTM to become stale.

PR #489, opened 2018-03-27
renaming annotations to be closer to standard

balopat:


balopat:
Standard in the sense that it is what the Spring Data community is used to.

ChengyuanZhao:
How so? It doesn't look like the mongodb project has these annotations (nor many of the other projects), though the JPA project does. I get the feeling we have some sort of unstated obligation to follow that project, but we clearly stated this is not a JPA implementation. I see project as a sibling to the JPA project, not a child, and therefore has the same relation with the JPA project as the mongodb one does with JPA.

joaoandremartins:
As I mentioned, the belonging of the annotations is already explicit from the package name, which is why this PR has my +1.

Honestly, a `SpannerTable` vs `Table` decision is so insignificant that it isn't worth that much discussion.

ChengyuanZhao:
I actually find great utility in knowing which annotation I'm using without having to lose my place in the file scrolling to the top to check which "Table" i'm importing.

joaoandremartins:
Are you using vim or something? :smile: :smile: :smile: 

Try using the control key + hovering over a type, if you're using IJ.

It is hardly a best practice to base a design decision on your choice of development tooling or coding habits IMO.

ChengyuanZhao:
Ironically that's actually exactly the point I'm making ! I am not assuming anything about IJ or some modern IDE being available to do this hover trick. We don't name variables "x" and "y", but we give them descriptive names for the same reason.

balopat:
This is not just in JPA, in spirit other Spring Data modules are doing the same and I want to align better with Spring Data. E.g. Mongo, Areospike uses `@Document` instead of `MongoDocument` and `AeroSpikeDocument`. They also use `@Query` for detecting annotated queries from the repos instead of `@XYZQuery`.

PR #493, opened 2018-03-29
Spanner custom conv

ChengyuanZhao:
Added custom conversions. It actually comes with many common conversions for free, though I also added constructors to provide user-defined conversions. 

tests added to take advantage of this.

### DIFF
--- a/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/ConversionUtils.java
+++ b/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/ConversionUtils.java
@@ -17,8 +17,12 @@
 package org.springframework.cloud.gcp.data.spanner.core.convert;
 
 import java.lang.reflect.Array;
+import java.util.Collection;
 
 import com.google.cloud.ByteArray;
+import com.google.common.collect.ImmutableSet;
+
+import org.springframework.core.convert.converter.Converter;
 
 
 /**
@@ -26,6 +30,18 @@ import com.google.cloud.ByteArray;
  * @author Chengyuan Zhao
  */
 class ConversionUtils {
+
+	/**
+	 * Converters from common types to those used by Spanner.
+	 */
+	public static final Collection<Converter> DEFAULT_SPANNER_WRITE_CONVERTERS = ImmutableSet
+			.of();
+
+	/**
+	 * Converters from common types to those used by Spanner.
+	 */
+	public static final Collection<Converter> DEFAULT_SPANNER_READ_CONVERTERS = ImmutableSet
+			.of();
 
 	static Class boxIfNeeded(Class propertyType) {
 		if (propertyType == null) {
--- a/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/MappingSpannerReadConverter.java
+++ b/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/MappingSpannerReadConverter.java
@@ -27,12 +27,14 @@ import com.google.cloud.Date;
 import com.google.cloud.Timestamp;
 import com.google.cloud.spanner.AbstractStructReader;
 import com.google.cloud.spanner.Struct;
+import com.google.cloud.spanner.Type;
 import com.google.common.collect.ImmutableMap;
 
 import org.springframework.cloud.gcp.data.spanner.core.mapping.SpannerDataException;
 import org.springframework.cloud.gcp.data.spanner.core.mapping.SpannerMappingContext;
 import org.springframework.cloud.gcp.data.spanner.core.mapping.SpannerPersistentEntity;
 import org.springframework.cloud.gcp.data.spanner.core.mapping.SpannerPersistentProperty;
+import org.springframework.data.convert.CustomConversions;
 import org.springframework.data.convert.EntityReader;
 import org.springframework.data.mapping.PersistentPropertyAccessor;
 import org.springframework.data.mapping.PropertyHandler;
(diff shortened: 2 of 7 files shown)


spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/ConversionUtils.java
```diff
@@ -27,6 +31,18 @@
  */
 class ConversionUtils {
 
+	/**
+	 * Converters from common types to those used by Spanner.
+	 */
+	public static final Collection<Converter> DEFAULT_SPANNER_WRITE_CONVERTERS = ImmutableSet
```
ChengyuanZhao: @balopat 

The DefaultConversionService we (and mongo) uses comes with about 90 typical conversions already handled.

However, I also added these here (empty for this PR) for our own conversions (such as from java Date to spanner Date). I want to add those custom conversions and tests for them in the next PR. This PR is just for setting up the classes to use them.

balopat: sounds good!

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/MappingSpannerReadConverter.java
```diff
@@ -114,14 +128,26 @@
...
 					if (ConversionUtils.isIterableNonByteArrayType(propType)) {
 						valueSet = attemptReadIterableValue(spannerPersistentProperty,
 								source, columnName, accessor);
 					}
 					else {
-						valueSet = attemptReadSingleItemValue(spannerPersistentProperty,
-								source, columnName, accessor);
+						Class sourceType = this.spannerColumnTypeToJavaTypeMapping
+								.get(source.getColumnType(columnName));
+						if (sourceType != null && canConvert(sourceType, propType)) {
```
balopat: nit: we can move this if into `attemptReadSingleItemValue`

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/MappingSpannerWriteConverter.java
```diff
@@ -156,19 +160,36 @@ private void writeProperty(WriteBuilder sink, PersistentPropertyAccessor accesso
...
 			valueSet = attemptSetIterableValue((Iterable) propertyValue, valueBinder,
 					property);
 		}
 		else {
-			valueSet = attemptSetSingleItemValue(propertyValue, valueBinder, property);
+			valueSet = attemptSetSingleItemValue(propertyValue, valueBinder,
+					propertyType);
+			if (!valueSet) {
+				for (Class targetType : this.singleItemType2ToMethodMap.keySet()) {
+					if (!canConvert(propertyType, targetType)) {
```
balopat: nit: same as above, I'd just move it inside `attemptSetSingleItemValue`

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/MappingSpannerWriteConverter.java
```diff
@@ -156,19 +160,36 @@ private void writeProperty(WriteBuilder sink, PersistentPropertyAccessor accesso
...
-		 * Due to type erasure, binder methods for Iterable properties must be manually specified.
-		 * ByteArray must be excluded since it implements Iterable, but is also explicitly
-		 * supported by spanner.
+		 * Due to type erasure, binder methods for Iterable properties must be manually
+		 * specified. ByteArray must be excluded since it implements Iterable, but is also
+		 * explicitly supported by spanner.
+		 *
+		 * Also due to type erasure, custom conversions cannot be used for iterables
+		 * because different conversions between lists of different types cannot be
+		 * distinguished.
```
balopat: I see, the problem is that targetType is just a list, so you can't just simply call `canConvert` on it, right?
It would be nice if we could just assume lists should be of a single type of element defined by `@ColumnInnerType` and then have a "map(x -> convert(x))" function to map...for that maybe we could introduce a `source` and `targetType` for the `@ColumnInnerType` annotation?

ChengyuanZhao: I agree that we could make the assumption that all items in the list would work for a single conversion function, however there are cases where we just don't have enough type information to do anything:
if the user specifies a conversion from a type Car to  List<String> and a different conversion from Car to List<Integer> (perhaps capturing different parts of the information in a Car), then our matching would just see two converters from Car to List and we wouldn't know which converter is correct (or perhaps always choose the first one that matches and never the other one). 

However, I hesitate to include inner-type annotations and information in this custom conversion picture. 
Let's say in our POJO we have a field List<List<Character>> and the user would like that to be stored in Spanner as List<String> (it's weird, but conceivable).  That would mean we need a pair of converters from List<Character> to String and back, but because of type erasure that would be just List-to-String, which means we're back at square one. 

I think we can overcome this by making our own type EnhancedConverter that holds the inner type. I think that could be the next PR as well.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/MappingSpannerWriteConverter.java
ChengyuanZhao: I will include this change in my next PR! (I didn't wanna make this LGTM stale by doing it here haha)

PR #495, opened 2018-03-30
Added common java to Spanner type converters

ChengyuanZhao:
added conversions from standard java types to the 3 spanner types


ChengyuanZhao:
@balopat I added some the list conversions stuff right after your LGTM ! would you mind looking again? (also added unit/integration tests)

PR #498, opened 2018-03-30
Spanner transactions

ChengyuanZhao:
@balopat PTAL, read-write transactions. 
the unit test is a good example of how these are used. 
This is like a real transaction where we provide a block of code and NOT just a list of mutations as in the POC.

### DIFF
--- a/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/ReadOnlyTransactionSpannerTemplate.java
+++ b/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/ReadOnlyTransactionSpannerTemplate.java
@@ -50,7 +50,7 @@ class ReadOnlyTransactionSpannerTemplate extends SpannerTemplate {
 	protected <T, U> void applyMutationTwoArgs(BiFunction<T, U, Mutation> function,
 			T arg1, U arg2) {
 		throw new SpannerDataException(
-				"A read-only transaction template cannot perform" + " mutations.");
+				"A read-only transaction template cannot perform mutations.");
 	}
 
 	@Override
--- a/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/ReadOnlyTransactionSpannerTemplate.java
+++ b/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/ReadOnlyTransactionSpannerTemplate.java
@@ -31,9 +31,10 @@ import org.springframework.cloud.gcp.data.spanner.core.mapping.SpannerMappingCon
 
 /**
  * A {@link SpannerTemplate} that performs all operations in a single transaction.
+ * This template is not intended for the user to directly instantiate.
  * @author Chengyuan Zhao
  */
-public class ReadOnlyTransactionSpannerTemplate extends SpannerTemplate {
+class ReadOnlyTransactionSpannerTemplate extends SpannerTemplate {
 
 	private ReadOnlyTransaction readOnlyTransaction;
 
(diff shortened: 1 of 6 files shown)


ChengyuanZhao:
@balopat read-only transactions also added now

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/ReadOnlyTransactionSpannerTemplate.java
```diff
@@ -0,0 +1,78 @@
...
+
+import org.springframework.cloud.gcp.data.spanner.core.convert.SpannerConverter;
+import org.springframework.cloud.gcp.data.spanner.core.mapping.SpannerDataException;
+import org.springframework.cloud.gcp.data.spanner.core.mapping.SpannerMappingContext;
+
+/**
+ * A {@link SpannerTemplate} that performs all operations in a single transaction.
+ * @author Chengyuan Zhao
+ */
+public class ReadOnlyTransactionSpannerTemplate extends SpannerTemplate {
```
balopat: nit: shouldn't this be package private too?

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/ReadOnlyTransactionSpannerTemplate.java
```diff
@@ -0,0 +1,78 @@
...
+import com.google.cloud.spanner.ReadContext;
+import com.google.cloud.spanner.ReadOnlyTransaction;
+
+import org.springframework.cloud.gcp.data.spanner.core.convert.SpannerConverter;
+import org.springframework.cloud.gcp.data.spanner.core.mapping.SpannerDataException;
+import org.springframework.cloud.gcp.data.spanner.core.mapping.SpannerMappingContext;
+
+/**
+ * A {@link SpannerTemplate} that performs all operations in a single transaction.
+ * @author Chengyuan Zhao
```
balopat: nit: we should clarify in the javadoc for both special templates that this is not meant to be directly instantiated by the user ever.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/ReadOnlyTransactionSpannerTemplate.java
```diff
@@ -0,0 +1,78 @@
...
+	@Override
+	public <T> T performReadWriteTransaction(Function<SpannerOperations, T> operations) {
+		throw new SpannerDataException("A read-only transaction template cannot be used"
+				+ " to perform sub-transactions.");
+	}
+
+	@Override
+	public <T> T performReadOnlyTransaction(Function<SpannerOperations, T> operations, SpannerReadOptions readOptions) {
+		throw new SpannerDataException("A read-only transaction template cannot be used"
+				+ " to perform sub-transactions.");
```
balopat: nit (please apply it for others too): "A read-only transaction is already under execution. Opening sub-transactions is not supported!"

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/ReadOnlyTransactionSpannerTemplate.java
ChengyuanZhao: yes, will fix!

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/ReadOnlyTransactionSpannerTemplate.java
ChengyuanZhao: yes will add!

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/ReadOnlyTransactionSpannerTemplate.java
ChengyuanZhao: fixed!

ChengyuanZhao:
@balopat changes made

PR #499, opened 2018-03-30
Enables autoconfiguration for EnableSpannerRepositories

### LINKED ISSUE #490: Move @EnableSpannerRepositories to GcpSpannerAutoconfiguration if possible from sample
`@EnableSpannerRepositories` is a Spring Data feature that enables the discovery of repositories.
We need to explore if it's possible to setup the annotation on the autoconfig level, thus potentially save the user this extra step to setup a spring data spanner application. 

The best place to start proving this is the Spring Data Spanner sample project.

joaoandremartins:
Users won't have to explicitly provide the annotation for it to be
enabled.

Fixes #490


spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/spanner/GcpSpannerAutoConfigurationTests.java
```diff
@@ -38,7 +37,8 @@
 
 	private ApplicationContextRunner contextRunner = new ApplicationContextRunner()
 			.withConfiguration(AutoConfigurations.of(GcpSpannerAutoConfiguration.class,
-					GcpContextAutoConfiguration.class))
+					GcpContextAutoConfiguration.class,
+					SpannerRepositoriesAutoConfiguration.class))
```
joaoandremartins: @artembilan do you know why this test is failing with:

```
Caused by: java.lang.IllegalStateException: Unable to retrieve @EnableAutoConfiguration base packages
	at org.springframework.boot.autoconfigure.AutoConfigurationPackages.get(AutoConfigurationPackages.java:76)
	at org.springframework.boot.autoconfigure.data.AbstractRepositoryConfigurationSourceSupport.getBasePackages(AbstractRepositoryConfigurationSourceSupport.java:79)
```

I can't add `@EnableAutoConfiguration` to `SpannerRepositoriesAutoConfiguration` since it will enable every other autoconfiguration in the package.
I added `SpannerRepositoriesAutoConfiguration` to `spring.factories`, shouldn't that be enough?

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/spanner/GcpSpannerAutoConfigurationTests.java
```diff
@@ -58,7 +58,6 @@ public void testTestRepositoryCreated() {
 		});
 	}
 
-	@EnableSpannerRepositories
```
artembilan: To answer to your question, @joaoandremartins .
Add here `@AutoConfigurationPackage` instead.

The main point is to let to Spring Boot to know where to find `Repository` (and other similar) classes.
In this case we are lucky that our `TestRepository` in the same package. :smile:

ChengyuanZhao:
Our integration tests for Spanner at `org.springframework.cloud.gcp.data.spanner.test.IntegrationTestConfiguration` has the `@EnableSpannerRepositories` for it to work currently. You should remove it from that class and if the integration tests still work then it should be good to go!

joaoandremartins:
@ChengyuanZhao the integration tests don't use auto-configuration, so I'm afraid that `@EnableJpaRepositories` will need to continue there :(

joaoandremartins:
@artembilan PTAL

PR #500, opened 2018-04-02
SpannerPersistentEntity can generate DDL 

ChengyuanZhao:
Added feature for spanner persistent entity to generate DDL.


spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/ConversionUtils.java
```diff
@@ -21,22 +21,126 @@
...
 import org.springframework.lang.Nullable;
+import org.springframework.util.Assert;
 
 /**
  * @author Balint Pato
  * @author Chengyuan Zhao
  */
 public class ConversionUtils {
 
+	public static final Map<Type, Class> SPANNER_COLUMN_TYPE_TO_JAVA_TYPE_MAPPING =
```
ChengyuanZhao: this was previously in the read converter, it was just moved here.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/ConversionUtils.java
```diff
@@ -21,22 +21,126 @@
...
+		JAVA_TYPE_TO_SPANNER_COLUMN_TYPE_MAPPING = builder.build();
+	}
+
+	private static String getTypeDDLString(Type type) {
+		Assert.notNull(type, "A valid Spanner column type is required.");
+		if (type.getCode() == Code.ARRAY) {
+			return "ARRAY<" + getTypeDDLString(type.getArrayElementType()) + ">";
+		}
+		return type.toString()
+				+ (type.getCode() == Code.STRING || type.getCode() == Code.BYTES ? "(MAX)"
```
ChengyuanZhao: we could add options to specify the size, but using MAX seems ok since it doesn't actually use any extra space in Spanner. however, if the user did want exceptions to be thrown for too-long inputs, then we do lose that functionality. It would be pretty simple to add via an annotation on the persistent property.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/ConversionUtils.java
```diff
@@ -21,22 +21,126 @@
...
+						: "");
+	}
+
+	private static Class getFirstFullyConvertableType(SpannerConverter spannerConverter,
+			Class originalType, boolean isArrayInnerType) {
+		Set<Class> spannerTypes = (isArrayInnerType
+				? MappingSpannerWriteConverter.iterablePropertyType2ToMethodMap
```
ChengyuanZhao: just wanted to mention that it doesn't matter if we're using the Read or Write method mappings here to see the available types, since we check that it's convertible from both directions in the following lines

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerOperations.java
```diff
@@ -37,6 +37,23 @@
...
 
+	/**
+	 * Gets the DDL string to create the table for the given entity in Spanner. This is just one of
+	 * the possible schemas that can support the given entity type. The specific schema is determined
+	 * by the configured property type converters used by the read and write methods in
+	 * this SpannerOperations and will be compatible with those methods.
+	 * @param entityClass the entity type.
+	 * @return the DDL string.
+	 */
+	String getCreateTableDDLString(Class entityClass);
```
balopat: I think this feature should go together with introducing a `SpannerAdminTemplate` that wraps the `DatabaseAdminClient` and then these could go on that template instead of `SpannerOperations`.

ChengyuanZhao: At first I thought so too, but generating just the *string* i dont think is a database admin client responsibility. I think the database admin's responsibility is running this string and actually creating the table. This is supported by:

- generating the schema requires extensive use of the converter
- requires extensive access to the underlying entities
- requires use of converter utils

All of these things would be strange to have inside just database admin class. I think the admin template should accept these strings, but not be in charge of actually generating them.

ChengyuanZhao:
@balopat how about getting this one in for M3? This one will be useful once we get to the parent-child PR and adding the ability to generate all the CREATE statements for the user if they request.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerOperations.java
balopat: Hmmm...maybe we should just create a separate class for it, with the right collaborators, e.g. `MappingSchemaGenerator`, it really doesn't belong to the DML world.

ChengyuanZhao: sounds good. coming right up!

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/ConversionUtils.java
balopat: Let's add the length annotation. It shouldn't be much and then we don't have to touch this again/create a new issue for it, etc.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/SpannerConverter.java
```diff
@@ -82,4 +82,44 @@
...
+
+	/**
+	 * Determines if this converter can handle the given types for writing to Spanner
+	 * @param type the type of the property that data will be converted from Spanner in to
+	 * @param spannerSupportedType the desired Spanner supported type for reading.
+	 * @return true if the conversion is possible. false otherwise
+	 * @throws org.springframework.cloud.gcp.data.spanner.core.mapping.SpannerDataException if
+	 * the given desired Spanner supported type is not actually a supported Spanner type
+	 */
+	boolean canHandlePropertyTypeForSingularRead(Class type, Class spannerSupportedType);
```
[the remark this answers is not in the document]
ChengyuanZhao: Do you mean to add the differences in these 4 functions as parameters to a single function?

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/ConversionUtils.java
```diff
@@ -21,22 +21,121 @@
...
+		Assert.notNull(type, "A valid Spanner column type is required.");
+		if (type.getCode() == Code.ARRAY) {
+			return "ARRAY<" + getTypeDDLString(type.getArrayElementType()) + ">";
+		}
+		return type.toString()
+				+ (type.getCode() == Code.STRING || type.getCode() == Code.BYTES ? "(MAX)"
+						: "");
+	}
+
+	private static Class getFullyConvertableType(SpannerConverter spannerConverter,
```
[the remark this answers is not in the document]
ChengyuanZhao: rename done. Also added the length annotation.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/SpannerConverter.java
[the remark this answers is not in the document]
balopat: * have these as private functions on SpannerConverter
* expose a `canConvert` function that by the way you already introduced in another PR

The following piece of code in `SchemaUtils` knows too much about conversion and can be just turned into a `converter.canConvert(...)` call: 

````
if (isIterableInnerType
+					&& spannerConverter.canHandlePropertyTypeForArrayRead(originalType,
+							spannerType)
+					&& spannerConverter.canHandlePropertyTypeForArrayWrite(originalType,
+							spannerType)) {
+				ret = spannerType;
+				break;
+			}
+			else if (!isIterableInnerType
+					&& spannerConverter.canHandlePropertyTypeForSingularRead(originalType,
+							spannerType)
+					&& spannerConverter.canHandlePropertyTypeForSingularWrite(
+							originalType, spannerType)) {
+				ret = spannerType;
+				break;
+			}

````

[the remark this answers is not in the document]
ChengyuanZhao: The "forArray" vs "forSinglular" distinction is important because Spanner can't handle the same types for array as for singular objects. For example, `double[]` can be a singular write but cannot be a an array write. I could combine the two forArrays and the two forSingulars, though. But I would need to name them "canConvertReadAndWriteForArray" and "canConvertReadAndWriteForSingular" to show that we can convert in both directions. How does that sound?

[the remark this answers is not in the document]
balopat: We can still have the 4 methods as private methods. Why do you need them to be non-private?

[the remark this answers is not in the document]
balopat: I am trying to simplify the API - the narrower the surface the better - makes it easier to implement for users for example if they wanted to extend it. I see that currently introducing these finer grained methods on the Converter interface is not justified.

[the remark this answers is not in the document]
ChengyuanZhao: good point! changed

ChengyuanZhao:
Thanks! sample and refdoc issue: #609

PR #505, opened 2018-04-04
Spanner parent-child relationships

ChengyuanZhao:
When a property is detected as a type also annotated Table, that means it is a Spanner table, and we check that its PKs obey the Spanner parent-child rules.

The reason is because we will soon take advantage of Spanner's ON DELETE CASCADE and INTERLEAVE features for managing parent-child (1-to-many) relationships. 

I also changed the mutation factory return types to give Lists of mutations because we will soon be issuing multiple mutations per write, since a mutation can only target a single table, and parents with children will need this.

However, Delete operations still use a single mutation, because Spanner's delete cascade feature will  take care of deleting children automatically.


ChengyuanZhao:
FYI @meltsufin 

The master branch has progressed a lot and now merging is impractical, so I will be starting a new PR eventually and closing this one. 

We recently discussed relationships in spring data  with Oliver, and arrived at some common points among spring data integrations:
- eager retrieval of children should be the default implementation, and it is OK to let the user know the limitations of it. this is the recommendation for single children or multiple children in a collection.
- lazy retrieval can be very complicated especially after involving other features like transactions, and should definitely come second if at all.
- the idea of aggregates (parents) is strong in spring data, and relationships should be reflected in POJOs one-way from aggregate to its sub-items (children). 
- recommend users use identifier properties instead of POJO references , which forces the user to do retrieval themselves.

Issue #19, opened 2017-05-26
Support PubSub topic resolution from message header


joaoandremartins:
Header would have precedence over `PubSubTemplate`'s `topic` field.

Some good examples are the Kafka and Kinesis adapters:
https://github.com/spring-projects/spring-integration-aws/blob/master/src/main/java/org/springframework/integration/aws/outbound/KinesisMessageHandler.java#L93
https://github.com/spring-projects/spring-integration-kafka/blob/master/src/main/java/org/springframework/integration/kafka/outbound/KafkaProducerMessageHandler.java#L147

More discussion [here](https://github.com/spring-cloud/spring-cloud-gcp/pull/14/files/a379cc599773c86a402b0a6b365331c8bdc39e7f#diff-b9e1539c9f93aaee09ea016f3c777faa).

artembilan:
The `topicExpression` should be as an option as well. The same Kinesis and Kafka sample.
All that is based on the [SpEL](http://docs.spring.io/spring/docs/4.2.9.RELEASE/spring-framework-reference/html/expressions.html) and its support in [Spring Integration](http://docs.spring.io/spring-integration/reference/html/spel.html).

PR #508, opened 2018-04-05
enable dynamic topics via Message Header

rpahli:
enable to override default topic via Message Header.

### DIFF
--- a/docs/src/main/asciidoc/spring-integration-pubsub.adoc
+++ b/docs/src/main/asciidoc/spring-integration-pubsub.adoc
@@ -81,7 +81,7 @@ On automatic acking, a message is acked with GCP Pub/Sub if the adapter sent it 
 no exceptions were thrown.
 If a `RuntimeException` is thrown while the message is processed, then the message is nacked.
 On manual acking, the adapter attaches an `AckReplyConsumer` object to the `Message` headers, which
-users can extract using the `GcpHeaders.ACKNOWLEDGEMENT` key and use to (n)ack a message.
+users can extract using the `GcpPubSubHeaders.ACKNOWLEDGEMENT` key and use to (n)ack a message.
 
 [source,java]
 ----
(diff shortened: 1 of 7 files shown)


joaoandremartins:
Thanks for the contribution, @rpahli!
I think this fixes #19, even though maybe only partially due to the lack of topic expression support.
@artembilan wdyt?

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/GcpHeaders.java
```diff
@@ -26,4 +26,6 @@
 	private static final String PREFIX = "gcp_";
 
 	public static final String ACKNOWLEDGEMENT = PREFIX + "pub_sub_acknowledgement";
+
+	public static final String DESTINATION = PREFIX + "destination_topic";
```
artembilan: I would just call this constant (and its value)  as `topic`:
```
public static final String TOPIC = PREFIX + "topic";
```

Seems for me the name of the class is too broad as well - `GcpPubSubHeaders` would sound better.
The reason that in other modules we may have a different set of headers and, therefore, there will be a new support class for headers constants. BTW the ``prefix should might be as `gcp_pubsub_` as well. But this can be done as a separate issue anyway.

Right, I'd prefer to see the constant and its value renamed.

rpahli:
ok I updated class to GcpPubSubHeaders and renamed the Topic const.

joaoandremartins:
@artembilan it seems to be an issue only with forks.

I think @balopat is looking into a way of disabling integration tests for forks, which should fix the situation.

joaoandremartins:
I didn't add a fix to #19 because I intend to go and add the expressions to the topic.

But I did not pay attention to the commit message. I'll fix it up next time.

Issue #501, opened 2018-04-03
Improve CredentialsProvider logic


joaoandremartins:
It would be nice if CredentialsProvider handling was such that the CF piece didn't show up in the regular auto-config at all. Currently, the logic works as follows:

. If the service is bound in CF, we use the information from VCAP
. If the user has set a property, we use that
. Otherwise, we use the global CredentialsProvider from the context

If the mapping between 1. and 2. is one-to-one, one idea is to create a special PropertySource that sets the property that is expected in 2. with the information from VCAP. This has several advantages:

- The code isn't aware of the CF specific bits
- The Environment clearly states why credentials are configured this way (you can see the property source using /actuator/env)
If that's not possible, then we need some sort of Provider<CredentialsProvider> where the implementation would apply the logic above. In this scenario, it would be beneficial that the CF pieces was registered independantly so that the provider doesn't have to know about it. An advantage of that is that if the application doesn't run in CF, then that piece isn't registered at all and we don't have to play with @Autowired(required=false). Unrelated but ObjectProvider<T>.getIfAvailable() is what we use nowadays for this.

artembilan:
Also see #538

joaoandremartins:
Given that there isn't any Cloud Foundry specific code anymore (besides the post processor), I think the main goal of this issue is to use a single `CredentialsProvider` everywhere and avoid [code like this](https://github.com/spring-cloud/spring-cloud-gcp/blob/master/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java#L79).

The only solution I can think of for this problem is to have multiple qualified `CredentialsProvider` (e.g., `pubsub-credentials-provider` or even `CredentialsProvider` sub-types (e.g., `PubSubCredentialsProvider`). I think the latter would probably be better, since we could rely on Java types and not have to manage beans by their name.

However, the sub-type solution will still require similar code to the one we want to avoid (we can't rely on `@ConditionalOnMissingBean`) and the advantages are not obvious, in the current state of the repo (i.e., we don't really need a `PubSubCredentialsProvider` anywhere outside of auto-configuration).

Thus, I can't think of another solution for this problem. If anyone can't provide a good solution for this problem, then I suggest we close this issue.

@meltsufin @artembilan @snicoll @balopat @ChengyuanZhao FYI

snicoll:
[...]

PR #509, opened 2018-04-05
Extracts Pub/Sub parameters from Cloud Foundry's environment

joaoandremartins:


joaoandremartins:
@artembilan @snicoll this PR needs more work for obvious reasons.
But please take a look and let me know if this is the direction you guys were thinking of.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/core/cloudfoundry/GcpCloudFoundryEnvironmentPostProcessor.java
```diff
@@ -0,0 +1,129 @@
...
+			Map<String, Object> vcapMap =
+					this.parser.parseMap(environment.getProperty(VCAP_SERVICES_ENVVAR));
+
+			Properties gcpCfPubSubProps = new Properties();
+
+			Arrays.stream(GcpCfService.values()).forEach(cfService ->
+					gcpCfPubSubProps.putAll(retrieveCfProperties(vcapMap, cfService)));
+
+			environment.getPropertySources()
+					.addLast(new PropertiesPropertySource("gcpCf", gcpCfPubSubProps));
```
snicoll: I think that's wrong. You want that the cloud information overrides anything you've set in your own config right? That's what's happening for the  `DataSource` support in CF (your `application.properties` refer to some DB you use locally but you want to use the one provided by the cloud when you push your app there.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/core/cloudfoundry/GcpCloudFoundryEnvironmentPostProcessor.java
```diff
@@ -0,0 +1,129 @@
...
+
+		public Map<String, String> getCfPropNameToGcp() {
+			return this.cfPropNameToGcp;
+		}
+	}
+
+	@Override
+	public void postProcessEnvironment(ConfigurableEnvironment environment,
+			SpringApplication application) {
+		if (CloudPlatform.CLOUD_FOUNDRY.isActive(environment)) {
```
snicoll: You might remove that check (and just check for the presence of VCAPs) as it would ease unit testing of this feature. Practically that code is doing that but it might change in the future so doing it explicitly is better IMO

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/core/cloudfoundry/GcpCloudFoundryEnvironmentPostProcessor.java
joaoandremartins: Ack. Should we take the same care to not override the command line parameters, like in https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/cloud/CloudFoundryVcapEnvironmentPostProcessor.java#L126 ?

joaoandremartins:
A couple of points about this PR:
- Unit tests still to come in this PR,
- Documentation to come in a separate PR;
- SQL integration will be dealt with separately in another PR, since it's a bit hairier.

spring-cloud-gcp-core/src/main/java/org/springframework/cloud/gcp/core/DefaultCredentialsProvider.java
```diff
@@ -75,13 +86,20 @@ public Credentials getCredentials() throws IOException {
...
 		if (!StringUtils.isEmpty(providedLocation)) {
 			this.wrappedCredentialsProvider = FixedCredentialsProvider
 					.create(GoogleCredentials.fromStream(
 							providedLocation.getInputStream())
 							.createScoped(scopes));
 		}
+		else if (!StringUtils.isEmpty(encodedKey)) {
+			this.wrappedCredentialsProvider = FixedCredentialsProvider.create(
+					GoogleCredentials.fromStream(
+							new ByteArrayInputStream(Base64.getDecoder().decode(encodedKey)))
```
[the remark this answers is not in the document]
joaoandremartins: Hmmm would you mind providing an alternative?
I could extract `GoogleCredentials.fromStream(...)`, but then the third branch doesn't use it, so it wouldn't look very pretty...

joaoandremartins:
Unit tests were merged with this PR. It's ready for full review.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubProperties.java
```diff
@@ -47,6 +47,14 @@
 	@NestedConfigurationProperty
 	private final Credentials credentials = new Credentials(GcpScope.PUBSUB.getUrl());
 
+	/** The specific topic that should be used. Useful for environments like Cloud Foundry. */
+	private String topicName;
```
meltsufin: How's this supposed to be used because it doesn't seem to be used by this project at all? Maybe we just need a better documentation, but this may confuse regular users.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/core/cloudfoundry/GcpCloudFoundryEnvironmentPostProcessor.java
```diff
@@ -0,0 +1,154 @@
...
+
+			environment.getPropertySources()
+					.addFirst(new PropertiesPropertySource("gcpCf", gcpCfPubSubProps));
+		}
+	}
+
+	private static Properties retrieveCfProperties(Map<String, Object> vcapMap,
+			GcpCfService service) {
+		Properties properties = new Properties();
+		List<Object> pubSubBindings = (List<Object>) vcapMap.get(service.getCfServiceName());
```
artembilan: I think the name of the variable is a bit misleading.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubProperties.java
artembilan: Well, that should be as good addition anyway.
We have similar approach for RabbitMQ and Kafka auto-configurations: https://github.com/spring-projects/spring-boot/blob/v2.0.1.RELEASE/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/KafkaProperties.java#L763

Let's really address it ASAP after merging this PR!

spring-cloud-gcp-core/src/main/java/org/springframework/cloud/gcp/core/DefaultCredentialsProvider.java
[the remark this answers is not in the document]
artembilan: OK. Never mind

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubProperties.java
meltsufin: OK, I can kind of understand the purpose of a default topic, but then this field should probably be renamed to reflect that.

balopat:
added some nits to the PR, let me know if you really don't agree with them. Otherwise LGTM.

balopat:
the "default topic/subscription" commit went a bit too far, I'll fix that in a sec.

docs/src/main/asciidoc/pubsub.adoc
```diff
@@ -115,15 +115,15 @@ starter.
 
 [source,java]
 ----
-public Topic createTopic(String topicName)
+public Topic createTopic(String defaultTopicName)
```
artembilan: I think this is coincidental after renaming property in the `GcpPubSubProperties`.
Should be reverted. We really don't talk here about a `default` one.

Thanks

balopat: agreed, I noticed it after pushing, fix is already in.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/core/cloudfoundry/GcpCloudFoundryEnvironmentPostProcessor.java
```diff
@@ -0,0 +1,153 @@
...
+			return properties;
+		}
+
+		Map<String, Object> serviceBinding = (Map<String, Object>) serviceBindings.get(0);
+		Map<String, String> credentialsMap = (Map<String, String>) serviceBinding.get("credentials");
+		String prefix = SPRING_CLOUD_GCP_PROPERTY_PREFIX + service.name() + ".";
+		service.getCfPropNameToGcp().forEach(
+				(cfPropKey, gcpPropKey) -> properties.put(
+						prefix + gcpPropKey,
+						credentialsMap.get(cfPropKey)));
```
artembilan: Is there some confirmation that `topic_name` or `topic-name` really maps to the `GcpPubSubProperties.defaultTopicName` ?

Thanks

joaoandremartins: Good catch!

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/core/cloudfoundry/GcpCloudFoundryEnvironmentPostProcessor.java
```diff
@@ -0,0 +1,153 @@
...
+
+		pubsub("google-pubsub",
+				ImmutableMap.of("ProjectId", "project-id",
+				"PrivateKeyData", "credentials.encodedKey",
+				"topic_name", "default-topic-name",
+				"subscription_name", "default-subscription-name")),
+		storage("google-storage",
+				ImmutableMap.of("ProjectId", "project-id",
+				"PrivateKeyData", "credentials.encodedKey",
+				"bucket_name", "bucket-name")),
```
artembilan: Well, the same applies to a new `defaultBucketName` as well :smile: !

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/core/cloudfoundry/GcpCloudFoundryEnvironmentPostProcessor.java
```diff
@@ -0,0 +1,153 @@
...
+
+	private JsonParser parser = JsonParserFactory.getJsonParser();
+
+	private int order = ConfigFileApplicationListener.DEFAULT_ORDER - 1;
+
+	private enum GcpCfService {
+
+		pubsub("google-pubsub",
+				ImmutableMap.of("ProjectId", "project-id",
+				"PrivateKeyData", "credentials.encodedKey",
```
artembilan: Shouldn't it be like `credentials.encoded-key` for consistency with all other *lower-case-dashed* ?

I've been told recently that Spring Boot `@ConditionalOnProperty` doesn't support relaxed values anymore.
It is now only like this:
```
	 * Use the dashed notation to specify each property, that is all lower case with a "-"
	 * to separate words (e.g. {@code my-long-property}).
	 * @return the names
	 */
	String[] name() default {};
```

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
```diff
@@ -76,7 +77,10 @@ public GcpPubSubAutoConfiguration(GcpPubSubProperties gcpPubSubProperties,
 		this.finalProjectIdProvider = gcpPubSubProperties.getProjectId() != null
 				? gcpPubSubProperties::getProjectId
 				: gcpProjectIdProvider;
-		this.finalCredentialsProvider = gcpPubSubProperties.getCredentials().getLocation() != null
+		Credentials credentialProperties = gcpPubSubProperties.getCredentials();
+		this.finalCredentialsProvider =
```
artembilan: Something is wrong going on here.
Would be better to have a `CredentialsProvider` as a conditional bean and inject it into all the `@Bean` method below.
With this approach you eliminate a possibility for the end-user to provide his/her own `CredentialsProvider` implementation.

This should be resolved in the separate issue though.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/storage/GcpStorageAutoConfiguration.java
```diff
@@ -57,11 +58,14 @@
 	public static Storage storage(CredentialsProvider credentialsProvider,
 			GcpStorageProperties gcpStorageProperties,
 			GcpProjectIdProvider projectIdProvider) throws IOException {
+		Credentials propertiesCredentials = gcpStorageProperties.getCredentials();
 		return StorageOptions.newBuilder()
-				.setCredentials(gcpStorageProperties.getCredentials().getLocation() != null
+				.setCredentials(propertiesCredentials.getLocation() != null
+						|| propertiesCredentials.getEncodedKey() != null
 						? new DefaultCredentialsProvider(gcpStorageProperties).getCredentials()
 						: credentialsProvider.getCredentials())
```
artembilan: I think this logic must be changed as well.
The user provided `CredentialsProvider` must have a precedence independently of the properties provided.
Therefore would be better to refactor this auto-configuration to extract a `CredentialsProvider` conditional bean population.

Different story as well though.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
meltsufin: @artembilan Can you please file an issue for this.

joaoandremartins: Feel free to comment on #501.

Though to address your comment, this behaviour is desired -- credentials from properties have precedence over the `CredentialsProvider` bean.

artembilan: > credentials from properties have precedence over the `CredentialsProvider` bean.

That's not correct, especially if I would like to test auto-config against some `MockCredentialsProvider` bean in the test `@Configuration`. That's not how things work in Spring Boot. And that's why we have all those `@Conditional...`

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/core/cloudfoundry/GcpCloudFoundryEnvironmentPostProcessor.java
joaoandremartins: Good catch! I thought I had gone this way since the beginning for encoded key...

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
artembilan: Fair enough! Let's continue discussion there!
Would be good theme for the next call. Today?

joaoandremartins:
@artembilan let me know if you find anything else or if you're OK to approve.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/storage/GcpStorageProperties.java
```diff
@@ -31,11 +31,22 @@
 public class GcpStorageProperties extends GoogleStorageProtocolResolverSettings implements
 		CredentialsSupplier {
 
+	/** The default bucket that should be used. Useful for environments like Cloud Foundry. */
```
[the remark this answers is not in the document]
joaoandremartins: Why do you say so?
It is valid Javadoc format, and I can't find anything in the [Spring style guide](https://github.com/spring-projects/spring-framework/wiki/Code-Style) that says otherwise..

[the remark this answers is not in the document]
artembilan: M-m-m. 
At least this is not consistent what you have so far in other similar new properties in this PR...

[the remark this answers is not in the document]
snicoll: +1 with @artembilan 

Also we consistently do not start configuration keys with "The", "A". It would be nice if you follow the same consistency (or any consistency that is reasonable).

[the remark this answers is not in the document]
artembilan: May you confirm that this JavaDocs really becomes as a proper description hint for the configuration properties assistant?
Thanks

artembilan:
Hm.
Have you seen Travis report?
```
Wanted but not invoked:
object.onSuccess(<any>);
-> at org.springframework.cloud.gcp.autoconfigure.pubsub.it.PubSubChannelAdaptersIntegrationTests.lambda$sendAndReceiveMessagePublishCallback$3(PubSubChannelAdaptersIntegrationTests.java:175)
Actually, there were zero interactions with this mock.
	at org.springframework.cloud.gcp.autoconfigure.pubsub.it.PubSubChannelAdaptersIntegrationTests.lambda$sendAndReceiveMessagePublishCallback$3(PubSubChannelAdaptersIntegrationTests.java:175)
```

Looks like there is some wrong Mockito usage.

Not related to this PR I guess anyway

joaoandremartins:
Travis is failing because the Pub/Sub integration tests are flaky.
Haven't been able to understand why, yet...

PR #510, opened 2018-04-05
Spanner getkey moved, added key conveniences 

ChengyuanZhao:
Moved getkey to its own schema operations class. added conveniences for users to make keys


spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/spanner/GcpSpannerAutoConfiguration.java
```diff
@@ -133,4 +134,12 @@ public SpannerMutationFactory spannerMutationFactory(
 			SpannerMappingContext spannerMappingContext) {
 		return new SpannerMutationFactoryImpl(spannerConverter, spannerMappingContext);
 	}
+
+	@Bean
+	@ConditionalOnMissingBean
+	public MappingSchemaOperations mappingSchemaGenerator(
```
balopat: How about`SpannerSchemaOperations`? Nevertheless, the method name should be synced up with the class name.

ChengyuanZhao: done

ChengyuanZhao:
@balopat travis rolled back a deployment and it's all good now!

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/support/SimpleSpannerRepository.java
```diff
@@ -129,4 +135,22 @@ public Iterable findAll(Sort sort) {
...
+		Key k;
+		boolean isIterable = Iterable.class.isAssignableFrom(key.getClass());
+		boolean isArray = Object[].class.isAssignableFrom(key.getClass());
+		if (isIterable || isArray) {
+			Key.Builder kb = Key.newBuilder();
+			for (Object keyPart : (isArray ? (Arrays.asList((Object[]) key))
+					: ((Iterable) key))) {
+				kb.appendObject(keyPart);
+			}
+			k = kb.build();
```
balopat: Should we throw an error when we pass in an empty array as a key?

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/admin/SpannerSchemaOperations.java
```diff
@@ -0,0 +1,56 @@
...
+
+import org.springframework.cloud.gcp.data.spanner.core.convert.SpannerConverter;
+import org.springframework.cloud.gcp.data.spanner.core.mapping.SpannerMappingContext;
+import org.springframework.cloud.gcp.data.spanner.core.mapping.SpannerPersistentEntity;
+import org.springframework.util.Assert;
+
+/**
+ * @author Chengyuan Zhao
+ */
+public class SpannerSchemaOperations {
```
balopat: Apologies - but I have a last rename idea: `SpannerEntityUtils` - because this is not about resource management, or exposing existing functionality. It's more like a helper class around entities, which if we wanted to we could even implement as a similar bytecode generator as Lombok does - getId() (and later getCreateDDL(), getDropDDL()) could live _on the entity class_ itself...

balopat: oh and let's add a javadoc to it too, so that users know what this is for. Thank you!

ChengyuanZhao: @balopat PTAL ! changes made

PR #511, opened 2018-04-06
GCP Storage integration tests

ChengyuanZhao:


spring-cloud-gcp-storage/src/test/java/org/springframework/cloud/gcp/storage/GoogleStorageIntegrationTests.java
```diff
@@ -0,0 +1,116 @@
...
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assume.assumeThat;
+
+/**
+ * @author Chengyuan Zhao
+ */
+@RunWith(SpringRunner.class)
+@SpringBootTest(classes = { GoogleStorageIntegrationTestsConfiguration.class })
```
artembilan: Looks like we don't need here Spring Boot overhead.
The simple `ContextConfiguration(classes)` should be enough.

And at this point I wonder how Spring Boot is here in the `spring-cloud-gcp-storage` as a dependency at all...

spring-cloud-gcp-storage/src/test/java/org/springframework/cloud/gcp/storage/GoogleStorageIntegrationTestsConfiguration.java
```diff
@@ -0,0 +1,69 @@
...
+			return new DefaultCredentialsProvider(Credentials::new);
+		}
+		catch (IOException e) {
+			throw new RuntimeException(e);
+		}
+	}
+
+	@Bean
+	public GoogleStorageProtocolResolverSettings googleStorageProtocolResolverSettings() {
+		return new GoogleStorageProtocolResolverSettings();
```
artembilan: M-m-m.
I think we should consider to change logic in the `GoogleStorageProtocolResolver` do not require this void bean.
The defaults can come even without such a bean.

Different story though: may be resolved as a separate fix.

spring-cloud-gcp-storage/pom.xml
```diff
@@ -28,5 +28,10 @@
 			<artifactId>spring-integration-file</artifactId>
 			<optional>true</optional>
 		</dependency>
-	</dependencies>
+    <dependency>
```
artembilan: Wrong indents for this new dependency definition.
We should use `tabs` everywhere unless it is ASCIIDoc

ChengyuanZhao: fixed! 

About that SpringBootTest annotation it was mistakenly used when just ContextConfig was necessary. I fixed the other location where I copied it from.

spring-cloud-gcp-storage/pom.xml
```diff
@@ -1,32 +1,51 @@
...
-			<groupId>com.google.cloud</groupId>
-			<artifactId>google-cloud-storage</artifactId>
-		</dependency>
-		<dependency>
-			<groupId>org.springframework.integration</groupId>
-			<artifactId>spring-integration-file</artifactId>
-			<optional>true</optional>
-		</dependency>
-	</dependencies>
+<project xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
```
joaoandremartins: Wrong indentation

PR #515, opened 2018-04-07
Reverting to 2579918763ca2f532915a6d2af3f4778c28cc4cc

ChengyuanZhao:


ChengyuanZhao:
Still trying to figure out why that one pubsub emulator integration test fails on this PR even though it's exactly the same state as 
2579918763ca2f532915a6d2af3f4778c28cc4cc 
and that one passes Travis just fine :

image

PR #516, opened 2018-04-07
Integrate Error Prone checking to spring-gcp

nglorioso:
Integrates Error Prone (http://errorprone.info) to the compilation process for spring-gcp.

This detects issues at compile time. 

Here, we make an explicit capture of an ignored return value with a comment: http://errorprone.info/bugpattern/CheckReturnValue
Additionally, in a test where a Stream is opened with Files.list(), we use try-with-resources to ensure the directory stream is closed (per documentation): http://errorprone.info/bugpattern/StreamResourceLeak

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -111,6 +111,8 @@
 				<plugin>
 					<artifactId>maven-compiler-plugin</artifactId>
 					<configuration>
+						<compilerId>javac-with-errorprone</compilerId>
+						<forceJavacCompilerUse>true</forceJavacCompilerUse>
 						<compilerArgs>
 							<!-- Enable all warnings -->
 							<compilerArg>-Xlint:all</compilerArg>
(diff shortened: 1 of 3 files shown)


joaoandremartins:
Thanks @nglorioso. It was a conscious decision to be using checkstyle instead of error prone.

@meltsufin @artembilan wdyt about adding error prone to the project, besides checkstyle? Overkill?

meltsufin:
@joaoandremartins Errorprone addresses somewhat different concerns than checkstyle. I think we should add this. Thanks for the contribution @nglorioso!

artembilan:
Yes, I think we can add this as well.

@nglorioso ,

Before starting review we need a confirmation from you that CLA is signed: https://cla.pivotal.io/

Thank you so far!

meltsufin:
@artembilan, @nglorioso is a Googler. So, I think he's covered by the corporate CLA.

artembilan:
OK. But that isn't clear by his GitHub account :smile:

meltsufin:
True! @nglorioso I think you'll need to make your organizations membership public for this to be accepted.

nglorioso:
Thanks for the thumbs up! I've made that membership public now.

balopat:
@nglorioso can you please update your branch with the latest on `master`? That should fix the Travis build by disabling integration tests. Thank you!

nglorioso:
Rebased on top of spring-cloud-gcp/master

PR #517, opened 2018-04-09
disabling integration tests for forks

balopat:
This PR is from a fork and it skips the integration tests.
The branch version of this PR runs the integration tests.

image

from fork: https://travis-ci.org/spring-cloud/spring-cloud-gcp/builds/364171963
from branch: https://travis-ci.org/spring-cloud/spring-cloud-gcp/builds/364173471

It is leveraging [`TRAVIS_SECURE_ENV_VARS`](https://docs.travis-ci.com/user/environment-variables/#Convenience-Variables) to determine if it is possible to decrypt the credentials required for integration tests on Travis.

With travis, there is just simply no way to run integration tests on forks that require the encrypted variables for cloud credentials. Emulator based tests will still work though.
Running integration tests would be still highly recommended. One way to de-risk broken integration tests in master would be to stage the changes in a local branch (after a thorough review) and have them checked on travis. Until someone comes up with a better idea, this can work, currently fork PRs are just simply broken.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ before_script:
   - while [ ! -f ~/.config/gcloud/emulators/pubsub/env.yaml ]; do sleep 1; done
   - $(gcloud beta emulators pubsub env-init)
 before_install:
-  - if [ "$TRAVIS_PULL_REQUEST" != "false" ]; then
+  - if [ "$encrypted_1ef8dfbdb114_key" != "" ]; then
         openssl aes-256-cbc -K $encrypted_1ef8dfbdb114_key -iv $encrypted_1ef8dfbdb114_iv -in travis.tar.gz.enc -out travis.tar.gz -d;
         tar -xzf travis.tar.gz;
         export INTEGRATION_TEST_FLAGS="-Dit.spanner=true -Dit.storage=true";
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,10 +21,10 @@ before_script:
 before_install:
   - if [ "$TRAVIS_PULL_REQUEST" != "false" ]; then
         openssl aes-256-cbc -K $encrypted_1ef8dfbdb114_key -iv $encrypted_1ef8dfbdb114_iv -in travis.tar.gz.enc -out travis.tar.gz -d;
+        tar -xzf travis.tar.gz;
         export INTEGRATION_TEST_FLAGS="-Dit.spanner=true -Dit.storage=true";
       fi
   - if [ ! -d "$HOME/google-cloud-sdk/bin" ]; then rm -rf $HOME/google-cloud-sdk; export CLOUDSDK_CORE_DISABLE_PROMPTS=1; curl https://sdk.cloud.google.com | bash; fi
   - source $HOME/google-cloud-sdk/path.bash.inc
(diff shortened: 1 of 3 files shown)


artembilan:
The build is still failed:
```
Results :
Tests in error: 
  createAndWriteTest(org.springframework.cloud.gcp.storage.GoogleStorageIntegrationTests): Failed to load ApplicationContext
Tests run: 31, Failures: 0, Errors: 1, Skipped: 0
```

.travis.yml
```diff
@@ -11,18 +11,20 @@ cache:
...
+  - ./mvnw test -B ${INTEGRATION_TEST_FLAGS}
 install:
   - ./mvnw install -DskipTests=true -Dmaven.javadoc.skip=true -B -V
 before_script:
   - gcloud beta emulators pubsub start &
   - while [ ! -f ~/.config/gcloud/emulators/pubsub/env.yaml ]; do sleep 1; done
   - $(gcloud beta emulators pubsub env-init)
 before_install:
-  - openssl aes-256-cbc -K $encrypted_1ef8dfbdb114_key -iv $encrypted_1ef8dfbdb114_iv -in travis.tar.gz.enc -out travis.tar.gz -d
+  - if [ "$encrypted_1ef8dfbdb114_key" != "" ]; then
```
joaoandremartins: So `$encrypted_1ef8dfbdb114_key` is an empty string in PRs from forks? Is that documented anywhere?

.travis.yml
```diff
@@ -19,7 +19,7 @@ before_script:
   - while [ ! -f ~/.config/gcloud/emulators/pubsub/env.yaml ]; do sleep 1; done
   - $(gcloud beta emulators pubsub env-init)
 before_install:
-  - if [ "$TRAVIS_PULL_REQUEST" != "false" ]; then
```
meltsufin: Why this didn't work?
From Travis docs:

> TRAVIS_PULL_REQUEST: The pull request number if the current job is a pull request, “false” if it’s not a pull req

ChengyuanZhao:
@balopat  Is the idea that we would copy fork changes into a branch to have the integration tests run?

balopat:
@ChengyuanZhao with travis, there is just simply no way to run integration tests on forks that require the encrypted variables for gcloud. Emulator based tests will still work though. 
Running integration tests would be still highly recommended. One way to de-risk broken integration tests in master would be, yes, to stage the changes in a local branch (after a thorough review) and have them checked on travis. Until someone comes up with a better idea, this can work.

ChengyuanZhao:
@balopat Since we should be making a branch within the repo to check every fork PR, why don't we just leave the block in place? Removing the block only opens the fork to be merged accidentally without checking the integration tests via a branch (which we agree we should be doing every time anyway).

balopat:
@ChengyuanZhao that is an interesting idea - however, the current version of the "block" stops the build and checks for nothing. We could do a full maven build, skip the auth integration tests, just to have some checks in place for iterating with contributors and we could still fail the build at the end - however in some cases that is an overkill - because the change doesn't involve any integration test impacts (see errorprone PR for example).

ChengyuanZhao:
@balopat perhaps we can add one more check on Github to commit that is if those travis secur env vars are present?

Issue #526, opened 2018-04-09
Eliminate SpannerRepository


joaoandremartins:
`SpannerRepository`'s goal appears to be to expose `SpannerOperations`.
However, if a user needs `SpannerOperations`, they should autowire it from an application context.
If there is no other compelling reason to keep it, it should be removed.

ChengyuanZhao:
I insist on keeping this. The other Spring Data projects have their own repositories that they inevitably start putting their storage-technology specific functionality in. If we get rid of this now it will be a pain to put it back in the future. @balopat what do you think

joaoandremartins:
The Template pattern is independent from Spring Data.
As the issue mentions, the right way to get `SpannerOperations` should be to autowire it, not get it from the template.

The way I see it, it's harder to revert launching an interface that has to be rolled back in later version (i.e., breaking change).

joaoandremartins:
As it turns out, it appears that other Spring Data projects follow the model of having their own repository implementation with vendor-specific functionality.
There appears to be some for Spanner as well, so it could make sense to leave this.

PR #529, opened 2018-04-09
Polishes the Spanner reference document

joaoandremartins:


docs/src/main/asciidoc/spanner.adoc
```diff
@@ -288,14 +316,14 @@ public class ConverterConfiguration {
...
-For Spring Data Spanner, the code inheriting `SpannerRepository` gets all the benefits of `CrudRepository` and `PagingAndSortingRepository` as well.
+{spring-data-commons-ref}/#repositories[Spring Data Repositories] are a powerful abstraction that
+can save you a lot of boilerplate code.
 
 For example:
 
 [source,java]
 ----
-public interface TradeRepository extends SpannerRepository<Trade> {
+public interface TradeRepository extends PagingAndSortingRepository<Trade> {
```
ChengyuanZhao: I think this is a point I missed from the last PR: 

There needs to be 2 type params here. So SpannerRepository or PagingAndSortingRepository with params <Trade, Key>. If you use SpannerRepository instead of the PandR one then you also get the methods the SpannerRepo has in addition. 

The second type param must be Key (the Spanner Key type), a descendant of Object[ ], or Iterable if there are multiple primary key columns, and can be any descendant of Object if there is only one PK column.

joaoandremartins: So it turns out that the key was required even before.
Passed in the key to `PagingAndSortingRepository`, and the plan is to deprecate `SpannerRepository` (#526).

PR #531, opened 2018-04-09
Added Read/write read-only transaction to Spanner Repo

ChengyuanZhao:
Added this Spanner-specific functionality to Spanner Repo

### DIFF
--- a/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/SpannerRepository.java
+++ b/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/SpannerRepository.java
@@ -16,7 +16,10 @@
 
 package org.springframework.cloud.gcp.data.spanner.repository;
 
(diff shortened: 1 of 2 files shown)


balopat:
Not sure if I'm a fan of these generic, "non entity-bound" extension points on a `Repository`. A `Repository` should be scoped for a single entity. Maybe an "entity-bound" version of a `SpannerOperations` would be valuable...
But in Mongo for example they introduced [`deleteBy*` methods](https://docs.spring.io/spring-data/mongodb/docs/current/reference/html/#mongodb.repositories.queries.delete) or other mongo specific but also "entity scoped" methods.

ChengyuanZhao:
@balopat that's a good point! I agree with you on the SpannerOperations being not-entity-bound while this repo is. However I think making SpannerOperations entity-specific is artificial, because much of the usefulness of the transaction is being to interact with many tables. 

I think the only conclusion is that these two methods do not belong in the SpannerRepository :)

ChengyuanZhao:
closing this PR because these features no longer feel like a good fit.

PR #536, opened 2018-04-09
[WIP] speed up maven build on Travis

balopat:


.travis.yml
```diff
@@ -11,9 +11,9 @@ cache:
     - $HOME/google-cloud-sdk
     - $HOME/.m2
 script:
-  - ./mvnw test -B ${INTEGRATION_TEST_FLAGS}
+  - ./mvnw -T 1.5C test -B ${INTEGRATION_TEST_FLAGS}
```
ChengyuanZhao: Wow I didn't know about this option!

balopat: unfortunately it didn't really help, because travis workers have 2 nodes :( 
we'll have to look at more options for parallelization more closely. 
these ideas are pretty good: https://docs.travis-ci.com/user/speeding-up-the-build/
there is a way to parallelize junit tests themselves too - but actually the spanner integration tests have some issues when paralellized.

PR #537, opened 2018-04-09
Spanner db admin

ChengyuanZhao:
created the db admin template class with simple function now to run DDL and create DBs. will be expanded in coming PRs to do more things like create the hierarchy of parent-child relationships and such.


spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/spanner/GcpSpannerAutoConfiguration.java
```diff
@@ -142,4 +144,17 @@ public SpannerSchemaUtils spannerSchemaUtils(
...
+	@ConditionalOnMissingBean
+	public SpannerDatabaseAdminTemplate spannerDatabaseAdminTemplate(
+			DatabaseAdminClient databaseAdminClient, DatabaseId databaseId) {
+		return new SpannerDatabaseAdminTemplate(databaseAdminClient, databaseId);
+	}
+
+	@Bean
+	@ConditionalOnMissingBean
+	public DatabaseAdminClient databaseAdminClient(Spanner spanner) {
+		return spanner.getDatabaseAdminClient();
```
artembilan: Does it really make sense to be so fine-grained? 
Can `DatabaseAdminClient` be created outside of the `Spanner`?
Doesn't this show us that we can have `DatabaseAdminClient` only via calling `Spanner` API?
Thanks

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/admin/SpannerDatabaseAdminTemplate.java
```diff
@@ -0,0 +1,126 @@
...
+					.updateDatabaseDdl(getInstanceId(), getDatabase(), ddlStrings, null)
+					.waitFor();
+		}
+	}
+
+	private String getInstanceId() {
+		return this.databaseId.getInstanceId().getInstance();
+	}
+
+	private String getDatabase() {
```
artembilan: I these two are pretty fine to be as a `public` API

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/admin/SpannerDatabaseAdminTemplate.java
```diff
@@ -0,0 +1,126 @@
...
+				.getValues()) {
+			if (getDatabase().equals(db.getId().getDatabase())) {
+				return true;
+			}
+		}
+		return false;
+	}
+
+	/**
+	 * Returns a map of parent and child table relationships in the database at the
```
artembilan: Any method level JavaDocs must be in the ` imperative style`: https://github.com/spring-projects/spring-framework/wiki/Code-Style#javadoc-formatting

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/admin/SpannerDatabaseAdminTemplate.java
```diff
@@ -0,0 +1,126 @@
...
+ * @author Balint Pato
+ */
+public class SpannerDatabaseAdminTemplate {
+
+	private final DatabaseAdminClient databaseAdminClient;
+
+	private final DatabaseId databaseId;
+
+	/**
+	 * Constructor.
```
artembilan: Need more info in the method description.
Also, please, some class-level JavaDocs.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/spanner/GcpSpannerAutoConfiguration.java
ChengyuanZhao: I don't think there's another way to get a db admin client, but I thought the alternative to having this bean here is to put the very broad Spanner object as the constructor argument to classes that otherwise would only need the DatabaseAdminClient and have them call getDatabaseAdminClient themselves.  I thought calling this method for them would be preferable. I could be missing something though...

artembilan: Well, my point is to inject `Spanner` to the previous `SpannerDatabaseAdminTemplate` bean definition and call this `getDatabaseAdminClient()` there for the ctor.

But I can live with this bean as well, if there is a good explanation "Why?".

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/admin/SpannerDatabaseAdminTemplate.java
```diff
@@ -74,11 +77,19 @@ else if (databaseExists()) {
 		}
 	}
 
-	private String getInstanceId() {
+	/**
+	 * Get the instance ID used to perform database operations.
+	 * @return
```
artembilan: `@return` should have some text as well

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/spanner/GcpSpannerAutoConfiguration.java
ChengyuanZhao: My preference to call .getDatabaseAdminClient() in this autoconfig class instead of the SpannerDatabaseAdminTemplate is that you can do much more than just db admin operations with a Spanner object, and we only want the admintemplate to have the dbadminclient.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/admin/SpannerDatabaseAdminTemplate.java
```diff
@@ -74,11 +77,19 @@ else if (databaseExists()) {
...
+	 */
+	public String getInstanceId() {
 		return this.databaseId.getInstanceId().getInstance();
 	}
 
-	private String getDatabase() {
+	/**
+	 * Get the database ID used to perform database operations.
+	 * @return
```
artembilan: DITTO

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/spanner/GcpSpannerAutoConfiguration.java
artembilan: I'm fine with this too. So, never mind.
Just to finish a discussion and close all the gaps, let me show what I mean anyway:
```
@Bean
@ConditionalOnMissingBean
public SpannerDatabaseAdminTemplate spannerDatabaseAdminTemplate(
		Spanner spanner, DatabaseId databaseId) {
	return new SpannerDatabaseAdminTemplate(spanner.getDatabaseAdminClient(), databaseId);
}
```
That's all. No any `DatabaseAdminClient` beans requirements.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/admin/SpannerDatabaseAdminTemplate.java
```diff
@@ -0,0 +1,137 @@
...
+
+	/**
+	 * Execute the given DDL strings in order and creates the database if it does not
+	 * exist.
+	 * @param ddlStrings
+	 * @param createDatabaseIfNotExists Has no effect if the database already exists.
+	 * Otherwise, if True then the database is created and the DDL strings are executed;
+	 * the DDL is not executed if False.
+	 */
+	public void executeDDLStrings(Iterable<String> ddlStrings,
```
meltsufin: I think the codestyle we're following requires capitalization of only the first letter of an acronym in identifiers.
So, `executeDdlStrings`.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/admin/SpannerDatabaseAdminTemplate.java
```diff
@@ -0,0 +1,137 @@
...
+			boolean createDatabaseIfNotExists) {
+		if (!databaseExists() && createDatabaseIfNotExists) {
```
meltsufin: I would change the order of the AND to short-circuit if user doesn't desire auto-creation of the database, since checking for the existence of the database requires a remote call.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/admin/SpannerDatabaseAdminTemplate.java
```diff
@@ -0,0 +1,137 @@
...
+			this.databaseAdminClient
+					.createDatabase(getInstanceId(), getDatabase(), ddlStrings).waitFor();
+		}
+		else if (databaseExists()) {
```
meltsufin: Maybe remove the `if` part and let it fail? Otherwise, you should have an `else` that throws an exception. Without that in place and the database no existing, the method will do nothing, and the user won't even be notified with an exception.

ChengyuanZhao:
@balopat now that M3 is out the door, PTAL

ChengyuanZhao:
@meltsufin PTAL changes made

PR #539, opened 2018-04-10
moves google cloud java bom out of our bom

### LINKED ISSUE #456: Remove Google Cloud Java BOM import from our BOM
@joaoandremartins The best practice seems to be to not include external dependnecies in the BOM, but rather keep them in the parent POM.
We seem to violate this rule in our BOM by importing the Google Cloud Java BOM there.
Also see: https://github.com/GoogleCloudPlatform/google-cloud-java/issues/2781

ChengyuanZhao:
fixes #456

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -41,6 +41,7 @@
 		<maven.compiler.target>1.8</maven.compiler.target>
 		<spring-cloud-release.version>Finchley.M9</spring-cloud-release.version>
 		<spring-boot-release.version>2.0.1.RELEASE</spring-boot-release.version>
+		<google-cloud-bom.version>0.43.0-alpha</google-cloud-bom.version>
 	</properties>
 
 	<dependencyManagement>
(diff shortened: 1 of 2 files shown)


meltsufin:
I'm actually not sure we should do this based on the comment from @snicoll on the issue:

> This BOM is providing support for the Google platform and, as such, should provide versions for Google-related artifacts.

ChengyuanZhao:
ok then let's close this as well as the underlying issue that's a blocker for M3 ? @joaoandremartins @meltsufin

joaoandremartins:
I'm OK with what this PR proposes due to my argument in #456.
Just wanted to hear other people's opinions before.
@meltsufin @artembilan @balopat if you agree that this is a good thing, I think we should merge it in.

artembilan:
Well, as far as all our modules provide the proper transitive dependencies from that Google BOM and our Boot starters have a good opinionated set set of dependencies, I think this is the way to go - merge this PR and reconsider if there are some concerns.

snicoll:
>  due to my argument in #456.

I don't think that's right. There is a difference between you providing a version (because you're using the bom internally) and *enforcing* a set of consistent versions (that's what dependency management does. 

If the Google BOM has been cleaned and doesn't leak 3rd party artifacts, I think we should include it in the bom.

joaoandremartins:
I guess if a user wants to force a version, setting it explicitly in their pom is still closer to the root than our BOM, so I don't have issues either way.

snicoll:
@joaoandremartins or they can just provide the version rather than no version at all. This is how it works in Spring Boot since day 1.

PR #542, opened 2018-04-10
Return SpannerTemplate instead of SpannerOperations in autoconfig

### LINKED ISSUE #533: GcpSpannerAutoConfiguration should expose SpannerTemplate
Currently, it only autoconfigures a `SpannerOperations` bean. Instead, we can just autoconfigure the `spannerTemplate` bean.

meltsufin:
Fixes #533.


spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/config/EnableSpannerRepositories.java
```diff
@@ -120,7 +120,7 @@
 	 *
 	 * @return
 	 */
-	String spannerOperationsRef() default "spannerOperations";
+	String spannerOperationsRef() default "spannerTemplate";
```
artembilan: `spannerTemplateRef` then?

At least that is what I see in the `@EnableMongoRepositories`:
```
	/**
	 * Configures the name of the {@link MongoTemplate} bean to be used with the repositories detected.
	 *
	 * @return
	 */
	String mongoTemplateRef() default "mongoTemplate";
```

artembilan:
Hm...
Have you tested this?
See Travis report:
```
Tests in error: 
  declarativeQueryMethodTest(org.springframework.cloud.gcp.data.spanner.repository.it.SpannerRepositoryIntegrationTests): Failed to load ApplicationContext
  readWriteTransactionTest(org.springframework.cloud.gcp.data.spanner.core.it.SpannerTemplateIntegrationTests): Failed to load ApplicationContext
  readOnlyTransactionTest(org.springframework.cloud.gcp.data.spanner.core.it.SpannerTemplateIntegrationTests): Failed to load ApplicationContext
  insertAndDeleteSequence(org.springframework.cloud.gcp.data.spanner.core.it.SpannerTemplateIntegrationTests): Failed to load ApplicationContext
```

meltsufin:
Tests passed locally.

artembilan:
I'm not sure...
See the root cause: 
```
Caused by: org.springframework.beans.factory.NoSuchBeanDefinitionException: No bean named 'spannerTemplate' available
```
This can't be so sporadic. This is fully related to your change.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/config/EnableSpannerRepositories.java
meltsufin: I'd rather not change this since it requires a lot of cascading changes.

artembilan: ???
You just do `Refactoring -> Rename...` and every consumer of this property will be change respectively.

Again: we lose consistency if we stick with the name as it is right now.
Pay attention to the `@EnableRedisRepositories` as well:
```
	/**
	 * Configures the bean name of the {@link RedisOperations} to be used. Defaulted to {@literal redisTemplate}.
	 *
	 * @return
	 */
	String redisTemplateRef() default "redisTemplate";
```

ChengyuanZhao: I think this makes sense to not change. The significance of this name is in `SpannerRepositoryConfigurationExtension`. "spannerOperationsRef" I believe is how the repository factory knows where to get an instance of the SpannerOperations needed to create a SimpleSpannerRepository. Following the guideline of the consumer (in this case the SimpleSpannerRepository) needing only a SpannerOperations and not the more specific SpannerTemplate, I think it makes sense that the factory use a "spannerOperationsRef" rather than a "spannerTemplateRef".

The value within being "spannerTemplate" just tells which bean function to call to get a SpannerOperations for making the SimpleSpannerRepository.

artembilan: I think this must be renamed, for consistency with all existing Spring Data projects.
One more sample: https://docs.spring.io/spring-data/cassandra/docs/current/api/org/springframework/data/cassandra/repository/config/EnableCassandraRepositories.html#cassandraTemplateRef--

That one indeed can stay as `operations`. Here is a sample:
```
public void postProcess(BeanDefinitionBuilder builder, AnnotationRepositoryConfigurationSource config) {

	AnnotationAttributes attributes = config.getAttributes();

	builder.addPropertyReference("mongoOperations", attributes.getString("mongoTemplateRef"));
	builder.addPropertyValue("createIndexesForQueryMethods", attributes.getBoolean("createIndexesForQueryMethods"));
}
```

@mp911de,

would you mind to share your feedback on the matter, please?

ChengyuanZhao:
@balopat I don't think this should be an M3 blocker. It's not an urgent change, and I don't think it really adds any value to M3 specifically. We could apply this change after M3 I think.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/config/EnableSpannerRepositories.java
ChengyuanZhao: That works for me too.

balopat:
I think it's a very low risk one and can fit in easily, it's basically done and aligns the development well on a community pattern. I wouldn't call it a blocker, but if we can fit in we should.

PR #546, opened 2018-04-10
Spanner refdoc additions and reorganizing

ChengyuanZhao:


docs/src/main/asciidoc/spanner.adoc
```diff
@@ -481,7 +481,7 @@ Queries on the Template API support sorting through standard SQL and also via Sp
 ----
 //...
 
-  List<Trade> trades = this.spannerTemplate.findAll(Trade.class, Sort.by("action")););
+  List<Trade> trades = this.spannerTemplate.queryAll(Trade.class, Sort.by("action")););
```
ChengyuanZhao: @balopat  I actually didn't rename this one "queryAll" because it doesn't take a Statement and from the outside there's nothing really query-ish about it. Another reason I kept it named "findAll" is that it is almost the same as the PagingAndSortingRepository findAll, which takes just the Sort.

PR #547, opened 2018-04-10
[WIP] Spanner better data rest support

balopat:
DO NOT MERGE this just yet.
@saturnism 
@ChengyuanZhao 
@mp911de 

please have a look - this is to support simple Spring Data REST integration for multikey setups.

### DIFF
--- a/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerMutationFactoryImpl.java
+++ b/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerMutationFactoryImpl.java
@@ -29,6 +29,7 @@ import com.google.cloud.spanner.Mutation.WriteBuilder;
 import org.springframework.cloud.gcp.data.spanner.core.convert.SpannerConverter;
 import org.springframework.cloud.gcp.data.spanner.core.mapping.SpannerMappingContext;
 import org.springframework.cloud.gcp.data.spanner.core.mapping.SpannerPersistentEntity;
+import org.springframework.cloud.gcp.data.spanner.core.mapping.WrappedKey;
 import org.springframework.data.mapping.PersistentProperty;
 import org.springframework.data.mapping.PersistentPropertyAccessor;
 import org.springframework.util.Assert;
@@ -87,7 +88,7 @@ public class SpannerMutationFactoryImpl implements SpannerMutationFactory {
 			PersistentPropertyAccessor accessor = persistentEntity
 					.getPropertyAccessor(entity);
 			PersistentProperty idProperty = persistentEntity.getIdProperty();
-			Key value = (Key) accessor.getProperty(idProperty);
+			Key value = ((WrappedKey) accessor.getProperty(idProperty)).getSpannerKey();
 
 			builder.addKey(value);
 		}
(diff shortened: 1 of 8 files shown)


balopat:
I reverted the original approach with `WrappedKey` as it was bringing a Spring Data REST concern to Spring Data Spanner, and it was hard to imagine how to configure it properly. 

The new approach I'm much happier with: 
- registers a `BackendIdConverter` by default (conditional on missing bean for `BackendIdConverter`) that converts `Key` to `String` and back 
- by default the logic is using `_` character for separation. It's overridable easily, if the user wants, can specify a `@Bean` like this: 

````
@Component
class MySpecialIdConverter extends SpannerKeyIdConverter {

    @Override
    protected String getUrlIdSeparator() {
        return ":";
    }
}
````

File: spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/spanner/SpannerKeyRestSupportAutoConfiguration.java [hunk 1]
```diff
@@ -0,0 +1,42 @@
...
+
+@Configuration
+@AutoConfigureAfter(GcpSpannerAutoConfiguration.class)
+@ConditionalOnMissingBean({ BackendIdConverter.class })
+@ConditionalOnClass({ BackendIdConverter.class })
+public class SpannerKeyRestSupportAutoConfiguration {
+
+	@Bean
+	@ConditionalOnMissingBean
+	public BackendIdConverter backendIdConverter() {
```
mp911de: Method names are picked as bean names. It can be unfortunate to pick a generic name and the generic `BackendIdConverter`. Anyone registering a `BackendIdConverter` will stop this bean definition from contributing the bean. Exposing `SpannerKeyIdConverter spannerKeyIdConverter()` would be a safer approach.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/spanner/SpannerKeyRestSupportAutoConfiguration.java
```diff
@@ -0,0 +1,42 @@
...
+import org.springframework.context.annotation.Configuration;
+import org.springframework.data.rest.webmvc.spi.BackendIdConverter;
+
+/**
+ * @author Balint Pato
+ */
+
+@Configuration
+@AutoConfigureAfter(GcpSpannerAutoConfiguration.class)
+@ConditionalOnMissingBean({ BackendIdConverter.class })
```
mp911de: `@ConditionalOnMissingBean` is a better fit for `@Bean` methods.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/spanner/SpannerKeyIdConverter.java
```diff
@@ -0,0 +1,53 @@
...
+	@Override
+	public String toRequestId(Serializable source, Class<?> entityType) {
+		Key id = (Key) source;
+		StringJoiner stringJoiner = new StringJoiner(getUrlIdSeparator());
+		id.getParts().forEach(p -> stringJoiner.add(p.toString()));
+		return stringJoiner.toString();
+	}
+
+	@Override
+	public boolean supports(Class<?> type) {
```
mp911de: Having no constraint here will apply the converter on every single entity, independent of its type. One can have multiple Spring Data modules on the classpath. It would make sense to validate the type through `PersistentEntities.getPersistentEntity(…)` and check the actual Id type.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/spanner/SpannerKeyIdConverter.java
```diff
@@ -0,0 +1,53 @@
...
+public class SpannerKeyIdConverter implements BackendIdConverter {
+
+	@Override
+	public Serializable fromRequestId(String id, Class<?> entityType) {
+		Object[] parts = id.split(getUrlIdSeparator());
+		return Key.of(parts);
+	}
+
+	protected String getUrlIdSeparator() {
+		return "_";
```
ChengyuanZhao: But isn't this also a totally legit character to have in a STRING column in Spanner that can a key column? Wouldn't that mean this has the same problem as the comma?

PR #549, opened 2018-04-11
Uses CF credentials for Trace

joaoandremartins:


spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/trace/StackdriverTraceAutoConfiguration.java
```diff
@@ -97,7 +97,9 @@ public StackdriverTraceAutoConfiguration(GcpProjectIdProvider gcpProjectIdProvid
 		this.finalProjectIdProvider = gcpTraceProperties.getProjectId() != null
 				? gcpTraceProperties::getProjectId
 				: gcpProjectIdProvider;
-		this.finalCredentialsProvider = gcpTraceProperties.getCredentials().getLocation() != null
+		this.finalCredentialsProvider =
+				gcpTraceProperties.getCredentials().getLocation() != null
+						|| gcpTraceProperties.getCredentials().getEncodedKey() != null
```
meltsufin: We are now repeating these two lines for the OR everywhere. I think it would be better to just add a method to `Credentials` like `hasKey()` so that the two lines are replaced with `gcpTraceProperties.getCredentials().hasKey()`.

PR #551, opened 2018-04-11
Adds Cloud Foundry to the refdoc

joaoandremartins:


docs/src/main/asciidoc/cloudfoundry.adoc
```diff
@@ -0,0 +1,15 @@
+== Cloud Foundry
+
+Spring Cloud GCP provides support for Cloud Foundry's
+https://docs.pivotal.io/partners/gcp-sb/index.html[GCP Service Broker].
+Our Pub/Sub, Spanner, Storage and Stackdriver Trace starters are Cloud Foundry aware and
+retrieve properties like project ID, credentials, etc., that are used in auto configuration
```
artembilan: Missed period in the end of sentence.
I'm also curious why don't you follow *one line per sentence* best practice: https://asciidoctor.org/docs/asciidoc-recommended-practices/#one-sentence-per-line?

joaoandremartins: @artembilan I do one sentence per line, but I also wrap the sentence at the end, so it's more easy to read in a text editor.

artembilan: Then it is not *one sentence per line*. 
And that is wrong editor then...

OK. Let it be. That's not a first thing I can't convince you to follow...

joaoandremartins: I had read the guideline wrong originally.

One of these days I'll do a swipe of the refdoc and apply that best practice.

PR #552, opened 2018-04-11
 Adds topic expression setting to Pub/Sub outbound CA

### LINKED ISSUE #19: Support PubSub topic resolution from message header

joaoandremartins:
Fixes #19

### DIFF
--- a/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/outbound/PubSubMessageHandler.java
+++ b/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/outbound/PubSubMessageHandler.java
@@ -28,6 +28,7 @@ import org.springframework.cloud.gcp.pubsub.core.PubSubOperations;
 import org.springframework.cloud.gcp.pubsub.support.GcpPubSubHeaders;
 import org.springframework.expression.EvaluationContext;
 import org.springframework.expression.Expression;
+import org.springframework.expression.common.LiteralExpression;
 import org.springframework.integration.expression.ExpressionUtils;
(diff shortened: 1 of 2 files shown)


spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/outbound/PubSubMessageHandler.java
```diff
@@ -189,6 +190,26 @@ public void setPublishCallback(ListenableFutureCallback<String> publishCallback)
...
 
+	public Expression getTopicExpression() {
+		return this.topicExpression;
+	}
+
+	/**
+	 * Set the SpEL expression for the topic this adapter sends messages to.
+	 * @param topicExpression SpEL expression representing the topic name
+	 */
+	public void setTopicExpression(Expression topicExpression) {
```
artembilan: Consider also to add ` setTopicExpressionString(String topicExpression)` and use a `EXPRESSION_PARSER` from the super class to parse provided String to the `Expression`.

This is pretty convenient for JavaConfig

PR #562, opened 2018-04-11
1.0.0.M3 Release

sobychacko:


joaoandremartins:
Now it's trying to get `spring-cloud-gcp-trace`, but that doesn't exist anymore...

joaoandremartins:
@sobychacko can you merge with master? Error should be gone after #563

PR #567, opened 2018-04-11
Custom SQL spanner queries use name-matched tags

ChengyuanZhao:


spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SqlSpannerQuery.java
```diff
@@ -63,20 +65,24 @@ public SqlSpannerQuery(Class type, QueryMethod queryMethod,
...
-		Matcher matcher = pattern.matcher(sql);
+	private List<String> getTags() {
 		List<String> tags = new ArrayList<>();
-		while (matcher.find()) {
-			// The initial '@' character must be excluded for Spanner
-			tags.add(matcher.group().substring(1));
+		Parameters parameters = getQueryMethod().getParameters();
+		for (int i = 0; i < parameters.getNumberOfParameters(); i++) {
+			Optional<String> paramName = parameters.getParameter(i).getName();
+			if (!paramName.isPresent() || paramName.get() == null) {
```
balopat: `paramName.get()` is superfluous

PR #570, opened 2018-04-12
Added Spanner-Repo domain-type specific transactions. 

ChengyuanZhao:
provides repo functions instead of template functions


meltsufin:
@ChengyuanZhao Have you looked at [Spring's transaction management](https://docs.spring.io/spring/docs/current/spring-framework-reference/data-access.html#transaction) and support for declaritive transactions using `@Transactional`? It seems like we should try to integrate into those abstractions.

artembilan:
I agree with @meltsufin that we need to consider to implement a *declarative transactions management support* as well. And that is indeed for the `@Transactional` annotation or for the `TransactionTemplate`. The main reason of this just because I may have several transaction resources in a single call and would like to perform atomic workflow. 

The current solution is also good. We have similar solution for Redis. This is its native tool so why don't let end-users to use it.

But at the same time something like `SpannerTransactionManager` should be considered as well.

Also pay attention how this TMs can be *chained*: http://gharshangupta.blogspot.com/

Also there is an epic article from very Dr. Dave Syer: https://www.javaworld.com/article/2077963/open-source-tools/distributed-transactions-in-spring--with-and-without-xa.html?page=2

balopat:
added https://github.com/spring-cloud/spring-cloud-gcp/issues/592

ChengyuanZhao:
@meltsufin would you mind approving again ? the merge from master made approval stale

PR #571, opened 2018-04-12
Spanner Projections

ChengyuanZhao:


spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/AbstractSpannerQuery.java
```diff
@@ -0,0 +1,84 @@
...
+			SpannerMappingContext spannerMappingContext) {
+		this.queryMethod = queryMethod;
+		this.entityType = type;
+		this.spannerOperations = spannerOperations;
+		this.spannerMappingContext = spannerMappingContext;
+	}
+
+	@Override
+	public Object execute(Object[] parameters) {
+		Object rawResult = executeRawResult(parameters);
```
balopat: Can you please either show an example that this is how in other Spring Data modules is done? 
My worry is that creating the full parent entity can cause surprises/unwanted performance impact.

ChengyuanZhao: Take a look at `AbstractJpaQuery.java` from the spring-data-jpa project. They do it in a nearly identical way.

Regardless, I think it would be a bad idea at this point to carry the projection information any higher up in the query/retrieval logic. Furthermore, while it is very very user-friendly to create a projection using spring-data  provided query method and result processor, they do not let you extract the needed projection requirements one  would need to actually that info upward to the query/retrieval code. 

Also, I think the assumption you're coming with that there might be some performance-limited situation where a projection-returning method works well but the actual domain type of the repository in which the projection-returning method lives would not is really not what I'm getting reading the projection docs from Spring Data. I'm really surprised there is an interpretation of the purpose of projections function as anything other than syntax-related convenience.

balopat: Thank you, I'm convinced that this solution is in line with the Spring Data design. 

I'm still not in love with it though :) - it's a hardcoded depedency between the entity creation and the project creation and doesn't allow you to be more surgical about it. But it's fine - we might want to mention though this in the user docs that user can define their subselect if they want to and relax the invariants on the entity for partial reading.
Let's add a section to the ref-doc to explain this?

ChengyuanZhao: Will do ! 

Though one thing we is we made the decision early on the throw an exception if a column in the POJO doesn't appear in the resultset from Spanner. We originally did this to ensure that POJOs were coming from their home table and to prevent people playing fast and loose. Since we're still using the same read converter code, if the user selects too few columns from Spanner in their query then they will get that error, because it is still the full POJO in the back. 

we could just remove that restriction. what do you think?

ChengyuanZhao: Oh i have an idea. By default we'll throw the exception but in the SpannerQueryOptions we'll add the setting to allow it . @balopat

balopat: I like the QueryOptions idea! Then we can add that into the ref-doc as well to the partial read sections - and update the Read vs Query matrix too.

ChengyuanZhao: changes made

spring-cloud-gcp-docs/src/main/asciidoc/spanner.adoc
```diff
@@ -794,6 +800,36 @@ List<Trade> fetchByActionNamedQuery(String action);
 
 This allows table names evaluated with SpEL to be used in custom queries.
 
+==== Projections
+
+Spring Data Spanner supports projections. You can define projection interfaces based
```
balopat: link to `https://docs.spring.io/spring-data/commons/docs/current/reference/html/#projections` - please note we have an asciidoc variable already for the spring data commons current docs root.

spring-cloud-gcp-docs/src/main/asciidoc/spanner.adoc
```diff
@@ -794,6 +800,36 @@ List<Trade> fetchByActionNamedQuery(String action);
...
+----
+public interface TradeProjection {
+
+	String getAction();
+
+	@Value("#{target.symbol + ' ' + target.action}")
+	String getSymbolAndAction();
+}
+
+public interface TradeRepository extends SpannerRepository<Trade, Key> {
```
balopat: This shouldn't be a `SpannerRepository`, but a `CrudRepository` unless I'm missing something.

spring-cloud-gcp-docs/src/main/asciidoc/spanner.adoc
```diff
@@ -794,6 +800,36 @@ List<Trade> fetchByActionNamedQuery(String action);
...
+Spring Data Spanner supports projections. You can define projection interfaces based
+on domain types and add query methods that return them in your repository:
+
+[source, java]
+----
+public interface TradeProjection {
+
+	String getAction();
+
+	@Value("#{target.symbol + ' ' + target.action}")
```
balopat: what is `target`? can you explain that below?

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerQueryOptions.java
```diff
@@ -40,6 +40,8 @@
 
 	private OptionalLong offset = OptionalLong.empty();
 
+	private boolean allowMissingResultSetColumns;
```
balopat: please rename to `allowPartialRead` as that's shorter and consistent with the terminology we use in the ref-docs too

spring-cloud-gcp-docs/src/main/asciidoc/spanner.adoc
ChengyuanZhao: I believe it is not something the user or even we can decide. The SpEL for these projection functions is handled by `SpelAwareProxyProjectionFactory` and it seems that "target" name is significant. Take a look at the projections page in the spring data commons project for a better explanation. https://github.com/spring-projects/spring-data-commons/blob/master/src/main/asciidoc/repository-projections.adoc

spring-cloud-gcp-docs/src/main/asciidoc/spanner.adoc
ChengyuanZhao: Both work, but `SpannerRepository` would give them some Spanner-specific methods.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerQueryOptions.java
ChengyuanZhao: changes made

spring-cloud-gcp-docs/src/main/asciidoc/spanner.adoc
balopat: Haha, I understand it, I meant can you write an explanation in the doc?

ChengyuanZhao: done!

PR #575, opened 2018-04-13
Made exec query and read methods public

### LINKED ISSUE #525: Make SpannerTemplate.executeQuery/executeRead public
`SpannerTemplate.executeQuery` and `SpannerTemplate.executeRead` could be useful when querying tables that are not mapped to entities.

ChengyuanZhao:
fixes #525

### DIFF
--- a/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/spanner/GcpSpannerAutoConfiguration.java
+++ b/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/spanner/GcpSpannerAutoConfiguration.java
@@ -37,6 +37,7 @@ import org.springframework.cloud.gcp.core.UsageTrackingHeaderProvider;
 import org.springframework.cloud.gcp.data.spanner.core.SpannerMutationFactory;
 import org.springframework.cloud.gcp.data.spanner.core.SpannerMutationFactoryImpl;
 import org.springframework.cloud.gcp.data.spanner.core.SpannerOperations;
+import org.springframework.cloud.gcp.data.spanner.core.SpannerSqlTemplate;
 import org.springframework.cloud.gcp.data.spanner.core.SpannerTemplate;
 import org.springframework.cloud.gcp.data.spanner.core.admin.SpannerDatabaseAdminTemplate;
 import org.springframework.cloud.gcp.data.spanner.core.admin.SpannerSchemaUtils;
@@ -120,6 +121,12 @@ public class GcpSpannerAutoConfiguration {
 			SpannerMutationFactory spannerMutationFactory) {
 		return new SpannerTemplate(databaseClient, mappingContext, spannerConverter,
 				spannerMutationFactory);
+	}
+
+	@Bean
+	@ConditionalOnMissingBean
+	public SpannerSqlTemplate spannerSqlTemplate(SpannerTemplate spannerTemplate) {
+		return spannerTemplate.getSpannerSqlTemplate();
 	}
 
 	@Bean
--- a/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
+++ b/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
@@ -38,6 +38,7 @@ import com.google.cloud.spanner.Statement;
 import com.google.cloud.spanner.TimestampBound;
 import com.google.cloud.spanner.TransactionContext;
 import com.google.cloud.spanner.TransactionRunner.TransactionCallable;
+import com.google.common.annotations.VisibleForTesting;
 
 import org.springframework.cloud.gcp.data.spanner.core.convert.SpannerConverter;
 import org.springframework.cloud.gcp.data.spanner.core.mapping.SpannerMappingContext;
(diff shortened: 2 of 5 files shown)


ChengyuanZhao:
I am ok with dropping this PR if Mike closes that issue as well. 

also "even JdbcTemplate" lol

meltsufin:
If you don't like returning the `ResultSet`, maybe expose something like `queryForList` where it returns a list of maps. The idea is to not require a struct.

ChengyuanZhao:
@meltsufin Do you mean like `List<Map<String,Object>>` where the String is the column name and the Object is the value?
There are some problems here though:
1. generic `Statements` allow the construction of STRUCT type columns, which  means embedded Map in that Object value. It would be nearly unusable if we had to let the user do all this reflection and inspection
2. the Struct (row ) type doesn't have a generic getObject(columnName) function, so we would basically be applying the full conversion logic we have, at which point it makes sense to just have them use the public Read/Query/Find template/repo methods.

meltsufin:
I'm not sure how popular the use of `STRUCT` is because it's not a common concept in relational databases AFAIK. What if we just ignored the `STRUCT` use-case? Would it make sense then?

ChengyuanZhao:
It feels wrong to be ignoring Spanner's in-query STRUCT construction abilities. Furthermore, my second point about there being no "getObject" method but only having type-specific getters still stands.

meltsufin:
I think I understand the confusion now. The result set of a Spanner query is basically an array of `com.google.cloud.spanner.Struct`. `Struct` class doesn't have a `getObject` method to help construct the `Map`. That was your second point. I think that can be resolved by utilizing the `Struct.getColumnType` method and calling the appropriate `Struct.get[type]` method.
Regarding the first point that a Spanner query can select a STRUCT, and `getColumnType` would itself return `Struct`, I would probably just keep that entry as `Struct` and not unwrap it further.

ChengyuanZhao:
@meltsufin you're exactly right about all of that. 

With that context, I just want to explain what my original two points were 

1.) The "Object" value can now be the 6 single-item types, the array or List of those types, or another Struct. I am very very skeptical that such an "Object" value would have much value, since it would require the user to either keep track in their head what the types are and do their own casting to use, or to do much "instanceof" kind of checking. 

2.) Because of the 6 single item types, numerous item array and List types, and the STRUCT type column, we would be duplicating much of the logic we provide via SpannerConverter in order to chose the correct getter method to call. As a result, I think the user should just be utilizing that converting/mapping logic directly via our main find / read/ query methods.

ChengyuanZhao:
It doesn't seem like `List<Struct>` would solve the problem of not exposing a Spanner type, though...
`Struct` is a Spanner type just like `ResultSet.

balopat:
I recommend the following signature: 

`List<Map<String, Object>> executeQuery(Statement sql)` 


That would

ChengyuanZhao:
@balopat @meltsufin PTAL. added the sqltemplate class

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/spanner/GcpSpannerAutoConfiguration.java
```diff
@@ -122,6 +123,12 @@ public SpannerTemplate spannerTemplate(DatabaseClient databaseClient,
 				spannerMutationFactory);
 	}
 
+	@Bean
+	@ConditionalOnMissingBean
+	public SpannerSqlTemplate spannerSqlTemplate(SpannerTemplate spannerTemplate) {
+		return spannerTemplate.getSpannerSqlTemplate();
```
meltsufin: I think this is backwards. You want to create a `SpannerSqlTemplate` bean and then have `SpannerTemplate` depend on it.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
```diff
@@ -92,6 +93,23 @@ public SpannerMappingContext getMappingContext() {
 		return this.mappingContext;
 	}
 
+	public SpannerSqlTemplate getSpannerSqlTemplate() {
+		return new SpannerSqlTemplate() {
```
meltsufin: Return a `spannerSqlTemplate` member instead. See my comment above.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/spanner/GcpSpannerAutoConfiguration.java
balopat: I'm sorry I haven't clarified this further - my comment became confusing.

After our discussion today it seems that a SpannerSqlTemplate is something that can live on the lower level and actually can expose Spanner types, similar to CqlTemplate in Cassandra. 

Also, as @meltsufin says SpannerSqlTemplate would be the foundation for SpannerTemplate this way.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
meltsufin: Why can't `SpannerTemplate` simply call `this.spannerSqlTemplate.executeQuery`?

ChengyuanZhao: because that method would be private right? Isn't that the whole point?

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/spanner/GcpSpannerAutoConfiguration.java
balopat: Let's put this to on-hold then.

ChengyuanZhao:
This one is out of date (being based on much older files) and I think it will be more practical to start new when we have time to pick this up again.

PR #576, opened 2018-04-13
Spanner POJO optional fields

### LINKED ISSUE #429: Spanner @optional properties for read and write

ChengyuanZhao:
fixes #429


meltsufin:
@ChengyuanZhao What about refdoc, readme, and sample app updates?

ChengyuanZhao:
@meltsufin will be in coming PR. We have a wave of these small improvements at the moment and wanted to put them all together in doc and sample update. Otherwise there would have been much conflict merging.

meltsufin:
Ideally, the docs should be updated in the feature PR. Let's create an issue for this one so we don't lose track of the these small doc updates. You can resolve multiple doc update issues in one PR though.

PR #583, opened 2018-04-16
Do not override Guava version in parent POM

### LINKED ISSUE #582: Inconsistent version of Guava
I was concerned with the way the [guava version is handled in this build](https://github.com/spring-cloud/spring-cloud-gcp/pull/466#issuecomment-373566292) that got no answer

A fresh project created with only `spring-cloud-gcp-starter` on start.spring.io brings `com.google.guava:guava:jar:18.0`

[This line is useless](https://github.com/spring-cloud/spring-cloud-gcp/blob/5dab9f4f3375118b153c04cadd8e38fb379b6551/pom.xml#L81) if you're not using guava yourself as it will fix the version for your build internally but will not promote the fact you have a preference for that version. For that to happen, you have to add a dependency explicitly on guava (which is quite messy if you ask me). Adding that to  `spring-cloud-gcp-core` might do what you want.

Guava is known to be very problematic so perhaps we should reconsider its usage?

meltsufin:
Fixes #582.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -75,10 +75,6 @@
 				<version>0.1.5</version>
 			</dependency>
 
-			<dependency>
-				<groupId>com.google.guava</groupId>
-				<artifactId>guava</artifactId>
-			</dependency>
 		</dependencies>
 	</dependencyManagement>
 
--- a/pom.xml
+++ b/pom.xml
@@ -78,7 +78,6 @@
 			<dependency>
 				<groupId>com.google.guava</groupId>
 				<artifactId>guava</artifactId>
-				<version>20.0</version>
 			</dependency>
 		</dependencies>
 	</dependencyManagement>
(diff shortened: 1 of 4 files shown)


meltsufin:
@snicoll The build fails now because Spring Cloud BOM imports Guava 18 and Google Cloud Java libraries need Guava 20. So, I think we actually need to specify the version of Guava in Spring Cloud GCP until the Spring Cloud BOM / Netflix BOM are fixed to not manage the Guava dependency since they do not provide it.

meltsufin:
@snicoll Our code doesn't require 20 and we can compile with 20 or 18. However, our transitive dependencies in Google Cloud libs require 20. Since Spring Cloud BOM via the Netflix BOM manages the guava dependency to version 18 there doesn't seem like there is anything we can do to override it in the user app, unless we add guava 20 to our own BOM.

I've tried adding guava 20 to spring-cloud-gcp-core and that still caused 18 to be imported in the user app.

snicoll:
@meltsufin you're right, that faulty bom takes precedence. Thanks for the feedback

meltsufin:
Pending fix in https://github.com/spring-cloud/spring-cloud-netflix/issues/2854 to be released.

Issue #581, opened 2018-04-14
Spanner - java.util.Date doesn't seem to be supported


saturnism:
Documentation suggests that `java.util.Date` is a supported type:
https://docs.spring.io/spring-cloud-gcp/docs/1.0.0.M3/reference/htmlsingle/#_supported_types

But when reading from a date field to `java.util.Date`, received the following exception:
```
org.springframework.cloud.gcp.data.spanner.core.mapping.SpannerDataException: The value in column with name birth_date could not be converted to the corresponding property in the entity. The property's type is class java.util.Date.
	at org.springframework.cloud.gcp.data.spanner.core.convert.MappingSpannerReadConverter.lambda$read$0(MappingSpannerReadConverter.java:147)
	at org.springframework.data.mapping.model.BasicPersistentEntity.doWithProperties(BasicPersistentEntity.java:328)
	at org.springframework.cloud.gcp.data.spanner.core.convert.MappingSpannerReadConverter.read(MappingSpannerReadConverter.java:110)
	at org.springframework.cloud.gcp.data.spanner.core.convert.MappingSpannerConverter.mapToList(MappingSpannerConverter.java:91)
	at org.springframework.cloud.gcp.data.spanner.core.convert.MappingSpannerConverter.mapToList(MappingSpannerConverter.java:83)
...
```

Similar issue w/ `java.sql.Date`.

ChengyuanZhao:
We do specify the java.util.Date converters here (https://github.com/spring-cloud/spring-cloud-gcp/blob/master/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/ConversionUtils.java).

Are you specifying a bean for SpannerConverter yourself rather than using the one from the starter project? If so, you will need to add the public static converters defined in ConversionUtils.java along with the custom converters you added yourself.

saturnism:
That explained it! Thanks @ChengyuanZhao

balopat:
I'm closing this manually - @ChengyuanZhao  you wrote "Addresses #581" in the PR - if you write "Fixes" or "Closes" Github automatically closes the issue!

PR #584, opened 2018-04-16
Add default converters to constructor taking user custom converters

ChengyuanZhao:
addresses #581


ChengyuanZhao:
@balopat tests added

PR #585, opened 2018-04-16
Spanner Date integ test and sql Date support

ChengyuanZhao:
@balopat 

java.util.Date works in integration test without any fix needed....

added converters for java.sql.Date


ChengyuanZhao:
#581

spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/test/domain/Trade.java
```diff
@@ -78,6 +84,13 @@ public static String dropDDL(String tableName) {
 		return "DROP table " + tableName;
 	}
 
+	private String getyyyymmddDate(Date date) {
```
balopat: you can use `DateUtil.truncateTime()` from assertj instead of this method

ChengyuanZhao: change made

Issue #579, opened 2018-04-14
Spanner - Using UUID as a PrimaryKey


saturnism:
I have a POJO that uses `java.util.UUID` as a primary key type mapped to Spanner type String(36). I created the converters.  However, when finding it by ID:
```
repository.findById(myUUID);
```

It throws an error:
```
java.lang.IllegalArgumentException: Unsupported type [java.util.UUID] for argument: f1cf9d40-6e99-4d1f-a717-6db99e60bd82
	at com.google.cloud.spanner.Key$Builder.appendObject(Key.java:187)
	at com.google.cloud.spanner.Key.of(Key.java:80)
	at org.springframework.cloud.gcp.data.spanner.repository.support.SimpleSpannerRepository.doIfKey(SimpleSpannerRepository.java:157)
```

It appears that we need to also convert the object to spanner type prior to constructing the `Key`.

ChengyuanZhao:
This is a really good point. I think we should address this after PR #500 is in. 

The reason is that in that PR I add some code for generating the mapping from given Java types to Spanner compatible types taking into account what custom converters are provided. Making the UUID and other Java PK types work utilizing the given custom converters would be the exact same logic, 
so I think this would be the perfect thing to fix right after #500 is in.

ChengyuanZhao:
@saturnism 
This is supposed to work now after #594.

balopat:
closing as it is resolved in #594

PR #594, opened 2018-04-16
spanner custom converters applied to key components

ChengyuanZhao:
addresses #579

### DIFF
--- a/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/ConversionUtils.java
+++ b/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/ConversionUtils.java
@@ -21,6 +21,7 @@ import java.time.Instant;
 import java.util.Calendar;
 import java.util.Collection;
 import java.util.Date;
+import java.util.Set;
 import java.util.stream.Collectors;
 import java.util.stream.StreamSupport;
 
@@ -128,7 +129,19 @@ public class ConversionUtils {
 			.of(SPANNER_TO_JAVA_DATE_CONVERTER, TIMESTAMP_INSTANT_CONVERTER, SPANNER_TO_JAVA_BYTE_ARRAY_CONVERTER,
 					SPANNER_TO_JAVA_TIMESTAMP_CONVERTER);
 
-	static Class boxIfNeeded(Class propertyType) {
+	/**
+	 * Converters to and from Spanner types.
+	 */
+	public static final Collection<Converter> DEFAULT_SPANNER_CONVERTERS = ImmutableSet
+			.<Converter>builder().addAll(DEFAULT_SPANNER_WRITE_CONVERTERS)
+			.addAll(DEFAULT_SPANNER_READ_CONVERTERS).build();
+
+	public static Set<Class> SPANNER_KEY_COMPATIBLE_TYPES = ImmutableSet.<Class>builder()
+			.add(Boolean.class).add(Integer.class).add(Long.class).add(Float.class)
+			.add(Double.class).add(String.class).add(ByteArray.class).add(Timestamp.class)
+			.add(com.google.cloud.Date.class).build();
+
+	public static Class boxIfNeeded(Class propertyType) {
 		if (propertyType == null) {
 			return null;
 		}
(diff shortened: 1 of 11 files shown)


spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/ConversionUtils.java
```diff
@@ -128,7 +129,19 @@ public ByteArray convert(byte[] bytes) {
...
 
-	static Class boxIfNeeded(Class propertyType) {
+	/**
+	 * Converters to and from Spanner types.
+	 */
+	public static final Collection<Converter> DEFAULT_SPANNER_CONVERTERS = ImmutableSet
+			.<Converter>builder().addAll(DEFAULT_SPANNER_WRITE_CONVERTERS)
+			.addAll(DEFAULT_SPANNER_READ_CONVERTERS).build();
+
+	public static Set<Class> SPANNER_KEY_COMPATIBLE_TYPES = ImmutableSet.<Class>builder()
```
balopat: I'm wondering why we needed to introduce this? Aren't these already in the mapping keySets anyway?

balopat: Oh I see - I guess we can't have arrays for keys, that's the reason, right? It's interesting that ByteArray is supported.

balopat: can we not have this as `public`? And maybe actually move it to the `MappingSpannerConverter` not a utility class? Same applies to `DEFAULT_SPANNER_CONVERTERS`

ChengyuanZhao: changes made

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/support/SimpleSpannerRepository.java
```diff
@@ -175,7 +184,7 @@ private Object convertKeyPart(Object object) {
 		 */
 		for (Class validKeyType : MappingSpannerWriteConverter.singleItemType2ToMethodMap
 				.keySet()) {
-			if (!ConversionUtils.SPANNER_KEY_COMPATIBLE_TYPES.contains(validKeyType)) {
+			if (!SPANNER_KEY_COMPATIBLE_TYPES.contains(validKeyType)) {
```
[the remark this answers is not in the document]
ChengyuanZhao: done

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/support/SimpleSpannerRepository.java
```diff
@@ -154,8 +157,36 @@ public Page findAll(Pageable pageable) {
...
-			k = key instanceof Key ? (Key) key : Key.of(key);
+			k = Key.class.isAssignableFrom(key.getClass()) ? (Key) key
+					: Key.of(convertKeyPart(key));
 		}
 		return operation.apply(k);
 	}
+
+	private Object convertKeyPart(Object object) {
+		if (MappingSpannerConverter
+				.isValidSpannerKeyType(ConversionUtils.boxIfNeeded(object.getClass()))) {
```
balopat: `isValidSpannerKeyType` should be non-static, even if the logic currently depends on a set of statically predefined values. Introducing these hard-coded static depedencies makes the code hard to test and extend later, it's too rigid of a design with no particular benefit.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/support/SimpleSpannerRepository.java
```diff
@@ -154,8 +157,36 @@ public Page findAll(Pageable pageable) {
...
+			return object;
+		}
+		/*
+		 * Iterate through the supported Key component types in the same order as the
+		 * write converter. For example, if a type can be converted to both String and
+		 * Double, we want both the this key conversion and the write converter to choose
+		 * the same.
+		 */
+		for (Class validKeyType : MappingSpannerWriteConverter.singleItemType2ToMethodMap
```
balopat: same thing here - I would rather hide the detail of `singleItemType2ToMethodMap.keySet()` and expose a non-static member function `writableTypes()`

ChengyuanZhao: changes made

Issue #599, opened 2018-04-17
Validate table name returned for SpannerPersistentEntity


balopat:
Due to SpEL expressions table names can end up invalid. 
Whenever we have an expression in `@Table(name=<exp>)`, we should validate what SpannerPersistentEntityImpl returns to let the user know early and to avoid potential security issues.

ChengyuanZhao:
the PR addressing this is in.

PR #603, opened 2018-04-17
Spanner validate table name

ChengyuanZhao:
addresses #599

### DIFF
--- a/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/mapping/SpannerPersistentEntityImpl.java
+++ b/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/mapping/SpannerPersistentEntityImpl.java
@@ -21,6 +21,7 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
+import java.util.regex.Pattern;
 
 import org.springframework.beans.BeansException;
 import org.springframework.context.ApplicationContext;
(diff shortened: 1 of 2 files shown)


spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/mapping/SpannerPersistentEntityImpl.java
```diff
@@ -163,9 +164,19 @@ public void verify() {
...
+		return validateTableName(this.tableNameExpression == null
 				? this.tableName
-				: this.tableNameExpression.getValue(this.context, String.class);
+						: this.tableNameExpression.getValue(this.context, String.class));
+	}
+
+	// Because SpEL expressions in table name definitions are allowed, validation is
+	// required.
+	private String validateTableName(String name) {
+		if (Pattern.compile("[^a-zA-Z0-9_]").matcher(name).find()) {
```
meltsufin: Can we avoid recompiling the pattern every time?

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/mapping/SpannerPersistentEntityImpl.java
```diff
@@ -163,9 +164,19 @@ public void verify() {
...
+			throw new SpannerDataException("Only letters, numbers, and underscores are "
+					+ "allowed in table names.");
```
meltsufin: Can we also output the table name in the message of the exception to ease debugging?

ChengyuanZhao: changes made

PR #611, opened 2018-04-18
Spanner db admin schema sample doc update

### LINKED ISSUE #609: Add Spanner schema generation to samples and ref doc

ChengyuanZhao:
fixes #609

### DIFF
--- a/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/admin/SpannerDatabaseAdminTemplate.java
+++ b/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/admin/SpannerDatabaseAdminTemplate.java
@@ -23,7 +23,11 @@ import java.util.Set;
 
 import com.google.cloud.spanner.Database;
 import com.google.cloud.spanner.DatabaseAdminClient;
+import com.google.cloud.spanner.DatabaseClient;
 import com.google.cloud.spanner.DatabaseId;
+import com.google.cloud.spanner.ResultSet;
+import com.google.cloud.spanner.Statement;
+import com.google.cloud.spanner.Struct;
 
 import org.springframework.cloud.gcp.data.spanner.core.mapping.SpannerDataException;
 import org.springframework.util.Assert;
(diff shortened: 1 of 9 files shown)


spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/admin/SpannerDatabaseAdminTemplate.java
```diff
@@ -140,4 +140,29 @@ public boolean databaseExists() {
 		}
 		return relationships;
 	}
+
+	/**
+	 * Return a set of the tables that currently exist in the database.
+	 * @return A set of table names.
+	 */
+	public Set<String> getTables() {
```
balopat: A rule of thumb: we really really don't want to parse SQL ever in this module unless it's very necessary. 
I found a better way to list tables: https://cloud.google.com/spanner/docs/information-schema

ChengyuanZhao: I definitely agree and wish there was a way to get these schema info from the DB admin client, 
but it looks like the way mentioned requires actually querying Spanner with SQL query
```
SELECT
  t.table_name
FROM
  information_schema.tables AS t
WHERE
  t.table_catalog = '' and t.table_schema = '
```

This feels a bit heavy handed. I thought it was an acceptable compromise because the DDL strings we get are one statement per item in the list, and indeed comprises the "CREATE TABLE" strings.

balopat: I agree that it is a bit inconvenient - however I'd rather inject a dbClient into the dbAdminClient and query it from the dedicated, already parsed schema info (which is btw the typical pattern in all relational databases) - then go with this approach which depends on parsing the schema.

ChengyuanZhao: Could you explain how it is more robust? 

It looks like we would be swapping the likelihood of "CREATE TABLE" statements changing vs the name of the schema table + column names changing.

balopat: Thank you for implementing the change. 

The ddl functionality has the following promise: "we give you executable DDL statements that will define a database schema exactly the same as your current database has". This happens to be in SQL. Which happens to be very stable and has a small chance of changing. But it is a secondary aspect of the "DDL" feature. 

The information schema actually has a higher chance of changing but has the following promise: "we give you a schema information table that you are recommended to depend on to query insight about the schema of your database". The primary and sole reason of this information schema is to retrieve things like tables. It seems like the right thing to use for our current use case, but more importantly it is a _forward looking approach in case we want to retrieve other schema related information_ (e.g. primary keys, indices, etc.) in an easy way as we evolve our codebase. 

I think the "choose the right tool for the right job" argument and the evolvability argument far trumps the counter-argument of wiring an extra bean into the DatabaseAdminClient.

PR #613, opened 2018-04-19
expose Spanner Options

### LINKED ISSUE #591: configurability of SpannerOptions and SessionPoolOptions as properties
Expose the options as properties available on [`SpannerOptions`](https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SpannerOptions.java), it is very important to have a look at `SessionPoolOptions` in particular as large use cases will want fine tuneability on these knobs. 

Please discuss here first the details before jumping into implementation.

ChengyuanZhao:
fixes #591

### DIFF
--- a/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/spanner/GcpSpannerAutoConfiguration.java
+++ b/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/spanner/GcpSpannerAutoConfiguration.java
@@ -117,9 +117,7 @@ public class GcpSpannerAutoConfiguration {
 		if (this.prefetchChunks >= 0) {
 			builder.setPrefetchChunks(this.prefetchChunks);
 		}
-		if (sessionPoolOptions != null) {
-			builder.setSessionPoolOption(sessionPoolOptions);
-		}
+		builder.setSessionPoolOption(sessionPoolOptions);
 		return builder.build();
 	}
 
--- a/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/spanner/GcpSpannerAutoConfiguration.java
+++ b/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/spanner/GcpSpannerAutoConfiguration.java
@@ -22,8 +22,10 @@ import com.google.api.gax.core.CredentialsProvider;
 import com.google.auth.Credentials;
 import com.google.cloud.spanner.DatabaseClient;
 import com.google.cloud.spanner.DatabaseId;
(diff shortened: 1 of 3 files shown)


spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/spanner/GcpSpannerAutoConfiguration.java
```diff
@@ -78,15 +94,59 @@ public GcpSpannerAutoConfiguration(GcpSpannerProperties gcpSpannerProperties,
...
 	@ConditionalOnMissingBean
-	public SpannerOptions spannerOptions() {
-		return SpannerOptions.newBuilder()
+	public SpannerOptions spannerOptions(SessionPoolOptions sessionPoolOptions) {
+		Builder builder = SpannerOptions.newBuilder()
 				.setProjectId(this.projectId)
 				.setHeaderProvider(new UsageTrackingHeaderProvider(this.getClass()))
-				.setCredentials(this.credentials).build();
+				.setCredentials(this.credentials);
+		if (this.numRpcChannels > 0) {
```
balopat: I think you meant `>=0` as the default is -1?

balopat: (this applies for the rest of the properties too)

ChengyuanZhao: fixed!

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/spanner/GcpSpannerAutoConfiguration.java
```diff
@@ -78,15 +94,59 @@ public GcpSpannerAutoConfiguration(GcpSpannerProperties gcpSpannerProperties,
...
 				.setHeaderProvider(new UsageTrackingHeaderProvider(this.getClass()))
-				.setCredentials(this.credentials).build();
+				.setCredentials(this.credentials);
+		if (this.numRpcChannels >= 0) {
+			builder.setNumChannels(this.numRpcChannels);
+		}
+		if (this.prefetchChunks >= 0) {
+			builder.setPrefetchChunks(this.prefetchChunks);
+		}
+		if (sessionPoolOptions != null) {
```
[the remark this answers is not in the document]
ChengyuanZhao: I will remove the check. My original thought is that  the user can override the sessionpooloptions bean with one of his own that does return null. (But in both cases the SpannerOptions  is ok with it being null)

PR #616, opened 2018-04-19
Spanner check converters cover all functions

### LINKED ISSUE #441: Add a unit test that checks the Spanner reader and writer classes for new functions
To check if there are new reading/writing functions handing types that we haven't manually included in our object mapper. The failure of the test would indicate that we need to manually add more cases to our type-method mapping.

ChengyuanZhao:
fixes #441


ChengyuanZhao:
@balopat PTAL, this is a pretty small one

PR #618, opened 2018-04-19
Spanner pageable sortable SQL methods

### LINKED ISSUE #601: Supporting Pageable and Sort in @Query annotations
This appears in the JPA project as an example

ChengyuanZhao:
fixes #601


spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SqlSpannerQuery.java
```diff
@@ -58,19 +63,32 @@ public SqlSpannerQuery(Class type, QueryMethod queryMethod,
 		this.tags = getTags();
 		this.evaluationContextProvider = evaluationContextProvider;
 		this.expressionParser = expressionParser;
-		this.sql = sql;
+		this.sql = StringUtils.trimTrailingCharacter(sql.trim(), ';');
+		if (this.sql.contains(";")) {
```
meltsufin: What if the query has a where clause like:
```
where sybol = ';'
```
This seems pretty fragile. What does Spring Data JDBC do in this scenario?

ChengyuanZhao: ah i keep forgetting that. I think we just shouldn't have that check there.

ChengyuanZhao: I'm not sure if this appears in spring-data-jdbc. In any case, with Spanner client lib we must specify the, limit, offset, and sort for queries in the SQL string.

ChengyuanZhao:
I removed a bunch of SpannerOperations methods in doing this issue, which I think is a nice side effect. 
We also used to have Pageable parameters in SpannerOperations, which was very confusing because SpannerQueryOperations also provided limit/offset settings. I took out the Pageable, so now it is less confusing. However, because the only way to do limit/offset in Spanner is inside the SQL string, and since Statement objects are immutable, I had to expose a SQL-string parameter in one the query methods.
@balopat

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
```diff
@@ -145,46 +154,35 @@ public SpannerConverter getSpannerConverter() {
...
-
-	@Override
-	public <T> Page<T> queryAll(Class<T> entityClass, Pageable pageable) {
-		return queryAll(entityClass, pageable, new SpannerQueryOptions()
-				.setOffset(pageable.getOffset()).setLimit(pageable.getPageSize()));
+	private String applySort(Sort sort, String sql) {
+		if (sort == null || sort.isUnsorted()) {
+			return sql;
+		}
+		String s = "SELECT * FROM (" + sql + ") ORDER BY ";
```
balopat: Cool idea for guarding against existing clauses :) I hope there are no extra cost to it (it seems that way, that Spanner is smart enough to optimize issues from little experiments on my console)

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
```diff
@@ -145,46 +154,36 @@ public SpannerConverter getSpannerConverter() {
...
+		return query(entityClass, applySortingPagingQueryOptions(options, sql), null,
+				null, options);
+	}
+
+	public String applySortingPagingQueryOptions(SpannerQueryOptions options,
+			String sql) {
+		String r = applySort(options.getSort(), sql);
 			if (options.hasLimit()) {
-				stringBuilder.append(" LIMIT " + options.getLimit());
+			r += " LIMIT " + options.getLimit();
```
balopat: why not StringBuilder? This creates an extra String instace.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SqlSpannerQuery.java
```diff
@@ -105,9 +117,47 @@ private String resolveEntityClassNames(String sql) {
...
+
+		for (Object param : parameters) {
+			Class paramClass = param.getClass();
+			if (isPageableSort(paramClass)) {
+				if (pageable != null || sort != null) {
+					throw new SpannerDataException(
+							"Only a single Pageable or Sort param is allowed.");
+				}
+				else {
+					if (Pageable.class.isAssignableFrom(paramClass)) {
```
balopat: at this point we are 4 nesting deep, please refactor it to smaller/cleaner!

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
```diff
@@ -145,46 +154,36 @@ public SpannerConverter getSpannerConverter() {
...
+	}
+
+	public String applySortingPagingQueryOptions(SpannerQueryOptions options,
+			String sql) {
+		StringBuilder r = new StringBuilder(applySort(options.getSort(), sql));
 			if (options.hasLimit()) {
-				stringBuilder.append(" LIMIT " + options.getLimit());
+				r.append(" LIMIT " + options.getLimit());
 			}
 			if (options.hasLimit()) {
```
meltsufin: Did you mean `hasOffset()`?

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
```diff
@@ -145,46 +154,36 @@ public SpannerConverter getSpannerConverter() {
...
-	}
-
-	@Override
-	public <T> Page<T> queryAll(Class<T> entityClass, Pageable pageable) {
-		return queryAll(entityClass, pageable, new SpannerQueryOptions()
-				.setOffset(pageable.getOffset()).setLimit(pageable.getPageSize()));
+	private String applySort(Sort sort, String sql) {
+		String s = "SELECT * FROM (" + sql + ")";
+		if (sort == null || sort.isUnsorted()) {
+			return s;
```
meltsufin: Why not just return `sql`? The additional subquery just adds an unnecessary performance hit.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SqlSpannerQuery.java
```diff
@@ -105,9 +117,45 @@ private String resolveEntityClassNames(String sql) {
...
+			if (isPageableSort(paramClass)) {
+				if (pageable != null || sort != null) {
+					throw new SpannerDataException(
+							"Only a single Pageable or Sort param is allowed.");
+				}
+				else if (Pageable.class.isAssignableFrom(paramClass)) {
+						pageable = (Pageable) param;
+				}
+				else {
+					sort = (Sort) param;
```
meltsufin: Will it necessarily be `Sort` if it's not `Pageable`? It might be safer to just verify before casting.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
ChengyuanZhao: Are you sure that it does? It seems like something that could be optimized away.

This is needed because the user might already have had a LIMIT and OFFSET at the end of the query and our stated policy in the issue is to apply this on top of given query.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SqlSpannerQuery.java
ChengyuanZhao: Yes. The outer if-statement checks if the type is a child of either pageable or sort. and if it isn't a child of pageable, then it must be a child of sort.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
ChengyuanZhao: will fix.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
ChengyuanZhao: To clarify, applySort is used by applySortingPagingQueryOptions, and even if there is no sort the LIMIT and OFFSET bits need to be applied by applySortingPagingQueryOptions after the inner SQL is surrounded by parentheses.

meltsufin: It seems like the wrapping into a subquery then maybe should be done before the call to `applySort`.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SqlSpannerQuery.java
meltsufin: I think you can simplify the logic in this method by just having a single Object `pageableOrSort` and then checking its type at the bottom. Just and idea.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
ChengyuanZhao: The downside to that is then applySort  cannot be used by the user without the user wrapping the SQL themselves.

meltsufin: But this is a private method. How would the user ever call it directly?

ChengyuanZhao: By "user" i meant any other function that may call this function. It doesn't feel clean to need to mess with the SQL string if you're about to call a function that also touches the contents of this SQL string. I prefer to keep all of the string-touching regarding sorting and pageable within these two functions.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SqlSpannerQuery.java
ChengyuanZhao: Thank you for the suggestion, but I think it is simple as it is.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SqlSpannerQuery.java
```diff
@@ -66,27 +70,29 @@ public SqlSpannerQuery(Class type, QueryMethod queryMethod,
...
+		this.sql = StringUtils.trimTrailingCharacter(sql.trim(), ';');
+	}
+
+	private boolean isPageableSort(Class type) {
+		return Pageable.class.isAssignableFrom(type) || Sort.class.isAssignableFrom(type);
 	}
 
 	private List<String> getParamTags() {
 		List<String> tags = new ArrayList<>();
 		Set<String> seen = new HashSet<>();
```
meltsufin: This is not used anymore.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SqlSpannerQuery.java
```diff
@@ -66,15 +70,23 @@ public SqlSpannerQuery(Class type, QueryMethod queryMethod,
 		super(type, queryMethod, spannerOperations, spannerMappingContext);
 		this.evaluationContextProvider = evaluationContextProvider;
 		this.expressionParser = expressionParser;
-		this.sql = sql;
+		this.sql = StringUtils.trimTrailingCharacter(sql.trim(), ';');
+	}
+
+	private boolean isPageableSort(Class type) {
```
balopat: recommend: `isPageableOrSort`

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerQueryOptions.java
```diff
@@ -135,4 +138,14 @@ public SpannerQueryOptions setAllowPartialRead(
 		this.allowPartialRead = allowPartialRead;
 		return this;
 	}
+
+	public Sort getSort() {
+		return this.sort;
+	}
+
```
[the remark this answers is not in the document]
ChengyuanZhao: changes made !

ChengyuanZhao:
@balopat changes have been made, PTAL

balopat:
added some nits, if you agree we can go ahead and merge it

ChengyuanZhao:
@balopat yup I'm good with them

PR #619, opened 2018-04-20
Introducing non-default constructors

### LINKED ISSUE #593: Support non-default constructors for mapped POJOs
With Spring Data's EntityInstantiator, it should be relatively straightforwad to support non-default constructors for mapped POJOs:

````
SpannerPersistentEntity<?> persistentEntity = this.spannerMappingContext
     .getPersistentEntity(type);
EntityInstantiator instantiator = entityInstantiators.getInstantiatorFor(persistentEntity);
instantiator.createInstance(persistentEntity, new ParameterValueProvider<SpannerPersistentProperty>() {
  @Override
  public <T> T getParameterValue(PreferredConstructor.Parameter<T, SpannerPersistentProperty> parameter) {
//...
  }
});

````

Note: We should pay attention to include testcases including complex types for parameters as well.

### PR DESCRIPTION
**Pull request by balopat**
Closes #593. 

- [x] refactor SpannerObjectMapper into two separate SpannerWrite/ReadConverterTest (#617)
- [x] refactor SpannerReadConverter to be more functional with `StructAccessor` so that it can be reused for the constructor parameter value accessor
- [x] write the first test to cover a simple valued constructor and implement changes
- [x] separate conversion logic from property read logic from struct (#631)
- [x] introduce more tests to cover more advanced scenarios
- [x] add documentation

### CODE REVIEW

**Comment by balopat** [comment]
@mp911de can you please have a look at this approach? Let me know if I'm missing something! 
Thank you!

File: spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/MappingSpannerReadConverter.java [hunk 1]
```diff
@@ -73,163 +78,201 @@
...
-			return false;
+		StructAccessor(Struct struct) {
+			this.struct = struct;
 		}
 
-		// due to checkstyle limit of 3 return statments per function, this variable is
-		// used.
-		boolean valueSet = false;
+		public Object get(SpannerPersistentProperty spannerPersistentProperty) {
+			return ConversionUtils.isIterableNonByteArrayType(spannerPersistentProperty.getType())
```
mp911de: `StructAccessor` currently has multiple responsibilities: Access and conversion.

Typically, conversion is part of the converter itself. Accessors retrieve the actual value and use a converter method (See [`MongoDbPropertyValueProvider`](https://github.com/spring-projects/spring-data-mongodb/blob/239304160b35e594acd584a75556bc971d68ad4b/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/convert/MappingMongoConverter.java#L1323) and [`MappingMongoConverter.read(…)`](https://github.com/spring-projects/spring-data-mongodb/blob/239304160b35e594acd584a75556bc971d68ad4b/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/convert/MappingMongoConverter.java#L1374))

This way, you can apply the same conversion for properties and constructor properties and keep responsibilities simple.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/MappingSpannerReadConverter.java
```diff
@@ -73,163 +78,201 @@
...
-					e);
+
+		public boolean isNull(String columnName) {
+			return this.struct.isNull(columnName);
 		}
-		return object;
-	}
 
+		public boolean hasColumn(String columnName) {
+			return this.struct.getType().getStructFields().stream()
```
mp911de: Using Java 8 `Stream` in conversion/mapping code paths has a negative impact on performance (in MongoDB we had a 100% performance impact, mapping took twice as long using Java 8 `Stream`).

mp911de:
The direction looks good. I left some comments.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/MappingSpannerReadConverter.java
```diff
@@ -73,163 +78,201 @@
...
-			String colName, PersistentPropertyAccessor accessor) {
-		Class targetType = spannerPersistentProperty.getType();
-		if (sourceType == null || !canConvert(sourceType, targetType)) {
-			return false;
-		}
-		BiFunction readFunction = singleItemReadMethodMapping
-				.get(ConversionUtils.boxIfNeeded(sourceType));
-		if (readFunction == null) {
-			return false;
+	private <R> boolean shouldSkipProperty(StructAccessor structAccessor,
```
ChengyuanZhao: This type parameter `<R>` isn't necessary actually. The reason is that the only place it is used is as a constraint in an argument. This is saying "give me a PreferredConstruct, whose first type param must extend R, but I never say what R should be and the only thing telling me what R is the PreferredConstructor itself". As  a result it's not actually a restriction at all.

For example, if instead you had this method signature as `private <R,S> boolean shouldSkipProperty`, 
and the argument as `PreferredConstructor<? extends S, SpannerPersistentProperty> persistenceConstructor` everything still compiles and works despite there being total ambiguity about what R and S are.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/MappingSpannerReadConverter.java
```diff
@@ -73,163 +78,201 @@
...
+				return true;
+			}
 		}
-		accessor.setProperty(spannerPersistentProperty,
-				convert(readFunction.apply(struct, colName), targetType));
-		return true;
+		return false;
+	}
+
+	private <R> PersistentEntityParameterValueProvider<SpannerPersistentProperty> getParameterValueProvider(
```
ChengyuanZhao: same comment as above

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/MappingSpannerReadConverter.java
```diff
@@ -73,163 +78,201 @@
...
+			SpannerPersistentEntity<? extends R> persistentEntity, StructAccessor structAccessor) {
+		return new PersistentEntityParameterValueProvider<>(
+				persistentEntity, new PropertyValueProvider<SpannerPersistentProperty>() {
+					@Override
+					public <T> T getPropertyValue(SpannerPersistentProperty property) {
```
ChengyuanZhao: the type param T is only used to cast the return and is not actually restricting anything because none of the arguments or methods called inside this method restrict T. You could just return Object type from getPropertyValue

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/MappingSpannerReadConverter.java
balopat: Thank you, I'll change that, and check for other parts too.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/MappingSpannerReadConverter.java
balopat: Sounds good.

balopat:
@mp911de thank you for your comments - it led me down a bigger refactoring - but I think it helped the design. The new version has a better separation of concerns. Please take a look! 

@ChengyuanZhao PTAL as well, I took your comments into account too, thank you!

mp911de:
Looks way cleaner now with clear separation of concerns.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/MappingSpannerReadConverter.java
```diff
@@ -218,30 +152,7 @@ else if (canConvert(sourceClass, targetType)) {
...
-		Type.Code innerTypeCode = struct.getColumnType(colName).getArrayElementType().getCode();
-		Class clazz = this.spannerTypeMapper.getSimpleJavaClassFor(innerTypeCode);
-		BiFunction<Struct, String, List> readMethod = readIterableMapping.get(clazz);
-		List listValue = readMethod.apply(struct, colName);
-
-		return convertList(listValue, spannerPersistentProperty.getColumnInnerType());
-	}
-
-	private List<Object> convertList(List listValue, Class targetInnerType) {
+	List<Object> convertList(List listValue, Class targetInnerType) {
```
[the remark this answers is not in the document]
balopat: Sounds good.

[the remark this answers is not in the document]
balopat: Actually I think we can even get rid of this - we have a `ConversionUtils.convertIterable`!

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/MappingSpannerReadConverter.java
```diff
@@ -115,44 +66,61 @@
 	public <R> R read(Class<R> type, Struct source, Set<String> includeColumns,
```
ChengyuanZhao: Now that we've gone down this road of separating concerns and such, I don't think a method named "read" that reads from a source and gets names of pieces of data (columns) to read belongs in a "converter". 

I think we should have a separate SpannerEntityReader class and that this converter should not implement it.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/MappingSpannerReadConverter.java
```diff
@@ -46,60 +40,17 @@
...
-			constructor.setAccessible(true);
-			object = constructor.newInstance();
-		}
-		catch (ReflectiveOperationException e) {
-			throw new SpannerDataException(
-					"Unable to create a new instance of entity using default constructor.",
-					e);
-		}
-		return object;
+		this.instantiators = new EntityInstantiators();
```
ChengyuanZhao: EntityInstantiators class has many constructors it seems. We should take this in the constructor to account for the potentially different EntityInstantiator instances that the user could provide.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/StructPropertyValueProvider.java
```diff
@@ -0,0 +1,82 @@
...
+							+ " could not be converted to the corresponding property in the entity."
+							+ " The property's type is %s.",
+					colName, propType));
+		}
+		return value;
+
+	}
+
+	private Object readSingleWithConversion(SpannerPersistentProperty spannerPersistentProperty,
+			StructAccessor struct) {
```
ChengyuanZhao: Why take in the StructAccessor as the parameter instead of using this.structaccessor??

spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/core/convert/TestEntities.java
```diff
@@ -105,9 +105,52 @@
...
+		SimpleConstructorTester(String id) {
+			this.id = id;
+		}
+	}
 
 	static class TestEntityWithListWithZeroTypeArgs {
 		@PrimaryKey
 		List zeroArgsListOfObjects;
 	}
+
```
ChengyuanZhao: Are constructors where the type of the arguments don't directly match the type of the properties supported?

for example 
`this.stringLength = givenString.length();` 
instead of `this.string = givenString;`

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/ConversionUtils.java
```diff
@@ -43,8 +43,8 @@ public static boolean isIterableNonByteArrayType(Class propType) {
 
 	static Iterable convertIterable(
```
ChengyuanZhao: Could the type parameter be added? 

`static <T> Iterable<T> convertIterable(Iterable source, Class<T> targetType....`

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/StructPropertyValueProvider.java
```diff
@@ -0,0 +1,82 @@
...
+		String colName = spannerPersistentProperty.getColumnName();
+		Class targetType = spannerPersistentProperty.getType();
+		Object value = struct.getSingleValue(colName);
+		return this.readConverter.convert(value, targetType);
+	}
+
+	private Object readIterableWithConversion(SpannerPersistentProperty spannerPersistentProperty,
```
ChengyuanZhao: The return type here can be `Iterable`

spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/core/convert/TestEntities.java
balopat: That's a really good test case, I think it should work, but let me add it.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/MappingSpannerReadConverter.java
```diff
@@ -115,44 +66,61 @@
...
+		// @formatter:on
+
+		EntityInstantiator instantiator = this.instantiators.getInstantiatorFor(persistentEntity);
+		Object instance = instantiator.createInstance(persistentEntity, parameterValueProvider);
+		PersistentPropertyAccessor accessor = persistentEntity.getPropertyAccessor(instance);
 
 		persistentEntity.doWithProperties(
 				(PropertyHandler<SpannerPersistentProperty>) spannerPersistentProperty -> {
-					if (!shouldSkipProperty(source,
+					if (!shouldSkipProperty(
```
ChengyuanZhao: This `shouldSkipProperty` function is now effectively just the few function calls with ORs among them. I think you should just do those OR checks inside this if-statement here. 

For somebody not familiar with the code this `shouldSkipProperty` function looks like a mysterious function that takes half a dozen very specific and uncommon arguments with obvious connections to each other.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/ConversionUtils.java
balopat: ack

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/MappingSpannerReadConverter.java
balopat: yeah, the MongoDb version has [setters defined in the abstract class ](https://github.com/spring-projects/spring-data-mongodb/blob/master/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/convert/AbstractMongoConverter.java) of the converter. 
@mp911de can you weigh in on the preferred approach around making instantiators configurable?

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/MappingSpannerReadConverter.java
```diff
@@ -217,36 +149,4 @@ else if (canConvert(sourceClass, targetType)) {
 		}
```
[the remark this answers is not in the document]
balopat: Sure! I'd file a separate issue for supporting custom Struct conversion and even Struct fields.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/StructPropertyValueProvider.java
balopat: artifact of moving things around, good catch!

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/StructPropertyValueProvider.java
balopat: ack

spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/core/convert/TestEntities.java
balopat: I tried it out and it is not supported. Mainly because I'm using `PersistentEntityParameterValueProvider` (similar to MongoDb) that assumes that constructor parameters are mapped to the properties of the entity. We can file a separate issue for making this work.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/MappingSpannerReadConverter.java
balopat: This is a simple readability method. 
It is simpler, shorter. 
It means that there is a method that has the property skipping logic, that depends on all of these things. If you're interested, have a look at it. 
I tend to use these as I believe that they result in [cleaner code](https://medium.com/coding-skills/clean-code-101-meaningful-names-and-functions-bf450456d90c).

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/MappingSpannerReadConverter.java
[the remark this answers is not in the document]
ChengyuanZhao: Why does this need to be a separate issue? What is the difference in a custom converter from the user config between an arbitrary type and Struct vs a customer converter from an arbitrary type to another arbitrary type.

spring-cloud-gcp-docs/src/main/asciidoc/spanner.adoc
```diff
@@ -114,6 +114,33 @@ public class Trader {
 Spring Data Spanner will ignore any property annotated with `@NotMapped`. These properties will
 not be written to or read from Spanner.
 
+==== Constructors
+
+Simple constructors are supported on POJOs. The constructor arguments can be a subset of the persistent properties or all of them. Constructor arguments that do not map to a persistent property are not supported.
```
ChengyuanZhao: You can leave out "or all of them". subsets also include the set itself

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerMutationFactoryImpl.java
```diff
@@ -44,7 +44,7 @@
 
 	/**
 	 * Constructor
-	 * @param spannerConverter The object mapper used to convert between objects and Spanner
+	 * @param spannerConverter The object mapper used to readConvert between objects and Spanner
```
balopat: I'll need to revert the comment changes

spring-cloud-gcp-docs/src/main/asciidoc/spanner.adoc
balopat: done

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/SpannerReadConverter.java
```diff
@@ -0,0 +1,48 @@
...
+		}
+		else if (super.canConvert(sourceClass, targetType)) {
+			result = super.convert(sourceValue, targetType);
+		}
+		return result;
+	}
+
+	@Override
+	protected boolean canConvert(Class sourceType, Class targetType) {
+		return super.canConvert(sourceType, targetType);
```
ChengyuanZhao: It looks like this is literally calling the parent version; is this overrided method necessary at all?

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/SpannerReadConverter.java
```diff
@@ -0,0 +1,48 @@
...
+ * @author Balint Pato
+ */
+public class SpannerReadConverter extends AbstractSpannerCustomConverter {
+
+	public SpannerReadConverter(CustomConversions customConversions) {
+		super(customConversions, null);
+	}
+
+	@Override
+	public <T> T convert(Object sourceValue, Class<T> targetType) {
```
ChengyuanZhao: It looks like the only difference between this method and the parent version is that the parent does a check if the desired type is the same as that of source rather than here if it is a child type. I think this difference is more an error and  we should have done the parent method the way you have done this. 

We should change the parent version to do the check for child type rather than equals-type. With that it doesn't seem like this class has a reason to exist anymore right?

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/MappingSpannerConverter.java
```diff
@@ -45,59 +46,67 @@
...
  */
-public class MappingSpannerConverter extends AbstractSpannerCustomConverter
-		implements SpannerConverter {
-
-	private static final Collection<Converter> DEFAULT_SPANNER_CONVERTERS = ImmutableSet
-			.<Converter>builder()
-			.addAll(SpannerConverters.DEFAULT_SPANNER_WRITE_CONVERTERS)
-			.addAll(SpannerConverters.DEFAULT_SPANNER_READ_CONVERTERS).build();
-
+public class MappingSpannerConverter implements SpannerConverter {
```
ChengyuanZhao: I think we should rename `SpannerConverter` and `MappingSpannerConverter` at this point. We're pretty heavily invested in not having "converters" do the reading or writing, so maybe the names could be `SpannerCorrespondent` and `MappingSpannerCorrespondent`

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/SpannerReadConverter.java
balopat: Oh, yeah I wanted to tackle that, thanks for catching it!

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/MappingSpannerConverter.java
```diff
@@ -246,8 +265,13 @@ public void write(Object source, WriteBuilder sink, Set<String> includeColumns)
 		this.writeConverter.write(source, sink, includeColumns);
 	}
 
+	@Override
+	public <R> R read(Class<R> type, Struct source, Set<String> includeColumns, boolean allowMissingColumns) {
+		return this.entityReader.read(type, source, includeColumns, allowMissingColumns);
+	}
+
 	@Override
 	public <R> R read(Class<R> type, Struct source) {
```
ChengyuanZhao: Does this overriding method have any effect? It looks like the `default` version defined calls the other more-fully defined `read` method, which is exactly what we're doing here because `this.entityReader.read` is just `default` version.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/SpannerConverter.java
```diff
@@ -75,15 +75,32 @@
...
+	boolean canWriteConvert(Class sourceType, Class targetType);
 
 	/**
 	 * Converts the given object to the desired target type.
 	 * @param source the object to convert
 	 * @param targetType the desired target type
 	 * @return the converted object.
 	 */
-	Object convert(Object source, Class targetType);
+	Object readConvert(Object source, Class targetType);
```
ChengyuanZhao: Not an error any thing, but it's interesting that this function is greyed out in my intellij and it's not used anywhere.

Also, it seems like the `canReadConvert` function is only used in a test, too. So perhaps these two functions aren't needed in this interface?

spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/core/convert/MappingSpannerConverterTests.java
```diff
@@ -99,13 +99,10 @@ public void canConvertDefaultTypesCustomConverters() {
 	private void verifyCanConvert(MappingSpannerConverter converter, Class javaType,
 			Class spannerType) {
 		MappingSpannerWriteConverter writeConverter = converter.getWriteConverter();
-		MappingSpannerReadConverter readConverter = converter.getReadConverter();
+		ConverterAwareMappingSpannerEntityReader readConverter = converter.getEntityReader();
 
-		assertTrue(converter.canConvert(javaType, spannerType));
-		assertTrue(converter.canConvert(spannerType, javaType));
-
-		assertTrue(writeConverter.canConvert(javaType, spannerType));
```
ChengyuanZhao: Why were these two lines removed?
`assertTrue(writeConverter.canConvert(javaType, spannerType));`
`assertTrue(readConverter.canConvert(spannerType, javaType));`

These are supposed to confirm that the read and write sub converters can handle the necessary conversions in the read and write directions, respectively.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/MappingSpannerWriteConverter.java
```diff
@@ -32,23 +32,25 @@
...
 import org.springframework.data.mapping.PersistentPropertyAccessor;
 import org.springframework.data.mapping.PropertyHandler;
 
 /**
  * @author Balint Pato
  * @author Chengyuan Zhao
  */
-public class MappingSpannerWriteConverter extends AbstractSpannerCustomConverter
-		implements SpannerEntityWriter {
+public class MappingSpannerWriteConverter implements SpannerEntityWriter {
```
ChengyuanZhao: Now that you've gone through the hard work of removing the convert functions from this, should it be renamed to `MappingSpannerWriter` ?

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/MappingSpannerConverter.java
```diff
@@ -31,78 +31,62 @@
...
+		this.readConverter = new SpannerReadConverter(readConverters);
+		this.entityReader = new ConverterAwareMappingSpannerEntityReader(spannerMappingContext, this.readConverter);
+		this.writeConverter = new SpannerWriteConverter(writeConverters);
+		this.entityWriter = new MappingSpannerWriteConverter(spannerMappingContext, this.writeConverter);
 	}
 
 	@VisibleForTesting
-	MappingSpannerReadConverter getReadConverter() {
-		return this.readConverter;
+	ConverterAwareMappingSpannerEntityReader getEntityReader() {
```
ChengyuanZhao: It looks like this method (and the one below it) aren't used anymore (including tests).

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/SpannerConverter.java
balopat: sure, we can remove this delegation to the specific converters, I'm up for that - we have to expose the right converters though for - which is already kind of happening. I'm going to experiment with it in the next commit.

balopat:
Re your points 1, 2, 3: 

1.) How about `ConverterAwareMappingSpannerEntityReader`, `ConverterAwareMappingSpannerEntityWriter` and the facade class would be `ConverterAwareMappingSpannerEntityProcessor` implementing `SpannerEntityProcessor`
2.) I don't understand the question - `read(Class<R> type, Struct source)` is defined in `EntityReader`, that I would like to keep, but we have a richer `read(Class<R> type, Struct source, Set<String> includeColumns, boolean allowMissingColumns)` that is also exposed and the basic `read` refers to it - like telescopic constructors...
3.) I removed these in the next commit coming up.

balopat:
@ChengyuanZhao I addressed all the above in the last commits, PTAL.

balopat:
@ChengyuanZhao can you please reapprove after my merge of master?

PR #621, opened 2018-04-20
Spanner SpEL generated tags

### LINKED ISSUE #589: SpEL expressions for queries
SpEL expressions allow for dynamic resolution of final SQL queries with beans used from the application context or query method arguments.
For design see for example [this](https://github.com/spring-projects/spring-data-jpa/blob/549cfd9cf8dd94a9f7fdcf73c167290550cfff1c/src/test/java/org/springframework/data/jpa/repository/sample/UserRepository.java#L447) implementation, with [this spec](https://jira.spring.io/browse/DATAJPA-1140)

ChengyuanZhao:
fixes #589

### DIFF
--- a/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SqlSpannerQuery.java
+++ b/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SqlSpannerQuery.java
@@ -17,8 +17,13 @@
 package org.springframework.cloud.gcp.data.spanner.repository.query;
 
 import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
+import java.util.Set;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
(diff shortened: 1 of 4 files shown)


spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SqlSpannerQuery.java
```diff
@@ -103,11 +117,93 @@ private String resolveEntityClassNames(String sql) {
...
+		Expression[] expressions = detectExpressions(queryTagValue.sql);
+		StringBuilder sb = new StringBuilder();
+		Map<Object, String> valueToTag = new HashMap<>();
+		int tagNum = 0;
+		for (Expression expression : expressions) {
+			if (expression instanceof LiteralExpression) {
+				sb.append(expression.getValue(String.class));
+			}
+			else if (expression instanceof SpelExpression) {
+				Object value = expression.getValue(this.evaluationContextProvider
```
balopat: we should get the evaluation context outside of this loop

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SqlSpannerQuery.java
```diff
@@ -103,11 +117,93 @@ private String resolveEntityClassNames(String sql) {
...
+						.getEvaluationContext(this.queryMethod.getParameters(),
+								queryTagValue.intialParams));
+				if (valueToTag.containsKey(value)) {
+					sb.append("@" + valueToTag.get(value));
+				}
+				else {
+					String newTag;
```
balopat: this is an interesting caching solution - first it made me look again, but I think it's fine. 
nit: can you please extract the do-while statement to a private method: `getNextSyntheticTag`?

balopat:
Overall looks great, see my nits comments above

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SqlSpannerQuery.java
ChengyuanZhao: I would rather keep it in this method. It is a small bit of code, and I find it more annoying than helpful to have it moved out.

other changes made.

Issue #625, opened 2018-04-21
Logging - JSON Logged Trace ID needs to conform format: projects/PROJECT_ID/traces/128_BIT_TRACEID


saturnism:
Currently the trace ID is outputted like this:
```
{"traceId":"58f6bbea713ab60e", ... }
```
It should be:
```
{"traceId":"projects/PROJECT_ID/[PREPENDED WITH 0]58f6bbea713ab60e",
```

saturnism:
@joaoandremartins is there anyway we could extrapolate the current project ID that's already configured? Perhaps w/ `springProperty`?
https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#_environment_properties

meltsufin:
@saturnism I wasn't aware of this format requirement. Is this to enable association with traces? Is this documented somewhere?

joaoandremartins:
@saturnism where do you want to have access to the configured project ID?

Spring property could be a good option, but you should take into account that each service has a project ID override, e.g., for Pub/Sub, you'd need to check first `spring.cloud.gcp.pubsub.project-id`, then `spring.cloud.gcp.project-id`.

meltsufin:
I kind of like the `springProperty` via the `Action`. Maybe we can give it try?

saturnism:
I took a look at this; it's a lot more involved. I.e., the `SpringPropertyAction` is created by [LogbackLoggingSystem](https://github.com/spring-projects/spring-boot/blob/v2.0.1.RELEASE/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/logging/logback/LogbackLoggingSystem.java), which is created by [LoggingSystem](https://github.com/spring-projects/spring-boot/blob/236b1bc4b634e06657f35829eb86ac996b36fed3/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/logging/LoggingSystem.java) static initializer... We can't really hook into it.

saturnism:
Also, the trace ID format is documented here: https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry

saturnism:
I think a compromise might be using `<springProperty/>` to try to grab `spring.cloud.gcp.project-id`. And if that's not set, override it w/ `DefaultGcpProjectIdProvider`.

PR #626, opened 2018-04-21
Make 64-bit Trace ID 128-bit, and also determine default Project ID

saturnism:
For #625

### DIFF
--- a/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/logging/StackdriverJsonLayout.java
+++ b/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/logging/StackdriverJsonLayout.java
@@ -29,6 +29,8 @@ import ch.qos.logback.contrib.json.classic.JsonLayout;
 import com.google.cloud.logging.TraceLoggingEnhancer;
 import com.google.gson.Gson;
 
+import org.springframework.cloud.gcp.core.DefaultGcpProjectIdProvider;
+import org.springframework.cloud.gcp.core.GcpProjectIdProvider;
 import org.springframework.util.StringUtils;
 
 /**
@@ -161,6 +163,17 @@ public class StackdriverJsonLayout extends JsonLayout {
 	 */
 	public void setIncludeExceptionInMessage(boolean includeExceptionInMessage) {
 		this.includeExceptionInMessage = includeExceptionInMessage;
+	}
+
+	@Override
+	public void start() {
+		super.start();
+
+		// If no Project ID set, then attempt to resolve it with the default project ID provider
+		if (StringUtils.isEmpty(this.projectId) || this.projectId.endsWith("_IS_UNDEFINED")) {
+			GcpProjectIdProvider projectIdProvider = new DefaultGcpProjectIdProvider();
+			this.projectId = projectIdProvider.getProjectId();
+		}
 	}
 
 	/**
--- a/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/logging/StackdriverJsonLayout.java
+++ b/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/logging/StackdriverJsonLayout.java
@@ -211,6 +211,15 @@ public class StackdriverJsonLayout extends JsonLayout {
 		return map;
 	}
 
+	protected String formatTraceId(final String traceId) {
+		// Trace IDs are either 64-bit or 128-bit, which is 16-digit hex, or 32-digit hex.
+		// If traceId is 64-bit (16-digit hex), then we need to prepend 0's to make a 32-digit hex.
+		if (traceId != null && traceId.length() == 16) {
+			return "0000000000000000" + traceId;
+		}
+		return traceId;
+	}
+
 	private void addTraceId(ILoggingEvent event, Map<String, Object> map) {
 		if (!this.includeTraceId) {
 			return;
@@ -223,7 +232,7 @@ public class StackdriverJsonLayout extends JsonLayout {
 		if (!StringUtils.isEmpty(traceId)
 			&& !StringUtils.isEmpty(this.projectId)
 			&& !this.projectId.endsWith("_IS_UNDEFINED")) {
-			traceId = "projects/" + this.projectId + "/traces/" + traceId;
+			traceId = "projects/" + this.projectId + "/traces/" + formatTraceId(traceId);
 		}
 
 		add(TRACE_ID_ATTRIBUTE, this.includeTraceId, traceId, map);
--- a/spring-cloud-gcp-autoconfigure/src/main/resources/org/springframework/cloud/gcp/autoconfigure/logging/logback-json-appender.xml
+++ b/spring-cloud-gcp-autoconfigure/src/main/resources/org/springframework/cloud/gcp/autoconfigure/logging/logback-json-appender.xml
@@ -3,7 +3,7 @@
 Stackdriver json-log format provided for import.
 -->
 <included>
-    <property name="projectId" value="${projectId:-${GOOGLE_CLOUD_PROJECT}}"/>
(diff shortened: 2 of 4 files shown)


spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/logging/StackdriverJsonLayout.java
```diff
@@ -163,6 +165,17 @@ public void setIncludeExceptionInMessage(boolean includeExceptionInMessage) {
 		this.includeExceptionInMessage = includeExceptionInMessage;
 	}
 
+	@Override
+	public void start() {
+		super.start();
+
+		// If no Project ID set, then attempt to resolve it with the default project ID provider
+		if (StringUtils.isEmpty(this.projectId) || this.projectId.endsWith("_IS_UNDEFINED")) {
```
joaoandremartins: Why do we do this?

If there is no configured project ID, should't we ask the user to provide one before jumping to auto-configuration?

spring-cloud-gcp-autoconfigure/src/main/resources/org/springframework/cloud/gcp/autoconfigure/logging/logback-json-appender.xml
```diff
@@ -3,7 +3,7 @@
 Stackdriver json-log format provided for import.
 -->
 <included>
-    <property name="projectId" value="${projectId:-${GOOGLE_CLOUD_PROJECT}}"/>
+    <property name="projectId" value="${projectId:-${SPRING_CLOUD_GCP_LOGGING_PROJECT_ID}}"/>
```
joaoandremartins: IIUC this aims at grabbing the value of the `spring.cloud.gcp.logging.project-id` property?
What if an application configured the project ID as `spring.cloud.gcp.project-id`?

spring-cloud-gcp-docs/src/main/asciidoc/logging.adoc
```diff
@@ -114,7 +114,16 @@ The following properties are available:
 [options="header",]
 |=======================================================================
 | Property | Default Value | Description
-| `projectId` |  | Only required for logging the correct traceId: `projects/[PROJECT-ID]/traces/[TRACE-ID]`. If `projectId` is not set, the logged `traceId` will match the one provided by `spring-cloud-gpc-trace`
+| `projectId`
+a| If not set, default value is determined in the following order:
+
+1. `SPRING_CLOUD_GCP_LOGGING_PROJECT_ID` Environmental Variable.
```
joaoandremartins: I'm not really sure about this since, as I mentioned earlier, if a user wants auto-configuration, why don't they use it explicitly?

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/logging/StackdriverJsonLayout.java
```diff
@@ -163,6 +165,17 @@ public void setIncludeExceptionInMessage(boolean includeExceptionInMessage) {
...
+			GcpProjectIdProvider projectIdProvider = new DefaultGcpProjectIdProvider();
```
meltsufin: I'm not sure about using the `DefaultGcpProjectIdProvider` here. If we want to be consistent with the rest of the project, we should add `projectId` property to `StackdriverLoggingProperties`.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/logging/StackdriverJsonLayout.java
saturnism: this is pretty similar to autoconfigurations. setting project ID per deployment/project is quite a hassle. we don't require it for the rest of spring cloud gcp.  i feel it should at least attempt to resolve it via the standard way.  If they need to use spring.cloud.gcp.project-id, they'll need to use `<springProperty/>`

spring-cloud-gcp-docs/src/main/asciidoc/logging.adoc
saturnism: The ideal situation is to consume Spring configurations from Layout. But it seems that the Layout is instantiated out of band, and has no access to the context.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/logging/StackdriverJsonLayout.java
saturnism: I wish there is a way to get that info from the Layout. I haven't found a way though.

spring-cloud-gcp-autoconfigure/src/main/resources/org/springframework/cloud/gcp/autoconfigure/logging/logback-json-appender.xml
saturnism: That won't be used. They'll need to manually add <springProperty> to reference to a project ID.

GOOGLE_CLOUD_PROJECT is actually not explicit enough since it could override a lot of things.  It's actually possible to send a multi-project trace in Stackdriver. I.e., traces from multiple projects show up in the same Trace view, but the log links to the originating project.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/logging/StackdriverJsonLayout.java
meltsufin: You're right. The problem is that it lives outside of the Spring Context and initialized by Logback.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/logging/StackdriverJsonLayout.java
joaoandremartins: Auto-configuration already supports credential discovery, like you're doing it here.

It's the last resort, and it's done through `DefaultCredentialsProvider`.

Which is why I think we should avoid this code duplication.

spring-cloud-gcp-docs/src/main/asciidoc/logging.adoc
joaoandremartins: I don't know what the layout is, but I'm assuming you're referring to something static/compile time?

But weren't we auto-configuring these before? Was the auto-configuration not being used then?

saturnism: The configuration is in the Logback Layout, which is initialized outside of Spring. I.e., it's initialized by Logback:
https://github.com/saturnism/spring-cloud-gcp/blob/ee0b421da3aaa9468286fca54b9e6fd01b5301a7/spring-cloud-gcp-autoconfigure/src/main/resources/org/springframework/cloud/gcp/autoconfigure/logging/logback-json-appender.xml#L10

Thus it has no access to the Spring context.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/logging/StackdriverJsonLayout.java
saturnism: This is because we have no access to the Spring context from the Layout :(

spring-cloud-gcp-docs/src/main/asciidoc/logging.adoc
joaoandremartins: It turns out we weren't even using the project ID in auto-configuration before.
Should we omit the `If not set, default value is ` bit since, even if users specify a project ID through a property, it won't be used anywhere?

PR #631, opened 2018-04-23
Spanner separate read from convert

balopat:
Related to #593, this refactoring achieves cleaner separation of concerns regarding reading values from structs and only then handing over values to the conversion logic, triggered by [this comment](https://github.com/spring-cloud/spring-cloud-gcp/pull/619#discussion_r182968838) from @mp911de on #619.


spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/MappingSpannerReadConverter.java
```diff
@@ -47,46 +46,69 @@
 class MappingSpannerReadConverter extends AbstractSpannerCustomConverter
 		implements SpannerEntityReader {
 
+	// @formatter:off
```
ChengyuanZhao: I totally didn't know that was a thing! i had been manually pressing return on this line the whole time after the formatter messes it up.

balopat: I know - I got soooo annoyed with it too :)

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/ConversionUtils.java
```diff
@@ -17,234 +17,34 @@
...
-
-	/**
-	 * Converters from common types to those used by Spanner.
-	 */
-	public static final Collection<Converter> DEFAULT_SPANNER_READ_CONVERTERS = ImmutableSet
-			.of(SPANNER_TO_JAVA_DATE_CONVERTER, TIMESTAMP_INSTANT_CONVERTER, SPANNER_TO_JAVA_BYTE_ARRAY_CONVERTER,
-					SPANNER_TO_JAVA_TIMESTAMP_CONVERTER, SPANNER_TO_JAVA_SQL_DATE_CONVERTER);
-
-	public static Class boxIfNeeded(Class propertyType) {
+	public static <T> Class<T> boxIfNeeded(Class<T> propertyType) {
```
ChengyuanZhao: I don't think the return type and param type here should both be `Class<T>`. That's saying the original type and the boxed type should be exactly the same  type. That seems to go against the purpose of this class because we are supposed to be giving back a different type than what is received. I think this only works because any type that does need boxing can be safely cast as its boxed type like `Integer` can be cast as `int`. I don't doubt that casting like this can always be safe, but I think it just makes the definition of this function strange and self-contradicting .

I see you need this type parameter here on this line in AbstractSpannerCustomConverter:
`Class<T> boxedTargetType = ConversionUtils.boxIfNeeded(targetType);`. 
I think you should just be casting to `Class<T>` on this line instead of in `boxIfNeeded`.

ChengyuanZhao: @balopat otherwise it looks good to me!

PR #633, opened 2018-04-24
Logging integration tests

joaoandremartins:
Test the header extractor and ensuring that the log was successfully
stored in a GCP project.
Also refactors logging module a little, by
1) Moving extractors to a new package,
2) Merging TraceIdExtractorType with TraceIdExtractor


spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/logging/extractors/TraceIdExtractor.java
```diff
@@ -32,4 +32,33 @@
...
+
+	/**
+	 * Enum values represent the order and combination of predefined trace ID extractors that
+	 * can be used.
+	 *
+	 * @see XCloudTraceIdExtractor
+	 * @see ZipkinTraceIdExtractor
+	 * @see CompositeTraceIdExtractor
+	 */
+	enum Type {
```
artembilan: The things doesn't work this way. 
You are mixing here a concern of the `TraceIdExtractor` abstraction and possible out-of-the-box types.
The author of any possible custom implementation would be confused. 

If you would like to have it as an inner class, this `Type` should go into the `StackdriverLoggingProperties`.
Outside of the auto-configuration the `enum` does not make sense: you are not going to instantiate the target instance based on the `enum` manually.

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/logging/it/StackdriverLoggingIntegrationTests.java
```diff
@@ -0,0 +1,121 @@
...
+
+	@Test
+	public void test() throws InterruptedException, IOException {
+		URL url = new URL("http://localhost:" + this.port);
+		HttpURLConnection connection = (HttpURLConnection) url.openConnection();
+		connection.setRequestProperty("x-cloud-trace-context", "everything-zen");
+		connection.getResponseMessage();
+		connection.disconnect();
+		// Allows the client library some time to flush the logs.
+		Thread.sleep(2000);
```
artembilan: That's not good even for integration test. 
Consider looping for the log entry instead with some small sleep in between and spawn the whole waiting window up to 20 second:
```
int n = 0;
while (n++ < 200 && [CHECK_FOR_LOG_ENTRY]) {
	Thread.sleep(100);
}
// asserts
```

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/logging/it/StackdriverLoggingIntegrationTests.java
```diff
@@ -0,0 +1,121 @@
...
+
+	@BeforeClass
+	public static void enableTests() {
+		assumeThat("it.logging").isEqualTo("true");
+	}
+
+	@Test
+	public void test() throws InterruptedException, IOException {
+		URL url = new URL("http://localhost:" + this.port);
+		HttpURLConnection connection = (HttpURLConnection) url.openConnection();
```
artembilan: Consider to use `TestRestTempalte` instead. And seems for me it is supplied with the proper `port` already.

joaoandremartins: I looked into `TestRestTemplate`, but it doesn't seem to support adding headers to the HTTP layer...

artembilan:
@joaoandremartins ,

Does it say anything to you?
```
Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 8.196 sec <<< FAILURE!
test(org.springframework.cloud.gcp.autoconfigure.logging.it.StackdriverLoggingIntegrationTests)  Time elapsed: 0.956 sec  <<< ERROR!
com.google.cloud.logging.LoggingException: io.grpc.StatusRuntimeException: PERMISSION_DENIED: Stackdriver Logging API has not been used in project 653574825147 before or it is disabled. Enable it by visiting https://console.developers.google.com/apis/api/logging.googleapis.com/overview?project=653574825147 then retry. If you enabled this API recently, wait a few minutes for the action to propagate to our systems and retry.
```

See Travis report.

meltsufin:
@artembilan We just needed to enable the Logging API. I just did and restarted the build.

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/logging/it/StackdriverLoggingIntegrationTests.java
```diff
@@ -0,0 +1,124 @@
...
+		do {
+			Thread.sleep(100);
+			page = logClient.listLogEntries(
+					Logging.EntryListOption.filter("textPayload:\"#$%^&" + NOW + "\" AND"
+							+ " logName=\"projects/" + projectIdProvider.getProjectId()
+							+ "/logs/spring.log\""));
+			for (LogEntry entry : page.getValues()) {
+				pageSize++;
+			}
+		} while (pageSize == 0 && counter++ < 2);
```
artembilan: Don't you mean here `counter++ < 20`?
Otherwise you are going to sleep `100 ms` only twice.

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/logging/it/StackdriverLoggingIntegrationTests.java
```diff
@@ -0,0 +1,124 @@
...
+
+	@RestController
+	@SpringBootApplication(exclude = {
+			GcpCloudSqlAutoConfiguration.class,
+			GcpStorageAutoConfiguration.class,
+			DataSourceAutoConfiguration.class
+	})
+	static class LoggingApplication {
+
+		public static void main(String[] args) {
```
artembilan: I don't think that this code brings some value here.

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/logging/it/StackdriverLoggingIntegrationTests.java
artembilan: Yeah, I also always have a hard time to determine how to perform simple GET via `RestTemplate` with headers.
You can do like this:
```
HttpHeaders headers = new HttpHeaders();
headers.set("x-cloud-trace-context", "everything-zen");
HttpEntity entity = new HttpEntity(headers);

ResponseEntity<String> response = testRestTemplate.exchange("/", HttpMethod.GET, entity, String.class);
```
But well: your code is so darn short that I won't insist anymore for changing it to `RestTemplate` attempt! :smiley:

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/logging/it/StackdriverLoggingIntegrationTests.java
```diff
@@ -0,0 +1,124 @@
...
+			SpringApplication.run(LoggingApplication.class, args);
+		}
+
+		@GetMapping
+		public void log() {
```
artembilan: Well, since it's `void` it sounds more like `HEAD`, but not `GET`: some inconsistency in REST protocol usage: https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html :smile:

PR #636, opened 2018-04-25
Sets project ID so that logging appender won't complain

joaoandremartins:


spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/storage/GcpStorageAutoConfigurationTests.java
```diff
@@ -59,6 +59,11 @@
 		})
 public class GcpStorageAutoConfigurationTests {
 
+	// Sets project ID so logging appender doesn't complain.
+	static {
+		System.setProperty("GOOGLE_CLOUD_PROJECT", "test-project");
```
artembilan: OK! How does it work on Travis and locally for you then?
Thanks

joaoandremartins: It works on Travis because gcloud sets the project ID for the integration tests, which doesn't happen in Jenkins.
Locally, it works if gcloud has the project ID set, but doesn't if it is unset.

Since we can't conditionally add `logback-appender.xml` for one test only, this was the only way I could think of resolving this.

artembilan: OK! And why then you don't reflect that in the commit message?

However I still think that this is not enough. Since you say we can't add that config conditionally, we have to figure out something different, e.g. create such a config on the fly in the particular test.

Nobody says that we are not going to fail for the same reason in other tests.

BTW, why does only this one fail, but not others? The `logback-appender.xml` is still there for them as well...

Too many unclear situations...
That makes me useless in helping you...

Thanks for understanding

joaoandremartins: I don't think it's possible to create the configuration on the fly, let me know if you know a way.
I also looked into specifying a config name to Logback so that different tests used different configs, but that doesn't seem to be possible either.

I suspect these tests failing has something to do with them being the only ones that use `@SpringBootTest`.

artembilan: One more time: that's not right.
Any other new `@SpringBootTest` is going to suffer the same problem. 
And that's clearly not its responsibility to worry about configuration not related stuff.
And from here it even sounds for me like all other tests are not configured for logging.

In other project we use Log4J2 as a logging background and trigger `org.slf4j.LoggerFactory` whenever we need to poll `logback-test.xml`. 

On the other hand I see that there is a property like:
```
<property name="projectId" value="test-project"/>
```

Why doesn't it work?
Does that `logback-json-appender.xml` include override this local property?

joaoandremartins: That property isn't an input anywhere to `LoggingAppender`. It gets project ID from `[LoggingOptions.getDefaultInstance().getProjectId()](https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/google-cloud-contrib/google-cloud-logging-logback/src/main/java/com/google/cloud/logging/logback/LoggingAppender.java#L195)`.

`logback-json-appender.xml` uses a different appender, so the same doesn't apply there.

I don't expect this to be a very big deal, since the best practice is to use `ApplicationContextRunner` from now on, as much as possible, and they don't seem to automatically read `logback-test.xml`. I also don't think we'll be adding those many more tests, let alone `@SpringBootTests`.

You say it's not right, but I haven't heard of a valid alternative. If there isn't one, then I'm afraid we'll either need to extract the logging module to its own, just for tests, or only have one `@SpringBootTest`, which I don't think makes a lot of sense.

meltsufin: How about we just have two different logging configurations and use maven profiles to control which one is included. See https://maven.apache.org/plugins/maven-resources-plugin/examples/filter.html

artembilan: Well, I have just debugged one of the non-`@SpringBootTests` test and looks like the same error happens there as well on the line:
```
cai.addError("RuntimeException in Action for tag [" + tagName + "]", e);
```
But that doesn't pop there I think, because there really might be some injection from Spring Boot to Logback context to fail fast.

So, I really would insist to go some other way for testing Stackdriver logging, rather than pollute all our tests.

artembilan: Right, according stack trace there is something from Spring Boot involved:
```
ERROR in ch.qos.logback.core.joran.spi.Interpreter@15:14 - RuntimeException in Action for tag [appender] java.lang.IllegalArgumentException: A project ID is required for this service but could not be determined from the builder or the environment.  Please set a project ID using the builder.
	at org.springframework.boot.logging.logback.LogbackLoggingSystem.loadConfiguration(LogbackLoggingSystem.java:166)
	at org.springframework.boot.logging.logback.LogbackLoggingSystem.reinitialize(LogbackLoggingSystem.java:212)
	at org.springframework.boot.logging.AbstractLoggingSystem.initializeWithConventions(AbstractLoggingSystem.java:75)
	at org.springframework.boot.logging.AbstractLoggingSystem.initialize(AbstractLoggingSystem.java:60)
	at org.springframework.boot.logging.logback.LogbackLoggingSystem.initialize(LogbackLoggingSystem.java:114)
	at org.springframework.boot.context.logging.LoggingApplicationListener.initializeSystem(LoggingApplicationListener.java:264)
	at org.springframework.boot.context.logging.LoggingApplicationListener.initialize(LoggingApplicationListener.java:237)
	at org.springframework.boot.context.logging.LoggingApplicationListener.onApplicationEnvironmentPreparedEvent(LoggingApplicationListener.java:200)
```

Looks like these:
```
org.springframework.boot.context.logging.ClasspathLoggingApplicationListener,\
org.springframework.boot.context.logging.LoggingApplicationListener,\
```
Are not involved when `ApplicationContextRunner` is used instead of fully blown `@SpringBootTest`

@snicoll , would you mind to clarify for sake of curiosity and complete picture on the matter?

(I still think that we shouldn't run Stackdriver if we are not interested in it...)

joaoandremartins:
PTAL I believe this should do as a better approach.

spring-cloud-gcp-autoconfigure/src/test/resources/logback-test.xml
```diff
@@ -23,6 +23,8 @@
 		<appender-ref ref="STDOUT" />
 	</logger>
 
+	<property name="GOOGLE_CLOUD_PROJECT" value="test-project" />
```
artembilan: I've just copied/pasted this code into my local (fresh) copy of the project and nothing changed when I run `GcpStorageAutoConfigurationTests`:
```
Caused by: java.lang.IllegalStateException: Logback configuration error detected: 
ERROR in ch.qos.logback.core.joran.spi.Interpreter@15:14 - RuntimeException in Action for tag [appender] java.lang.IllegalArgumentException: A project ID is required for this service but could not be determined from the builder or the environment.  Please set a project ID using the builder.
	at org.springframework.boot.logging.logback.LogbackLoggingSystem.loadConfiguration(LogbackLoggingSystem.java:166)
```

Am I missing anything else?

Thanks

.travis.yml
```diff
@@ -11,14 +11,13 @@ cache:
     - $HOME/google-cloud-sdk
     - $HOME/.m2
 script:
-  - ./mvnw -T 1.5C test -P it -B ${INTEGRATION_TEST_FLAGS}
+  - ./mvnw -T 1.5C test -B ${INTEGRATION_TEST_FLAGS}
 install:
-  - ./mvnw -T 1.5C install -DskipTests=true -Dmaven.javadoc.skip=true -B -V
+  - ./mvnw -T 1.5C install -P it -DskipTests=true -Dmaven.javadoc.skip=true -B -V
```
artembilan: But we have here `-DskipTests=true`. Doesn't that mean that IT tests won't go as well?

joaoandremartins: I suspect that the tests aren't run, but it is here that the test structure is built.
i.e., it is here that what we're doing in https://github.com/spring-cloud/spring-cloud-gcp/pull/636/files#diff-2d9cc6b52cccb7c9ee54209384519466 kicks in

At least, I tried to set it in the test and it didn't work. Let's see if this works...

joaoandremartins: Oof. I think I finally got it :)
Now Jenkins should also work because we're not using the `it` profile there (`unit` is used by default), and the `STACKDRIVER` appender is never included.

artembilan: Does it mean it's ready for the final review?
Thanks

PR #639, opened 2018-04-26
Updates to Finchley.RC1 and removes Guava from internal BOM

### LINKED ISSUE #582: Inconsistent version of Guava

joaoandremartins:
Fixes #582

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -39,7 +39,7 @@
 	<properties>
 		<maven.compiler.source>1.8</maven.compiler.source>
 		<maven.compiler.target>1.8</maven.compiler.target>
-		<spring-cloud-release.version>Finchley.M9</spring-cloud-release.version>
+		<spring-cloud-release.version>Finchley.RC1</spring-cloud-release.version>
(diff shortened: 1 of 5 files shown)


joaoandremartins:
I don't think GitHub will truncate this commit line.
Should that be enough?

PR #640, opened 2018-04-26
Removes redundant label extractor from auto-configuration

joaoandremartins:


meltsufin:
Why is it redundant now? What has changed?

joaoandremartins:
It's already being declared in https://github.com/spring-cloud/spring-cloud-gcp/blob/master/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/trace/StackdriverTraceAutoConfiguration.java#L132

artembilan:
Good. So, I insist to include such an information into the commit message.

(I can reject my request if there are some strong arguments against)

Thanks

PR #643, opened 2018-04-30
Spanner custom conversions for Struct

### LINKED ISSUE #634: Add support for Structs as fields and/or CustomConversion for Structs
See for context: [this comment](https://github.com/spring-cloud/spring-cloud-gcp/pull/619/files#r183794813)

The idea is that if the user wants to create their own `Struct` serialization, we can relatively easily expose that via the CustomConverters API. This would entail changing that above mentioned if statement and writing some tests.

ChengyuanZhao:
fixes #634

### DIFF
--- a/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/StructAccessor.java
+++ b/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/StructAccessor.java
@@ -35,6 +35,7 @@ import com.google.common.collect.ImmutableMap;
  * knowing their type.
  *
  * @author Balint Pato
+ * @author Chengyuan Zhao
  */
 class StructAccessor {
 
@@ -79,17 +80,20 @@ class StructAccessor {
 		this.columnNamesIndex = indexColumnNames();
 	}
 
-	public Object getSingleValue(String colName) {
+	Object getSingleValue(String colName) {
 		Type colType = this.struct.getColumnType(colName);
 		Type.Code code = colType.getCode();
 		Class sourceType = code.equals(Type.Code.ARRAY)
 				? this.spannerTypeMapper.getArrayJavaClassFor(colType.getArrayElementType().getCode())
 				: this.spannerTypeMapper.getSimpleJavaClassFor(code);
 		BiFunction readFunction = singleItemReadMethodMapping.get(sourceType);
+		if (readFunction == null) {
+			return null;
+		}
 		return readFunction.apply(this.struct, colName);
 	}
 
-	public List getListValue(String colName) {
+	List getListValue(String colName) {
 		Type.Code innerTypeCode = this.struct.getColumnType(colName).getArrayElementType().getCode();
 		Class clazz = this.spannerTypeMapper.getSimpleJavaClassFor(innerTypeCode);
 		BiFunction<Struct, String, List> readMethod = readIterableMapping.get(clazz);
--- a/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/StructAccessor.java
+++ b/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/StructAccessor.java
@@ -28,7 +28,10 @@ import com.google.cloud.Timestamp;
 import com.google.cloud.spanner.AbstractStructReader;
 import com.google.cloud.spanner.Struct;
 import com.google.cloud.spanner.Type;
+import com.google.cloud.spanner.Type.Code;
 import com.google.common.collect.ImmutableMap;
+
+import org.springframework.cloud.gcp.data.spanner.core.mapping.SpannerDataException;
 
 /**
  * A convenience wrapper class around Struct to make reading columns easier without
@@ -94,6 +97,9 @@ class StructAccessor {
 	}
 
 	List getListValue(String colName) {
+		if (this.struct.getColumnType(colName).getCode() != Code.ARRAY) {
+			throw new SpannerDataException("Column is not an ARRAY type: " + colName);
+		}
 		Type.Code innerTypeCode = this.struct.getColumnType(colName).getArrayElementType().getCode();
 		Class clazz = this.spannerTypeMapper.getSimpleJavaClassFor(innerTypeCode);
 		BiFunction<Struct, String, List> readMethod = readIterableMapping.get(clazz);
--- a/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/StructAccessor.java
+++ b/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/StructAccessor.java
@@ -91,6 +91,8 @@ class StructAccessor {
 				: this.spannerTypeMapper.getSimpleJavaClassFor(code);
 		BiFunction readFunction = singleItemReadMethodMapping.get(sourceType);
 		if (readFunction == null) {
+			// This case should only occur if the POJO field is non-Iterable, but the column type
+			// is ARRAY of STRUCT, TIMESTAMP, DATE, BYTES, or STRING. This use-case is not supported.
 			return null;
 		}
 		return readFunction.apply(this.struct, colName);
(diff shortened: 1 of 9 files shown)


spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/StructAccessor.java
```diff
@@ -79,17 +80,20 @@
...
 
-	public Object getSingleValue(String colName) {
+	Object getSingleValue(String colName) {
 		Type colType = this.struct.getColumnType(colName);
 		Type.Code code = colType.getCode();
 		Class sourceType = code.equals(Type.Code.ARRAY)
 				? this.spannerTypeMapper.getArrayJavaClassFor(colType.getArrayElementType().getCode())
 				: this.spannerTypeMapper.getSimpleJavaClassFor(code);
 		BiFunction readFunction = singleItemReadMethodMapping.get(sourceType);
+		if (readFunction == null) {
```
balopat: when can this realistically happen? (i.e. not in _test_ code)

spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/core/convert/ConverterAwareMappingSpannerEntityReaderTest.java
```diff
@@ -70,12 +75,37 @@ public void readNestedStructTest() {
...
+		OuterTestEntity result = this.spannerEntityReader.read(OuterTestEntity.class,
+				outerStruct);
 		assertEquals("key1", result.id);
 		assertEquals(1, result.innerTestEntities.size());
 		assertEquals("value", result.innerTestEntities.get(0).value);
 	}
 
+	@Test
+	public void readConvertedNestedStructTest() {
+		Struct innerStruct = Struct.newBuilder().add("value", Value.string("value"))
```
balopat: nit suggestions: 
* rename "value" -> "string_col" 
* rename innerStruct to structColumn
* rename outerStruct to rowStruct

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/StructAccessor.java
ChengyuanZhao: I'm not sure, but I think it's good to return null here. In the reader we throw the exception if at the end the read value was null.

spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/core/convert/ConverterAwareMappingSpannerEntityReaderTest.java
ChengyuanZhao: done

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/StructAccessor.java
ChengyuanZhao: I got this one to come up by not having a needed custom converter supplied. The readfunction in that was null, and that caused a nullpointerexception.

balopat: This doesn't seem to have to do much with converters. The `singleItemReadMethodMapping` keyset should cover the column types...

ChengyuanZhao: I lost my original example, but I remember it came up when the POJO had a singular property (like int) and the corresponding column was ARRAY<STRUCT>. This causes `sourceType` to be `Struct[].class`, which isn't in `singleItemReadMethodMapping`, so readFunction is null. 

If we return null then we get the exception later saying "int couldn't be read from Struct[ ] " which makes sense to me.  It looks like this would happen if the column was some other types like "ARRAY<BYTES>" and the POJO property was singular, too. 

Alternatively we could explicitly check for singular-property vs. array-column situation, but it seems like  we naturally detect these situations based on the way we have the type/function maps set up because we will just fail to find a usable function (which  feels pretty elegant).

meltsufin:
@ChengyuanZhao Please include doc updates with the PR. Thanks!

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/StructAccessor.java
balopat: 1. we need to have an example in unit test form - otherwise someone might just remove this piece of code and wouldn't know that a valid case was broken
2. StructAccessor accesses a Struct. This method gets a colName. There are no converters in the picture. This must mean that there is a column type that we are not capturing in the `singleItemReadMethodMapping`. And I guess that could be the `Struct[]` or `Struct`? The question is how do we want to handle on this abstraction level these missing column types.

ChengyuanZhao: I "misspoke" about the converter. Please ignore that brief mention from earlier. The case I wanted to illustrate is the longer last comment I had.

spring-cloud-gcp-docs/src/main/asciidoc/spanner.adoc
```diff
@@ -370,6 +370,27 @@ public class ConverterConfiguration {
 }
 ----
 
+==== Structs in Columns
```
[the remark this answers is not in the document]
ChengyuanZhao: This section best follows the custom converter section, because its behavior is very heavily dependent on knowing about customer converters. The behavior of reading of STRUCTs is entirely dependent on the inclusion of a custom converter. In fact I considered not having a header at all. 

I also think the very small example is better than extending the running example of Trade and Trader. That is a great example to show the other features of this project that require more context, but this specific feature can be demonstrated in isolation. The reason is there is a lot of irrelevant code and context in the example you included. It took some effort to focus on what feature is specifically being demonstrated. I think the smaller example presents the relevant features in a easier-to-digest format.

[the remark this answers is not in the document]
meltsufin: Why is it heavily dependent on custom converters? I just tried the feature out without having to define any converters.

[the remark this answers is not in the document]
ChengyuanZhao: The goal of this PR is to make this line true: 
"If a `Converter<Struct, A>` is provided, then properties of type `List<A>` can be used."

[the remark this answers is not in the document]
meltsufin: I guess there are two things going on here. Support for structs, and customer converters for structs. Perhaps, the documentation should be separated into two sections then: "Arrays of Structs" and "Customer converters for Structs". Does it make sense?

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/StructAccessor.java
balopat: can you just simply please write a realistic testcase that exercises this path? Then we can figure out the right design for it.

ChengyuanZhao: I actually did in a commit earlier today.

take a look at 02fdef6

spring-cloud-gcp-docs/src/main/asciidoc/spanner.adoc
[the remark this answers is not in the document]
ChengyuanZhao: Yes that sounds good. I'll move things around.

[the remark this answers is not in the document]
ChengyuanZhao: done.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/StructAccessor.java
ChengyuanZhao: Wow that's a really subtle find! I would never have expected the default conversion service we're using can convert from single-item lists to a single item!

I think it's a good idea to add `.put(Struct[].class, (struct, col) -> null)` as a safe-guard. This situation is looks the same for ByteArray[].class/getByteArrayList, Timestamp[].class/getTimestampList, etc. , and we ought to add the null-returning anon function for those too, right?

balopat: As discussed, just adding a comment will suffice. Thanks.

PR #645, opened 2018-04-30
Reduces log verbosity

joaoandremartins:
Test cases that start Spring Boot apps will no longer log the
boilerplate Spring Boot app boot logs. Warnings and errors will still be
logged.
After this change, we will still need to clear the Spanner compilation
and Javadoc warnings.


spring-cloud-gcp-data-spanner/src/test/resources/logback-test.xml
```diff
@@ -0,0 +1,4 @@
+<configuration>
+
+  <root level="warn"/>
```
artembilan: Does it STDOUT by default?

And since we are here already, I would like to share our experience: we add entries for root packages of the module we test against, something like:
```
<logger name="org.springframework.cloud.gcp.data.spanner" level="warn"/>
```
This way it is going to be pretty easy to change it to `debug` if something wrong happens.

Just an idea. Feel free to reject.

spring-cloud-gcp-data-spanner/src/test/resources/logback.xml
```diff
@@ -1,19 +0,0 @@
...
-    </encoder>
-  </appender>
-
-  <logger name="io.grpc" level="info"/>
-  <logger name="io.netty" level="info"/>
-  <logger name="org.springframework" level="info"/>
-  <logger name="org.springframework.cloud" level="debug"/>
-
-  <root level="debug">
-    <appender-ref ref="STDOUT" />
```
meltsufin: You need to keep this. Logback will not output to `STDOUT` by default.

spring-cloud-gcp-data-spanner/src/test/resources/logback-test.xml
```diff
@@ -1,4 +1,14 @@
...
 <configuration>
+	<appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
+		<!-- encoders are assigned the type
+		ch.qos.logback.classic.encoder.PatternLayoutEncoder by default -->
+		<encoder>
+			<pattern>%d{HH:mm:ss.SSS} [%thread] %-5level %logger{36} - %msg%n</pattern>
+		</encoder>
+	</appender>
 
-  <root level="warn"/>
```
artembilan: I think `root` still has to be there:
```
<root level="warn">
	<appender-ref ref="STDOUT"/>
</root>
```

No? Does it work properly even without that?
I mean do we meet the general purpose of this PR if we don't have `root`?

Thanks

artembilan:
Travis has failed with some tests, but looks like your fix did the trick for verbosity.
I think we also can turn off Spring Boot banner, if that is a noise as well:
```
spring.main.banner-mode = off
```
https://docs.spring.io/spring-boot/docs/2.0.1.RELEASE/reference/htmlsingle/#boot-features-banner

joaoandremartins:
Will add that line.

artembilan:
Why logs like this:
```
class org.springframework.cloud.gcp.data.spanner.core.it.SpannerTemplateIntegrationTests - schema created: [CREATE TABLE trades_1525191361570 ( id STRING(MAX) , age INT64 , tradeTime TIMESTAMP , tradeDate DATE , action STRING(MAX) , price FLOAT64 , shares FLOAT64 , symbol STRING(MAX) , trader_id STRING(MAX) , executionTimes ARRAY<TIMESTAMP> ) PRIMARY KEY ( id , trader_id )]
Integration database cleaned up!
```
Are not covered by this fix?
Are they just `System.out.println()` instead of logging?

joaoandremartins:
That is indeed a good question for @ChengyuanZhao to take a look into, while he's eliminating the logging noise on the Spanner side.

ChengyuanZhao:
@joaoandremartins That line indeed is coming from System.out.println() . it is in `AbstractSpannerIntegrationTest`. It's not an important line actually. I can take care of it in the other PR about spanner warnings.

PR #648, opened 2018-05-01
Uses Zipkin's Stackdriver sender

### LINKED ISSUE #485: Use Stackdriver Trace Zipkin Sender
Update Spring Cloud GCP Trace to use Stackdriver Zipkin Sender.

joaoandremartins:
The HttpClientParser and HttpServerParser portions of the
auto-configuration were left intact, since they currently enrich the
traces in a way that Zipkin/Brave can't (as far as I'm aware of).

This change depends on a new version of zipkin-translation-stackdriver
due to a bug.

Fixes #485


joaoandremartins:
@marcingrzejszczak @adriancole FYI, this change depends on https://github.com/openzipkin/zipkin-gcp/pull/78 (and `zipkin-gcp`'s subsequent version).

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/trace/StackdriverTraceAutoConfiguration.java
```diff
@@ -112,25 +96,24 @@ public SleuthProperties stackdriverSleuthProperties(SleuthProperties sleuthPrope
...
 	@Bean
 	@ConditionalOnMissingBean
-	public Reporter<zipkin2.Span> reporter(
-			FlushableTraceConsumer traceConsumer,
-			SpanTranslator spanTranslator) {
-		return new StackdriverTraceReporter(
-				this.finalProjectIdProvider.getProjectId(),
-				traceConsumer,
-				spanTranslator);
+	public Reporter<Span> reporter(Sender sender, BytesEncoder<Span> encoder) {
```
adriancole: probably should just leave this one out, right? Otherwise it will ignore all sleuth's property overrides for the reporter, right?

joaoandremartins: Interesting. I tried to leave it out and the spans weren't being sent.
Maybe I messed up, or maybe I missed any (auto-configuration) dependency?

spring-cloud-gcp-docs/src/main/asciidoc/trace.adoc
```diff
@@ -92,12 +70,9 @@ spring.sleuth.web.skipPattern=(^cleanup.*|.+favicon.*)  # Ignore some URL paths.
 Spring Cloud GCP Trace does override some Sleuth configurations:
 
 - Always uses 128-bit Trace IDs. This is required by Stackdriver Trace.
-- Does not use Span joins. Span joins will share the span ID between the client and server Spans.
-  Stackdriver requires that every Span ID within a Trace to be unique, so Span joins are not supported.
-- Uses `StackdriverHttpClientParser` and `StackdriverHttpServerParser` by default to populate Stackdriver
-  related fields.
+- Does not use Span joins. Span joins will share the span ID between the client and server Spans. Stackdriver requires that every Span ID within a Trace to be unique, so Span joins are not supported.
```
adriancole: maybe make it hard-fail (by auto-wiring the property for sharing span ID and 128-bit ID and checking them when producing the sender bean)

adriancole:
well I tried this a few weeks ago and didn't need to override the
asyncreporter..

https://github.com/openzipkin/sleuth-webmvc-example/blob/use-stackdriver-sender/src/main/java/sleuth/webmvc/StackdriverConfig.java

adriancole:
sorry few days ago

joaoandremartins:
I just tried it again and it doesn't work.
By the way, I only added `zipkin-sender-stackdriver` to my dependencies. Perhaps I should have added another dependency that auto-configures the reporter based on the sender and encoder?

adriancole:
the sample is using spring-cloud-sleuth.. you are using this? also, maybe
run the example itself for sanity check?

joaoandremartins:
Yep, the example works.
Also, we're tying with Sleuth's properties and using `TraceHttpAutoConfiguration` which is in Sleuth land.

spring-cloud-gcp-docs/src/main/asciidoc/trace.adoc
joaoandremartins: We're already forcing 128 bit trace IDs and not supporting trace joins in autoconfiguration.

```
@Bean
@Primary
public SleuthProperties stackdriverSleuthProperties(SleuthProperties sleuthProperties) {
	sleuthProperties.setSupportsJoin(false);
	sleuthProperties.setTraceId128(true);
	return sleuthProperties;
}
```

joaoandremartins:
Version now set to `0.4.1`.

This is now ready for another pass. Please let me know if I missed any earlier requirement. I'd be interested in how we can not have to specify the reporter, but I'm afraid we're missing some auto-configuration...

adriancole:
rename the sender bean to just "sender" and see if you can remove your own
reporter config. the only thing wrong with this change is the redundantly
declared reporter.

I am going on holiday so sorry I can't comment more for a couple weeks.

joaoandremartins:
@adriancole I made that test and it doesn't work.

I suggest we merge it this way and get back to this later, then.

@meltsufin can you please review?

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/trace/StackdriverTraceAutoConfiguration.java
```diff
@@ -112,25 +96,24 @@ public SleuthProperties stackdriverSleuthProperties(SleuthProperties sleuthPrope
...
 	}
 
 	@Bean
 	@ConditionalOnMissingBean
-	public SpanTranslator spanTranslator(LabelExtractor labelExtractor) {
-		return new SpanTranslator(labelExtractor);
+	public Sender stackdriverSender() throws IOException {
+		return StackdriverSender.newBuilder()
+				.projectId(this.finalProjectIdProvider.getProjectId())
+				.callOptions(CallOptions.DEFAULT.withCallCredentials(
```
meltsufin: I think we need to expose the rest of the `CallOptions` through `GcpTraceProperties`.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/trace/StackdriverTraceAutoConfiguration.java
```diff
@@ -155,88 +138,24 @@ public Sampler defaultTraceSampler(SamplerProperties config) {
 	}
 
 	@Configuration
-	@ConditionalOnProperty(name = "spring.sleuth.http.enabled", havingValue = "true", matchIfMissing = true)
+	@ConditionalOnProperty(name = "spring.sleuth.http.enabled",
+			havingValue = "true", matchIfMissing = true)
 	@AutoConfigureBefore(TraceHttpAutoConfiguration.class)
 	public static class StackdriverTraceHttpAutoconfiguration {
```
meltsufin: Are these client and server parsers still needed? If so, why aren't they part of the Sender lib?

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/trace/StackdriverTraceAutoConfiguration.java
joaoandremartins: Shall we do that in another PR?
There are a lot of parameters to be exposed.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/trace/StackdriverTraceAutoConfiguration.java
joaoandremartins: These parsers are "enriching" the spans, as in, adding information to them like a better name (localhost:8080 instead of no host name), some headers, etc.
They are still required, although I don't know why they aren't in the new zipkin libs...

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/trace/StackdriverTraceAutoConfiguration.java
meltsufin: Sure, feel free to create a linked issue.

joaoandremartins: https://github.com/spring-cloud/spring-cloud-gcp/issues/662

adriancole:
> @adriancole <https://github.com/adriancole> I made that test and it
> doesn't work.
>
> I suggest we merge it this way and get back to this later, then.
>
I'm on holiday, but I disagree on punting this part (about the extra
declaration of the AsyncReporter). Folks who own this repo can feel free to
merge eventhough I disagree, but I'd recommend them to not release it until
it is fixed.

On that note, seems someone else could actually try this (not just look at
it): While redundant declaration looks small but in practice it hints
something isn't working, or should be understood. Maybe someone from google
or sleuth can help. Try on sleuth's gitter? I will be back online 21 May,
in worst case.

saturnism:
i thought if we simply rely on sleuth zipkin starter (not just sleuth) that also has all the asyncreporter constructs; from there, all we need to do is to provide the Sender bean, right?

joaoandremartins:
We needed to depend on `ZipkinAutoConfiguration`, which is where the reporter gets auto-configured.

@balopat can you please review this here? @meltsufin and @adriancole are away, but I think all of their concerns are addressed.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/trace/StackdriverTraceAutoConfiguration.java
```diff
@@ -73,17 +59,13 @@
...
 @Configuration
-@EnableConfigurationProperties({ SamplerProperties.class, GcpTraceProperties.class, SleuthProperties.class })
+@EnableConfigurationProperties(
+		{ SamplerProperties.class, GcpTraceProperties.class, SleuthProperties.class })
 @ConditionalOnProperty(value = "spring.cloud.gcp.trace.enabled", matchIfMissing = true)
 @ConditionalOnClass(TraceConsumer.class)
-@Import({ StackdriverTraceAutoConfiguration.TraceConsumerConfiguration.class,
-		StackdriverTraceAutoConfiguration.StackdriverTraceHttpAutoconfiguration.class })
-@AutoConfigureBefore({ TraceAutoConfiguration.class })
+@AutoConfigureBefore({ ZipkinAutoConfiguration.class })
```
joaoandremartins: FYI `ZipkinAutoConfiguration` is already `@AutoConfigureBefore(TraceAutoConfiguration.class)`, which is why I think it's safe to remove it from here.

joaoandremartins:
@ChengyuanZhao can you please review this PR next week whenever you have some time?

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/trace/StackdriverTraceAutoConfigurationTests.java
```diff
@@ -66,50 +63,19 @@ public void test() {
 			SleuthProperties sleuthProperties = context.getBean(SleuthProperties.class);
 			assertThat(sleuthProperties.isTraceId128()).isTrue();
 			assertThat(sleuthProperties.isSupportsJoin()).isFalse();
-
```
ChengyuanZhao: I get that a lot of things were deleted, and a lot of the responsibility has shifted from this module to the dependency. But would it still be good to have some tests that see if the things are still working as they should ? (perhaps that would be more of an integration test).

joaoandremartins: I wanted Trace integration tests too. I looked a while back into implementing them, but found out the client library won't let you query traces by a key, which makes the integration tests infeasible..
I did test this manually, though, and it works.

PR #651, opened 2018-05-01
Uses idomatic values for Config configuration

### LINKED ISSUE #244: Spring Cloud GCP Config should use idiomatic config values as default
Currently spring-cloud-gcp-config requires `spring.cloud.gcp.config.name` and sets profile w/ `spring.cloud.gcp.config.profile`.  It should use the idiomatic settings `spring.application.name` and `spring.profiles.active`.

joaoandremartins:
Application name is now obtained from the spring.application.name
property. Likewise, the profile is also now the comma-delimited list
of the active profiles from spring.profiles.active property.

Fixes #244

### DIFF
--- a/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/config/GcpConfigProperties.java
+++ b/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/config/GcpConfigProperties.java
@@ -16,11 +16,14 @@
 
 package org.springframework.cloud.gcp.autoconfigure.config;
 
+import org.springframework.beans.factory.annotation.Value;
 import org.springframework.boot.context.properties.ConfigurationProperties;
 import org.springframework.boot.context.properties.NestedConfigurationProperty;
 import org.springframework.cloud.gcp.core.Credentials;
 import org.springframework.cloud.gcp.core.CredentialsSupplier;
 import org.springframework.cloud.gcp.core.GcpScope;
+import org.springframework.core.env.Environment;
+import org.springframework.util.StringUtils;
 
 /**
  * Configuration for {@link GoogleConfigPropertySourceLocator}.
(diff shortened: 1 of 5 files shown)


spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/config/GcpConfigProperties.java
```diff
@@ -32,25 +35,46 @@
...
+	 */
+	@Value("${spring.application.name:application}")
 	private String name;
 
-	/** Profile under which app is running (e.g., "prod", "dev", "test", etc.). */
-	private String profile = "default";
+	/**
+	 *  Comma-delimited string of profiles under which the app is running.
+	 */
+	private String profile;
```
meltsufin: Is the intent to allow for this property to be overwritten, but it defaults to `spring.profiles.active`? If so, the comment should reflect that.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/config/GcpConfigProperties.java
```diff
@@ -32,25 +35,46 @@
...
+	 */
 	@NestedConfigurationProperty
 	private final Credentials credentials = new Credentials(GcpScope.RUNTIME_CONFIG_SCOPE.getUrl());
 
+	public GcpConfigProperties(Environment environment) {
+		String[] profiles = environment.getActiveProfiles();
+		if (profiles.length == 0) {
+			profiles = environment.getDefaultProfiles();
+		}
+		this.profile = StringUtils.arrayToCommaDelimitedString(profiles);
```
meltsufin: So, you might end up with configuration name like `myapp_prod,xyz,perf` for the config name?
Are commas even allowed in the config name.
I think we're trying to be too smart here. I would just collapse everything into a simple `spring.cloud.gcp.config.name` like you and Ray suggested in the issue. Why are we not doing that here?

joaoandremartins: Because [this is exactly what Spring Cloud Config Client is doing](https://github.com/spring-cloud/spring-cloud-config/blob/master/spring-cloud-config-client/src/main/java/org/springframework/cloud/config/client/ConfigClientProperties.java#L109), and I love that they solved this issue for us.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/config/GcpConfigProperties.java
joaoandremartins: Added clarification, though my preference is actually that this property never gets used with our property prefix (`spring.cloud.gcp.config...`). Although it's there, in case an override is necessary for any reason...

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/config/GcpConfigProperties.java
```diff
@@ -32,25 +35,48 @@
...
+	/**
+	 *  Name of the application.
+	 */
+	@Value("${spring.application.name:application}")
 	private String name;
 
-	/** Profile under which app is running (e.g., "prod", "dev", "test", etc.). */
-	private String profile = "default";
+	/**
+	 *  Comma-delimited string of profiles under which the app is running.
```
meltsufin: It appears that Spring Cloud Config, parses the list of profiles and applies them all at the same time. Since we do not have that ability on GCP currently, I would suggest instead to just use the last profile in the list, not the whole comma-delimited list.

PR #652, opened 2018-05-02
Version written to properties file

### LINKED ISSUE #630: UsageTrackingHeaderProvider determines wrong project version
This may be caused by `getClass().getPackage().getImplementationVersion()` not working correctly when the project jars are repackaged. We should find an alternative way of capturing the version. Perhaps by writing it to a file during the Maven build.

ChengyuanZhao:
fixes #630


spring-cloud-gcp-core/src/main/resources/tracking-header.properties
```diff
@@ -0,0 +1 @@
+metrics.version=@parent.version@
```
meltsufin: Can we scope this file to a package to avoid potential collisions with user code?
Also, maybe call the property `project.version`?

spring-cloud-gcp-core/src/main/java/org/springframework/cloud/gcp/core/UsageTrackingHeaderProvider.java
```diff
@@ -49,7 +78,7 @@ public UsageTrackingHeaderProvider(Class clazz) {
 
 		headers.put("User-Agent",
 				"Spring/" + this.getClass().getPackage().getImplementationVersion()
```
meltsufin: What about this line?

ChengyuanZhao: Is this one supposed to always identical to the other? 

To confirm, there are two components  "Spring/" and "spring-cloud-gcp/" 
The former uses "this.getClass()" while the latter uses "this.clazz". The second is the one I replaced. Is the first one also supposed to be replaced identically, or is it supposed to potentially show the different repackaged version?

spring-cloud-gcp-core/src/main/resources/tracking-header.properties
ChengyuanZhao: Could you elaborate on what you mean? I'm not familiar with properties files. What do you mean by "scope this file to a package?"

artembilan: Well, our practice is do not have extra files on the matter, but just compile in the version into some class. The best sample is: https://github.com/spring-projects/spring-integration/blob/master/spring-integration-core/src/main/java/org/springframework/integration/config/xml/AbstractIntegrationNamespaceHandler.java#L47

But we can try some other solution, e.g. generate such a class during maven build. 

Saying that -1 from me for this extra file which may be lost during repackaging as you mention in the issue.

meltsufin: I just mean putting it in a subdirectory like `resources/org/springframework/cloud/core/` like we did for the logging config files.

spring-cloud-gcp-core/src/main/java/org/springframework/cloud/gcp/core/UsageTrackingHeaderProvider.java
meltsufin: I think the idea was that some modules within the project may have different versions. This would actually mean that you need to generated this file separately for each module.
@joaoandremartins

spring-cloud-gcp-core/src/main/resources/tracking-header.properties
meltsufin: @ChengyuanZhao On a second thought, I think @artembilan might be right. Depending on how people are repacking our code, this file can be lost, or generated differently. The safest thing might be to just hard-code the version in a Java class.

artembilan: There is this for your consideration: http://www.mojohaus.org/templating-maven-plugin/index.html

ChengyuanZhao:
I'm gonna close this PR and start a new one. We're going with hard-coded version string in java, and pretty much everything here would be reverted anyway.

PR #653, opened 2018-05-02
Pub/Sub Stream Binder code sample

joaoandremartins:
This is a code sample that sends messages using the Spring Cloud Stream
binder.

### DIFF
--- a/spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-binder-sample/README.adoc
+++ b/spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-binder-sample/README.adoc
@@ -1,8 +1,8 @@
 = Spring Cloud GCP Stream Binder for Pub/Sub Code Sample
 
-This code sample demonstrates how to use the Spring Cloud Stream binders for Google Cloud Pub/Sub.
+This code sample demonstrates how to use the Spring Cloud Stream binder for Google Cloud Pub/Sub.
 The sample app prompts a user for a message and their user name.
-That data is added to a `Message` object, together with the time of message creation, and is sent through Google Cloud Pub/Sub to a sink which simply logs the message.
+That data is added to a `UserMessage` object, together with the time of message creation, and is sent through Google Cloud Pub/Sub to a sink which simply logs the message.
 
 If the topic and subscription for the sink and source don't exist, the binder will automatically create them in Google Cloud Pub/Sub based on the values in link:src/main/resources/application.properties[application.properties].
 
--- a/spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-binder-sample/README.adoc
+++ b/spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-binder-sample/README.adoc
@@ -1,0 +1,29 @@
+= Spring Cloud GCP Stream Binder for Pub/Sub Code Sample
+
+This code sample demonstrates how to use the Spring Cloud Stream binders for Google Cloud Pub/Sub.
+The sample app prompts a user for a message and their user name.
+That data is added to a `Message` object, together with the time of message creation, and is sent through Google Cloud Pub/Sub to a sink which simply logs the message.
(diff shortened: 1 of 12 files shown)


joaoandremartins:
It would be great if anyone from Spring Cloud Stream could take a look here and let me know if there is anything else we should be conveying in this sample.
We don't really support consumer groups or partitioning FYI..

spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-binder-sample/README.adoc
```diff
@@ -0,0 +1,29 @@
+= Spring Cloud GCP Stream Binder for Pub/Sub Code Sample
+
+This code sample demonstrates how to use the Spring Cloud Stream binders for Google Cloud Pub/Sub.
```
artembilan: s/binders/Binder ?

spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-binder-sample/src/main/java/com/example/Message.java
```diff
@@ -0,0 +1,64 @@
...
+ */
+
+package com.example;
+
+import java.time.LocalDateTime;
+
+/**
+ * @author João André Martins
+ */
+public class Message {
```
artembilan: Maybe call it `UserMessage` do not confuse with many existing `Message` classes ?

spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-binder-sample/src/main/java/com/example/SourceExample.java
```diff
@@ -0,0 +1,46 @@
...
+ */
+@EnableBinding(Source.class)
+@RestController
+public class SourceExample {
+
+	@Autowired
+	private Source source;
+
+	@PostMapping("/newMessage")
+	public Message sendMessage(@RequestParam("messageBody") String messageBody,
```
artembilan: So, why `@Publisher(Source.OUTPUT)` doesn't work here for you?

sobychacko:
@joaoandremartins I just glanced through the code and it looks good to me. What happens if someone tries to set consumer group or partitoning properties? Does that throw an exception or no-op?

spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-binder-sample/src/main/java/com/example/SourceExample.java
joaoandremartins: Humm `@Publisher(Source.OUTPUT)` causes this method to return a 404. Do you have an idea why?

artembilan: Well, no idea, indeed.
Would you mind to share stack trace from logs on the matter?
Thanks

joaoandremartins:
@sobychacko they're no-ops.
Setting a consumer group names adds the group suffix to the Pub/Sub subscription, but if I recall from previous conversations, there isn't actual consumer group distribution since Pub/Sub doesn't support that notion. i.e., a message sent to a topic X with subscriptions X.a and X.b is delivered to both subscriptions instead of just one.

artembilan:
Checkstyle has failed for this build.
Also I still see a lot of compilation warning.

Didn't you fix them recently? What am I missing ?

spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-binder-sample/src/main/java/com/example/SourceExample.java
artembilan: Right, there is a proxying.
So, let's never mind for now!
It's not important for this sample.

PR #654, opened 2018-05-02
Hardcoded metrics version

### LINKED ISSUE #630: UsageTrackingHeaderProvider determines wrong project version

ChengyuanZhao:
fixes #630 

as agreed in that issue, we're setting all modules to the same hard coded version.  unit test now is a check for us to manually update the version if needed

### DIFF
--- a/spring-cloud-gcp-core/src/test/java/org/springframework/cloud/gcp/core/UsageTrackingHeaderProviderIT.java
+++ b/spring-cloud-gcp-core/src/test/java/org/springframework/cloud/gcp/core/UsageTrackingHeaderProviderIT.java
@@ -16,8 +16,6 @@
 
 package org.springframework.cloud.gcp.core;
 
-import java.util.regex.Pattern;
-
 import org.junit.Test;
 
 import static org.assertj.core.api.Assertions.assertThat;
@@ -28,17 +26,21 @@ import static org.assertj.core.api.Assertions.assertThat;
  *
  * @author João André Martins
  * @author Mike Eltsufin
+ * @author Chengyuan Zhao
  */
 public class UsageTrackingHeaderProviderIT {
 
+	/**
+	 * This test is check if the hard-coded version needs to be manually updated.
+	 */
 	@Test
 	public void testGetHeaders() {
 		UsageTrackingHeaderProvider subject = new UsageTrackingHeaderProvider(this.getClass());
 
-		String versionRegex = "\\d+\\.\\d+\\.\\d+\\.[BUILD-SNAPSHOT|M\\d+|RC\\d+|RELEASE]$";
+		String builtVersion = this.getClass().getPackage().getImplementationVersion();
 		assertThat(subject.getHeaders()).containsKey("User-Agent");
-		assertThat(subject.getHeaders().get("User-Agent")).matches(
-				Pattern.compile("Spring/" + versionRegex + " spring-cloud-gcp-core/" + versionRegex));
+		assertThat(subject.getHeaders().get("User-Agent")).isEqualTo(
+				"Spring/" + builtVersion + " spring-cloud-gcp-core/" + builtVersion);
 		assertThat(subject.getHeaders().size()).isEqualTo(1);
 	}
 }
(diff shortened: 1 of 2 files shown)


spring-cloud-gcp-core/src/test/java/org/springframework/cloud/gcp/core/UsageTrackingHeaderProviderIT.java
```diff
@@ -28,17 +26,21 @@
  *
  * @author João André Martins
  * @author Mike Eltsufin
+ * @author Chengyuan Zhao
  */
 public class UsageTrackingHeaderProviderIT {
 
+	/**
+	 * This test is check if the hard-coded version needs to be manually updated.
+	 */
```
meltsufin: I would add here another note like,
"If you find this test failing, you might be repackaging the source or changing the version.
Unless, you're doing an official release of the project, it's safe to `@Ignore` this test."

artembilan: I think we should admit that we develop here Open Source and that doesn't sound reasonable to say end-user that he/she is not right: you get a code from here and do whatever you want and can. 

The point is that this metrics pursuing is not so valuable regarding those "custom" artifacts. Maximum what we can ensure is our official release. Anything out of our hands is already out of our responsibility as well and it really might not be traced.are lying to the public.

meltsufin: How does this imply that someone re-packing the source is "not right" or being lied to?
If someone is modifying the source, sure, do whatever you want.
The goal of my comment was to avoid confusing users who might be just rebuilding the source and seeing this non-functional test fail.

artembilan: I just mean that we shouldn't test it at all or leave it as is, because if you change something in the code, you definitely may fail on this or other test.

The current comment on the test is fine with me. Maybe it is necessary to add a link to the class where to fix such a version.

meltsufin: The only reason you should fail on this code is if you changed the project version directly or indirectly via some kind of repackaging.
Another thing we can do, like you said, is to not have the test always enabled. We can instead execute this test only in a "release" Maven profile, but we will need to make sure that this profile is enabled in our release scripts. WDYT?

artembilan: Right, if someone repackage our project his/her own way, that's already out of our control and therefore failed test is fully fine and I wouldn't pay attention. 

OTOH if they fail, they may come to us with the issue and we will ask them why do they repackage our project.

I fully don't understand why do you pursue the goal do not allow to rebuild this Open Source project...

ChengyuanZhao: I can add an option like we do for our integration tests that Travis runs but is usually an ignored test.

meltsufin: @artembilan We really have no problem with the rebuilds. We just want to avoid inadvertently inaccurate metrics reporting.

@ChengyuanZhao I'm ok with that idea, but I think Artem disagrees. I don't feel strongly enough about it. So, if Artem insists on keeping this test on by default, I'm fine with that.

artembilan: As I said in one comment above: we can give guaranties about metrics only for our official release.
If they rebuild project some other non-standard way, let's treat that as not our library!

We spent already too much time to try to protect ourselves from they rare situation.
And again: since it is Open Source, any smart hacker will find out the way to mock our metrics anyway.

If you want one more Maven profile, I don't mind. 
I'll rethink my vision on the matter on my own.
Since it looks like you have a hard time to agree with me, then my position is mistaken somehow...

meltsufin: OK, let's merge this as is and we can deal with profiles separately.

PR #655, opened 2018-05-03
Stream Binder refdoc section

### LINKED ISSUE #293: Add remaining Javadoc and documentation to Cloud Stream binder
#145 was submitted with minimal Javadoc and no documentation.
We should:
- [ ] Add an entry to the reference document,
- [x] Add a code sample,
- [x] Add a README,
- [x] Add Javadoc.

joaoandremartins:
Adds a section about the Spring Cloud Stream Binder for Google Cloud
Pub/Sub to the reference document.

Fixes #293

### DIFF
--- a/spring-cloud-gcp-docs/src/main/asciidoc/index.adoc
+++ b/spring-cloud-gcp-docs/src/main/asciidoc/index.adoc
@@ -36,6 +36,8 @@ include::sql.adoc[]
 
 include::spring-integration.adoc[]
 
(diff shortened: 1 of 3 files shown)


joaoandremartins:
I added a similar `Overview` entry to the section, even though I can't think of anything to write there.
The behaviour of the binder is a little basic, since we don't support partitioning or consumer groups.
The configurable bits are also limited to resource auto creation, but we can add more to the document once we have it.

PR #658, opened 2018-05-07
Enables header filtering in SI adapters

### LINKED ISSUE #63: Sanitize PubsubMessage headers
Currently, the SI-generated headers, like ID and timestamp, are being preserved on message conversion (Spring to PubsubMessage).

We should remove those messages on conversion, as well as any GcpHeaders, while at the same time, providing some means of preserving those headers if the users wants.

More details [here](https://github.com/spring-cloud/spring-cloud-gcp/pull/61#discussion_r124899136).

joaoandremartins:
Makes use of the HeaderMapper pattern.

Fixes #63

### DIFF
--- a/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/PubSubHeaderMapper.java
+++ b/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/PubSubHeaderMapper.java
@@ -40,7 +40,7 @@ public class PubSubHeaderMapper implements HeaderMapper<Map<String, String>> {
 	/**
 	 * Patterns of headers to map in {@link #fromHeaders(MessageHeaders, Map)}.
 	 */
-	private String[] outboundHeaderPatternsToMap = {
+	private String[] outboundHeaderPatterns = {
 			"!" + MessageHeaders.ID,
 			"!" + MessageHeaders.TIMESTAMP,
 			"*"};
@@ -48,28 +48,28 @@ public class PubSubHeaderMapper implements HeaderMapper<Map<String, String>> {
 	/**
 	 * Patterns of headers to map in {@link #toHeaders(Map)}.
 	 */
-	private String[] inboundHeaderPatternsToMap = {"*"};
+	private String[] inboundHeaderPatterns = {"*"};
 
 	/**
 	 * Set the patterns of the headers to be mapped in {@link #fromHeaders(MessageHeaders, Map)}.
-	 * @param outboundHeaderPatternsToMap header patterns to be mapped
+	 * @param outboundHeaderPatterns header patterns to be mapped
 	 */
-	public void setOutboundHeaderPatternsToMap(String... outboundHeaderPatternsToMap) {
-		Assert.notNull(outboundHeaderPatternsToMap, "Header patterns can't be null.");
-		Assert.noNullElements(outboundHeaderPatternsToMap, "No header pattern can be null.");
-		this.outboundHeaderPatternsToMap =
-				Arrays.copyOf(outboundHeaderPatternsToMap, outboundHeaderPatternsToMap.length);
+	public void setOutboundHeaderPatterns(String... outboundHeaderPatterns) {
+		Assert.notNull(outboundHeaderPatterns, "Header patterns can't be null.");
+		Assert.noNullElements(outboundHeaderPatterns, "No header pattern can be null.");
+		this.outboundHeaderPatterns =
+				Arrays.copyOf(outboundHeaderPatterns, outboundHeaderPatterns.length);
 	}
 
 	/**
 	 * Set the patterns of the headers to be mapped in {@link #toHeaders(Map)}.
-	 * @param inboundHeaderPatternsToMap header patterns to be mapped
+	 * @param inboundHeaderPatterns header patterns to be mapped
 	 */
-	public void setInboundHeaderPatternsToMap(String... inboundHeaderPatternsToMap) {
-		Assert.notNull(inboundHeaderPatternsToMap, "Header patterns can't be null.");
-		Assert.noNullElements(inboundHeaderPatternsToMap, "No header pattern can be null.");
-		this.inboundHeaderPatternsToMap =
-				Arrays.copyOf(inboundHeaderPatternsToMap, inboundHeaderPatternsToMap.length);
+	public void setInboundHeaderPatterns(String... inboundHeaderPatterns) {
+		Assert.notNull(inboundHeaderPatterns, "Header patterns can't be null.");
+		Assert.noNullElements(inboundHeaderPatterns, "No header pattern can be null.");
+		this.inboundHeaderPatterns =
+				Arrays.copyOf(inboundHeaderPatterns, inboundHeaderPatterns.length);
 	}
 
 	/**
@@ -86,7 +86,7 @@ public class PubSubHeaderMapper implements HeaderMapper<Map<String, String>> {
 		messageHeaders.entrySet().stream()
 				.filter(entry -> Boolean.TRUE.equals(
 						PatternMatchUtils.smartMatch(entry.getKey(),
-								this.outboundHeaderPatternsToMap)))
+								this.outboundHeaderPatterns)))
 				.forEach(entry -> pubsubMessageHeaders.put(
 						entry.getKey(), entry.getValue().toString()));
 	}
@@ -104,7 +104,7 @@ public class PubSubHeaderMapper implements HeaderMapper<Map<String, String>> {
 		return pubsubMessageHeaders.entrySet().stream()
 				.filter(entry -> Boolean.TRUE.equals(
 						PatternMatchUtils.smartMatch(entry.getKey(),
-								this.inboundHeaderPatternsToMap)))
+								this.inboundHeaderPatterns)))
 				.collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
 	}
 }
--- a/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/PubSubHeaderMapper.java
+++ b/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/PubSubHeaderMapper.java
@@ -30,7 +30,9 @@ public class PubSubHeaderMapper implements HeaderMapper<Map<String, String>> {
 	/**
 	 * Header patterns to map in {@link #fromHeaders(MessageHeaders, Map)}.
 	 */
-	private String[] outboundHeaderPatternsToMap = {"*"};
+	private String[] outboundHeaderPatternsToMap = {"*",
+			"!" + MessageHeaders.ID,
+			"!" + MessageHeaders.TIMESTAMP};
 
 	/**
 	 * Header patterns to map in {@link #toHeaders(Map)}.
@@ -70,10 +72,8 @@ public class PubSubHeaderMapper implements HeaderMapper<Map<String, String>> {
 	public void fromHeaders(MessageHeaders messageHeaders,
 			final Map<String, String> pubsubMessageHeaders) {
 		messageHeaders.entrySet().stream()
-				.filter(entry -> this.outboundHeaderPatternsToMap != null
-						? PatternMatchUtils.smartMatch(entry.getKey(),
-								this.outboundHeaderPatternsToMap)
-						: false)
+				.filter(entry -> PatternMatchUtils.smartMatch(entry.getKey(),
+						this.outboundHeaderPatternsToMap))
 				.forEach(entry -> pubsubMessageHeaders.put(
 						entry.getKey(), entry.getValue().toString()));
 	}
@@ -89,10 +89,8 @@ public class PubSubHeaderMapper implements HeaderMapper<Map<String, String>> {
 	@Override
 	public Map<String, Object> toHeaders(Map<String, String> pubsubMessageHeaders) {
 		return pubsubMessageHeaders.entrySet().stream()
-				.filter(entry -> this.inboundHeaderPatternsToMap != null
-						? PatternMatchUtils.smartMatch(entry.getKey(),
-								this.inboundHeaderPatternsToMap)
-						: false)
+				.filter(entry -> PatternMatchUtils.smartMatch(entry.getKey(),
+						this.inboundHeaderPatternsToMap))
 				.collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
 	}
 }
--- a/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/PubSubHeaderMapper.java
+++ b/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/PubSubHeaderMapper.java
@@ -21,9 +21,9 @@ import java.util.Map;
 import java.util.stream.Collectors;
 
 import org.springframework.integration.mapping.HeaderMapper;
+import org.springframework.integration.util.PatternMatchUtils;
 import org.springframework.messaging.MessageHeaders;
 import org.springframework.util.Assert;
-import org.springframework.util.PatternMatchUtils;
 
 public class PubSubHeaderMapper implements HeaderMapper<Map<String, String>> {
 
@@ -71,8 +71,8 @@ public class PubSubHeaderMapper implements HeaderMapper<Map<String, String>> {
 			final Map<String, String> pubsubMessageHeaders) {
 		messageHeaders.entrySet().stream()
 				.filter(entry -> this.outboundHeaderPatternsToMap != null
-						? PatternMatchUtils.simpleMatch(
-								this.outboundHeaderPatternsToMap, entry.getKey())
+						? PatternMatchUtils.smartMatch(entry.getKey(),
+								this.outboundHeaderPatternsToMap)
 						: false)
 				.forEach(entry -> pubsubMessageHeaders.put(
 						entry.getKey(), entry.getValue().toString()));
@@ -90,8 +90,8 @@ public class PubSubHeaderMapper implements HeaderMapper<Map<String, String>> {
 	public Map<String, Object> toHeaders(Map<String, String> pubsubMessageHeaders) {
 		return pubsubMessageHeaders.entrySet().stream()
 				.filter(entry -> this.inboundHeaderPatternsToMap != null
-						? PatternMatchUtils.simpleMatch(
-								this.inboundHeaderPatternsToMap, entry.getKey())
+						? PatternMatchUtils.smartMatch(entry.getKey(),
+								this.inboundHeaderPatternsToMap)
 						: false)
 				.collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
 	}
--- a/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/PubSubHeaderMapper.java
+++ b/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/PubSubHeaderMapper.java
@@ -1,0 +1,83 @@
+/*
+ *  Copyright 2018 original author or authors.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package org.springframework.cloud.gcp.pubsub.integration;
+
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.Map;
+
+import org.springframework.messaging.MessageHeaders;
+import org.springframework.messaging.support.HeaderMapper;
+import org.springframework.util.Assert;
+
+public class PubSubHeaderMapper implements HeaderMapper<Map<String, String>> {
+
+	/**
+	 * Headers to exclude in the {@link #fromHeaders(MessageHeaders, Map)}.
+	 */
+	private String[] doNotMapHeaderNames = {
+			MessageHeaders.ID, MessageHeaders.TIMESTAMP
+	};
+
+	private boolean filterHeaders;
+
+	/**
+	 * Set the names of the headers to filtered out in {@link #fromHeaders(MessageHeaders, Map)}.
+	 * @param doNotMapHeaderNames
+	 */
+	public void setDoNotMapHeaderNames(String... doNotMapHeaderNames) {
+		Assert.notNull(doNotMapHeaderNames, "Header names can't be null.");
+		Assert.noNullElements(doNotMapHeaderNames, "No header name can be null.");
+		this.doNotMapHeaderNames = doNotMapHeaderNames;
+	}
+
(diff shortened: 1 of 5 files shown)


joaoandremartins:
A few unit tests to come.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/PubSubHeaderMapper.java
```diff
@@ -0,0 +1,83 @@
...
+	private boolean filterHeaders;
+
+	/**
+	 * Set the names of the headers to filtered out in {@link #fromHeaders(MessageHeaders, Map)}.
+	 * @param doNotMapHeaderNames
+	 */
+	public void setDoNotMapHeaderNames(String... doNotMapHeaderNames) {
+		Assert.notNull(doNotMapHeaderNames, "Header names can't be null.");
+		Assert.noNullElements(doNotMapHeaderNames, "No header name can be null.");
+		this.doNotMapHeaderNames = doNotMapHeaderNames;
```
artembilan: `Arrays.copyOf(doNotMapHeaderNames, doNotMapHeaderNames.length);` must be used to avoid the cases when end-user may mutate the original array.
On the other hand SonarQube doesn't like this code - exactly for the reason I've explained.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/PubSubHeaderMapper.java
```diff
@@ -0,0 +1,83 @@
...
+	 * @param messageHeaders headers to map from
+	 * @param pubsubMessageHeaders headers in their final format
+	 */
+	@Override
+	public void fromHeaders(MessageHeaders messageHeaders,
+			final Map<String, String> pubsubMessageHeaders) {
+		messageHeaders.forEach((key, value) -> pubsubMessageHeaders.put(key, value.toString()));
+
+		if (this.filterHeaders) {
+			Arrays.stream(this.doNotMapHeaderNames).forEach(messageHeaders::remove);
```
artembilan: What is the goal to remove headers from the original message?
I thought an idea was like do not map headers to the target `pubsubMessageHeaders`...

Must be `messageHeaders.stream().filter().forEach()`.

No? What am I missing?

And why do we need this `filterHeaders` at all?

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/PubSubHeaderMapper.java
```diff
@@ -0,0 +1,83 @@
...
+	/**
+	 * Generate headers in {@link org.springframework.messaging.Message} format from
+	 * {@code Map<String, String>}.
+	 * @param pubsubMessageHeaders headers in {@link com.google.pubsub.v1.PubsubMessage} format
+	 * @return a map with headers in the {@link org.springframework.messaging.Message} format
+	 */
+	@Override
+	public MessageHeaders toHeaders(Map<String, String> pubsubMessageHeaders) {
+		Map<String, Object> typifiedHeaders = new HashMap<>();
+		pubsubMessageHeaders.forEach((key, value) -> typifiedHeaders.put(key, value));
```
artembilan: Why do you map all headers from here?
Shouldn't we filter them as well?

Shouldn't we have a separate set of headers to map `to` and `from`?
See, `StompHeaderMapper` for example: https://github.com/spring-projects/spring-integration/blob/master/spring-integration-stomp/src/main/java/org/springframework/integration/stomp/support/StompHeaderMapper.java

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/PubSubHeaderMapper.java
```diff
@@ -0,0 +1,83 @@
...
+
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.Map;
+
+import org.springframework.messaging.MessageHeaders;
+import org.springframework.messaging.support.HeaderMapper;
+import org.springframework.util.Assert;
+
+public class PubSubHeaderMapper implements HeaderMapper<Map<String, String>> {
```
artembilan: Please, consider to implement exactly `org.springframework.integration.mapping.HeaderMapper`.
The problem with the current `org.springframework.messaging.support.HeaderMapper` that it requires to return `MessageHeaders`, which is much heavier object and requires special treatment to avoid extra re-creations.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/PubSubHeaderMapper.java
joaoandremartins: I can filter the maps as I add them, if you want.
If you notice, first I add all the headers, then I remove the ones to be filtered out.
Since it is a map, accesses are O(1), so I thought it would be fine, but more readable like this.
I can change it if you like, though.

Pub/Sub doesn't really add any headers to the messages and I'm not sure a set of `toHeaders` headers to filter would be useful. YAGNI?

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/PubSubHeaderMapper.java
joaoandremartins: Oops, my mistake, the code was intended to remove headers from `pubsubMessageHeaders`.

`filterHeaders` is needed to disable header filtering. In fact, the default for the outbound channel adapter will be `false`.

joaoandremartins: So the first `forEach()`, besides copying all the headers, also casts the `Map<String, Object>` to `Map<String, String>`.

I like having the filtering out because otherwise, we'd have to do O(n) passes over `doNotMapHeaderNames`, since it isn't a set but an array.

I prefer to leave things like this, let me know if you see any issues though.

artembilan: That's not what we do in other implementations.
Actually we rely on the `pattern`. And if you want to map all, you use `*`.

The mentioned `StompHeaderMapper` should be a good sample.

Right, you can use `*` by default.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/PubSubHeaderMapper.java
artembilan: ???
This is  `toHeaders()` method and I fully don't see any filtering logic here, therefore your explanation just does not make sense to me.

Please, try to express your mind one more time.

Sorry for confusing you, but I can't accept what I don't understand.

joaoandremartins: Sorry, I realised that late and edited my comment.
I'm not sure if it's worth to filter headers on `PubsubMessage`s because Pub/Sub doesn't add any headers. It feels like a feature that isn't necessary?

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/PubSubHeaderMapper.java
joaoandremartins: Sorry, I'm not following.
What do you mean with `pattern` and `*`?
After looking at that Stomp header mapper, I'm still not sure what you mean or what the issue with the current code is.
Thanks!

artembilan: The issue with the current solution that it doesn't follow a pattern we have in other implementations.

It definitely doesn't matter if we map or filter. From the math perspective it's really the same, but from consistency it would be better to follow what we have so far in other places. 

We do expose `headerPatternsToMap` and allow to configure the patters, for example `foo_*`. Therefore all the headers with the `foo_` prefix will be mapped. 

Your current solution requires from me to map all my `foo_` header names.

You are missing the fact of pattern matching in the mentioned `StompHeaderMapper`:
https://github.com/spring-projects/spring-integration/blob/master/spring-integration-stomp/src/main/java/org/springframework/integration/stomp/support/StompHeaderMapper.java#L212

If you still would like to `filter`, we suggest a solution with the `!` operator: https://docs.spring.io/spring-integration/docs/5.0.4.RELEASE/reference/html/amqp.html#amqp-message-headers

artembilan:
Oh! Thank you for the reminder, @joaoandremartins !

This is what I meant on the meeting: https://github.com/spring-projects/spring-integration-aws/blob/master/src/main/java/org/springframework/integration/aws/support/AbstractMessageAttributesHeaderMapper.java

This the latest implementation of the header mapping pattern. 
Pay attention to the `smartMatch` and JavaDocs on the `setOutboundHeaderNames()`.

spring-cloud-gcp-pubsub/src/test/java/org/springframework/cloud/gcp/pubsub/integration/PubSubHeaderMapperTests.java
```diff
@@ -0,0 +1,74 @@
...
+ * @author João André Martins
+ */
+public class PubSubHeaderMapperTests {
+
+	@Test
+	public void testFilterHeaders() {
+		PubSubHeaderMapper mapper = new PubSubHeaderMapper();
+		mapper.setOutboundHeaderPatternsToMap(
+				"!" + MessageHeaders.ID,
+				"!" + MessageHeaders.TIMESTAMP,
```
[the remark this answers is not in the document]
joaoandremartins: Right, Google Cloud Pub/Sub users are billed by data volume, so it makes sense to exclude them.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/PubSubHeaderMapper.java
```diff
@@ -0,0 +1,96 @@
...
+
+	/**
+	 * Header patterns to map in {@link #fromHeaders(MessageHeaders, Map)}.
+	 */
+	private String[] outboundHeaderPatternsToMap = {"*",
+			"!" + MessageHeaders.ID,
+			"!" + MessageHeaders.TIMESTAMP};
```
artembilan: Are you sure that the order must be like this?
Do you have a test-case to confirm that we don't send `id` and `timestamp` by default?

See `smartMatch` JavaDocs:
```
 * Pattern match against the supplied patterns; also supports negated ('!')
 * patterns. First match wins (positive or negative).
```

Therefore that asterisk in the first position will always match to `id` and `timestamp` and these negative patterns will never be accessed.

See SI-AWS solution one more time:
https://github.com/spring-projects/spring-integration-aws/blob/master/src/main/java/org/springframework/integration/aws/support/AbstractMessageAttributesHeaderMapper.java#L51

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/PubSubHeaderMapper.java
```diff
@@ -0,0 +1,96 @@
...
+import java.util.Arrays;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+import org.springframework.integration.mapping.HeaderMapper;
+import org.springframework.integration.util.PatternMatchUtils;
+import org.springframework.messaging.MessageHeaders;
+import org.springframework.util.Assert;
+
+public class PubSubHeaderMapper implements HeaderMapper<Map<String, String>> {
```
artembilan: We need class level JavaDocs.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/PubSubHeaderMapper.java
```diff
@@ -0,0 +1,96 @@
...
+	 * <p>Will map only the headers that match the patterns in {@code outboundHeaderPatternsMap}.
+	 * @param messageHeaders headers to map from
+	 * @param pubsubMessageHeaders headers in their final format
+	 */
+	@Override
+	public void fromHeaders(MessageHeaders messageHeaders,
+			final Map<String, String> pubsubMessageHeaders) {
+		messageHeaders.entrySet().stream()
+				.filter(entry -> PatternMatchUtils.smartMatch(entry.getKey(),
+						this.outboundHeaderPatternsToMap))
```
artembilan: The `java.util.function.Predicate` expects primitive `boolean` as the return.
Meanwhile `smartMatch` returns `Boolean` wrapper, including `null` as a third state:
```
* @return true for positive match; false for negative; null if no pattern matches.
```
Therefore condition here must be like this:
```
Boolean.TRUE.equals(PatternMatchUtils.smartMatch(entry.getKey(), this.outboundHeaderPatternsToMap))
```

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/PubSubHeaderMapper.java
```diff
@@ -0,0 +1,96 @@
...
+				.forEach(entry -> pubsubMessageHeaders.put(
+						entry.getKey(), entry.getValue().toString()));
+	}
+
+	/**
+	 * Generate headers in {@link org.springframework.messaging.Message} format from
+	 * {@code Map<String, String>}.
+	 *
```
artembilan: No blank lines in the method JavaDocs: https://github.com/spring-projects/spring-framework/wiki/Code-Style#javadoc-formatting
Why do we return to the code style concerns from time to time?..

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapter.java
```diff
@@ -135,4 +136,8 @@ public MessageConverter getMessageConverter() {
 	public void setMessageConverter(MessageConverter messageConverter) {
 		this.messageConverter = messageConverter;
 	}
+
+	public void setHeaderPatternsToMap(String... patterns) {
```
artembilan: Why do we lose a flexibility to let end-user to inject their own `HeaderMapper` implementation?
This pattern style is just a `PubSubHeaderMapper` responsiblity, therefore here we need just expect `HeaderMapper<Map<String, String>>`.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/PubSubHeaderMapper.java
joaoandremartins: Checkstyle won't let me add a paragraph without a blank line before...

artembilan: Then Checkstyle config has to be fixed.
I try to get consistency throughout our projects you pointed in the beginning, but this somehow doesn't help...

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapter.java
```diff
@@ -137,7 +138,12 @@ public void setMessageConverter(MessageConverter messageConverter) {
...
 
-	public void setHeaderPatternsToMap(String... patterns) {
-		this.headerMapper.setInboundHeaderPatternsToMap(patterns);
+	/**
+	 * Set the header mapper to map headers from incoming {@link PubsubMessage} into
+	 * {@link Message}.
+	 * @param headerMapper the header mapper
+	 */
+	public void setHeaderMapper(HeaderMapper<Map<String, String>> headerMapper) {
+		this.headerMapper = headerMapper;
```
artembilan: `Assert,notNull()` ?

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/outbound/PubSubMessageHandler.java
```diff
@@ -67,7 +68,7 @@
 
 	private ListenableFutureCallback<String> publishCallback;
 
-	private PubSubHeaderMapper headerMapper = new PubSubHeaderMapper();
+	private HeaderMapper headerMapper = new PubSubHeaderMapper();
```
artembilan: Have you missed generics ?

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/PubSubHeaderMapper.java
```diff
@@ -0,0 +1,108 @@
...
+			"*"};
+
+	/**
+	 * Header patterns to map in {@link #toHeaders(Map)}.
+	 */
+	private String[] inboundHeaderPatternsToMap = {"*"};
+
+	/**
+	 * Set the patterns of the headers to be mapped in {@link #fromHeaders(MessageHeaders, Map)}.
+	 * @param outboundHeaderPatternsToMap
```
artembilan: No JavaDocs description.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/PubSubHeaderMapper.java
```diff
@@ -0,0 +1,108 @@
...
+		Assert.noNullElements(outboundHeaderPatternsToMap, "No header pattern can be null.");
+		this.outboundHeaderPatternsToMap =
+				Arrays.copyOf(outboundHeaderPatternsToMap, outboundHeaderPatternsToMap.length);
+	}
+
+	/**
+	 * Set the patterns of the headers to be mapped in {@link #toHeaders(Map)}.
+	 * @param inboundHeaderPatternsToMap
+	 */
+	public void setInboundHeaderPatternsToMap(String[] inboundHeaderPatternsToMap) {
```
artembilan: Why this is not a `varargs`?

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/PubSubHeaderMapper.java
```diff
@@ -0,0 +1,108 @@
...
+		Assert.notNull(inboundHeaderPatternsToMap, "Header patterns can't be null.");
+		Assert.noNullElements(inboundHeaderPatternsToMap, "No header pattern can be null.");
+		this.inboundHeaderPatternsToMap = inboundHeaderPatternsToMap;
```
artembilan: Why here is no `Arrays.copyOf()` ?

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/PubSubHeaderMapper.java
```diff
@@ -0,0 +1,108 @@
...
+import org.springframework.integration.mapping.HeaderMapper;
+import org.springframework.integration.util.PatternMatchUtils;
+import org.springframework.messaging.MessageHeaders;
+import org.springframework.util.Assert;
+
+/**
+ * Maps headers from {@link com.google.pubsub.v1.PubsubMessage}s to
+ * {@link org.springframework.messaging.Message}s and vice-versa.
+ *
+ * <p>Filters out headers called "id" or "timestamp" on the
```
artembilan: ... by default...

When you provide a mapping, the `id` and `timestamp` rules go away.
You should to include them back to your custom patterns set, if you want to map or not.

joaoandremartins:
PTAL. Sorry it took a while longer than it should to get this right...

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/PubSubHeaderMapper.java
```diff
@@ -62,12 +63,13 @@ public void setOutboundHeaderPatternsToMap(String... outboundHeaderPatternsToMap
 
 	/**
 	 * Set the patterns of the headers to be mapped in {@link #toHeaders(Map)}.
-	 * @param inboundHeaderPatternsToMap
+	 * @param inboundHeaderPatternsToMap header patterns to be mapped
 	 */
-	public void setInboundHeaderPatternsToMap(String[] inboundHeaderPatternsToMap) {
+	public void setInboundHeaderPatternsToMap(String... inboundHeaderPatternsToMap) {
```
artembilan: Do we really need to have `ToMap` suffix in the method name?

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/PubSubHeaderMapper.java
```diff
@@ -0,0 +1,110 @@
...
+ * @author João André Martins
+ */
+public class PubSubHeaderMapper implements HeaderMapper<Map<String, String>> {
+
+	/**
+	 * Patterns of headers to map in {@link #fromHeaders(MessageHeaders, Map)}.
+	 */
+	private String[] outboundHeaderPatterns = {
+			"!" + MessageHeaders.ID,
+			"!" + MessageHeaders.TIMESTAMP,
```
artembilan: We need to add to this set to exclude also a `GcpPubSubHeaders.ACKNOWLEDGEMENT` header as well.
This one is populated by the `PubSubInboundChannelAdapter` and should not be sent to the Pub/Sub back.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/PubSubHeaderMapper.java
artembilan: Any decision made here, please?
Thanks

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapter.java
```diff
@@ -52,6 +53,8 @@
 
 	private MessageConverter messageConverter;
 
+	private HeaderMapper<Map<String, String>> headerMapper = new PubSubHeaderMapper();
```
artembilan: Please, don't forget to update a Copyright in all the affected classes to reflect the current year, e.g. here it must be like `2017-2018`.

joaoandremartins: Changed.

Would be great if this could be automated somehow with IntelliJ...

artembilan: Out of scope of this PR...
We have to follow policy this or other way - nothing personal.

joaoandremartins:
I suggest we address documentation in another PR after this one, so we don't drag here for much longer.

PR #659, opened 2018-05-07
Pubsub jackson payloads whitelisted packages

### LINKED ISSUE #361: Support object payload in PubSubTemplate
A way to send an object and marshall w/ converter, e.g., send(topic, Object message) & use converter, potentially detect Jackson in classpath and automatically register a JSON converter.

Issue spun off from #92.

ChengyuanZhao:
fixes #361


artembilan:
Something happened unexpectedly: 
```
Failed tests:   testCreatePublishPullNextAndDelete(org.springframework.cloud.gcp.autoconfigure.pubsub.it.PubSubTemplateIntegrationTests): expected:<null> but was:<name: "projects/spring-cloud-gcp-ci/topics/tarkus"(..)
```

ChengyuanZhao:
I wonder what happened, because that integration test does delete that topic. Perhaps during a run it was interrupted? I'll delete it manually.

ChengyuanZhao:
That test failure has now been fixed in another PR. PTAL

spring-cloud-gcp-docs/src/main/asciidoc/pubsub.adoc
```diff
@@ -61,6 +61,8 @@ public ListenableFuture<String> publish(String topic, byte[] payload, Map<String
 public ListenableFuture<String> publish(String topic, ByteString payload, Map<String, String> headers)
 
 public ListenableFuture<String> publish(String topic, PubsubMessage pubsubMessage)
+
+public <T> ListenableFuture<String> publish(String topic, T payload, Map<String, String> headers) throws JsonProcessingException;
```
joaoandremartins: Can probably leave the `throws ...` out of the docs.

spring-cloud-gcp-pubsub/src/test/java/org/springframework/cloud/gcp/pubsub/core/test/allowed/AllowedPayload.java
```diff
@@ -0,0 +1,27 @@
...
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package org.springframework.cloud.gcp.pubsub.core.test.allowed;
+
+/**
+ * @author Chengyuan Zhao
+ */
+public class AllowedPayload {
```
joaoandremartins: Could declare this within `PubSubTemplateTests`.

spring-cloud-gcp-pubsub/src/test/java/org/springframework/cloud/gcp/pubsub/core/test/disallowed/DisallowedPayload.java
```diff
@@ -0,0 +1,27 @@
...
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package org.springframework.cloud.gcp.pubsub.core.test.disallowed;
+
+/**
+ * @author Chengyuan Zhao
+ */
+public class DisallowedPayload {
```
joaoandremartins: Ditto on declaration location.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
```diff
@@ -79,6 +82,8 @@ public GcpPubSubAutoConfiguration(GcpPubSubProperties gcpPubSubProperties,
 		this.finalCredentialsProvider = gcpPubSubProperties.getCredentials().hasKey()
 				? new DefaultCredentialsProvider(gcpPubSubProperties)
 				: credentialsProvider;
+		this.trustedPackages = gcpPubSubProperties.getTrustedPackages()
```
joaoandremartins: Inline this in the `pubSubTemplate()` method.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubProperties.java
```diff
@@ -43,6 +47,8 @@
 	 */
 	private String emulatorHost;
 
+	private List<String> trustedPackages = Collections.emptyList();
```
joaoandremartins: Add some Javadoc here?
I'd argue that `jsonMapperTrustedPackages` would be a more elucidating name.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubOperations.java
```diff
@@ -111,4 +125,15 @@
...
+	/**
+	 * Deserializes the payload of the given message into the desired type.
+	 * @param pubsubMessage the message containing the payload data
+	 * @param objectType the class object of the desired type
+	 * @param <T> the desired type of the returned object
+	 * @return an instance of the object from the payload
+	 * @throws IOException if the payload could not be deserialized into the requested
+	 * type
+	 */
+	<T> T getPayloadFromMessage(PubsubMessage pubsubMessage, Class<T> objectType) throws IOException;
```
joaoandremartins: I'm wondering if this should belong to `PubSubOperations` or just be a helper method in `PubSubTemplate`? WDYT?

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
```diff
@@ -66,13 +74,19 @@
...
-	 * @param subscriberFactory the {@link com.google.cloud.pubsub.v1.Subscriber} factory to
-	 *                          subscribe to subscriptions
+	 * publish to topics
+	 * @param subscriberFactory the {@link com.google.cloud.pubsub.v1.Subscriber} factory
+	 * to subscribe to subscriptions
+	 * @param trustedPackages the trusted Java packages for deserialization. If null or
+	 * empty, then default type information is not included in serialization.
 	 */
-	public PubSubTemplate(PublisherFactory publisherFactory, SubscriberFactory subscriberFactory) {
+	public PubSubTemplate(PublisherFactory publisherFactory,
```
joaoandremartins: Can we keep the old constructor and add this one separately, so as not to break the API?

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubProperties.java
artembilan: Well, there is similar deserialization vulnerabilities with the regular Java `Serializable`.
When we introduce such a converter we should ensure to use similar property as well.
So, I think to narrow the name to the `json` would eliminate a possibility to reuse the property in other places based on the trusted sources.

So, my point is to leave the name as is.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
ChengyuanZhao: Will do. Somehow I didn't see we already have a `this.gcpPubSubProperties` haha

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubOperations.java
ChengyuanZhao: I'm fine with that. @meltsufin @artembilan does that sound ok?

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubProperties.java
joaoandremartins: Then let's change it to `deserializationTrustedPackages`.

Anything more informative than `trustedPackages`. Trusted packages of what? For what? This was my initial reaction when I saw this field.

artembilan: Right. That's why we always encourage to add JavaDocs to this kind of configuration properties.

spring-cloud-gcp-docs/src/main/asciidoc/pubsub.adoc
```diff
@@ -95,6 +97,13 @@ to be configured.
 
 `PubSubTemplate` uses a special subscriber generated by its `SubscriberFactory` to pull messages.
 
+If the message payload contains a serialized POJO, it can be retrieved as a `Class` compatible with that serialized payload:
```
meltsufin: Maybe it's just me, but in the absence of additional context, the word "serialization" means the Java built-in serialization. So, to avoid confusion, I would instead use "JSON-encoded" or "JSON-serialized".

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
```diff
@@ -40,39 +43,65 @@
...
+			SubscriberFactory subscriberFactory, String[] trustedPackages) {
 		this.publisherFactory = publisherFactory;
 		this.subscriberFactory = subscriberFactory;
+		this.objectMapper = trustedPackages == null || trustedPackages.length == 0
+				? new ObjectMapper()
```
meltsufin: I think the creation of the `ObjectMapper` should be completely contained in the auto-config class. You shouldn't need to call `new` here or pass in the `trustedPackages`.

spring-cloud-gcp-docs/src/main/asciidoc/pubsub.adoc
```diff
@@ -95,6 +97,13 @@ to be configured.
...
+
+[source,java]
+----
+this.pubSubTemplate.getPayloadFromMessage(message, MyPojo.class);
```
meltsufin: I believe the class information / default typing is not needed, if you are passing in the type.

ChengyuanZhao: @meltsufin I tried getting the reader mapper without providing the class param, but it would not work. Jackson complains that it couldn't create its reader mapper object.

I infer from their  documentation (https://github.com/FasterXML/jackson-docs/wiki/JacksonPolymorphicDeserialization#11-global-default-typing) that this type information is not for the purpose of omitting this class param (which we can do if we did Java serialization), but for purposes of using super-types with serialized children.

spring-cloud-gcp-docs/src/main/asciidoc/pubsub.adoc
ChengyuanZhao: I didn't think it was appropriate to reveal to the user specifically how we serialize their objects. Therefore I  didn't want to mention JSON/not-java/binary serialization specifically. I think "serialized" has just the connotation of sending an object over the network and not specifically Java Serializable serialization.

meltsufin: It's important because the receiving end might not be the same application / code base. So, it will need to deal with deserialization.

spring-cloud-gcp-docs/src/main/asciidoc/pubsub.adoc
meltsufin: Is this supposed to work: `this.pubSubTemplate.getPayloadFromMessage(message, Object.class)`?
@artembilan mentioned that he will share some more details about how he has done this in the latest project.

artembilan: No, I promised to do that for the `HeaderMapper` PR #658 .

I will come back to you with review some time later. Don't know when though: the pool of tasks never exhausts...

meltsufin: Sorry, my mistake. :)

spring-cloud-gcp-docs/src/main/asciidoc/pubsub.adoc
artembilan: Not only from docs: I wouldn't stick just only with JSON at all.
Therefore it would be pretty confusing to me to do some other configuration, but see a JSON-specific exception in the contract.

spring-cloud-gcp-docs/src/main/asciidoc/pubsub.adoc
artembilan: Right, @meltsufin , that should work:
See source code of the `GenericJackson2JsonRedisSerializer` where borrowed an idea about default typing:
```
public <T> T deserialize(@Nullable byte[] source, Class<T> type) throws SerializationException {

	Assert.notNull(type,
			"Deserialization type must not be null! Pleaes provide Object.class to make use of Jackson2 default typing.");

	if (SerializationUtils.isEmpty(source)) {
		return null;
	}

	try {
		return mapper.readValue(source, type);
	} catch (Exception ex) {
		throw new SerializationException("Could not read JSON: " + ex.getMessage(), ex);
	}
}
```
https://github.com/spring-projects/spring-data-redis/blob/master/src/main/java/org/springframework/data/redis/serializer/GenericJackson2JsonRedisSerializer.java

spring-cloud-gcp-docs/src/main/asciidoc/pubsub.adoc
```diff
@@ -267,4 +276,7 @@ Google Cloud Pub/Sub API, if different from the ones in the
 | `spring.cloud.gcp.pubsub.credentials.scopes` |
 https://developers.google.com/identity/protocols/googlescopes[OAuth2 scope] for Spring Cloud GCP
 Pub/Sub credentials | No | https://www.googleapis.com/auth/pubsub
+| `spring.cloud.gcp.pubsub.deserializationTrustedPackages[n]` |
```
artembilan: This sounds like very long configuration property.
The `trustedPackages` is pretty self-explanatory, plus the JavaDocs on the property should do the trick to determine the purpose.

spring-cloud-gcp-pubsub/pom.xml
```diff
@@ -30,6 +30,10 @@
 			<artifactId>spring-integration-core</artifactId>
 			<optional>true</optional>
 		</dependency>
+		<dependency>
+			<groupId>com.fasterxml.jackson.core</groupId>
+			<artifactId>jackson-databind</artifactId>
```
artembilan: I guess `optional` as well?

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubOperations.java
```diff
@@ -75,6 +78,17 @@
 	 */
 	ListenableFuture<String> publish(String topic, ByteString payload, Map<String, String> headers);
 
+	/**
+	 * Send a message to Pub/Sub.
+	 * @param topic the name of an existing topic
+	 * @param payload an object that will be serialized and sent
+	 * @param headers map of String to String headers
+	 * @return the listenable future of the call
+	 * @throws JsonProcessingException when the payload cannot be serialized into JSON
```
artembilan: No this exception. You can't be sure that we always use only JSON encoding/decoding.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubOperations.java
artembilan: This doesn't look like a part of the `PubSubOperations` contract...

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
ChengyuanZhao: Could you elaborate on that? I think the logic of "if no trusted packages are specified , then don't include type information in JSON" is very specific to this specific class.

spring-cloud-gcp-docs/src/main/asciidoc/pubsub.adoc
ChengyuanZhao: I don't have an opinion on this one. At the moment there is both the long name and the javadoc comment. @meltsufin If you don't have a problem with the shorter name I can change it back.

spring-cloud-gcp-docs/src/main/asciidoc/pubsub.adoc
ChengyuanZhao: done.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubOperations.java
ChengyuanZhao: done.

spring-cloud-gcp-docs/src/main/asciidoc/pubsub.adoc
ChengyuanZhao: done.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubOperations.java
ChengyuanZhao: removed.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
artembilan: ??? Why don't always include type information? 
It is very convenient way do not worry about type on the other side, especially when you deal with container classes, like `List` or `Message`.

To trust or not trust is already an end-user choice and his/her security team.
However we can have an opinion on the matter and always be secured providing some default set of packages. For example in Spring Integration it is:
```
private static final List<String> TRUSTED_PACKAGES =
			Arrays.asList(
					"java.util",
					"java.lang",
					"org.springframework.messaging.support",
					"org.springframework.integration.support",
					"org.springframework.integration.message",
					"org.springframework.integration.store"
			);
```

To *trust all* we just require an asterisk (`*`)":
```
if (trustedPackages != null) {
		for (String whiteListClass : trustedPackages) {
			if ("*".equals(whiteListClass)) {
				this.trustedPackages.clear();
				break;
			}
			else {
				this.trustedPackages.add(whiteListClass);
			}
		}
}
```

So, this way you will never use `ObjectMapper` without `trustedPackages` setting. However I would still defer the choice to end user and would let him/her to inject any possible `ObjectMapper`, but that's already a `MessageConverter` (or `PayloadConverter`) responsibility.

ChengyuanZhao: If the user doesn't provide any trusted packages, then is that supposed to mean "trust all" ?

spring-cloud-gcp-docs/src/main/asciidoc/pubsub.adoc
meltsufin: Shorted name is ok by me.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
artembilan: In Spring Kafka we use `"java.util", "java.lang"` as default trusted packages.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
```diff
@@ -40,39 +43,65 @@
...
  * @author Vinicius Carvalho
  * @author João André Martins
  * @author Mike Eltsufin
+ * @author Chengyuan Zhao
  */
 public class PubSubTemplate implements PubSubOperations, InitializingBean {
 
 	private static final Log LOGGER = LogFactory.getLog(PubSubTemplate.class);
 
+	private final ObjectMapper objectMapper;
```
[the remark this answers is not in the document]
ChengyuanZhao: By this do you mean I should create an interface called `MessageConverter`, and then have an implementation called `JacksonMessageConverter`? The latter would then hold the `ObjectMapper` ?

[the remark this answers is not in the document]
artembilan: Exactly! This way I will be able to create a `Protobuf` or `SchemaRegistry` or just plain old `JavaSerialization` converter and inject it into this `PubSubTemplate`.
I even think that it make sense really call it as a `PubSubMessageConverter` with contract like `<T> T fromMessage(PubSubMessage m, Class<T> payloadType)` and `PubSubMessage toMessage(Object payload)`. See Spring AMQP on the matter: https://github.com/spring-projects/spring-amqp/blob/master/spring-amqp/src/main/java/org/springframework/amqp/support/converter/MessageConverter.java

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubProperties.java
```diff
@@ -43,6 +47,12 @@
 	 */
 	private String emulatorHost;
 
+	/**
+	 * List of packages containing types that are whitelisted for deserialization from message
+	 * payloads.
+	 */
+	private List<String> trustedPackages = Collections.emptyList();
```
meltsufin: `String[]` maybe, since that's what you need to pass to Jackson.

spring-cloud-gcp-pubsub/pom.xml
ChengyuanZhao: @meltsufin what would you prefer? make it non-optional again?

meltsufin: optional is fine, we just need to document that the dependency is needed if you want to take advantage of object to json conversion.

ChengyuanZhao: done

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubProperties.java
ChengyuanZhao: done

spring-cloud-gcp-docs/src/main/asciidoc/pubsub.adoc
```diff
@@ -75,6 +74,18 @@ public void publishMessage() {
 }
 ----
 
+
+The default serialization and deserialization of POJOs uses Jackson JSON via the `JacksonMessageConverter` class.
+However, the Jackson dependency is required to utilize that implementation:
```
artembilan: That's something wrong.
People are going to argue that we require Jackson to use with the `PubSubTemplate`, but a dependency on the matter is not included.
Or let's remove `optional` from there, or don't instantiate Jackson converter in the `PubSubTemplate`. The dependency between our classes eliminates an optional nature of the external dependency.

meltsufin: Ah, that's right, required is fine then.

artembilan: We can come back to `optional` later, when we will be able to introduce some other converter implementation. However we can't regain it to the `optional` until the next major `2.0` version though :-)

ChengyuanZhao: @meltsufin done, PTAL

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
```diff
@@ -40,39 +41,68 @@
...
+	}
+
+	public PubSubMessageConverter getMessageConverter() {
+		return this.messageConverter;
+	}
+
+	/**
+	 * {@link PubSubTemplate} constructor that uses a {@link JacksonMessageConverter} to
+	 * serialize and deserialize payloads.
+	 *
```
artembilan: No blank lines in the method JavaDocs: https://github.com/spring-projects/spring-framework/wiki/Code-Style#javadoc-formatting

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/converter/PubSubMessageConverter.java
```diff
@@ -0,0 +1,44 @@
...
+ * @author Chengyuan Zhao
+ */
+public interface PubSubMessageConverter {
+
+	/**
+	 * Convert a Java object to a byte array payload.
+	 * @param object the object to convert
+	 * @return a byte array ready to be sent with PubSub
+	 */
+	byte[] toPayload(Object object) throws IOException;
```
artembilan: I think this one has to be aligned with other `MessageConverter`s. Also for the simetry this one  has to deal with the `PubsubMessage`:

```
PubsubMessage toMessage(Object payload, Map headers);
```

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/converter/JacksonMessageConverter.java
```diff
@@ -0,0 +1,61 @@
...
+import com.google.pubsub.v1.PubsubMessage;
+
+import org.springframework.integration.support.json.JacksonJsonUtils;
+
+/**
+ * A message converter using Jackson JSON.
+ *
+ * @author Chengyuan Zhao
+ */
+public class JacksonMessageConverter implements PubSubMessageConverter {
```
artembilan: This is a bit wide name. To avoid conflicts with existing imple I'd rename this to the `JacksonPubSubMessageConverter`

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/converter/JacksonMessageConverter.java
```diff
@@ -0,0 +1,61 @@
...
+
+	private final ObjectMapper objectMapper;
+
+	/**
+	 * Constructor
+	 * @param trustedPackages the packages trusted for deserialization in addition to the
+	 * default trusted packages listed in {@link JacksonJsonUtils}.
+	 */
+	public JacksonMessageConverter(String... trustedPackages) {
```
artembilan: There must be a separate ctor which allows end-user to inject his/her own `ObjectMapper`.
Seems for me we had such a discussion in other place in this PR...

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
```diff
@@ -40,39 +41,68 @@
...
  * @author Vinicius Carvalho
  * @author João André Martins
  * @author Mike Eltsufin
+ * @author Chengyuan Zhao
  */
 public class PubSubTemplate implements PubSubOperations, InitializingBean {
 
 	private static final Log LOGGER = LogFactory.getLog(PubSubTemplate.class);
 
+	private final PubSubMessageConverter messageConverter;
```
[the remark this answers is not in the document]
ChengyuanZhao: done

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
ChengyuanZhao: done

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/converter/PubSubMessageConverter.java
ChengyuanZhao: done

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/converter/JacksonMessageConverter.java
ChengyuanZhao: done

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/converter/JacksonMessageConverter.java
ChengyuanZhao: done

artembilan: Please, don't send us this kind of spam emails with `done` word.

The proper commit message is enough for me to understand that you agree for the changes request.

That makes sense to comment when you have some other opinion.

I have thousands emails from hundred repos on GH I follow per day, so extra noise in my mail box decrease my efficiency :smile:

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
```diff
@@ -40,41 +41,58 @@
...
 import org.springframework.cloud.gcp.pubsub.support.PublisherFactory;
 import org.springframework.cloud.gcp.pubsub.support.SubscriberFactory;
+import org.springframework.cloud.gcp.pubsub.support.converter.JacksonPubSubMessageConverter;
+import org.springframework.cloud.gcp.pubsub.support.converter.PubSubMessageConverter;
 import org.springframework.util.Assert;
 import org.springframework.util.concurrent.ListenableFuture;
 import org.springframework.util.concurrent.SettableListenableFuture;
 
 /**
+ * // @formatter:off
```
artembilan: Why do we need this?

Seems for me I fully don't use formatting for JavaDocs in my IDEA - to much inconsistency with required code style.
I prefer to reformat it manually.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/converter/JacksonPubSubMessageConverter.java
```diff
@@ -0,0 +1,71 @@
...
+
+	private final ObjectMapper objectMapper;
+
+	/**
+	 * Constructor
+	 * @param trustedPackages the packages trusted for deserialization in addition to the
+	 * default trusted packages listed in {@link JacksonJsonUtils}.
+	 */
+	public JacksonPubSubMessageConverter(String... trustedPackages) {
+		this.objectMapper = JacksonJsonUtils.messagingAwareMapper(trustedPackages);
```
artembilan: One more time: are we agreed to have Spring Integration as a `required` dependency?
Seems for me the story was to have `PubSubTemplate` independent of Spring Integration.

With the required Jackson dependency for the `PubSubTemplate` defaults, we end up with required Spring Integration because of this "fancy" `messagingAwareMapper()` utility. 

Is that really expected?

Thanks

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/converter/JacksonMessageConverter.java
ChengyuanZhao: My bad. I won't do that anymore. We do that internally to let you know that your comments addressed.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/converter/PubSubMessageConverter.java
```diff
@@ -0,0 +1,62 @@
...
+		return pubsubMessageBuilder.build();
+	}
+
+	/**
+	 * Convert the payload of a given PubSub message to a desired Java type.
+	 * @param message the message containing the payload of the object
+	 * @param payloadType the desired type of the object
+	 * @return the object converted from the message's payload
+	 */
+	<T> T fromMessage(PubsubMessage message, Class<T> payloadType) throws IOException;
```
artembilan: Looks like we can have a `default` impl here as well!
```
default <T> T fromMessage(PubsubMessage message, Class<T> expectedType) throws IOException {
    return fromPayload(message.getData(), expectedType);
}
```

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
ChengyuanZhao: My auto-formatter (using the xml style file in this repo) will move that `<p>` to another line in certain situations (including this one), which fails the checkstyle.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/converter/JacksonPubSubMessageConverter.java
ChengyuanZhao: I am with having it required. @meltsufin said he was fine with it being required, too.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
artembilan: Well, then we have to fix that auto-formatter, because having `<p>` on the same like as fist sentence of the paragraph is indeed our code style: https://github.com/spring-projects/spring-framework/wiki/Code-Style#javadoc-formatting.

Let's just try do not use this IDE-specific `@formatter:off`!

PR #661, opened 2018-05-07
Storage protocol resolver cache beans

### LINKED ISSUE #606: Cache protocol resolver settings for Storage
as mentioned in #574

ChengyuanZhao:
fixes #606

### DIFF
--- a/spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageProtocolResolver.java
+++ b/spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageProtocolResolver.java
@@ -47,6 +47,10 @@ public class GoogleStorageProtocolResolver
 
 	private ConfigurableListableBeanFactory beanFactory;
 
+	private Storage storage;
+
+	private GoogleStorageProtocolResolverSettings googleStorageProtocolResolverSettings;
+
 	GoogleStorageProtocolResolver() {
 	}
 
@@ -54,6 +58,21 @@ public class GoogleStorageProtocolResolver
 	public void postProcessBeanFactory(
 			ConfigurableListableBeanFactory beanFactory) throws BeansException {
 		this.beanFactory = beanFactory;
+	}
+
+	private Storage getStorage() {
+		if (this.storage == null) {
+			this.storage = this.beanFactory.getBean(Storage.class);
+		}
+		return this.storage;
+	}
+
+	private GoogleStorageProtocolResolverSettings getGoogleStorageProtocolResolverSettings() {
+		if (this.googleStorageProtocolResolverSettings == null) {
+			this.googleStorageProtocolResolverSettings = this.beanFactory
+					.getBean(GoogleStorageProtocolResolverSettings.class);
+		}
+		return this.googleStorageProtocolResolverSettings;
 	}
 
 	@Override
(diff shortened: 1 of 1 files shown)


spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageProtocolResolver.java
```diff
@@ -56,6 +60,21 @@ public void postProcessBeanFactory(
 		this.beanFactory = beanFactory;
 	}
 
+	private Storage getStorage() {
```
joaoandremartins: A getter that changes the state of the object?

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageProtocolResolver.java
```diff
@@ -56,6 +60,21 @@ public void postProcessBeanFactory(
...
+		if (this.storage == null) {
+			this.storage = this.beanFactory.getBean(Storage.class);
+		}
+		return this.storage;
+	}
+
+	private GoogleStorageProtocolResolverSettings getGoogleStorageProtocolResolverSettings() {
```
joaoandremartins: Ditto.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageProtocolResolver.java
```diff
@@ -69,7 +88,7 @@ public void setResourceLoader(ResourceLoader resourceLoader) {
 
 	private GoogleStorageProtocolResolverSettings getSettings() {
 		try {
-			return this.beanFactory.getBean(GoogleStorageProtocolResolverSettings.class);
+			return getGoogleStorageProtocolResolverSettings();
```
joaoandremartins: Why not inlining those getters here?
e.g.,
```
if (this.googleStorageProtocolResolverSettings == null) {
this.googleStorageProtocolResolverSettings = this.beanFactory.getBean(GoogleStorageProtocolResolverSettings.class);
}
return this.googleStorageProtocolResolverSettings;
```

Even better, do it in `postProcessBeanFactory()`, after `this.beanFactory = beanFactory`.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageProtocolResolver.java
artembilan: This is `private` method, so harmless

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageProtocolResolver.java
ChengyuanZhao: I had that originally , but it appears things are not all ready to be used by that point. Specifically the GoogleProtocolResolverSettings isn't correct to be retrieved from the beanFactory at that point and unit tests fail.

artembilan: We can't do that in the `postProcessBeanFactory()`. That is going to be *evil* called `early BeanFactory` access.
That's why we resolve these bean lazy-load manner.

PR #667, opened 2018-05-11
Spanner embeddable classes with Ids

### LINKED ISSUE #588: Embedded Id support
Provide support for user defined ID classes to capture composite keys.
This would be supported beyond the existing `@PrimaryKey` annotation as an alternative way to map composite keys. Similar to [`@EmbeddedId` ](https://www.objectdb.com/java/jpa/entity/id#Embedded_Primary_Key_) annotation in JPA

ChengyuanZhao:
fixes #588


ChengyuanZhao:
@meltsufin PTAL

ChengyuanZhao:
@meltsufin @dmitry-s PTAL

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/ConverterAwareMappingSpannerEntityWriter.java
```diff
@@ -141,11 +140,20 @@ public void write(Object source, MultipleValueBinder sink,
...
-						return;
+					if (spannerPersistentProperty.isEmbedded()) {
+						Object embeddedObject = accessor
+								.getProperty(spannerPersistentProperty);
+						if (embeddedObject != null) {
+							write(embeddedObject, sink, includeColumns);
+						}
+					}
+					else {
+						if (!writeAllColumns && !includeColumns
```
dmitry-s: Please refactor this block to make positive checks instead of negative, to improve readability. This construct
```
if (!a && !b){
    return;
}
code 
```
can be expressed in this way:
```
if (a || b){
    code
} 
```
Also, after this modification it would probably makes sense to merge that `if` with previous `else`.

spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/core/admin/SpannerSchemaUtilsTests.java
```diff
@@ -64,10 +65,10 @@ public void getDropDDLTest() {
 
 	@Test
 	public void getCreateDDLTest() {
-		assertEquals("CREATE TABLE custom_test_table ( id STRING(MAX) , id2 INT64 "
-				+ ", custom_col STRING(MAX) , other STRING(333) , bytes BYTES(MAX) "
-				+ ", bytesList ARRAY<BYTES(111)> , integerList ARRAY<INT64> "
-				+ ", doubles ARRAY<FLOAT64> ) PRIMARY KEY ( id , id2 )",
+		assertEquals("CREATE TABLE custom_test_table ( id STRING(MAX) , id3 INT64 , "
+						+ "id2 STRING(MAX) , bytes2 BYTES(MAX) , custom_col STRING(MAX) , other STRING(333) ,"
```
dmitry-s: id3 has (keyOrder = 3), while id2 has (keyOrder = 2)
why does id2 appears after id3?

spring-cloud-gcp-docs/src/main/asciidoc/spanner.adoc
```diff
@@ -257,6 +257,67 @@ Column naming is generated by the `PropertyNameFieldNamingStrategy` by default d
...
+    id STRING(MAX),
+    value STRING(MAX),
+    child_value STRING(MAX),
+    age INT64
+) PRIMARY KEY (grandParentId, parentId, id)
+----
+
+Note that embedded properties' column names must all be unique.
+
+
```
dmitry-s: Please also improve spanner-sample and AbstractSpannerIntegrationTest to include embedded properties.

spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/core/convert/ConverterAwareMappingSpannerEntityWriterTest.java
```diff
@@ -98,6 +99,11 @@ public void writeTest() {
 		t.bytesList = new ArrayList<>();
 		t.bytesList.add(t.bytes);
 
+		t.testEmbeddedColumns = new TestEmbeddedColumns();
```
dmitry-s: All test classes should be named \*Test**s** for consistency

spring-cloud-gcp-docs/src/main/asciidoc/spanner.adoc
ChengyuanZhao: We actively did not choose to add complex functionality to the samples to not overwhelm somebody checking this out, so I don't think we want to include it in there. 

As for the integration test, it is already tested in integration tests. Take a look at the Trade.java entity type we use for the integration tests, and it has an embedded property.

spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/core/convert/ConverterAwareMappingSpannerEntityWriterTest.java
ChengyuanZhao: will fix.

spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/core/admin/SpannerSchemaUtilsTests.java
ChengyuanZhao: I wanted to write the TestEntity class with the keys out of order to make sure that we're not depending on that written order, but only on the key order. 

As a side effect in the SQL it comes out in the same reversed order. However, what actually matters is that final line where the order does matter, and it is correct there: `PRIMARY KEY ( id , id2 , id3 )`

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/ConverterAwareMappingSpannerEntityWriter.java
ChengyuanZhao: will fix!

spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/core/convert/ConverterAwareMappingSpannerEntityWriterTest.java
ChengyuanZhao: changes made, PTAL

spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/core/admin/SpannerSchemaUtilsTests.java
dmitry-s: ok I see

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/admin/SpannerSchemaUtils.java
```diff
@@ -143,23 +143,50 @@ public Key getKey(Object object) {
...
+			}
+			else {
+				keyStrings.add(keyProp.getColumnName());
+			}
 		}
+	}
 
-		stringBuilder.append(keyStrings.toString()).append(" )");
-		return stringBuilder.toString();
+	private <T> void addColumnDDLStrings(
```
meltsufin: Generally, we should camel-case acronyms as well, which would make it `addColumnDdlStrings`.

ChengyuanZhao: Will fix. To be fair nearly all of the functionality we prioritized we haven't received any feedback from the community that would indicate adoption or not :/ 

However, embeddable classes had been discussed internally for a while, and was seen as a somewhat well-known feature from JPA. Furthermore, I think it fits nicely into the plan to support parent-child relationships where the PK components would need to be repeated in parent and child, and where them being embedded would be very convenient.

ChengyuanZhao:
@dmitry-s any further changes?

Issue #668, opened 2018-05-14
Pubsub emulator configuration is not working correctly


Andy2003:
With the changes of https://github.com/spring-cloud/spring-cloud-gcp/pull/484 the simple use of the emulator host is no longer working.

PR #669, opened 2018-05-14
#668 improve emulator host configuration

Andy2003:


spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
```diff
@@ -79,9 +81,19 @@ public GcpPubSubAutoConfiguration(GcpPubSubProperties gcpPubSubProperties,
 		this.finalProjectIdProvider = gcpPubSubProperties.getProjectId() != null
 				? gcpPubSubProperties::getProjectId
 				: gcpProjectIdProvider;
-		this.finalCredentialsProvider = gcpPubSubProperties.getCredentials().hasKey()
-				? new DefaultCredentialsProvider(gcpPubSubProperties)
-				: credentialsProvider;
+
+		if (StringUtils.isEmpty(gcpPubSubProperties.getEmulatorHost())
+				|| "false".equals(gcpPubSubProperties.getEmulatorHost())) {
```
joaoandremartins: What does this second condition test? How can the `emulatorHost` property be `false`?

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubEmulatorConfiguration.java
```diff
@@ -36,7 +36,7 @@
  * @author Andreas Berger
  */
 @Configuration
-@ConditionalOnProperty(prefix = "spring.cloud.gcp.pubsub", name = "emulatorHost")
+@ConditionalOnProperty(prefix = "spring.cloud.gcp.pubsub", name = "emulator-host")
```
joaoandremartins: Could you please make the change everywhere else in documentation, etc.?

joaoandremartins:
Also, why is it so important to have a `NoCredentialsProvider` for the emulator? It'll still work if you send it actual credentials.
Could you describe your use case a bit, so I can understand better? Thanks!

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
Andy2003: `@ConditionalOnProperty(prefix = "spring.cloud.gcp.pubsub", name = "emulator-host")` matches also if I set `spring.cloud.gcp.pubsub.emulator-host=` (to empty)

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
```diff
@@ -79,9 +81,19 @@ public GcpPubSubAutoConfiguration(GcpPubSubProperties gcpPubSubProperties,
...
+			this.finalCredentialsProvider = gcpPubSubProperties.getCredentials().hasKey()
+					? new DefaultCredentialsProvider(gcpPubSubProperties)
+					: credentialsProvider;
+		}
+		else {
+			// Since we cannot create a general NoCredentialsProvider if the emulator host is enabled
+			// (because it would also be used for the other components), we have to create one here
+			// for this particular case.
+			this.finalCredentialsProvider = NoCredentialsProvider.create();
```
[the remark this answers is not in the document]
Andy2003: Yeah, I think so too. But right now, there's no better way to do this. In case you have any better idea: I am open for every hint.

Andy2003:
@joaoandremartins Not all developers of our project have necessarily stored real access data for pubsub locally. This results in an exception if you start the spring application, even though you have an emulator set up and running locally.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
joaoandremartins: Apologies if I'm missing something, but isn't that what the first condition tests?

Andy2003: yes, if I set `spring.cloud.gcp.pubsub.emulator-host=`
* `@ConditionalOnProperty(prefix = "spring.cloud.gcp.pubsub", name = "emulator-host")` will match and
* `StringUtils.isEmpty(gcpPubSubProperties.getEmulatorHost())` will match

if I set `spring.cloud.gcp.pubsub.emulator-host=false`
* `@ConditionalOnProperty(prefix = "spring.cloud.gcp.pubsub", name = "emulator-host")` will not match and
* `"false".equals(gcpPubSubProperties.getEmulatorHost())` will match

So a simple `null` check and the test to `flase` will match `@ConditionalOnProperty(prefix = "spring.cloud.gcp.pubsub", name = "emulator-host")` better, I will adjust this.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
```diff
@@ -79,9 +80,18 @@ public GcpPubSubAutoConfiguration(GcpPubSubProperties gcpPubSubProperties,
 		this.finalProjectIdProvider = gcpPubSubProperties.getProjectId() != null
 				? gcpPubSubProperties::getProjectId
 				: gcpProjectIdProvider;
-		this.finalCredentialsProvider = gcpPubSubProperties.getCredentials().hasKey()
-				? new DefaultCredentialsProvider(gcpPubSubProperties)
-				: credentialsProvider;
+
+		if (gcpPubSubProperties.getEmulatorHost() == null || "false".equals(gcpPubSubProperties.getEmulatorHost())) {
```
artembilan: I guess `!"true".equals(gcpPubSubProperties.getEmulatorHost())` is much better for us.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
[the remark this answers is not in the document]
artembilan: I don't understand where is the concern for now. 
If you really need a `NoCredentialsProvider` instance for the emulator, then it is fine and really explained in the comments above.

If you have some other ideas (which might not be related to this issue), then would be great to do that separately.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubEmulatorConfiguration.java
artembilan: That's good catch, indeed. However you still can use it as `spring.cloud.gcp.pubsub.emulatorHost`. Therefore might be no need to change docs everywhere else.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
Andy2003: Since the property refers to the emulator host, I cannot check for `!true` here. The check for `false` is the only way to override an already set emulator-host, or is there a better way? All other issues has already been addressed.

artembilan: Oh! I got it.
I think I wound't have any comments here if the story would about a separate `enableEmulator` property.
Although that's just about testing, so I'm fine if you have document that to disable it you need to specify `false` for the `emulatorHost` property. A bit awkward, but let it be if that is what you have agreed here with others...

artembilan:
> This branch cannot be rebased due to conflicts

@Andy2003 ,

please, address conflicts with the current `master`.

Andy2003:
@joaoandremartins Is there anything else that needs to be done? If not, I'd be happy about a merge.

joaoandremartins:
Go for it!

joaoandremartins:
@artembilan that might be some caching issue, I was able to merge the PR just fine.

PR #670, opened 2018-05-14
Improves lack of core credentials message

joaoandremartins:
Currently, if no core credentials are set, but either the app doesn't
need them or they're set at the service level, an error message is
printed with a stack trace.

This leads users to believe there is an error with the app, when things
are actually fine. Only saying no core credentials are set is fine,
things will blow up when a service actually needs credentials but
doesn't have them.

### DIFF
--- a/spring-cloud-gcp-core/src/main/java/org/springframework/cloud/gcp/core/DefaultCredentialsProvider.java
+++ b/spring-cloud-gcp-core/src/main/java/org/springframework/cloud/gcp/core/DefaultCredentialsProvider.java
@@ -125,7 +125,9 @@ public class DefaultCredentialsProvider implements CredentialsProvider {
 			}
 		}
 		catch (IOException ioe) {
-			LOGGER.error("No core credentials are set.");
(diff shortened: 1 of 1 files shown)


spring-cloud-gcp-core/src/main/java/org/springframework/cloud/gcp/core/DefaultCredentialsProvider.java
```diff
@@ -125,7 +125,7 @@ else if (credentials instanceof ComputeEngineCredentials) {
 			}
 		}
 		catch (IOException ioe) {
-			LOGGER.error("No credentials were found.", ioe);
+			LOGGER.error("No core credentials are set.");
```
artembilan: Won't it help to extend this message to say what is going to happen then?
For example mention those credentials in the target module.

I mean that always was good experience for me working with Spring that it has all those good explanations in the logs when we have this or that warning.

Actually why is this `error`, but not `warn` if this isn't going to be fatal for the app?

PR #671, opened 2018-05-14
Documents SI Pub/Sub header mapping

joaoandremartins:
Also addresses some JavaDoc issues.

Removes the redundant `pub_sub` prefix on the acknowledgement header.

### DIFF
--- a/spring-cloud-gcp-docs/src/main/asciidoc/spring-integration-pubsub.adoc
+++ b/spring-cloud-gcp-docs/src/main/asciidoc/spring-integration-pubsub.adoc
@@ -169,4 +169,4 @@ NOTE: The order in which the patterns are declared in `PubSubHeaderMapper.setOut
 The first patterns have precedence over the following ones.
 
 In the previous example, the `"*"` pattern means every header is mapped.
-However, because it comes last in the list, the previous patterns take precedence.
+However, because it comes last in the list, https://docs.spring.io/spring-integration/api/org/springframework/integration/util/PatternMatchUtils.html#smartMatch-java.lang.String-java.lang.String...-[the previous patterns take precedence].


spring-cloud-gcp-docs/src/main/asciidoc/spring-integration-pubsub.adoc
```diff
@@ -144,3 +144,12 @@ public void handleMessage(Message<?> msg) throws MessagingException {
 ----
 
 It is also possible to set an SpEL expression for the topic with the `setTopicExpression()` or `setTopicExpressionString()` methods.
+
+==== Header mapping
```
artembilan: Would be great to mention somehow in this paragraph our `PubSubHeaderMapper` implementation.
Otherwise it isn't clear how end-user can configure a desired set of headers.
Some code snippet would be great addition too, as well as a sample project modification on the matter.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/PubSubHeaderMapper.java
```diff
@@ -30,16 +30,17 @@
...
+ * the {@link org.springframework.messaging.Message} to {@link com.google.pubsub.v1.PubsubMessage}
  * header conversion.
  *
  * @author João André Martins
  */
 public class PubSubHeaderMapper implements HeaderMapper<Map<String, String>> {
 
 	/**
-	 * Patterns of headers to map in {@link #fromHeaders(MessageHeaders, Map)}.
+	 * Patterns of headers to map in {@link #fromHeaders(MessageHeaders, Map)}. First headers take
```
artembilan: Would be better to start a new sentence from a new line.
Also it is really not about *headers*. This is about *patterns*. You can configure it like `gcp_*` and all headers with the `gcp_` prefix will be mapped.

spring-cloud-gcp-docs/src/main/asciidoc/spring-integration-pubsub.adoc
```diff
@@ -153,3 +153,20 @@ The outbound channel adapter maps every header from Spring messages into Google
...
+PubSubHeaderMapper headerMapper = new PubSubHeaderMapper();
+headerMapper.setOutboundHeaderPatterns("!foo", "!bar", "!prefix_*", "*");
+adapter.setHeaderMapper(headerMapper);
+----
+
+NOTE: The order in which the patterns are declared in `PubSubHeaderMapper.setOutboundHeaderPatterns()` and `PubSubHeaderMapper.setInboundHeaderPatterns()` matters.
+The first patterns have precedence over the following ones.
+
+In the previous example, the `"*"` pattern means every header is mapped.
+However, because it comes last in the list, the previous patterns take precedence.
```
artembilan: You may send end-user from here to JavaDocs of the `org.springframework.integration.support.utils.PatternMatchUtils.smartMatch()`.

PR #672, opened 2018-05-14
Makes the SI Pub/Sub IT use the real service

joaoandremartins:
The emulator is no longer used in these tests.

### DIFF
--- a/spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/pubsub/it/PubSubChannelAdaptersIntegrationTests.java
+++ b/spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/pubsub/it/PubSubChannelAdaptersIntegrationTests.java
@@ -132,18 +132,18 @@ public class PubSubChannelAdaptersIntegrationTests {
 
 			PollableChannel channel = context.getBean("outputChannel", PollableChannel.class);
 
-			Message<?> message = channel.receive(5000);
+			Message<?> message = channel.receive(10000);
 			assertThat(message).isNotNull();
 			AckReplyConsumer acker =
 					(AckReplyConsumer) message.getHeaders().get(GcpPubSubHeaders.ACKNOWLEDGEMENT);
 			assertThat(acker).isNotNull();
 			acker.nack();
-			message = channel.receive(2000);
+			message = channel.receive(10000);
 			assertThat(message).isNotNull();
 			acker = (AckReplyConsumer) message.getHeaders().get(GcpPubSubHeaders.ACKNOWLEDGEMENT);
 			assertThat(acker).isNotNull();
 			acker.ack();
-			message = channel.receive(1000);
+			message = channel.receive(10000);
 			assertThat(message).isNull();
 		});
 	}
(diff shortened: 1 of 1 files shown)


spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/pubsub/it/PubSubChannelAdaptersIntegrationTests.java
```diff
@@ -69,19 +68,16 @@
...
-					GcpPubSubEmulatorConfiguration.class,
+					GcpContextAutoConfiguration.class,
 					GcpPubSubAutoConfiguration.class))
 			.withUserConfiguration(
-					PubSubChannelAdaptersIntegrationTests.IntegrationConfiguration.class)
-			.withPropertyValues("spring.cloud.gcp.pubsub.emulatorHost=${PUBSUB_EMULATOR_HOST}");
+					PubSubChannelAdaptersIntegrationTests.IntegrationConfiguration.class);
 
 	@BeforeClass
 	public static void checkEmulatorIsRunning() {
```
ChengyuanZhao: This method is still named `...EmulatorIsRunning`

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/pubsub/it/PubSubChannelAdaptersIntegrationTests.java
```diff
@@ -204,28 +200,20 @@ public MessageChannel outputChannel() {
...
-
 		@Bean
 		public SubscriberFactory defaultSubscriberFactory(
 				@Qualifier("subscriberExecutorProvider") ExecutorProvider executorProvider,
 				TransportChannelProvider transportChannelProvider,
-				PubSubAdmin pubSubAdmin) {
+				PubSubAdmin pubSubAdmin,
+				GcpProjectIdProvider projectIdProvider,
+				CredentialsProvider credentialsProvider) {
 			if (pubSubAdmin.getSubscription("doralice") == null) {
```
ChengyuanZhao: Could it become a problem with these constant names like we did with the pubsub IT tests when several were running (from multiple PRs). Perhaps append a random bit at the end? (topic and subscription)

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/pubsub/it/PubSubChannelAdaptersIntegrationTests.java
```diff
@@ -141,7 +138,7 @@ public void sendAndReceiveMessageManualAck() {
 					(AckReplyConsumer) message.getHeaders().get(GcpPubSubHeaders.ACKNOWLEDGEMENT);
 			assertThat(acker).isNotNull();
 			acker.nack();
-			message = channel.receive(1000);
+			message = channel.receive(2000);
```
artembilan: To honor slowness of the Travis container and those CI servers, in other project we agreed to use a 10 seconds to wait for a message from the channel.

PR #675, opened 2018-05-16
Boilerplate datastore base classes

### LINKED ISSUE #676: Datastore Persistent Entities, Properties, Mapping Context, Annotations
- Classes are annotated with `@Entity(name = “kind_name”)`. This is because in Datastore documentation the “Kind” is analogous to a table in SQL, and an `entity` is analogous to a row in a table. 
Note that many different classes can be annotated with the same Kind name, because DS allows different schemas for entities of the same Kind. 

- The Spring Data `@Id` annotation is optional. If used on a property, then that property is used as the key in Datastore AND also stored as a field. 
If no properties are annotated with it, then DS will auto-generate an ID.

- Optional single property of type `Iterable<KeyValue<String, Object>>` can be annotated `@Ancestors`. 

- This property will not be stored as a field. It will be used as the ancestors of this entity, where the String component is that ancestor’s Kind, and the Object component is the specific key of that ancestor. 
POJO properties:
1. can be annotated `@Reference`, which means that POJO must also be of a type annotated with `@Kind`, and then the DS Key of that entity is stored in this field in DS. (the Key type is a data type that a field can have in DS).
2. can be annotated `@Embedded`, which means they will be stored as an embedded entity in DS. This is also a data type that a field can have in DS. Embedded entities do not need to have a Key or Kind, and are stored in their entirety in that field in DS.
3. Both of these also apply to Iterable properties where the inner type is a POJO, because DS can store any singular field type as elements of an array column type.

ChengyuanZhao:
fixes #676 

These are mostly boilerplate implementations of Spring Data interfaces that all spring data implementations do similarly.


spring-cloud-gcp-data-datastore/src/test/java/org/springframework/cloud/gcp/data/datastore/core/mapping/DatastoreMappingContextTests.java
```diff
@@ -0,0 +1,70 @@
...
+		DatastorePersistentEntityImpl mockEntity = mock(
+				DatastorePersistentEntityImpl.class);
+		DatastoreMappingContext context = createSpannerMappingContextWith(mockEntity);
+
+		context.createPersistentEntity(ClassTypeInformation.from(Object.class));
+
+		verifyZeroInteractions(mockEntity);
+	}
+
+	private DatastoreMappingContext createSpannerMappingContextWith(
```
dmitry-s: did you really mean "spanner"?

ChengyuanZhao: Ah yes fixed!

meltsufin:
@ChengyuanZhao I think we need to take another look at the things that this PR introduced, like the `@Ancestors`. In general, I don't think this PR really got the thorough review from everyone that it deserved. Let's schedule a meeting to take another look at this PR.

ChengyuanZhao:
@meltsufin sure we can meet about this one. Just want to note that the `@Ancestors` annotation is consistent with what we most recently discussed with the Spring Data team.

PR #689, opened 2018-05-18
Adds publish() without headers

### LINKED ISSUE #686: Add publish(topic, payload) to PubSubOperations
We should add a `publish(topic, payload)` convenience method to `PubSubOperations`.

Should we also add different flavours of payload, like `String`, `byte[]` and `ByteString`, or just `String`?

joaoandremartins:
Users wanting to publish without headers would have had to call
publish(topic, payload, null), which isn't very idiomatic.

Fixes #686

### DIFF
--- a/spring-cloud-gcp-docs/src/main/asciidoc/pubsub.adoc
+++ b/spring-cloud-gcp-docs/src/main/asciidoc/pubsub.adoc
@@ -46,8 +46,8 @@ utilization.
 ==== Publishing to a topic
 
 `PubSubTemplate` provides asynchronous methods to publish messages to a Google Cloud Pub/Sub topic.
-It supports different types of payloads, including `Strings` with different encodings, `byte[]`,
-`ByteString`, POJOs, and `PubsubMessage`.
+It supports different types of payloads, including `Strings` with different encodings, `byte[]`, `ByteString`, POJOs, and `PubsubMessage`.
+For each type of payload, there are equivalent `publish()` methods without headers.
 
 [source,java]
 ----
--- a/spring-cloud-gcp-docs/src/main/asciidoc/pubsub.adoc
+++ b/spring-cloud-gcp-docs/src/main/asciidoc/pubsub.adoc
@@ -46,24 +46,13 @@ utilization.
 ==== Publishing to a topic
 
 `PubSubTemplate` provides asynchronous methods to publish messages to a Google Cloud Pub/Sub topic.
-It supports different types of payloads, including `Strings` with different encodings, `byte[]`, `ByteString`, POJOs, and `PubsubMessage`.
-For each type of payload, there are equivalent `publish()` methods without headers.
+The `publish()` method takes in a topic name to post the message to, a payload of a generic type and, optionally, a map with the message headers.
 
-[source,java]
-----
-public ListenableFuture<String> publish(String topic, String payload, Map<String, String> headers)
-
-public ListenableFuture<String> publish(String topic, String payload, Map<String, String> headers,
-Charset charset)
-
-public ListenableFuture<String> publish(String topic, byte[] payload, Map<String, String> headers)
-
-public ListenableFuture<String> publish(String topic, ByteString payload, Map<String, String> headers)
-
-public ListenableFuture<String> publish(String topic, PubsubMessage pubsubMessage)
-
-public <T> ListenableFuture<String> publish(String topic, T payload, Map<String, String> headers)
-----
+* If the payload is a string, it is decoded using the active character set.
+** A character set can be set using the `setCharset()` method in `PubSubTemplate`.
+* If the payload is a `ByteString`, it is sent as-is.
+* If the payload is a `byte[]`, then a `ByteString` is created from it.
+* Otherwise, the payload is serialized into JSON and sent.
 
 Here is an example of how to publish a message to a Google Cloud Pub/Sub topic:
 
@@ -73,7 +62,6 @@ public void publishMessage() {
     this.pubSubTemplate.publish("topic", "your message payload", ImmutableMap.of("key1", "val1"));
 }
 ----
-
 
 The default serialization and deserialization of POJOs uses Jackson JSON via the `JacksonMessageConverter` class.
 
--- a/spring-cloud-gcp-docs/src/main/asciidoc/pubsub.adoc
+++ b/spring-cloud-gcp-docs/src/main/asciidoc/pubsub.adoc
@@ -52,7 +52,7 @@ The `publish()` method takes in a topic name to post the message to, a payload o
 ** A character set can be set using the `setCharset()` method in `PubSubTemplate`.
 * If the payload is a `ByteString`, it is sent as-is.
 * If the payload is a `byte[]`, then a `ByteString` is created from it.
-* Otherwise, the payload is serialized into JSON and sent.
+* Otherwise, the payload is serialized and sent.
 
 Here is an example of how to publish a message to a Google Cloud Pub/Sub topic:
 
@@ -64,6 +64,7 @@ public void publishMessage() {
 ----
 
 The default serialization and deserialization of POJOs uses Jackson JSON via the `JacksonMessageConverter` class.
+To set a new message converter in `PubSubTemplate`, use the `setMessageConverter()` method.
 
 
 ==== Subscribing to a subscription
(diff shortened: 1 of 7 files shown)


spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubOperations.java
```diff
@@ -81,12 +81,52 @@
...
+	 */
+	ListenableFuture<String> publish(String topic, ByteString payload);
+
+	/**
+	 * Send a message to Pub/Sub.
+	 * @param topic the name of an existing topic
+	 * @param payload an object that will be serialized and sent
+	 * @return the listenable future of the call
+	 */
+	<T> ListenableFuture<String> publish(String topic, T payload) throws IOException;
```
artembilan: Why do we need all those overloaded methods, when we have just this one and all the hard conversion logic according the provided payload type can be encapsulated in the implementation details?

Otherwise this is pretty confusing API when I try to use it as auto-completion feature in the IDE.

joaoandremartins: Did that, but kept the publish with string and charset, because there's no elegant way to collapse that one into the new generic one.
Let me know if you disagree.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubOperations.java
```diff
@@ -81,12 +81,28 @@
...
 			Map<String, String> headers) throws IOException;
 
+	/**
+	 * Send a message to Pub/Sub.
+	 * @param topic the name of an existing topic
+	 * @param payload the message String payload
+	 * @param charset charset to decode the payload
+	 * @return the listenable future of the call
+	 */
+	ListenableFuture<String> publish(String topic, String payload, Charset charset);
```
artembilan: Well, you can have a configurable `Charset` on the template for common use-cases. And if you really need a logic based on the different charsets for different messages, you just need to perform a simple `payload.getBytes(charset)` operation before calling template. 

I just fully don't see reason to overhead template API with such a simple logic just to treat Strings some other way, where a difference is only in a small charset conversion step.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
```diff
@@ -133,6 +133,31 @@ public PubSubTemplate setMessageConverter(PubSubMessageConverter messageConverte
...
+			result = publish(topic, payload, null);
+		}
+		else if (payload instanceof byte[]) {
+			result = publish(topic, payload, null);
+		}
+		else if (payload instanceof ByteString) {
+			result = publish(topic, payload, null);
+		}
+		else {
+			result = publish(topic, payload, null);
```
artembilan: Why didn't you fix the ambiguity for the `publish()` with headers?

I mean why don't collapse them to a single generic one as well?

spring-cloud-gcp-docs/src/main/asciidoc/pubsub.adoc
```diff
@@ -46,24 +46,13 @@ utilization.
...
-public ListenableFuture<String> publish(String topic, byte[] payload, Map<String, String> headers)
-
-public ListenableFuture<String> publish(String topic, ByteString payload, Map<String, String> headers)
-
-public ListenableFuture<String> publish(String topic, PubsubMessage pubsubMessage)
-
-public <T> ListenableFuture<String> publish(String topic, T payload, Map<String, String> headers)
-----
+* If the payload is a string, it is decoded using the active character set.
+** A character set can be set using the `setCharset()` method in `PubSubTemplate`.
```
artembilan: Why double asterisk ?

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubOperations.java
```diff
@@ -42,50 +40,19 @@
...
-	 * Send a message to Pub/Sub.
-	 * @param topic the name of an existing topic
-	 * @param payload the message payload on the {@link PubsubMessage} payload format
-	 * @param headers map of String to String headers
+	 * @param payload an object that will be serialized and sent
 	 * @return the listenable future of the call
 	 */
-	ListenableFuture<String> publish(String topic, ByteString payload, Map<String, String> headers);
+	<T> ListenableFuture<String> publish(String topic, T payload,
+			Map<String, String> headers) throws IOException;
```
artembilan: Can't we don't expose this exception as a public contract in this method?
May we wrap into the `UncheckedIOException` whenever an `IOException` is thrown in the target implementation?

We can mention `UncheckedIOException` in the JavaDocs though for better behavior understanding.

What worries me here that we not always get an `IOException` in the impl, therefore no reason to enforce end-user to always handle it.

Does it make sense to you?
Thanks

spring-cloud-gcp-docs/src/main/asciidoc/pubsub.adoc
joaoandremartins: To make it a sub-item of the first one.
It will be a little bit indented and the bullet also looks different.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
```diff
@@ -132,15 +133,20 @@ else if (payload instanceof byte[]) {
...
 		}
 		else {
-			pubsubMessageBuilder.setData(
-					ByteString.copyFrom(this.messageConverter.toPayload(payload)));
+			try {
+				pubsubMessageBuilder.setData(
+						ByteString.copyFrom(this.messageConverter.toPayload(payload)));
+			}
+			catch (IOException ioe) {
+				throw new UncheckedIOException("Error serializing payload into JSON. ", ioe);
```
artembilan: `into JSON`?

Are you sure that `messageConverter` is always about JSON?

Can't I inject any other possible `MessageConverter` impl even if JSON is a default one?

spring-cloud-gcp-docs/src/main/asciidoc/pubsub.adoc
```diff
@@ -46,24 +46,13 @@ utilization.
...
+* If the payload is a `ByteString`, it is sent as-is.
+* If the payload is a `byte[]`, then a `ByteString` is created from it.
+* Otherwise, the payload is serialized into JSON and sent.
```
artembilan: > Omits JSON in exception message

Good, but why haven't you fixed a JSON sticking here as well?

Am I missing anything?

spring-cloud-gcp-pubsub/src/test/java/org/springframework/cloud/gcp/pubsub/integration/outbound/PubSubMessageHandlerTests.java
```diff
@@ -16,6 +16,7 @@
 
 package org.springframework.cloud.gcp.pubsub.integration.outbound;
 
+import java.io.IOException;
```
artembilan: Redundant import?

spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-sample/src/main/java/com/example/WebController.java
```diff
@@ -58,8 +60,8 @@ public RedirectView createSubscription(@RequestParam("topicName") String topicNa
 
 	@GetMapping("/postMessage")
 	public RedirectView publish(@RequestParam("message") String message,
-			@RequestParam("topicName") String topicName) {
-		this.pubSubTemplate.publish(topicName, message, null);
+			@RequestParam("topicName") String topicName) throws IOException {
```
artembilan: Seems for me we have just talked that there is no `IOException` requirement any more. 
Let's not overcomplicate samples!

PR #694, opened 2018-05-23
Spanner queries ignore case support

### LINKED ISSUE #691: Add support for Sort.Order.ignoreCase()
Currently, when the sorting options are being validated and the `ORDER BY` clause is being constructed in the `SpannerTemplate ` class, the` Sort.Order.ignoreCase() ` flag is not being consulted, therefore, the `ORDER BY` clause is not quite correct and the returned data set is not sorted as expected.

These are reference links in the spring data repo about this issue:

- [Jira ticket in the SpringData project](https://jira.spring.io/browse/DATAJPA-296)
- [SpringData commit 1](https://github.com/kraymondksc/spring-data-jpa/commit/c3adce0bd36799d3754a5f4c61aee64982abd7e0)
- [SpringData commit 2](https://github.com/spring-projects/spring-data-jpa/commit/4c65ad1df6ea281036b5d35c25473a57f0bc3ac0)

ChengyuanZhao:
fixes #691

### DIFF
--- a/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
+++ b/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
@@ -49,7 +49,6 @@ import org.springframework.cloud.gcp.data.spanner.core.convert.SpannerEntityProc
 import org.springframework.cloud.gcp.data.spanner.core.mapping.SpannerMappingContext;
 import org.springframework.cloud.gcp.data.spanner.core.mapping.SpannerPersistentEntity;
 import org.springframework.cloud.gcp.data.spanner.repository.query.SpannerStatementQueryExecutor;
-import org.springframework.data.domain.Sort;
 import org.springframework.util.Assert;
 
 /**
@@ -170,7 +169,8 @@ public class SpannerTemplate implements SpannerOperations {
 
 	public String applySortingPagingQueryOptions(SpannerQueryOptions options,
 			String sql) {
-		StringBuilder sb = applySort(options.getSort(), wrapAsSubSelect(sql));
+		StringBuilder sb = SpannerStatementQueryExecutor.applySort(options.getSort(),
+				wrapAsSubSelect(sql), o -> o.getProperty());
 			if (options.hasLimit()) {
 				sb.append(" LIMIT ").append(options.getLimit());
 			}
@@ -178,17 +178,6 @@ public class SpannerTemplate implements SpannerOperations {
 				sb.append(" OFFSET ").append(options.getOffset());
 			}
 		return sb.toString();
-	}
-
-	private StringBuilder applySort(Sort sort, StringBuilder sql) {
-		if (sort == null || sort.isUnsorted()) {
-			return sql;
-		}
-		sql.append(" ORDER BY ");
-		StringJoiner sj = new StringJoiner(" , ");
-		sort.iterator().forEachRemaining(
-				o -> sj.add(o.getProperty() + (o.isAscending() ? " ASC" : " DESC")));
-		return sql.append(sj);
 	}
 
 	private StringBuilder wrapAsSubSelect(String sql) {
--- a/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
+++ b/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
@@ -68,7 +68,8 @@ public class SpannerTemplate implements SpannerOperations {
 	private final SpannerMutationFactory mutationFactory;
 
 	public SpannerTemplate(DatabaseClient databaseClient,
-			SpannerMappingContext mappingContext, SpannerEntityProcessor spannerEntityProcessor,
+			SpannerMappingContext mappingContext,
+			SpannerEntityProcessor spannerEntityProcessor,
 			SpannerMutationFactory spannerMutationFactory) {
 		Assert.notNull(databaseClient,
 				"A valid database client for Spanner is required.");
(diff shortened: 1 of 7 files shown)


spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
```diff
@@ -170,7 +169,8 @@ public SpannerEntityProcessor getSpannerEntityProcessor() {
 
 	public String applySortingPagingQueryOptions(SpannerQueryOptions options,
 			String sql) {
-		StringBuilder sb = applySort(options.getSort(), wrapAsSubSelect(sql));
+		StringBuilder sb = SpannerStatementQueryExecutor.applySort(options.getSort(),
+				wrapAsSubSelect(sql), o -> o.getProperty());
 			if (options.hasLimit()) {
```
meltsufin: Wrong indentation.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryExecutor.java
```diff
@@ -167,19 +168,30 @@ private static void buildWhere(PartTree tree, SpannerPersistentEntity<?> persist
...
 					String tag = "tag" + tags.size();
 					tags.add(tag);
 					String andString = persistentEntity.getPersistentProperty(segment)
 							.getColumnName();
+					String sortedTag = "@" + tag;
+					if (part.shouldIgnoreCase() == IgnoreCaseType.ALWAYS) {
+						andString = "LOWER(" + andString + ")";
+						sortedTag = "LOWER(" + sortedTag + ")";
+					}
+					else if (part.shouldIgnoreCase() == IgnoreCaseType.WHEN_POSSIBLE) {
```
meltsufin: If you would like the condition to be true to the actual exception message, this should be 
`else if (part.shouldIgnoreCase() != IgnoreCaseType.NEVER)`

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryExecutor.java
```diff
@@ -167,19 +168,30 @@ private static void buildWhere(PartTree tree, SpannerPersistentEntity<?> persist
 				StringJoiner andStrings = new StringJoiner(" AND ");
 
 				orPart.forEach(part -> {
-
 					String segment = part.getProperty().getSegment();
 					String tag = "tag" + tags.size();
 					tags.add(tag);
 					String andString = persistentEntity.getPersistentProperty(segment)
 							.getColumnName();
+					String sortedTag = "@" + tag;
```
meltsufin: Why are you calling this `sortedTag`? How is this related to sorting?

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryExecutor.java
```diff
@@ -135,23 +139,20 @@ private static void buildFrom(SpannerPersistentEntity<?> persistentEntity,
...
-				order.getProperty()).getColumnName() + " " + toString(order);
-	}
-
-	private static String toString(Order order) {
-		return order.isAscending() ? "ASC" : "DESC";
+		sql.append(" ORDER BY ");
+		StringJoiner sj = new StringJoiner(" , ");
+		sort.iterator().forEachRemaining(o -> {
+			String sortedPropertyName = sortedPropertyNameFunction.apply(o);
+			String sortedProperty = o.isIgnoreCase() ? "LOWER(" + sortedPropertyName + ")"
```
meltsufin: What if the `sortedPropertyName` not a String, but a something like a number? Wouldn't passing it to `LOWER()` cause a failure?

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryExecutor.java
ChengyuanZhao: I will rename

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
ChengyuanZhao: I don't think so, I tried moving this to the previous line and the auto format moves it back to being like this.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryExecutor.java
ChengyuanZhao: Yes will change for cosmetic reasons. However there are only 3 options in this enum and it's logically impossible to add another one. (always, never, sometimes cover all possible cases).

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryExecutor.java
meltsufin: OK, got it.

meltsufin:
@ChengyuanZhao I think this is good now. I would just like to see some update to the refdoc mentioning that we support IngoreCase, but not the `WHEN_POSSIBLE` option. Thanks!

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
```diff
@@ -343,12 +332,10 @@ private void logReadOptions(SpannerReadOptions options, StringBuilder logs) {
...
 
-	private StringBuilder logColumns(String tableName, KeySet keys, Iterable<String> columns) {
-		StringBuilder logSb = new StringBuilder("Executing read on table "
-				+ tableName
-				+ " with keys: "
-				+ keys
-				+ " and columns: ");
+	private StringBuilder logColumns(String tableName, KeySet keys,
+			Iterable<String> columns) {
+		StringBuilder logSb = new StringBuilder("Executing read on table " + tableName
```
meltsufin: `logSb` should only be constructed if the `DEBUG` log level is on.

ChengyuanZhao: the intention is that this `private` function is only to be used when DEBUG is on. Indeed, the only usage of it is inside such an if-statement.

meltsufin: Thanks, I missed that context. :)

ansspage:
Ами супер

PR #709, opened 2018-05-30
Documents non-ack pull()

joaoandremartins:


spring-cloud-gcp-docs/src/main/asciidoc/pubsub.adoc
```diff
@@ -80,15 +77,19 @@ The Spring Boot starter for Google Cloud Pub/Sub auto-configures a `SubscriberFa
...
-Google Cloud Pub/Sub supports the synchronous pulling of messages from a subscription.
+Google Cloud Pub/Sub supports synchronous pulling of messages from a subscription.
 This is different from subscribing to a subscription, in the sense that subscribing is an
 asynchronous task which polls the subscription on a set interval.
 
-The `pullNext()` method allows for a single message to be pulled from a subscription.
-The `pull()` method pulls a number of messages from a subscription, allowing for the retry settings
-to be configured.
+The `pullNext()` method allows for a single message to be pulled and automatically acknowledged from a subscription.
+The `pull()` method pulls a number of messages from a subscription, allowing for the retry settings to be configured.
```
meltsufin: I think it's worth saying here that `pull()` does not auto-ack, but returns a list of `AcknowledgeablePubsubMessage`s

spring-cloud-gcp-docs/src/main/asciidoc/pubsub.adoc
```diff
@@ -33,15 +33,12 @@ It provides the common set of operations needed to interact with Google Cloud Pu
...
 to interact with Google Cloud Pub/Sub.
 
-`PubSubTemplate` depends on a `PublisherFactory`, which is a functional interface to provide a
-Google Cloud Java Client for Pub/Sub `Publisher`.
-The Spring Boot starter for GCP Pub/Sub auto-configures a `PublisherFactory` with default settings
-and uses the `GcpProjectIdProvider` and `CredentialsProvider` auto-configured by the Spring Boot GCP
-starter.
+`PubSubTemplate` depends on a `PublisherFactory` and a `SubscriberFactory`.
+The `PublisherFactory` provides a Google Cloud Java Client for Pub/Sub `Publisher`.
+The `SubscriberFactory` provides the `Subscriber` for asynchronous message pulling, as well as a `SubscriberStub` for synchronous pulling and an `Acknowledger`, to automatically acknowledge incoming messages if a user so desires.
```
meltsufin: how does the `Acknowledger` automatically ack incoming messages? Isn't the idea that you have to invoke it explicitly?

meltsufin: I guess the part where you say "if a user so desires" is confusing me.

meltsufin: I think I would just remove the part of the sentence: ", to automatically acknowledge incoming messages if a user so desires.".

Issue #270, opened 2017-12-15
Copy/pasted documention in READMEs


snicoll:
Part of the reference guide is copy/pasted in the README. This makes it quite hard to be consistent. The PR I've recently submitted makes the readme of `spring-cloud-starter-config` inconsistent.

meltsufin:
What would you suggest as the solution to this?

snicoll:
The readme can have a link to ref guide section if you want to keep it.

joaoandremartins:
I think @snicoll means fixes like [this](https://github.com/spring-cloud/spring-cloud-gcp/pull/268/files#diff-53772f03cf390c50569d1bd513a3a0fdR35).

I can sweep our READMEs next week to find similar issues.

PR #712, opened 2018-05-31
consolidate sql-starter's documentation and remove a broken link

dmitry-s:
Part of #270


spring-cloud-gcp-docs/src/main/asciidoc/spring-stream.adoc
```diff
@@ -23,10 +23,6 @@ dependencies {
 }
 ----
 
-=== Getting started
-
-There is a link:../spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-binder-sample[code sample] that demonstrates how to build a simple app to send messages over a Spring Cloud Stream Binder for Google Cloud Pub/Sub.
```
joaoandremartins: Hmm why removing this? The link doesn't work like this, but it would still be possible to link to the sample by using `https://github.com/spring-cloud/spring-cloud-gcp/tree/master/spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-binder-sample[code sample]`

dmitry-s: It would point to a snapshot version of the sample which uses a different version than documentation describes.

joaoandremartins: It would point to the GitHub repository at master, which is fine.
The only issue I see is if we would get rid of the sample, or relocate it. Having a deadlink is a small price to pay for having this information in the reference document.
Any chance we can add it back?

dmitry-s: @meltsufin what do you think?

meltsufin: As I understand, the link was already broken. If we want to link to samples in our refdoc, we should do it consistently for all features.
@joaoandremartins Why do you think it's so important to link to the sample and risk having it out of date with the refdoc release version? Is the binder documentation in the refdoc insufficient on its own?

joaoandremartins: It was broken because I was trying to use an internal link from the refdoc to a GH README, which obviously doesn't make sense. It could have been fixed just by pointing to the sample with a full URL.
Agreed it should be done for all features. Since this one is already in, I don't see why we should take it out to bring it back in later.
I've seen quick starts with samples as a Spring refdoc practice. [Example](http://cloud.spring.io/spring-cloud-static/Finchley.RC2/single/spring-cloud.html#_quick_start)
If by "sample version", you mean the version of the dependencies it uses, then wouldn't it always be `1.0.0.BUILD-SNAPSHOT`? I don't think it makes sense to correlate refdoc with samples, since samples aren't released with the rest of the code.

meltsufin: Right, the samples may be out of sync with the refdoc. Hence, my concern about linking to them from the refdoc.

joaoandremartins: If anything, old refdocs will be pointing to version of the samples that use more recent code.
In that unlikely case, the user will know they should update to the newer version, and we're all good.

meltsufin: Let's continue the discussion here: #715

Issue #363, opened 2018-01-29
Add SubscriberConfiguration and PublisherConfiguration


meltsufin:
Spun off from #92.

@saturnism Please clarify what you meant by this.

joaoandremartins:
I think the requirement here was to add all the little configurable bits from the Pub/Sub classes to a `@ConfigurationProperties` class that we would have to map manually.

I think the agreement was to abandon this idea and instead add any settings on demand, but @saturnism can let me know if I'm wrong.

meltsufin:
Currently we autoconfigure `DefaultPublisherFactory` and `DefaultSubscriberFactory`. However, there are many different properties that can be set for each, which we are not setting. The idea is to expose more properties to be used in the autoconfiguration of these beans.

meltsufin:
Retry settings come to mind. Maybe let's get a list and try to go from there?

ChengyuanZhao:
- DefaultPublisherFactory
- - RetrySettings
- - - setTotalTimeout
- - - setInitialRetryDelay
- - - setRetryDelayMultiplier
- - - setMaxRetryDelay
- - - setMaxAttempts
- - - setJittered
- - - setInitialRpcTimeout
- - - setInitialRpcTimeout
- - - setInitialRpcTimeout
- - - setInitialRpcTimeout

ChengyuanZhao:
@saturnism PTAL

ChengyuanZhao:
Would it make more sense to just have the user provide their own beans for many of these settings instead of us putting them all the way through as spring boot config settings?

meltsufin:
It looks like if we do flow control settings, retry settings, and batching settings, we got most of everything covered. Wherever a whole implementation is needed, we can skip those.

ChengyuanZhao:
Which retry settings and flow control settings ? (Several of the sub-settings use those types)

This is part of the reason I am for having the user provide the instances of these objects instead of exposing all of the repeated underlying numbers/strings as properties settings.

ChengyuanZhao:
@meltsufin What do you think? Instead of exposing so many individual values in properties files we can make bean functions that take in the entire retry setting and flow control setting objects for those various settings and rely on name-matching (with  `@Qualifier`) instead of type-matching (since there are several params of the same type).

meltsufin:
[...]

PR #719, opened 2018-06-01
pubsub factory direct settings and compound-setting beans

ChengyuanZhao:
discussed in #363

This PR adds the documentation and properties for settings that are not the compound setting objects (retry , batch, and flow control settings). Beans are added for these compound settings.

Exposing those beans as properties will come in the next PR.

### DIFF
--- a/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
+++ b/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
@@ -19,17 +19,22 @@ package org.springframework.cloud.gcp.autoconfigure.pubsub;
 import java.io.IOException;
 import java.util.concurrent.Executors;
 
+import com.google.api.core.ApiClock;
+import com.google.api.gax.batching.BatchingSettings;
+import com.google.api.gax.batching.FlowControlSettings;
 import com.google.api.gax.core.CredentialsProvider;
 import com.google.api.gax.core.ExecutorProvider;
 import com.google.api.gax.core.FixedExecutorProvider;
 import com.google.api.gax.core.NoCredentialsProvider;
 import com.google.api.gax.grpc.InstantiatingGrpcChannelProvider;
+import com.google.api.gax.retrying.RetrySettings;
 import com.google.api.gax.rpc.HeaderProvider;
 import com.google.api.gax.rpc.TransportChannelProvider;
 import com.google.cloud.pubsub.v1.SubscriptionAdminClient;
 import com.google.cloud.pubsub.v1.SubscriptionAdminSettings;
 import com.google.cloud.pubsub.v1.TopicAdminClient;
 import com.google.cloud.pubsub.v1.TopicAdminSettings;
+import org.threeten.bp.Duration;
 
 import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.boot.autoconfigure.AutoConfigureAfter;
@@ -52,6 +57,7 @@ import org.springframework.cloud.gcp.pubsub.support.converter.JacksonPubSubMessa
 import org.springframework.cloud.gcp.pubsub.support.converter.PubSubMessageConverter;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
+import org.springframework.lang.Nullable;
 
 /**
  * @author João André Martins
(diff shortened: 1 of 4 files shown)


ChengyuanZhao:
@meltsufin PTAL

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubProperties.java
```diff
@@ -50,6 +52,47 @@
 	 */
 	private String[] trustedPackages;
 
+	/**
+	 * The optional pull endpoint setting for the subscriber factory.
+	 */
+	private String subscriberPullEndpoint;
```
meltsufin: Would it make sense to aggregate all of the `subscriber*` properties into its own class?

ChengyuanZhao: Yes can do. I would prefer to make this an inner class named "Subscriber" rather than a separate class and then annotated with `@NestedConfigurationProperty`. Lemme know where I'm wrong, but as far as I know: the reason is we would prefer to use spring.cloud.gcp.pubsub.subscriber.blah, which would mean we would need a Subscriber.java class, which sounds like a misleading name for a java file and class if all it held were some settings. Therefore, I think it makes sense to have an inner class called GcpPubSubProperties.Subscriber. @meltsufin what do you think about that

artembilan: That's what we do in other places in Spring Boot: 

https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/KafkaProperties.java

https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/RabbitProperties.java

Although you don't need `@NestedConfigurationProperty` is the class is really nested. See those samples.

meltsufin: Inner class should be fine.

ChengyuanZhao: @meltsufin I also moved the publisher setting to its own class. PTAL

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
```diff
@@ -124,30 +131,107 @@ public PubSubTemplate pubSubTemplate(PublisherFactory publisherFactory,
 				.setMessageConverter(pubSubMessageConverter);
 	}
 
+	@Bean
+	@ConditionalOnMissingBean(name = "subscriberSystemExecutorProvider")
+	public ExecutorProvider subscriberSystemExecutorProvider() {
+		return null;
```
joaoandremartins: Let's not provide `null` beans. If we don't have any opinion, let's not provide any beans at all.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubProperties.java
```diff
@@ -29,11 +31,18 @@
  */
 @ConfigurationProperties("spring.cloud.gcp.pubsub")
 public class GcpPubSubProperties implements CredentialsSupplier {
-	/** Number of threads used by every {@link com.google.cloud.pubsub.v1.Subscriber}. */
-	private int subscriberExecutorThreads = 4;
 
-	/** Number of threads used by every {@link com.google.cloud.pubsub.v1.Publisher}. */
-	private int publisherExecutorThreads = 4;
+	private final Subscriber subscriber = new Subscriber();
```
joaoandremartins: If we go down this route, I think you'll need `@NestedConfigurationProperties` here and below for `Publisher` too.

spring-cloud-gcp-docs/src/main/asciidoc/pubsub.adoc
```diff
@@ -261,7 +261,7 @@ The Spring Boot starter for Google Cloud Pub/Sub provides the following configur
 | `spring.cloud.gcp.pubsub.enabled` | Enables or disables Pub/Sub auto-configuration | No | `true`
 | `spring.cloud.gcp.pubsub.subscriber-executor-threads` | Number of threads used by `Subscriber`
```
joaoandremartins: Missed `subscriber.executor-threads`.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
ChengyuanZhao: I didn't know these didn't need to be specified if they were nullable! Will remove.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubProperties.java
ChengyuanZhao: Mike and Artem are for the inner class , though.

spring-cloud-gcp-docs/src/main/asciidoc/pubsub.adoc
ChengyuanZhao: ah will fix.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubProperties.java
artembilan: We don't need a `@NestedConfigurationProperties` if options classes are inner.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
```diff
@@ -128,26 +135,67 @@ public PubSubTemplate pubSubTemplate(PublisherFactory publisherFactory,
 	@ConditionalOnMissingBean
 	public SubscriberFactory defaultSubscriberFactory(
 			@Qualifier("subscriberExecutorProvider") ExecutorProvider executorProvider,
+			@Qualifier("subscriberSystemExecutorProvider")
+			@Nullable ExecutorProvider systemExecutorProvider,
+			@Qualifier("subscriberFlowControlSettings") @Nullable FlowControlSettings flowControlSettings,
```
joaoandremartins: Can we get rid of these `@Nullable` and maybe replace them by `@Autowired(required = false)` and Mike suggested?

meltsufin: What's the difference?

joaoandremartins: I would say more Spring idiomatic?

artembilan: No `@Autowired` in the auto-configurations, please.
There is something like `ObjectProvider` in Spring Boot to consider instead.

artembilan:
> Merged

I guess my comment has been just missed: https://github.com/spring-cloud/spring-cloud-gcp/pull/719#discussion_r192790620

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
meltsufin: @artembilan what's the best practice in this situation? Is `@Nullable` commonly used?
@ChengyuanZhao If so, can we do a post-merge fix?

ChengyuanZhao:
Would changing the Autowired back to Nullable be acceptable? I can make that change in the next PR that also exposes the retry , batch, and flow control properties

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
artembilan: The practice is to use an `ObjectProvider<?>`.
See here fore example: https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jdbc/DataSourceJmxConfiguration.java#L58

ChengyuanZhao: I can make the change to use the objectprovider in the next PR.

artembilan: Correct, because this one is merged already.

PR #723, opened 2018-06-04
consolidate spanner docs

dmitry-s:
Part of #270


spring-cloud-gcp-docs/src/main/asciidoc/spanner.adoc
```diff
@@ -29,7 +29,7 @@ dependencies {
 ----
 
 This setup takes care of bringing in the latest compatible version of Cloud Java Cloud Spanner libraries
-as well.
+as well. It doesn't provide any bean to be used by an end-user.
```
meltsufin: I think this comment is really strange and confusing. It's not even true either. For example, `SpannerTemplate` bean is provided that can definitely be used by the end-user.

meltsufin: I would just leave it out.

PR #724, opened 2018-06-04
Pubsub factory properties

### LINKED ISSUE #363: Add SubscriberConfiguration and PublisherConfiguration

ChengyuanZhao:
fixes #363

### DIFF
--- a/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
+++ b/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
@@ -28,6 +28,7 @@ import com.google.api.gax.core.FixedExecutorProvider;
 import com.google.api.gax.core.NoCredentialsProvider;
 import com.google.api.gax.grpc.InstantiatingGrpcChannelProvider;
 import com.google.api.gax.retrying.RetrySettings;
+import com.google.api.gax.retrying.RetrySettings.Builder;
 import com.google.api.gax.rpc.HeaderProvider;
 import com.google.api.gax.rpc.TransportChannelProvider;
 import com.google.cloud.pubsub.v1.SubscriptionAdminClient;
@@ -36,7 +37,7 @@ import com.google.cloud.pubsub.v1.TopicAdminClient;
 import com.google.cloud.pubsub.v1.TopicAdminSettings;
 import org.threeten.bp.Duration;
 
-import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.ObjectProvider;
 import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.boot.autoconfigure.AutoConfigureAfter;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
@@ -132,45 +133,54 @@ public class GcpPubSubAutoConfiguration {
 	}
 
 	@Bean
+	@ConditionalOnMissingBean(name = "subscriberRetrySettings")
+	public RetrySettings subscriberRetrySettings() {
+		return buildRetrySettings(this.gcpPubSubProperties.getSubscriber().getRetry());
+	}
+
+	@Bean
+	@ConditionalOnMissingBean(name = "subscriberFlowControlSettings")
+	public FlowControlSettings subscriberFlowControlSettings() {
+		return buildFlowControlSettings(
+				this.gcpPubSubProperties.getSubscriber().getFlowControl());
+	}
+
+	private FlowControlSettings buildFlowControlSettings(
+			GcpPubSubProperties.FlowControl flowControl) {
+		FlowControlSettings.Builder builder = FlowControlSettings.newBuilder();
+		flowControl.getLimitExceededBehavior()
+				.ifPresent(builder::setLimitExceededBehavior);
+		flowControl.getMaxOutstandingElementCount()
+				.ifPresent(builder::setMaxOutstandingElementCount);
+		flowControl.getMaxOutstandingRequestBytes()
+				.ifPresent(builder::setMaxOutstandingRequestBytes);
+		return builder.build();
+	}
+
+	@Bean
 	@ConditionalOnMissingBean
 	public SubscriberFactory defaultSubscriberFactory(
 			@Qualifier("subscriberExecutorProvider") ExecutorProvider executorProvider,
 			@Qualifier("subscriberSystemExecutorProvider")
-			@Autowired(required = false) ExecutorProvider systemExecutorProvider,
-			@Qualifier("subscriberFlowControlSettings") @Autowired(required = false)
-					FlowControlSettings flowControlSettings,
-			@Qualifier("subscriberApiClock") @Autowired(required = false) ApiClock apiClock,
-			@Qualifier("subscriberRetrySettings") @Autowired(required = false) RetrySettings retrySettings,
+			ObjectProvider<ExecutorProvider> systemExecutorProvider,
+			@Qualifier("subscriberFlowControlSettings")
+					ObjectProvider<FlowControlSettings> flowControlSettings,
+			@Qualifier("subscriberApiClock") ObjectProvider<ApiClock> apiClock,
+			@Qualifier("subscriberRetrySettings") ObjectProvider<RetrySettings> retrySettings,
 			TransportChannelProvider transportChannelProvider) {
 		DefaultSubscriberFactory factory = new DefaultSubscriberFactory(this.finalProjectIdProvider);
 		factory.setExecutorProvider(executorProvider);
 		factory.setCredentialsProvider(this.finalCredentialsProvider);
 		factory.setHeaderProvider(this.headerProvider);
 		factory.setChannelProvider(transportChannelProvider);
-		if (systemExecutorProvider != null) {
-			factory.setSystemExecutorProvider(systemExecutorProvider);
-		}
-		if (flowControlSettings != null) {
-			factory.setFlowControlSettings(flowControlSettings);
-		}
-		if (apiClock != null) {
-			factory.setApiClock(apiClock);
-		}
-		if (retrySettings != null) {
-			factory.setSubscriberStubRetrySettings(retrySettings);
-		}
-		if (this.gcpPubSubProperties.getSubscriber().getMaxAckDurationSeconds()
-				.isPresent()) {
-			factory.setMaxAckDurationPeriod(Duration.ofSeconds(
-					this.gcpPubSubProperties.getSubscriber()
-							.getMaxAckDurationSeconds().get()));
-		}
-		if (this.gcpPubSubProperties.getSubscriber().getParallelPullCount()
-				.isPresent()) {
-			factory.setParallelPullCount(
-					this.gcpPubSubProperties.getSubscriber()
-							.getParallelPullCount().get());
-		}
+		systemExecutorProvider.ifAvailable(factory::setSystemExecutorProvider);
+		flowControlSettings.ifAvailable(factory::setFlowControlSettings);
+		apiClock.ifAvailable(factory::setApiClock);
+		retrySettings.ifAvailable(factory::setSubscriberStubRetrySettings);
+		this.gcpPubSubProperties.getSubscriber().getMaxAckDurationSeconds()
+				.ifPresent(x -> factory.setMaxAckDurationPeriod(Duration.ofSeconds(x)));
+		this.gcpPubSubProperties.getSubscriber().getParallelPullCount()
+				.ifPresent(factory::setParallelPullCount);
 		if (this.gcpPubSubProperties.getSubscriber()
 				.getPullEndpoint() != null) {
 			factory.setPullEndpoint(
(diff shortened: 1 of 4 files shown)


spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubProperties.java
```diff
@@ -129,6 +143,18 @@ public void setExecutorThreads(int executorThreads) {
 		 */
 		private Optional<Integer> parallelPullCount = Optional.empty();
```
artembilan: I'm not sure that this is a good variant for the configuration property.
I mean no `Optional` for them, please.

@snicoll ,
WDYT  here?

Consider to refactor this class properly: https://github.com/spring-projects/spring-framework/wiki/Code-Style#java-source-file-organization

I mean no getters before properties.

Also all the configuration properties must have JavaDocs - for the future metadata for the auto-completion assistant and so on.

ChengyuanZhao: Are you proposing just leaving things as null instead of Optional.empty to tell that the property wasn't set?

artembilan: That's correct. At least this is what we do in other Spring Boot places for simple types.
And that's why I invited @snicoll to judge us: may be this is really legit to use `Optional` since I see that you have tested it in your changes.

snicoll: Thanks Artem. Using `Optional` on properties [is not really recommended](https://stackoverflow.com/questions/29033518/is-it-a-good-practice-to-use-optional-as-an-attribute-in-a-class) which is the main reason we don't support it.

ChengyuanZhao: @snicoll Has something changed since this ticket? https://github.com/spring-projects/spring-boot/issues/5110

It looks like you personally added support for Optional there.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
```diff
@@ -179,25 +192,78 @@ public SubscriberFactory defaultSubscriberFactory(
...
+				.apply(wasSet(retryProperties.getMaxRpcTimeoutSeconds(),
+						x -> builder.setMaxRpcTimeout(Duration.ofSeconds(x)))
+				.apply(wasSet(retryProperties.getRetryDelayMultiplier(), builder::setRetryDelayMultiplier)
+				.apply(wasSet(retryProperties.getTotalTimeoutSeconds(),
+						x -> builder.setTotalTimeout(Duration.ofSeconds(x)))
+				.apply(wasSet(retryProperties.getRpcTimeoutMultiplier(), builder::setRpcTimeoutMultiplier)
+				.apply(false))))))))) ? builder.build() : null;
+	}
+
+	private <T> Function<Boolean, Boolean> wasSet(Optional<T> optional, Consumer<T> consumer) {
```
joaoandremartins: Can you add some Javadoc to this method?

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubProperties.java
meltsufin: I guess support of a feature is not necessarily an endorsement of it. :smile: 
I'm with @snicoll regarding not using it in the class attributes though. Can we go back to `@Nullable` or `required=false`?

ChengyuanZhao: It sounded like you thought it was a good idea enough to confirm to that individual that Spring would be adding support for what he wanted...

Separately, @meltsufin  this isn't related to the Nullable or Autowire(required=false) issue. That was instead covered by the ObjectProvider thing.

snicoll: > It sounded like you thought it was a good idea enough to confirm to that individual that Spring would be adding support for what he wanted...

We can't really enforce our opinions on everything. Using `Optional` on class attributes is a bad idea IMO. Yet, if we support `Optional`, then  we should support that consistently. 

If a property isn't available in the `Environment` it should return `null` and not `Optional.empty()` for consistency with `Environment#getProperty`

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubProperties.java
```diff
@@ -36,14 +36,6 @@
 
 	private final Publisher publisher = new Publisher();
```
artembilan: These properties must have some JavaDoc description as well.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubProperties.java
```diff
@@ -36,14 +36,6 @@
...
 
-	public Subscriber getSubscriber() {
-		return this.subscriber;
-	}
-
-	public Publisher getPublisher() {
-		return this.publisher;
-	}
-
 	/** Overrides the GCP project ID specified in the Core module. */
```
artembilan: Please, consider to refactor JavaDocs in this class for the proper multi-line format.
This one looks ugly and breaks our code style.
Looks like @joaoandremartins didn't fix them all when we had and argument on the matter...

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubProperties.java
```diff
@@ -98,6 +98,24 @@ public void setTrustedPackages(String[] trustedPackages) {
 		 */
 		private int executorThreads = 4;
 
+		/**
+		 * Properties for {@link com.google.api.gax.retrying.RetrySettings}
```
artembilan: The description of the configuration property must be simple, single like text. No any JavaDoc tags there, please: https://docs.spring.io/spring-boot/docs/2.0.2.RELEASE/reference/htmlsingle/#configuration-metadata-property-attributes

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubProperties.java
snicoll: Actually they must not. This will create a sub namespace and not a property (Or am I missing something?)

artembilan: Doh. Sure! We don't need JavaDocs on them, but I guess that won't hurt since it isn't clear what is the point of this property reading source code or JavaDocs of the class.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubProperties.java
```diff
@@ -32,11 +32,15 @@
 @ConfigurationProperties("spring.cloud.gcp.pubsub")
 public class GcpPubSubProperties implements CredentialsSupplier {
 
+	// Contains settings specific to the subscriber factory
 	private final Subscriber subscriber = new Subscriber();
 
+	// Contains settings specific to the publisher factory
```
artembilan: That's not JavaDocs comment...

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubProperties.java
```diff
@@ -32,11 +32,15 @@
...
 	private final Publisher publisher = new Publisher();
 
-	/** Overrides the GCP project ID specified in the Core module. */
+	/**
+	 * Overrides the GCP project ID specified in the Core module.
```
artembilan: Why haven't you fixed all similar issues in this class?
I understand that it's not your PR responsibility, but that is just simple refactoring...
Unfortunately I don't have time to walk through the project for such a fix, so I comment for flaws whenever I see them.

Thanks for understanding.

ChengyuanZhao: I fixed the other one in this file that my eye didn't catch the first time.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubProperties.java
ChengyuanZhao: I'm really confused. I thought he said there "must not" be one. 

So I added a comment for your benefit and for any reader who is wondering what this is.

spring-cloud-gcp-docs/src/main/asciidoc/pubsub.adoc
```diff
@@ -277,4 +277,32 @@ Package names that contain the types that are whitelisted for deserializing mess
 | `spring.cloud.gcp.pubsub.subscriber.parallel-pull-count` | The number of pull workers | No | The available number of processors
 | `spring.cloud.gcp.pubsub.subscriber.max-ack-duration-seconds` | The maximum period a message ack deadline will be extended in seconds | No | 3600
 | `spring.cloud.gcp.pubsub.subscriber.pull-endpoint` | The endpoint for synchronous pulling messages | No | pubsub.googleapis.com:443
+| `spring.cloud.gcp.pubsub.subscriber.retry.total-timeout-seconds`| pass-through setting of same name from `com.google.api.gax.retrying.RetrySetting` for subscriber |No |
```
meltsufin: Is there some issue with actually copying in the descriptions of the properties rather than saying "pass-through setting"? I think it would be more user-friendly.

ChengyuanZhao: The downside is that the description gets updated and also to reduce duplication of text. This also makes it clear to the user that he's directly touching that underlying setting on the client library and we aren't clouding the picture, which has its own value.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubProperties.java
artembilan: OK. As you wish. But what I believe @snicoll meant exactly `do not have to`, not `must not`: https://www.gymglish.com/en/english-grammar/must-vs-have-to. And his point that this JavaDoc on such a complex property won't be visible in the meta-data, just because it isn't a simple linear property, but the entire namespace.

P.S. Pay attention to the first paragraph in that "must-vs-have-to" article. That's why I used `must` there - my opinion, not a rule.

(Sorry for the spam though: English is not my native language :smile: )

ChengyuanZhao: Thanks for the link.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubProperties.java
```diff
@@ -32,10 +32,14 @@
 @ConfigurationProperties("spring.cloud.gcp.pubsub")
 public class GcpPubSubProperties implements CredentialsSupplier {
 
-	// Contains settings specific to the subscriber factory
+	/**
+	 * Contains settings specific to the subscriber factory
```
artembilan: I know, I did a lot of spam in this PR, but let me one more time to point to the rule about this kind of JavaDocs: https://docs.spring.io/spring-boot/docs/2.0.2.RELEASE/reference/htmlsingle/#configuration-metadata-property-attributes

> A short description of the group that can be displayed to users. If no description is available, it may be omitted. It is recommended that descriptions be short paragraphs, with the first line providing a concise summary. The last line in the description should end with a period (`.`).

The main point is a `period` in the end of the sentence.

Would you mind to fix all of them in this class?

Thanks

spring-cloud-gcp-docs/src/main/asciidoc/pubsub.adoc
meltsufin: I understand the hesitation to duplicate the documentation, but unless you can link directly to the description of this property in `com.google.api.gax.retrying.RetrySetting`, I would copy it in here and make the lives of our users easier, even if that makes our lives a bit harder. :smile:
The Pub/Sub library is GA now. So, we don't really expect these properties to change in the near future.

ChengyuanZhao: OK it sounds like you want to copy the descriptions here. I can do that.

meltsufin: Thanks!

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubProperties.java
ChengyuanZhao: Alright, will add the period. (Just want to note that in English we do not add periods to the end of incomplete sentences.)

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubProperties.java
```diff
@@ -161,4 +207,241 @@ public void setExecutorThreads(int executorThreads) {
 			this.executorThreads = executorThreads;
 		}
 	}
+
+	public static class Retry {
+
+		/**
+		 * Property for setting of the same name in {@link com.google.api.gax.retrying.RetrySettings}
```
artembilan: These all properties must also have a proper description and don't have any JavaDoc tags.
The description from here becomes a meta-data for auto-configuration properties which is used by IDEs or other visualization tools.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubProperties.java
artembilan: Thanks for the English lesson :smile: ! That's English, but here it is JavaDocs, more over a description for the auto-configuration property and there is a rule in Spring Boot. That's what I request - consistency and clean Spring code. Fair?

spring-cloud-gcp-docs/src/main/asciidoc/pubsub.adoc
ChengyuanZhao: Done. please take a look.

snicoll:
FTR we are considering a build tool that generates documentation appendix based on the metadata (so that the copy/paste is not necessary). Didn't get a chance to add that comment before the PR got merged :)

Issue #726, opened 2018-06-05
Upgrade to Finchley.RELEASE when it becomes available


meltsufin:
See: https://github.com/spring-cloud/spring-cloud-release/releases

At this point we should stop setting Spring Boot version explicitly as well.

saturnism:
Finchley is out :D

PR #727, opened 2018-06-05
Upgrade Boot to 2.0.2 and Cloud to Finchley.RC2

meltsufin:


artembilan:
Just for a note: once we have Finchley GA, I believe that we don't need explicit Boot version...

meltsufin:
@artembilan Thanks, I made a note of it in #726.

PR #730, opened 2018-06-06
Remove hard dependency on WebMVC from Logging

### LINKED ISSUE #713: Remove hard-dependency on WebMVC from Logging Starter
Logging starter and auto-configuration currently depends on WebMVC. However, not all Spring apps are Web-based. Some can be CLI only, but would still need to use Logging.

consider WebVMC as a optional dependency.

meltsufin:
The logging starter can now also be used without Web MVC.
In this case the Web MVC interceptor for trace id extraction will not
automatically kick in.

Fixes #713.


joaoandremartins:
Outside of Spring MVC, how do we expect this module to be used?
After all, `TraceIdLoggingWebMvcInterceptor` is Spring MVC-specific and the `TraceIdExtractors` are pretty trivial and not Spring-specific at all.

meltsufin:
This module would be used to get the Logback appender configs.
It's still useful even without the interceptor because MDC context is supported.

joaoandremartins:
The most obvious aspect to me is that we're completely getting rid of what little Spring there is in this module. :)
For the MDC to be used, there has to be something that writes to it. That's not clear from the documentation.
If users want someone else's appenders, shouldn't they be importing them on their own? Or are we assuming an auto-configuration role a-la-Spring Boot, where we're automating configuration for users? :)

meltsufin:
Sleuth handles writing to the MDC and we will have our specialized trace ID extractor after you PR in openzipkin is merged.

PR #731, opened 2018-06-07
Improve message conversion in Pub/Sub channel adapter

### LINKED ISSUE #161: Support content types other than UTF-8 string in SI inbound channel adapter
As suggested in [here](https://github.com/spring-cloud/spring-cloud-gcp/pull/151/files#r140067514), the SI Pub/Sub inbound channel adapter should support incoming payload types other than UTF-8 strings.

`String payload = message.getData().toStringUtf8();`

We could possibly add an encoding header to `PubsubMessage` that can be used to decode the payload on the receiving end.
We might also be able to use a similar mechanism to (de)serialize Objects into byte arrays.

meltsufin:
Replaces `Message` -> `String` -> `byte[]` message converter requirement with the direct `Message` -> `byte[]` converter, thus avoiding the UTF-8 encoding issue.

Fixes #161.

### DIFF
--- a/spring-cloud-gcp-docs/src/main/asciidoc/spring-integration-pubsub.adoc
+++ b/spring-cloud-gcp-docs/src/main/asciidoc/spring-integration-pubsub.adoc
@@ -41,6 +41,11 @@ It converts new messages to an internal Spring
 https://docs.spring.io/spring-integration/reference/html/messaging-construction-chapter.html#message[`Message`]
 and then sends it to the bound output channel.
 
+Google Pub/Sub treats message payloads as byte arrays. So, by default, the inbound channel adapter
+will construct the Spring `Message` with `byte[]` as the payload. However, you can also specify
+a `MessageConverter` that would then be used to convert the `byte[]` payload coming from
+Pub/Sub to a Spring `Message`.
+
 To use the inbound channel adapter, a `PubSubInboundChannelAdapter` must be provided and configured
 on the user application side.
 
@@ -101,8 +106,13 @@ public MessageHandler messageReceiver() {
 
 `PubSubMessageHandler` is the outbound channel adapter for GCP Pub/Sub that listens for new messages
 on a Spring `MessageChannel`.
-It uses `PubSubTemplate` to convert new `Message` instances to the GCP Pub/Sub format and post them
-to a GCP Pub/Sub topic.
+It uses `PubSubTemplate` to post them to a GCP Pub/Sub topic.
+
+To construct a Pub/Sub representation of the message, the outbound channel adapter needs to
+convert the Spring `Message` payload to a `byte[]` representation expected by Pub/Sub. If the
+`Message` payload is already of type `PubsubMessage`, `byte[]`, or `ByteString`, then a message
+converter doesn't need to be specified. Otherwise, you can specify a `MessageConverter` that
+should convert the `Object` payload of the Spring `Message` to a `byte[]` representation.
 
 To use the outbound channel adapter, a `PubSubMessageHandler` bean must be provided and configured
 on the user application side.
--- a/spring-cloud-gcp-docs/src/main/asciidoc/spring-integration-pubsub.adoc
+++ b/spring-cloud-gcp-docs/src/main/asciidoc/spring-integration-pubsub.adoc
@@ -89,7 +89,7 @@ users can extract using the `GcpPubSubHeaders.ACKNOWLEDGEMENT` key and use to (n
 @ServiceActivator(inputChannel = "pubsubInputChannel")
 public MessageHandler messageReceiver() {
     return message -> {
-        LOGGER.info("Message arrived! Payload: " + message.getPayload());
+        LOGGER.info("Message arrived! Payload: " + new String((byte[]) message.getPayload()));
         AckReplyConsumer consumer =
               message.getHeaders().get(GcpPubSubHeaders.ACKNOWLEDGEMENT, AckReplyConsumer.class);
         consumer.ack();
--- a/spring-cloud-gcp-docs/src/main/asciidoc/spring-integration-pubsub.adoc
+++ b/spring-cloud-gcp-docs/src/main/asciidoc/spring-integration-pubsub.adoc
@@ -41,10 +41,9 @@ It converts new messages to an internal Spring
 https://docs.spring.io/spring-integration/reference/html/messaging-construction-chapter.html#message[`Message`]
 and then sends it to the bound output channel.
 
-Google Pub/Sub treats message payloads as byte arrays. So, by default, the inbound channel adapter
-will construct the Spring `Message` with `byte[]` as the payload. However, you can also specify
-a `MessageConverter` that would then be used to convert the `byte[]` payload coming from
-Pub/Sub to a Spring `Message`.
+Google Pub/Sub treats message payloads as byte arrays.
+So, by default, the inbound channel adapter will construct the Spring `Message` with `byte[]` as the payload.
+However, you can also specify a `MessageConverter` that would then be used to convert the `byte[]` payload coming from Pub/Sub to a Spring `Message`.
 
 To use the inbound channel adapter, a `PubSubInboundChannelAdapter` must be provided and configured
 on the user application side.
@@ -108,11 +107,9 @@ public MessageHandler messageReceiver() {
 on a Spring `MessageChannel`.
 It uses `PubSubTemplate` to post them to a GCP Pub/Sub topic.
 
-To construct a Pub/Sub representation of the message, the outbound channel adapter needs to
-convert the Spring `Message` payload to a `byte[]` representation expected by Pub/Sub. If the
-`Message` payload is already of type `PubsubMessage`, `byte[]`, or `ByteString`, then a message
-converter doesn't need to be specified. Otherwise, you can specify a `MessageConverter` that
-should convert the `Object` payload of the Spring `Message` to a `byte[]` representation.
+To construct a Pub/Sub representation of the message, the outbound channel adapter needs to convert the Spring `Message` payload to a `byte[]` representation expected by Pub/Sub.
+If the `Message` payload is already of type `PubsubMessage`, `byte[]`, or `ByteString`, then a message converter doesn't need to be specified.
+Otherwise, you can specify a `MessageConverter` that should convert the `Object` payload of the Spring `Message` to a `byte[]` representation.
 
 To use the outbound channel adapter, a `PubSubMessageHandler` bean must be provided and configured
 on the user application side.
--- a/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapter.java
+++ b/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapter.java
@@ -26,10 +26,11 @@ import org.springframework.cloud.gcp.pubsub.core.PubSubOperations;
 import org.springframework.cloud.gcp.pubsub.integration.AckMode;
 import org.springframework.cloud.gcp.pubsub.integration.PubSubHeaderMapper;
 import org.springframework.cloud.gcp.pubsub.support.GcpPubSubHeaders;
-import org.springframework.core.convert.converter.Converter;
 import org.springframework.integration.endpoint.MessageProducerSupport;
 import org.springframework.integration.mapping.HeaderMapper;
 import org.springframework.integration.support.MessageBuilder;
+import org.springframework.messaging.MessageHeaders;
+import org.springframework.messaging.converter.MessageConverter;
 import org.springframework.util.Assert;
 
 /**
@@ -49,7 +50,7 @@ public class PubSubInboundChannelAdapter extends MessageProducerSupport {
 
 	private AckMode ackMode = AckMode.AUTO;
 
-	private Converter<byte[], Object> payloadConverter;
+	private MessageConverter messageConverter;
 
 	private HeaderMapper<Map<String, String>> headerMapper = new PubSubHeaderMapper();
 
@@ -76,16 +77,16 @@ public class PubSubInboundChannelAdapter extends MessageProducerSupport {
 		}
 
 		try {
-			Object payload;
-
-			if (this.payloadConverter != null) {
-				payload = this.payloadConverter.convert(pubsubMessage.getData().toByteArray());
+			if (this.messageConverter != null) {
+				sendMessage(this.messageConverter.toMessage(pubsubMessage.getData().toByteArray(),
+						new MessageHeaders(messageHeaders)));
 			}
 			else {
-				payload = pubsubMessage.getData().toByteArray();
+				sendMessage(MessageBuilder.withPayload(pubsubMessage.getData().toByteArray())
+						.copyHeaders(messageHeaders).build());
 			}
 
-			sendMessage(MessageBuilder.withPayload(payload).copyHeaders(messageHeaders).build());
+
 		}
 		catch (RuntimeException re) {
 			if (this.ackMode == AckMode.AUTO) {
@@ -117,18 +118,18 @@ public class PubSubInboundChannelAdapter extends MessageProducerSupport {
 		this.ackMode = ackMode;
 	}
 
-	public Converter<byte[], Object> getPayloadConverter() {
-		return this.payloadConverter;
+	public MessageConverter getMessageConverter() {
+		return this.messageConverter;
 	}
 
 	/**
-	 * Set the {@link Converter} to convert an incoming Pub/Sub message payload to an
+	 * Set the {@link MessageConverter} to convert an incoming Pub/Sub message payload to an
 	 * {@code Object}. If the converter is null, the payload is unchanged.
-	 * @param payloadConverter converts a {@link PubsubMessage} payload to a
+	 * @param messageConverter converts a {@link PubsubMessage} payload to a
 	 * {@link org.springframework.messaging.Message} payload. Can be set to null.
 	 */
-	public void setPayloadConverter(Converter<byte[], Object> payloadConverter) {
-		this.payloadConverter = payloadConverter;
+	public void setMessageConverter(MessageConverter messageConverter) {
+		this.messageConverter = messageConverter;
 	}
 
 	/**
--- a/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapter.java
+++ b/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapter.java
@@ -81,7 +81,7 @@ public class PubSubInboundChannelAdapter extends MessageProducerSupport {
 			Message<?> internalMessage = null;
 
 			if (this.messageConverter != null) {
-				this.messageConverter.toMessage(
+				internalMessage = this.messageConverter.toMessage(
 						pubsubMessage.getData().toByteArray(),
 						new MessageHeaders(messageHeaders));
 			}
--- a/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapter.java
+++ b/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapter.java
@@ -26,12 +26,10 @@ import org.springframework.cloud.gcp.pubsub.core.PubSubOperations;
 import org.springframework.cloud.gcp.pubsub.integration.AckMode;
 import org.springframework.cloud.gcp.pubsub.integration.PubSubHeaderMapper;
 import org.springframework.cloud.gcp.pubsub.support.GcpPubSubHeaders;
+import org.springframework.core.convert.converter.Converter;
 import org.springframework.integration.endpoint.MessageProducerSupport;
 import org.springframework.integration.mapping.HeaderMapper;
 import org.springframework.integration.support.MessageBuilder;
-import org.springframework.messaging.Message;
-import org.springframework.messaging.MessageHeaders;
-import org.springframework.messaging.converter.MessageConverter;
 import org.springframework.util.Assert;
 
 /**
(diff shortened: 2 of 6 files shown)


spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/outbound/PubSubMessageHandler.java
```diff
@@ -89,16 +88,18 @@ protected void handleMessageInternal(Message<?> message) throws Exception {
...
 			pubsubPayload = ByteString.copyFrom((byte[]) payload);
 		}
 		else if (payload instanceof ByteString) {
 			pubsubPayload = (ByteString) payload;
 		}
 		else {
-			pubsubPayload =	ByteString.copyFrom(
-					(String) this.messageConverter.fromMessage(message, String.class),
-					Charset.defaultCharset());
+			pubsubPayload = ByteString.copyFrom(message.toString().getBytes());
```
artembilan: Instead of this not so friendly operation I suggest to have a `GenericMessageConverter` as a default converter and don't allow to overrride it to `null`.

The `GenericMessageConverter` is base on the Spring's `ConversionService` and provide a lot of options to convert from one type to another: https://docs.spring.io/spring/docs/5.0.6.RELEASE/spring-framework-reference/core.html#core-convert-ConversionService-API

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/outbound/PubSubMessageHandler.java
```diff
@@ -127,6 +128,15 @@ public MessageConverter getMessageConverter() {
 		return this.messageConverter;
 	}
 
+	/**
+	 * Set the {@link MessageConverter} to convert an outgoing message to a {@code byte[]}
+	 * payload for sending to Pub/Sub. If {@code messageConverter} is null and payload is not
+	 * already of type {@code byte[]} or {@code {@link ByteString}}, the
```
artembilan: or `{@code }` or `{@link}` - otherwise it is something wrong.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapter.java
```diff
@@ -82,14 +78,18 @@ private void receiveMessage(PubsubMessage pubsubMessage, AckReplyConsumer consum
...
-							? MessageBuilder.withPayload(pubsubMessage.getData().toByteArray())
-							.copyHeaders(messageHeaders)
-							.build()
-							: this.messageConverter.toMessage(
-									pubsubMessage.getData().toStringUtf8(),
-							new MessageHeaders(messageHeaders));
+			Message<?> internalMessage = null;
+
+			if (this.messageConverter != null) {
+				this.messageConverter.toMessage(
```
ChengyuanZhao: Should you set `internalMessage` to this?

meltsufin: good catch!

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/outbound/PubSubMessageHandler.java
meltsufin: Should we do the same in the `PubSubInboundChannelAdapter`?

ChengyuanZhao: This `message.toString()` seems to just give the name of the class and the hashcode, and not the actual contents. Why not keep the StringMessageConverter that we had previously and use that to provide some semblance of the contents?

artembilan: No, because we don't know the target type preferences. Just clear the situation in the Docs (and JavaDocs) that without any converter we are going to produce a message with just `byte[]` payload.

meltsufin: The weird thing is that we also have the `PubSubMessageConverter` and that is only used in the `PubSubTemplate`. Shouldn't we try to reuse it somehow in the channel adapters?

joaoandremartins: I think you mean `message.getPayload()`?

meltsufin: Yes

artembilan: I think it doesn't matter here on the `PubSubMessageHandler` since both `PubSubMessageConverter` and Spring's `MessageConverter` are going to do the same functionality - extract body from the `Message`. I would still vote for the `MessageConverter` just because we deal here with the Spring `Message<?>` meanwhile `PubSubTemplate` doesn't.

meltsufin: Could `GenericMessageConverter` also fail to convert `Message` to `byte[]`? What should we do then?

artembilan: In this case we throw an exception to the caller notifying him/her that the payload is not compatible for sending to Pub/Sub. Fair? We may mention such a limitation in the Docs. 
Otherwise all this `toString().getBytes()`  does not make sense.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/outbound/PubSubMessageHandler.java
```diff
@@ -89,16 +88,18 @@ protected void handleMessageInternal(Message<?> message) throws Exception {
...
 			pubsubPayload = ByteString.copyFrom((byte[]) payload);
 		}
 		else if (payload instanceof ByteString) {
 			pubsubPayload = (ByteString) payload;
 		}
 		else {
-			pubsubPayload =	ByteString.copyFrom(
-					(String) this.messageConverter.fromMessage(message, String.class),
-					Charset.defaultCharset());
+			pubsubPayload = ByteString.copyFrom(message.getPayload().toString().getBytes());
```
ChengyuanZhao: Don't you already have the `payload` variable that you're checking the instanceof  ? `Object payload = message.getPayload();`

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/outbound/PubSubMessageHandler.java
meltsufin: If we make the converter required, then we never fall into the other branches of the `if` that allow for `byte[]` and `ByteString`. How would you handle that?

artembilan: You need to swap that `if...else` logic then. No reason to convert to `byte[]` if it is already `byte[]`. Isn't it ?

meltsufin: I'm not sure. Maybe someone would want to because `byte[]` can be anything. We would also then need to explain that the converter is ignored if the payload is `byte[]` or `ByteString`. i'm fine with that, if you are.

artembilan: Well, we have some convenient built-in functionality, anything else, fancy enough, can be done in advance before sending Spring message to this component. Remember: we deal here with Spring Integration, so any complexity an be done with other endpoints? Also would be great to let to send a `PubSubMessage` here per se. I mean as `payload` of the Spring message. Some custom `transfomer` upfornt can buid that for us.

meltsufin: Agreed, fancy transformations can be done in the SI flow before it reaches the adapter. I'll update accordingly. Thanks!

meltsufin:
@artembilan It doesn't look like the `GenericMessageConverter` can convert from `String` to `byte[]` by default.

artembilan:
Well, might be. Let's consider then a variant like `Converter<Object, byte[]>` instead. We indeed need only convert a `payload` here if it's not of type we expect. We may choose a `SerializingConverter` by default.

I did something similar in the SI-AWS `KinesisMessageHandler`: https://github.com/spring-projects/spring-integration-aws/blob/master/src/main/java/org/springframework/integration/aws/outbound/KinesisMessageHandler.java#L61. 

You can also borrow a logic from there: https://github.com/spring-projects/spring-integration-aws/blob/master/src/main/java/org/springframework/integration/aws/outbound/KinesisMessageHandler.java#L209

meltsufin:
@artembilan The tests are still failing, but can you look at my last commit to see if that's what you're thinking? Thanks!

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapter.java
```diff
@@ -51,17 +49,13 @@
 
 	private AckMode ackMode = AckMode.AUTO;
 
-	private MessageConverter messageConverter;
+	private Converter<byte[], Object> payloadConverter;
```
artembilan: I think this change is not true.
We still have to apply here a `MessageConverter.toMessage(Object payload, @Nullable MessageHeaders headers)`.
The difference with Kinesis that we don't have there headers to consult what target payload type to choose. That's why a plain `Converter<byte[], ?>` only can be applied there.

Here it is with a slightly different story, where we really map headers from the `PubSubMessage`.

Let's consider to comeback to the `MessageConverter`. Not sure if that still should be a mentioned `StringMessageConverter` or something different, but we definitely can restore an old behavior and reconsider an *inbound* part in the separate issue.

For the *outbound* side we strictly know that the target type is a `byte[]` and it doesn't matter for us what is the payload type - the provide `Converter<Object, byte[]>` should take care about that. That's why I don't see reason to manage the logic there with the `MessageConverter`. 

If you want, let's come back to the `MessageConverter` there as well for symmetry! Only the problem that we need to introduce some default implementation with the serialization capabilities...

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/outbound/PubSubMessageHandler.java
```diff
@@ -46,17 +45,20 @@
...
+ * @author Mike Eltsufin
  */
 public class PubSubMessageHandler extends AbstractMessageHandler {
 
 	private static final long DEFAULT_PUBLISH_TIMEOUT = 10000;
 
 	private final PubSubOperations pubSubTemplate;
 
-	private MessageConverter messageConverter = new StringMessageConverter();
+	private Converter<Object, byte[]> payloadConverter;
```
artembilan: If this is the part of ctor, then it has to be `final`

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/outbound/PubSubMessageHandler.java
```diff
@@ -71,8 +73,14 @@
...
 
 	public PubSubMessageHandler(PubSubOperations pubSubTemplate, String topic) {
+		this(pubSubTemplate, topic, new SerializingConverter());
+	}
+
+	public PubSubMessageHandler(PubSubOperations pubSubTemplate, String topic,
+			Converter<Object, byte[]> payloadConverter) {
 		this.pubSubTemplate = pubSubTemplate;
 		this.topicExpression = new LiteralExpression(topic);
+		this.payloadConverter = payloadConverter;
```
artembilan: I guess we need here `Assert` for all the ctor args. Otherwise we end up with `NPE` at runtime.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/outbound/PubSubMessageHandler.java
```diff
@@ -123,14 +131,21 @@ else if (payload instanceof ByteString) {
...
-				"The specified message converter can't be null.");
-		this.messageConverter = messageConverter;
+	/**
+	 * Set the {@link Converter} to convert an outgoing message payload to a {@code byte[]}
+	 * payload for sending to Pub/Sub. If the payload is already of type {@link PubsubMessage},
+	 * {@code byte[]} or {@link ByteString}, the converter is ignored.
+	 * @param payloadConverter converts {@link Message} to a payload of the outgoing message
+	 * to Pub/Sub. Cannot be set to null.
+	 */
+	public void setPayloadConverter(Converter<Object, byte[]> payloadConverter) {
```
artembilan: Right, or just let's not introduce an new ctor since this option is err... optional.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapter.java
meltsufin: @artembilan I considered the header conversion issue as well, but since we already have a `headerMapper` in this class, I think it's really OK, and even better to just use the payload converter. WDYT?

artembilan: No, you didn't get it. We receive a raw `byte[]` and headers may have an information into what type to covert it for the message to send. That's exactly what a `MessageConverter.toMessage(Object payload, @Nullable MessageHeaders headers)` suggests for us. With the raw `Converter` we don't have such an ability and maximum what can be useful for us is a standard Java Serialization model.

So, IMHO that's not OK in this case and we really should come back to the `MessageConverter` on the inbound side.

meltsufin: I see your point, but here's my concern. If we have `MessageConverter`, then what do we do with the `headerMapper`, because it seems confusing to use a `MessageConverter` just to convert the payload, but not the whole message. The other nuisance is the asymmetry between the inbound and the outbound. Can we just say that conversion dependent on headers is an advanced use-case that should be handled outside of the adapter somewhere else in the integration flow?
Otherwise, another way to say what you're asking for is `Converter<Message, byte[]>`, which I guess we can do without switching to `MessageConverter`.

artembilan: Not sure how a `HeaderMapper` confuses you since it does only one purpose: creates a plain Java `Map` from the native header representation: `Map<String, Object> toHeaders(T source)`. As you see there is nothing about a `Message` at all. The `MessageConverter` is really the next step in the logic to build a proper `Message` from the native payload and converted headers.

The `Converter<Message, byte[]>` still doesn't bring too much value since you don't provide `headers` contract and don't propagate them to the converter.

Ass I said: we can bring `MessageConverter` to the outbound side as well to let you do not be confused, but that doesn't bring any values since we always expect only a `byte[]` as outcome from the conversion.

I don't see reason to document anything as a limitation since we have handy `HeaderMapper` and `MessageConverter` as out-of-the-box solution in Spring already.

Please, take a look to existing Channel Adapter implementations. I would say we have already came through similar architectural concerns in other places and that's how we ended up with the 
`HeaderMapper` and `MessageConverter`  pair. The outbound side really can be questionable, but IMO it's overheader to bring `MessageConverter` there since we always expect only `byte[]`.

meltsufin: I was actually looking at the outbound, while you were talking about the inbound. Sorry!
Now what you are saying makes sense! :blush:

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapter.java
```diff
@@ -82,16 +77,16 @@ private void receiveMessage(PubsubMessage pubsubMessage, AckReplyConsumer consum
...
-							? MessageBuilder.withPayload(pubsubMessage.getData().toByteArray())
-							.copyHeaders(messageHeaders)
-							.build()
-							: this.messageConverter.toMessage(
-									pubsubMessage.getData().toStringUtf8(),
-							new MessageHeaders(messageHeaders));
-
-			sendMessage(internalMessage);
+			if (this.messageConverter != null) {
+				sendMessage(this.messageConverter.toMessage(pubsubMessage.getData().toByteArray(),
```
joaoandremartins: nit: single call to `sendMessage()`

meltsufin:
@artembilan I went back to using a `MessageConverter` in the inbound adapter. I'm still not completely satisfied with it because (1) `MessageConverter` is two-way, but we only use it in one direction and (2) `String`s don't work that great because the outbound adapter will serialize them by default, but the inbound will not deserialize. I'm not sure how to make it better though.

artembilan:
The `MessageConverter` is two-way because it really can be used on both sides and it has two APIs for consistent data model on producer and consumer. As I said: we can use `MessageConverter` on the outbound side as well: in most cases the same instance will be injected to both channel adapters. Or even more: when we think about Spring Boot, we definitely may just auto-configure some `MessageConverter` on the matter (JSON in most cases?) and suggest it for injection in both channel adapter. I feel like I see now the reason we have `MessageConverter` API on both side. 

May I get my words back and request a `MessageConverter` for the outbond side as well? 😄 


I don't see what you mean about `String` on the outbound side, but if we really don't have any hints from the `PubsubMessage` consumed about data type out-of-the-box, then we don't have choice unless to let `MessageConverter` to deal with that `byte[]`. I might miss something - won't mind any explanations from the Pub/Sub perspective.

meltsufin:
@artembilan I updated to use the `MessageConverter`. I also made it so it fails hard if the message payload is not `byte[]`/`ByteString`/`PubsubMessage` and the converter is not specified.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/outbound/PubSubMessageHandler.java
```diff
@@ -82,30 +86,35 @@ protected void handleMessageInternal(Message<?> message) throws Exception {
...
+						(byte[]) this.messageConverter.fromMessage(message, byte[].class));
+			}
+			else if (payload instanceof byte[]) {
+				pubsubPayload = ByteString.copyFrom((byte[]) payload);
+			}
+			else if (payload instanceof ByteString) {
+				pubsubPayload = (ByteString) payload;
+			}
+			else {
+				throw new PubSubException("Unable to convert payload of type "
```
artembilan: Sounds like a `MessageConversionException`? Seems for me it's a bit early to fail as a Pub/Sub.

But that's IMO I guess.

meltsufin: Sure, that works.
I'll update the refdoc tomorrow.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapter.java
```diff
@@ -82,16 +78,16 @@ private void receiveMessage(PubsubMessage pubsubMessage, AckReplyConsumer consum
...
-							.copyHeaders(messageHeaders)
-							.build()
-							: this.messageConverter.toMessage(
-									pubsubMessage.getData().toStringUtf8(),
-							new MessageHeaders(messageHeaders));
-
-			sendMessage(internalMessage);
+			Message message;
+			if (this.messageConverter != null) {
+				message = this.messageConverter.toMessage(pubsubMessage.getData().toByteArray(),
```
meltsufin: @artembilan Maybe we should pass in the whole `PubsubMessage` to the converter instead of just its payload?

artembilan: No, that's not, because that way the whole `PubsubMessage` is going to be as a payload of the Spring message to send. Just take a look into the `MappingJackson2MessageConverter.convertToInternal()`. We do there `this.objectMapper.writeValue(generator, payload);`, I'm not sure that everybody is going to be happy that you try to convert the whole `PubsubMessage` there...

meltsufin: OK, thanks! Any remaining concerns about the PR?

spring-cloud-gcp-docs/src/main/asciidoc/spring-integration-pubsub.adoc
```diff
@@ -41,6 +41,11 @@ It converts new messages to an internal Spring
 https://docs.spring.io/spring-integration/reference/html/messaging-construction-chapter.html#message[`Message`]
 and then sends it to the bound output channel.
 
+Google Pub/Sub treats message payloads as byte arrays. So, by default, the inbound channel adapter
+will construct the Spring `Message` with `byte[]` as the payload. However, you can also specify
```
artembilan: What happened with the rules *Sentence per line* : https://asciidoctor.org/docs/asciidoc-recommended-practices/#one-sentence-per-line ?

PR #736, opened 2018-06-11
Update Runtime Config, Spanner doc mention beta

### LINKED ISSUE #735: Update refdoc to reflect that Runtime config, Spanner support is Beta/Preview

ChengyuanZhao:
fixes #735


spring-cloud-gcp-docs/src/main/asciidoc/spanner.adoc
```diff
@@ -31,6 +31,8 @@ dependencies {
 This setup takes care of bringing in the latest compatible version of Cloud Java Cloud Spanner libraries
 as well.
 
+Spring Data Cloud Spanner and the underlying Cloud Spanner Java Client Library are beta status.
```
meltsufin: I think we wanted to update the main header of the section
"Spring Data Cloud Spanner BETA"
Also, we need to make it clear that for now you will be able to get the artifacts for Spanner in the Spring Milestones repository.

ChengyuanZhao: Should we retag the underlying issue to GA or RC? These two things you mention make sense if this PR is supposed to go in after M4 but before RC/GA.

meltsufin: Sure, we can retag, although I'm not too worried about it being in M4 as well.

ChengyuanZhao: It's just that it doesn't seem to make sense putting it into M4:

1. we have a section saying all the modules are in the spring milestone repo in the readme applying to all modules right now. I assume this is going away in the RC/GA version of the project, at which point it would make sense to specifically say the Spanner module can be found in the milestones repo

2. since we're not at GA/RC for M4, doesn't that mean technically every single module is beta? Then it wouldn't make sense to specifically label this module as Beta as none of the other are.

I'll add the awaiting tag and put this into the release milestone. I'll wait until M4 is out to merge in this  PR.

meltsufin: Make sense. We should also add a blocking issue for GA to update the refdoc regarding the milestones repo.

ChengyuanZhao:
@meltsufin I added the BETA to the titles and also added these to the runtime config ref doc. (I didn't add the specific mention that these are from the spring milestones repo, because that should be done when the other modules are finally available in maven central and we make the doc changes for that?)

spring-cloud-gcp-docs/src/main/asciidoc/config.adoc
```diff
@@ -31,6 +31,8 @@ dependencies {
 
 A https://github.com/spring-cloud/spring-cloud-gcp/tree/{git_version}/spring-cloud-gcp-samples/spring-cloud-gcp-config-sample[sample application] is available.
 
+The Google Cloud Runtime Configuration service is in beta status.
+
```
meltsufin: We should mention somewhere which versions of the project it's available in -- i.e. 1.0.0.M4.
I'm not sure what the best way to do it is though. Maybe on the project page: http://cloud.spring.io/spring-cloud-gcp/.

ChengyuanZhao: I'll make an issue of this

PR #742, opened 2018-06-12
fix PubSubTemplateIntegrationTests

dmitry-s:


spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/pubsub/it/PubSubTemplateIntegrationTests.java
```diff
@@ -16,10 +16,9 @@
 
 package org.springframework.cloud.gcp.autoconfigure.pubsub.it;
 
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.UUID;
+import java.util.*;
```
artembilan: No asterisk imports, please. 
Seems for me we have a Checkstyle rule on the matter.
Otherwise we need to add it.

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/pubsub/it/PubSubTemplateIntegrationTests.java
```diff
@@ -110,9 +109,18 @@ public void testPullAndAck() {
...
-			pubSubTemplate.publish(topicName, "valedictory");
-			pubSubTemplate.publish(topicName, "the-runaway");
+			List<Future<String>> futures = new ArrayList<>();
+			futures.add(pubSubTemplate.publish(topicName, "free-hand"));
+			futures.add(pubSubTemplate.publish(topicName, "valedictory"));
+			futures.add(pubSubTemplate.publish(topicName, "the-runaway"));
+
+			futures.parallelStream().forEach( f -> {
+				try {
+					f.get();
```
artembilan: You can't do this `get()` which is infinite by nature. You will block CI server forever if there is something wrong with the target Pub/Sub system.

I wonder if there is some callback where we can use a `CountDownLatch` with its `await(10_000)` afterwards to assert that we really get a result during reasonable timeout.

artembilan:
Oh, wait!

There *is* a Checkstyle for you to follow:
```
/home/travis/build/spring-cloud/spring-cloud-gcp/spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/pubsub/it/PubSubTemplateIntegrationTests.java:19: error: Using the '.*' form of import should be avoided - java.util.*.
/home/travis/build/spring-cloud/spring-cloud-gcp/spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/pubsub/it/PubSubTemplateIntegrationTests.java:117:46: error: '(' is followed by whitespace.
/home/travis/build/spring-cloud/spring-cloud-gcp/spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/pubsub/it/PubSubTemplateIntegrationTests.java:120:17: error: '}' at column 5 should be alone on a line.
```

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/pubsub/it/PubSubTemplateIntegrationTests.java
ChengyuanZhao: There is `Future.get(long timeout, TimeUnit unit)` which allows you to give that time limit, too.

artembilan: Well, looking to the test-case closely I don't think that we need such a kung-fu about the publish `Future`.

Our test relies on the `PubSubTemplate.pull()`. Seems for me it must be blocked at least for some time. Isn't it?
Or what am I missing?

Thanks

dmitry-s: I'm going to change this test significantly. In it's current implementation it doesn't actually test what we want it to test. For example, there is no only-once delivery guarantee for messages, on which this test relies when it tries to count how many messages we received.

joaoandremartins: I don't really see the need for this test to be changed "significantly".
In practice, we find that the messages do end up being delivered once.

dmitry-s: I think that's one of the reasons this test is failing from time to time.

joaoandremartins: So we get more than 5 messages?
Do you have a sample Travis output?

dmitry-s: @joaoandremartins I've seen it locally

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubProperties.java
```diff
@@ -136,7 +136,7 @@ public void setExecutorThreads(int executorThreads) {
 		/**
 		 * The optional max ack duration in seconds for the subscriber factory.
```
meltsufin: Update comment please.

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/pubsub/it/PubSubTemplateIntegrationTests.java
```diff
@@ -106,35 +117,53 @@ public void testPullAndAck() {
 			String topicName = "peel-the-paint" + UUID.randomUUID();
 			String subscriptionName = "i-lost-my-head" + UUID.randomUUID();
 			pubSubAdmin.createTopic(topicName);
-			pubSubAdmin.createSubscription(subscriptionName, topicName);
+			pubSubAdmin.createSubscription(subscriptionName, topicName, 10);
```
meltsufin: Can we lower it to like 3 seconds, so that the test runs faster?

dmitry-s: no, 10 is the minimum value

meltsufin:
@artembilan Do you have any more concerns on this? Thanks!

Issue #718, opened 2018-06-01
Remove `google-api-client` version.


joaoandremartins:
#716 set the version of `google-api-client` internally as `1.23.0`.
Ideally, we wouldn't have had to set any version and our Google libraries would just use the version from `dependencyManagement`.

1) Ask the Cloud SQL sockets factory team to update the version of `com.google.apis:google-api-services-sqladmin` to `v1beta4-rev26-1.23.0` in `com.google.cloud.sql:postgres-socket-factory` and `com.google.cloud.sql:mysql-socket-factory`,
2) Have them, as well as `com.google.apis:google-api-services-sqladmin`, start getting dependency versions from `dependencyManagement`.

meltsufin:
@patflynn FYI as it relates to the diamond dependency issues.

patflynn:
thanks @meltsufin.

Have you filed any bugs related to this with library authors?

meltsufin:
@joaoandremartins ?

joaoandremartins:
It isn't clear if `com.google.apis:google-api-services-sqladmin` has any GitHub repo.
Making the sockets factory change is trivial, I'll submit a PR some time today.

joaoandremartins:
https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/pull/76 will allow us to not set the version of `google-api-client` on our side (provide a `1.24.0` one isn't released in the meantime and things break again...)

PR #745, opened 2018-06-13
Update to latest google-cloud-java and Cloud SQL

### LINKED ISSUE #711: Cloud SQL usage metrics not working

meltsufin:
Fixes #711, #718

### DIFF
--- a/spring-cloud-gcp-autoconfigure/pom.xml
+++ b/spring-cloud-gcp-autoconfigure/pom.xml
@@ -78,16 +78,6 @@
             <groupId>com.google.cloud.sql</groupId>
             <artifactId>mysql-socket-factory</artifactId>
             <optional>true</optional>
-            <exclusions>
-                <exclusion>
-                    <groupId>mysql</groupId>
-                    <artifactId>mysql-connector-java</artifactId>
-                </exclusion>
-                <exclusion>
-                    <groupId>com.google.api-client</groupId>
-                    <artifactId>google-api-client</artifactId>
-                </exclusion>
-            </exclusions>
         </dependency>
 
         <dependency>
--- a/spring-cloud-gcp-dependencies/pom.xml
+++ b/spring-cloud-gcp-dependencies/pom.xml
@@ -17,7 +17,7 @@
 
 	<properties>
 		<google-cloud-bom.version>0.51.0-alpha</google-cloud-bom.version>
-		<google-auth.version>0.9.1</google-auth.version>
+		<google-auth-library-oauth2-http.version>0.9.1</google-auth-library-oauth2-http.version>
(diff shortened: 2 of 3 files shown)


joaoandremartins:
As part of this PR, could you also remove the declaration of `google-api-client` in the root `pom.xml`?

spring-cloud-gcp-autoconfigure/pom.xml
```diff
@@ -78,16 +78,6 @@
             <groupId>com.google.cloud.sql</groupId>
             <artifactId>mysql-socket-factory</artifactId>
             <optional>true</optional>
-            <exclusions>
-                <exclusion>
-                    <groupId>mysql</groupId>
```
joaoandremartins: We might still need this one?

spring-cloud-gcp-dependencies/pom.xml
```diff
@@ -119,25 +120,13 @@
 			<dependency>
 				<groupId>com.google.cloud.sql</groupId>
 				<artifactId>mysql-socket-factory</artifactId>
-				<version>1.0.8</version>
-				<exclusions>
```
joaoandremartins: Is this library no longer pulled in by them?

spring-cloud-gcp-autoconfigure/pom.xml
meltsufin: This dependency is now marked as `provided`. So, I don't think we need to exclude it.
https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/blob/28b266b13ec4a9b6204226a864453d3903095f7e/connector-j-5/pom.xml#L25

spring-cloud-gcp-dependencies/pom.xml
meltsufin: They exclude it: https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/blob/28b266b13ec4a9b6204226a864453d3903095f7e/core/pom.xml#L29

spring-cloud-gcp-dependencies/pom.xml
```diff
@@ -16,8 +16,9 @@
 	<packaging>pom</packaging>
 
 	<properties>
-		<google-cloud-bom.version>0.50.0-alpha</google-cloud-bom.version>
-		<cloud-trace-java.version>0.5.0</cloud-trace-java.version>
+		<google-cloud-bom.version>0.51.0-alpha</google-cloud-bom.version>
+		<google-auth.version>0.9.1</google-auth.version>
```
joaoandremartins: nit, but should we call this `google-auth-library-oauth2-http.version`, not to confuse with other Google auth libraries?

joaoandremartins:
This also fixes #718

PR #746, opened 2018-06-13
Uses new x-cloud-trace-context extractor

### LINKED ISSUE #703: Add extractor for x-cloud-trace-context header
After #695, tracing will support the `X-B3-TraceId` header. To work in Google Cloud Platform, we should also add support for the `x-cloud-trace-context` header.

The suggestion is to modify the existing [`B3Propagation`](https://github.com/openzipkin/brave/blob/master/brave/src/main/java/brave/propagation/B3Propagation.java) extractor to also contemplate the `x-cloud-trace-context` header. However, since there will be no span IDs or parent trace IDs at initial incoming request time, we must still decide on what to do [there](https://github.com/openzipkin/brave/blob/master/brave/src/main/java/brave/propagation/B3Propagation.java#L134).

As reference, here is [another extractor](https://github.com/openzipkin/brave/tree/master/propagation/aws) implementation that takes full control of the propagation. I don't think we want to do that, it will probably do just to adapt the current `B3Propagation`.

joaoandremartins:
Our Trace integration now supports the x-cloud-trace-context trace key
used in GCP, along with the B3 keys, X-B3-TraceId, X-B3-SpanId, etc.

Fixes #703

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -41,7 +41,7 @@
 		<maven.compiler.target>1.8</maven.compiler.target>
 		<spring-cloud-release.version>Finchley.RC2</spring-cloud-release.version>
 		<spring-boot-release.version>2.0.2.RELEASE</spring-boot-release.version>
-		<zipkin-gcp.version>0.6.1-SNAPSHOT</zipkin-gcp.version>
+		<zipkin-gcp.version>0.6.2</zipkin-gcp.version>
 	</properties>
 
 	<dependencyManagement>
--- a/pom.xml
+++ b/pom.xml
@@ -78,7 +78,13 @@
 			<dependency>
 				<groupId>io.zipkin.gcp</groupId>
 				<artifactId>zipkin-sender-stackdriver</artifactId>
-				<version>0.4.1</version>
+				<version>0.6.1-SNAPSHOT</version>
+			</dependency>
+
+			<dependency>
+				<groupId>io.zipkin.gcp</groupId>
+				<artifactId>brave-propagation-stackdriver</artifactId>
+				<version>0.6.1-SNAPSHOT</version>
 			</dependency>
 
 			<dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -41,6 +41,7 @@
 		<maven.compiler.target>1.8</maven.compiler.target>
 		<spring-cloud-release.version>Finchley.RC2</spring-cloud-release.version>
 		<spring-boot-release.version>2.0.2.RELEASE</spring-boot-release.version>
+		<zipkin-gcp.version>0.6.1-SNAPSHOT</zipkin-gcp.version>
 	</properties>
 
 	<dependencyManagement>
@@ -78,13 +79,13 @@
 			<dependency>
 				<groupId>io.zipkin.gcp</groupId>
 				<artifactId>zipkin-sender-stackdriver</artifactId>
-				<version>0.6.1-SNAPSHOT</version>
+				<version>${zipkin-gcp.version}</version>
 			</dependency>
 
 			<dependency>
 				<groupId>io.zipkin.gcp</groupId>
 				<artifactId>brave-propagation-stackdriver</artifactId>
-				<version>0.6.1-SNAPSHOT</version>
+				<version>${zipkin-gcp.version}</version>
 			</dependency>
 
 			<dependency>
--- a/spring-cloud-gcp-docs/src/main/asciidoc/trace.adoc
+++ b/spring-cloud-gcp-docs/src/main/asciidoc/trace.adoc
@@ -44,7 +44,7 @@ This integration enables Brave to use the https://github.com/openzipkin/zipkin-g
 
 A propagation is responsible for extracting trace context from an entity (e.g., a HTTP servlet request) and for injecting trace context into an entity.
 A canonical example of the propagation usage is a web server that receives a HTTP request, which triggers other HTTP requests from the server before returning a HTTP response to the original caller.
-In the case of `StackdriverTracePropagation`, first it looks for trace context from the `x-cloud-trace-context` key (e.g., a HTTP request header).
+In the case of `StackdriverTracePropagation`, first it looks for trace context in the `x-cloud-trace-context` key (e.g., a HTTP request header).
 The value of the `x-cloud-trace-context` key can be formatted in three different ways:
 
 * `x-cloud-trace-context: TRACE_ID`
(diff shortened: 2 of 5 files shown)


joaoandremartins:
This PR is just waiting for `zipkin-gcp` version 0.6.1 to be available from Maven Central.

pom.xml
```diff
@@ -78,7 +78,13 @@
 			<dependency>
 				<groupId>io.zipkin.gcp</groupId>
 				<artifactId>zipkin-sender-stackdriver</artifactId>
-				<version>0.4.1</version>
+				<version>0.6.1-SNAPSHOT</version>
```
meltsufin: Maybe use a maven property to avoid duplicating the version number in two places?

spring-cloud-gcp-docs/src/main/asciidoc/trace.adoc
```diff
@@ -32,20 +27,43 @@ dependencies {
...
-without modifying existing applications.
+If you are already using a Zipkin server capturing trace information from multiple platform/frameworks, you also use a https://cloud.google.com/trace/docs/zipkin[Stackdriver Zipkin proxy] to forward those traces to Stackdriver Trace without modifying existing applications.
 ====
 
+=== Tracing
+
+Spring Cloud Sleuth uses the https://github.com/openzipkin/brave[Brave tracer] to generate traces.
+This integration enables Brave to use the https://github.com/openzipkin/zipkin-gcp/tree/master/propagation-stackdriver[`StackdriverTracePropagation`] propagation.
+
+A propagation is responsible for extracting trace context from an entity (e.g., a HTTP servlet request) and for injecting trace context into an entity.
```
meltsufin: "an HTTP"

spring-cloud-gcp-docs/src/main/asciidoc/trace.adoc
```diff
@@ -32,20 +27,43 @@ dependencies {
...
+A canonical example of the propagation usage is a web server that receives a HTTP request, which triggers other HTTP requests from the server before returning a HTTP response to the original caller.
```
meltsufin: Again "a HTTP" -> "an HTTP"

spring-cloud-gcp-docs/src/main/asciidoc/trace.adoc
```diff
@@ -32,20 +27,43 @@ dependencies {
...
+In the case of `StackdriverTracePropagation`, first it looks for trace context from the `x-cloud-trace-context` key (e.g., a HTTP request header).
```
meltsufin: looks for trace context (from -> in)

spring-cloud-gcp-docs/src/main/asciidoc/trace.adoc
```diff
@@ -32,20 +27,43 @@ dependencies {
...
+The value of the `x-cloud-trace-context` key can be formatted in three different ways:
```
meltsufin: Do you mean that `StackdriverTracePropagation` allows all of these formats?

spring-cloud-gcp-docs/src/main/asciidoc/trace.adoc
```diff
@@ -32,20 +27,43 @@ dependencies {
...
+* `x-cloud-trace-context: TRACE_ID`
+* `x-cloud-trace-context: TRACE_ID/SPAN_ID`
+* `x-cloud-trace-context: TRACE_ID/SPAN_ID;o=TRACE_TRUE`
+
+`TRACE_ID` is a 32-character hexadecimal value that encodes a 128-bit number.
+
+`SPAN_ID` is an unsigned long.
+Since Stackdriver Trace doesn't support span joins, a new span ID is always generated, regardless of the one specified in `x-cloud-trace-context`.
+
+`TRACE_TRUE` can either be `0` is the entity should be untraced, or `1` if it should be traced.
```
meltsufin: "is" doesn't sound right here.
Maybe:
>`TRACE_TRUE` can either be `0` meaning that the entity should not be traced, or `1` if it should be traced.

spring-cloud-gcp-docs/src/main/asciidoc/trace.adoc
```diff
@@ -84,4 +102,6 @@ Spring Cloud GCP Trace does override some Sleuth configurations:
 
 === Integration with Logging
 
-Logs can also be associated with traces. See link:logging.adoc[Stackdriver Logging Support] for how to configure logging such that it carries Trace ID metadata.
+Integration with Stackdriver Logging is available through the link:logging.adoc[Stackdriver Logging Support].
+If the Trace integration is used together with the Logging one, the request logs will be associated to the trace.
```
meltsufin: "associated to the corresponding traces."

spring-cloud-gcp-docs/src/main/asciidoc/trace.adoc
joaoandremartins: I think "a HTTP" request is actually the correct english, since it's not a mute H.

spring-cloud-gcp-docs/src/main/asciidoc/trace.adoc
joaoandremartins: It's not `StackdriverTracePropagation` that allows them, it's how the convention is, and the propagation follows it.

spring-cloud-gcp-docs/src/main/asciidoc/trace.adoc
joaoandremartins: It should be `if` :)

spring-cloud-gcp-docs/src/main/asciidoc/trace.adoc
meltsufin: http://blog.apastyle.org/apastyle/2012/04/using-a-or-an-with-acronyms-and-abbreviations.html

pom.xml
```diff
@@ -41,6 +41,7 @@
 		<maven.compiler.target>1.8</maven.compiler.target>
 		<spring-cloud-release.version>Finchley.RC2</spring-cloud-release.version>
 		<spring-boot-release.version>2.0.2.RELEASE</spring-boot-release.version>
+		<zipkin-gcp.version>0.6.2</zipkin-gcp.version>
```
adriancole: when you feel ready. we can bump this to 1.0.0 ok. it is your call

meltsufin: @adriancole Can you please comment on #749? I wonder if this requires changes to the extractor. Thanks!

Issue #705, opened 2018-05-25
Run samples on CI


dmitry-s:
Currently we do not automatically check if samples work, so we should have CI verify that everything is in working order

meltsufin:
It's a great idea, and I think this will all serve the goals of #171.

meltsufin:
One thing to consider here is adding `@SpringBootTest` [tests](https://docs.spring.io/spring-boot/docs/current/reference/html/boot-features-testing.html#boot-features-testing-spring-boot-applications) to samples.

dzou:
At this point in time, all samples have tests covering them so I will close this issue.

PR #752, opened 2018-06-14
Fixes Pub/Sub SI sample

### LINKED ISSUE #750: PubSub Integration Sample doesn't have converter for String to byte[]

joaoandremartins:
#731 caused our SI adapters to expect receiving byte arrays, as well as
producing them, leaving the actual conversion to the client.
Therefore, the sample should convert the strings to/from the byte[].

Also puts sentences from the doc in one line.

Fixes #750

### DIFF
--- a/spring-cloud-gcp-samples/spring-cloud-gcp-integration-pubsub-sample/spring-cloud-gcp-integration-pubsub-sample-sender/src/main/java/com/example/SenderApplication.java
+++ b/spring-cloud-gcp-samples/spring-cloud-gcp-integration-pubsub-sample/spring-cloud-gcp-integration-pubsub-sample-sender/src/main/java/com/example/SenderApplication.java
@@ -66,6 +66,6 @@ public class SenderApplication {
 	@MessagingGateway(defaultRequestChannel = "pubSubOutputChannel")
 	public interface PubSubOutboundGateway {
 
-		void sendToPubSub(String text);
+		void sendToPubSub(byte[] text);
 	}
 }
--- a/spring-cloud-gcp-samples/spring-cloud-gcp-integration-pubsub-sample/README.adoc
+++ b/spring-cloud-gcp-samples/spring-cloud-gcp-integration-pubsub-sample/README.adoc
@@ -1,31 +1,21 @@
 = Spring Integration Channel Adapters for Google Cloud Pub/Sub Code Sample
 
 This code sample contains a couple of apps.
-One app gets a message provided by the user through a webpage and publishes it to a Google Cloud
-Pub/Sub topic.
+One app gets a message provided by the user through a webpage and publishes it to a Google Cloud Pub/Sub topic.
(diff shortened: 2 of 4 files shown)


spring-cloud-gcp-samples/spring-cloud-gcp-integration-pubsub-sample/spring-cloud-gcp-integration-pubsub-sample-sender/src/main/java/com/example/SenderApplication.java
```diff
@@ -66,6 +66,6 @@ public void onSuccess(String result) {
 	@MessagingGateway(defaultRequestChannel = "pubSubOutputChannel")
 	public interface PubSubOutboundGateway {
 
-		void sendToPubSub(String text);
+		void sendToPubSub(byte[] text);
```
artembilan: Can't we really make an explicit `String -> byte[]` conversion in the `PubSubMessageHandler`?

Seems for me not a big deal to call there `getBytes()` instead of requesting some external converters/paylaodMappers.

joaoandremartins: That would make us also ask users for an encoding at the adapter level, which is totally fine by me.
But that would be better suited for another PR I think
@meltsufin wdyt?

artembilan: Correct. So, can we hold merging this PR off until we come up with the solution for target adapters?

joaoandremartins: This PR fixes the sample to be according to the current state of `master`, which should have been done on #731.

If anything, I think this PR should be merged and, later, if we decide on any such changes, they should be accompanied by the respective sample changes.

artembilan: OK! So, the next time let's consider to fix samples together with the actual core changes.
Otherwise it doesn't looks like productive development: we are going to merge the change we we already know will be dropped by the next change. 🤷‍♂️

meltsufin: In the future, we will have the samples run automatically as part of build and will catch these kinds of issues. See: #705 .

PR #755, opened 2018-06-15
Guidance on pre-release process

joaoandremartins:
So that the required actions on Pivotal's side during release are as
minimal as possible, here is a list of things to do before asking
Pivotal to run release scripts.

### DIFF
--- a/RELEASING.adoc
+++ b/RELEASING.adoc
@@ -5,7 +5,7 @@ In order to simplify that process, here is a list of prerequisites to be perform
 
 1. Comment out the `spring-cloud-gcp-samples` module from the root `pom.xml` file, so that the samples aren't released.
 
-2. For any non-GA modules (e.g., Spanner), also comment them out from the `pom.xml` file.
+2. Comment out any non-GA modules from the `pom.xml` file.
 
 3. Manually change the value of `TRACKING_HEADER_PROJECT_VERSION` in link:spring-cloud-gcp-core/src/main/java/org/springframework/cloud/gcp/core/UsageTrackingHeaderProvider.java[`UsageTrackingHeaderProvider`] to the version being released from the interim `BUILD-SNAPSHOT` value.
 
--- a/RELEASING.adoc
+++ b/RELEASING.adoc
@@ -1,7 +1,7 @@
 = Spring Cloud GCP release process
 
-The actual release process is triggered by Pivotal.
-In order to simplify that process, here is a list of prerequisites to be performed before asking Pivotal to run the release scripts.
+The actual release process is triggered by the release admin.
+In order to simplify that process, here is a list of prerequisites to be performed before asking the release admin to run the release scripts.
 
 1. Comment out the `spring-cloud-gcp-samples` module from the root `pom.xml` file, so that the samples aren't released.
 
@@ -11,4 +11,4 @@ In order to simplify that process, here is a list of prerequisites to be perform
 
 4. Change the value of the `<version>` elements of the `pom.xml` of each module in the project being released to the name of the version being released from the interim `BUILD-SNAPSHOT` values. This can be done manually or using the https://www.mojohaus.org/versions-maven-plugin/set-mojo.html[mvn versions:set] command.
 
-After the release scripts have been run by Pivotal, revert the version names from steps 3 and 4 to the interim `BUILD-SNAPSHOT` name.
+After the release scripts have been run by the release admin, revert the version names from steps 3 and 4 to the interim `BUILD-SNAPSHOT` name.
--- a/RELEASING.adoc
+++ b/RELEASING.adoc
@@ -7,8 +7,8 @@ In order to simplify that process, here is a list of prerequisites to be perform
 
 2. For any non-GA modules (e.g., Spanner), also comment them out from the `pom.xml` file.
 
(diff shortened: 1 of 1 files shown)


joaoandremartins:
@dmitry-s should anything be said here about that set-up to have the refdoc point at the latest code samples?

RELEASING.adoc
```diff
@@ -0,0 +1,12 @@
+= Spring Cloud GCP release process
+
+The actual release process is triggered by Pivotal.
```
artembilan: This is something obvious. The project is owned by Pivotal per se.
I fully don't think that there should be something to tell like "ask Pivotal" in the project files.

I don't mind about such an instruction, but it must be as formal as possible and should come from the project perspective. You can have this instruction on your, Google side, but not here in the Pivotal project.

RELEASING.adoc
```diff
@@ -0,0 +1,12 @@
...
+In order to simplify that process, here is a list of prerequisites to be performed before asking Pivotal to run the release scripts.
+
+1. Comment out the `spring-cloud-gcp-samples` module from the root `pom.xml` file, so that the samples aren't released.
+
+2. For any non-GA modules (e.g., Spanner), also comment them out from the `pom.xml` file.
```
ChengyuanZhao: I don't think you need to mention any specific modules here; it would just be another thing to forget when things change in the future.

RELEASING.adoc
```diff
@@ -0,0 +1,12 @@
...
+
+3. Manually change the value of `TRACKING_HEADER_PROJECT_VERSION` in link:spring-cloud-gcp-core/src/main/java/org/springframework/cloud/gcp/core/UsageTrackingHeaderProvider.java[`UsageTrackingHeaderProvider`] to the version being released.
+
+After the release scripts have been run by Pivotal, revert all these changes.
```
ChengyuanZhao: This line mentions scripts being run by Pivotal, implying that that process is outside of the process described herein. As a result, I don't think you can remove the first line in this doc if you're going to keep this one, because it would be without context.

meltsufin: Wherever you're saying "Pivotal" right now, you can just say "project admins".

RELEASING.adoc
```diff
@@ -0,0 +1,12 @@
+= Spring Cloud GCP release process
```
meltsufin: I think there should be here steps that require the change of the version of the project from `SNAPSHOT` to release, and also changing it back to `SNAPSHOT` after the release from master.

RELEASING.adoc
ChengyuanZhao: @meltsufin that could be less informative. I can imagine myself as new to the team and thinking we are the project admins since we are in charge of code reviews and writing the code. Saying "Pivotal" is totally clear and probably will stay accurate for the foreseeable future .

artembilan: Well, then this document should be internal for your team.
One more time: it doesn't make sense to say in the Pivtoal doc that we need to ask Pivotal.

This project is Pivotal proprietary and it is really clear from context of the project that who is responsible for release process.

Well, we really can say `project admins` and then do not consider your team as in the admin role.
You are indeed not admins: you don't perform release 🤷‍♂️

RELEASING.adoc
ChengyuanZhao: Added another step for this in latest commit.

RELEASING.adoc
```diff
@@ -7,6 +7,8 @@ In order to simplify that process, here is a list of prerequisites to be perform
 
 2. For any non-GA modules (e.g., Spanner), also comment them out from the `pom.xml` file.
 
-3. Manually change the value of `TRACKING_HEADER_PROJECT_VERSION` in link:spring-cloud-gcp-core/src/main/java/org/springframework/cloud/gcp/core/UsageTrackingHeaderProvider.java[`UsageTrackingHeaderProvider`] to the version being released.
+3. Manually change the value of `TRACKING_HEADER_PROJECT_VERSION` in link:spring-cloud-gcp-core/src/main/java/org/springframework/cloud/gcp/core/UsageTrackingHeaderProvider.java[`UsageTrackingHeaderProvider`] to the version being released from the interim `SNAPSHOT` value.
```
artembilan: Well, actually it is called `BUILD-SNAPSHOT` do not mislead a consumer of this document.

ChengyuanZhao: Changed. 

I thought it would be clear which value I am referring to, since there are also numbers in front of the "BUILD-" that we cannot mention here.

RELEASING.adoc
```diff
@@ -0,0 +1,14 @@
...
+The actual release process is triggered by Pivotal.
+In order to simplify that process, here is a list of prerequisites to be performed before asking Pivotal to run the release scripts.
+
+1. Comment out the `spring-cloud-gcp-samples` module from the root `pom.xml` file, so that the samples aren't released.
+
+2. For any non-GA modules (e.g., Spanner), also comment them out from the `pom.xml` file.
+
+3. Manually change the value of `TRACKING_HEADER_PROJECT_VERSION` in link:spring-cloud-gcp-core/src/main/java/org/springframework/cloud/gcp/core/UsageTrackingHeaderProvider.java[`UsageTrackingHeaderProvider`] to the version being released from the interim `BUILD-SNAPSHOT` value.
+
+4. Change the value of the `<version>` elements of the `pom.xml` of each module in the project being released to the name of the version being released from the interim `BUILD-SNAPSHOT` values.
```
meltsufin: We can even suggest using the [`mvn versions:set`](https://www.mojohaus.org/versions-maven-plugin/set-mojo.html).

ChengyuanZhao:
@meltsufin made your change.

PR #756, opened 2018-06-15
Reuse PubSubTemplate conversion

joaoandremartins:
PubSubTemplate already offers type conversion to/from JSON.
On top of that, we found that the current MessageConverters are not
only very expressive, but they don't provide the kind of conversion we
need (arguably, the don't provide any conversion, but rather message
building/mapping).

### DIFF
--- a/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapter.java
+++ b/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapter.java
@@ -29,9 +29,6 @@ import org.springframework.cloud.gcp.pubsub.support.GcpPubSubHeaders;
 import org.springframework.integration.endpoint.MessageProducerSupport;
 import org.springframework.integration.mapping.HeaderMapper;
 import org.springframework.integration.support.MessageBuilder;
-import org.springframework.messaging.Message;
-import org.springframework.messaging.MessageHeaders;
-import org.springframework.messaging.converter.MessageConverter;
 import org.springframework.util.Assert;
 
 /**
@@ -50,8 +47,6 @@ public class PubSubInboundChannelAdapter extends MessageProducerSupport {
 	private Subscriber subscriber;
 
 	private AckMode ackMode = AckMode.AUTO;
-
-	private MessageConverter messageConverter;
 
 	private HeaderMapper<Map<String, String>> headerMapper = new PubSubHeaderMapper();
 
@@ -78,16 +73,9 @@ public class PubSubInboundChannelAdapter extends MessageProducerSupport {
 		}
 
 		try {
-			Message message;
-			if (this.messageConverter != null) {
-				message = this.messageConverter.toMessage(pubsubMessage.getData().toByteArray(),
-						new MessageHeaders(messageHeaders));
-			}
-			else {
-				message = MessageBuilder.withPayload(pubsubMessage.getData().toByteArray())
-						.copyHeaders(messageHeaders).build();
-			}
-			sendMessage(message);
+			sendMessage(
+					MessageBuilder.withPayload(pubsubMessage.getData().toByteArray())
+							.copyHeaders(messageHeaders).build());
 		}
 		catch (RuntimeException re) {
 			if (this.ackMode == AckMode.AUTO) {
@@ -117,20 +105,6 @@ public class PubSubInboundChannelAdapter extends MessageProducerSupport {
 	public void setAckMode(AckMode ackMode) {
 		Assert.notNull(ackMode, "The acknowledgement mode can't be null.");
 		this.ackMode = ackMode;
-	}
-
-	public MessageConverter getMessageConverter() {
-		return this.messageConverter;
-	}
-
-	/**
-	 * Set the {@link MessageConverter} to convert an incoming Pub/Sub message payload to an
-	 * {@code Object}. If the converter is null, the payload is unchanged.
-	 * @param messageConverter converts a {@link PubsubMessage} payload to a
-	 * {@link org.springframework.messaging.Message} payload. Can be set to null.
-	 */
-	public void setMessageConverter(MessageConverter messageConverter) {
-		this.messageConverter = messageConverter;
 	}
 
 	/**
--- a/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapter.java
+++ b/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapter.java
@@ -22,6 +22,7 @@ import com.google.cloud.pubsub.v1.AckReplyConsumer;
 import com.google.cloud.pubsub.v1.Subscriber;
 import com.google.pubsub.v1.PubsubMessage;
 
+import org.springframework.cloud.gcp.pubsub.core.PubSubException;
 import org.springframework.cloud.gcp.pubsub.core.PubSubTemplate;
 import org.springframework.cloud.gcp.pubsub.integration.AckMode;
 import org.springframework.cloud.gcp.pubsub.integration.PubSubHeaderMapper;
@@ -80,11 +81,11 @@ public class PubSubInboundChannelAdapter extends MessageProducerSupport {
 					.copyHeaders(messageHeaders)
 					.build());
 		}
-		catch (Exception re) {
+		catch (RuntimeException re) {
 			if (this.ackMode == AckMode.AUTO) {
 				consumer.nack();
 			}
-			throw new RuntimeException(re);
+			throw new PubSubException("Spring Message construction from Pub/Sub message failed.", re);
 		}
 
 		if (this.ackMode == AckMode.AUTO) {
--- a/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapter.java
+++ b/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapter.java
@@ -85,7 +85,7 @@ public class PubSubInboundChannelAdapter extends MessageProducerSupport {
 			if (this.ackMode == AckMode.AUTO) {
 				consumer.nack();
 			}
-			throw new PubSubException("Spring Message construction from Pub/Sub message failed.", re);
+			throw new PubSubException("Sending Spring message failed.", re);
 		}
 
 		if (this.ackMode == AckMode.AUTO) {
--- a/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapter.java
+++ b/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapter.java
@@ -22,15 +22,13 @@ import com.google.cloud.pubsub.v1.AckReplyConsumer;
 import com.google.cloud.pubsub.v1.Subscriber;
 import com.google.pubsub.v1.PubsubMessage;
 
-import org.springframework.cloud.gcp.pubsub.core.PubSubOperations;
+import org.springframework.cloud.gcp.pubsub.core.PubSubTemplate;
 import org.springframework.cloud.gcp.pubsub.integration.AckMode;
 import org.springframework.cloud.gcp.pubsub.integration.PubSubHeaderMapper;
 import org.springframework.cloud.gcp.pubsub.support.GcpPubSubHeaders;
 import org.springframework.integration.endpoint.MessageProducerSupport;
 import org.springframework.integration.mapping.HeaderMapper;
-import org.springframework.integration.mapping.InboundMessageMapper;
 import org.springframework.integration.support.MessageBuilder;
-import org.springframework.messaging.Message;
 import org.springframework.util.Assert;
 
 /**
@@ -44,7 +42,7 @@ public class PubSubInboundChannelAdapter extends MessageProducerSupport {
 
 	private final String subscriptionName;
 
-	private final PubSubOperations pubSubTemplate;
+	private final PubSubTemplate pubSubTemplate;
 
 	private Subscriber subscriber;
 
@@ -52,9 +50,9 @@ public class PubSubInboundChannelAdapter extends MessageProducerSupport {
 
 	private HeaderMapper<Map<String, String>> headerMapper = new PubSubHeaderMapper();
 
-	private InboundMessageMapper<byte[]> messageConverter;
+	private Class payloadType = byte[].class;
 
-	public PubSubInboundChannelAdapter(PubSubOperations pubSubTemplate, String subscriptionName) {
+	public PubSubInboundChannelAdapter(PubSubTemplate pubSubTemplate, String subscriptionName) {
 		this.pubSubTemplate = pubSubTemplate;
 		this.subscriptionName = subscriptionName;
 	}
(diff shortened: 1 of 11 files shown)


spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapter.java
```diff
@@ -78,16 +73,9 @@ private void receiveMessage(PubsubMessage pubsubMessage, AckReplyConsumer consum
...
-				message = this.messageConverter.toMessage(pubsubMessage.getData().toByteArray(),
-						new MessageHeaders(messageHeaders));
-			}
-			else {
-				message = MessageBuilder.withPayload(pubsubMessage.getData().toByteArray())
-						.copyHeaders(messageHeaders).build();
-			}
-			sendMessage(message);
+			sendMessage(
+					MessageBuilder.withPayload(pubsubMessage.getData().toByteArray())
```
artembilan: Well, and this way you do nothing, but just send a `byte[]` as a payload. In other words you don't do any conversions at all.
Is it what we would like to do here on the *inbound* side?

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/outbound/PubSubMessageHandler.java
```diff
@@ -88,33 +78,10 @@ protected void handleMessageInternal(Message<?> message) throws Exception {
...
-			}
+		Map<String, String> headers = new HashMap<>();
+		this.headerMapper.fromHeaders(message.getHeaders(), headers);
 
-			Map<String, String> headers = new HashMap<>();
-			this.headerMapper.fromHeaders(message.getHeaders(), headers);
-
-			pubsubFuture = this.pubSubTemplate.publish(topic, pubsubPayload, headers);
-		}
+		pubsubFuture = this.pubSubTemplate.publish(topic, payload, headers);
```
artembilan: This side is fine with me as far as end-user can supply any custom `PubSubMessageConverter` into the target `PubSubTemplate`.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapter.java
joaoandremartins: Not quite. No conversions **on the adapter itself**, we just reuse `PubSubTemplate`'s conversion, instead of literally duplicating it. (Now you're not looking at my commit message ;) )

If we ever want to have conversion happening in the adapter, I think we should start using those `InboundMessageConverter`/`OutboundMessageConverter`. But, for now, I believe this is the better state.

artembilan: ??? Show me, please, where is the conversion here `MessageBuilder.withPayload(pubsubMessage.getData().toByteArray())`.
You receive a `PubsubMessage` and get its `body` as a `byte[]`. 
What am I missing?

Your change doesn't reflect what you say in commit message. 🤷‍♂️ 

Sorry, if I'm so ignorant, but the change doesn't look like flexible and the goal we are trying to achieve with the conversion.

joaoandremartins: This is embarrassing, but I thought I was commenting on `PubSubMessageHandler` :no_mouth:

joaoandremartins:
I'll delegate to @meltsufin to remove the `InboundMessageMapper` from the inbound adapter if we ever add it to `PubSubTemplate`.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapter.java
```diff
@@ -51,10 +50,10 @@
 
 	private AckMode ackMode = AckMode.AUTO;
 
-	private MessageConverter messageConverter;
-
 	private HeaderMapper<Map<String, String>> headerMapper = new PubSubHeaderMapper();
 
+	private InboundMessageMapper<byte[]> messageConverter;
```
artembilan: Let's call it `messageMapper` then!

OTOH I won't mind if we evolve our `PubSubMessageConverter` abstraction and will really reuse it here as well, even if there is nothing to do with the `PubSubTemplate` here in the `PubSubInboundChannelAdapter`.

At least that is how we do in other places like Kafka or AMQP, or even JMS.

That's really embarrassing from my side that I led you to that generics `MessageConverter`  in the beginning: i just wanted to reuse an existing abstraction as much as possible without extra code to support here, but as we all realized already it doesn't fit our model. So, let's try to look into some other existing (hope working) solution in the mentioned projects!

meltsufin: I think we can just add support in the `PubSubTemplate` for something more convenient than 
`com.google.cloud.pubsub.v1.MessageReceiver`, something that has 

```
void receiveMessage(Object payload, Map<String,String> headers, AckReplyConsumer consumer);
```
rather than
```
void receiveMessage(final PubsubMessage message, final AckReplyConsumer consumer);
```
This way we can remove conversion concerns from the inbound side as well.

artembilan: ??? That's something not related. `PubSubTemplate`  is for single, user-driven operations. Meanwhile this `PubSubInboundChannelAdapter` is really message-driven, based on events generated by the `Consumer` in the background thread. So, it isn't going to be so convenient to inject `PubSubTemplate` just for the conversion reasons.

meltsufin: I'm not sure I understand. `PubSubInboundChannelAdapter` already has the `PubSubTemplate` injected, and it relies on the `MessageReceiver`.
```java
	@Override
	protected void doStart() {
		super.doStart();

		this.subscriber =
				this.pubSubTemplate.subscribe(this.subscriptionName, this::receiveMessage);
	}

	private void receiveMessage(PubsubMessage pubsubMessage, AckReplyConsumer consumer) {
...
```

I'm just suggesting an enhancement to the `PubSubTemplate` that can also be taken advantage of in the adapter.

artembilan: Oh! Sorry. I really missed something in this project.
So, I see now that `pubSubTemplate` has that `subscribe()` operation. A bit unusual for the stateless component when the *template* abstraction should be just only for a use-driven operations, but let it be - you broke the approach and you are in responsibility to support it.
That wasn't my call to pull `subscribe` to the `pubSubTemplate`: https://github.com/spring-cloud/spring-cloud-gcp/pull/148

Nevertheless I disagree with the `receiveMessage(Object payload, Map<String,String> headers)` API. You still have a `HeaderMapper` in the Channel Adapter. Are you going to move that one to the template one day as well  ? 🤷‍♂️ 

For the conversion part you really can call a getter for converter from that template and still be good in the Channel Adapter.

With all of these discussions and massive reworks I feel like you would like to get rid of Spring Integration support at all and you try to pull as much of its functionality into one component as possible. Meanwhile we still should keep in mind the level of responsibility and loosely coupling concern.

meltsufin: I believe in the early discussions there was a drive to do all the hard work in the template, and have the channel adapters reuse the template as much as possible. I agree that may well violate the loose coupling and separation of concerns.

I really just want to keep message conversion in one place, be it the template or the channel adapters. If you think it belongs best in the channel adapters, then let's take it out of the template along with convenience methods like 
```java
	@Override
	public <T> ListenableFuture<String> publish(String topic, T payload) {
		return publish(topic, payload, null);
	}
```

where you can pass an arbitrary object rather than the `PubsubMessage`.
If you think that's better, I'm OK with it. I was never really a supporter of these convenience methods, and rather wanted to keep the template as minimal as possible.

artembilan: Well, actually that's fine to keep those convenient method. As long as we have similar API in Spring AMQP and Spring Kafka.

My only point is to get a `PubSubMessageConverter` from the template for conversion in the Channel Adapter since such a `subscribe()` is really out of template scope and that your `receive()` suggestion is a bit confusing. 

Well, it really confuses me more and more since template should expose only atomic stateless operations.

With the  `PubSubMessageConverter`  abstraction you have just introduced we really have a one-stop-shop for conversion which can be reused in the template as well as in the Channel Adapters.

Let's for now reuse it via getter from the template to complete this issue and we can revise the template architecture a bit later!

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapter.java
```diff
@@ -78,22 +75,16 @@ private void receiveMessage(PubsubMessage pubsubMessage, AckReplyConsumer consum
...
+					.copyHeaders(messageHeaders)
+					.build());
 		}
-		catch (RuntimeException re) {
+		catch (Exception re) {
 			if (this.ackMode == AckMode.AUTO) {
 				consumer.nack();
 			}
-			throw re;
+			throw new RuntimeException(re);
```
artembilan: May we throw `PubSubException` here to make it easier to trace logs for errors?

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/converter/PubSubMessageConverter.java
```diff
@@ -34,13 +34,13 @@
...
+	PubsubMessage toPubSubMessage(Object payload, Map<String, String> headers);
 
 	/**
 	 * Convert the payload of a given {@code PubsubMessage} to a desired Java type.
 	 * @param message the message containing the payload of the object
 	 * @param payloadType the desired type of the object
 	 * @return the object converted from the message's payload
 	 */
-	<T> T fromMessage(PubsubMessage message, Class<T> payloadType);
+	<T> T fromPubSubMessage(PubsubMessage message, Class<T> payloadType);
```
artembilan: I think we need to revise this contract to avoid generics.
See the same `MessageConverter` for example:
```
Object fromMessage(Message<?> message, Class<?> targetClass);
```

So, we provide target class to expect, but that doesn't mean that return is going to be in exactly expected type.

The point is that the mentioned Jackson's default typing is really ignoring provided type and consults for that the provided type info in the JSON. Something similar matters with standard Java deserialization: that really does not have any effect what type you provide since I'm just going to deserialize from bytes.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapter.java
meltsufin: Typically this would fail due to message conversion, not necessarily an issue with Pub/Sub per se, but we can do that. I also wounder if we should just make the `PubSubMessageConversionException` inherit from `PubSubException`.

artembilan: Well, that's not true. 
You see, you wrap here even `sendMessage()` and this involves the whole Spring Integration flow downstream. So, anything can happen there: starting with any `MessagingException` and ending with the `TransactionTimedOutException` and so on.

Therefore I would insist to wrap it into the generic `PubSubException` and proceed. Just because there downstream can be anything and not only conversion, but with this specific `PubSubException` we are going to make logs more traceable.

Am I missing anything?

artembilan: BTW, if you really want a fine-grained tracing, you really can make that `PubSubMessageConversionException` as a `PubSubException` and extract conversion to its own `try...catch()`, but again: the `sendMessage()` is a different story and has to be treated respectively.

meltsufin: I see. I'll just wrap into `PubSubException` for now like you suggested.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapter.java
```diff
@@ -80,11 +81,11 @@ private void receiveMessage(PubsubMessage pubsubMessage, AckReplyConsumer consum
 					.copyHeaders(messageHeaders)
 					.build());
 		}
-		catch (Exception re) {
+		catch (RuntimeException re) {
 			if (this.ackMode == AckMode.AUTO) {
 				consumer.nack();
 			}
-			throw new RuntimeException(re);
+			throw new PubSubException("Spring Message construction from Pub/Sub message failed.", re);
```
artembilan: Why "construction"?
I've explained in the previous comment that there can be anything downstream in the `sendMessage()`...

How transaction timeout is related to the "construction"? 🤷‍♂️

spring-cloud-gcp-pubsub/src/test/java/org/springframework/cloud/gcp/pubsub/support/converter/SimplePubSubMessageConverterTests.java
```diff
@@ -106,7 +106,7 @@ public void testNullHeaders() {
 		SimplePubSubMessageConverter converter = new SimplePubSubMessageConverter();
 
 		// test extraction from PubsubMessage to T
-		assertEquals(TEST_STRING, toString.convert(converter.fromPubSubMessage(PubsubMessage.newBuilder()
+		assertEquals(TEST_STRING, toString.convert((T) converter.fromPubSubMessage(PubsubMessage.newBuilder()
```
artembilan: I think we need to reflect reality in the test as well - and get rid of the generics here too!

meltsufin: This would cause me to do a bunch more casts in the individual tests. Is there sufficient benefit?

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/converter/PubSubMessageConverter.java
meltsufin: @artembilan Going back to this point, in the Jackson case, can't the client just call `fromPubSubMessage(message, Object.class)` or otherwise risk the `ClassCastException`?

artembilan: That's correct. That's exactly what happens in the `GenericJackson2JsonRedisSerializer` where I studied this approach from:
```
@Override
public Object deserialize(@Nullable byte[] source) throws SerializationException {
	return deserialize(source, Object.class);
}
```
And there is no any `ClassCastException`

meltsufin: The casting would be done in our `JacksonPubSubMessageConverter`.

meltsufin: I'm thinking of changing it back to how I had it before with the generics.

artembilan: Well, that's why I commented somewhere that there must not be any generics at all.
That's why  a `MessageConverter` has a contract like:
```
Object fromMessage(Message<?> message, Class<?> targetClass);
```
So, this really says us that even if we specify some type, that doesn't mean that a result object is going to be of that type.

meltsufin: I understand, but what's wrong with actually requiring the result object to match the specified type?

PR #759, opened 2018-06-15
Remove dependence on spring-integration-core in PubSubTemplate

meltsufin:
@artembilan WDYT?
We need to figure out what to do with JSON support in PubSubTemplate because it's creating the dependence on `spring-integration-core`.


spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
```diff
@@ -62,7 +59,7 @@
 
 	private static final Log LOGGER = LogFactory.getLog(PubSubTemplate.class);
 
-	private PubSubMessageConverter messageConverter = new JacksonPubSubMessageConverter();
+	private Converter<Object, ByteString> publishPayloadConverter;
```
artembilan: This fix is colliding with the #756 .
I believe we can stay with the current situation.

The `JacksonPubSubMessageConverter` can be resolved on demand when Spring Integration dependency is on class path.

You can add a classpath check static variable and only after that create a `JacksonPubSubMessageConverter` instance.
Something like this:
```
public static final boolean SPRING_INTEGRATION_PRESENT =
			ClassUtils.isPresent("org.springframework.integration.config.EnableIntegration",
					PubSubTemplate.class.getClassLoader());
```

meltsufin: If we insist that `PubSubTemplate` should be independent of Spring Integration, this seems to compromise that position. Besides, why is conversion to JSON so tightly coupled to Spring Integration? It seems like Spring should just have this functionality somewhere in core.

joaoandremartins: Strong +1 to being independent from SI in our core Pub/Sub module, especially if that functionally is not related to SI at all.

artembilan: That's easy. Just follow commit history: https://github.com/spring-cloud/spring-cloud-gcp/pull/659 !

I told there that we are going to be tied with Spring Integration...

meltsufin:
@artembilan I did some refactoring to make `JacksonPubSubMessageConverter` independent of `spring-integration-core`. I think we really just aught to delegate details of how users want to construct the `ObjectMapper` to them. Maybe they just want to use the `@JsonTypeInfo` to require the `@class` property instead of the default typing.
Also, now we have two implementations of the `PubSubMessageConverter`. I'm thinking of introducing a `spring.cloud.gcp.pubsub.message-converter` property to the autoconfig to switch between `json` and `simple`.
WDYT?

artembilan:
Thanks.

I will take a look at this tomorrow.

Not sure about the property. What if we introduce some other converter implementation which is not so easy to configure, but you already confuse people with that property. 

I remember we tried something with opinionated property for the Stackdrive extractors, but somehow we ended up with some other more robust solution. 

So, I suggest to stick with the `SimplePubSubMessageConverter` by default and switch to the Jackson one, if an `ObjectMapper` bean is present. This can simply be done with the particular dependency and existing auto-configuration on the matter: https://github.com/spring-projects/spring-boot/blob/v2.0.3.RELEASE/spring-boot-project/spring-boot-starters/spring-boot-starter-json/pom.xml

meltsufin:
We actually still use the property for Stackdriver logging extractors. :) Thanks for the reminder though. I think we need to take another look at that.

Regarding the `ObjectMapper` presence hinting that JSON converter is desired seems like a good idea, but we would need to clearly document that.

artembilan:
Right, of course.

To be honest that's not a choice to explicitly rely on the Jackson. There are rumors it is not so fast at the beginning, that we in Boot are looking to some other options instead. So, this way Jackson is still going to be an `optional` dependency and that's already end-user choice to make such a configuration explicitly auto-selected if Jackson is in the classpath.

spring-cloud-gcp-autoconfigure/pom.xml
```diff
@@ -62,7 +62,7 @@
         <dependency>
             <groupId>org.springframework.integration</groupId>
             <artifactId>spring-integration-core</artifactId>
-            <scope>test</scope>
+            <optional>true</optional>
```
artembilan: If we don't rely on Spring Integration anymore, do we still need this dependency here?

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
```diff
@@ -120,9 +121,14 @@ public ExecutorProvider subscriberExecutorProvider() {
 
 	@Bean
 	@ConditionalOnMissingBean
-	public PubSubMessageConverter pubSubMessageConverter() {
-		return new JacksonPubSubMessageConverter(
-				this.gcpPubSubProperties.getTrustedPackages());
+	public ObjectMapper objectMapper() {
```
artembilan: Sorry, but seems for me we talked with you that no an explicit `ObjectMapper` bean.
Everything should be deferred to end-user decision.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
```diff
@@ -120,9 +121,14 @@ public ExecutorProvider subscriberExecutorProvider() {
...
+		return new ObjectMapper();
+	}
+
+	@Bean
+	@ConditionalOnMissingBean
+	public JacksonPubSubMessageConverter pubSubMessageConverter(ObjectMapper objectMapper) {
```
artembilan: Therefore this one should be here only if there is an `ObjectMapper` bean.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/converter/JacksonPubSubMessageConverter.java
```diff
@@ -53,18 +46,32 @@ public JacksonPubSubMessageConverter(ObjectMapper objectMapper) {
...
+
+			if (headers != null) {
+				pubsubMessageBuilder.putAllAttributes(headers);
+			}
+
+			return pubsubMessageBuilder.build();
 		}
 		catch (JsonProcessingException e) {
-			throw new IOException("Jackson serialization of object failed: " + object, e);
+			throw new RuntimeException("JSON serialization of an object of type " +
```
artembilan: I suggest to throw here an explicit `MessageConversionException` instead. It is indeed the conversion issue. Isn't it?

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/converter/JacksonPubSubMessageConverter.java
```diff
@@ -53,18 +46,32 @@ public JacksonPubSubMessageConverter(ObjectMapper objectMapper) {
...
 	@Override
-	public <T> T fromPayload(ByteString payload, Class<T> payloadType)
-			throws IOException {
-		return this.objectMapper.readerFor(payloadType).readValue(payload.toByteArray());
+	public <T> T fromMessage(PubsubMessage message, Class<T> payloadType) {
+		try {
+			return this.objectMapper.readerFor(payloadType).readValue(message.getData().toByteArray());
+		}
+		catch (IOException e) {
+			throw new RuntimeException("JSON deserialization of an object of type " +
```
artembilan: The same here about `MessageConversionException`

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/converter/SimplePubSubMessageConverter.java
```diff
@@ -0,0 +1,90 @@
...
+ * @author Mike Eltsufin
+ */
+public class SimplePubSubMessageConverter implements PubSubMessageConverter {
+
+	private Charset charset = Charset.defaultCharset();
+
+	/**
+	 * Set charset for converting strings.
+	 */
+	public void setCharset(Charset charset) {
```
artembilan: Can't we have a `charset` as a ctor arg?
One ctor is without args and delegates to that one with the `charset`.
Much efficient during runtime since that `charset` propety is going to be `final` and it is much convenient to declare instances of this class.

Thanks

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/converter/SimplePubSubMessageConverter.java
```diff
@@ -0,0 +1,90 @@
...
+package org.springframework.cloud.gcp.pubsub.support.converter;
+
+import java.nio.charset.Charset;
+import java.util.Map;
+
+import com.google.protobuf.ByteString;
+import com.google.pubsub.v1.PubsubMessage;
+
+/**
+ * A simple {@link SimplePubSubMessageConverter} that that directly maps payloads of type
```
artembilan: Double `that`. 
I mean: we don't need double that.

Although I also see double `simple` 😄

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/converter/SimplePubSubMessageConverter.java
```diff
@@ -0,0 +1,90 @@
...
+			convertedPayload = ByteString.copyFrom(((String) payload).getBytes(this.charset));
+		}
+		else if (payload instanceof ByteString) {
+			convertedPayload = (ByteString) payload;
+		}
+		else if (payload instanceof byte[]) {
+			convertedPayload = ByteString.copyFrom((byte[]) payload);
+		}
+		else {
+			throw new IllegalArgumentException("Unable to convert payload of type "
```
artembilan: How about `ByteBuffer` support ?

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
```diff
@@ -120,15 +122,16 @@ public ExecutorProvider subscriberExecutorProvider() {
 	}
 
 	@Bean
+	@ConditionalOnBean(type = "com.fasterxml.jackson.databind.ObjectMapper")
```
meltsufin: @artembilan What am I doing wrong here? It doesn't seem to be working. I get 
`No qualifying bean of type 'org.springframework.cloud.gcp.pubsub.support.converter.PubSubMessageConverter' available: expected at least 1 bean which qualifies as autowire candidate. Dependency annotations: {}`.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/converter/JacksonPubSubMessageConverter.java
meltsufin: `MessageConversionException` belongs to `spring-messaging`. Do we want the template to depend on it?

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/converter/SimplePubSubMessageConverter.java
meltsufin: Done.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/converter/JacksonPubSubMessageConverter.java
artembilan: Well, that's fine. The `spring-messaging` is a generic dependency - there is nothing about separate Spring Integration project.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
artembilan: ??? You don't have an `ObjectMapper` bean in your application context, therefore this `JacksonPubSubMessageConverter` is not created as well. And your consumer `PubSubTemplate` fails to be inject because there is no any beans on the matter.

However I see you have there already `SimplePubSubMessageConverter...

So, what, where and how fails.

Maybe you can share more stack trace to determine?

Thanks

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/converter/JacksonPubSubMessageConverter.java
meltsufin: The description of `MessageConversionException` is `An exception raised by {@link MessageConverter} implementations.` Since we do not implement `MessageConverter`, I think it might be misleading.

artembilan: Yeah... Agreed - unfortunate to have such a JavaDocs there. Although it doesn't break a purpose and functionality of the exception. So, what do you suggest: introduce our own exception on the matter?

I just don't like such a generic `RuntimeException` just for the goal to hide checked exception. 😢

meltsufin: Our own will do. :)

artembilan:
Does it say anything to you?
```
googTests run: 11, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 1.504 sec <<< FAILURE!
testToByteBuffer(org.springframework.cloud.gcp.pubsub.support.converter.SimplePubSubMessageConverterTests)  Time elapsed: 0.005 sec  <<< ERROR!
java.lang.IllegalArgumentException: Unable to convert Pub/Sub message to payload of type java.nio.ByteBuffer.
	at org.springframework.cloud.gcp.pubsub.support.converter.SimplePubSubMessageConverter.fromMessage(SimplePubSubMessageConverter.java:92)
	at org.springframework.cloud.gcp.pubsub.support.converter.SimplePubSubMessageConverterTests.doToTestForType(SimplePubSubMessageConverterTests.java:109)
	at org.springframework.cloud.gcp.pubsub.support.converter.SimplePubSubMessageConverterTests.testToByteBuffer(SimplePubSubMessageConverterTests.java:76)
```

See Travis report.

meltsufin:
I fixed that. The autoconfig issue is still there though.

```
testSubscriberPullConfig(org.springframework.cloud.gcp.autoconfigure.pubsub.GcpPubSubEmulatorConfigurationTests)  Time elapsed: 0.022 sec  <<< ERROR!
java.lang.IllegalStateException: Unstarted application context org.springframework.boot.test.context.assertj.AssertableApplicationContext[startupFailure=org.springframework.beans.factory.UnsatisfiedDependencyException] failed to start
        at org.springframework.boot.test.context.assertj.AssertProviderApplicationContextInvocationHandler.getStartedApplicationContext(AssertProviderApplicationContextInvocationHandler.java:163)
        at org.springframework.boot.test.context.assertj.AssertProviderApplicationContextInvocationHandler.invokeApplicationContextMethod(AssertProviderApplicationContextInvocationHandler.java:154)
        at org.springframework.boot.test.context.assertj.AssertProviderApplicationContextInvocationHandler.invoke(AssertProviderApplicationContextInvocationHandler.java:86)
        at com.sun.proxy.$Proxy55.getBean(Unknown Source)
        at org.springframework.cloud.gcp.autoconfigure.pubsub.GcpPubSubEmulatorConfigurationTests.lambda$testSubscriberPullConfig$1(GcpPubSubEmulatorConfigurationTests.java:102)
        at org.springframework.boot.test.context.runner.AbstractApplicationContextRunner.accept(AbstractApplicationContextRunner.java:327)
        at org.springframework.boot.test.context.runner.AbstractApplicationContextRunner.lambda$run$0(AbstractApplicationContextRunner.java:279)
        at org.springframework.boot.test.util.TestPropertyValues.applyToSystemProperties(TestPropertyValues.java:130)
        at org.springframework.boot.test.context.runner.AbstractApplicationContextRunner.run(AbstractApplicationContextRunner.java:277)
        at org.springframework.cloud.gcp.autoconfigure.pubsub.GcpPubSubEmulatorConfigurationTests.testSubscriberPullConfig(GcpPubSubEmulatorConfigurationTests.java:100)
        at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
        at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
        at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
        at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
        at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
        at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
        at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:57)
        at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
        at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
        at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
        at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
        at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)
        at org.junit.runners.ParentRunner.run(ParentRunner.java:363)
        at org.apache.maven.surefire.junit4.JUnit4Provider.execute(JUnit4Provider.java:252)
        at org.apache.maven.surefire.junit4.JUnit4Provider.executeTestSet(JUnit4Provider.java:141)
        at org.apache.maven.surefire.junit4.JUnit4Provider.invoke(JUnit4Provider.java:112)
        at org.apache.maven.surefire.util.ReflectionUtils.invokeMethodWithArray(ReflectionUtils.java:189)
        at org.apache.maven.surefire.booter.ProviderFactory$ProviderProxy.invoke(ProviderFactory.java:165)
        at org.apache.maven.surefire.booter.ProviderFactory.invokeProvider(ProviderFactory.java:85)
        at org.apache.maven.surefire.booter.ForkedBooter.runSuitesInProcess(ForkedBooter.java:115)
        at org.apache.maven.surefire.booter.ForkedBooter.main(ForkedBooter.java:75)
Caused by: org.springframework.beans.factory.UnsatisfiedDependencyException: Error creating bean with name 'pubSubTemplate' defined in org.springframework.cloud.gcp.autoconfigure.pubsub.GcpPubSubAutoConfiguration: Unsatisfied dependency expressed through method 'pubSubTemplate' parameter 2; nested exception is org.springframework.beans.factory.NoSuchBeanDefinitionException: No qualifying bean of type 'org.springframework.cloud.gcp.pubsub.support.converter.PubSubMessageConverter' available: expected at least 1 bean which qualifies as autowire candidate. Dependency annotations: {}
        at org.springframework.beans.factory.support.ConstructorResolver.createArgumentArray(ConstructorResolver.java:729)
        at org.springframework.beans.factory.support.ConstructorResolver.instantiateUsingFactoryMethod(ConstructorResolver.java:470)
        at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.instantiateUsingFactoryMethod(AbstractAutowireCapableBeanFactory.java:1254)
        at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.createBeanInstance(AbstractAutowireCapableBeanFactory.java:1103)
        at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.doCreateBean(AbstractAutowireCapableBeanFactory.java:541)
        at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.createBean(AbstractAutowireCapableBeanFactory.java:501)
        at org.springframework.beans.factory.support.AbstractBeanFactory.lambda$doGetBean$0(AbstractBeanFactory.java:317)
        at org.springframework.beans.factory.support.DefaultSingletonBeanRegistry.getSingleton(DefaultSingletonBeanRegistry.java:228)
        at org.springframework.beans.factory.support.AbstractBeanFactory.doGetBean(AbstractBeanFactory.java:315)
        at org.springframework.beans.factory.support.AbstractBeanFactory.getBean(AbstractBeanFactory.java:199)
        at org.springframework.beans.factory.support.DefaultListableBeanFactory.preInstantiateSingletons(DefaultListableBeanFactory.java:760)
        at org.springframework.context.support.AbstractApplicationContext.finishBeanFactoryInitialization(AbstractApplicationContext.java:869)
        at org.springframework.context.support.AbstractApplicationContext.refresh(AbstractApplicationContext.java:550)
        at org.springframework.boot.test.context.runner.AbstractApplicationContextRunner.configureContext(AbstractApplicationContextRunner.java:322)
        at org.springframework.boot.test.context.runner.AbstractApplicationContextRunner.createAndLoadContext(AbstractApplicationContextRunner.java:299)
        at org.springframework.boot.test.context.assertj.AssertProviderApplicationContextInvocationHandler.getContextOrStartupFailure(AssertProviderApplicationContextInvocationHandler.java:62)
        at org.springframework.boot.test.context.assertj.AssertProviderApplicationContextInvocationHandler.<init>(AssertProviderApplicationContextInvocationHandler.java:49)
        at org.springframework.boot.test.context.assertj.ApplicationContextAssertProvider.get(ApplicationContextAssertProvider.java:111)
        at org.springframework.boot.test.context.runner.AbstractApplicationContextRunner.createAssertableContext(AbstractApplicationContextRunner.java:292)
        at org.springframework.boot.test.context.runner.AbstractApplicationContextRunner.lambda$run$0(AbstractApplicationContextRunner.java:278)
        ... 32 more
Caused by: org.springframework.beans.factory.NoSuchBeanDefinitionException: No qualifying bean of type 'org.springframework.cloud.gcp.pubsub.support.converter.PubSubMessageConverter' available: expected at least 1 bean which qualifies as autowire candidate. Dependency annotations: {}
        at org.springframework.beans.factory.support.DefaultListableBeanFactory.raiseNoMatchingBeanFound(DefaultListableBeanFactory.java:1509)
        at org.springframework.beans.factory.support.DefaultListableBeanFactory.doResolveDependency(DefaultListableBeanFactory.java:1104)
        at org.springframework.beans.factory.support.DefaultListableBeanFactory.resolveDependency(DefaultListableBeanFactory.java:1065)
        at org.springframework.beans.factory.support.ConstructorResolver.resolveAutowiredArgument(ConstructorResolver.java:815)
        at org.springframework.beans.factory.support.ConstructorResolver.createArgumentArray(ConstructorResolver.java:721)
        ... 51 more
```

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
```diff
@@ -119,10 +122,16 @@ public ExecutorProvider subscriberExecutorProvider() {
...
-	public PubSubMessageConverter pubSubMessageConverter() {
-		return new JacksonPubSubMessageConverter(
-				this.gcpPubSubProperties.getTrustedPackages());
+	public JacksonPubSubMessageConverter pubSubMessageConverter(ObjectMapper objectMapper) {
+		return new JacksonPubSubMessageConverter(objectMapper);
+	}
+
+	@Bean
+	@ConditionalOnMissingBean
+	public SimplePubSubMessageConverter pubSubMessageConverter() {
```
artembilan: How does it work if you change both return types to the interface  - `PubSubMessageConverter` ?

meltsufin: I thought I tested that before, but now it seems to work.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
artembilan: Why do you use a type name instead of class?
It doesn't bring me too much sense since you are going to use this type explicitly in the bean method argument.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
```diff
@@ -118,11 +121,17 @@ public ExecutorProvider subscriberExecutorProvider() {
...
+	public PubSubMessageConverter pubSubMessageConverter(ObjectMapper objectMapper) {
+		return new JacksonPubSubMessageConverter(objectMapper);
+	}
+
 	@Bean
 	@ConditionalOnMissingBean
 	public PubSubMessageConverter pubSubMessageConverter() {
-		return new JacksonPubSubMessageConverter(
-				this.gcpPubSubProperties.getTrustedPackages());
+		return new SimplePubSubMessageConverter();
```
artembilan: Hm... Do we really need a bean for this variant?
I see it is a default instance in the target `PubSubTemplate`.
So, maybe we really can live without a default bean?
The `PubSubTemplate` bean has to be fixed for the `ObjectProvider<PubSubMessageConverter>` instead though...

spring-cloud-gcp-autoconfigure/pom.xml
meltsufin: Looks like it's needed in the tests. Would you prefer to have `scope=test`?

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
meltsufin: Sounds good.

spring-cloud-gcp-autoconfigure/pom.xml
artembilan: Correct. `test` is better then.

meltsufin:
@artembilan Should be ready for final review. I will tackle completing #756 after this is merged.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
```diff
@@ -118,20 +121,27 @@ public ExecutorProvider subscriberExecutorProvider() {
 				this.gcpPubSubProperties.getSubscriber().getExecutorThreads()));
 	}
 
+	@Bean
+	@ConditionalOnBean(type = "com.fasterxml.jackson.databind.ObjectMapper")
```
artembilan: Looks like my comment on some of your commit has lost: https://github.com/spring-cloud/spring-cloud-gcp/pull/759#discussion_r196527435

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
artembilan: You missed to remove this bean.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/converter/PubSubMessageConversionException.java
```diff
@@ -14,14 +14,21 @@
...
  */
 
-package org.springframework.cloud.gcp.pubsub.core.test.disallowed;
+package org.springframework.cloud.gcp.pubsub.support.converter;
+
+import org.springframework.core.NestedRuntimeException;
 
 /**
- * @author Chengyuan Zhao
+ * An exception raised by {@link PubSubMessageConverter} implementations.
```
artembilan: Let's don't repeat an error from the `spring-messaging`!
I mean let's don't mention `PubSubMessageConverter` at all since such an exception can be used in any conversion place, where Pub/Sub is involved.

That's why I would be good to still reuse that `MessageConversionException` in our classes, ever if that has such a misleading JavaDoc.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/converter/SimplePubSubMessageConverter.java
```diff
@@ -0,0 +1,101 @@
...
+			result = (T) message.getData();
+		}
+		else if (payloadType == String.class) {
+			result = (T) new String(payload, this.charset);
+		}
+		else if (payloadType == ByteBuffer.class) {
+			result = (T) ByteBuffer.wrap(message.getData().toByteArray());
+		}
+		else if (payloadType == byte[].class) {
+			result = (T) message.getData().toByteArray();
```
ChengyuanZhao: you've repeated `message.getData().toByteArray();` here  from line 80.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/converter/SimplePubSubMessageConverter.java
```diff
@@ -0,0 +1,101 @@
...
+		byte[] payload = message.getData().toByteArray();
+
+		if (payloadType == ByteString.class) {
+			result = (T) message.getData();
+		}
+		else if (payloadType == String.class) {
+			result = (T) new String(payload, this.charset);
+		}
+		else if (payloadType == ByteBuffer.class) {
+			result = (T) ByteBuffer.wrap(message.getData().toByteArray());
```
ChengyuanZhao: you've repeated `message.getData().toByteArray();` here  from line 80.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/converter/PubSubMessageConverter.java
```diff
@@ -16,58 +16,31 @@
...
-	 */
-	byte[] toPayload(Object object) throws IOException;
-
-	/**
-	 * Convert a payload from a PubSub message into an object of the desired type
-	 * @param payload the {@link ByteString} payload from the message
-	 * @param payloadType the desired type of the object
-	 * @return the object converted from the payload of the message
-	 */
-	<T> T fromPayload(ByteString payload, Class<T> payloadType) throws IOException;
```
ChengyuanZhao: Why get rid of these two methods? By removing these two and the default implementations of toMessage and fromMessage you've now repeated the code for dealing with the headers and also message.getData() calls in both the jackson and simple message converter classes.

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubEmulatorConfigurationTests.java
```diff
@@ -43,8 +43,6 @@
 	private ApplicationContextRunner contextRunner = new ApplicationContextRunner()
 			.withPropertyValues("spring.cloud.gcp.pubsub.emulator-host=localhost:8085",
 					"spring.cloud.gcp.projectId=test-project",
-					"spring.cloud.gcp.pubsub.trusted-packages[0]=a",
```
ChengyuanZhao: For all the exposing-config related work we've pushed for recently, it feels like a big step in the opposite direction to take away this trusted packages config functionality. 

From reading it seems like JSON Jackson deserialization is never to be used for serious business without trusted packages settings, and taking this away really feels like we're leaving the user out in the cold providing and configuring his or her own objectmapper.

spring-cloud-gcp-docs/src/main/asciidoc/pubsub.adoc
```diff
@@ -64,8 +58,9 @@ public void publishMessage() {
 }
 ----
 
-The default serialization and deserialization of POJOs uses Jackson JSON via the `JacksonMessageConverter` class.
-To set a new message converter in `PubSubTemplate`, use the `setMessageConverter()` method.
+By default, the `SimplePubSubMessageConverter` is used to convert payloads of type `byte[]`, `ByteString`, `ByteBuffer`, and `String` to Pub/Sub messages.
+
+For serialization and deserialization of POJOs using Jackson JSON, configure the `PubSubTemplate` to use the `JacksonPubSubMessageConverter` by using the `setMessageConverter()` method.
```
ChengyuanZhao: You tell the user to manually set their own Jackson converter here, but in the autoconfig you provide a bean for Jackson converter using an object mapper. The combination of these two implies that you want to user to provide a bean for his or her own objectmapper, but not a jackson converter, and then you want the user to get the jackson converter from the context manually and set it in the template class manually?  Why not set it up in such a way that if the user has provided an objectmapper then the jackson converter is automatically used?

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/converter/PubSubMessageConverter.java
meltsufin: I think it's better handled by a utility class if we need this.
I feel it's best to keep the interface as minimal as possible. There are cases where you would want to override `toMessage` and `fromMessage`, and it would be additional burden for those implementation to also implement `toPayload` and `fromPayload`.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/converter/SimplePubSubMessageConverter.java
meltsufin: good catch!

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubEmulatorConfigurationTests.java
meltsufin: There so many ways to do serialization that I don't think we need to over-prescribe it. Even on the surface the property doesn't make sense to be part `spring.cloud.gcp.pubsub.*` because it has nothing to do with Pub/Sub. Furthermore, the Jackson documentation encourages the use of the annotations instead, like I did in the `AllowedPayload`.
Lastly, this was Artem's suggestion to introduce `trusted-packages` and he doesn't seem to mind this change.

spring-cloud-gcp-docs/src/main/asciidoc/pubsub.adoc
meltsufin: You're right, I need to add some clarification here.

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubEmulatorConfigurationTests.java
artembilan: Right, I'm fine to dropping this one, since it definitely is out of this project scope.

Or annotations way or provided `ObjectMapper` instance configuration.
Since we have came to the situation when we would like do not depend here on Spring Integration, we don't have choice unless just document it with an appropriate sample, e.g. pointing to that Spring Integration's `JacksonJsonUtils` and so on.

We have many other similar approaches in Spring Kafka and Spring AMQP (maybe others) and I don't see that we expose there a `trusted packages` property. So, we are going to be good here to live without that as well.

Sorry for confusing in the begging: this project is not my full-time responsibility to think about all the features properly 😢

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/converter/PubSubMessageConverter.java
ChengyuanZhao: Could you elaborate? I don't follow. Anybody implementing this interface necessarily implements the functionality of both To/From message and To/From Payload. The difference that we're discussing is whether that is split into 2 methods or 4 methods.  In fact, reusing 2 out of those 4 methods reduces burden. Currently it looks like Jackson and Simple both repeat what was in the to/from Message methods' default implementations, so that actually is the additional burden.

meltsufin: If you imagine a converter that relies on headers to make the decision, for example, one that uses something like a mime-type header, 2 of the 4 methods would just become a burden. As a best practice, interfaces should always be kept as minimal as possible, preferably even limited to a single method to be compatible with lambdas.

artembilan: Well, I'd prefer to keep only two, message-based methods. Even if our out-of-the-box implementations have some similar code which may go to the `default` methods, there is no guarantee that some other implementations won't have fully different logic, so end-user will have to implement 4 methods instead of 2, where those `to(from)Payload()` may be just with empty bodies because they are not going to be called from others.

Plus with 4 methods there can be some confuse which method to choose to use in the API consumer.

meltsufin:
@artembilan Do you have any more unaddressed concerns on this PR?

PR #762, opened 2018-06-19
Fixed link to enable API runtime config

ChengyuanZhao:


spring-cloud-gcp-samples/spring-cloud-gcp-config-sample/README.adoc
```diff
@@ -63,4 +62,4 @@ $ gcloud beta runtime-config configs variables set \
 ....
 $ curl -XPOST http://localhost:8080/actuator/refresh
 ....
-10. Verify the values are updated at http://localhost:8080/
\ No newline at end of file
+10. Verify the values are updated at http://localhost:8080/
```
meltsufin: what changed here?

ChengyuanZhao: Nothing as far as I know. I did make this change using the in-browser editor, so maybe that caused this?

artembilan: Well, there may be just came a new line symbol which could be missed in the previous file version: https://github.com/spring-cloud/spring-cloud-gcp/pull/748/files

PR #764, opened 2018-06-20
remove trace id extractors options

### LINKED ISSUE #761: Remove all of the options for the TraceIdExtractor for logging
These extractors are now only relevant if someone is not using Sleuth but wants to have log grouping by trace id. In this case, we should only use the `XCLOUD` option because others simply don't make sense without Sleuth/Zipkin.

@saturnism FYI

ChengyuanZhao:
fixes #761


spring-cloud-gcp-docs/src/main/asciidoc/logging.adoc
```diff
@@ -39,17 +39,6 @@ For use in Web MVC-based applications, `TraceIdLoggingWebMvcInterceptor` is prov
 
 WARNING: If Spring Cloud GCP Trace is enabled, the logging module disables itself and delegates log correlation to Spring Cloud Sleuth.
 
-The behavior of the trace ID extraction can be configured with the `spring.cloud.gcp.logging.traceIdExtractor` property:
-
-[options="header",]
-|=======================================================================
-| Setting | Behavior
-| `XCLOUD` | Checks the `x-cloud-trace-context` HTTP header
```
meltsufin: We should probably still mention where the trace ID will be picked up from -- `x-cloud-trace-context`.
And if Sleuth is used, then the trace ID will be picked up from the MDC.

ChengyuanZhao: done

PR #770, opened 2018-06-21
Add integration test for the JPA sample

meltsufin:


spring-cloud-gcp-samples/spring-cloud-gcp-data-jpa-sample/src/test/java/com/example/ApplicationTests.java
```diff
@@ -0,0 +1,55 @@
...
+import static org.junit.Assert.assertTrue;
+
+/**
+ * @author Mike Eltsufin
+ */
+@RunWith(SpringRunner.class)
+@SpringBootTest(classes = {
+		DemoApplication.class }, properties = {
+				"spring.cloud.gcp.sql.database-name=houses",
+				"spring.cloud.gcp.sql.instance-connection-name=spring-cloud-gcp-ci:us-central1:testmysql",
```
meltsufin: I need to figure out how to keep these configuration properties in the `travis.yaml` instead of here.

spring-cloud-gcp-samples/spring-cloud-gcp-data-jpa-sample/src/test/java/com/example/ApplicationTests.java
```diff
@@ -26,18 +27,33 @@
 import org.springframework.boot.test.rule.OutputCapture;
 import org.springframework.test.context.junit4.SpringRunner;
 
+import static org.hamcrest.Matchers.is;
 import static org.junit.Assert.assertTrue;
+import static org.junit.Assume.assumeThat;
 
 /**
  * @author Mike Eltsufin
```
meltsufin: Add your name here please.

.travis.yml
```diff
@@ -22,7 +22,11 @@ before_install:
   - if [ "$TRAVIS_SECURE_ENV_VARS" != "false" ]; then
         openssl aes-256-cbc -K $encrypted_1ef8dfbdb114_key -iv $encrypted_1ef8dfbdb114_iv -in travis.tar.gz.enc -out travis.tar.gz -d;
         tar -xzf travis.tar.gz;
-        export INTEGRATION_TEST_FLAGS="-Dit.spanner=true -Dit.storage=true -Dit.config=true -Dit.pubsub=true -Dit.logging=true";
+        export INTEGRATION_TEST_FLAGS="-Dit.spanner=true -Dit.storage=true -Dit.config=true -Dit.pubsub=true -Dit.logging=true
+            -Dit.jpa-sample=true
+            -Dspring.cloud.gcp.sql.database-name=houses
+            -Dspring.cloud.gcp.sql.instance-connection-name=spring-cloud-gcp-ci:us-central1:testmysql
```
ChengyuanZhao: I think we should stay consistent and choose one way to pass these in, and I prefer the way we have this in the other integration tests: https://github.com/spring-cloud/spring-cloud-gcp/blob/master/spring-cloud-gcp-data-spanner/src/test/resources/application-test.properties

The downside of this travis file method is that we can't use this if we decide to run the tests from intellij or mvn commandline if we don't also type these out in the command

.travis.yml
```diff
@@ -22,7 +22,11 @@ before_install:
   - if [ "$TRAVIS_SECURE_ENV_VARS" != "false" ]; then
         openssl aes-256-cbc -K $encrypted_1ef8dfbdb114_key -iv $encrypted_1ef8dfbdb114_iv -in travis.tar.gz.enc -out travis.tar.gz -d;
         tar -xzf travis.tar.gz;
-        export INTEGRATION_TEST_FLAGS="-Dit.spanner=true -Dit.storage=true -Dit.config=true -Dit.pubsub=true -Dit.logging=true";
+        export INTEGRATION_TEST_FLAGS="-Dit.spanner=true -Dit.storage=true -Dit.config=true -Dit.pubsub=true -Dit.logging=true
+            -Dit.jpa-sample=true
```
meltsufin: Can we just use `-Dit.cloudsql`? This would indicate that the intent is to run all Cloud SQL tests, and it would be consistent with the other `it.*` flags.

.travis.yml
meltsufin: I would be OK with `spring.cloud.gcp.sql.database-name=houses` being in a configuration file, but here I really don't want to mention the CI project name, `spring-cloud-gcp-ci`, in the tests code.

ChengyuanZhao: My opinion is that it's ok to have that project name / instance-connection-name in an application-test.properties file buried deep in the test directories. I don't think it's any more visible than this travis.yml file at the very top directory :/ 

It's up to you!

meltsufin: It's not only about visibility, it's also about the fact that external users won't have access to this project and won't be able to run the test even if they followed the same naming conventions as described in the README.

ChengyuanZhao: I don't think I follow; isn't what you say true regardless of what method we provide these settings ? (in application-test.properties or travis.yml)

ChengyuanZhao: This also constitutes an argument to change our other integration tests that use GCP-global config values to use travis.yml to provide those settings, correct? (for example the bucket for storage integration tests). should we make an issue of that then ?

meltsufin: Yep we should have an issue for that too.

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/core/cloudfoundry/GcpCloudFoundryEnvironmentPostProcessorTests.java
```diff
@@ -46,6 +46,7 @@
 			.withInitializer(context ->
 					new GcpCloudFoundryEnvironmentPostProcessor()
 							.postProcessEnvironment(context.getEnvironment(), null))
+			.withSystemProperties(new String[] {"spring.cloud.gcp.sql.instance-connection-name="})
```
ChengyuanZhao: Would we need to add lines like this for every setting for the other integration tests that we move from their application-test.properties files to travis.yml?

meltsufin:
LGTM, but I can't approve it myself.

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/core/cloudfoundry/GcpCloudFoundryEnvironmentPostProcessorTests.java
```diff
@@ -46,6 +46,11 @@
 			.withInitializer(context ->
 					new GcpCloudFoundryEnvironmentPostProcessor()
 							.postProcessEnvironment(context.getEnvironment(), null))
+			.withSystemProperties(new String[] {
+					"spring.cloud.gcp.sql.instance-connection-name=test-connection",
+					"spring.cloud.gcp.sql.database-name=test-dbname",
```
ChengyuanZhao: Can I get an answer to the question I had here previously?

artembilan:
Any ideas why Travis has failed for this PR?
Thanks

Issue #425, opened 2018-02-07
Join the Spring Cloud release train


spencergibb:
Finchley [is due to GA in March](https://github.com/spring-cloud/spring-cloud-release/milestones?direction=asc&sort=due_date&state=open)

If that is too early, we can shoot for the next train, Greenwich.

meltsufin:
Thanks for the heads up! March might be too early for us, but we'll discuss it.

spencergibb:
Not a problem

meltsufin:
Related: #951.
Currently targeting Spring Cloud Greenwich release train.

meltsufin:
We're in master of Spring Cloud Release now!
https://github.com/spring-cloud/spring-cloud-release/pull/102

PR #772, opened 2018-06-22
Upgrade to Finchley.RELEASE and Boot 2.0.3

### LINKED ISSUE #726: Upgrade to Finchley.RELEASE when it becomes available

meltsufin:
Fixes #726.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -6,7 +6,7 @@
 	<parent>
 		<groupId>org.springframework.cloud</groupId>
 		<artifactId>spring-cloud-build</artifactId>
-		<version>2.0.0.RELEASE</version>
+		<version>2.0.2.RELEASE</version>
 		<relativePath/><!-- lookup parent from repository -->
 	</parent>
 
--- a/pom.xml
+++ b/pom.xml
@@ -40,7 +40,6 @@
 		<maven.compiler.source>1.8</maven.compiler.source>
 		<maven.compiler.target>1.8</maven.compiler.target>
 		<spring-cloud-release.version>Finchley.RELEASE</spring-cloud-release.version>
-		<spring-boot-release.version>2.0.3.RELEASE</spring-boot-release.version>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -39,8 +39,8 @@
 	<properties>
 		<maven.compiler.source>1.8</maven.compiler.source>
 		<maven.compiler.target>1.8</maven.compiler.target>
-		<spring-cloud-release.version>Finchley.RC2</spring-cloud-release.version>
-		<spring-boot-release.version>2.0.2.RELEASE</spring-boot-release.version>
+		<spring-cloud-release.version>Finchley.RELEASE</spring-cloud-release.version>
```
artembilan: You know, according the architecture of the Spring Cloud train, we can't rely on the Spring Cloud BOM, just because eventually we are going to be a part of that train #425.

Therefore we need to consider some other solution like use BOM from the particular Spring Cloud dependencies, for example for the `spring-cloud-context` we have to use `spring-cloud-commons-dependencies`: https://github.com/spring-cloud/spring-cloud-aws/pull/347 and so.

Also we should consider somehow do not use `spring-boot` version explicitly. I somehow believe that `spring-cloud-build` parent does the trick for us

meltsufin: I added #773 regarding your first comment.
I'd rather not handle this right now though.
The latest `spring-cloud-build` specifies version 2.0.2 of Spring Boot, which is not latest.

artembilan: Re. Spring Boot version: well, unfortunate, agreed,. but that's not the way all other Spring Cloud projects go...
So, if we are going to be a part of the Spring Cloud Train anyway, we have to follow rules there and be aligned with dependencies are dictated from there. 🤷‍♂️

meltsufin: Looks like we don't even need to specify the boot version. So, I removed that altogether.

PR #778, opened 2018-06-25
Add accept-once filter to SI storage sample

### LINKED ISSUE #777: integration-storage-sample - fix don't try transferring the same file more than once
I followed the instructions from the integration-storage-sample and seeing this error continually appearing:

```
The remote file 'Blob{bucket=test-test3, name=file1, generation=1529694509798668, size=4, 
content-type=application/octet-stream, metadata=null}' has not been transferred to the existing local file
```

Looks like the file is delivered to the local fs, and then for some reason, the app tries to re-write it.
The output bucket is filled with copies of the file.

meltsufin:
Fixes #777.

### DIFF
--- a/spring-cloud-gcp-samples/spring-cloud-gcp-integration-storage-sample/src/main/java/com/example/GcsSpringIntegrationApplication.java
+++ b/spring-cloud-gcp-samples/spring-cloud-gcp-integration-storage-sample/src/main/java/com/example/GcsSpringIntegrationApplication.java
@@ -29,7 +29,6 @@ import org.springframework.boot.SpringApplication;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
 import org.springframework.cloud.gcp.storage.integration.GcsRemoteFileTemplate;
 import org.springframework.cloud.gcp.storage.integration.GcsSessionFactory;
-import org.springframework.cloud.gcp.storage.integration.filters.GcsPersistentAcceptOnceFileListFilter;
 import org.springframework.cloud.gcp.storage.integration.inbound.GcsInboundFileSynchronizer;
 import org.springframework.cloud.gcp.storage.integration.inbound.GcsInboundFileSynchronizingMessageSource;
 import org.springframework.cloud.gcp.storage.integration.inbound.GcsStreamingMessageSource;
@@ -40,7 +39,7 @@ import org.springframework.integration.annotation.Poller;
 import org.springframework.integration.annotation.ServiceActivator;
 import org.springframework.integration.core.MessageSource;
 import org.springframework.integration.expression.ValueExpression;
-import org.springframework.integration.metadata.SimpleMetadataStore;
+import org.springframework.integration.file.FileHeaders;
 import org.springframework.messaging.MessageHandler;
 
 /**
@@ -129,6 +128,9 @@ public class GcsSpringIntegrationApplication {
 		GcsMessageHandler outboundChannelAdapter = new GcsMessageHandler(new GcsSessionFactory(gcs));
 		outboundChannelAdapter.setRemoteDirectoryExpression(
 				new ValueExpression<>(this.gcsWriteBucket));
+		outboundChannelAdapter
+				.setFileNameGenerator(message -> message.getHeaders().get(FileHeaders.REMOTE_FILE, String.class) + "."
+						+ System.currentTimeMillis());
 
 		return outboundChannelAdapter;
 	}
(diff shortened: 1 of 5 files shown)


artembilan:
@ChengyuanZhao ,

Please, see my comment in the issue how it has to be addressed by default.

I don't understand about your file names concern. The `AbstractRemoteFileStreamingMessageSource` provides this information in the headers:
```
return getMessageBuilderFactory()
					.withPayload(session.readRaw(remotePath))
					.setHeader(IntegrationMessageHeaderAccessor.CLOSEABLE_RESOURCE, session)
					.setHeader(FileHeaders.REMOTE_DIRECTORY, file.getRemoteDirectory())
					.setHeader(FileHeaders.REMOTE_FILE, file.getFilename())
					.setHeader(FileHeaders.REMOTE_FILE_INFO,
							this.fileInfoJson ? file.toJson() : file);
```

The `FileReadingMessageSource`, used by the `AbstractInboundFileSynchronizingMessageSource`, provides this information in headers:
```
return getMessageBuilderFactory()
				.withPayload(file)
				.setHeader(FileHeaders.RELATIVE_PATH,
						file.getAbsolutePath()
								.replaceFirst(Matcher.quoteReplacement(
										this.directory.getAbsolutePath() + File.separator), ""))
				.setHeader(FileHeaders.FILENAME, file.getName())
				.setHeader(FileHeaders.ORIGINAL_FILE, file);
```


Am I missing anything?

ChengyuanZhao:
@artembilan Gotcha, thank you for that info! Then i think we should demonstrate that the file name info is still there and can be used when writing to the output bucket in this sample, because in the sample currently it's not used and instead the files are written with random names.

artembilan:
Fine with me to address it here or in the separate PR, @ChengyuanZhao !

spring-cloud-gcp-samples/spring-cloud-gcp-integration-storage-sample/src/main/java/com/example/GcsSpringIntegrationApplication.java
```diff
@@ -85,20 +106,27 @@ public MessageHandler handleNewFileFromSynchronizer() {
...
+	/**
+	 * A service activator that connects to a channel with messages containing
+	 * {@code InputStream} payloads and copies the file data to a remote directory on GCS.
+	 */
 	@Bean
 	@ServiceActivator(inputChannel = "copy-channel")
 	public MessageHandler outboundChannelAdapter(Storage gcs) {
-		GcsMessageHandler outboundChannelAdapter =
-				new GcsMessageHandler(new GcsSessionFactory(gcs));
+		GcsMessageHandler outboundChannelAdapter = new GcsMessageHandler(new GcsSessionFactory(gcs));
```
artembilan: For the reusing remote file name concern we need to add this:
```
outboundChannelAdapter.setFileNameGenerator(message -> 
		message.getHeaders().get(FileHeaders.REMOTE_FILE, String.class));
```

artembilan:
Also fix, please, Checkstyle violations:
```
/home/travis/build/spring-cloud/spring-cloud-gcp/spring-cloud-gcp-samples/spring-cloud-gcp-integration-storage-sample/src/main/java/com/example/GcsSpringIntegrationApplication.java:32:8: error: Unused import - org.springframework.cloud.gcp.storage.integration.filters.GcsPersistentAcceptOnceFileListFilter.
/home/travis/build/spring-cloud/spring-cloud-gcp/spring-cloud-gcp-samples/spring-cloud-gcp-integration-storage-sample/src/main/java/com/example/GcsSpringIntegrationApplication.java:43:8: error: Unused import - org.springframework.integration.metadata.SimpleMetadataStore.
```

See Travis report.

spring-cloud-gcp-samples/spring-cloud-gcp-integration-storage-sample/src/main/java/com/example/GcsSpringIntegrationApplication.java
```diff
@@ -129,6 +128,9 @@ public MessageHandler outboundChannelAdapter(Storage gcs) {
 		GcsMessageHandler outboundChannelAdapter = new GcsMessageHandler(new GcsSessionFactory(gcs));
 		outboundChannelAdapter.setRemoteDirectoryExpression(
 				new ValueExpression<>(this.gcsWriteBucket));
+		outboundChannelAdapter
+				.setFileNameGenerator(message -> message.getHeaders().get(FileHeaders.REMOTE_FILE, String.class) + "."
+						+ System.currentTimeMillis());
```
artembilan: Is it really what we want to add in the end of file name as an extension?

meltsufin: Seemed like a way to avoid filename collision every time the sample is run. I can remove it.

artembilan: Well, you may just document that default file writing mode is `FileExistsMode.REPLACE`.
So, a new file with the same name will override existing one in the `gcsWriteBucket`.
Or you can configure it to the `FAIL` to let end-user to know that file is already there and has to be removed before the next sample cycle.

PR #779, opened 2018-06-25
Add conditional on web app logging autoconfig

### LINKED ISSUE #766: Add Webflux WebFilter for Logging trace ID extraction
Currently the `StackdriverLoggingAutoConfiguration` class is dependent on a web-mvc application by importing a `WebMvcConfigurer` implementation to add the `TraceIdLoggingWebMvcInterceptor` as an mvc-style interceptor.

It would be ideal if this auto-configuration was also supported in webflux environments.

ChengyuanZhao:
please see #766 for context

### DIFF
--- a/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/logging/StackdriverLoggingAutoConfiguration.java
+++ b/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/logging/StackdriverLoggingAutoConfiguration.java
@@ -22,6 +22,7 @@ import org.springframework.boot.autoconfigure.AutoConfigureAfter;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnWebApplication;
 import org.springframework.cloud.gcp.autoconfigure.logging.extractors.TraceIdExtractor;
 import org.springframework.cloud.gcp.autoconfigure.logging.extractors.XCloudTraceIdExtractor;
 import org.springframework.cloud.gcp.autoconfigure.trace.StackdriverTraceAutoConfiguration;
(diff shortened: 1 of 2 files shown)


spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/logging/StackdriverLoggingAutoConfiguration.java
```diff
@@ -42,6 +43,7 @@
 @ConditionalOnClass({HandlerInterceptorAdapter.class, LoggingAppender.class})
 @ConditionalOnMissingBean(type = "org.springframework.cloud.sleuth.autoconfig.SleuthProperties")
 @AutoConfigureAfter(StackdriverTraceAutoConfiguration.class)
+@ConditionalOnWebApplication
```
artembilan: The problem that this might not be just enough. 
See its JavaDocs:
```
 * {@link Conditional} that matches when the application is a web application. By default,
 * any web application will match but it can be narrowed using the {@link #type()}
 * attribute.
```

I don't think that `TraceIdLoggingWebMvcInterceptor` is what we expect in the non-`SERVLET` environment...

meltsufin:
@ChengyuanZhao This PR actually does not intend to fix the main point of #766. We need a different issue for what you're doing here.

artembilan:
Why do we need a separate issue?
We can just keep PR as an issue reference and don't close that one after merging here.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/logging/StackdriverLoggingAutoConfiguration.java
```diff
@@ -42,6 +44,7 @@
 @ConditionalOnClass({HandlerInterceptorAdapter.class, LoggingAppender.class})
 @ConditionalOnMissingBean(type = "org.springframework.cloud.sleuth.autoconfig.SleuthProperties")
 @AutoConfigureAfter(StackdriverTraceAutoConfiguration.class)
+@ConditionalOnWebApplication(type = Type.SERVLET)
```
artembilan: Any chances to add a test-case with the non-`SERVLET` environment to be sure that this configuration is not applied and there is no any collisions: https://docs.spring.io/spring-boot/docs/2.0.3.RELEASE/reference/htmlsingle/#boot-features-testing-spring-boot-applications-detecting-web-app-type ?

PR #793, opened 2018-06-25
Patch spanner request ID parsing

### LINKED ISSUE #792: Spanner Spring Data autoconfiguration must handle null ID in POST request
When using `@RepositoryRestResource`, it is not currently possible to perform a POST request against the collection resource since it expects a non-null ID here:
https://github.com/spring-cloud/spring-cloud-gcp/blob/master/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/spanner/SpannerKeyIdConverter.java#L41

This should return null when the `id` is null.
For collection `guestbookMessages`:
``` 
curl -XPOST -H "content-type: application/json" -d '{"name": "John Doe", "message": "Hello Cloud Spanner"}'  
```
http://localhost:8080/guestbookMessages currently throws a `NullPointerException`

jabubake:
Fixes #792 792


jabubake:
@jsimonweb FYI

PR #796, opened 2018-06-26
fix timing issue for nack-ed messages

dmitry-s:


artembilan:
@meltsufin ,

??? What made you think that this is a good fix so you decided to merge it?

That's too long to sleep and there definitely no guarantee that the mentioned ack will happen during this period.
Well, to be honest, this sleep never gives the proper delay. Only what it does for us is a suspending of the current thread and lets other resources to obtain CPU. But from here no one will give you an answer that CPU is really going to do something with that resource you would like to wait for.

Saying that I mean that we need to find out some hook on the Pub/Sub to check the real state of the message, but don't sleep almost for nothing in unaware situation.

Hope I am clear.

And thanks for understanding

meltsufin:
@artembilan The issue is that minimum ack deadline on Pub/Sub is 10 seconds. That's a limitation of the service. I don't know if we have a lower-level API to find out the real status of the message. It's worth exploring though. This change does not materially change the waiting because the deleted code was waiting for 10 seconds as well.

artembilan:
Well, what's wrong with `polling until` approach?
My point that even this 10 seconds sleep doesn't give you guarantee that message is going to be available for you there.
The fact that you have moved the sleep from one place to another doesn't advocate that the solution was right previously.

In my experience all the sleep-based tests are weak and don't give the real picture of the situation around resources.

The async solutions are really hard and we need to think twice and definitely together how and what to test. And  investigate what tools and approaches we can use to reach better and stable test coverage.

For example in Spring Integration we have the tool like:
```
List myList = ...;
Assert.assertThat(0, eventually(100, 100, equalsResult(myList::size)));
```
So, we poll the size of the `list` at least `100` times until it is `0` with the `100 ms`  delay between calls.
Right, we still sleep, but we do that with short periods and there is more chances that we won't sleep for nothing too long.

Just an idea how to be. Anyway each use-case requires its own investigation and solution.

As far as for this "minimum ack deadline on Pub/Sub is 10 seconds" issue. Doesn't look reasonable for me to block each test for so long time.
There should be at least some reflection option to shorten this period. And so on. 

Like we say in our country: 
> Seven times measure - cut once. 

😉

dmitry-s:
@artembilan what we actually trying to do here is test that ack worked. To do that we need to make sure that after ackDeadline the message is not back in the queue. If we don't wait long enough - we can't be sure.

PR #800, opened 2018-06-26
Spanner client lib API changes

ChengyuanZhao:
Moved to newer version of Spanner client lib for new functionality of structs. 

It involved some API changes for us, and also support for reading Struct columns as Structs. Added new unit tests to cover this case. 

Note that these changes were necessary when going to 0.53.0-alpha BOM for the GCP client libs, which we have already done for the 1.0.x branch.

### DIFF
--- a/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/StructAccessor.java
+++ b/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/StructAccessor.java
@@ -70,6 +70,9 @@ class StructAccessor {
 					.put(double[].class, AbstractStructReader::getDoubleArray)
 					.put(long[].class, AbstractStructReader::getLongArray)
 					.put(boolean[].class, AbstractStructReader::getBooleanArray)
+					// Note that Struct.class appears in this map. While we support
+					// converting structs into POJO fields of POJOs, the value in this map is for
+					// the case where the field within the POJO is Struct.
 					.put(Struct.class, Struct::getStruct).build();
 
 	private final SpannerTypeMapper spannerTypeMapper;
--- a/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/StructAccessor.java
+++ b/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/StructAccessor.java
@@ -69,7 +69,8 @@ class StructAccessor {
 					.put(ByteArray.class, AbstractStructReader::getBytes)
 					.put(double[].class, AbstractStructReader::getDoubleArray)
 					.put(long[].class, AbstractStructReader::getLongArray)
-					.put(boolean[].class, AbstractStructReader::getBooleanArray).build();
+					.put(boolean[].class, AbstractStructReader::getBooleanArray)
+					.put(Struct.class, Struct::getStruct).build();
 
 	private final SpannerTypeMapper spannerTypeMapper;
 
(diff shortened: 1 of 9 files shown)


ChengyuanZhao:
@meltsufin PTAL

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/StructAccessor.java
```diff
@@ -69,7 +69,11 @@
 					.put(ByteArray.class, AbstractStructReader::getBytes)
 					.put(double[].class, AbstractStructReader::getDoubleArray)
 					.put(long[].class, AbstractStructReader::getLongArray)
-					.put(boolean[].class, AbstractStructReader::getBooleanArray).build();
+					.put(boolean[].class, AbstractStructReader::getBooleanArray)
+					// Note that Struct.class appears in this map. While we support
+					// converting structs into POJO fields of POJOs, the value in this map is for
+					// the case where the field within the POJO is Struct.
```
meltsufin: Do we also support the case where the type of the field in the entity is a `Value`?

spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/core/convert/SpannerWriteMethodCoverageTest.java
```diff
@@ -45,7 +47,18 @@ public void allKnownMappingTypesTest() throws NoSuchFieldException {
 			}
 
 			Class<?> paramType = ConversionUtils.boxIfNeeded(method.getParameterTypes()[0]);
-			if (ConversionUtils.isIterableNonByteArrayType(paramType)) {
+			if (paramType.equals(Struct.class) || paramType.equals(Value.class)) {
+				/*
+				 * 1. there is a method for binding a Struct value, but because Struct
+				 * values cannot be written to table columns we will ignore it. 2. there
```
meltsufin: What about using structs in the `WHERE` clause? Does that work? Do we support it?

spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/core/convert/ConverterAwareMappingSpannerEntityReaderTest.java
```diff
@@ -83,20 +85,50 @@ public void readNestedStructTest() {
 		assertEquals("value", result.innerTestEntities.get(0).value);
 	}
 
+	@Test
+	public void readNestedStructsAsStructsTest() {
+		Struct innerStruct = Struct.newBuilder().set("value").to(Value.string("value"))
```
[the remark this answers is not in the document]
ChengyuanZhao: fixed

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/StructAccessor.java
ChengyuanZhao: We don't support `Value` here because all columns from the result set come wrapped in `Value`, and what we do with all the convert classes is to get the value from within. 

I suppose somebody could want us to do absolutely no converting, though this would short-circuit a lot of our code, and could be a somewhat large job. I can make an issue of it so we don't lose track of that.

spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/core/convert/SpannerWriteMethodCoverageTest.java
ChengyuanZhao: This class is specifically only testing the methods for writing to columns during creating a mutation to send to spanner, and as a result we don't support structs because columns in spanner cannot store structs. 

Separately, we also can't do equality comparisons on entire structs.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/StructAccessor.java
ChengyuanZhao: #822

meltsufin: Thanks!

ChengyuanZhao: Struct does not have a way to read out just `Value`, so the conversion to the underlying type always needs to be done. Please see the linked issue for more details.

PR #811, opened 2018-06-28
Spanner LIKE and CONTAINS queries

### LINKED ISSUE #809: Spring Data Spanner - "findXByYLike" doesn't seem to work
```
@RepositoryRestResource
public interface EventRestRepository extends SpannerRepository<Event, Long> {
	public List<Event> findEventsByNameLike(String name);
}
```

When I call this via REST API, e.g.,
```
curl http://localhost:8080/events/search/findEventsByNameLike?name=Java
{"timestamp":"2018-06-27T21:00:33.142+0000","status":500,"error":"Internal Server Error","message":"INVALID_ARGUMENT: Syntax error: Illegal input character \"%\" [at 1:39]\nSELECT * FROM event WHERE ( name LIKE %@tag0 );\n                                      ^","path":"/events/search/findEventsByNameLike"}
```

### LINKED ISSUE #810: Spring Data Spanner - implement Contains Query
```
@RepositoryRestResource
public interface EventRestRepository extends SpannerRepository<Event, Long> {
	public List<Event> findEventsByNameContains(String name);
}
```
This doesn't get translated into the query that can be simulated w/ `LIKE %name%`

ChengyuanZhao:
fixes #809
fixes #810 

For LIKE and NOT LIKE queries the user needs to provide the `_` or `%` matching characters in the param.
https://cloud.google.com/spanner/docs/functions-and-operators#operators

For CONTAINS and NOT  CONTAINS queries, the string the user provides is a regex. 
https://cloud.google.com/spanner/docs/functions-and-operators#regexp_contains

### DIFF
--- a/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryExecutor.java
+++ b/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryExecutor.java
@@ -188,7 +188,18 @@ public class SpannerStatementQueryExecutor {
 
 					switch (part.getType()) {
 					case LIKE:
-						andString += " LIKE %" + insertedTag;
+						andString += " LIKE " + insertedTag;
+						break;
+					case NOT_LIKE:
+						andString += " NOT LIKE " + insertedTag;
+						break;
+					case CONTAINING:
+						andString = " REGEXP_CONTAINS(" + andString + "," + insertedTag
+								+ ") =TRUE";
+						break;
+					case NOT_CONTAINING:
+						andString = " REGEXP_CONTAINS(" + andString + "," + insertedTag
+								+ ") =FALSE";
 						break;
 					case SIMPLE_PROPERTY:
 						andString += "=" + insertedTag;
--- a/spring-cloud-gcp-docs/src/main/asciidoc/spanner.adoc
+++ b/spring-cloud-gcp-docs/src/main/asciidoc/spanner.adoc
@@ -869,6 +869,20 @@ a non-case-sensitive matching. The `IgnoreCase` phrase may only be appended to f
 to columns of type STRING or BYTES.
 The Spring Data "AllIgnoreCase" phrase appended at the end of the method name is not supported.
 
+The `Like` or `NotLike` naming conventions:
+[source, java]
+----
+List<Trade> findBySymbolLike(String symbolFragment);
+----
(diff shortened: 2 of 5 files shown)


spring-cloud-gcp-docs/src/main/asciidoc/spanner.adoc
```diff
@@ -869,6 +869,20 @@ a non-case-sensitive matching. The `IgnoreCase` phrase may only be appended to f
 to columns of type STRING or BYTES.
 The Spring Data "AllIgnoreCase" phrase appended at the end of the method name is not supported.
 
+The `Like` or `NotLike` naming conventions:
```
meltsufin: Can we document all of the keywords that we support like the table here? https://docs.spring.io/spring-data/data-jpa/docs/1.3.4.RELEASE/reference/html/repository-query-keywords.html

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryExecutor.java
```diff
@@ -188,7 +188,18 @@ else if (part.shouldIgnoreCase() != IgnoreCaseType.NEVER) {
...
 					switch (part.getType()) {
 					case LIKE:
-						andString += " LIKE %" + insertedTag;
+						andString += " LIKE " + insertedTag;
+						break;
+					case NOT_LIKE:
+						andString += " NOT LIKE " + insertedTag;
+						break;
+					case CONTAINING:
+						andString = " REGEXP_CONTAINS(" + andString + "," + insertedTag
```
meltsufin: I'm not sure what `CONTAINING` should mean, but Ray said `LIKE %search%`, not REGEX.

spring-cloud-gcp-docs/src/main/asciidoc/spanner.adoc
ChengyuanZhao: Yes please make an issue of it

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryExecutor.java
ChengyuanZhao: Turns out we cannot just append and prepend the % symbols. Those symbols must be part of the actual param. We have not laid out this code to allow actual changing of the params, which Spring Data provides to us as  Objects. Therefore, unless we want to make things hacky and ugly, and unless there is a straight up CONTAINS(string) function from Cloud Spanner, this REGEXP_CONTAINS is the best choice.

PR #813, opened 2018-06-28
Re-ordered method parameters

dhoard:
Re-ordered method parameters to be more consistent with other methods within the class.

While this is not a functional change, it makes it more clean, clear, and consistent.

### DIFF
--- a/spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-sample/src/main/java/com/example/WebController.java
+++ b/spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-sample/src/main/java/com/example/WebController.java
@@ -57,8 +57,8 @@ public class WebController {
 	}
 
 	@GetMapping("/postMessage")
-	public RedirectView publish(@RequestParam("message") String message,
-			@RequestParam("topicName") String topicName) {
+	public RedirectView publish(@RequestParam("topicName") String topicName,
+			@RequestParam("message") String message) {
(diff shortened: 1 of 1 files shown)


meltsufin:
@dhoard Thank you for the contribution!
@artembilan I'm thinking that for now, we should make all changes that apply to both branches in `master`, and at some point before RC2 or GA we can do one PR to sync relevant deltas up to `1.0.x`.

meltsufin:
@dhoard Would mind opening a similar PR against `master`? We can merge this one too.

artembilan:
That makes sense, @meltsufin .

Let's open issue on the matter and gather there commits from `master` which should go into the `1.0.x` as back-porting task!

This way we indeed just will need to ask contributors to PR against `master`.

Thank you! Both

Let's push merge button then already!

meltsufin:
Sounds good! https://github.com/spring-cloud/spring-cloud-gcp/issues/814

dhoard:
@meltsufin PR against `master` created

meltsufin:
Thanks! I guess we can just close this one and deal with the merge into `1.0.x` later along with other things.

artembilan:
Does it matter, @meltsufin ?

We have this PR here already and it is really ready to be merged.
There is just need to push `Merge` button though...

I don't see reason to move this work to some other place... 
🤷‍♂️

meltsufin:
I don't think it would add much additional work either way, but sure, I'll merge! :)

Issue #816, opened 2018-06-28
PubSubTemplate can't be used for a publish only use case without a SubscriberFactory


dhoard:
My team and I have various use cases, most of which are subscribe only ... but we have few publish only use cases.

The `PubSubTemplate` class has methods for both publish and subscribe use cases.

From a subscribe only use case, from looking at the code, I know I can just pass `null` for the `PublisherFactory` and use it only for the subscribe use case.

From a publish only use case, I can't use `PubSubTemplate` without a creating a `SubscriberFactory` because of the code in the constructor.

Quick fix ...

1) make the constructor of `PubSubTemplate` check for a null `SubscriberFactory`.  In that case, assume that `PubSubTemplate` is only going to be used for a publish use case and don't make the methods calls in the constructor that use the `SubscriberFactory`

Clean fix ...

1) separate the interface `PubSubOperations` into `PubSubPublisherOperations` (publish methods) and `PubSubSubscriberOperations` (subscription / pull methods)

2) create a `PubSubPublisherTemplate` that implements `PubSubPublisherOperations`

3) create a `PubSubSubscriberTemplate` that implements `PubSubSubscriberOperations`

4) change `PubSubTemplate` to implement `PubSubPublisherOperations` and `PubSubSubscriberOperations`

5) change `PubSubTemplate` to contain instances of `PubSubPublisherTemplate` and `PubSubSubscriberTemplate`.  Have the methods in `PubSubTemplate` delegate to the appropriate methods in the contained instances

meltsufin:
I like the idea of separating the concerns of publishing and subscribing - the clean fix. I'll take a closer look at the PR a bit later. Thanks for the thorough issue description and the PR!

dhoard:
@meltsufin I create PR #819 to implement a solution that should be considered clean.

meltsufin:
@dhoard Thanks! I did the same. So, we'll need to resolve differences. :)
I'm happy to accept your PR once any differences are resolved.

PR #817, opened 2018-06-29
Fix to allow a publish only use case without a SubscriberFactory

dhoard:
Quick fix suggested in issue #816

Added additional comments in the default constructor which clarifies
that only a PublisherFactory is required when wanting to use publish
methods and only a SubscriberFactory is required when wanting to use
subscribe / pull methods.

Changed constructor code to handle a null SubscriberFactory - common
for a subscriber / pull only use case.

Added constructor that only requires a PublisherFactory - common
for a publish only use case.

Added constructor that only requires a SubscriberFactory - common
for a subscriber / pull only use case.

### DIFF
--- a/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
+++ b/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
@@ -73,7 +73,7 @@ public class PubSubTemplate implements PubSubOperations, InitializingBean {
 	 * Default {@link PubSubTemplate} constructor that uses {@link SimplePubSubMessageConverter}
 	 * to serialize and deserialize payloads.
 	 * @param publisherFactory the {@link com.google.cloud.pubsub.v1.Publisher} factory to
-	 * publish to topic. Only required if using publish methods
+	 * publish to topics. Only required if using publish methods
 	 * @param subscriberFactory the {@link com.google.cloud.pubsub.v1.Subscriber} factory
 	 * to subscribe to subscriptions. Only required if using subscriber or pull methods
 	 */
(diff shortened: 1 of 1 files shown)


spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
```diff
@@ -72,16 +73,43 @@
...
+			this.subscriberStub = null;
+			this.acknowledger = null;
+		}
+	}
+
+	/**
+	 * {@link PubSubTemplate} constructor that uses {@link SimplePubSubMessageConverter}
+	 * to serialize and deserialize payloads.
+	 * @param publisherFactory the {@link com.google.cloud.pubsub.v1.Publisher} factory to
+	 * publish to topics. Subscribe and pull methods are not supported
```
artembilan: There is a rule for code style that new public API must have it version reflected via `@since`: https://github.com/spring-projects/spring-framework/wiki/Code-Style#use-of-since

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
```diff
@@ -72,16 +73,43 @@
...
+	}
+
+	/**
+	 * {@link PubSubTemplate} constructor that uses {@link SimplePubSubMessageConverter}
+	 * to serialize and deserialize payloads.
+	 * @param subscriberFactory the {@link com.google.cloud.pubsub.v1.Subscriber} factory
+	 * to subscribe to subscriptions. Publish methods are not supported
+	 */
+	public PubSubTemplate(SubscriberFactory subscriberFactory) {
+		this(null, subscriberFactory);
```
artembilan: With this two `null`-based constructor you have missed the fact to `Assert.notNull()` factories in the target method.
Otherwise, when I don provide a `PublisherFactory`, my first `PubSubTemplate.publsih()` will fail in the `public ListenableFuture<String> publish(final String topic, PubsubMessage pubsubMessage) {`, because of NPE.
That's not good and so informative: we need to be proactive and fail with the reasonable message using particular Spring utils.

I don't tell that you need new unit-tests, but with them you would get an issue early and proceed respectively.

Thanks for understanding!

P.S. We may come back to our rule "always unit tests in PRs!"

dhoard:
@artembilan given comments by @meltsufin in https://github.com/spring-cloud/spring-cloud-gcp/issues/816 I am going to close this PR.

Splitting the publish / subscription operations is a much cleaner implementation.

PR #819, opened 2018-06-29
Added specific publish and subscriber interfaces and classes

dhoard:
This pull request implements code changes to resolve https://github.com/spring-cloud/spring-cloud-gcp/issues/816 by refactoring the class `PubSubTemplate` into one that delegates to an encapsulated instance of either a `PubSubPublishTemplate` or a `PubSubSubscribeTemplate`.

Additionally, I have introduced `PubSubPublishOperations` and `PubSubSubscribeOperations` interfaces.

For a publish - subscriber / pull use case, the class `PubSubTemplate` should be used.

For a publish only use case, the class `PubSubPublishTemplate` should be used.

For a subscribe / pull only use case, the class `PubSubSubscribeTemplate` should be used.

PR #820, opened 2018-06-29
Split PubSubTemplate into Publisher and Subscriber [WIP]

### LINKED ISSUE #816: PubSubTemplate can't be used for a publish only use case without a SubscriberFactory

meltsufin:
Changes PubSubTemplate to contain instances of PubSubPublisherTemplate
and PubSubSubscriberTemplate and delegates all method calls to them.

Fixes #816.

### DIFF
--- a/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
+++ b/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
@@ -38,6 +38,8 @@ import com.google.cloud.pubsub.v1.SubscriptionAdminClient;
 import com.google.cloud.pubsub.v1.SubscriptionAdminSettings;
 import com.google.cloud.pubsub.v1.TopicAdminClient;
 import com.google.cloud.pubsub.v1.TopicAdminSettings;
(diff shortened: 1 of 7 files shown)


meltsufin:
I still need to fix the tests, but otherwise it should be ready for review.

meltsufin:
Closing in favor of #819.

PR #825, opened 2018-07-02
Use bifurcated Pub/Sub template

### LINKED ISSUE #823: Use PubSubPublisherTemplate and PubSubSubscriberTemplate
As a follow-up to #819:

- [ ] Update `GcpPubSubAutoConfiguration` to create and use `PubSubPublisherTemplate` and `PubSubSubscriberTemplate`.
- [ ] The Pub/Sub outbound channel adapter should use `PubSubPublisherTemplate` rather than `PubSubTemplate`.
- [ ] The Pub/Sub inbound channel adapter should use `PubSubSubscriberTemplate` rather than `PubSubTemplate`.

meltsufin:
Updates GcpPubSubAutoConfiguration to create and use PubSubPublisherTemplate and PubSubSubscriberTemplate.

The Pub/Sub outbound channel adapter now uses PubSubPublisherTemplate rather than PubSubTemplate.

The Pub/Sub inbound channel adapter now uses PubSubSubscriberTemplate rather than PubSubTemplate.

Fixes #823.

### DIFF
--- a/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapter.java
+++ b/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapter.java
@@ -80,6 +80,7 @@ public class PubSubInboundChannelAdapter extends MessageProducerSupport {
 
 		try {
 			sendMessage(MessageBuilder.withPayload(
+					// TODO: need to get the message converter from somewhere.
 					this.pubSubSubscriberOperations.getMessageConverter().fromPubSubMessage(pubsubMessage, this.payloadType))
 					.copyHeaders(messageHeaders)
 					.build());
--- a/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapter.java
+++ b/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapter.java
@@ -24,6 +24,7 @@ import com.google.pubsub.v1.PubsubMessage;
 
 import org.springframework.cloud.gcp.pubsub.core.PubSubException;
 import org.springframework.cloud.gcp.pubsub.core.PubSubTemplate;
+import org.springframework.cloud.gcp.pubsub.core.subscriber.PubSubSubscriberOperations;
 import org.springframework.cloud.gcp.pubsub.integration.AckMode;
 import org.springframework.cloud.gcp.pubsub.integration.PubSubHeaderMapper;
 import org.springframework.cloud.gcp.pubsub.support.GcpPubSubHeaders;
@@ -43,7 +44,7 @@ public class PubSubInboundChannelAdapter extends MessageProducerSupport {
 
 	private final String subscriptionName;
 
-	private final PubSubTemplate pubSubTemplate;
+	private final PubSubSubscriberOperations pubSubSubscriberOperations;
 
 	private Subscriber subscriber;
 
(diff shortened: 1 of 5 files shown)


spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapter.java
```diff
@@ -77,7 +80,8 @@ private void receiveMessage(PubsubMessage pubsubMessage, AckReplyConsumer consum
 
 		try {
 			sendMessage(MessageBuilder.withPayload(
-					this.pubSubTemplate.getMessageConverter().fromPubSubMessage(pubsubMessage, this.payloadType))
+					// TODO: need to get the message converter from somewhere.
```
meltsufin: @artembilan Since the subscriber template now doesn't have the message converter, we should either add it there, even though it will not be used, or have a public method that can do the conversion there, or inject the converter into this class. Any preference?

artembilan: OK. Since there is no conversion anymore, then we should move it here.
Otherwise it does not make sense until we really don't implement the conversion there.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapter.java
```diff
@@ -141,4 +143,14 @@ public void setHeaderMapper(HeaderMapper<Map<String, String>> headerMapper) {
 		Assert.notNull(headerMapper, "The header mapper can't be null.");
 		this.headerMapper = headerMapper;
 	}
+
+	public PubSubMessageConverter getMessageConverter() {
+		return this.pubSubMessageConverter;
+	}
+
+	public void setMessageConverter(
```
artembilan: New API?
How about Javadocs and `@since 1.1` ?

meltsufin:
@artembilan Can you take another look pls? Thanks!

meltsufin:
@artembilan friendly ping :)

artembilan:
This PR in some shape is clashing with the #841 .
So, we need to decide first of all which direction we go, then review one or another.
Otherwise it is going to be a wasting time.
No? What am I missing?

meltsufin:
I don't think it's really clashing. Let's merge this in because it's less controversial. #841 is orthogonal to this.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/outbound/PubSubMessageHandler.java
```diff
@@ -48,7 +48,7 @@
 
 	private static final long DEFAULT_PUBLISH_TIMEOUT = 10000;
 
-	private final PubSubOperations pubSubTemplate;
+	private final PubSubOperations pubSubPublisherOperations;
```
artembilan: `PubSubPublisherOperations` then ?

meltsufin: good catch

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/outbound/PubSubMessageHandler.java
```diff
@@ -62,10 +63,10 @@
 
 	private HeaderMapper<Map<String, String>> headerMapper = new PubSubHeaderMapper();
 
-	public PubSubMessageHandler(PubSubOperations pubSubTemplate, String topic) {
-		Assert.notNull(pubSubTemplate, "Pub/Sub template cannot be null.");
+	public PubSubMessageHandler(PubSubOperations pubSubPublisherOperations, String topic) {
```
artembilan: Yeah... You didn't get me.
This one must be a `PubSubPublisherOperations` as well 😉

meltsufin: missed it

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapter.java
```diff
@@ -137,4 +143,24 @@ public void setHeaderMapper(HeaderMapper<Map<String, String>> headerMapper) {
...
+	public PubSubMessageConverter getMessageConverter() {
+		return this.pubSubMessageConverter;
+	}
+
+	/**
+	 * Sets the {@link PubSubMessageConverter} used to convert the {@code byte[]} Pub/Sub
+	 * payload to an object of the configured payload type.
+	 * @since 1.1
+	 */
+	public void setMessageConverter(
```
artembilan: Please, move all the setters and getter just after ctors: `protected` and `private` method should go after: https://github.com/spring-projects/spring-framework/wiki/Code-Style#java-source-file-organization

The same requirement applies for the `PubSubMessageHandler` as well

PR #829, opened 2018-07-03
Fixed constructor null parameter error messages

dhoard:
Fixed error messages when null parameters are passed to the constructors of `PubSubTemplate`, `PubSubPublisherTemplate`, and `PubSubSubscriberTemplate`

### DIFF
--- a/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/publisher/PubSubPublisherTemplate.java
+++ b/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/publisher/PubSubPublisherTemplate.java
@@ -48,7 +48,7 @@ public class PubSubPublisherTemplate implements PubSubPublisherOperations {
 
 	private static final Log LOGGER = LogFactory.getLog(PubSubPublisherTemplate.class);
 
-	private PubSubMessageConverter messageConverter = new SimplePubSubMessageConverter();
+	private PubSubMessageConverter pubSubMessageConverter = new SimplePubSubMessageConverter();
 
 	private final PublisherFactory publisherFactory;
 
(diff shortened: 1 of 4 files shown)


spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/publisher/PubSubPublisherTemplate.java
```diff
@@ -48,7 +48,7 @@
 
 	private static final Log LOGGER = LogFactory.getLog(PubSubPublisherTemplate.class);
 
-	private PubSubMessageConverter messageConverter = new SimplePubSubMessageConverter();
+	private PubSubMessageConverter pubSubMessageConverter = new SimplePubSubMessageConverter();
```
meltsufin: My concern with this private member rename is that it now creates an inconsistency with the public accessor `getMessageConverter()`. It also seems auxiliary to the purpose of this PR. I would skip this rename.

artembilan: I would say that I'm fine with the renaming - this is just a private property, although this is really unfortunate that setter and getter can't be renamed...

dhoard: Since the core interface for all of the Pub / Sub converters is `PubSubMessageConverter` ...

I propose ...

1) keep the private member name change
2) adding new methods `getPubSubMessageConverter` and `setPubSubMessageConverter`
3) deprecated `getMessageConverter` and `setMessageConverter`

artembilan: OK. But I'm fine to stay with the current situation even if property doesn't reflect the class it is based on...

meltsufin: I would avoid renaming the getter and setter for now. Let's just merge as is, since I don't mind the private member rename too much.

PR #830, opened 2018-07-03
Make Pub/Sub integration test more robust

meltsufin:


spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/pubsub/it/PubSubTemplateIntegrationTests.java
```diff
@@ -157,9 +157,17 @@ public void testPullAndAck() {
 			});
 
 			Thread.sleep(11_000);
-			ackableMessages = pubSubTemplate.pull(subscriptionName, 4, true);
-			assertThat(ackableMessages.size()).as("check that we get both nacked messages back").isEqualTo(2);
-			ackableMessages.forEach(AcknowledgeablePubsubMessage::ack);
+			// pull the 2 nacked messages with retries for up to 10s
+			int messagesCount = 0;
+			int tries = 100;
+			while (messagesCount < 2 && tries > 0) {
```
artembilan: Do we still need to sleep those 11 seconds above with solution?

ChengyuanZhao: I think you should just keep track of tries or turn the messageCount threshold here to something much larger than 2. 

The reason is you might stop doing this loop even if there are extra unexpected messages left, and then this test would pass but there are unexpected messages that you didn't check.

meltsufin: We still do because we're trying to verify that the ack'ed message actually did get ack'ed and will not come back after 10 seconds. If it wasn't properly ack'ed, we will get 3 messages instead of 2.

meltsufin: We are already waiting >10s for the unexpected messages. This is attempting to fix the flakiness where we don't get enough messages.

PR #833, opened 2018-07-04
Added max ackdeadline check

dhoard:
Added max ackDeadline check to `createSubscription`

### DIFF
--- a/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/PubSubAdmin.java
+++ b/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/PubSubAdmin.java
@@ -44,6 +44,10 @@ import org.springframework.util.Assert;
  */
 public class PubSubAdmin {
 
+	private static final int MINIMUM_ACK_DEADLINE_SECONDS = 10;
+
+	private static final int MAXIMUM_ACK_DEADLINE_SECONDS = 600;
+
 	private final String projectId;
 
 	private final TopicAdminClient topicAdminClient;
@@ -196,8 +200,10 @@ public class PubSubAdmin {
 
 		int finalAckDeadline = this.defaultAckDeadline;
 		if (ackDeadline != null) {
-			Assert.isTrue(ackDeadline >= 10 && ackDeadline <= 600,
-					"The acknowledgement deadline must be between 10 and 600 seconds.");
+			Assert.isTrue(ackDeadline >= MINIMUM_ACK_DEADLINE_SECONDS
+							&& ackDeadline <= MAXIMUM_ACK_DEADLINE_SECONDS,
+					String.format("The acknowledgement deadline must be between %d and %d seconds.",
+							MINIMUM_ACK_DEADLINE_SECONDS, MAXIMUM_ACK_DEADLINE_SECONDS));
 
 			finalAckDeadline = ackDeadline;
 		}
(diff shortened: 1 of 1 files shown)


spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/PubSubAdmin.java
```diff
@@ -196,8 +196,9 @@ public Subscription createSubscription(String subscriptionName, String topicName
 
 		int finalAckDeadline = this.defaultAckDeadline;
 		if (ackDeadline != null) {
-			Assert.isTrue(ackDeadline >= 0,
-					"The acknowledgement deadline value can't be negative.");
+			Assert.isTrue(ackDeadline >= 10 && ackDeadline <= 600,
```
ChengyuanZhao: Should these two special numbers be made into constants ?

dhoard: I'll refactor them into constants

meltsufin: Something like [this](https://github.com/GoogleCloudPlatform/google-cloud-java/blob/27c90f1298bb27a618ff59b3cd17321c577f3a6b/google-cloud-clients/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/Subscriber.java#L97).
Unfortunately it's package private there.

dhoard:
Refactored min and max ack deadlines into constants.  Should we refactor ...

`private int defaultAckDeadline = 10;`

... into a constant, `DEFAULT_ACK_DEADLINE_SECONDS = 10;` and use the constant?

meltsufin:
`private int defaultAckDeadline = MIN_ACK_DEADLINE_SECONDS;`
should be good

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/PubSubAdmin.java
```diff
@@ -196,8 +200,11 @@ public Subscription createSubscription(String subscriptionName, String topicName
 
 		int finalAckDeadline = this.defaultAckDeadline;
 		if (ackDeadline != null) {
-			Assert.isTrue(ackDeadline >= 0,
-					"The acknowledgement deadline value can't be negative.");
+			Assert.isTrue(ackDeadline >= MIN_ACK_DEADLINE_SECONDS
+							&& ackDeadline <= MAX_ACK_DEADLINE_SECONDS,
+					String.format("The acknowledgement deadline must be between %d and %d seconds.",
```
meltsufin: I'm not sure if `String.format` is a great idea here since the compiler can't optimize that.
The regular concatenation with `+` is probably significantly more efficient.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/PubSubAdmin.java
```diff
@@ -272,8 +279,10 @@ public int getDefaultAckDeadline() {
 	 * @param defaultAckDeadline default acknowledgement deadline value in seconds, must be between 10 and 600 seconds.
 	 */
 	public void setDefaultAckDeadline(int defaultAckDeadline) {
-		Assert.isTrue(defaultAckDeadline >= 10 && defaultAckDeadline <= 600,
-				"The acknowledgement deadline must be between 10 and 600 seconds.");
+		Assert.isTrue(defaultAckDeadline >= MIN_ACK_DEADLINE_SECONDS
+						&& defaultAckDeadline <= MAX_ACK_DEADLINE_SECONDS,
+				String.format("The acknowledgement deadline must be between %d and %d seconds.",
```
meltsufin: Same here.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/PubSubAdmin.java
dhoard: Changed.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/PubSubAdmin.java
dhoard: Changed.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/PubSubAdmin.java
```diff
@@ -196,8 +200,14 @@ public Subscription createSubscription(String subscriptionName, String topicName
 
 		int finalAckDeadline = this.defaultAckDeadline;
 		if (ackDeadline != null) {
-			Assert.isTrue(ackDeadline >= 0,
-					"The acknowledgement deadline value can't be negative.");
+			Assert.isTrue(ackDeadline >= MIN_ACK_DEADLINE_SECONDS
```
dmitry-s: probably makes sense to move that check to a separate method this to avoid duplication

PR #836, opened 2018-07-06
Use Spring Boot autoconfigure-processor

### LINKED ISSUE #834: Use Spring Boot autoconfigure-processor to optimize auto-configurations
As explained in [the its reference documentation](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-docs/src/main/asciidoc/spring-boot-features.adoc#autoconfigure-module), Spring Boot provides a `spring-boot-autoconfigure-processor`.

This should be added as an optional dependency to the module that contains the AutoConfiguration classes (and contains the `spring.factories` file declaring them).

Once configured, this annotation processor will generate metadata at compile time and will help Spring Boot process those auto-configurations without loading the actual classes in some cases.

In this project, the `spring-cloud-gcp-autoconfigure` module declares auto-configurations but doesn't use this annotation processor. Could you add it to the build file of this module?

```xml
<dependency>
	<artifactId>spring-boot-autoconfigure-processor</artifactId>
	<optional>true</optional>
</dependency>
```

Once added, this should generate a `spring-autoconfigure-metadata.properties` file in the `target/classes/META-INF` folder.

meltsufin:
Fixes #834.

### DIFF
--- a/spring-cloud-gcp-autoconfigure/pom.xml
+++ b/spring-cloud-gcp-autoconfigure/pom.xml
@@ -29,6 +29,11 @@
             <artifactId>spring-boot-configuration-processor</artifactId>
             <optional>true</optional>
         </dependency>
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-autoconfigure-processor</artifactId>
+            <optional>true</optional>
+        </dependency>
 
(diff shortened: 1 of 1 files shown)


artembilan:
Has to be back-ported to `1.0.x`

Issue #831, opened 2018-07-03
Create a Pub/Sub Spring Integration sample that uses JSON payloads


meltsufin:

nbrinks:
I think this would be a great idea! I would also love to see a JSON sample without spring-integration as well. I am using 1.1.0.BUILD-SNAPSHOT and following the progress of this library closely. I started out with the Spring Integration sample but the plain `PubSubTemplate` fit my use case better. 

Here is some relevant feedback:

I am using Jackson in my project, which automatically enabled the `JacksonPubSubMessageConverter` thanks to `org.springframework.cloud.gcp.autoconfigure.pubsub.GcpPubSubAutoConfiguration`. This was a pleasant surprise. 👍

I think the most confusing part I ran into was that, AFAIK, I can publish messages using this method `org.springframework.cloud.gcp.pubsub.core.publisher.PubSubPublisherTemplate#publish(java.lang.String, T, java.util.Map<java.lang.String,java.lang.String>)` and the payload will be automatically converted thanks to `JacksonPubSubMessageConverter` but when I subscribe to messages via `org.springframework.cloud.gcp.pubsub.core.PubSubTemplate#subscribe` the received messages (via MessageReceiver interface) are not utilizing any message converter and I have to parse them manually. Perhaps this is not the intended way to subscribe to incoming messages. I think a sample would clear things up.

dhoard:
I would guess in most people's use cases, the messages are in the same format, for example JSON objects. An auto-configured message converter makes sense for these scenarios.  

But, what about the scenario where messages in a topic may not be the same type? JSON vs XML? This use case is not common, not ideal, and quite frankly bad practice for a topic ... but it may occur where legacy systems are involved.

dhoard:
Possible solution ...

1. Add the`PubSubMessageConverter` to the `PubSubSubscriberTemplate`

2. Create a Spring Pub / Sub specific `PubSubMessageReceiver` interface that has the `PubSubMessageConverter` as a parameter ...

```
public interface PubSubMessageReceiver {
	
	void receiveMessage(PubsubMessage pubsubMessage, AckReplyConsumer ackReplyConsumer, PubSubMessageConverter pubSubMessageConverter);
}
```

3. Add a new `subscribe` method into the `PubSubSubscriberOperations` and `PubSubSubscriberTemplate` that accepts the new `PubSubMessageReceiver` ...

```
@Override
[...]

PR #841, opened 2018-07-09
Add message conversion to PubSubSubscriberTemplate

meltsufin:
Exploring ideas related to #831.

@dhoard @artembilan


dhoard:
@meltsufin Thoughts on adding the `PubSubMessage` `messageId` and `publishTime` into the `Map<String, String> headers` using constants?

meltsufin:
@artembilan I'm not sure `ConvertedMessageReceiver` as a sub-interface of `MessageReceiver` would work because it would also need the converter to be specified to implement the default method.

If we were to implement `pullAndConvert`, what would be the return type? `Message<T>`?

@dhoard I'll try adding the timestamp and the id, but my concern is that the timestamp may not be well represented as a `String`.

artembilan:
> also need the converter to be specified to implement the default method.

```
PubSubMessageConverter getConverter();
```
to be called from the `default` method should work on the interface. No?

If you are not going to add `pullAndConvert()` API, then I expressed my position: I don't see reason to overhead a template with the `converter` almost for nothing. And yes: `Message<T>` with appropriate headers sounds like a good compromise. However having a wish to stay away from Spring Messaging there as far as possible, I don't see any other way unless just lose manual acknowledge and return plain `T`. Something similar what we have in Spring AMQP.

dhoard:
I would rather have manual acknowledgement than message conversion.

I can implement conversion code like the sample code I put in #831, but I can’t write any code to unacknowledge a message.

meltsufin:
@dhoard If we go with something like `Message<T> pullAndConvert()`, we can add `AckReplyConsumer` to the headers of the returned `Message`, like we do in the `PubSubInboundChannelAdapter`.
At that point we would make the inbound channel adapter essentially a dumb wrapper around the the template, which might be fine, but I definitely hear @artembilan 's concern about putting too much responsibility into the template.

dhoard:
@artembilan After reviewing the `PubSubInboundChannelAdapter` source / class, I can definitely see your concerns and concur.

Adding the message conversion really contorts the vision and purpose of the `PubSubSubscriberTemplate`.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberTemplate.java
```diff
@@ -142,6 +147,24 @@ public Subscriber subscribe(String subscription, MessageReceiver messageReceiver
 				returnImmediately));
 	}
 
+	@Override
+	public <T> List<Message<T>> pullAndConvert(String subscription, Integer maxMessages, Boolean returnImmediately,
```
meltsufin: @artembilan @dhoard I added this. Let me know how this feels.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/ConvertedMessageReceiver.java
```diff
@@ -0,0 +1,30 @@
...
+
+import com.google.cloud.pubsub.v1.AckReplyConsumer;
+
+/**
+ * @author Mike Eltsufin
+ *
+ * @since 1.1
+ */
+public interface ConvertedMessageReceiver<T> {
+	void receiveMessage(T payload, Map<String, Object> headers, AckReplyConsumer ackReplyConsumer);
```
meltsufin: We can potentially replace this with a plain old `MessageHandler` if the `AckReplyConsumer` is stored in the headers.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberTemplate.java
```diff
@@ -104,6 +147,24 @@ public Subscriber subscribe(String subscription, MessageReceiver messageReceiver
...
+			Class<T> payloadType) {
+		List<AcknowledgeablePubsubMessage> pubsubMessages = this.pull(subscription, maxMessages, returnImmediately);
+
+		return pubsubMessages.stream().map(
+				m -> {
+					Map<String, Object> headers = this.headerMapper.toHeaders(m.getMessage().getAttributesMap());
+					headers.put(GcpPubSubHeaders.PUBLISH_TIME, m.getMessage().getPublishTime());
+					headers.put(GcpPubSubHeaders.MESSAGE_ID, m.getMessage().getMessageId());
+					headers.put(GcpPubSubHeaders.ACKNOWLEDGEMENT, m);
+					return new GenericMessage<>(
```
artembilan: You have to use here a `MessageBuilder`.
More over it would be better to use a `MessageBuilderFactory` from Spring Integration.
And one more concern: there are `id` and `timestamp` headers which end-user might would like do not have generated on the message.

In other words: you pull too much Spring Integration into this `PubSubSubscriberTemplate`.

Also your idea about `MessageHandler` as callback is also a bit awkward - this also come from Spring Integration world.

I'm not sure why you so don't like Spring Integration and don't want to postpone such a hard work there using appropriate constructions, but with this your current solution it fully looks as you are mixing all the Spring Integration concerns in one place.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberTemplate.java
```diff
@@ -17,18 +17,27 @@
...
 import com.google.pubsub.v1.PullResponse;
 
+import org.springframework.cloud.gcp.pubsub.integration.PubSubHeaderMapper;
 import org.springframework.cloud.gcp.pubsub.support.AcknowledgeablePubsubMessage;
+import org.springframework.cloud.gcp.pubsub.support.GcpPubSubHeaders;
 import org.springframework.cloud.gcp.pubsub.support.PubSubAcknowledger;
 import org.springframework.cloud.gcp.pubsub.support.SubscriberFactory;
+import org.springframework.cloud.gcp.pubsub.support.converter.PubSubMessageConverter;
+import org.springframework.cloud.gcp.pubsub.support.converter.SimplePubSubMessageConverter;
+import org.springframework.integration.mapping.HeaderMapper;
```
artembilan: We can't use it here since it is from Spring Integration which is an `optional` dependency.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberOperations.java
```diff
@@ -66,6 +80,20 @@
...
+	 * Pull a number of messages from a Google Cloud Pub/Sub subscription and convert them to Spring messages with
+	 * the desired payload type.
+	 * @param subscription the subscription name
+	 * @param maxMessages the maximum number of pulled messages
+	 * @param returnImmediately returns immediately even if subscription doesn't contain enough
+	 * messages to satisfy {@code maxMessages}
+	 * @param payloadType the type to which the payload of the Pub/Sub messages should be converted
+	 * @return the list of received acknowledgeable messages
+	 */
+	<T> List<Message<T>> pullAndConvert(String subscription, Integer maxMessages, Boolean returnImmediately,
```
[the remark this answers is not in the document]
meltsufin: I don't (yet) think it's an issue, but let's have the conversation in the other PR.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberTemplate.java
meltsufin: I like Spring Integration. :) I'm just a fan of the template as well, and trying to make sure that it is complete.

`MessageHandler` comes from `spring-messaging` which I thought is not strictly for SI and could be used in the template.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberTemplate.java
meltsufin: Good point.

artembilan:
Sorry, for confusing. Yes, my point was about a `subscribe` side where you are going to poll now a `MessageHandler` abstraction.
The `pull` operations are fully fine. Even conversion is OK there - as long as we have it in AMQP and JMS. Only the problem is with the acknowledge where we are forced to use a Messaging API because of the particular header to transfer.
That's why in Spring AMQP and JMS we don't have any manual ack'ing: all the `receive`  APIs are auto ack'ing. That's my point to not over complicate.
With your current solution we are going to have already 3 different approaches for the same `pull` operation. Not sure that it is going to be so convenient for end-user. 
The plain PubSubMessage and subsequent conversion in the target project feels much straightforward.

We definitely can give a shot for your solution: there is no harm for target users, but I'm not going to support it in the community 😉 

Let's merge everything what you have so far! We spent already too much time without any gain for me. That's really sad I can't find arguments to convince you do not over complicate API and it looks like only me is always against your suggestions.
Maybe it's time for you to stop asking me for review?.. 😉

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberTemplate.java
dhoard: I have code in production that would benefit from this method ...

pseudo code ...
```
@Scheduled(cron = "0 * * * *")
public void runHourly() {
    List<JSONObject> list = pubSubSubscriber.pullAndConvert(...);

    // Convert to batch format
    // Send batch format to third party via HTTP

    // ack list of messages
}
``` 

This can be done with Spring Integration ... just requires more code.

meltsufin:
I think the real source of your sense of it being overcomplicated might be the subscribe aspect of the template. However, that was introduced a long time ago now, and I'm not defending it. 😄

I'm sorry that some discussions get long and tedious, but I hope in the end we get a better API and happier users. We really value your reviews! Your contributions are really critical to the project.

I will clean up this PR a bit more, and then we'll take a final look at whether it does more good than harm. :)

meltsufin:
@dhoard @artembilan I introduced a class called `ConvertedAcknowledgeableMessage` which I think makes things a bit cleaner. PTAL.

dhoard:
This seems odd to me.

Given the similarity of  ...

```
List<AcknowledgeablePubsubMessage> pull(String subscription, Integer maxMessages, Boolean returnImmediately);
```

... and ...

```
<T> List<ConvertedAcknowledgeableMessage<T>> pullAndConvert(String subscription, Integer maxMessages, Boolean returnImmediately, Class<T> payloadType);
```

... I would expect `ConvertedAcknowledgeableMessage` to be named `ConvertedAcknowledgeablePubSubMessage` and extend `AcknowledgeablePubsubMessage`.

meltsufin:
@dhoard I wanted to have `ConvertedAcknowledgeableMessage` extend `AcknowledgeablePubsubMessage` too, but there is a little wrinkle there that make it difficult to do that and at same time reuse it on the subscribe side.

The issue is that the `AcknowledgeablePubsubMessage` contains the ack ID of the message, which enables the option to collect several of those messages into a batch and ack them all at once via the `PubSubAcknowledger`. On the subscribe side, the underlying API doesn't provide us with the ack id. I just created an [issue](https://github.com/GoogleCloudPlatform/google-cloud-java/issues/3479) for this in the Pub/Sub library. So, `ConvertedAcknowledgeableMessage` does away with ack id to make it usable for pull and subscribe.

A agree though that we need to have better consistency there. Maybe acking messages in batch is actually not that useful?

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberTemplate.java
```diff
@@ -240,4 +285,73 @@ public String toString() {
...
+
+	private class ConvertedPulledAcknowledgeablePubsubMessage<T> extends PulledAcknowledgeablePubsubMessage
+			implements ConvertedAcknowledgeablePubsubMessage<T> {
+
+		private final Message<T> message;
+
+		ConvertedPulledAcknowledgeablePubsubMessage(AcknowledgeablePubsubMessage ackableMessage,
+				T payload) {
+			super(ackableMessage.getPubsubMessage(), ackableMessage.getAckId(), ackableMessage.getSubscriptionName());
+			this.message = MessageBuilder.withPayload(payload)
```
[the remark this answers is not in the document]
artembilan: > actually extends `Message<T>`

Right, that's exactly a part I don't like: or we stick with our own entities and stay away of Messaging. Or we just go Messaging way and only use standard, existing (and portable) `Message` implementations.

Fully not sure why you have decided to mix...

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberTemplate.java
```diff
@@ -75,6 +96,16 @@ public Subscriber subscribe(String subscription, MessageReceiver messageReceiver
 		return subscriber;
 	}
 
+	@Override
+	public <T> Subscriber subscribeAndConvert(String subscription,
+			Consumer<ConvertedBasicAcknowledgeablePubsubMessage<T>> messageConsumer, Class<T> payloadType) {
+		return this.subscribe(subscription,
```
artembilan: We don't use `this.` for method and you are still missing a `Assert.notNull(messageConsumer)`

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
```diff
@@ -132,16 +137,36 @@ public PubSubTemplate setMessageConverter(PubSubMessageConverter messageConverte
...
+	@Override
+	public Subscriber subscribe(String subscription, Consumer<BasicAcknowledgeablePubsubMessage> messageConsumer) {
+		return getPubSubSubscriberTemplate().subscribe(subscription, messageConsumer);
+	}
+
+	@Override
+	public <T> Subscriber subscribeAndConvert(String subscription,
+			Consumer<ConvertedBasicAcknowledgeablePubsubMessage<T>> messageConsumer,
+			Class<T> payloadType) {
+		return this.pubSubSubscriberTemplate.subscribeAndConvert(subscription, messageConsumer, payloadType);
```
artembilan: Why inconsistency? In the previous method you use `getPubSubSubscriberTemplate()`, but here it is a `this.pubSubSubscriberTemplate`...

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberOperations.java
```diff
@@ -38,14 +42,34 @@
  */
 public interface PubSubSubscriberOperations {
 
+	/**
+	 * @deprecated As of 1.1, use {@link #subscribe(String, Consumer)} instead.
```
artembilan: Huh? Deja Vu?

Seems for me we discussed with you that this should be fixed.

What am I missing?

artembilan: Oh! I know "why?". This PR is based on your in-progress one #923 

Can't we somehow do not combine work-in-progress in different PR?

Should I review the same code twice and comment about this "As" all the time ? 😉

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberTemplate.java
```diff
@@ -67,14 +76,47 @@ public PubSubSubscriberTemplate(SubscriberFactory subscriberFactory) {
...
+	public PubSubMessageConverter getMessageConverter() {
+		return this.pubSubMessageConverter;
+	}
+
+	public PubSubSubscriberTemplate setMessageConverter(PubSubMessageConverter pubSubMessageConverter) {
+		Assert.notNull(pubSubMessageConverter, "The pubSubMessageConverter can't be null.");
+
+		this.pubSubMessageConverter = pubSubMessageConverter;
+
+		return this;
```
artembilan: Why have you decided to make it in a "chain" style?

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberTemplate.java
```diff
@@ -67,14 +76,47 @@ public PubSubSubscriberTemplate(SubscriberFactory subscriberFactory) {
...
+				this.subscriberFactory.createSubscriber(subscription,
+						(message, ackReplyConsumer) -> messageConsumer.accept(
+								new PushedAcknowledgeablePubsubMessage(message, ackReplyConsumer, subscription)));
+		subscriber.startAsync();
+		return subscriber;
+	}
+
+	@Override
+	public <T> Subscriber subscribeAndConvert(String subscription,
+			Consumer<ConvertedBasicAcknowledgeablePubsubMessage<T>> messageConsumer, Class<T> payloadType) {
```
artembilan: `Assert.notNull(messageConsumer)` ?

artembilan:
I suggest you to close this PR to avoid confusion.

Or mark with the prefix `[DO NOT REVIEW YET]` 😃

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
meltsufin: the previous one is wrong

meltsufin:
@artembilan Sorry for the confusion, I'll use a label for it 'on-hold'.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberTemplate.java
meltsufin: Good question. I didn't even notice. I copied it from `PubSubTemplate` where it was introduced in #659 by @ChengyuanZhao, and none of us noticed it then.
It's not too late to change it to `void`.

artembilan: Well, I'm OK with chaining and it still works with XML configuration. The parser continues to treat it as a setter.
Nevertheless we need to be consistent and if we talk about setter, it must follow Java Beans specification.
Nothing more.

Thanks

meltsufin: I don't think there is any need for supporting chaining here. So, for consistency, I'm removing it.

meltsufin:
Tests and doc changes are still needed, but please take a look to see if this is on the right track.
@artembilan @dhoard

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/converter/ConvertedAcknowledgeablePubsubMessage.java
```diff
@@ -0,0 +1,29 @@
...
+package org.springframework.cloud.gcp.pubsub.support.converter;
+
+import org.springframework.cloud.gcp.pubsub.support.AcknowledgeablePubsubMessage;
+
+/**
+ * @author Mike Eltsufin
+ *
+ * @since 1.1
+ */
+public interface ConvertedAcknowledgeablePubsubMessage<T> extends AcknowledgeablePubsubMessage,
```
artembilan: Need some class level Javadocs.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/converter/ConvertedBasicAcknowledgeablePubsubMessage.java
```diff
@@ -0,0 +1,28 @@
...
+package org.springframework.cloud.gcp.pubsub.support.converter;
+
+import org.springframework.cloud.gcp.pubsub.support.BasicAcknowledgeablePubsubMessage;
+
+/**
+ * @author Mike Eltsufin
+ *
+ * @since 1.1
+ */
+public interface ConvertedBasicAcknowledgeablePubsubMessage<T> extends BasicAcknowledgeablePubsubMessage {
```
artembilan: DITTO

meltsufin:
@artembilan I've updated the docs. So, I think this PR is ready for a final review.

spring-cloud-gcp-docs/src/main/asciidoc/pubsub.adoc
```diff
@@ -85,19 +90,14 @@ Any messages received by `pull()` are not automatically acknowledged.
 Instead, since they are of the kind `AcknowledgeablePubsubMessage`, you can acknowledge them by calling the `ack()` method, or negatively acknowledge them by calling the `nack()` method.
 The `pullAndAck()` method does the same as the `pull()` method and, additionally, acknowledges all received messages.
 
-To acknowledge multiple messages received from `pull()` at once, you can use the `PubSubTemplate.ack()` method.
+The `pullAndConvert()` method does the same as the `pull()` method and, additionally, converts the Pub/Sub binary payload to an object of the desired type, using the converter configured in the template.
+
+To acknowledge multiple messages received from `pull()` or `pullAndConvert` at once, you can use the `PubSubTemplate.ack()` method.
```
artembilan: Missed braces: `pullAndConvert()`

spring-cloud-gcp-pubsub/src/test/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapterTests.java
```diff
@@ -65,31 +64,30 @@
 
 	private String value;
 
-	private BasicAcknowledgeablePubsubMessage message;
+	private ConvertedBasicAcknowledgeablePubsubMessage message;
```
artembilan: Is there some consumer of this property?
I mean do we really need to have it on the class level, but not locally only in the `setUp()` ?

Issue #837, opened 2018-07-06
Add @since 1.1 to all Spanner and Config classes


meltsufin:
Since Spanner and Config were pulled out of the 1.0 track, we should label those classes with `@since 1.1`.

dhoard:
@meltsufin I have checked out a branch and added the `since 1.1` in all of the classes in the module `spring-cloud-gcp-data-spanner`.

Can you clarify exactly which module / classes are you referring to as `Config` and I'll make those Javadoc changes.

meltsufin:
Config is all contained in the `autoconfigure.config` package. Thanks!

PR #842, opened 2018-07-09
Added Javadoc since 1.1

dhoard:
Added Javadoc since 1.1 to resolve https://github.com/spring-cloud/spring-cloud-gcp/issues/837

### DIFF
--- a/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/config/GcpConfigBootstrapConfiguration.java
+++ b/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/config/GcpConfigBootstrapConfiguration.java
@@ -31,6 +31,7 @@ import org.springframework.context.annotation.Configuration;
  * @author Jisha Abubaker
  * @author João André Martins
  * @author Mike Eltsufin
+ * @since 1.1
  */
 @Configuration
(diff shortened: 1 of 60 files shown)


spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/config/GcpConfigBootstrapConfiguration.java
```diff
@@ -31,6 +31,7 @@
  * @author Jisha Abubaker
  * @author João André Martins
  * @author Mike Eltsufin
+ * @since 1.1
```
artembilan: Everything is good, but as we discussed before `@since` on class level should be separated with the blank line, but not stick to the `@author` list.
WDYT?

dhoard: I'll make the change ... no worries.

It seems this isn't consistent across Spring Java code, which makes it confusing.

artembilan: Yeah, agreed.

I guess eventually we all come up with this foundation: https://github.com/spring-io/spring-javaformat/

dhoard:
@artembilan code changed.

PR #844, opened 2018-07-10
Added ackDeadline validation using common method

dhoard:
Added ackDeadline validation using common method


meltsufin:
@dhoard I think you need to rebase your branch because all of the old commits are showing up too.

dhoard:
@meltsufin let me clean this issue up / create new PR if required.

meltsufin:
@dhoard I don't think you need another PR, the delta seems fine against mater.

PR #847, opened 2018-07-10
Add new constructor that accepts AckMode

dhoard:
Add new constructor that accepts AckMode.


dhoard:
@artembilan apologies. I have create Issue https://github.com/spring-cloud/spring-cloud-gcp/issues/848 in which I describe the "Why?" for the constructor, as well as a change to deprecated the existing constructor.

meltsufin:
I think our general policy is to not add optional parameters to the constructor arguments. So, I think we will have to decline this PR. However, I love the constructor javadoc and we need that for the original constructor. :)

dhoard:
Per conversation ... **Closed - no plans to implement**.

PR #851, opened 2018-07-11
Clarified Javadoc

dhoard:
Clarified Javadoc around `AckMode.MANUAL` and `AckMode.AUTO`

### DIFF
--- a/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/AckMode.java
+++ b/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/AckMode.java
@@ -24,14 +24,17 @@ package org.springframework.cloud.gcp.pubsub.integration;
 public enum AckMode {
 
 	/**
-	 * The framework does not ack/nack and the
-	 * {@link com.google.cloud.pubsub.v1.AckReplyConsumer} is sent back to the user.
+	 * The framework does not ack / nack the message. The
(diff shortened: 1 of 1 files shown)


spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/AckMode.java
```diff
@@ -24,14 +24,17 @@
 public enum AckMode {
 
 	/**
-	 * The framework does not ack/nack and the
-	 * {@link com.google.cloud.pubsub.v1.AckReplyConsumer} is sent back to the user.
+	 * The framework does not ack / nack the message. The
+	 * {@link com.google.cloud.pubsub.v1.AckReplyConsumer} is sent back to the user
+	 * for acking or nacking the {@link com.google.pubsub.v1.PubsubMessage}.
```
artembilan: That's does bring to much value. We have a link to the `AckReplyConsumer`, so, essentially you are repeating its JavaDocs here and that's already an end-user responsibility what to do with such a `AckReplyConsumer`.

But yeah... Let it. It's not so critical to spend more time on this.

meltsufin:
@artembilan We can add it on merge.

dhoard:
@meltsufin is it cleaner to close this and have me open a new PR with just the Javadoc change for the existing constructor?

meltsufin:
I'm fine with merging this, but I'll let @artembilan make the call.

meltsufin:
I changed the commit message on merge.

PR #852, opened 2018-07-11
Datastore operations and converter interfaces

ChengyuanZhao:
interfaces for operations and converter


ChengyuanZhao:
fixes #679 .

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/DatastoreDataUtils.java
```diff
@@ -0,0 +1,112 @@
...
+			accessor.setProperty(idProp, key.getNameOrId());
+		}
+		else {
+			key = useKeyComponent(idVal, () -> keyFactory.newKey((Long) idVal),
+					() -> keyFactory.newKey((String) idVal));
+		}
+		return key;
+	}
+
+	private static <T> T useKeyComponent(Object val, Supplier<T> longFunc,
```
dmitry-s: what exactly does this method do? please add comments

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreReadOptions.java
```diff
@@ -0,0 +1,62 @@
...
+import org.springframework.util.Assert;
+
+/**
+ * Options for performing reads and queries in Cloud Datastore.
+ *
+ * @author Chengyuan Zhao
+ *
+ * @since 1.1
+ */
+public class DatastoreReadOptions {
```
dmitry-s: Looks like it's just a wrapper around List<ReadOption>, not sure if it is needed.
Also, there is only one possible ReadOption: https://googlecloudplatform.github.io/google-cloud-java/google-cloud-clients/apidocs/com/google/cloud/datastore/ReadOption.EventualConsistency.html
So, not sure if we even need a list.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
```diff
@@ -0,0 +1,159 @@
...
+		this.datastore = datastore;
+		this.datastoreEntityConverter = datastoreEntityConverter;
+		this.datastoreMappingContext = datastoreMappingContext;
+	}
+
+	@Override
+	public <T> T read(Class<T> entityClass, Key key,
+			@Nullable DatastoreReadOptions readOption) {
+		List<T> results = read(entityClass, ImmutableList.of(key), readOption);
+		return results == null || results.isEmpty() ? null : results.get(0);
```
dmitry-s: maybe make it Optional<T> to avoid returning null?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreOperations.java
```diff
@@ -0,0 +1,114 @@
...
+	/**
+	 * Get a list of entities using a query.
+	 * @param entityClass the type of the entity to retrieve.
+	 * @param query the query to run
+	 * @param readOption optional read options.
+	 * @param <T> the class type of the entity.
+	 * @return a list of entities found.
+	 */
+	<T> List<T> query(Class<T> entityClass, Query<Entity> query,
+			@Nullable DatastoreReadOptions readOption);
```
dmitry-s: probably would it should be readOption**s**

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreReadOptions.java
ChengyuanZhao: This will be used to hold paging and sorting options later on (take a look at the spanner module's read and query option objects). Furthermore, all the APIs from the client lib can take varag read options, so I think we should at least support that much. In the future there may be more settings.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
ChengyuanZhao: The convention seems to be to return the underlying type T instead of Optional<T> for these template methods. I think null is ok here.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreOperations.java
```diff
@@ -0,0 +1,114 @@
...
+
+	/**
+	 * Get a list of all entities of the given class type.
+	 * @param entityClass the class type of entity to retrieve.
+	 * @param <T> the type of the entity.
+	 * @param readOption optional read options.
+	 * @return a list of entities of the given class type.
+	 */
+	<T> List<T> readAll(Class<T> entityClass, @Nullable DatastoreReadOptions readOption);
```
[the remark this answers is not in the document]
ChengyuanZhao: fixed

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/DatastoreDataUtils.java
```diff
@@ -0,0 +1,112 @@
...
+				.getPropertyAccessor(entity);
+		DatastorePersistentProperty ancestorProp = datastorePersistentEntity
+				.ancestorProperty();
+		if (ancestorProp != null) {
+			List<Pair<String, Object>> ancestors = (List<Pair<String, Object>>) accessor
+					.getProperty(ancestorProp);
+			if (ancestors != null) {
+				ancestors.forEach(pair -> {
+					Object val = pair.getSecond();
+					PathElement pathElement;
```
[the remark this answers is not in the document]
ChengyuanZhao: fixed

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreOperations.java
ChengyuanZhao: fixed

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreOperations.java
```diff
@@ -0,0 +1,114 @@
...
+
+	/**
+	 * Get an entity based on a key.
+	 * @param entityClass the type of the entity to get.
+	 * @param key the key of the entity
+	 * @param <T> the class type of the entity.
+	 * @param readOptions optional read options.
+	 * @return the entity that was found with that key.
+	 */
+	<T> T read(Class<T> entityClass, Key key, @Nullable DatastoreReadOptions readOptions);
```
meltsufin: Looking at `MongoTemplate` and [Spring Data JDBC](https://github.com/spring-projects/spring-data-jdbc/blob/master/src/main/java/org/springframework/data/jdbc/core/JdbcAggregateOperations.java) and some other Spring Data examples, `findById` seems more common.  Why are we straying from that pattern?
Also in the examples I've seen the key is passed in as an `Object`. I'm not sure why though.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreReadOptions.java
meltsufin: `ReadOption` has a very specific meaning in Datastore around consistency. I'm not sure if it's a good idea to mix in paging and sorting into the concept.

ChengyuanZhao:
@meltsufin @dmitry-s PTAL and LGTM

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreOperations.java
```diff
@@ -0,0 +1,37 @@
...
+ *
+ * @author Chengyuan Zhao
+ *
+ * @since 1.1
+ */
+public interface DatastoreOperations {
+
+	/**
+	 * Get an entity based on a id.
+	 * @param id the id of the entity
```
meltsufin: We probably need to document the possible options for the id: `String`, `Long`, `Key`?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/DatastoreEntityConverter.java
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * An interface for converting objects to Datastore Entities and vice versa.
+ *
+ * @author Chengyuan Zhao
+ *
+ * @since 1.1
+ */
+public interface DatastoreEntityConverter extends
+		EntityReader<Object, Entity>, EntityWriter<Object, Entity.Builder> {
```
meltsufin: What's the benefit of implementing reader and writer in a single class?

ChengyuanZhao: There is in fact a Spring Data `EntityConverter` class that combines both reader and writer, but we aren't implementing that because it requires the source of the read to be the same type as the type of the write. For example I believe the mongodb one implements that converter interface since it has Document as the source and sink type. 

Having them together in one class also requires that want bi-directional conversion, which I think is always going to be true here.

Do you have a reason why they should be separate?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreOperations.java
ChengyuanZhao: @meltsufin I'll make this change.

This PR is already nearly 100 lines. We would be at ~900 lines with tests. Removing the options class only saves like 50 lines. 

Please let's stick with 100 line for now.

ChengyuanZhao:
@meltsufin PTAL changes made.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/DatastoreEntityConverter.java
dmitry-s: Let's keep them together for now, it looks like it logically belongs together. We can always change if needed.

ChengyuanZhao:
@dmitry-s  would you mind approving? with that LGTM?

dmitry-s:
done

PR #855, opened 2018-07-11
PubSubMessageConverter with generics

### LINKED ISSUE #854: JacksonPubSubMessageConverter with generics
Current:
`public Object fromPubSubMessage(PubsubMessage message, Class<?> payloadType)`

Proposed:
`public <T> T fromPubSubMessage(PubsubMessage message, Class<T> payloadType)`

If we decide to make this change, I think it should be pushed into the 1.0.x stream.

@artembilan

meltsufin:
Provides stricter generics for
PubSubMessageConverter.fromPubSubMessage().

Fixes #854.

### DIFF
--- a/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/converter/JacksonPubSubMessageConverter.java
+++ b/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/converter/JacksonPubSubMessageConverter.java
@@ -65,7 +65,7 @@ public class JacksonPubSubMessageConverter implements PubSubMessageConverter {
 	}
 
 	@Override
-	public Object fromPubSubMessage(PubsubMessage message, Class<?> payloadType) {
+	public <T> T fromPubSubMessage(PubsubMessage message, Class<T> payloadType) {
 		try {
 			return this.objectMapper.readerFor(payloadType).readValue(message.getData().toByteArray());
 		}
--- a/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/converter/PubSubMessageConverter.java
+++ b/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/converter/PubSubMessageConverter.java
@@ -42,5 +42,5 @@ public interface PubSubMessageConverter {
 	 * @param payloadType the desired type of the object
 	 * @return the object converted from the message's payload
 	 */
-	Object fromPubSubMessage(PubsubMessage message, Class<?> payloadType);
+	<T> T fromPubSubMessage(PubsubMessage message, Class<T> payloadType);
 }
--- a/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/converter/SimplePubSubMessageConverter.java
+++ b/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/converter/SimplePubSubMessageConverter.java
@@ -75,21 +75,21 @@ public class SimplePubSubMessageConverter implements PubSubMessageConverter {
 	}
 
 	@Override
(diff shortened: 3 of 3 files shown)


meltsufin:
I don't really understand this argument because the type is already provided as a parameter. So, what's the risk in forcing the return to be of the same type. It's sort of the contract of the method.
If one wants to get objects back, they can call the method with `fromPubSubMessage(message, Object.class)`. What am I missing here?

artembilan:
As I said: I don't feel that it is going to be useful and convenient. I only trust the contract we have so far in other projects: the same aforementioned `org.springframework.messaging.converter.MessageConverter`.

We probably could nail it with the generic on the class level. In this case a Json one would have it as `Object`, so that `fromMessage()` would get a strict contract as well.
However it isn't going to help with that `pullAndConvert()` API...

I would like to hear someone else opinion here...
I'm confused and don't see where it would lead us... 

Thanks

dhoard:
I'm curious ... Is there any other discussion / history as to why a `MessageConverter` isn't constrained to conversion from / to a specific type via generics?

ChengyuanZhao:
I don't think it's an issue of constrained or not. It's that the data is delivered either as bytes or string, and we don't get that extra arbitrary type information for free.

meltsufin:
> (T) message.getData() the getData method always returns a String. However, you're casting it to T, and you're not restricting T to be castable from String. It just doesn't make sense that a function is basically saying "Give me an arbitrary type T and I'll cast a String to T for you", because the underlying object actually is a String and NOT a T. it would fail if T is just not castable from String. Furthermore, it provides no value to the user, because the String cannot behave as any arbitrary T.

`message.getData()` always returns a *ByteString*. Since, we already checked that the requested type `T` is `ByteString`, there is no danger in the cast.

> (T) new String(payload, this.charset) same problem here.

How so? `payloadType` == `T` == `String.class`. Consider the code:
```java	
       else if (payloadType == String.class) {
			result = new String(payload, this.charset);
			result = (T) new String(payload, this.charset);
 	}
```

> (T) ByteBuffer.wrap(payload) same problem here. what's the point of casting to an arbitrary type T if the actual object is only ever going to be a ByteBuffer ?

I think you're missing the fact that T is not arbitrary, but strictly specified as the parameter `payloadType`.


I might still be missing something, but I would really be convinced by seeing a concrete example where this breaks down. I haven't seen one yet.

meltsufin:
@ChengyuanZhao How's this really any different from `<T> T read(Class<T> entityClass, Key key)` in `SpannerTemplate` that you wrote?

ChengyuanZhao:
@meltsufin `<T> T read(Class<T> entityClass, Key key)` will actually instantiate an real instance of T and then populate that real T instance and then return it. 

This situation doesn't do that. It's actually more like a misuse of the type parameter. There are actually only  exactly 4 supported types  (string, bytestring, bytebuffer ,byte[] ). Normally when you have a small finite number of return types supported you just write 4 different functions with  those 4 return types. Doing this checking and unchecked casting is king of just bypassing the whole type system.

ChengyuanZhao:
@meltsufin In contrast, I would say the jackson pubsub message converter actually is using the type parameter by using Jackson to create a real object that backs up the type T provided, and is not limited to a fixed set of acceptable types.

dhoard:
> There are actually only exactly 4 supported types (string, bytestring, bytebuffer ,byte[] ). Normally when you have a small finite number of return types supported you just write 4 different functions with those 4 return types. Doing this if-statement and casting is kind of just bypassing the whole type system.

Changing `PubSubMessageConverter` to be constrained to perform conversion to / from a single type seem much cleaner.

```
public interface PubSubMessageConverter<T> {

	PubsubMessage toPubSubMessage(T payload, Map<String, String> headers);

	<T> T fromPubSubMessage(PubsubMessage message);
}
```

If I really need to handle multiple types, then I would define ...

```
public interface ObjectPubSubMessageConverter implements PubSubMessageConverter<Object>
```

Not good ... but neither is trying to be clever for 4 specific types.

PR #856, opened 2018-07-12
Added new AckMode.AUTO_ACK

### LINKED ISSUE #850: Add third acknowledgment mode to PubSubInboundChannelAdapter / AckMode
`AckMode` currently defines two modes ...

`MANUAL` - the message receiver / sink must ack or nack the message.

`AUTO` - the framework acks the message when the message receiver / sink successfully processes the message. When An exception occurs processing the message, the framework nacks the message. 

The third acknowledgment mode, which is not modeled, is the scenario where an exception occurs when processing the message, but we do not want to nack the message.

This is a very common use case that leverages the Pub / Sub ack deadline timeout as a delay mechanism for message retry. 

Implementation would require adding the new mode in `AckMode` (name ?) and changing `PubSubInboundChannelAdapter`

dhoard:
Added new AckMode.AUTO_ACK and implemented changes in
PubSubInboundChannelAdapter to process the new value.

AckMode.AUTO_ACK adds a new use case to PubSubInboundChannelAdapter in
which a message is only ack'ed if successfully processed. If an
exception occurs processing the message, the message will not be
nack'ed ... resulting in a re-delivery delay by Google Pub / Sub.

Closes #850.

### DIFF
--- a/spring-cloud-gcp-docs/src/main/asciidoc/spring-integration-pubsub.adoc
+++ b/spring-cloud-gcp-docs/src/main/asciidoc/spring-integration-pubsub.adoc
@@ -72,24 +72,32 @@ public PubSubInboundChannelAdapter messageChannelAdapter(
 }
 ----
 
-In the example, we first specify the `MessageChannel` where the adapter is going to write incoming
-messages to.
+In the example, we first specify the `MessageChannel` where the adapter is going to write incoming messages to.
 The `MessageChannel` implementation isn't important here.
-Depending on your use case, you might want to use a `MessageChannel` other than
-`PublishSubscribeChannel`.
+Depending on your use case, you might want to use a `MessageChannel` other than `PublishSubscribeChannel`.
 
 Then, we declare a `PubSubInboundChannelAdapter` bean.
-It requires the channel we just created and a `SubscriberFactory`, which creates `Subscriber`
-objects from the Google Cloud Java Client for Pub/Sub.
+It requires the channel we just created and a `SubscriberFactory`, which creates `Subscriber` objects from the Google Cloud Java Client for Pub/Sub.
 The Spring Boot starter for GCP Pub/Sub provides a configured `SubscriberFactory`.
 
-It is also possible to set the message acknowledgement mode on the adapter, which is automatic by
-default.
-On automatic acking, a message is acked with GCP Pub/Sub if the adapter sent it to the channel and
-no exceptions were thrown.
-If a `RuntimeException` is thrown while the message is processed, then the message is nacked.
-On manual acking, the adapter attaches an `AckReplyConsumer` object to the `Message` headers, which
-users can extract using the `GcpPubSubHeaders.ACKNOWLEDGEMENT` key and use to (n)ack a message.
+The `PubSubInboundChannelAdapter` supports three acknowledgement modes, with `AckMode.AUTO` being the default value;
+
+Automatic acking (`AckMode.AUTO`)
+
+A message is acked with GCP Pub/Sub if the adapter sent it to the channel and no exceptions were thrown.
+If a RuntimeException is thrown while the message is processed, then the mssage is nacked.
+
+Automatic acking OK (`AckMode.AUTO_ACK`)
+
+A message is acked with GCP Pub/Sub if the adapter sent it to the channel and no exceptions were thrown.
+If a `RuntimeException` is thrown while the message is processed, then the message is neither neither acked / nor nacked.
+
+This is useful when using the subscription's ack deadline timeout as a retry delivery backoff mechanism.
+
+Manually acking (`AckMode.MANUAL`)
+
+The adapter attaches an `AckReplyConsumer` object to the `Message` headers.
+Users can extract the `AckReplyConsumer` using the `GcpPubSubHeaders.ACKNOWLEDGEMENT` key and use it to (n)ack a message.
 
 [source,java]
 ----
--- a/spring-cloud-gcp-docs/src/main/asciidoc/spring-integration-pubsub.adoc
+++ b/spring-cloud-gcp-docs/src/main/asciidoc/spring-integration-pubsub.adoc
@@ -85,12 +85,12 @@ The `PubSubInboundChannelAdapter` supports three acknowledgement modes, with `Ac
 Automatic acking (`AckMode.AUTO`)
 
 A message is acked with GCP Pub/Sub if the adapter sent it to the channel and no exceptions were thrown.
-If a RuntimeException is thrown while the message is processed, then the mssage is nacked.
+If a `RuntimeException` is thrown while the message is processed, then the message is nacked.
 
 Automatic acking OK (`AckMode.AUTO_ACK`)
 
 A message is acked with GCP Pub/Sub if the adapter sent it to the channel and no exceptions were thrown.
-If a `RuntimeException` is thrown while the message is processed, then the message is neither neither acked / nor nacked.
+If a `RuntimeException` is thrown while the message is processed, then the message is neither acked / nor nacked.
 
 This is useful when using the subscription's ack deadline timeout as a retry delivery backoff mechanism.
 
--- a/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/AckMode.java
+++ b/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/AckMode.java
@@ -44,7 +44,6 @@ public enum AckMode {
 	 * sent to the channel and processed successfully.
 	 * <p>The framework does not immediately nack the message when the exception occurs,
 	 * and allows the eventual redelivery to take effect.
-	 *
 	 * @since 1.1
 	 */
 	AUTO_ACK
--- a/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/AckMode.java
+++ b/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/AckMode.java
@@ -20,6 +20,7 @@ package org.springframework.cloud.gcp.pubsub.integration;
  * Determines acknowledgement policy for incoming messages from Google Cloud Pub/Sub.
  *
  * @author João André Martins
+ * @author Doug Hoard
  */
 public enum AckMode {
 
@@ -36,5 +37,16 @@ public enum AckMode {
 	 * <p>The framework nacks the {@link com.google.pubsub.v1.PubsubMessage} when an exception
 	 * occurs while processing the message.
 	 */
-	AUTO
+	AUTO,
+
+	/**
+	 * The framework acks the {@link com.google.pubsub.v1.PubsubMessage} after it is
+	 * sent to the channel and processed successfully.
+	 * <p>The framework does nothing (no ack / no nack) of the {@link com.google.pubsub.v1.PubsubMessage}
+	 * when an exception occurs while processing the message.
+	 *
+	 * @since 1.1
+	 */
+	AUTO_ACK
+
 }
--- a/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/AckMode.java
+++ b/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/AckMode.java
@@ -42,8 +42,8 @@ public enum AckMode {
 	/**
 	 * The framework acks the {@link com.google.pubsub.v1.PubsubMessage} after it is
 	 * sent to the channel and processed successfully.
-	 * <p>The framework does nothing (no ack / no nack) of the {@link com.google.pubsub.v1.PubsubMessage}
-	 * when an exception occurs while processing the message.
+	 * <p>The framework does not immediately nack the message when the exception occurs,
+	 * and allows the eventual redelivery to take effect.
 	 *
 	 * @since 1.1
 	 */
--- a/spring-cloud-gcp-pubsub/src/test/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapterTests.java
+++ b/spring-cloud-gcp-pubsub/src/test/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapterTests.java
@@ -16,10 +16,6 @@
 
 package org.springframework.cloud.gcp.pubsub.integration.inbound;
 
-import java.util.concurrent.CountDownLatch;
-import java.util.concurrent.TimeUnit;
-import java.util.concurrent.atomic.AtomicReference;
-
 import com.google.cloud.pubsub.v1.AckReplyConsumer;
 import com.google.cloud.pubsub.v1.MessageReceiver;
 import com.google.protobuf.ByteString;
@@ -49,12 +45,16 @@ import static org.mockito.Mockito.when;
  * @author João André Martins
  * @author Doug Hoard
  */
-@RunWith(MockitoJUnitRunner.class)
+@RunWith(MockitoJUnitRunner.Silent.class)
 public class PubSubInboundChannelAdapterTests {
 
 	public static final String ACK = "ACK";
 
 	public static final String NACK = "NACK";
+
+	public static final String EXCEPTION_MESSAGE = "Forced exception sending message";
+
+	public static final String EXPECTED_EXCEPTION = "Expected exception";
 
 	private PubSubOperations pubSubOperations;
 
@@ -62,32 +62,24 @@ public class PubSubInboundChannelAdapterTests {
 
 	private MessageChannel messageChannel;
 
-	private AtomicReference<String> returnValue;
-
-	private CountDownLatch countDownLatch;
+	private String value;
 
 	private AckReplyConsumer ackReplyConsumer;
-
-	private TestAnswer testAnswer;
 
 	@Before
 	public void setUp() {
 		this.pubSubOperations = mock(PubSubOperations.class);
 		this.pubSubSubscriberOperations = mock(PubSubSubscriberOperations.class);
 		this.messageChannel = mock(MessageChannel.class);
-
-		when(this.messageChannel.send(any())).thenThrow(new RuntimeException("Forced exception sending message"));
-
-		this.returnValue = new AtomicReference<>();
-		this.countDownLatch = new CountDownLatch(1);
-
-		NackAnswer nackAnwser = new NackAnswer(this.returnValue);
-
 		this.ackReplyConsumer = mock(AckReplyConsumer.class);
+		this.value = null;
 
-		doAnswer(nackAnwser).when(this.ackReplyConsumer).nack();
+		when(this.messageChannel.send(any())).thenThrow(
+				new RuntimeException(EXCEPTION_MESSAGE));
 
-		this.testAnswer = new TestAnswer(this.countDownLatch, this.ackReplyConsumer);
+		when(this.pubSubSubscriberOperations.subscribe(
+				anyString(), any(MessageReceiver.class))).then(
+						new MessageReceiverAnswer(this.ackReplyConsumer));
 	}
 
 	@Test(expected = IllegalArgumentException.class)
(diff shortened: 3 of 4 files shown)


spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/AckMode.java
```diff
@@ -36,5 +37,16 @@
 	 * <p>The framework nacks the {@link com.google.pubsub.v1.PubsubMessage} when an exception
 	 * occurs while processing the message.
 	 */
-	AUTO
+	AUTO,
+
+	/**
+	 * The framework acks the {@link com.google.pubsub.v1.PubsubMessage} after it is
+	 * sent to the channel and processed successfully.
+	 * <p>The framework does nothing (no ack / no nack) of the {@link com.google.pubsub.v1.PubsubMessage}
```
meltsufin: I would say "the framework does not immediately nack the message when the exception occurs, and allows the eventual redelivery to take effect."

dhoard:
@artembilan I have added unit tests for both `AckMode.AUTO` and `AckMode,AUTO_ACK` Please review / advise. Thanks.

spring-cloud-gcp-pubsub/src/test/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapterTests.java
```diff
@@ -31,13 +47,166 @@
...
-				this.pubSubTemplate, "testSubscription");
+				pubSubTemplate, "testSubscription");
+
 		adapter.setAckMode(null);
 	}
+
+	@Test
+	public void testAckModeAuto() throws Exception {
+		PubSubTemplate pubSubTemplate = Mockito.mock(PubSubTemplate.class);
+		QueueChannel queueChannel = Mockito.mock(QueueChannel.class);
```
meltsufin: I'm wondering what causes the send to this channel to always fail? Just because it's a mock?

spring-cloud-gcp-pubsub/src/test/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapterTests.java
```diff
@@ -31,13 +47,166 @@
 @RunWith(MockitoJUnitRunner.class)
```
meltsufin: Please add your name above, as the file now mostly your code anyway. :wink:

spring-cloud-gcp-pubsub/src/test/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapterTests.java
```diff
@@ -31,13 +47,166 @@
...
+		adapter.start();
+
+		countDownLatch.await();
+
+		Assert.assertEquals(0, queueChannel.getQueueSize());
+		Assert.assertEquals(NACK, returnValue.value);
+	}
+
+	@Test
+	public void testAckModeAutoAck() throws Exception {
```
meltsufin: Almost all of the code is repeated in this method. Can you please refactor so that the mock setup is only done in one place?

spring-cloud-gcp-pubsub/src/test/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapterTests.java
artembilan: The problem that it is really an anti-pattern to mock against classes. Would be much better to do that against interfaces.
On the other hand I fully don't see reason to mock a `QueueChannel`: according the test body below, there is no any Mockito stubbing and verification. 🤷‍♂️

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/AckMode.java
```diff
@@ -36,5 +37,16 @@
...
 	 */
-	AUTO
+	AUTO,
+
+	/**
+	 * The framework acks the {@link com.google.pubsub.v1.PubsubMessage} after it is
+	 * sent to the channel and processed successfully.
+	 * <p>The framework does not immediately nack the message when the exception occurs,
+	 * and allows the eventual redelivery to take effect.
+	 *
```
artembilan: No blank lines in method/properties JavaDocs.

spring-cloud-gcp-pubsub/src/test/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapterTests.java
```diff
@@ -31,13 +47,166 @@
...
+	@Test
+	public void testAckModeAuto() throws Exception {
+		PubSubTemplate pubSubTemplate = Mockito.mock(PubSubTemplate.class);
+		QueueChannel queueChannel = Mockito.mock(QueueChannel.class);
+
+		final Holder<String> returnValue = new Holder<>();
+		final CountDownLatch countDownLatch = new CountDownLatch(1);
+
+		PubSubInboundChannelAdapter adapter = new PubSubInboundChannelAdapter(
+				pubSubTemplate, "testSubscription");
```
artembilan: I see you stubbing below, but that would be much stable from the Mockito perspective to stub before assigning.

spring-cloud-gcp-pubsub/src/test/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapterTests.java
```diff
@@ -31,13 +47,166 @@
...
+
+		adapter.setAckMode(AckMode.AUTO);
+		adapter.setSendTimeout(10);
+
+		Answer<MessageReceiver> answer = new Answer<MessageReceiver>() {
+			public MessageReceiver answer(InvocationOnMock invocation) throws Throwable {
+				Thread thread = new Thread(new Runnable() {
```
artembilan: What makes you, please, to start a new `Thread` here?

spring-cloud-gcp-pubsub/src/test/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapterTests.java
```diff
@@ -31,13 +47,166 @@
...
+				return null;
+			}
+		};
+
+		when(pubSubTemplate.subscribe(anyString(), any(MessageReceiver.class))).then(answer);
+
+		adapter.setOutputChannel(queueChannel);
+		adapter.start();
+
+		countDownLatch.await();
```
artembilan: That's fully not good to wait in tests forever.
Moreover you need to assert a result of wait.
Out typical pattern is like this:
```
assertTrue(latch.await(10, TimeUnit.SECONDS));
```

spring-cloud-gcp-pubsub/src/test/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapterTests.java
```diff
@@ -31,13 +47,166 @@
...
+							};
+
+							Answer<Boolean> nackAnswer = new Answer<Boolean>() {
+								public Boolean answer(InvocationOnMock invocationOnMock) throws Throwable {
+									returnValue.value = NACK;
+									return null;
+								}
+							};
+
+							AckReplyConsumer ackReplyConsumer = Mockito.mock(AckReplyConsumer.class);
```
artembilan: You can't stub in the active code.
you definitely have to do that in the main thread before any assignments.

spring-cloud-gcp-pubsub/src/test/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapterTests.java
```diff
@@ -31,13 +47,166 @@
...
+
 		adapter.setAckMode(null);
 	}
+
+	@Test
+	public void testAckModeAuto() throws Exception {
+		PubSubTemplate pubSubTemplate = Mockito.mock(PubSubTemplate.class);
+		QueueChannel queueChannel = Mockito.mock(QueueChannel.class);
+
+		final Holder<String> returnValue = new Holder<>();
```
artembilan: What is this `Holder`, please?
Why standard Java `AtomicReference` is not enough?

spring-cloud-gcp-pubsub/src/test/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapterTests.java
```diff
@@ -31,13 +47,166 @@
...
+		};
+
+		when(pubSubTemplate.subscribe(anyString(), any(MessageReceiver.class))).then(answer);
+
+		adapter.setOutputChannel(queueChannel);
+		adapter.start();
+
+		countDownLatch.await();
+
+		Assert.assertEquals(0, queueChannel.getQueueSize());
```
artembilan: Fully unclear why do you need this if your `Answer` doesn't return any value to be wrapped to message to send...
You test here an `AckMode`, not sending/receiving. The unit tests should be as unit as possible.

spring-cloud-gcp-pubsub/src/test/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapterTests.java
```diff
@@ -31,13 +47,166 @@
...
+							Answer<Boolean> ackAnswer = new Answer<Boolean>() {
+								public Boolean answer(InvocationOnMock invocationOnMock) throws Throwable {
+									returnValue.value = ACK;
+									return null;
+								}
+							};
+
+							Answer<Boolean> nackAnswer = new Answer<Boolean>() {
+								public Boolean answer(InvocationOnMock invocationOnMock) throws Throwable {
+									returnValue.value = ACK;
```
artembilan: Why it is about `ACK`, not `NACK`?

I don't know what you are testing here, but this is very complicated and doesn't look like necessary.
But that's again - IMHO. I know people here are tended to disagree with me 😉

spring-cloud-gcp-pubsub/src/test/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapterTests.java
dhoard: Done.

dhoard:
Refactored code to be cleaner.

I still need to document a new `AckMode` in the `spring-integration-pubsub.adoc`

spring-cloud-gcp-pubsub/src/test/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapterTests.java
```diff
@@ -16,28 +16,171 @@
...
+	}
+
+	@Test
+	public void testAckModeAutoAck() throws Exception {
+		PubSubInboundChannelAdapter adapter = new PubSubInboundChannelAdapter(
+				this.pubSubSubscriberOperations, "testSubscription");
+
+		adapter.setAckMode(AckMode.AUTO_ACK);
+
+		when(this.pubSubSubscriberOperations.subscribe(anyString(), any(MessageReceiver.class))).then(this.testAnswer);
```
artembilan: No, you can't stub the same method on the same mock object, independently of the place.
Although I see you expect and produce the same answer. So, would be great to move this stubbing in that `setUp()` - and all your test will still work well!

spring-cloud-gcp-pubsub/src/test/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapterTests.java
```diff
@@ -16,28 +16,171 @@
...
+
+		adapter.setOutputChannel(this.messageChannel);
+
+		try {
+			adapter.start();
+		}
+		catch (Throwable t) {
+
```
artembilan: Would be great to verify expected exception here

spring-cloud-gcp-pubsub/src/test/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapterTests.java
```diff
@@ -16,28 +16,171 @@
...
+		private AtomicReference<String> returnValue;
+
+		NackAnswer(AtomicReference<String> returnValue) {
+			this.returnValue = returnValue;
+		}
+
+		public Boolean answer(InvocationOnMock invocationOnMock) throws Throwable {
+			this.returnValue.set(NACK);
+
+			return null;
```
artembilan: I don't see reason in the separate class for such a simple logic in body. More over it is much cleaner now with Java 8 lambdas.

dhoard:
@artembilan refactored the code to be cleaner ... though left the `Answer` class definitions in the code; I think they make the code easier to read ... but can refactor if you feel strongly about changing them.

spring-cloud-gcp-pubsub/src/test/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapterTests.java
```diff
@@ -49,45 +45,41 @@
  * @author João André Martins
  * @author Doug Hoard
  */
-@RunWith(MockitoJUnitRunner.class)
+@RunWith(MockitoJUnitRunner.Silent.class)
```
artembilan: OK. Each of us has its own codding style and vision how it must look in the end.
I can your PR since it looks like we are not going to agree in some compromise, but I would like to quote JavaDoc of this class:
```
* Using this implementation of the runner is not recommended.
* Engineers should care for removing unused stubbings because they are dead code,
* they add unnecessary details, potentially making the test code harder to comprehend.
* If you have good reasons to use the silent runner, let us know at the mailing list
* or raise an issue in our issue tracker.
* The purpose of silent implementation is to satisfy edge/unanticipated use cases,
* and to offer users an opt-out.
```

So, did you find the real reason to use it in your tests? Have you contacted with Mockito team about your reason? 😉

spring-cloud-gcp-pubsub/src/test/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapterTests.java
```diff
@@ -126,59 +119,60 @@ public void testAckModeAutoAck() throws Exception {
...
+			messageReceiver.receiveMessage(pubsubMessage, this.ackReplyConsumer);
 
 			return null;
 		}
 	}
 
-	private class NackAnswer implements Answer<Boolean> {
+	private class CallbackAnswer implements Answer<Void> {
+
+		private PubSubInboundChannelAdapterTests callback;
```
artembilan: I don't see reason in this inner class because there is a simple access to the `PubSubInboundChannelAdapterTests` as a natural outer-inner classes relationship.
Therefore there is also no need in the `setValue()` because you simple can set it from the lambda:
```
doAnswer(invocation -> {
     this.value = NACK;
     return null;
}).when(this.ackReplyConsumer).nack();
```

That's it. I don't see where is a complexity and overhead during reading.
But with your explicit `PubSubInboundChannelAdapterTests` instance propagation I have to think about its reason and that's confusing. 🤷‍♂️

spring-cloud-gcp-pubsub/src/test/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapterTests.java
```diff
@@ -126,59 +119,60 @@ public void testAckModeAutoAck() throws Exception {
 				this.pubSubSubscriberOperations, "testSubscription");
 
 		adapter.setAckMode(AckMode.AUTO_ACK);
-
-		when(this.pubSubSubscriberOperations.subscribe(anyString(), any(MessageReceiver.class))).then(this.testAnswer);
-
 		adapter.setOutputChannel(this.messageChannel);
 
+		doAnswer(new CallbackAnswer(this, NACK)).when(
+				this.ackReplyConsumer).nack();
```
artembilan: You're stubbing the same method with the same answer.
Actually with Mockito you can't re-stub at all. That's not how things work there, although it may pass for you, but may fail for me and so on. Right, it is a limitation, but still Mockito is very powerful tool.
You have to move this stubbing into the `setUp()`.

spring-cloud-gcp-pubsub/src/test/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapterTests.java
dhoard: I changed to use `MockitoJUnitRunner.Silent.class` because I wanted to set the mock for `nack` even though it shouldn't be called.  A call to `nack` would result in the `value` being set, and hence an error.

With the refactored code, I have reverted this change.

spring-cloud-gcp-pubsub/src/test/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapterTests.java
dhoard: Fixed.

dhoard:
@artembilan `spring-integration-pubsub.adoc` updated. PTAL.

artembilan:
Thanks, but that was @meltsufin 's request: https://github.com/spring-cloud/spring-cloud-gcp/pull/856#pullrequestreview-138722457 😉

spring-cloud-gcp-docs/src/main/asciidoc/spring-integration-pubsub.adoc
```diff
@@ -83,13 +83,29 @@ It requires the channel we just created and a `SubscriberFactory`, which creates
...
-If a `RuntimeException` is thrown while the message is processed, then the message is nacked.
-On manual acking, the adapter attaches an `AckReplyConsumer` object to the `Message` headers, which
-users can extract using the `GcpPubSubHeaders.ACKNOWLEDGEMENT` key and use to (n)ack a message.
+The `PubSubInboundChannelAdapter` supports three acknowledgement modes, with
+`AckMode.AUTO` being the default value;
+
+Automatic acking (`AckMode.AUTO`)
+
+A message is acked with GCP Pub/Sub if the adapter sent it to the channel and no
+exceptions were thrown. If a RuntimeException is thrown while the message is processed,
```
artembilan: Please, follow the rule "One sentence per line": https://asciidoctor.org/docs/asciidoc-recommended-practices/#one-sentence-per-line

dhoard:
@artembilan documentation updated.

There are lines in the file that don't follow the "One sentence per line" rule and were wrapped at 80 - 90 characters, so I was trying to be consistent with previous work.

spring-cloud-gcp-docs/src/main/asciidoc/spring-integration-pubsub.adoc
```diff
@@ -72,24 +72,32 @@ public PubSubInboundChannelAdapter messageChannelAdapter(
...
-no exceptions were thrown.
-If a `RuntimeException` is thrown while the message is processed, then the message is nacked.
-On manual acking, the adapter attaches an `AckReplyConsumer` object to the `Message` headers, which
-users can extract using the `GcpPubSubHeaders.ACKNOWLEDGEMENT` key and use to (n)ack a message.
+The `PubSubInboundChannelAdapter` supports three acknowledgement modes, with `AckMode.AUTO` being the default value;
+
+Automatic acking (`AckMode.AUTO`)
+
+A message is acked with GCP Pub/Sub if the adapter sent it to the channel and no exceptions were thrown.
+If a RuntimeException is thrown while the message is processed, then the mssage is nacked.
```
meltsufin: mssage -> message.
RuntimeException -> `RuntimeException`

spring-cloud-gcp-docs/src/main/asciidoc/spring-integration-pubsub.adoc
```diff
@@ -72,24 +72,32 @@ public PubSubInboundChannelAdapter messageChannelAdapter(
...
+
+Automatic acking OK (`AckMode.AUTO_ACK`)
+
+A message is acked with GCP Pub/Sub if the adapter sent it to the channel and no exceptions were thrown.
+If a `RuntimeException` is thrown while the message is processed, then the message is neither neither acked / nor nacked.
```
meltsufin: neither neither.

PR #859, opened 2018-07-13
Added the ability to use asynchronous acknowledgement

### LINKED ISSUE #739: Create async ack and nack methods
Currently there are only sync methods for ack and nack:
https://github.com/spring-cloud/spring-cloud-gcp/blob/master/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/DefaultPubSubAcknowledger.java

In order to implement async versions, we can use unaryCallable.futureCall
https://github.com/googleapis/gax-java/blob/master/gax/src/main/java/com/google/api/gax/rpc/UnaryCallable.java

dhoard:
Added the ability to use asynchronous acknowledgement for use cases
where the developer wants maximum performance.

Closes #739.


meltsufin:
@dmitry-s PTAL since you might have some good context on the matter.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/DefaultPubSubAcknowledger.java
```diff
@@ -27,23 +27,35 @@
  * ack IDs and subscription name.
  *
  * @author João André Martins
+ * @author Doug Hoard
  */
 class DefaultPubSubAcknowledger implements PubSubAcknowledger {
 
 	private SubscriberStub subscriberStub;
 
+	private boolean asynchronousAcknowledgement;
```
artembilan: I thought about an overloaded `(n)ackAsync()` and let end-user to choose.
The `boolean asyncAck`  could be as an option on the `PubSubSubscriberTemplate`. Or we could think about an appropriate `AckMode`...

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/DefaultSubscriberFactory.java
```diff
@@ -162,6 +165,12 @@ public void setSubscriberStubRetrySettings(RetrySettings subscriberStubRetrySett
 		this.subscriberStubRetrySettings = subscriberStubRetrySettings;
 	}
 
+	public void setAcknowledger(PubSubAcknowledger acknowledger) {
```
artembilan: This is something wrong: it looks like an `acknowledger` abstraction is very tied with the `SubscriberStub`, which is produced exactly by this `SubscriberFactory`. With such a setter you don't give a protection that provided `PubSubAcknowledger` is not going to be based on some other subscriber instance.

Am I missing anything?

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/DefaultPubSubAcknowledger.java
dhoard: I am not opposed to this approach.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/DefaultSubscriberFactory.java
dhoard: You are correct ... to the `setAcknowledger` method will cause issues.  Coming to a concensus on the `boolean asyncAck` on the `PubSubSubscriberTemplate` negates this issue.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/DefaultPubSubAcknowledger.java
```diff
@@ -27,23 +27,35 @@
...
+
 	DefaultPubSubAcknowledger(SubscriberStub subscriberStub) {
 		this.subscriberStub = subscriberStub;
 	}
 
+	public void setAsynchronousAcknowledgement(boolean asynchronousAcknowledgement) {
```
artembilan: Why just an `setAsync(boolean)` with good JavaDocs doesn't work for you?
We are clearly in the `DefaultPubSubAcknowledger` context, so no reason to overhead with its properties.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/DefaultSubscriberFactory.java
artembilan: I don't see that you have really resolved the issue: the setter is still here.
Am I missing anything else?

Thanks

dhoard:
@artembilan I haven't made the changes yet ... I wanted to get your consensus on approach of `setAsync(boolean)` vs `(n)ackAsync()`.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/DefaultPubSubAcknowledger.java
```diff
@@ -27,23 +27,35 @@
...
 
 	private SubscriberStub subscriberStub;
 
+	private boolean asyncAck;
+
 	DefaultPubSubAcknowledger(SubscriberStub subscriberStub) {
 		this.subscriberStub = subscriberStub;
 	}
 
+	public void setAsyncAck(boolean asyncAck) {
```
artembilan: Why do you still insist in adding `Ack` suffix to the obvious option?
The `DefaultPubSubAcknowledger` does nothing more unless `(n)ack`.
So, what is the point to be so verbose?
Plus where is a JavaDoc on this new public API?

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/DefaultSubscriberFactory.java
```diff
@@ -156,12 +159,22 @@ public void setApiClock(ApiClock apiClock) {
...
+	 *                                    including jitter logic, timeout, and exponential backoff
 	 */
 	public void setSubscriberStubRetrySettings(RetrySettings subscriberStubRetrySettings) {
 		this.subscriberStubRetrySettings = subscriberStubRetrySettings;
 	}
 
+	/**
+	 * Set the (n)acknowledgement of messages to be asynchronous
+	 * @param asyncAck
+	 *
```
artembilan: I think there should be some option in the IDE to disable JavaDocs auto-formating and, therefore, there won't be this extra blank line if you don't add it explicitly.

Thanks for understanding.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/DefaultSubscriberFactory.java
```diff
@@ -203,8 +216,7 @@ public Subscriber createSubscriber(String subscriptionName, MessageReceiver rece
 	}
 
 	@Override
-	public PullRequest createPullRequest(String subscriptionName, Integer maxMessages,
-			Boolean returnImmediately) {
+	public PullRequest createPullRequest(String subscriptionName, Integer maxMessages, Boolean returnImmediately) {
```
artembilan: This looks like automatic reformatting in your IDE.
I personally prefer not more than `120` symbols, but guys might have different vision, e.g. `90`.
That's may be why we had such a code before.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/DefaultSubscriberFactory.java
```diff
@@ -156,12 +159,22 @@ public void setApiClock(ApiClock apiClock) {
 	/**
 	 * Set the retry settings for the generated subscriber stubs.
 	 * @param subscriberStubRetrySettings parameters for retrying pull requests when they fail,
-	 * including jitter logic, timeout, and exponential backoff
+	 *                                    including jitter logic, timeout, and exponential backoff
```
artembilan: Although this makes a Javadoc more readable, it doesn't look like a necessary change.
Might be just an auto-refactoring artifact in the IDE...

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/PubSubAcknowledger.java
```diff
@@ -22,9 +22,18 @@
...
  */
 public interface PubSubAcknowledger {
 
+	/**
+	 * (N)acknowledgement of messages will be asynchronous
+	 * @param asyncAck
+	 *
+	 * @since 1.1
+	 */
+	void setAsyncAck(boolean asyncAck);
```
artembilan: I don't this that is going to be as a general contract of the `PubSubAcknowledger`.
Some custom implementation might be just far away from the standard Pub/Sub.

This is an API of the `DefaultSubscriberFactory` indeed, but it is really fully tied with the `DefaultPubSubAcknowledger`.

dhoard: Fixed - removed from the constract of `PubSubAcknowledger`

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/DefaultSubscriberFactory.java
dhoard: Fixed - changed to 90 characters or less.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/DefaultSubscriberFactory.java
dhoard: Fixed - IDE issue.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/DefaultSubscriberFactory.java
dhoard: Fixed.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/DefaultPubSubAcknowledger.java
dhoard: Removed the verbosity since the method is implicit for the classes function.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/DefaultSubscriberFactory.java
```diff
@@ -155,16 +159,27 @@ public void setApiClock(ApiClock apiClock) {
...
 	 */
-	public void setSubscriberStubRetrySettings(RetrySettings subscriberStubRetrySettings) {
+	public void setSubscriberStubRetrySettings(
+			RetrySettings subscriberStubRetrySettings) {
 		this.subscriberStubRetrySettings = subscriberStubRetrySettings;
 	}
 
+	/**
+	 * Set the (n)acknowledgement of messages to be asynchronous
+	 * @param asyncAck
```
artembilan: The parameter must have some description.

Let's turn on a Checkstyle rule on the matter: http://checkstyle.sourceforge.net/config_javadoc.html#NonEmptyAtclauseDescription

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/PubSubAcknowledger.java
```diff
@@ -22,6 +22,7 @@
  * Acknowledges Pub/Sub messages based on their subscription name and ack ID.
  *
  * @author João André Martins
+ * @author Doug Hoard
```
artembilan: Does looks like you have done anything to this class 😃

dhoard: Forgot to revert that ... will fix :)

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/DefaultSubscriberFactory.java
dhoard: Created https://github.com/spring-cloud/spring-cloud-gcp/issues/869 since this issue occurs in various places within the project.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/DefaultSubscriberFactory.java
```diff
@@ -104,7 +107,8 @@ public void setCredentialsProvider(CredentialsProvider credentialsProvider) {
 	}
 
 	/**
-	 * Set the provider for the HTTP headers to be added to the subscribers' REST API calls.
```
meltsufin: Why all of the comment reformatting? It makes the PR harder to read. Can we leave the old formatting?

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/DefaultSubscriberFactory.java
```diff
@@ -263,4 +281,12 @@ public SubscriberStub createSubscriberStub() {
 		}
 	}
 
+	@Override
+	public PubSubAcknowledger createAcknowledger() {
+		DefaultPubSubAcknowledger acknowledger =
+				new DefaultPubSubAcknowledger(createSubscriberStub());
+		acknowledger.setAsync(this.asyncAck);
```
meltsufin: Since the only place this method is currently used is in the `PubSubSubscriberTemplate` constructor, making `asyncAck` a settable property could be pretty confusing because if you change it after the application context is loaded, it won't actually affect anything.

Perhaps adding a `createAsyncAcknowledger` is better, and the caller should really decide which one they want rather than depending on the configuration in the `DefaultSubscriberFactory`. It almost feels that this whole method is unnecessary because the template can simply `DefaultPubSubAcknowledger` on its own by first calling `createSubscriberStub()`.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/DefaultSubscriberFactory.java
```diff
@@ -69,21 +70,23 @@
 
 	private RetrySettings subscriberStubRetrySettings;
 
+	private boolean asyncAck;
```
meltsufin: This property should be exposed in `GcpPubSubProperties`.

spring-cloud-gcp-docs/src/main/asciidoc/spring-integration-pubsub.adoc
```diff
@@ -105,6 +105,10 @@ public MessageHandler messageReceiver() {
 }
 ----
 
+==== Asynchronous acknowledgement / negative acknowledgement of messages
+
+`DefaultSubscriberFactory` provides the ability to perform Asynchronous acknowledgement / negative acknowledgement by setting `setAsyncAck` to `true`.
```
artembilan: @meltsufin ,

Is this doc not enough for you here?

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/DefaultSubscriberFactory.java
artembilan: IMO, the `createAsyncAcknowledger()` is not what could be a part of the `SubscriberFactory` contract.
That's really an implementation details.

And yes, you are right that `template` can mock our `acknowledger` API using direct access. That's why I didn't see the reason of this `createAcknowledger()` contract in the beginning at all...

meltsufin: Let's just remove `createAcknowledger` and do it in 1.0.x as well.

dhoard:
@artembilan @meltsufin with the creation of https://github.com/spring-cloud/spring-cloud-gcp/pull/875 should this be put a hold / closed?

artembilan:
i think async (n)ack still make sense.
This is definitely on hold until we come up with the solution around existing `Acknowledger` abstraction, but I don't see reason to close for now.

Thanks

dhoard:
Adding reference to https://github.com/spring-cloud/spring-cloud-gcp/pull/892 , which will invalidate most of this implementation.

dhoard:
@artembilan @meltsufin I have re-implemented asynchronous (n)ack and modifyAckDeadline based on the recent work done as part of https://github.com/spring-cloud/spring-cloud-gcp/pull/892.

PTAL and let me know your opinion on the approach.  (Documentation and tests still need to be implemented.)

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberTemplate.java
```diff
@@ -171,10 +195,23 @@ private void ack(String subscriptionName, Collection<String> ackIds) {
 		this.subscriberStub.acknowledgeCallable().call(acknowledgeRequest);
 	}
 
+	private void ackAsync(String subscriptionName, Collection<String> ackIds) {
+		AcknowledgeRequest acknowledgeRequest = AcknowledgeRequest.newBuilder()
```
meltsufin: This is duplicating code in `ack()`. There's only one line difference in all of these new `private` implementation methods. You can reduce code duplication by just changing the `private` methods to have an additional `boolean async` parameter.

dhoard:
@meltsufin I like `ack(boolean async)` better than `ackAsync()` ... but thought that the later was more inline with standards.  I don't feel strongly on either way.

meltsufin:
@dhoard I gues `ackAsync()` is more in line with best practices. So we can keep it at the public API level, but for the private implementation methods, let's use the `boolean` param.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
```diff
@@ -162,17 +162,33 @@ public void ack(Collection<AcknowledgeablePubsubMessage> acknowledgeablePubsubMe
 		this.pubSubSubscriberTemplate.ack(acknowledgeablePubsubMessages);
 	}
 
+	@Override
+	public void ackAsync(Collection<AcknowledgeablePubsubMessage> acknowledgeablePubsubMessages) {
+		this.pubSubSubscriberTemplate.ack(acknowledgeablePubsubMessages);
```
artembilan: I believe you mean here `ackAsync()` as a delegate method...

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
```diff
@@ -162,17 +162,33 @@ public void ack(Collection<AcknowledgeablePubsubMessage> acknowledgeablePubsubMe
...
+	}
+
 	@Override
 	public void nack(Collection<AcknowledgeablePubsubMessage> acknowledgeablePubsubMessages) {
 		this.pubSubSubscriberTemplate.nack(acknowledgeablePubsubMessages);
 	}
 
+	@Override
+	public void nackAsync(Collection<AcknowledgeablePubsubMessage> acknowledgeablePubsubMessages) {
+		this.pubSubSubscriberTemplate.nack(acknowledgeablePubsubMessages);
```
artembilan: DITTO, although of course I mean `nackAsync()` 😄

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
```diff
@@ -162,17 +162,33 @@ public void ack(Collection<AcknowledgeablePubsubMessage> acknowledgeablePubsubMe
...
 	@Override
 	public void modifyAckDeadline(Collection<AcknowledgeablePubsubMessage> acknowledgeablePubsubMessages,
 			int ackDeadlineSeconds) {
 		this.pubSubSubscriberTemplate.modifyAckDeadline(acknowledgeablePubsubMessages, ackDeadlineSeconds);
 	}
 
+	@Override
+	public void modifyAckDeadlineAsync(Collection<AcknowledgeablePubsubMessage> acknowledgeablePubsubMessages,
+			int ackDeadlineSeconds) {
+		this.pubSubSubscriberTemplate.modifyAckDeadline(acknowledgeablePubsubMessages, ackDeadlineSeconds);
```
artembilan: DITTO?

Well, looking that you don't use expected delegates, I wonder if you have done that intentionally.

What am I missing in this your work, please?

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/AcknowledgeablePubsubMessage.java
```diff
@@ -38,9 +39,26 @@
 	 */
 	String getSubscriptionName();
 
+	/**
+	 * Acknowledge the message asynchronously
+	 */
+	void ackAsync();
```
meltsufin: I think we should return `java.util.concurrent.Future` for all of these async methods.

artembilan: That's all OK, but I see that downstream API is like this:
```
 public ApiFuture<ResponseT> futureCall(RequestT request) {
```

Shouldn't we expose a `org.springframework.util.concurrent.ListenableFuture` as a return type for possible use-cases when end-user would like to react for the call result or failure?

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberTemplate.java
artembilan: That's all OK, but see my comment about `ListenableFuture`

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/AcknowledgeablePubsubMessage.java
meltsufin: How would we turn `ApiFuture` into `org.springframework.util.concurrent.ListenableFuture`? It doesn't look like `ApiFuture` has an error callback support.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
meltsufin: Nice catch! If we had the async methods return something, this typo could've been avoided. :)

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/AcknowledgeablePubsubMessage.java
meltsufin: Never mind, I see we did it on the publish side.

artembilan:
That is good catch, @meltsufin !

Let's proceed that way then!

`ListenableFuture<> (n)ack()`

and always perform `futureCall()` on the delegate!
 👍

dhoard:
@meltsufin @artembilan I like the idea of `ListenableFuture<String> (n)ack()` but it won't work with the interface definitions as defined.

`AckReplyConsumer` defines the (n)ack as `public void ack()` and `public void nack()`

I can add `ListenableFuture<String> ack()` and `ListenableFuture<String> nack()` into `BasicAcknowledgeablePubsubMessage` and remove the `extends AckReplyConsumer`

Thoughts?

Also, what return type were you all thinking for `ack(Collection<AcknowledgeablePubsubMessage> acknowledgeablePubsubMessages)` ?

`Collection<ListenableFuture<String>>` ?

meltsufin:
@dhoard I think it should be OK to get rid of the `extends AckReplyConsumer`.
For acking a collection, I think you'll want `ListenableFuture<Collection<String>>`.

artembilan:
I think so, too.
It might be a bit tricky to adapt (colect) several `ApiFuture`s into a single `ListenableFuture`, but that is indeed a way we go.

dhoard:
@meltsufin looking for at the return type you suggested, I'm not seeing how `ListenableFuture<Collection<String>>` fits.

For example, doing a `modifyAckDeadline` of a batch of ackIds, each call returns a `ListenableFuture<String>` ... which would leads me to ...

```
	@Override
	public List<ListenableFuture<String>> modifyAckDeadline(
			Collection<AcknowledgeablePubsubMessage> acknowledgeablePubsubMessages, int ackDeadlineSeconds) {
		Assert.notEmpty(acknowledgeablePubsubMessages, "The acknowledgeablePubsubMessages cannot be null.");
		Assert.isTrue(ackDeadlineSeconds >= 0, "The ackDeadlineSeconds must not be negative.");
		
		Map<String, List<String>> acknowledgeablePubsubMessagesMap =
				groupAcknowledgeableMessages(acknowledgeablePubsubMessages);
		
		List<ListenableFuture<String>> listenableFutureList = new LinkedList<>();
		for (Map.Entry<String, List<String>> entry : acknowledgeablePubsubMessagesMap.entrySet()) {
			listenableFutureList.add(modifyAckDeadline(entry.getKey(), entry.getValue(), ackDeadlineSeconds);
		}
		
		return listenableFutureList;
	}
```

What am I missing?

meltsufin:
@dhoard Something like [this](https://stackoverflow.com/a/41905325/9356209)?

dhoard:
@meltsufin that works.

Do you know if the Spring codebase has any similar code? ... I know SO posts / code is licensed under https://creativecommons.org/licenses/by-sa/3.0/ but not sure the legalities with regards to the attribution requirements / Spring codebase requirements.

dhoard:
@meltsufin @artembilan I have pushed code (non-compiling) for direction; This seems to be a mess. 

`BasicAcknowledgableMessage` needs ...

```
	/**
	 * Acknowledge (ack) the message asynchronously
	 * @return ListenableFuture&lt;String&gt;
	 */
	ListenableFuture<String> ack();

	/**
	 * Negatatively achnowledge (nack) the message asynchronously
	 * @return ListenableFuture&lt;String&gt;
	 */
	ListenableFuture<String> nack();
```

`PushedAcknowledgeablePubsubMessage` uses the `AckReplyConsumer`  for the (n)ack ...

```
		@Override
		public void ack() {
			this.ackReplyConsumer.ack();
		}

		@Override
		public void nack() {
			this.ackReplyConsumer.nack();
		}
```

... so it's not clear what could be returned.

Also, the (n)ack and modifyAckDeadline calls return `ApiFuture<Empty>` ... I am returning a `ListenableFuture<Collection<String>>` where the `Collection<String>` are the ackIds ... but don't think that is correct.

:confused:

artembilan:
I think those have to be `subscriptionName` because we really call API in batches, therefore a reply `Future` is about batch and per subscription. You can also have it like `Listenable<Map<String, Collection<String>>>` to be fair with the reality :smile:

dhoard:
@artembilan not sure I follow.

I changed changed the `ListenableFuture<Collection<String>>` and `ListenableFuture<String>` to contain the subscription name(s).

The code compiles ... :scream::cold_sweat::see_no_evil:

meltsufin:
Since the underlying response value is `Empty`, shouldn't we just use `ListenableFuture<Void>` here?
Note that it's different from the publish side where, the underlying service call returns a `String`.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberTemplate.java
```diff
@@ -139,39 +146,63 @@ public Subscriber subscribe(String subscription,
...
+				groupAcknowledgeableMessages(acknowledgeablePubsubMessages);
+
+		Collection<String> subscriptionNames = new LinkedList<>();
+		for (Map.Entry<String, List<String>> entry : acknowledgeablePubsubMessagesMap.entrySet()) {
+			modifyAckDeadline(entry.getKey(), entry.getValue(), ackDeadlineSeconds);
+			subscriptionNames.add(entry.getKey());
+		}
+
+		SettableListenableFuture<Collection<String>> settableListenableFuture = new SettableListenableFuture<>();
+		settableListenableFuture.set(subscriptionNames);
```
meltsufin: This doesn't work. The `SettableListenableFuture` you're returning is completely unhinged from the actual calls to `modifyAckDeadline`.

dhoard: @meltsufin Agreed ... I'll probably more direction from you and @artembilan. The whole `ListenableFuture` use seems to be baffling to me for the receive case.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/BasicAcknowledgeablePubsubMessage.java
```diff
@@ -36,9 +43,15 @@
...
-	 * @return the name of the subscription
+	 * Acknowledge (ack) the message asynchronously
+	 * @return ListenableFuture&lt;String&gt;
 	 */
-	String getSubscriptionName();
+	ListenableFuture<String> ack();
+
+	/**
+	 * Negatatively achnowledge (nack) the message asynchronously
+	 * @return ListenableFuture&lt;String&gt;
```
artembilan: We need some description what the `ListenableFuture` is about. Also consider to use a `{@code }` instead of this ugly manual encoding

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberTemplate.java
dhoard: @meltsufin Would something like this work?

```
	@Override
	public ListenableFuture<Collection<Result>> modifyAckDeadline(
			Collection<AcknowledgeablePubsubMessage> acknowledgeablePubsubMessages, int ackDeadlineSeconds) {
		Assert.notEmpty(acknowledgeablePubsubMessages, "The acknowledgeablePubsubMessages cannot be null.");
		Assert.isTrue(ackDeadlineSeconds >= 0, "The ackDeadlineSeconds must not be negative.");
		
		Collection<Result> results = new LinkedList<>();
		
		SettableListenableFuture<Collection<Result>> settableListenableFuture = new SettableListenableFuture<>();
		settableListenableFuture.set(results);
		
		Map<String, List<String>> acknowledgeablePubsubMessagesMap =
				groupAcknowledgeableMessages(acknowledgeablePubsubMessages);
		
		for (Map.Entry<String, List<String>> entry : acknowledgeablePubsubMessagesMap.entrySet()) {
			Result result = new Result(entry.getKey(), entry.getValue());
			
			ApiFuture<Empty> apiFuture = modifyAckDeadline(entry.getKey(), entry.getValue(), ackDeadlineSeconds);
			ApiFutures.addCallback(apiFuture, new ApiFutureCallback<Empty>() {
				
				@Override
					result.setThrowabe(throwable);
				}

				@Override
				public void onSuccess(Empty empty) {

				}
			});
			
			results.add(result);
		}

		return settableListenableFuture;
	}
```

... where Result is ...

```	
	class Result {

		private String subscription;

		private Collection<String> ackIds;

		private Throwable throwable;
		
		public Result(String subscription, Collection<String> ackIds) {
			this.subscription = subscription;
			this.ackIds = ackIds;
		}
		
		public void setThrowabe(Throwable throwable) {
			this.throwable = throwable;
		}
		
		public String getSubscription() {
			return this.subscription;
		}
		
		public Collection getAckIds() {
			return this.ackIds;
		}
		
		public Throwable getThrowable() {
			return this.throwable;
		}
	};
```

dhoard:
@meltsufin While trying to wrap my head around the `ListenableFuture`, I think I may have found a a design issue with the `PubSubSubscriberTemplate`.

The code appears to be written such that is assumes any `PubSubSubscriberTemplate` can (n)ack / modifyAckDeadline a message regardless of the subscription.

For the `PubSubSubscriberTemplate`, messages are grouped by subscription, then (n)ack is called ...

```
@Override
	public void ack(Collection<AcknowledgeablePubsubMessage> acknowledgeablePubsubMessages) {
		Assert.notEmpty(acknowledgeablePubsubMessages, "The acknowledgeablePubsubMessages cannot be null.");

		groupAcknowledgeableMessages(acknowledgeablePubsubMessages).forEach(this::ack);
	}
```

... calling `ack()` on the message ...

```
@Override
		public void ack() {
			PubSubSubscriberTemplate.this.ack(Collections.singleton(this));
		}
```

... uses the enclosing `PubSubSubscriberTemplate` to make the call ...

```
private void ack(String subscriptionName, Collection<String> ackIds) {
		AcknowledgeRequest acknowledgeRequest = AcknowledgeRequest.newBuilder()
				.addAllAckIds(ackIds)
				.setSubscription(subscriptionName)
				.build();

		this.subscriberStub.acknowledgeCallable().call(acknowledgeRequest);
	}
```

.... but because of projectId and authentication, the `subscriberStub` is really only valid for messages for a specific subscription.

dhoard:
@meltsufin @artembilan PTAL.  I changed the return types for (n)ack and modifyAckDeadline to return `ListenableFuture<Empty>` to be consistent with the return type from the underlying GCP method calls.  I also changed code to force the `Collection<AcknowledgeablePubsubMessage>` passed to the (n)ack and modifyAckDeadline methods to validate the are from the correct project and subscription.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberOperations.java
```diff
@@ -22,10 +22,12 @@
 
 import com.google.cloud.pubsub.v1.MessageReceiver;
 import com.google.cloud.pubsub.v1.Subscriber;
+import com.google.protobuf.Empty;
```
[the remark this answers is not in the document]
dhoard: Makes sense to me.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberTemplate.java
```diff
@@ -288,28 +365,54 @@ public String toString() {
...
-			super(message, subscriptionName);
+			super(projectId, message, subscriptionName);
 			this.ackReplyConsumer = ackReplyConsumer;
 		}
 
 		@Override
-		public void ack() {
-			this.ackReplyConsumer.ack();
+		public ListenableFuture<Empty> ack() {
+			Executor executor = MoreExecutors.directExecutor();
```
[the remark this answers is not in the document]
dhoard: @meltsufin are you referring to the implements of `PulledAcknowledgeablePubsubMessage::ack()` or `PubSubSubscriberTemplate.this.ack()` ?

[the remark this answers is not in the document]
meltsufin: I didn't look at it carefully enough. In the pushed case, we don't really have an async option afforded by the underlying API. So, we should't try to simulate it either. Let me think about how to do that.

meltsufin:
@dhoard Here's something that seems to work to mix in the `void ack()` and `ListenableFuture<Void> ack()` in the same type hierarchy.

`BasicAcknowledgeablePubsubMessage`:
```
<T> T ack();
```

`PushedAcknowledgeablePubsubMessage`:
```
@Override
public Void ack() {
	this.ackReplyConsumer.ack();
	return null;
}
```

This is a bit unconventional, but at least it allows us to avoid simulating asynchronous ack.
WDYT?

dhoard:
@meltsufin I don't like that approach. Seems like a hack to get around an implementation issue. 

Returning a `SettableListenableFuture<Void>` that has already been set (executed) seems valid to me.

Couldn't this happen if the execution of the method were extremely fast / happened before the return?

meltsufin:
Sure, your approach is fine by me too. I just didn't like that direct executor part, I see you've removed it now. @artembilan WDYT?

meltsufin:
@artembilan Friendly ping. Any thoughts on this? I think its a good direction and we should get this in, once the docs are updated.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberOperations.java
```diff
@@ -84,27 +85,32 @@
 	PubsubMessage pullNext(String subscription);
 
 	/**
-	 * Acknowledge a batch of messages. The method will group the messages by subscription name
-	 * and acknowledge them in batches.
+	 * Acknowledge a batch of messages. The messages must have the same project id and be from the same subscription.
 	 * @param acknowledgeablePubsubMessages messages to be acknowledged
+	 * @return {@code ListenableFuture<Void>}
```
artembilan: This should be a `@link` and there indeed must be some description for the return  value.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberOperations.java
```diff
@@ -84,27 +85,32 @@
 	PubsubMessage pullNext(String subscription);
 
 	/**
-	 * Acknowledge a batch of messages. The method will group the messages by subscription name
-	 * and acknowledge them in batches.
+	 * Acknowledge a batch of messages. The messages must have the same project id and be from the same subscription.
```
artembilan: > The messages must have the same project id and be from the same subscription

Why? What is the point then have an argument as as `Collection<AcknowledgeablePubsubMessage>`?
Just add two more args: `projectId` and `subscription`. And request just `Collection<String> ackIds`.

What am I missing? What was wrong with grouping?

Thanks

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberTemplate.java
```diff
@@ -152,84 +153,141 @@ public Subscriber subscribe(String subscription,
...
 	public SubscriberFactory getSubscriberFactory() {
 		return this.subscriberFactory;
 	}
 
 	@Override
-	public void ack(Collection<AcknowledgeablePubsubMessage> acknowledgeablePubsubMessages) {
-		Assert.notEmpty(acknowledgeablePubsubMessages, "The acknowledgeablePubsubMessages can't be null.");
+	public ListenableFuture<Void> ack(
+			Collection<AcknowledgeablePubsubMessage> acknowledgeablePubsubMessages) {
+		Assert.notEmpty(acknowledgeablePubsubMessages, "The acknowledgeablePubsubMessages cannot be null.");
```
artembilan: It's not about `null`. It's really `notEmpty` check.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberTemplate.java
```diff
@@ -152,84 +153,141 @@ public Subscriber subscribe(String subscription,
...
-		groupAcknowledgeableMessages(acknowledgeablePubsubMessages).forEach(this::ack);
+		AcknowledgeablePubsubMessage acknowledgeablePubsubMessage = acknowledgeablePubsubMessages.iterator().next();
+
+		String projectId = acknowledgeablePubsubMessage.getProjectId();
+		String subscriptionName = acknowledgeablePubsubMessage.getSubscriptionName();
+		List<String> ackIds = collectAckIds(projectId, subscriptionName, acknowledgeablePubsubMessages);
+
+		SettableListenableFuture<Void> settableListenableFuture = new SettableListenableFuture<>();
+
+		ApiFuture<Empty> apiFuture = ack(subscriptionName, ackIds);
```
artembilan: Why do we worry about a `projectId` if we do nothing with it in the downstream logic?

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/SubscriberFactory.java
```diff
@@ -28,9 +28,17 @@
...
+ * @author Doug Hoard
  */
 public interface SubscriberFactory {
 
+	/**
+	 * Method to get the project id
+	 * @return the project id
+	 * @since 1.1
+	 */
+	String getProjectId();
```
artembilan: So, what is the reason to have all these new overhead with the `projectId`?

Where is the reason in the commit message?
I didn't get its requirements from the code, too.

Please, elaborate

Thanks

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberOperations.java
dhoard: Agreed - will fix.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberTemplate.java
dhoard: Agreed - will fix.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberOperations.java
dhoard: Given the follow statements ...

1. The `PubSubSubscriberTemplate` requires a `SubscriberFactory`.
2. The `SubscriberFactory` has a `projectId` and requires a `CredentialsProvider`.
3. The GCP Pub / Sub READ role for a subscription can be specific to a user / service account, and it not project global.

This would require any `Collection<AcknowledgeablePubsubMessage>` to contain only messages which have the same `projectId` and subscription name so that they match the `PubSubSubscriberTemplate` (or likewise configured `PubSubSubscriptionTemplate`) that generated the messages. 

The method *could* be changed to accept a `Collection<String> ackIds`, as well as as `projectId` and `subscription` ... but given the `PubSubSubscriberTemplate` deals with `AcknowledgeablePubsubMessage`, in my opinion it is much cleaner and developer friendly to accept `Collection<AcknowledgeablePubsubMessage>` and have the method deal with the work to collection the `Collection<String> ackIds`.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberTemplate.java
dhoard: The `projectId` is used in `collectAckIds()` for validation that the `Collection< AcknowledgeablePubsubMessages>` all contain the same `projectId`.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/SubscriberFactory.java
dhoard: The method `getProjectId()` was added so that the `PubSubSubscriberTemplate` can add the `projectId` to messages.

This allows the `PubSubSubscriberTemplate` batch ack 
 method (`ack(Collection<AcknowledgeablePubsubMessage> acknowledgeablePubsubMessages)`) to be able to validate the messages can be acked by the `PubSubSubscriberTemplate`.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberTemplate.java
dhoard: If the subscriptionId (name) is a String, then I think most developers would expect it to match the plain subscription id (name) when creating the subscription in GCP. (I'm using id (name) since it is inconsistent ... the API uses `id` ... the GCP UI uses `name`).

If we changed it to a `ProjectSubscriptionName`, then it would have the `projectId` and `subscriptionId` (name) ( `https://googleapis.github.io/googleapis/java/all/latest/apidocs/com/google/pubsub/v1/ProjectSubscriptionName.html`

meltsufin: Sure, I think using `ProjectSubscriptionName` also makes sense.

dhoard:
@meltsufin @artembilan I believe the code, tests, and documentation are ready for review.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberOperations.java
```diff
@@ -114,27 +115,32 @@
...
 	 * @param acknowledgeablePubsubMessages messages to be acknowledged
+	 * @return {@code ListenableFuture<Void>} the ListenableFuture for the asynchronous execution
+	 * @since 1.1
```
meltsufin: I don't think putting `@since 1.1` is correct here, given that the method was available in the `1.0` release, granted with a `void` return, but still.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberOperations.java
```diff
@@ -114,27 +115,32 @@
...
+	ListenableFuture<Void> ack(Collection<AcknowledgeablePubsubMessage> acknowledgeablePubsubMessages);
 
 	/**
-	 * Negatively acknowledge a batch of messages. The method will group the messages by subscription name
-	 * and acknowledge them in batches.
+	 * Negatively acknowledge a batch of messages. The messages must have the same project id and be from the same
+	 * subscription.
 	 * @param acknowledgeablePubsubMessages messages to be negatively acknowledged
+	 * @return {@code ListenableFuture<Void>} the ListenableFuture for the asynchronous execution
+	 * @since 1.1
```
meltsufin: See my comment above about `@since`.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberOperations.java
```diff
@@ -114,27 +115,32 @@
...
 	 */
-	void nack(Collection<AcknowledgeablePubsubMessage> acknowledgeablePubsubMessages);
+	ListenableFuture<Void> nack(Collection<AcknowledgeablePubsubMessage> acknowledgeablePubsubMessages);
 
 	/**
-	 * Modify the ack deadline of a batch of messages. The method will group the messages by subscription name
-	 * and modify their ack deadline in batches.
+	 * Modify the ack deadline of a batch of messages. The messages must have the same project id and be from the
```
meltsufin: Why can't we just group automatically by `ProjectSubscriptionName` like before?

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberTemplate.java
```diff
@@ -189,92 +204,136 @@ public Subscriber subscribe(String subscription,
...
 	public SubscriberFactory getSubscriberFactory() {
 		return this.subscriberFactory;
 	}
 
 	@Override
-	public void ack(Collection<AcknowledgeablePubsubMessage> acknowledgeablePubsubMessages) {
-		Assert.notEmpty(acknowledgeablePubsubMessages, "The acknowledgeablePubsubMessages can't be null.");
+	public ListenableFuture<Void> ack(
+			Collection<AcknowledgeablePubsubMessage> acknowledgeablePubsubMessages) {
+		Assert.notEmpty(acknowledgeablePubsubMessages, "The acknowledgeablePubsubMessages cannot be empty.");
```
meltsufin: I think we try to use `can't` everywhere.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberTemplate.java
```diff
@@ -189,92 +204,136 @@ public Subscriber subscribe(String subscription,
...
-	 * @return a map from subscription to list of ack IDs.
-	 */
-	private Map<String, List<String>> groupAcknowledgeableMessages(
+	private List<String> collectAckIds(ProjectSubscriptionName projectSubscriptionName,
 			Collection<AcknowledgeablePubsubMessage> acknowledgeablePubsubMessages) {
-		return acknowledgeablePubsubMessages.stream()
-				.collect(Collectors.groupingBy(AcknowledgeablePubsubMessage::getSubscriptionName,
-						Collectors.mapping(AcknowledgeablePubsubMessage::getAckId, Collectors.toList())));
+
+		List<String> ackIds = new LinkedList<>();
```
meltsufin: Why not `ArrayList` initialized to the size of the `acknowledgeablePubsubMessages`?

spring-cloud-gcp-pubsub/src/test/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberTemplateTests.java
```diff
@@ -114,29 +185,239 @@ public void setUp() {
 	}
 
 	@Test
-	public void testSubscribeAndConvert() {
+	public void testSubscribeAndAck() {
```
meltsufin: A bunch of tests below look almost identical with a lot of code duplication. Can we reduce the code duplication? We can  add utility methods and/or introduce a reusable test method that accepts a strategy for behavior that varies among the tests.
We can also just leave this to a future refactoring.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberTemplate.java
```diff
@@ -189,92 +204,136 @@ public Subscriber subscribe(String subscription,
...
+	public ListenableFuture<Void> modifyAckDeadline(
+			Collection<AcknowledgeablePubsubMessage> acknowledgeablePubsubMessages, int ackDeadlineSeconds) {
+		Assert.notEmpty(acknowledgeablePubsubMessages, "The acknowledgeablePubsubMessages cannot be empty.");
+		Assert.isTrue(ackDeadlineSeconds >= 0, "The ackDeadlineSeconds must not be negative.");
+
+		AcknowledgeablePubsubMessage acknowledgeablePubsubMessage = acknowledgeablePubsubMessages.iterator().next();
+		ProjectSubscriptionName projectSubscriptionName = acknowledgeablePubsubMessage.getProjectSubscriptionName();
+		List<String> ackIds = collectAckIds(projectSubscriptionName, acknowledgeablePubsubMessages);
 
-		groupAcknowledgeableMessages(acknowledgeablePubsubMessages)
```
meltsufin: I liked the automatic grouping by project-subscription. What's wrong with keeping it? Is it just too difficult to convert into the `ListenableFuture`?

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/BasicAcknowledgeablePubsubMessage.java
```diff
@@ -16,18 +16,27 @@
...
+ * @author Doug Hoard
  *
  * @since 1.1
  */
-public interface BasicAcknowledgeablePubsubMessage extends AckReplyConsumer {
+public interface BasicAcknowledgeablePubsubMessage {
+
+	/**
+	 * Accessor for the project subscription name of the Pub/Sub message.
+	 * @return the project id
```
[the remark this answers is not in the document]
dhoard: Fixed.

spring-cloud-gcp-docs/src/main/asciidoc/pubsub.adoc
```diff
@@ -94,7 +94,12 @@ The `pullAndConvert()` method does the same as the `pull()` method and, addition
 
 To acknowledge multiple messages received from `pull()` or `pullAndConvert()` at once, you can use the `PubSubTemplate.ack()` method.
 You can also use the `PubSubTemplate.nack()` for negatively acknowledging messages.
-Using these methods for acknowledging messages in batches is more efficient than acknowledging messages individually.
+
+Using these methods for acknowledging messages in batches is more efficient than acknowledging messages individually, but they *require* the collection of messages to be from the same project and subscription.
+
+
```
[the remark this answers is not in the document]
dhoard: Fixed.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberOperations.java
dhoard: Fixed.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberOperations.java
dhoard: Fixed.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberTemplate.java
dhoard: Fixed.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberTemplate.java
dhoard: Fixed (no specific reason it wasn't set.)

spring-cloud-gcp-pubsub/src/test/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberTemplateTests.java
dhoard: Agreed.  If the tests are valid, then I would like to open an issue to refactor them and defer so we can make sure to get this code in sooner than later.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberOperations.java
dhoard: Grouping by `ProjectSubscriptionName` doesn't solve the issue that messages need to have the same project id as the `PubSubSubscriber` as well as the same permissions to the subscription which is associated with the ackId.

Assuming the constraint above is met ...

Returning a `Collection` of `ListenableFuture`s is not enough information for developer to process / understand what happened / which messages were an issue. I don't see how that provides any value.

Completely my opinion, but I can see a developer thinking ...

"I made the method call with a single `Collection` of messages, but I get a `Collection` of `ListenableFuture`s as a return value.  I can see that the second `ListenableFuture` failed ... but what messages are in that list?"

meltsufin: I wasn't thinking of returning a collection of `ListenableFuture`, but maybe rather `ListenableFuture<Collection<ProjectSubscriptionName>>` to indicate which groups of messages were acked.

PR #864, opened 2018-07-16
Datastore template

### LINKED ISSUE #679: Datastore template and operations

ChengyuanZhao:
fixes #679

### DIFF
--- a/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
+++ b/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
@@ -77,7 +77,7 @@ public class DatastoreTemplate implements DatastoreOperations {
 		if (id instanceof String) {
 			key = keyFactory.newKey((String) id);
 		}
-		else if (id == long.class) {
+		else if (id instanceof Long) {
 			key = keyFactory.newKey((long) id);
 		}
 		else {
--- a/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
+++ b/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
@@ -27,6 +27,7 @@ import com.google.cloud.datastore.Entity.Builder;
 import com.google.cloud.datastore.Key;
 import com.google.cloud.datastore.KeyFactory;
 import com.google.cloud.datastore.Query;
+import com.google.common.annotations.VisibleForTesting;
 
 import org.springframework.cloud.gcp.data.datastore.core.convert.DatastoreEntityConverter;
 import org.springframework.cloud.gcp.data.datastore.core.mapping.DatastoreDataException;
@@ -63,7 +64,8 @@ public class DatastoreTemplate implements DatastoreOperations {
 		this.datastoreMappingContext = datastoreMappingContext;
 	}
 
-	private Key getKeyFromId(Object id, Class entityClass) {
+	@VisibleForTesting
+	Key getKeyFromId(Object id, Class entityClass) {
 		Assert.notNull(id, "Cannot get key for null ID value.");
 		if (id instanceof Key) {
 			return (Key) id;
@@ -94,7 +96,8 @@ public class DatastoreTemplate implements DatastoreOperations {
 				this.datastore.get(getKeyFromId(id, entityClass)));
 	}
 
-	private Key getKey(Object entity) {
+	@VisibleForTesting
+	Key getKey(Object entity) {
 		DatastorePersistentEntity datastorePersistentEntity = this.datastoreMappingContext
 				.getPersistentEntity(entity.getClass());
 		PersistentProperty idProp = datastorePersistentEntity.getIdProperty();
@@ -151,6 +154,9 @@ public class DatastoreTemplate implements DatastoreOperations {
 
 	private <T> Collection<T> convertEntities(Iterator<Entity> entities,
 			Class<T> entityClass) {
+		if (entities == null) {
+			return null;
+		}
 		List<T> results = new ArrayList<>();
 		entities.forEachRemaining(entity -> results
 				.add(this.datastoreEntityConverter.read(entityClass, entity)));
--- a/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
+++ b/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
@@ -64,57 +64,10 @@ public class DatastoreTemplate implements DatastoreOperations {
 		this.datastoreMappingContext = datastoreMappingContext;
 	}
 
-	@VisibleForTesting
-	Key getKeyFromId(Object id, Class entityClass) {
-		Assert.notNull(id, "Cannot get key for null ID value.");
-		if (id instanceof Key) {
-			return (Key) id;
-		}
-		KeyFactory keyFactory = this.datastore.newKeyFactory();
-		keyFactory.setKind(
-				this.datastoreMappingContext.getPersistentEntity(entityClass).kindName());
-		Key key;
-		if (id instanceof String) {
-			key = keyFactory.newKey((String) id);
-		}
-		else if (id instanceof Long) {
-			key = keyFactory.newKey((long) id);
-		}
-		else {
-			// We will use configurable custom converters to try to convert other types to
-			// String or long
-			// in the future.
-			throw new DatastoreDataException(
-					"Keys can only be created using String or long values.");
-		}
-		return key;
-	}
-
 	@Override
 	public <T> T findById(Object id, Class<T> entityClass) {
 		return this.datastoreEntityConverter.read(entityClass,
 				this.datastore.get(getKeyFromId(id, entityClass)));
-	}
-
-	@VisibleForTesting
-	Key getKey(Object entity) {
-		DatastorePersistentEntity datastorePersistentEntity = this.datastoreMappingContext
-				.getPersistentEntity(entity.getClass());
-		PersistentProperty idProp = datastorePersistentEntity.getIdProperty();
-		if (idProp == null) {
-			throw new DatastoreDataException(
-					"Cannot construct key for entity types without ID properties: "
-							+ entity.getClass());
-		}
-		return getKeyFromId(
-				datastorePersistentEntity.getPropertyAccessor(entity).getProperty(idProp),
-				entity.getClass());
-	}
-
-	private Entity convertToEntity(Object entity) {
-		Builder builder = Entity.newBuilder(getKey(entity));
-		this.datastoreEntityConverter.write(entity, builder);
-		return builder.build();
 	}
 
 	@Override
@@ -152,17 +105,6 @@ public class DatastoreTemplate implements DatastoreOperations {
 				entityClass);
 	}
 
-	private <T> Collection<T> convertEntities(Iterator<Entity> entities,
-			Class<T> entityClass) {
-		if (entities == null) {
-			return null;
-		}
-		List<T> results = new ArrayList<>();
-		entities.forEachRemaining(entity -> results
-				.add(this.datastoreEntityConverter.read(entityClass, entity)));
-		return results;
-	}
-
 	@Override
 	public <T> Collection<T> findAll(Class<T> entityClass) {
 		return convertEntities(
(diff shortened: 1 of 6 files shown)


ChengyuanZhao:
@meltsufin @dmitry-s PTAL

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
```diff
@@ -0,0 +1,180 @@
...
+				"A non-null DatastoreEntityConverter is required.");
+		Assert.notNull(datastoreMappingContext,
+				"A non-null DatastoreMappingContext is required.");
+		this.datastore = datastore;
+		this.datastoreEntityConverter = datastoreEntityConverter;
+		this.datastoreMappingContext = datastoreMappingContext;
+	}
+
+	@VisibleForTesting
+	Key getKeyFromId(Object id, Class entityClass) {
```
meltsufin: Package and `private` methods should come after `public` methods according to the Spring code-style.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
```diff
@@ -0,0 +1,180 @@
...
+			return (Key) id;
+		}
+		KeyFactory keyFactory = this.datastore.newKeyFactory();
+		keyFactory.setKind(
+				this.datastoreMappingContext.getPersistentEntity(entityClass).kindName());
+		Key key;
+		if (id instanceof String) {
+			key = keyFactory.newKey((String) id);
+		}
+		else if (id == long.class) {
```
meltsufin: Does this work? Why not `id instanceof Long`
Also, what about `Integer`?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
```diff
@@ -0,0 +1,180 @@
...
+			key = keyFactory.newKey((long) id);
+		}
+		else {
+			// We will use configurable custom converters to try to convert other types to
```
meltsufin: Why not externalize id conversion into its own interface, so that this becomes more configurable from the start?
You would also not need to use the `@VisibleForTesting` since the logic would be in its own class easily testable.

ChengyuanZhao: The conversion from other types to the String or Long type indeed would come from outside. I plan to be utilizing Spring Data's custom converters and conversion service classes and making our implementations separately like we did in Spanner and also that is used in Mongodb and older implementations.

However this specific function must use the template's own instance of the Datastore service, and an instance of the mapping context. it might make sense to put it into some sort of util class,  but at this point in the interest of keeping things as simple as needed until we see otherwise i think we should keep this method in this template class because of it's use of 2 things the template holds and also the rather specific Key-related functionality. 

Totally open to moving it out in the future when we have more of these utility style functions though.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
ChengyuanZhao: Oh my goodness, my bad, will fix.

Regarding not including Integer or int or Long: these will be handled by the custom converters and the Spring Data conversion service that we will introduce later. So, I don't want to include these handful of similar-ish types because that will be redundant because the conversion service we will be using by default handles 90 or so of these common type conversions.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
ChengyuanZhao: will fix.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
```diff
@@ -0,0 +1,181 @@
...
+			return (Key) id;
+		}
+		KeyFactory keyFactory = this.datastore.newKeyFactory();
+		keyFactory.setKind(
+				this.datastoreMappingContext.getPersistentEntity(entityClass).kindName());
+		Key key;
+		if (id instanceof String) {
+			key = keyFactory.newKey((String) id);
+		}
+		else if (id instanceof Long) {
```
dmitry-s: just to clarify, did you mean long or Long?

ChengyuanZhao: So the function newKey takes  long, (small l). However you can't do "instance of long". If you do "id.getClass() == long.class", then that will always be false because a long.getClass() is actually Long.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
```diff
@@ -0,0 +1,181 @@
...
+				"A non-null DatastoreEntityConverter is required.");
+		Assert.notNull(datastoreMappingContext,
+				"A non-null DatastoreMappingContext is required.");
+		this.datastore = datastore;
+		this.datastoreEntityConverter = datastoreEntityConverter;
+		this.datastoreMappingContext = datastoreMappingContext;
+	}
+
+	@Override
+	public <T> T findById(Object id, Class<T> entityClass) {
```
meltsufin: Since the key factory seems to only support Strings and longs, I think it might be more user-friendly to overload this method with `findById(String id, ...)` and `findById(long id, ...)`. We can also have `findByKey(Key key,...)`.

ChengyuanZhao: Doing this would preclude the use of the converters that handle this for the user. Also, the examples I followed, the JDBC template and operations and mongo template and operations, support Object id type. 

I think overloading it would shift the responsibility of this conversion to the user when it's not necessary because we will have access to the converters in this template eventually anyway.

meltsufin: This may be an instance where we deviate from `MongoTemplate` if it makes sense. How would it preclude the use of custom converters? Any method in the template can still call a converter?

ChengyuanZhao: I didn't mean we couldn't use the custom converters anywhere else in the class, but only that we wouldn't be able to use them for the ID conversion when the user wants to do a fast convenient find operation. For example, let's say he has a Car object and a converter of Car to its ID string that isn't just a trivial getter method (perhaps it synthesizes it from several attributes of the car). He wouldn't be able to find by the Car object. 

@mp911de @olivergierke what do you think?

mp911de: From an API perspective, the Template API comes already with a lot of methods so adding overloads increases the API surface. 

Adding overloads would triple the number of methods that accept an Id. I propose to stick with `Object id` which is a common approach and it allows for extensibility in the future (Custom conversions, additional Id types).

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
```diff
@@ -0,0 +1,181 @@
...
+	}
+
+	@Override
+	public <T> void delete(T entity) {
+		this.datastore.delete(getKey(entity));
+	}
+
+	@Override
+	public void deleteAll(Class<?> entityClass) {
+		this.datastore.delete(findAll(entityClass).parallelStream().map(this::getKey)
```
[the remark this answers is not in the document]
ChengyuanZhao: done

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
```diff
@@ -0,0 +1,181 @@
...
+				entityClass);
+	}
+
+	@Override
+	public <T> boolean existsById(Object id, Class<T> entityClass) {
+		return findById(id, entityClass) != null;
+	}
+
+	@VisibleForTesting
+	Key getKeyFromId(Object id, Class entityClass) {
```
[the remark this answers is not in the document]
ChengyuanZhao: done

spring-cloud-gcp-data-datastore/src/test/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplateTests.java
```diff
@@ -0,0 +1,262 @@
...
+
+		when(this.datastore.get(ArgumentMatchers.<Key[]>any()))
+				.thenReturn(ImmutableList.of(e1, e2).iterator());
+		assertThat(this.datastoreTemplate.findAllById(keys, Object.class),
+				contains(ob1, ob2));
+	}
+
+	@Test
+	public void saveTest() {
+		DatastoreTemplate spy = spy(this.datastoreTemplate);
```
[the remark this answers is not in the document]
ChengyuanZhao: I removed the `spy` calls due to the key and id functions. However, some of the functions like deleteAll, count, exists do depend on the findAll and findById functions, which I think is correct. So ,tests for those functions still have spy mocks for the findAll and findById functions.

ChengyuanZhao:
@meltsufin @dmitry-s changes made , PTAL

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
```diff
@@ -0,0 +1,163 @@
...
+	}
+
+	@Override
+	public <T> void delete(T entity) {
+		this.datastore.delete(getKey(entity));
+	}
+
+	@Override
+	public void deleteAll(Class<?> entityClass) {
+		this.datastore.delete(findAllKeys(entityClass));
```
dmitry-s: that looks much better

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/ObjectToKeyFactory.java
```diff
@@ -0,0 +1,93 @@
...
+ *
+ * @since 1.1
+ */
+public interface ObjectToKeyFactory {
+
+	/**
+	 * Gets a key factory for creating new Keys.
+	 * @return a key factory.
+	 */
+	KeyFactory getKeyFactory();
```
meltsufin: This interface is more like a template pattern implementation and not a true contract.
It's like a `KeyFactory` provider because all you need to implement is the `getKeyFactory` method.
Instead I think the `KeyFactory` should be kept out of the interface because it's more of an implementation detail.
So, I would make the default methods just interface methods, remove `getKeyFactory`, and then move the `default` methods to a default implementation of our key factory interface.

ChengyuanZhao: done, PTAL

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
```diff
@@ -0,0 +1,163 @@
...
+				this.datastore.run(Query.newEntityQueryBuilder()
+						.setKind(this.datastoreMappingContext
+								.getPersistentEntity(entityClass).kindName())
+						.build()),
+				entityClass);
+	}
+
+	@Override
+	public <T> boolean existsById(Object id, Class<T> entityClass) {
+		return findById(id, entityClass) != null;
```
meltsufin: Since datastore doesn't seem to support direct `existsById` (or does it), do we really need this method?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/DatastoreServiceObjectToKeyFactory.java
```diff
@@ -0,0 +1,87 @@
...
+public class DatastoreServiceObjectToKeyFactory implements ObjectToKeyFactory {
+
+	private final Datastore datastore;
+
+	public DatastoreServiceObjectToKeyFactory(Datastore datastore) {
+		Assert.notNull(datastore, "A non-null Datastore service is required.");
+		this.datastore = datastore;
+	}
+
+	private KeyFactory getKeyFactory() {
```
meltsufin: private methods should be after public ones according to Spring style guide.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
ChengyuanZhao: I think so. The other spring data modules have this in the Operations interface. Also, this is needed by the Repository interfaces that we must implement.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/DatastoreServiceObjectToKeyFactory.java
ChengyuanZhao: @meltsufin changes made. PTAL

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
meltsufin: OK.

Issue #865, opened 2018-07-16
Stop auto-configuring JacksonPubSubMessageConverter


meltsufin:
Ever since we decided to auto-configure the `JacksonPubSubMessageConverter` based on the presence of the Jackson dependencies in the classpath, we've had multiple issues due to the surprising aspect of that. For example, see #863.

This issue is a proposal require explicit configuration of the `JacksonPubSubMessageConverter`.

saturnism:
Agreed on the surprise. An alternative is to use create a `CompositePubSubMessageConverter`, and delegate String to `Simple` converter, and unrecognized objects to `Jackson` converter.

meltsufin:
One issue I see with the composite idea is that certain basic types like `byte[]` and `String` could be handled by `Jackson` or the `Simple` converter. How would we choose which one to give priority to? If we say use `Simple` in this case, it could be surprising to someone assuming everything is JSON.
Maybe it's not an issue because no one would want to transfer strings or bytes as JSON. I dunno.

PR #867, opened 2018-07-16
Pubsub binder sample fix

### LINKED ISSUE #863: Bug: spring-cloud-gcp-pubsub-binder-sample
Hi,

I was trying to run spring-cloud-gcp/spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-binder-sample/ and the have replaced the values in the application.properties file.

The StreamListener in SinkExample does not work. I have tried sending a simple String. No Luck.

There are no error logs even.

Also, the actuator endpoint does not show the details.

`
binders: {
status: "UNKNOWN",
details: {
pubsub: {
status: "UNKNOWN"
}
}
}
`

dmitry-s:
fixes #863

### DIFF
--- a/spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-binder-sample/src/main/java/com/example/PubSubBinderApplication.java
+++ b/spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-binder-sample/src/main/java/com/example/PubSubBinderApplication.java
@@ -18,12 +18,19 @@ package com.example;
 
 import org.springframework.boot.SpringApplication;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
+import org.springframework.cloud.gcp.pubsub.support.converter.SimplePubSubMessageConverter;
+import org.springframework.context.annotation.Bean;
 
 /**
  * @author João André Martins
  */
 @SpringBootApplication
 public class PubSubBinderApplication {
+
+	@Bean
+	public SimplePubSubMessageConverter messageConverter() {
+		return new SimplePubSubMessageConverter();
+	}
 
 	public static void main(String[] args) {
 		SpringApplication.run(PubSubBinderApplication.class, args);


spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-binder-sample/src/main/java/com/example/PubSubBinderApplication.java
```diff
@@ -18,13 +18,20 @@
...
+import org.springframework.context.annotation.Bean;
 
 /**
  * @author João André Martins
  */
 @SpringBootApplication
 public class PubSubBinderApplication {
 
+	@Bean
+	public SimplePubSubMessageConverter messageConverter() {
```
artembilan: Wait. What is wrong with existing configuration?
The current `PubSubInboundChannelAdapter` has this by default:
```
private PubSubMessageConverter pubSubMessageConverter = new SimplePubSubMessageConverter();
```

So, what am I missing?

However I can guess that an issue is fully related to what @meltsufin has just reported with the #865.

On the other side I fully agree that it's not OK to rely on some pre-configured converter. According Spring Cloud Stream logic we always have to produce `byte[]` from the Binder: https://docs.spring.io/spring-cloud-stream/docs/Elmhurst.RELEASE/reference/htmlsingle/#content-type-management

I remember as I mentioned that once, but it's hard to dig a conversion log now.
So, my point is that this is really a good catch, but it must be fixed explicitly with some `PathThroughConverter` to return just plain `byte[]` or change a `PubSubInboundChannelAdapter` back to the state where we can configure a converter to `null` to just return `byte[]` as is without any conversion attempts.

artembilan:
One more time: the fix is not correct even if it leads to the expected outcome.
Not sure where you see philosophy, if we talk about cold code and proper approach.
More over this is an application code you are fixing here, not a Framework. 
Are you going to ask all the people to add such a converter to their applications to make binder working?..

ChengyuanZhao:
Is it so bad to have the user specify a pubsub message converter like this? I don't think so, and I think that's up to opinion/philosophy. This PR is just for fixing the sample application given the current state of the repository. If we decide to switch up the auto config of these converters later then we can change this sample then. 

Until then I don't think we should let big questions block small fixes for things that definitely need fixing at the moment.

artembilan:
I don't see reason to postpone a proper fix and apply a temporary fix if we have all the players on field already. This is not so productive.
My main concern here that Binder's consumer always must produce a message with `byte[]`. So, no any custom conversions from the application perspective should be exposed.
We shoudn't encourage end-users to think some other way since we always expect `byte[]` on the Spring Cloud Stream.

If you still disagree with me, let it be: feel free to merge it as is! 
I don't see arguments why would I proceed with the wrong solution when the proper one is in hands already...

meltsufin:
This might be a little off topic, but I'm looking into this issue and I'm realizing that the conversion exception is being swallowed somewhere by the binder. It looks like it's because the binder is receiving the messages asynchronously. @artembilan Any ideas on how to make sure that the exception bubbles up to the console?

meltsufin:
Replaced by #872. Closing.

PR #871, opened 2018-07-18
Changed autoconfigured PubSubMessageConverter type

dhoard:
Changed auto-configured `PubSubMessageConverter` type to
`SimplePubSubMessageConverter` to prevent a `JacksonPubSubMessageConverter`
from being used simply because Jackson was included in the project
dependencies.

There are various use cases where Jackson may be included, yet JSON is
not used.  For example, If the project uses an `XmlMapper` (a subclass of `ObjectMapper`) to serialize / de-serialize XML, the auto-configure code would infer that the developer wants a `JacksonPubSubMessageConverter` configured ... even though the usage of the `XmlMapper` may be completely unrelated to any Pub / Sub functionality.

Fixes https://github.com/spring-cloud/spring-cloud-gcp/issues/865

### DIFF
--- a/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
+++ b/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
@@ -21,7 +21,6 @@ import java.util.concurrent.Executors;
 import java.util.function.Consumer;
 import java.util.function.Function;
 
-import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.api.core.ApiClock;
 import com.google.api.gax.batching.BatchingSettings;
 import com.google.api.gax.batching.FlowControlSettings;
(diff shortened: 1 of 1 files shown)


spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
```diff
@@ -344,14 +342,10 @@ public TransportChannelProvider transportChannelProvider() {
...
-
-		@Bean
-		@ConditionalOnMissingBean
-		public PubSubMessageConverter pubSubMessageConverter(ObjectMapper objectMapper) {
-			return new JacksonPubSubMessageConverter(objectMapper);
-		}
+	@Bean
+	@ConditionalOnMissingBean
+	public PubSubMessageConverter pubSubMessageConverter() {
+		return new SimplePubSubMessageConverter();
```
meltsufin: I don't think we need to configure it at all, since it's used by default anyway.

dhoard:
I'll work on the changes.

meltsufin:
@dhoard Please hold off on that. I created #872 to take care of everything.

meltsufin:
Replaced by #872. Closing.

PR #872, opened 2018-07-18
Stop auto-configuring JacksonPubSubMessageConverter

### LINKED ISSUE #865: Stop auto-configuring JacksonPubSubMessageConverter

meltsufin:
Auto-configuring JacksonPubSubMessage converter based on the
presence of ObjectMappber bean in the application context
has turned out to be too magical and surprising.

This PR ensures that by default only the SimplePubSubMessageConverter
will be used.

Fixes #865, #863.

Supersedes #867 and #871.

### DIFF
--- a/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
+++ b/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
@@ -21,7 +21,6 @@ import java.util.concurrent.Executors;
 import java.util.function.Consumer;
 import java.util.function.Function;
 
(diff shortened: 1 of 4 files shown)


meltsufin:
@dmitry-s I believe this should also be applied to the 1.0.x branch, but I'm not sure why #863 is not an issue there?

meltsufin:
@dmitry-s Never mind, I understand now why the Binder sample works in the `1.0.x` branch. It's because by using the converter configured in the `PubSubTemplate` on the inbound side, it avoids the situation of having a different converter on outbound vs. inbound.

In `master`, we basically end up with `JacksonPubSubConverter` outbound, and `SimplePubSubConverter` inbound. Hence the issue.

PR #875, opened 2018-07-19
Remove SubscriberFactory.createAcknowledger

meltsufin:
Currently this default method is just a passthrough,
and the acknowledger can be easily created in the subscriber template.

See [discussion](https://github.com/spring-cloud/spring-cloud-gcp/pull/859#discussion_r203773939).

### DIFF
--- a/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberTemplate.java
+++ b/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberTemplate.java
@@ -27,6 +27,7 @@ import com.google.pubsub.v1.PullRequest;
 import com.google.pubsub.v1.PullResponse;
 
 import org.springframework.cloud.gcp.pubsub.support.AcknowledgeablePubsubMessage;
+import org.springframework.cloud.gcp.pubsub.support.DefaultPubSubAcknowledger;
 import org.springframework.cloud.gcp.pubsub.support.PubSubAcknowledger;
 import org.springframework.cloud.gcp.pubsub.support.SubscriberFactory;
 import org.springframework.util.Assert;
@@ -63,7 +64,7 @@ public class PubSubSubscriberTemplate implements PubSubSubscriberOperations {
 
 		this.subscriberFactory = subscriberFactory;
 		this.subscriberStub = this.subscriberFactory.createSubscriberStub();
-		this.acknowledger = this.subscriberFactory.createAcknowledger();
+		this.acknowledger = new DefaultPubSubAcknowledger(this.subscriberFactory.createSubscriberStub());
 	}
 
 	@Override
--- a/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/DefaultPubSubAcknowledger.java
+++ b/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/DefaultPubSubAcknowledger.java
@@ -28,11 +28,11 @@ import com.google.pubsub.v1.ModifyAckDeadlineRequest;
  *
  * @author João André Martins
  */
-class DefaultPubSubAcknowledger implements PubSubAcknowledger {
+public class DefaultPubSubAcknowledger implements PubSubAcknowledger {
 
 	private SubscriberStub subscriberStub;
 
-	DefaultPubSubAcknowledger(SubscriberStub subscriberStub) {
+	public DefaultPubSubAcknowledger(SubscriberStub subscriberStub) {
 		this.subscriberStub = subscriberStub;
 	}
 
--- a/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/SubscriberFactory.java
+++ b/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/SubscriberFactory.java
@@ -58,11 +58,4 @@ public interface SubscriberFactory {
 	 */
 	SubscriberStub createSubscriberStub();
 
-	/**
-	 * Create a {@link PubSubAcknowledger} to (negatively) acknowledge messages in bulk.
-	 */
-	default PubSubAcknowledger createAcknowledger() {
-		return new DefaultPubSubAcknowledger(createSubscriberStub());
-	}
-
 }


spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/DefaultPubSubAcknowledger.java
```diff
@@ -28,11 +28,11 @@
  *
  * @author João André Martins
  */
-class DefaultPubSubAcknowledger implements PubSubAcknowledger {
+public class DefaultPubSubAcknowledger implements PubSubAcknowledger {
```
artembilan: Why must it become as a `public`?

Seems for me the idea around a `PubSubAcknowledger` abstraction that end-user may decide to implement his/her own customization in his/her own `SubscriberFactory`. So, we have only single place to control a subscriber and `acknowledger`. This way the `PubSubSubscriberTemplate` is free from any implementation details and we have a good distribution of responsibility and dependency injection power.

Yes, template could factor an instance from the `createSubscriberStub()`, but if we don't do that but just delegate to the `SubscriberFactory`, everything is good and no one in a harm.

The `async` option is really a `DeafultSubscriberFactory` responsibility: I don't see any issues on the matter...

So, -1 from me for this change

dhoard:
With this change, wouldn't the `acknowledger` need to be `protected` / non-final so that `PubSubSubscriberTemplate` could be subclassed to add the custom `PubSubAcknowledger`?

With synchronous / asynchronous acknowledgement functionality in `DefaultPubSubAcknowledger`, I would speculate it would rarely be changed..

I don't really see the value of potentially requiring someone to subclass `PubSubSubscriberTemplate` to inject a custom `PubSubAcknowledger`.

meltsufin:
The goal of the `PubSubAcknowledger` abstraction is to allow (n)acking messages in a batch.
We can definitely just move its methods to the `PubSubSubscriberTemplate`. Since, there isn't much support for this change. I will think of other approaches.

artembilan:
But bear in mind you can't introduce class tangle, when you create an `AcknowledgeablePubsubMessage` in the `PubSubSubscriberTemplate` and inject the last one into the first one.
That's how the `PubSubAcknowledger` helps us, but we definitely could do the same with just a `SubscriberStub` injection in the message.

dhoard:
+1 for injecting the `SubscriberStub` into the message and getting rid of `PubSubAcknowledger` / class tangle.

For asynchronous (n)acking, batching should implemented for performance reasons.

meltsufin:
@artembilan @dhoard PTAL. I removed the acknowledger and moved the batch ack/nack methods to the template. I'm also considering changing the signature of the `(n)ack` to take the `Collection<AcknowledgeablePubsubMessage>`  instead of the `Collection<String> ackIds`.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberTemplate.java
```diff
@@ -112,7 +111,7 @@ public Subscriber subscribe(String subscription, MessageReceiver messageReceiver
 
 		List<AcknowledgeablePubsubMessage> ackableMessages = pull(pullRequest);
 
-		this.acknowledger.ack(ackableMessages.stream()
+		this.ack(ackableMessages.stream()
```
artembilan: No `this.` in the method calls.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/AcknowledgeablePubsubMessage.java
```diff
@@ -16,16 +16,17 @@
...
 import com.google.pubsub.v1.PubsubMessage;
 
 /**
  * A {@link PubsubMessage} wrapper that allows it to be acknowledged.
  *
  * <p>To acknowledge {@link AcknowledgeablePubsubMessage} in bulk, using a
- * {@link PubSubAcknowledger} is recommended. To do that, a user must first extract the ack IDs and
- * group them by subscription name.
+ * {@link org.springframework.cloud.gcp.pubsub.core.subscriber.PubSubSubscriberOperations#ack(java.util.Collection,
+ * String)} is recommended. To do that, a user must first extract the ack IDs and group them by subscription name.
```
artembilan: Let's consider to make such a "hard" logic in the template! I mean your idea to proceed with the `AcknowledgeablePubsubMessage` -based API makes sense. and we can group them by the `subscriptionName` already there.

dhoard:
@meltsufin I would prefer changing the method to `Collection<AcknowledgeablePubsubMessage>`.  It seems cleaner and would be easier to implement batch / asynchronous (n)acking.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberTemplate.java
```diff
@@ -111,9 +112,7 @@ public Subscriber subscribe(String subscription, MessageReceiver messageReceiver
 
 		List<AcknowledgeablePubsubMessage> ackableMessages = pull(pullRequest);
 
-		this.ack(ackableMessages.stream()
-				.map(AcknowledgeablePubsubMessage::getAckId)
-				.collect(Collectors.toList()), pullRequest.getSubscription());
+		this.ack(ackableMessages);
```
artembilan: You still use `this.` here. 
This is kinda not our code style and the problem that we can't enforce it via Checkstyle rule...

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberTemplate.java
```diff
@@ -131,7 +130,27 @@ public SubscriberFactory getSubscriberFactory() {
 	}
 
 	@Override
-	public void ack(Collection<String> ackIds, String subscriptionName) {
+	public void ack(Collection<AcknowledgeablePubsubMessage> acknowledgeablePubsubMessages) {
+		groupAcknowledgeableMessages(acknowledgeablePubsubMessages).forEach(this::ack);
```
artembilan: I think we need to add `Assert.notEmpty()`. Otherwise NPE in the `groupAcknowledgeableMessages()`.

PR #879, opened 2018-07-19
Datastore entity converter

### LINKED ISSUE #680: Datastore entity POJO reader/writer, Single items
This includes adding support for custom converters, type-method maps.

dmitry-s:
DatastoreCustomConverter will be added in the next PR

fixes #680

### DIFF
--- a/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/DatastoreEntityConverterImpl.java
+++ b/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/DatastoreEntityConverterImpl.java
@@ -1,0 +1,130 @@
+/*
+ *  Copyright 2018 original author or authors.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package org.springframework.cloud.gcp.data.datastore.core.convert;
+
+import java.util.Map;
+import java.util.function.Function;
+
+import com.google.cloud.Timestamp;
+import com.google.cloud.datastore.Blob;
+import com.google.cloud.datastore.BlobValue;
+import com.google.cloud.datastore.BooleanValue;
+import com.google.cloud.datastore.DoubleValue;
+import com.google.cloud.datastore.Entity;
+import com.google.cloud.datastore.EntityValue;
+import com.google.cloud.datastore.Key;
+import com.google.cloud.datastore.KeyValue;
+import com.google.cloud.datastore.LatLng;
+import com.google.cloud.datastore.LatLngValue;
+import com.google.cloud.datastore.LongValue;
+import com.google.cloud.datastore.StringValue;
+import com.google.cloud.datastore.TimestampValue;
+import com.google.cloud.datastore.Value;
+import com.google.common.collect.ImmutableMap;
+
+import org.springframework.cloud.gcp.data.datastore.core.mapping.DatastoreMappingContext;
+import org.springframework.cloud.gcp.data.datastore.core.mapping.DatastorePersistentEntity;
+import org.springframework.cloud.gcp.data.datastore.core.mapping.DatastorePersistentProperty;
+import org.springframework.data.convert.EntityInstantiator;
+import org.springframework.data.convert.EntityInstantiators;
+import org.springframework.data.mapping.PersistentPropertyAccessor;
+import org.springframework.data.mapping.PreferredConstructor;
+import org.springframework.data.mapping.PropertyHandler;
+import org.springframework.data.mapping.model.ParameterValueProvider;
+import org.springframework.data.mapping.model.PersistentEntityParameterValueProvider;
+
+/**
+ * Class for object to entity and entity to object conversions
+ *
+ * @author Dmitry Solomakha
+ *
+ * @since 1.1
+ */
+public class DatastoreEntityConverterImpl implements DatastoreEntityConverter {
+	private DatastoreMappingContext mappingContext;
+
+	private EntityInstantiators instantiators = new EntityInstantiators();
+
+	private static Map<Class<?>, Function<?, Value<?>>> valFactories;
+
+	static {
+		ImmutableMap.Builder<Class<?>, Function<?, Value<?>>> builder = new ImmutableMap.Builder<>();
+
+		builder.put(Blob.class, (Function<Blob, Value<?>>) BlobValue::of);
+		builder.put(Boolean.class, (Function<Boolean, Value<?>>) BooleanValue::of);
+		builder.put(Double.class, (Function<Double, Value<?>>) DoubleValue::of);
+		builder.put(Entity.class, (Function<Entity, Value<?>>) EntityValue::of);
+		builder.put(Key.class, (Function<Key, Value<?>>) KeyValue::of);
+		builder.put(LatLng.class, (Function<LatLng, Value<?>>) LatLngValue::of);
+		builder.put(Long.class, (Function<Long, Value<?>>) LongValue::of);
+		builder.put(String.class, (Function<String, Value<?>>) StringValue::of);
+		builder.put(Timestamp.class, (Function<Timestamp, Value<?>>) TimestampValue::of);
+
+		valFactories = builder.build();
+	}
+
+	DatastoreEntityConverterImpl(DatastoreMappingContext mappingContext) {
+		this.mappingContext = mappingContext;
+	}
+
+	@Override
+	@SuppressWarnings("unchecked")
+	public <R> R read(Class<R> aClass, Entity entity) {
+		DatastorePersistentEntity<R> persistentEntity = (DatastorePersistentEntity<R>) mappingContext
+				.getPersistentEntity(aClass);
+		PreferredConstructor<?, DatastorePersistentProperty> persistenceConstructor = persistentEntity
+				.getPersistenceConstructor();
+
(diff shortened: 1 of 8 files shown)


spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/DatastoreEntityConverterImpl.java
```diff
@@ -0,0 +1,130 @@
...
+				(PropertyHandler<DatastorePersistentProperty>) datastorePersistentProperty -> {
+					writeProperty(sink, accessor, datastorePersistentProperty);
+				});
+	}
+
+	private <T> void writeProperty(Entity.Builder sink, PersistentPropertyAccessor accessor,
+			DatastorePersistentProperty persistentProperty) {
+
+		Object propertyVal = accessor.getProperty(persistentProperty);
+		Function<T, Value<?>> valueFactory = (Function<T, Value<?>>) valFactories.get(persistentProperty.getType());
```
ChengyuanZhao: This function might not be found if the type isn't supported, and would result in a npe on the next line. I think you should check for null here and throw an exception with a message about unsupported type or return false to signal that the write failed.  I think you should also add a unit test that tests this case where a property is an unsupported type.

spring-cloud-gcp-data-datastore/src/test/java/org/springframework/cloud/gcp/data/datastore/core/convert/DatastoreEntityConverterImplTests.java
```diff
@@ -0,0 +1,176 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * @author Dmitry Solomakha
+ *
+ * @since 1.1
+ */
+
+public class DatastoreEntityConverterImplTests {
```
ChengyuanZhao: I think you should add tests for those two cases where we should throw exceptions:

- when the read value is null (can't be assigned to the needed type)
- when the value can't be written using the Value binder methods in that map

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/DatastoreEntityConverterImpl.java
dmitry-s: good point

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/DatastoreEntityConverterImpl.java
```diff
@@ -0,0 +1,130 @@
...
+
+		valFactories = builder.build();
+	}
+
+	DatastoreEntityConverterImpl(DatastoreMappingContext mappingContext) {
+		this.mappingContext = mappingContext;
+	}
+
+	@Override
+	@SuppressWarnings("unchecked")
```
meltsufin: Can we get rid of this?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/EntityPropertyValueProvider.java
```diff
@@ -0,0 +1,70 @@
...
+			throw new DatastoreDataException(String.format(
+					"The value in column with name %s"
+							+ " could not be converted to the corresponding property in the entity."
+							+ " The property's type is %s.",
+					fieldName, propType));
+		}
+		return (T) value;
+	}
+
+	private Object readSingleWithConversion(DatastorePersistentProperty persistentProperty) {
```
meltsufin: Can we change this to return `<T> T`?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/EntityPropertyValueProvider.java
```diff
@@ -0,0 +1,70 @@
...
+		Object val = entity.getValue(persistentProperty.getFieldName()).get();
+		Class<?> targetType = persistentProperty.getType();
+		if (ClassUtils.isAssignable(targetType, val.getClass())) {
+			return val;
+		}
+		return null;
```
meltsufin: Maybe handle the `null` case here?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/EntityPropertyValueProvider.java
```diff
@@ -0,0 +1,70 @@
...
+import org.springframework.util.ClassUtils;
+
+/**
+ * A {@link PropertyValueProvider} for Datastore entities
+ *
+ * @author Dmitry Solomakha
+ *
+ * @since 1.1
+ */
+public class EntityPropertyValueProvider implements PropertyValueProvider<DatastorePersistentProperty> {
```
meltsufin: Let's have a test for this class as well, even though it's used by the converter.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/DatastoreEntityConverterImpl.java
dmitry-s: actually we can't - it's because of mappingContext.getPersistentEntity(...), which extends AbstractMappingContext

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/DatastoreEntityConverterImpl.java
```diff
@@ -0,0 +1,140 @@
...
+		PersistentPropertyAccessor accessor = persistentEntity
+				.getPropertyAccessor(source);
+		persistentEntity.doWithProperties(
+				(DatastorePersistentProperty datastorePersistentProperty) ->
+					writeProperty(sink, accessor, datastorePersistentProperty)
+				);
+	}
+
+	@SuppressWarnings("unchecked")
+	private <T> void writeProperty(Entity.Builder sink, PersistentPropertyAccessor accessor,
```
ChengyuanZhao: Because this still required the unchecked suppress line, I think we should get rid of the type param T because it is just verbose now. 

You can change the following lines:

`Function valueFactory = valFactories.get(persistentProperty.getType());`

`Value val = (Value) valueFactory.apply(propertyVal);`

Getting rid of the `T` and the `?` is fine I think. We were just casting with those anyway.

spring-cloud-gcp-data-datastore/src/test/java/org/springframework/cloud/gcp/data/datastore/core/convert/DatastoreEntityConverterImplTests.java
```diff
@@ -0,0 +1,167 @@
...
+				.set("blobField", Blob.copyFrom(bytes))
+				.build();
+		DatastoreEntityConverter entityConverter = new DatastoreEntityConverterImpl(new DatastoreMappingContext());
+		TestDatastoreItem item = entityConverter.read(TestDatastoreItem.class, entity);
+
+		assertThat(item.getStringField()).as("validate null field").isNull();
+	}
+
+	@Test(expected = DatastoreDataException.class)
+	public void readTestException() {
```
ChengyuanZhao: This could have a more descriptive test name (about specifically what is wrong).

spring-cloud-gcp-data-datastore/src/test/java/org/springframework/cloud/gcp/data/datastore/core/convert/DatastoreEntityConverterImplTests.java
```diff
@@ -0,0 +1,167 @@
...
+		entityConverter.write(item, builder);
+
+		Entity entity = builder.build();
+
+		assertThat(entity.getValue("stringField").equals(new NullValue()))
+				.as("validate null field").isTrue();
+	}
+
+	@Test(expected = DatastoreDataException.class)
+	public void writeTestException() {
```
ChengyuanZhao: This could have a more descriptive name about specifically what is wrong that is causing the exception.

spring-cloud-gcp-data-datastore/src/test/java/org/springframework/cloud/gcp/data/datastore/core/convert/EntityPropertyValueProviderTests.java
```diff
@@ -0,0 +1,91 @@
...
+				.as("validate latLng field").isEqualTo(LatLng.of(10, 20));
+		assertThat((Timestamp) provider.getPropertyValue(this.persistentEntity.getPersistentProperty("timestampField")))
+				.as("validate timestamp field")
+				.isEqualTo(Timestamp.ofTimeSecondsAndNanos(30, 40));
+		assertThat((Blob) provider.getPropertyValue(this.persistentEntity.getPersistentProperty("blobField")))
+				.as("validate blob field").isEqualTo(Blob.copyFrom(bytes));
+	}
+
+	@Test(expected = DatastoreDataException.class)
+	public void testException() {
```
ChengyuanZhao: More descriptive name please

spring-cloud-gcp-data-datastore/src/test/java/org/springframework/cloud/gcp/data/datastore/core/convert/TestDatastoreItem2.java
```diff
@@ -0,0 +1,44 @@
...
+ */
+
+package org.springframework.cloud.gcp.data.datastore.core.convert;
+
+import java.math.BigInteger;
+
+/**
+ * @author Dmitry Solomakha
+ */
+class TestDatastoreItem2 {
```
ChengyuanZhao: This one seems like a class that is supposed to cause problems in the tests , right ? I think it should be named "TestDatastoreItemUnsupportedFields" or something. I think that would make more sense  when BigInteger and other such fields will be supported (because then there would look like there are 2 working test classes like this)

PR #884, opened 2018-07-24
Datastore template and operations modifications

### LINKED ISSUE #882: Datastore template and operations modifications
- Document what happened if delete something not existing
- Return T from save, such as when it has filled in the ID value 
- deleteAll returns number of deleted
- add deleteAllByID

ChengyuanZhao:
fixes #882

### DIFF
--- a/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreOperations.java
+++ b/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreOperations.java
@@ -83,6 +83,7 @@ public interface DatastoreOperations {
 	/**
 	 * Delete all entities of a given domain type.
 	 * @param entityClass the domain type to delete from Cloud Datastore.
+	 * @return the number of entities that were deleted.
 	 */
 	long deleteAll(Class<?> entityClass);
 
--- a/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreOperations.java
+++ b/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreOperations.java
@@ -41,8 +41,7 @@ public interface DatastoreOperations {
 	/**
 	 * Saves an instance of an object to Cloud Datastore. Behaves as update or insert.
 	 * @param instance the instance to save.
-	 * @return the instance that was saved. This is will have the ID property populated if the
-	 * original saved object had no ID and one was assigned by Cloud Datastore.
+	 * @return the instance that was saved.
 	 */
 	<T> T save(T instance);
 
--- a/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreOperations.java
+++ b/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreOperations.java
@@ -48,7 +48,7 @@ public interface DatastoreOperations {
 
 	/**
 	 * Delete an entity from Cloud Datastore. Deleting IDs that do not exist in Cloud Datastore
-	 * is OK.
+	 * will result in no operation.
 	 * @param id the ID of the entity to delete. If this is actually a
 	 * {@link com.google.cloud.datastore.Key}
 	 * then it will be used. Otherwise it will be attempted to be converted
@@ -61,7 +61,7 @@ public interface DatastoreOperations {
 
 	/**
 	 * Delete multiple IDs from Cloud Datastore. Deleting IDs that do not exist in Cloud Datastore
-	 * is OK.
+	 * will result in no operation.
 	 * @param ids the IDs to delete. If any of these is actually a
 	 * {@link com.google.cloud.datastore.Key}
 	 * then it will be used. Otherwise it will be attempted to be converted
@@ -74,7 +74,7 @@ public interface DatastoreOperations {
 
 	/**
 	 * Delete an entity from Cloud Datastore. Deleting entities that don't exist in Cloud
-	 * Datastore is OK.
+	 * Datastore will result in no operation.
 	 * @param entity the entity to delete.
 	 * @param <T> the entity type
 	 */
--- a/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreOperations.java
+++ b/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreOperations.java
@@ -41,11 +41,14 @@ public interface DatastoreOperations {
 	/**
 	 * Saves an instance of an object to Cloud Datastore. Behaves as update or insert.
 	 * @param instance the instance to save.
+	 * @return the instance that was saved. This is will have the ID property populated if the
+	 * original saved object had no ID and one was assigned by Cloud Datastore.
 	 */
-	<T> void save(T instance);
+	<T> T save(T instance);
 
 	/**
-	 * Delete an entity from Cloud Datastore.
+	 * Delete an entity from Cloud Datastore. Deleting IDs that do not exist in Cloud Datastore
+	 * is OK.
 	 * @param id the ID of the entity to delete. If this is actually a
 	 * {@link com.google.cloud.datastore.Key}
 	 * then it will be used. Otherwise it will be attempted to be converted
@@ -57,7 +60,21 @@ public interface DatastoreOperations {
 	<T> void deleteById(Object id, Class<T> entityClass);
 
 	/**
-	 * Delete an entity from Cloud Datastore.
+	 * Delete multiple IDs from Cloud Datastore. Deleting IDs that do not exist in Cloud Datastore
+	 * is OK.
+	 * @param ids the IDs to delete. If any of these is actually a
+	 * {@link com.google.cloud.datastore.Key}
+	 * then it will be used. Otherwise it will be attempted to be converted
+	 * to an integer or string value and it will be assumed to be a root key value with the
+	 * Kind determined by the entityClass.
+	 * @param entityClass the type of the
+	 * @param <T> ths entity type
+	 */
+	<T> void deleteAllById(Iterable<?> ids, Class<T> entityClass);
+
+	/**
+	 * Delete an entity from Cloud Datastore. Deleting entities that don't exist in Cloud
+	 * Datastore is OK.
 	 * @param entity the entity to delete.
 	 * @param <T> the entity type
 	 */
@@ -67,7 +84,7 @@ public interface DatastoreOperations {
 	 * Delete all entities of a given domain type.
 	 * @param entityClass the domain type to delete from Cloud Datastore.
 	 */
-	void deleteAll(Class<?> entityClass);
+	long deleteAll(Class<?> entityClass);
 
 	/**
 	 * Count all occurrences of entities of the given domain type.
--- a/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
+++ b/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
@@ -34,7 +34,6 @@ import com.google.cloud.datastore.QueryResults;
 import org.springframework.cloud.gcp.data.datastore.core.convert.DatastoreEntityConverter;
 import org.springframework.cloud.gcp.data.datastore.core.convert.ObjectToKeyFactory;
 import org.springframework.cloud.gcp.data.datastore.core.mapping.DatastoreMappingContext;
-import org.springframework.cloud.gcp.data.datastore.core.mapping.DatastorePersistentEntity;
 import org.springframework.util.Assert;
 
 /**
@@ -79,11 +78,7 @@ public class DatastoreTemplate implements DatastoreOperations {
 
 	@Override
 	public <T> T save(T instance) {
-		Entity saved = this.datastore.put(convertToEntity(instance));
-		DatastorePersistentEntity datastorePersistentEntity = this.datastoreMappingContext
-				.getPersistentEntity(instance.getClass());
-		datastorePersistentEntity.getPropertyAccessor(instance).setProperty(
-				datastorePersistentEntity.getIdProperty(), saved.getKey().getNameOrId());
+		this.datastore.put(convertToEntity(instance));
 		return instance;
 	}
 
@@ -161,7 +156,7 @@ public class DatastoreTemplate implements DatastoreOperations {
 	}
 
 	private Key getKey(Object entity) {
-		return this.objectToKeyFactory.getKeyFromObject(entity,
+		return this.objectToKeyFactory.getOrAllocateKeyFromObject(entity,
 				this.datastoreMappingContext.getPersistentEntity(entity.getClass()));
 	}
 
--- a/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
+++ b/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
@@ -77,8 +77,9 @@ public class DatastoreTemplate implements DatastoreOperations {
 	}
 
 	@Override
-	public <T> void save(T instance) {
-		this.datastore.put(convertToEntity(instance));
+	public <T> T save(T instance) {
+		Entity saved = this.datastore.put(convertToEntity(instance));
+		return this.datastoreEntityConverter.read((Class<T>) instance.getClass(), saved);
 	}
 
 	@Override
@@ -87,13 +88,21 @@ public class DatastoreTemplate implements DatastoreOperations {
 	}
 
 	@Override
+	public <T> void deleteAllById(Iterable<?> ids, Class<T> entityClass) {
+		List<Key> keys = getKeysFromIds(ids, entityClass);
+		this.datastore.delete(keys.toArray(new Key[keys.size()]));
+	}
+
+	@Override
 	public <T> void delete(T entity) {
 		this.datastore.delete(getKey(entity));
 	}
 
 	@Override
-	public void deleteAll(Class<?> entityClass) {
-		this.datastore.delete(findAllKeys(entityClass));
+	public long deleteAll(Class<?> entityClass) {
+		Key[] keysToDelete = findAllKeys(entityClass);
+		this.datastore.delete(keysToDelete);
+		return keysToDelete.length;
 	}
 
 	@Override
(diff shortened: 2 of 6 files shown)


ChengyuanZhao:
@meltsufin @dmitry-s this addresses the issues we discussed on call with spring data team today.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreOperations.java
```diff
@@ -41,11 +41,14 @@
...
+	 * @return the instance that was saved. This is will have the ID property populated if the
+	 * original saved object had no ID and one was assigned by Cloud Datastore.
 	 */
-	<T> void save(T instance);
+	<T> T save(T instance);
 
 	/**
-	 * Delete an entity from Cloud Datastore.
+	 * Delete an entity from Cloud Datastore. Deleting IDs that do not exist in Cloud Datastore
+	 * is OK.
```
meltsufin: "OK" seems ambiguous. Results in a no-op?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreOperations.java
```diff
@@ -57,7 +60,21 @@
 	<T> void deleteById(Object id, Class<T> entityClass);
 
 	/**
-	 * Delete an entity from Cloud Datastore.
+	 * Delete multiple IDs from Cloud Datastore. Deleting IDs that do not exist in Cloud Datastore
+	 * is OK.
```
meltsufin: s/OK/no-op.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreOperations.java
```diff
@@ -57,7 +60,21 @@
...
+	 * to an integer or string value and it will be assumed to be a root key value with the
+	 * Kind determined by the entityClass.
+	 * @param entityClass the type of the
+	 * @param <T> ths entity type
+	 */
+	<T> void deleteAllById(Iterable<?> ids, Class<T> entityClass);
+
+	/**
+	 * Delete an entity from Cloud Datastore. Deleting entities that don't exist in Cloud
+	 * Datastore is OK.
```
meltsufin: no-op

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreOperations.java
```diff
@@ -67,7 +84,7 @@
 	 * Delete all entities of a given domain type.
 	 * @param entityClass the domain type to delete from Cloud Datastore.
 	 */
-	void deleteAll(Class<?> entityClass);
+	long deleteAll(Class<?> entityClass);
```
meltsufin: Please add to javadoc the fact that it returns the count.

dmitry-s: please add `@return`to the documentation

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
```diff
@@ -77,23 +77,32 @@ public DatastoreTemplate(Datastore datastore,
 	}
 
 	@Override
-	public <T> void save(T instance) {
-		this.datastore.put(convertToEntity(instance));
+	public <T> T save(T instance) {
+		Entity saved = this.datastore.put(convertToEntity(instance));
+		return this.datastoreEntityConverter.read((Class<T>) instance.getClass(), saved);
```
meltsufin: This seems wasteful to do reverse conversion. Does datastore generate the ID on `put`, or do we create the ID?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
```diff
@@ -77,23 +77,32 @@ public DatastoreTemplate(Datastore datastore,
...
 	}
 
 	@Override
 	public <T> void deleteById(Object id, Class<T> entityClass) {
 		this.datastore.delete(getKeyFromId(id, entityClass));
 	}
 
+	@Override
+	public <T> void deleteAllById(Iterable<?> ids, Class<T> entityClass) {
```
meltsufin: We should return a count here too, since non-existing IDs will cause no-op.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
```diff
@@ -77,23 +77,32 @@ public DatastoreTemplate(Datastore datastore,
...
+		List<Key> keys = getKeysFromIds(ids, entityClass);
+		this.datastore.delete(keys.toArray(new Key[keys.size()]));
+	}
+
 	@Override
 	public <T> void delete(T entity) {
```
meltsufin: Can we return a `boolean` determining whether the delete occurred? Otherwise, we need to make sure that an exception is thrown when delete doesn't happen.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreOperations.java
```diff
@@ -57,7 +60,21 @@
...
+	 * is OK.
+	 * @param ids the IDs to delete. If any of these is actually a
+	 * {@link com.google.cloud.datastore.Key}
+	 * then it will be used. Otherwise it will be attempted to be converted
+	 * to an integer or string value and it will be assumed to be a root key value with the
+	 * Kind determined by the entityClass.
+	 * @param entityClass the type of the
+	 * @param <T> ths entity type
+	 */
+	<T> void deleteAllById(Iterable<?> ids, Class<T> entityClass);
```
meltsufin: Let's return count.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreOperations.java
```diff
@@ -57,7 +60,21 @@
 	<T> void deleteById(Object id, Class<T> entityClass);
```
meltsufin: I think we need to return a `boolean` to indicate whether the delete happened.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
ChengyuanZhao: We won't be aware of how many are actually deleted, because the client lib's delete function returns void and literally says that deleting non-existing keys is "OK". So, it will not be possible to give a count unless we check separately to see how many of those IDs exist beforehand.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
ChengyuanZhao: we cannot. The delete function from the client lib is void and it says literally that deleting non-existing keys is "OK".

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreOperations.java
ChengyuanZhao: We won't be aware of how many are actually deleted, because the client lib's delete function returns void and literally says that deleting non-existing keys is "OK". So, it will not be possible to give a count unless we check separately to see how many of those IDs exist beforehand.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreOperations.java
ChengyuanZhao: The delete function from the client lib is void and it says literally that deleting non-existing keys is "OK".

spring-cloud-gcp-data-datastore/src/test/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplateTests.java
```diff
@@ -192,6 +196,15 @@ public void deleteByIdTest() {
 		verify(this.datastore, times(1)).delete(same(key1));
 	}
 
+	@Test
+	public void deleteAllByIdTest() {
+		Key key1 = createFakeKey();
+		when(this.objectToKeyFactory.getKeyFromId(same(key1), any())).thenReturn(key1);
+		this.datastoreTemplate.deleteAllById(ImmutableList.of(key1, key1),
```
dmitry-s: would be better to have 2 distinct keys

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
ChengyuanZhao: Actually I think we can just extract specifically that key from the returned Entity and set that. will change.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreOperations.java
ChengyuanZhao: done.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreOperations.java
ChengyuanZhao: done

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreOperations.java
ChengyuanZhao: done

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreOperations.java
ChengyuanZhao: done

ChengyuanZhao: done

spring-cloud-gcp-data-datastore/src/test/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplateTests.java
ChengyuanZhao: done

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
ChengyuanZhao: done.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
ChengyuanZhao: done

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
ChengyuanZhao: We can't tell if something was deleted or not. though of course if it did exist and the delete failed we would have an exception.

ChengyuanZhao:
@dmitry-s @meltsufin changes made, PTAL

spring-cloud-gcp-data-datastore/src/test/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplateTests.java
ChengyuanZhao: sorry about that! (i didn't know it emailed you even if you weren't previously on the specific PR. )

artembilan: Well, I'm a commiter on the project, so that is my responsibility to monitor (watch) all the PRs and issues in the project.

I'm not commenting on your PRs because I might just don't understand stuff and my questions for explanation might be just a noise, but that's all doesn't mean I don't review PRs 😄 

Anyway I believe that `done` is a spam not only for me. GitHub has been improved significantly over years to make our, reviewers, life easier.
That's why I prefer to have silent agreement and a proper next commit's message to determine all your changes in a single email.

Thanks and sorry for this spam from me 😉

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
```diff
@@ -77,23 +78,36 @@ public DatastoreTemplate(Datastore datastore,
...
 		this.datastore.delete(getKey(entity));
 	}
 
 	@Override
-	public void deleteAll(Class<?> entityClass) {
-		this.datastore.delete(findAllKeys(entityClass));
+	public long deleteAll(Class<?> entityClass) {
+		Key[] keysToDelete = findAllKeys(entityClass);
```
meltsufin: Since the underlying client library does not return the count, I think we shouldn't do the wasteful find operation, just to get the count. Remember, every operation costs money.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
```diff
@@ -77,23 +78,36 @@ public DatastoreTemplate(Datastore datastore,
 	}
 
 	@Override
-	public <T> void save(T instance) {
-		this.datastore.put(convertToEntity(instance));
+	public <T> T save(T instance) {
+		Entity saved = this.datastore.put(convertToEntity(instance));
+		DatastorePersistentEntity datastorePersistentEntity = this.datastoreMappingContext
+				.getPersistentEntity(instance.getClass());
+		datastorePersistentEntity.getPropertyAccessor(instance).setProperty(
```
meltsufin: I don't understand the need for all this logic. It looks like `getKeyFromObject` already updates the ID property of the `instance`. You should be able to just return `instance` after the operation and it will have the id allocated. What am I missing?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/DatastoreServiceObjectToKeyFactory.java
```diff
@@ -76,9 +77,18 @@ public Key getKeyFromObject(Object entity,
...
 							+ entity.getClass());
 		}
-		return getKeyFromId(
-				datastorePersistentEntity.getPropertyAccessor(entity).getProperty(idProp),
-				datastorePersistentEntity.kindName());
+		PersistentPropertyAccessor accessor = datastorePersistentEntity
+				.getPropertyAccessor(entity);
+		Object idVal = accessor.getProperty(idProp);
+		if (idVal == null) {
+			Key key = this.datastore.allocateId(getKeyFactory()
```
meltsufin: Since you're allocating IDs here, the method should be called something like `getOrCreateKeyFromObject`.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
ChengyuanZhao: This is necessary to get the actual keys to delete. It's NOT just to get the count.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
ChengyuanZhao: You're right. I forgot that it did that!. Will fix

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/DatastoreServiceObjectToKeyFactory.java
ChengyuanZhao: @meltsufin changes made, PTAL.

meltsufin: Thinking about it a bit more, I don't really like a single method having the responsibility for both extracting and allocating keys. The decision whether/how to allocate the keys should probably be at the caller level.
Perhaps, change this method back to `getKeyFromObject` and just return null, if null; and also just add a method called `allocateKeyForObject`. The caller would decide whether it needs to use `allocateKey`. WDYT?

ChengyuanZhao: Yes, it sounds good to me to let the caller decide. I'll make a follow up issue for adding the mechanism by which the user chooses (like an annotation `@AutoAllocateId` )

ChengyuanZhao: #891

PR #892, opened 2018-07-25
Create an interface for AcknowledgeablePubsubMessage

meltsufin:
Since pulled messages contain ack IDs, but subscribed don't,
we need a common abstraction for both that simply includes
ack and nack methods without exposing the details of ack IDs.
This change acomplishes that by turning AcknowledgeablePubsubMessage
into an interface and creating PulledAcknowledgeablePubsubMessage as
the only implementation so far. However, a PushedAcknowledgeablePubsubMessage
is also envisioned on the subscribe side.

See: https://github.com/GoogleCloudPlatform/google-cloud-java/issues/3479

This will help us get closer to completing #841.


spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/PulledAcknowledgeablePubsubMessage.java
```diff
@@ -0,0 +1,86 @@
...
+ *
+ * <p>To acknowledge {@link PulledAcknowledgeablePubsubMessage} in bulk, using a
+ * {@link org.springframework.cloud.gcp.pubsub.core.subscriber.PubSubSubscriberOperations#ack(java.util.Collection)}
+ * is recommended.
+ *
+ * @author João André Martins
+ * @author Mike Eltsufin
+ */
+public class PulledAcknowledgeablePubsubMessage implements AcknowledgeablePubsubMessage {
+
```
dhoard: Thoughts on keep state so that ...

acking an acked message is a no-op
nacking a nacked message is a no-op
acking an nacked message throws an Exception
nacking an acked message throws an Exception

... ?

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/AcknowledgeablePubsubMessage.java
```diff
@@ -16,67 +16,21 @@
...
-	public void nack() {
-		ModifyAckDeadlineRequest modifyAckDeadlineRequest = ModifyAckDeadlineRequest.newBuilder()
-				.setAckDeadlineSeconds(0)
-				.addAckIds(this.ackId)
-				.setSubscription(this.subscriptionName)
-				.build();
-
-		this.subscriberStub.modifyAckDeadlineCallable().call(modifyAckDeadlineRequest);
-	}
+	PubsubMessage getPubsubMessage();
```
dhoard: Thoughts on adding `getSubscriptionName()` ?  I can see a scenario where code processing a message may want to know which subscription the message is from.

dhoard: Also what about the scenario where a developer wants to modify a messages ack deadline?

This would necessitate a `public void modifyAckDeadline(int ackDeadlineSeconds)` type of method in the interface.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/PulledAcknowledgeablePubsubMessage.java
meltsufin: I feel there might be some risk of getting out of sync with the state in the service. What's the benefit of this?

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/AcknowledgeablePubsubMessage.java
meltsufin: Adding `subscriptionName` should be OK.
Adding `modifyAckDeadline` would be an issue in the subscribe case because all we have there is the `AckReplyConsumer`, but can be add it to `PulledAcknowledgeablePubsubMessage`.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/PulledAcknowledgeablePubsubMessage.java
dhoard: If a developer has a bug, which results in a `nack()` call on a message which was previously `acked()`, etc. ... what is will happen?

It may be enough to just document the behavior.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/AcknowledgeablePubsubMessage.java
dhoard: You are correct, `modifyAckDeadline` would only make sense for a `PulledAcknowledgeablePubsubMessage`

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberTemplate.java
```diff
@@ -133,25 +132,33 @@ public SubscriberFactory getSubscriberFactory() {
 	public void ack(Collection<AcknowledgeablePubsubMessage> acknowledgeablePubsubMessages) {
 		Assert.notEmpty(acknowledgeablePubsubMessages, "The acknowledgeablePubsubMessages cannot be null.");
 
-		groupAcknowledgeableMessages(acknowledgeablePubsubMessages).forEach(this::ack);
+		groupPulledAcknowledgeableMessages(acknowledgeablePubsubMessages).forEach(this::ack);
+
+		acknowledgeablePubsubMessages.stream().filter(a -> !(a instanceof PulledAcknowledgeablePubsubMessage))
```
artembilan: This is something what should not be here...
What is the point to mix pulled and not-pulled messages in the same collection for ack'ing?

I wouldn't encourage to collect not-pulled and would just make an API like `public void ack(Collection<PulledAcknowledgeablePubsubMessage> acknowledgeablePubsubMessages)`.

I might miss something though...

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/PulledAcknowledgeablePubsubMessage.java
artembilan: I wouldn't bother myself with those details since it has to be done on the target client or service level, but not here in our API.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/AcknowledgeablePubsubMessage.java
artembilan: I would expose an API like `nack(ModifyAckDeadlineRequest)`. Looks like it is a pretty simple builder and can cover all your requirements.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberTemplate.java
meltsufin: I was trying to avoid exposing yet another class to the users. I would want to keep `PulledAcknowledgeablePubsubMessage` as an internal class. What is the right approach in Spring for internal classes? Put then in `support` package?

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/AcknowledgeablePubsubMessage.java
meltsufin: I don't think it's very user-friendly to expose `ModifyAckDeadlineRequest`. All we need is the seconds.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberTemplate.java
artembilan: Even if you move it into the `support` that doesn't mean I won't be able to instantiate it.
If you really want to hide details from user, you need to make a class as a `package protected`.
Well, a `package protected` is also fine, but you still need somehow to instantiate it and I wouldn't encourage you to play with reflection...

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/AcknowledgeablePubsubMessage.java
artembilan: M-m-m. Not sure what you mean...
Why does then Google do its API not user-friendly?..

This `ModifyAckDeadlineRequest` is really a part of Pub/Sub ack'ing API.
I'm sure there is nothing wrong to use it directly, without any Spring and that is really not a Spring goal to disable all the access to low-level client.
The simple sample is `JdbcTemplate` : it doesn't hide from us a fact of the `DataSource` fact. See `public DataSource getDataSource()`.

Something similar I have in the Spring Integration AWS, where alongside with our Spring API we are good to allow end-user to use existing AWS SDK abstractions.
There is just no reason to re-invent the wheel if existing one can be simply reused.

I don't mind if that not a case here though...

Feel free to do whatever you want and think is reasonable...

meltsufin: I think the low-level API route is using `SubscriberFactory` that is exposed via the `template.getSubscriberFactory().getSubscriberStub()...`.

artembilan: OK. Let's go ahead with the solution and ship it to the field for feedback from target users!

No sense baulk any more! 😄

meltsufin:
@dhoard @artembilan I added the `modifyAckDeadline` method, and also changed the pull operation to return the `PulledAcknowledgeablePubsubMessage`. I was trying to avoid exposing this class, but I guess there isn't much harm in it.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberTemplate.java
```diff
@@ -130,28 +128,38 @@ public SubscriberFactory getSubscriberFactory() {
...
 
-		groupAcknowledgeableMessages(acknowledgeablePubsubMessages).forEach(this::nack);
+		groupPulledAcknowledgeableMessages(pulledAcknowledgeablePubsubMessages).forEach(this::nack);
+	}
+
+	@Override
+	public void modifyAckDeadline(Collection<PulledAcknowledgeablePubsubMessage> pulledAcknowledgeablePubsubMessages,
+			int ackDeadlineSeconds) {
+		Assert.notEmpty(pulledAcknowledgeablePubsubMessages, "The pulledAcknowledgeablePubsubMessages cannot be null.");
+		Assert.isTrue(ackDeadlineSeconds >= 0, "The ackDeadlineSeconds must not be negative.");
```
artembilan: Why don't you allow end-user to specify it as `0`?
Look into your `nack(String subscriptionName, Collection<String> ackIds)`, so, it's pretty fine to call this `modifyAckDeadline()` with `0`...

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/PulledAcknowledgeablePubsubMessage.java
```diff
@@ -0,0 +1,90 @@
...
+	private PubsubMessage message;
+
+	private String ackId;
+
+	private String subscriptionName;
+
+	private SubscriberStub subscriberStub;
+
+	public PulledAcknowledgeablePubsubMessage(PubsubMessage message, String ackId,
```
artembilan: May this should not be `public`? Maybe only template can create instances of this class?...

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberTemplate.java
dhoard: @artembilan I see `ackDeadlineSeconds >= 0` ... am I missing something?

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/PulledAcknowledgeablePubsubMessage.java
```diff
@@ -0,0 +1,90 @@
...
+		this.subscriberStub.acknowledgeCallable().call(acknowledgeRequest);
+	}
+
+	@Override
+	public void nack() {
+		modifyAckDeadline(0);
+	}
+
+	public void modifyAckDeadline(int ackDeadlineSeconds) {
+		ModifyAckDeadlineRequest modifyAckDeadlineRequest = ModifyAckDeadlineRequest.newBuilder()
```
dhoard: Need to check for negative `ackDeadlineSeconds` ...

```
Assert.isTrue(ackDeadlineSeconds >= 0, "The ackDeadlineSeconds must not be negative.");
```

dhoard:
@meltsufin I like this approach.  I think we need to get this into the codebase.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberTemplate.java
artembilan: Doh... Thanks, never mind then: my eyes were blind. 😄

dhoard:
@meltsufin @artembilan PTAL at

https://github.com/dhoard/spring-cloud-gcp/blob/dhoard.acknowledgeable-interface/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/PulledAcknowledgeablePubsubMessage.java

https://github.com/dhoard/spring-cloud-gcp/blob/dhoard.acknowledgeable-interface/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberTemplate.java

Converted `PulledAcknowledgeablePubsubMessage` to an interface and added asynchronous (n)ack and modifyAckDeadline methods.

Implemented an inner class in `PubSubSubscriberTemplate` to implement `PulledAcknowledgeablePubsubMessage`.

Not sure if this approach is better or worse than a class `PulledAcknowledgeablePubsubMessage` with a public constructor.

artembilan:
Do we have some vision for the `PushedAcknowledgablePubsubMessage` already?
Does it make sense now to bother ourselves with some extra interfaces?
Can't we just have that `PulledAcknowledgeablePubsubMessage` as a class without `public` ctor?

I would prefer to see `asyncAck()`/`asyncNack()` instead for clear auto-completion assistant in the IDE and proper distribution of concerns.

meltsufin:
@dhoard RE: RANT
I completely agree as evident by the [issue](https://github.com/GoogleCloudPlatform/google-cloud-java/issues/3479) I submitted to the client lib team.
We should also probably think about how to insulate our users from a potential future when the subscribe side is fixed.

I'll look into the suggestion to hide the constructor for `PulledAcknowledgeablePubsubMessage`.

dhoard:
@artembilan I would be find with `asyncAck()` / `asyncNack()` or even `ackAsync()` / `nackAsync()`.

meltsufin:
I introduced the `BatchAcknowledgeablePubsubMessage` interface and made `PulledAcknowledgeablePubsubMessage` an inner class of the template. This looks better to me and also creates less code duplication. WDYT?

dhoard:
@meltsufin I agree ... this looks clean to me.

For the batch pull use case, it makes it very straightforward ...

```
while (!shutdown) {
    List<BatchAcknowledgeablePubsubMessage> list = pubSubSubscriberTemplate.pull("mySubscription", 1000, false);

    // Process the list ... i.e. batch of messages

    pubSubSubscriberTemplate.ack(list);
}
```

artembilan:
1. I don't see any problem with the `public PulledAcknowledgeablePubsubMessage`, but without `public`  ctor.
2. A new `BatchAcknowledgeablePubsubMessage` does't bring for me too much value and even confusing: its name says about some "batch", but there is really no any batch logic in the message.

Maybe if you name the interface as `PulledAcknowledgeablePubsubMessage` and have a `TemplatePulledAcknowledgeablePubsubMessage` as an inner class, then it would make me happier...
But that's again: my opinion...

meltsufin:
I'm struggling with the name for the interface. The reason I don't like `Pulled..` is because I'm anticipating a day when subscribe side will also have the ack ID available and then this name will be bad.
I've toyed with some potential names:

- `FullyAcknowledgeablePubsubMessage`
- `AcknowledgeablePubsubMessagePlus`
- `ExtendedAcknowledgeablePubsubMessage`

I don't really love any of these alternatives and not completely happy with `BatchAcknowledgeablePubsubMessage` either, but I'm open to other suggestions.

dhoard:
What about ...

`AcknowledgeablePubsubMessage` renamed to `BasicAcknowledgeablePubsubMessage`
`PulledAcknowledgeablePubsubMessage` renamed to `AcknowledgeablePubsubMessage`

or

`AcknowledgeablePubsubMessage` renamed to `BaseAcknowledgeablePubsubMessage`
`PulledAcknowledgeablePubsubMessage` renamed to `AcknowledgeablePubsubMessage`

artembilan:
Can't we call it `PullResultMessage` or even just `PullResult`?
As I said before I don't see reason in the extra interface in between. I don't believe that we are going to have more its implementations or will proxy such a Java bean...

meltsufin:
@artembilan I'm expecting to reuse it on the subscribe side when https://github.com/GoogleCloudPlatform/google-cloud-java/issues/3479 is resolved. So, I really want to avoid the word "Pull" in the name. The interface is useful because of the current inner-class implementation as well.
I don't mind any of the @dhoard's suggestions for naming. Do you have a preference among the suggestions?

artembilan:
@meltsufin  didn't answer me in our private chat, by looks like we are going the route with the `public interface BasicAcknowledgeablePubsubMessage extends AcknowledgeablePubsubMessage`...

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberTemplate.java
```diff
@@ -164,12 +172,62 @@ private void ack(String subscriptionName, Collection<String> ackIds) {
...
 	}
+
+
+	private class PulledAcknowledgeablePubsubMessage implements AcknowledgeablePubsubMessage {
+
+		private PubsubMessage message;
+
+		private String ackId;
+
+		private String subscriptionName;
```
artembilan: All 3 properties must be `final`

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberTemplate.java
```diff
@@ -164,12 +172,62 @@ private void ack(String subscriptionName, Collection<String> ackIds) {
...
+				.setAckDeadlineSeconds(ackDeadlineSeconds)
 				.addAllAckIds(ackIds)
 				.setSubscription(subscriptionName)
 				.build();
 
 		this.subscriberStub.modifyAckDeadlineCallable().call(modifyAckDeadlineRequest);
 	}
+
+
+	private class PulledAcknowledgeablePubsubMessage implements AcknowledgeablePubsubMessage {
```
artembilan: I would like to see a `toString()` on this class - pretty useful during application debugging

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/AcknowledgeablePubsubMessage.java
```diff
@@ -16,67 +16,19 @@
...
-	}
+public interface AcknowledgeablePubsubMessage extends BasicAcknowledgeablePubsubMessage {
 
-	public void nack() {
-		ModifyAckDeadlineRequest modifyAckDeadlineRequest = ModifyAckDeadlineRequest.newBuilder()
-				.setAckDeadlineSeconds(0)
-				.addAckIds(this.ackId)
-				.setSubscription(this.subscriptionName)
-				.build();
+	String getAckId();
```
artembilan: Would be great that all the public API has Javadocs.

Thanks

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberTemplate.java
ChengyuanZhao: I see that this isn't a `private static class` because of the `PubSubSubscriberTemplate.this`. Just curious , how come you  rather made a private class than passing the PubSubSubscriberTemplate in the constructor and having this be a separate class? If a whole PubSub template is too much functionality, perhaps you could just provide two `Consumer`s that do the ack and the modify ack deadline. It just felt a little strange to see an implementation of a message here in the template.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberTemplate.java
```diff
@@ -164,12 +172,70 @@ private void ack(String subscriptionName, Collection<String> ackIds) {
...
+			this.ackId = ackId;
+			this.subscriptionName = subscriptionName;
+		}
+
+		@Override
+		public PubsubMessage getPubsubMessage() {
+			return this.message;
+		}
+
+		public String getAckId() {
```
ChengyuanZhao: This one and the method below should be annotated `@Override` right?

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberTemplate.java
artembilan: When you move it into the top-level class, you get a class tangle which we would like to avoid.
There is nothing wrong having this class as an inner one as long as we have an `interface`  as a contract.

However I agree with the `Consumer` approach...

meltsufin: Having `PulledAcknowledgeablePubsubMessage` as an inner class works out well here in my opinion. We're coding to an interface, and the whole `PulledAcknowledgeablePubsubMessage` is just an implementation detail completely invisible to the users.

Regarding the `Consumer` approach, I don't fully understand the vision with that on the pull side. Can you elaborate?

ChengyuanZhao: Yes, of course: 

So it seems like this is a `private class` rather than a `private static class` because you would like to access the specific PubSubSubscriberTemplate instance that contains this inner class. However, you only need this PubSubSubscriberTemplate instance to do 2 very specific things, and you don't actually need the full functionality of it. So , the idea was to instead make this a private static class and pass those 2 specific operations you need in as two `Consumer`s. 

I'm just elaborating on the idea here, not saying it's blocking or anything. I'm fine with it as it is, too.

dhoard:
LGTM

PR #895, opened 2018-07-26
Spanner allow missing columns in STRUCTs

### LINKED ISSUE #894: Spanner cloud framework not allowing nested struct queries if not all columns are queried
We are trying to leverage allowmissingcolumns=true on org.springframework.cloud.gcp.data.spanner.core.SpannerTemplate.query(Class<T>, String, List<String>, Object[], SpannerQueryOptions)

Query used : select p.name as program_name, ARRAY(select as struct t.tier_id as tier_id, t.created_by as tier_created_by from tier t where t.PROGRAM_ID = p.PROGRAM_ID) as tiers from PROGRAM p

When it retireves parent entity there was no exception of a missing column but when it executed inner query to retrieve tiers in the below example we are seeing an exception 

```

org.springframework.cloud.gcp.data.spanner.core.mapping.SpannerDataException: Unable to read column from Spanner results: PROGRAM_ID
	at org.springframework.cloud.gcp.data.spanner.core.convert.ConverterAwareMappingSpannerEntityReader.isMissingColumn(ConverterAwareMappingSpannerEntityReader.java:129)
	at org.springframework.cloud.gcp.data.spanner.core.convert.ConverterAwareMappingSpannerEntityReader.shouldSkipProperty(ConverterAwareMappingSpannerEntityReader.java:120)
	at org.springframework.cloud.gcp.data.spanner.core.convert.ConverterAwareMappingSpannerEntityReader.lambda$read$0(ConverterAwareMappingSpannerEntityReader.java:94)
	at org.springframework.data.mapping.model.BasicPersistentEntity.doWithProperties(BasicPersistentEntity.java:329)
	at org.springframework.cloud.gcp.data.spanner.core.convert.ConverterAwareMappingSpannerEntityReader.read(ConverterAwareMappingSpannerEntityReader.java:92)
	at org.springframework.cloud.gcp.data.spanner.core.convert.SpannerEntityReader.read(SpannerEntityReader.java:35)
	at org.springframework.cloud.gcp.data.spanner.core.convert.StructPropertyValueProvider.convertOrRead(StructPropertyValueProvider.java:85)
	at org.springframework.cloud.gcp.data.spanner.core.convert.StructPropertyValueProvider.lambda$convertOrReadIterable$0(StructPropertyValueProvider.java:101)
	at java.util.ArrayList.forEach(Unknown Source)
	at java.util.Collections$UnmodifiableCollection.forEach(Unknown Source)
	at org.springframework.cloud.gcp.data.spanner.core.convert.StructPropertyValueProvider.convertOrReadIterable(StructPropertyValueProvider.java:101)

```

Please see the attached stackTrace where when its doing a fetch of inner query, entityReader is sending allowmissingcolumns as false. 

stacktrace

ChengyuanZhao:
fixes #894

This passes down the allow-missing-column setting so that nested STRUCTs can also allow missing parts.

### DIFF
--- a/spring-cloud-gcp-docs/src/main/asciidoc/spanner.adoc
+++ b/spring-cloud-gcp-docs/src/main/asciidoc/spanner.adoc
@@ -640,6 +640,7 @@ Sort.by(Order.desc("action").ignoreCase())
 ===== Partial read
 
 Partial read is only possible when using Queries. In case the rows returned by query have fewer columns than the entity that it will be mapped to, Spring Data will map the returned columns and leave the rest as they of the columns are.
+This setting also applies to nested structs and their corresponding nested POJO properties.
 
 [source,java]
 ----
--- a/spring-cloud-gcp-docs/src/main/asciidoc/spanner.adoc
+++ b/spring-cloud-gcp-docs/src/main/asciidoc/spanner.adoc
@@ -639,7 +639,7 @@ Sort.by(Order.desc("action").ignoreCase())
 
 ===== Partial read
 
-Partial read is only possible when using Queries. In case the rows returned by query have fewer columns than the entity that it will be mapped to, Spring Data will map the returned columns and leave the rest as they of the columns are.
+Partial read is only possible when using Queries. In case the rows returned by the query have fewer columns than the entity that it will be mapped to, Spring Data will map the returned columns only.
 This setting also applies to nested structs and their corresponding nested POJO properties.
 
 [source,java]
@@ -648,6 +648,7 @@ List<Trade> trades = this.spannerTemplate.query(Trade.class, Statement.of("SELEC
     new SpannerQueryOptions().setAllowMissingResultSetColumns(true));
 ----
 
+If the setting is set to `false`, then an exception will be thrown if there are missing columns in the query result.
 
 ===== Summary of options for Query vs Read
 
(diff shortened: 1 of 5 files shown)


ChengyuanZhao:
@meltsufin Added explicit clarification. Based on the original motivating feedback, I thought this change was making the behavior more intuitive or closer to what a user would assume, so did not think explicit explanation was needed.

ChengyuanZhao:
@meltsufin as a quick summary:

StructPropertyValueProvider.java is the main class we use to read columns from rows / structs. (the original row and the nested structs are the same type).  You see I added passing that allowMissingColumns param down to this reader class because it recursively reads nested structs. Originally this value was defaulted to `false`.

spring-cloud-gcp-docs/src/main/asciidoc/spanner.adoc
```diff
@@ -640,6 +640,7 @@ Sort.by(Order.desc("action").ignoreCase())
 ===== Partial read
 
 Partial read is only possible when using Queries. In case the rows returned by query have fewer columns than the entity that it will be mapped to, Spring Data will map the returned columns and leave the rest as they of the columns are.
+This setting also applies to nested structs and their corresponding nested POJO properties.
```
meltsufin: The line above contains "rest as they of the columns are" which needs to be fixed.

PR #897, opened 2018-07-26
Datastore Simple Repo CRUD operations

### LINKED ISSUE #681: Datastore CRUD repository implementation

ChengyuanZhao:
fixes #681

Adds the CRUD repo functions and unit tests. 

separate  issue and PR for the paging and sorting functionality.


spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreOperations.java
```diff
@@ -45,6 +45,13 @@
 	 */
 	<T> T save(T instance);
 
+	/**
+	 * Saves multiple instances of objects to Cloud Datastore. Behaves as update or insert.
+	 * @param entities the objects to save.
+	 * @return the entities that were saved.
+	 */
+	<T> Iterable<T> saveAll(Iterable<T> entities);
```
ChengyuanZhao: I added this and also deleteAll to implement the CRUD repo deleteAll and saveAll functions. I wanted to take advantage of the client lib's version that can handle multiple entities, so I created these new functions.

spring-cloud-gcp-data-datastore/src/test/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplateTests.java
```diff
@@ -134,24 +132,48 @@ public void saveTest() {
...
+		when(this.objectToKeyFactory.getKeyFromObject(same(object2), any()))
+				.thenReturn(key2);
+
+		when(this.datastore.put(any(), any()))
+				.thenReturn(ImmutableList.of(entity1, entity2));
+
+		this.datastoreTemplate.saveAll(ImmutableList.of(object1, object2));
+		verify(this.datastore, times(1)).put(eq(entity1), eq(entity2));
+		verify(this.datastoreEntityConverter, times(1)).write(same(object1), notNull());
+		verify(this.datastoreEntityConverter, times(1)).write(same(object2), notNull());
```
dmitry-s: should we verify that ids were allocated?

ChengyuanZhao: That functionality is in the `objectToKeyFactory`, so instead I have a separate unit test class for those that does check that the allocation happens and it is set properly. Here I just am mocking the allocate calls.

dmitry-s: Oh I see. Maybe rename the test then, because the name suggest that it tests allocate id.

PR #906, opened 2018-07-27
Using Spring Data common annotations

### LINKED ISSUE #902: Remove @Ancestors because Key already holds that info
For now, we're requiring the use of `Key` for the ID type, if using ancestors/parents.

### LINKED ISSUE #904: Use org.springframework.data.annotation.* wherever possible
This means replacing `@NotMapped` with `@Transient`, and `@Embedded` with `@Reference`.

ChengyuanZhao:
fixes #902 fixes #904


spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/mapping/DatastorePersistentEntityImpl.java
```diff
@@ -72,14 +72,6 @@ protected boolean hasTableName() {
 		return this.kind != null && StringUtils.hasText(this.kind.name());
 	}
 
-	@Override
-	public void addPersistentProperty(DatastorePersistentProperty property) {
-		if (!property.isMapped()) {
```
ChengyuanZhao: It turns out that since `@Transient` already exists, this part is already done by default.

PR #908, opened 2018-07-27
Datastore repository factory and config settings

### LINKED ISSUE #898: Datastore repository instantiation classes
These are the classes that are somewhat boilerplate implementations of Spring Data interfaces that are involved in creating instances of repositories for the user's types and methods.

ChengyuanZhao:
fixes #898

These are boilerplate type classes that tell Spring the names of the repository, template, and some annotation classes.

### DIFF
--- a/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/config/EnableDatastoreRepositories.java
+++ b/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/config/EnableDatastoreRepositories.java
@@ -1,0 +1,137 @@
+/*
+ *  Copyright 2018 original author or authors.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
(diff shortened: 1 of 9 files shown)


spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/config/EnableDatastoreRepositories.java
```diff
@@ -0,0 +1,137 @@
...
+ * @author Chengyuan Zhao
+ *
+ * @since 1.1
+ */
+@Target(ElementType.TYPE)
+@Retention(RetentionPolicy.RUNTIME)
+@Inherited
+@Documented
+@Import(DatastoreRepositoriesRegistrar.class)
+public @interface EnableDatastoreRepositories {
```
ChengyuanZhao: From the past we found out that all of these many default values are needed or instantiation of repositories will fail.

meltsufin:
Can we go over this one in a meeting?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/config/EnableDatastoreRepositories.java
meltsufin: Can we extend `org.springframework.data.repository.config.EnableRepositories`?

ChengyuanZhao: @meltsufin I'm having trouble finding this file, but I did find this: 

https://github.com/spring-projects/spring-data-commons/blob/master/src/test/java/org/springframework/data/repository/config/EnableRepositories.java

Was this the one? This one seems to just be a test file used with the test "DummyRegistrar"

meltsufin: Oops :)

Issue #853, opened 2018-07-11
Spanner eager Parent-Child (1-to-many) relationships


ChengyuanZhao:
We recently discussed relationships in spring data with Oliver, and arrived at some common points among spring data integrations:

- eager retrieval of children should be the default implementation, and it is OK to let the user know the limitations of it. this is the recommendation for single children or multiple children in a collection.
- lazy retrieval can be very complicated especially after involving other features like transactions, and should definitely come second if at all.
- the idea of aggregates (parents) is strong in spring data, and relationships should be reflected in POJOs one-way from aggregate to its sub-items (children).
- recommend users use identifier properties instead of POJO references , which forces the user to do retrieval themselves.

ChengyuanZhao:
Main points

- if property is a type (or inner type if iterable) annotated with Table(name = ...)  and it's not annotated embedded, then it is a child POJO relationship

Reads
- single-item children will be will be immediately recursively retrieved
- collection children will be retrieved on first touch (to prevent exponentially increasing number of retrievals). The subsequent "lazy" retrieval is done using the same SpannerOperations, and we already have functionality to create single-transaction-SpannerOperations (so even though it's "lazy" the data is exactly the same as immediate retrieval) 

Writes
- client lib allows batch write of mutations, so a pojo and all of its children can be saved in the same request
- single-item children will be saved 
- collection children that have been touched will be saved.

nov1vinni:
It would be good to have annotation defined for onjetomany/manytoone , we can specify if this needs to be eagerly/lazy loaded something like below . We want to define this on spanner persistent entity to save parent/child records at one shot and also to retrieve them at one shot.

It would be good to expose a detailed doc explaining features. 

@onetomany("getChild",LoadType.Eager) 
or
@onetomany
@Eager

ChengyuanZhao:
Tasks, in order:

- [ ] create admin client methods for checking table schema of given child tables if they are interleaved 
- [ ] create `@OneToMany` annotation to mark child collections
- [ ] read method recursively eagerly gets children
- [ ] write method creates single batch write for all children recursively 
- [ ] ref doc updates

vineela-gampa:
@ChengyuanZhao 
1) Does this parent/child or @onetoMany relationship defined on an entity works only if the tables are INTERLEAVED?
2) is there a plan to integrate more queryOptions for filtering/Grouping results. Currently support exists for sort/pagination.

vineela-gampa:
[...]

PR #910, opened 2018-07-28
Spanner schema admin functions for parent-child relationships

ChengyuanZhao:
related #853


spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/mapping/SpannerPersistentEntityImpl.java
```diff
@@ -208,6 +209,12 @@ private void verifyEmbeddedColumnNameOverlap(Set<String> seen,
 		spannerPersistentEntity.doWithProperties(
 				(PropertyHandler<SpannerPersistentProperty>) spannerPersistentProperty -> {
 					if (spannerPersistentProperty.isEmbedded()) {
+						if (ConversionUtils.isIterableNonByteArrayType(
+								spannerPersistentProperty.getType())) {
+							throw new SpannerDataException(
+									"Embedded properties cannot be collections: "
```
meltsufin: Is it a limitation of Spanner, or is it just to disambiguate with one-to-many relationship?

ChengyuanZhao: I would say this is a way to disambiguate. There's nothing preventing a spanner parent row from having just 1 child row interleaved, but there is also no mechanism to enforce this 1 limit in Spanner.  As a result, I think there's much less opportunity for confusion and debugging work if we reflect that in our lib too: that it's inherently a list and there is no mechanism to enforce it to be 1 item always.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/admin/SpannerSchemaUtils.java
```diff
@@ -156,6 +115,98 @@ public Key getKey(Object object) {
...
+			throw new SpannerDataException(
+					"Could not find suitable Cloud Spanner column type for property "
+							+ "type :" + columnType);
+		}
+
+		return columnName + getTypeDdlString(spannerColumnType, spannerJavaType.isArray(),
+					spannerPersistentProperty.getMaxColumnLength());
+	}
+
+	private void getCreateTableDdlStringsForHierarchy(String parentTable,
```
ChengyuanZhao: More explicitly name to say "interleaved"

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/admin/SpannerSchemaUtils.java
```diff
@@ -156,6 +115,98 @@ public Key getKey(Object object) {
...
+			Class entityClass, List<String> ddlStrings, Set<Class> seenClasses) {
+		if (seenClasses.contains(entityClass)) {
+			return;
+		}
+		ddlStrings.add(getCreateTableDdlString(entityClass) + (parentTable == null ? ""
+				: ", INTERLEAVE IN PARENT " + parentTable + " ON DELETE CASCADE"));
```
meltsufin: Should `ON DELETE CASCADE` an option through a property?

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/admin/SpannerSchemaUtils.java
```diff
@@ -156,6 +115,98 @@ public Key getKey(Object object) {
...
+		SpannerPersistentEntity spannerPersistentEntity = this.mappingContext
+				.getPersistentEntity(entityClass);
+		spannerPersistentEntity.doWithChildCollectionProperties(
+				(PropertyHandler<SpannerPersistentProperty>) spannerPersistentProperty ->
+						getCreateTableDdlStringsForHierarchy(
+						spannerPersistentEntity.tableName(),
+						spannerPersistentProperty.getColumnInnerType(), ddlStrings,
+						seenClasses));
+		seenClasses.add(entityClass);
```
ChengyuanZhao: need to add this to seen before recursion call

ChengyuanZhao: add test case where two properties have the same inner type to catch this

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/admin/SpannerSchemaUtils.java
```diff
@@ -189,13 +240,31 @@ public Key getKey(Object object) {
...
 
-	/**
-	 * Gets the DDL string to drop the table for the given entity in Spanner.
-	 * @param entityClass the entity type.
-	 * @return the DDL string.
-	 */
-	public String getDropTableDdlString(Class entityClass) {
-		return "DROP TABLE "
-				+ this.mappingContext.getPersistentEntity(entityClass).tableName();
+	private void getDropTableDdlStringsForHierarchy(Class entityClass,
```
ChengyuanZhao: drop DDL and create DDL is very similar, should refactor

spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/core/admin/SpannerSchemaUtilsTests.java
```diff
@@ -186,8 +221,60 @@ public void getIdTest() {
...
+		@Embedded
+		EmbeddedColumns embeddedColumns;
+
+		@Column(name = "custom_col")
+		String something;
+
+		@Column(name = "")
+		String other;
+
+		@OneToMany
```
ChengyuanZhao: Perhaps `@Interleaved`

spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/core/admin/SpannerSchemaUtilsTests.java
```diff
@@ -155,6 +157,39 @@ public void getIdTest() {
 				this.spannerSchemaUtils.getKey(t));
 	}
 
+	@Test
+	public void getCreateDdlHierarchyTest() {
+		List<String> createStrings = this.spannerSchemaUtils
+				.getCreateTableDdlStringsForHierarchy(ParentEntity.class);
+		assertEquals(3, createStrings.size());
```
ChengyuanZhao: assert equals on array of all 3 strings?

ChengyuanZhao:
@meltsufin @dmitry-s changes made from in-person discussion, PTAL

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/spanner/GcpSpannerAutoConfiguration.java
```diff
@@ -86,6 +86,8 @@
 
 		private final int keepAliveIntervalMinutes;
 
+		private final boolean createTableDdlCascadeOnDelete;
```
meltsufin: This applies only to interleaved tables, correct?

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/spanner/GcpSpannerProperties.java
```diff
@@ -40,6 +40,10 @@
 
 	private String database;
 
+	// if True then create-table statements generated will cascade on delete. no-action on delete
```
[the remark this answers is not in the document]
ChengyuanZhao: technically (grammatically) these aren't sentences , but will change.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/admin/SpannerSchemaUtils.java
```diff
@@ -47,76 +52,60 @@
...
+
+	/**
+	 * Constructor. Generates create-table DDL statements that cascade deletes for
+	 * interleaved tables.
+	 * @param mappingContext a mapping context used to obtain persistent entity metadata
+	 * for generating DDL statements.
+	 * @param spannerEntityProcessor an entity processor that is queried for types that it
+	 * can convert for determining compatible column types when generating DDL statements.
+	 */
 	public SpannerSchemaUtils(SpannerMappingContext mappingContext,
```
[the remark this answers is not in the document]
ChengyuanZhao: Removed. Made requested changes.

@dmitry-s

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/admin/SpannerSchemaUtils.java
```diff
@@ -47,76 +52,47 @@
 
 	private final SpannerTypeMapper spannerTypeMapper;
 
+	private final boolean createTableDdlDeleteOnCascade;
```
meltsufin: This is still different from `createTableStatementsCascadeOnDelete`. Also, I think a better name would actually indicate that this is for interleaved tables. Maybe `createInterleavedTableCascadeOnDelete`?

ChengyuanZhao: Yes I can change to that new name everywhere. (originally I felt it was OK to have "Ddl" in this class because so many things in this class have those letters in the name, and it would have more meaning given the context of this class and all the "Ddl" related functions.)

meltsufin: `createTableStatementsCascadeOnDelete` to use `DDL` in the name.
Actually, I'm leaning towards this name now:
`createInterleavedTableDdlOnDeleteCascade`

meltsufin:
@ChengyuanZhao I think you need to resolve conflicts before merging.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/admin/SpannerSchemaUtils.java
ChengyuanZhao: Renamed in all 3 places to `createInterleavedTableDdlOnDeleteCascade` and merged in other PRs and resolved conflicts. 

I'll update the ref doc PR for this new name, too.

PR #911, opened 2018-07-29
Spanner parent-child eager write save

ChengyuanZhao:
related #853 

Mutations for parents and children are written in a single request

### DIFF
--- a/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerMutationFactory.java
+++ b/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerMutationFactory.java
@@ -16,6 +16,7 @@
 
 package org.springframework.cloud.gcp.data.spanner.core;
 
+import java.util.List;
 import java.util.Optional;
 import java.util.Set;
 
@@ -38,7 +39,7 @@ public interface SpannerMutationFactory {
 	 * @param object The object to store.
 	 * @return The mutation operation which will store the object.
 	 */
-	Mutation insert(Object object);
+	List<Mutation> insert(Object object);
 
 	/**
 	 * Updates or inserts a single object in Spanner. The columns' values corresponding to
@@ -48,7 +49,7 @@ public interface SpannerMutationFactory {
 	 * Optional, then all columns are used.
 	 * @return The mutation operation to perform the action.
 	 */
-	Mutation upsert(Object object, Optional<Set<String>> includeColumns);
+	List<Mutation> upsert(Object object, Optional<Set<String>> includeColumns);
 
 	/**
 	 * Updates a single object in Spanner. The columns' values corresponding to the
@@ -58,7 +59,7 @@ public interface SpannerMutationFactory {
 	 * Optional, then all columns are used.
 	 * @return The mutation operation to perform the action.
 	 */
-	Mutation update(Object object, Optional<Set<String>> includeColumns);
+	List<Mutation> update(Object object, Optional<Set<String>> includeColumns);
 
 	/**
 	 * Deletes several objects from Spanner.
--- a/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerMutationFactory.java
+++ b/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerMutationFactory.java
@@ -35,29 +35,36 @@ import com.google.cloud.spanner.Mutation;
 public interface SpannerMutationFactory {
 
 	/**
-	 * Stores a single object in Spanner.
+	 * Stores an object and its interleaved child properties in Cloud Spanner.
+	 * There will be 1 mutation for each entity corresponding to a row in Cloud Spanner. If there
+	 * are no interleaved children then the returned list will contain the single mutation for the
+	 * given object.
 	 * @param object The object to store.
-	 * @return The mutation operation which will store the object.
+	 * @return The mutation operations which will store the object and its children.
 	 */
 	List<Mutation> insert(Object object);
 
 	/**
-	 * Updates or inserts a single object in Spanner. The columns' values corresponding to
-	 * the object's fields are treated according to Mutation.Op.INSERT_OR_UPDATE.
+	 * Updates or inserts an object and its interleaved child properties in Cloud Spanner.
+	 * There will be 1 mutation for each entity corresponding to a row in Cloud Spanner. If there
+	 * are no interleaved children then the returned list will contain the single mutation for the
+	 * given object.
 	 * @param object The object to update or newly insert.
 	 * @param includeColumns The columns to use in the operation. if null or empty
 	 * Optional, then all columns are used.
-	 * @return The mutation operation to perform the action.
+	 * @return The mutation operations which will store the object and its children.
 	 */
 	List<Mutation> upsert(Object object, Optional<Set<String>> includeColumns);
 
 	/**
-	 * Updates a single object in Spanner. The columns' values corresponding to the
-	 * object's fields are treated according to Mutation.Op.UPDATE.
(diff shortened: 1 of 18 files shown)


spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerMutationFactoryImpl.java
```diff
@@ -130,13 +132,37 @@ public Mutation delete(Class entityClass, Key key) {
...
+		PersistentPropertyAccessor parentAccessor = parentEntity
+				.getPropertyAccessor(parentObject);
+		PersistentPropertyAccessor childAccessor = childEntity
+				.getPropertyAccessor(childObject);
+		for (int i = 0; i < parentPK.length; i++) {
+			if (!parentAccessor.getProperty(parentPK[i])
+					.equals(childAccessor.getProperty(childPK[i]))) {
+				throw new SpannerDataException(
+						"A child entity's common primary key columns with its parent must "
+								+ "have the same values. Primary key component " + i
```
meltsufin: Can we include the property name in the exception?

ChengyuanZhao: Yes, added the column name to the message.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerMutationFactoryImpl.java
```diff
@@ -114,13 +118,49 @@ public Mutation delete(Class entityClass, Key key) {
...
+							this.spannerMappingContext.getPersistentEntity(
+									spannerPersistentProperty.getColumnInnerType()), child);
+					mutations.addAll(saveObject(op, child, includeColumns));
+				}
+			}
+		});
+		return mutations;
+	}
+
+	private void verifyChildHasParentId(SpannerPersistentEntity parentEntity,
```
dmitry-s: this change is a part of #912, should it be here as well?

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/mapping/SpannerPersistentEntityImpl.java
```diff
@@ -173,6 +173,17 @@ public void doWithChildCollectionProperties(
 				});
 	}
 
+	@Override
+	public void doWithColumnBackedProperties(
```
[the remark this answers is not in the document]
ChengyuanZhao: This only exists in both because it's not in master branch yet. I should have included this function in an earlier PR.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerMutationFactoryImpl.java
ChengyuanZhao: This one is different.

This one ensures that the values of the PKs are the same, so that one child doesn't get saved into some other parent. 

The other one is ensuring that the child POJO has the necessary properties.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
```diff
@@ -241,22 +243,29 @@ public void upsert(Object object, Optional<Set<String>> includeColumns) {
 
 	@Override
 	public void delete(Object entity) {
-		applyMutationUsingEntity(this.mutationFactory::delete, entity);
+		applyMutationUsingEntity(
+				x -> Collections.singletonList(this.mutationFactory.delete(x)), entity);
```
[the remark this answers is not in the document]
ChengyuanZhao: Yes good idea, will do!

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerMutationFactory.java
```diff
@@ -48,7 +49,7 @@
 	 * Optional, then all columns are used.
 	 * @return The mutation operation to perform the action.
 	 */
-	Mutation upsert(Object object, Optional<Set<String>> includeColumns);
+	List<Mutation> upsert(Object object, Optional<Set<String>> includeColumns);
 
 	/**
 	 * Updates a single object in Spanner. The columns' values corresponding to the
```
ChengyuanZhao: Update javadoc to say it's for all children, too. and mention for objects without interleaved it's just 1

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerMutationFactoryImpl.java
```diff
@@ -114,13 +118,49 @@ public Mutation delete(Class entityClass, Key key) {
...
 				.getPersistentEntity(object.getClass());
+		List<Mutation> mutations = new ArrayList<>();
 		Mutation.WriteBuilder writeBuilder = writeBuilder(op,
 				persistentEntity.tableName());
 		this.spannerEntityProcessor.write(object, writeBuilder::set, includeColumns);
-		return writeBuilder.build();
+		mutations.add(writeBuilder.build());
+
+		persistentEntity.doWithChildCollectionProperties(spannerPersistentProperty -> {
+			Iterable kids = (Iterable) persistentEntity.getPropertyAccessor(object)
```
ChengyuanZhao: rename children

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerMutationFactoryImpl.java
```diff
@@ -114,13 +118,49 @@ public Mutation delete(Class entityClass, Key key) {
...
+				}
+			}
+		});
+		return mutations;
+	}
+
+	private void verifyChildHasParentId(SpannerPersistentEntity parentEntity,
+			Object parentObject, SpannerPersistentEntity childEntity,
+			Object childObject) {
+		SpannerPersistentProperty[] parentPk = parentEntity.getPrimaryKeyProperties();
```
ChengyuanZhao: see if possible to zip with streams for smaller function

spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/core/SpannerMutationFactoryImplTests.java
```diff
@@ -59,23 +62,85 @@ public void setUp() {
...
+		parentMutation = mutations.get(0);
+		Mutation childMutation = mutations.get(1);
+		assertEquals("custom_test_table", parentMutation.getTable());
+		assertEquals(Op.UPDATE, parentMutation.getOperation());
+		assertEquals("child_test_table", childMutation.getTable());
+		assertEquals(Op.UPDATE, childMutation.getOperation());
 	}
 
 	@Test
 	public void upsertTest() {
```
ChengyuanZhao: parameterize the insert or update enum value?

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerMutationFactoryImpl.java
ChengyuanZhao: @meltsufin I looked into alternatives and short of bringing in third party libs like Guava , there does not seem to be zip functionality for plain old Java streams.

PR #912, opened 2018-07-29
Spanner verify child schema includes parent PKs

ChengyuanZhao:
related #853 

During persistent entity creation, check that the child entity includes the PK columns of the parent in the correct order.


spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/mapping/SpannerPersistentEntityImpl.java
```diff
@@ -182,9 +193,39 @@ public boolean hasIdProperty() {
...
+								+ "parent in the same order starting the first "
+								+ "column with additional" + " key columns after.");
+			}
+			for (int i = 0; i < primaryKeyProperties.length; i++) {
+				SpannerPersistentProperty parentKey = primaryKeyProperties[i];
+				SpannerPersistentProperty childKey = childKeyProperties[i];
+				if (!parentKey.getColumnName().equals(childKey.getColumnName())
+						|| !parentKey.getType().equals(childKey.getType())) {
+					throw new SpannerDataException(
+							"The child primary key column at position " + (i + 1)
```
meltsufin: Why not report zero-based index? You can also name the property and its type for both parent and child to aid debugging.

meltsufin:
@ChengyuanZhao Why are we requiring the child entities to map and repeat the keys of the parent? Can we just copy the parent components of the key to the child table when needed? This duplication of keys in the children seems to be causing a lot of pain in keeping them consistent.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/mapping/SpannerPersistentEntityImpl.java
ChengyuanZhao: The annotations where the user specify the key order starts with 1 and increases from there, so I wanted to show this number that the user input in the exception. I also wanted to say the number instead of the names of the 2 properties because I felt it was more concise and exact to say that what we detected was wrong was based on the position, and not that the names mismatched. This sounds like a semantic difference, but I think it is less confusing and doesn't imply to the user that there could be a typo or two real columns but just ordered wrong.

ChengyuanZhao:
@meltsufin There must not be any asymmetry in how the Child POJO and the Parent POJO are treated by the library and in general. In Spanner and here the POJO must hold all columns as properties. We have the embedded columns key  object feature to make this more convenient. However, if the user wanted to use the child as the root of the further family tree then we must not assume that this child has a further parent. 

This check isn't strictly necessary, because the all calls will ultimately fail at spanner if these things don't match, but I thought it would be good to notify the user early. I think this is an appropriate place to do this check. Spring Data gives us this verify() function to override, and this is also we have checked for PK key columns order and such  in the past, and that looked good to the Spring Data folks back then too.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/mapping/SpannerPersistentEntityImpl.java
meltsufin: Ok on the 1-based index, but what's the wrong with including both, the position and the name/type that was found to be mismatched in the exception? As a developer, I would want as much information in the exception as possible to aid debugging.

ChengyuanZhao: gotcha, will add.

spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/core/mapping/SpannerPersistentEntityImplTests.java
```diff
@@ -214,32 +215,55 @@ public void doWithChildrenCollectionsTest() {
...
 		@PrimaryKey
 		String id;
+
+		@PrimaryKey(keyOrder = 2)
+		String id2;
+	}
+
+	private static class ParentInRelationshipMismatchedKeyName {
+		@PrimaryKey
+		String idNameDifferentThanKids;
```
meltsufin: s/Kids/Children

ChengyuanZhao: change made; would you mind re-approving

PR #916, opened 2018-07-30
Spanner eager read parent-child relationships

### LINKED ISSUE #926: Do not use `SELECT *` in Spring Data Spanner
Instead only select the columns that are mapped.

ChengyuanZhao:
related #853 
fixes #926 

Eagerly retrieves child collections. 

The child retrieval uses the parent's PK columns. PKs in spanner are automatically indexed( https://cloud.google.com/spanner/docs/secondary-indexes ) The parent PKs appear in the WHERE  clause so that the index can be utilized. The child rows with the given parent PK values are also stored together physically.

### DIFF
--- a/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryExecutor.java
+++ b/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryExecutor.java
@@ -80,13 +80,15 @@ public class SpannerStatementQueryExecutor {
 	 * of all child rows having the parent's key values is efficient.
 	 * @param parentPersistentEntity the persistent entity of the parent table.
 	 * @param parentObject the parent object whose children to get.
-	 * @param childTable the name of the child table.
+	 * @param childPersistentEntity the persistent entity of the child table.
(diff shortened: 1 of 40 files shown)


spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryExecutor.java
```diff
@@ -73,6 +74,37 @@
...
+	 * is intended to be used with parent-child interleaved tables, so that the retrieval
+	 * of all child rows having the parent's key values is efficient.
+	 * @param parentPersistentEntity the persistent entity of the parent table.
+	 * @param parentObject the parent object whose children to get.
+	 * @param childTable the name of the child table.
+	 * @return the Spanner statement to perform the retrieval.
+	 */
+	public static Statement getChildrenRowsQuery(
+			SpannerPersistentEntity parentPersistentEntity, Object parentObject,
+			String childTable) {
```
ChengyuanZhao: Add child persistent entity to mention only those columns from that metadata in this select statement

ChengyuanZhao:
@meltsufin @dmitry-s I removed the select * from everywhere (actually was easier than i thought). PTAL

ChengyuanZhao:
@meltsufin PTAL.

PR #923, opened 2018-08-01
Overload subscribe() to allow use of BasicAcknowledgeablePubsubMessage

meltsufin:


meltsufin:
#841 will expand on this by providing `subscribeAndConvert()` with a `Consumer<>`.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberOperations.java
```diff
@@ -47,6 +49,16 @@
...
 
+	/**
+	 * Add a callback method to an existing subscription.
+	 * <p>The created {@link Subscriber} is returned so it can be stopped.
+	 * @param subscription the name of an existing subscription
+	 * @param messageConsumer the callback method triggered when new messages arrive
+	 * @return subscriber listening to new messages
+	 * @since 1.1
+	 */
+	Subscriber subscribe(String subscription, Consumer<BasicAcknowledgeablePubsubMessage> messageConsumer);
```
artembilan: I guess we need an implementation for this method.
Why then do we need a `MessageReceiver`-based API then?
Aren't you going to hide such a detail behind our own API?

meltsufin: The implementation is coming. I just wanted to get some preliminary feedback on the idea.
The two reasons I'm thinking of keeping the `MessageReceiver` API are (1) it's already in `RC1` and people are using it, and (2) it's the low-level API that Pub/Sub provides.
(1) is my main concern because I don't think we can have a breaking change like removing that method in `1.1`. WDYT?

artembilan: We can deprecate it instead, although I'm fine to keep it for time being.

The API idea is OK I believe...

dhoard: Leave the method and deprecate it would be developer friendly.

meltsufin: Sounds good.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplate.java
```diff
@@ -132,10 +134,16 @@ public PubSubTemplate setMessageConverter(PubSubMessageConverter messageConverte
 	}
 
 	@Override
+	@Deprecated
```
artembilan: I believe you still need a `@SupressWarning("deprecation")` since you call a deprecated `pubSubSubscriberTemplate.subscribe()`

meltsufin: The  compiler doesn't seem to be complaining.

artembilan: OK. We'll see.
That's why I don't like Maven: too noise during build - we may just miss some simple warning about deprecation violation...

meltsufin:
@artembilan @ChengyuanZhao Can you do a final pass on this one pls?

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberOperations.java
```diff
@@ -38,14 +40,21 @@
  */
 public interface PubSubSubscriberOperations {
 
+	/**
+	 * @deprecated As of 1.1, use {@link #subscribe(String, Consumer)} instead.
```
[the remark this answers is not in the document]
meltsufin: Where else?

meltsufin:
@artembilan I checked the `pubsub.adoc` but it only mentions using the `subscribe` method without referring to `MessageReceiver`. So, I think we're good on that front.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberTemplate.java
```diff
@@ -68,13 +71,29 @@ public PubSubSubscriberTemplate(SubscriberFactory subscriberFactory) {
...
 		Subscriber subscriber =
 				this.subscriberFactory.createSubscriber(subscription, messageReceiver);
 		subscriber.startAsync();
 		return subscriber;
 	}
 
+	@Override
+	public Subscriber subscribe(String subscription,
+			Consumer<BasicAcknowledgeablePubsubMessage> messageConsumer) {
+		Assert.notNull(subscriberFactory, "The messageConsumer can't be null.");
```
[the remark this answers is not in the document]
meltsufin: Haha, bad copy paste. :)

meltsufin:
@artembilan Are you OK with merging as it's now?

PR #924, opened 2018-08-02
Spanner allow resetting columns to null

### LINKED ISSUE #919: Change a field to null is ignored when a save is perform
I'm trying to change a field of an entity to null, but when I save the entity the previous value in the data base remain the same.

I found this in spring-cloud-gcp-data-spanner-1.0.0.BUILD-TVC-SNAPSHOT-sources.jar
org.springframework.cloud.gcp.data.spanner.core.convert
.ConverterAwareMappingSpannerEntityWriter. writeProperty
if (propertyValue == null) {
return;
}

Is this intentional or a bug? What is the suggested work around to remove an association id of another table? setting the associated id field to empty?
Thanks so far

ChengyuanZhao:
fixes #919 

Previously we did not expect users would save an entity and then save it again with some columns set to null. However based on user feedback and that the client lib supports this directly, it has been enabled. Conveniently, the client lib allows binding of any column type to null when creating a mutation.

### DIFF
--- a/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/ConverterAwareMappingSpannerEntityWriter.java
+++ b/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/ConverterAwareMappingSpannerEntityWriter.java
@@ -297,8 +297,7 @@ public class ConverterAwareMappingSpannerEntityWriter implements SpannerEntityWr
 							iterablePropertyType2ToMethodMap.get(targetType);
 					toMethod.accept(valueBinder,
 							value == null ? null
-									:
-							ConversionUtils.convertIterable(value, targetType, this.writeConverter));
+									: ConversionUtils.convertIterable(value, targetType, this.writeConverter));
 					valueSet = true;
 					break;
 				}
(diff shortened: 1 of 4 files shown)


spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/ConverterAwareMappingSpannerEntityWriter.java
```diff
@@ -321,7 +319,9 @@ private boolean attemptSetIterableValue(Iterable<Object> value,
 			return false;
 		}
 		// We're just checking for the bind to have succeeded, we don't need to chain the result.
-		Object ignored = toMethod.apply(valueBinder, this.writeConverter.convert(value, targetType));
+		// Spanner allows binding of null values.
+		Object ignored = toMethod.apply(valueBinder,
```
meltsufin: Why are we capturing the `ignored` method return value?

ChengyuanZhao: Let me check, but if I recall it was for a compiler warning? I'll check

ChengyuanZhao: Ah it was this:
```
[ERROR] /usr/local/google/home/cccz/IdeaProjects/spring-cloud-gcp/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/ConverterAwareMappingSpannerEntityWriter.java:[323,31] [ReturnValueIgnored] Return value of this method must be used
```

That is what happens if the return value isn't stored to something.

benjacarp:
Hi, when is this going to be released?
Thanks

ChengyuanZhao:
@benjacarp This change appears now in the 1.1.0.BUILD-SNAPSHOT, and will also be included in our next milestone release (1.1.0.M2) when that comes out.

Issue #930, opened 2018-08-06
Support custom host for Cloud Spanner


olavloite:
The Cloud Spanner client library allows a custom host to be specified instead of using the default https://spanner.googleapis.com. Setting a custom host makes it possible to connect to a (local) emulator instead of a real Cloud Spanner instance for testing/development purposes. Adding a 'host' configuration property to GcpSpannerAutoConfiguration could solve this.

ChengyuanZhao:
Thank you for the issue and PR. However, we can't merge this since we don't plan to support 3rd party emulators/services endpoints right now.

PR #931, opened 2018-08-06
add configuration for setting the Spanner host to use

olavloite:
This PR adds the functionality described in #930 

If a host is specified in the configuration file, that host will be used when creating a Spanner instance. This could be a (local) emulator instead of a real Cloud Spanner instance.
If no host is specified, the default Google Cloud Spanner host will automatically be used.

See http://www.googlecloudspanner.com/2018/07/create-trial-account-for-cloud-spanner.html for a short description on how to create an account for an emulator that you could use to test this.

### DIFF
--- a/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/spanner/GcpSpannerAutoConfiguration.java
+++ b/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/spanner/GcpSpannerAutoConfiguration.java
@@ -63,6 +63,8 @@ import org.springframework.data.rest.webmvc.spi.BackendIdConverter;
 public class GcpSpannerAutoConfiguration {
 
 	static class CoreSpannerAutoConfiguration {
+		
+		private final String host;
 
 		private final String projectId;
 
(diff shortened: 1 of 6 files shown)


spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/spanner/GcpSpannerProperties.java
```diff
@@ -34,6 +34,8 @@
 	private final Credentials credentials = new Credentials(
 			GcpScope.SPANNER_DATA.getUrl(), GcpScope.SPANNER_ADMIN.getUrl());
 
+	private String host;
```
artembilan: This is a mix of concerns: we are here in a cloud and that fully doesn't look like someone is going to configure a host explicitly in the cloud environment.

I'm pretty OK to have an `emulator-host` property and auto-configure something on the matter, but that's already must be outside of the standard cloud-based auto-configuration.

See what we have with the `GcpPubSubEmulatorConfiguration` as a source for ideas.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/spanner/GcpSpannerAutoConfiguration.java
```diff
@@ -64,6 +64,8 @@
 
 	static class CoreSpannerAutoConfiguration {
 
+		private final String host;
```
artembilan: The same problem here: an  emulator matter has to be separate from the standard configuration, even if a difference only with the one property.

ChengyuanZhao:
Closing this PR. Please see the linked issue.

PR #932, opened 2018-08-06
Datastore custom converters

### LINKED ISSUE #888: Datastore converter custom converters

dmitry-s:
fixes #888


spring-cloud-gcp-data-datastore/src/test/java/org/springframework/cloud/gcp/data/datastore/core/convert/ConverterAwareMappingDatastoreEntityConverterTests.java
```diff
@@ -52,23 +54,28 @@ public void setUp() {
...
+		Entity entity = getaEntityBuilder()
+				.set("durationField", "PT24H")
 				.set("stringField", "string value")
 				.set("boolField", true)
 				.set("doubleField", 3.1415D)
 				.set("longField", 123L)
 				.set("latLngField", LatLng.of(10, 20))
 				.set("timestampField", Timestamp.ofTimeSecondsAndNanos(30, 40))
 				.set("blobField", Blob.copyFrom(bytes))
+				.set("intField", 99)
```
ChengyuanZhao: I think you forgot the assertThat line that checks this int field

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/StorageAwareConversions.java
```diff
@@ -0,0 +1,25 @@
...
+ *  limitations under the License.
+ */
+
+package org.springframework.cloud.gcp.data.datastore.core.convert;
+
+
+/**
+ * @author Dmitry Solomakha
+ */
+public interface StorageAwareConversions {
```
ChengyuanZhao: Is this interface necessary? it's only referred to in the implements-line of DatastoreSimpleTypes

spring-cloud-gcp-data-datastore/src/test/java/org/springframework/cloud/gcp/data/datastore/core/convert/ConverterAwareMappingDatastoreEntityConverterTests.java
```diff
@@ -77,58 +84,71 @@ public void readTest() {
...
+				new ConverterAwareMappingDatastoreEntityConverter(new DatastoreMappingContext());
 		TestDatastoreItem item = entityConverter.read(TestDatastoreItem.class, entity);
 
 		assertThat(item.getStringField()).as("validate null field").isNull();
 	}
 
 	@Test(expected = DatastoreDataException.class)
 	public void testWrongTypeReadException() {
-				Entity entity = Entity.newBuilder(this.datastore.newKeyFactory().setKind("aKind").newKey("1"))
+		Entity entity = getaEntityBuilder()
```
ChengyuanZhao: I think this needs to be "getAnEntity..." or "getOneEntity..." or just "getEntity..."

dmitry-s: It actually returns a builder, so renaming it to getEntity might be misleading

ChengyuanZhao: Ah ok, then "getAnEntity" or "getOneEntity" or "get1Entity". It's just that "getaEntity" isn't correct because the "a" isn't capitalized and also "entity" starts with a vowel so it must be preceded by "an"

dmitry-s: oh, you right 
that was an IDE automatic refactoring generated name :)

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/ConverterAwareMappingDatastoreEntityConverter.java
```diff
@@ -71,7 +73,11 @@
 
 	private final GenericConversionService conversionService = new DefaultConversionService();
```
ChengyuanZhao: I think this looks rather confusing when finally seeing what using 2-layers of converters actually looks like.

There are now 4 things to somebody not deeply-in-the-know that look nearly interchangeable:

```
private final GenericConversionService conversionService
private final GenericConversionService toNativeConversionService
private final DatastoreSimpleTypes toNativeConversions
private final DatastoreCustomConversions customConversions
```

I much prefer reading and understanding the previous version where we don't use the 2-layer shoe-horned Spring Data Core custom conversions. It's also harder to reason about if our test cases are comprehensive enough when dealing with this multi-layer conversion.

At the end of the day if somebody is switching from Mongo to Datastore , a whole different storage tech, is that they need to supply a Double->Long converter rather than the Double->Int converter really a material inconvenience at all?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/ConverterAwareMappingDatastoreEntityConverter.java
```diff
@@ -0,0 +1,194 @@
...
+ * @author Dmitry Solomakha
+ *
+ * @since 1.1
+ */
+public class ConverterAwareMappingDatastoreEntityConverter implements DatastoreEntityConverter {
+	private DatastoreMappingContext mappingContext;
+
+	private final EntityInstantiators instantiators = new EntityInstantiators();
+
+	private final GenericConversionService conversionService = new DefaultConversionService();
```
meltsufin: s/GenericConversionService/ConversionService

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/ConverterAwareMappingDatastoreEntityConverter.java
```diff
@@ -0,0 +1,194 @@
...
+
+	private final GenericConversionService toNativeConversionService = new DefaultConversionService();
```
meltsufin: Don't need another conversion service, just use the one above in the second stage.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/ConverterAwareMappingDatastoreEntityConverter.java
```diff
@@ -0,0 +1,194 @@
...
+
+	private final DatastoreSimpleTypes toNativeConversions = new DatastoreSimpleTypes(this.conversionService);
+
+	private final DatastoreCustomConversions customConversions;
```
meltsufin: Change type to `CustomConversions`.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/ConverterAwareMappingDatastoreEntityConverter.java
```diff
@@ -0,0 +1,194 @@
...
+					" with name " + persistentProperty.getFieldName()
+					+ ". The property's simple type is " + propertyVal.getClass());
+		}
+
+		Value val = (Value) valueFactory.apply(propertyVal);
+		sink.set(persistentProperty.getFieldName(), val);
+	}
+
+	private Object convertToSimpleType(@Nullable Object obj, DatastorePersistentProperty prop) {
+		if (obj == null || DatastoreSimpleTypes.isSimple(obj.getClass())) {
```
meltsufin: You only need to check for `null`, and rest should be taken care of by `getCustomWriteTarget`.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/ConverterAwareMappingDatastoreEntityConverter.java
```diff
@@ -0,0 +1,194 @@
...
+				(DatastorePersistentProperty datastorePersistentProperty) ->
+						writeProperty(sink, accessor, datastorePersistentProperty));
+	}
+
+	@SuppressWarnings("unchecked")
+	private void writeProperty(Entity.Builder sink, PersistentPropertyAccessor accessor,
+			DatastorePersistentProperty persistentProperty) {
+		Object propertyVal = convertToSimpleType(accessor.getProperty(persistentProperty), persistentProperty);
+		if (propertyVal == null) {
+			sink.set(persistentProperty.getFieldName(), new NullValue());
```
meltsufin: You can just add this to the `DatastoreSimpleTypes` map with `null` key.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/ConverterAwareMappingDatastoreEntityConverter.java
```diff
@@ -0,0 +1,194 @@
...
+	private final GenericConversionService toNativeConversionService = new DefaultConversionService();
+
+	private final DatastoreSimpleTypes toNativeConversions = new DatastoreSimpleTypes(this.conversionService);
+
+	private final DatastoreCustomConversions customConversions;
+
+	private static final Map<Class<?>, Function<?, Value<?>>> VAL_FACTORIES;
+
+	static {
+		VAL_FACTORIES = ImmutableMap.<Class<?>, Function<?, Value<?>>>builder()
```
meltsufin: `DATASTORE_TYPE_WRAPPERS`?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/ConverterAwareMappingDatastoreEntityConverter.java
```diff
@@ -0,0 +1,194 @@
...
+		Object propertyVal = convertToSimpleType(accessor.getProperty(persistentProperty), persistentProperty);
+		if (propertyVal == null) {
+			sink.set(persistentProperty.getFieldName(), new NullValue());
+			return;
+		}
+		Optional<Class<?>> basicTargetType = this.toNativeConversions.getCustomWriteTarget(propertyVal.getClass());
+		if (basicTargetType.isPresent()) {
+			propertyVal = this.toNativeConversionService.convert(propertyVal, basicTargetType.get());
+		}
+		Function valueFactory = VAL_FACTORIES.get(propertyVal.getClass());
```
meltsufin: maybe call it `wrapper`?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/ConverterAwareMappingDatastoreEntityConverter.java
```diff
@@ -0,0 +1,194 @@
...
+
+		if (valueFactory == null) {
+			throw new DatastoreDataException("Simple type value factory is not found for property" +
```
meltsufin: fractory -> wrapper
..."simple type is not supported by Datastore"

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/ConverterAwareMappingDatastoreEntityConverter.java
```diff
@@ -0,0 +1,194 @@
...
+import org.springframework.lang.Nullable;
+
+/**
+ * A class for object to entity and entity to object conversions
+ *
+ * @author Dmitry Solomakha
+ *
+ * @since 1.1
+ */
+public class ConverterAwareMappingDatastoreEntityConverter implements DatastoreEntityConverter {
```
meltsufin: Let's rename this class to `DefaultDatastoreEntityConverter`.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/EntityPropertyValueProvider.java
```diff
@@ -52,13 +62,22 @@
 	private <T> T readSingleWithConversion(DatastorePersistentProperty persistentProperty) {
 		Object val = this.entity.getValue(persistentProperty.getFieldName()).get();
 		Class<?> targetType = persistentProperty.getType();
-		if (val == null) {
-			return null;
-		}
-		if (ClassUtils.isAssignable(targetType, val.getClass())) {
+
+		if (val == null || ClassUtils.isAssignable(targetType, val.getClass())) {
 			return (T) val;
```
meltsufin: This will not work for a case of `Object` -> `String`, where `String` is a JSON string that you need to convert.

Move it to the final case.

spring-cloud-gcp-data-datastore/src/test/java/org/springframework/cloud/gcp/data/datastore/core/convert/TestDatastoreItemUnsupportedFields.java
```diff
@@ -16,15 +16,16 @@
...
 package org.springframework.cloud.gcp.data.datastore.core.convert;
 
-import java.math.BigInteger;
+
+import java.util.Objects;
 
 /**
  * @author Dmitry Solomakha
  */
 class TestDatastoreItemUnsupportedFields {
```
meltsufin: See if you can add a field here that is of type `Object` and uses a converter to convert to `String`. Something like where your Object is actually a string and you store it in reverse.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/DefaultDatastoreEntityConverter.java
```diff
@@ -0,0 +1,245 @@
...
+ * @author Dmitry Solomakha
+ *
+ * @since 1.1
+ */
+public class DefaultDatastoreEntityConverter implements DatastoreEntityConverter {
+	private DatastoreMappingContext mappingContext;
+
+	private final EntityInstantiators instantiators = new EntityInstantiators();
+
+	private final GenericConversionService conversionService = new DefaultConversionService();
```
ChengyuanZhao: It looks like this same conversion service registered with the same set of converters is used for both Step 1 and Step 2. To somebody without the extensive context of the discussions, it begs the questions:
-  why just 2 steps if you're going to be connecting converters together? it wouldn't be much code to extend it to do some search of arbitrary steps ?
- why is there this very specific set of allowable intermediate types between the 2 steps?

My point is that I think there should be some explicit javadoc or something about the rationale for this 2-step method and how the Spring Data core CustomConversions's set of simple types is the hidden motivation here

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/EntityPropertyValueProvider.java
```diff
@@ -51,14 +62,30 @@
...
 		if (val == null) {
 			return null;
 		}
-		if (ClassUtils.isAssignable(targetType, val.getClass())) {
-			return (T) val;
+
+		Class<?> targetType = persistentProperty.getType();
+		T result = null;
+
+		//To mirror write conversion, try 1-step conversion first, then try 2-step conversion,
```
ChengyuanZhao: It's great that you noted here that we specifically want to do the conversions with converters in the exact reverse direction and same intermediate type as the write direction.  I think it would be safer to make some sort of new single class and interface that packages this bi-directional-same-path in a single place so somebody later on doesn't make changes and forget to preserve this property.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/DefaultDatastoreEntityConverter.java
meltsufin: @dmitry-s I thought we decided that we would use a separate instance of the `DefaultConversionService` for the second stage?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/DefaultDatastoreEntityConverter.java
```diff
@@ -0,0 +1,281 @@
...
+				" with name " + persistentProperty.getFieldName()
+				+ " to Datastore supported type. The property's type is " + propertyVal.getClass());
+	}
+
+	static class TwoStepConversion {
+		private Class<?> firstStepTarget;
+
+		private Class<?> secondStepTarget;
+
+		TwoStepConversion(Class<?> firstStepSource, CustomConversions customConversions,
```
meltsufin: Since you're doing all this work in the constructor, maybe just create a private method instead, and return a `TwoStepConversion` bean.

meltsufin:
@dmitry-s The build needs to be fixed too. Also, we probably need a separate issue for updating the docs related to type conversion.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/DefaultDatastoreEntityConverter.java
```diff
@@ -0,0 +1,245 @@
...
+
+		static final List<Class<?>> DATASTORE_NATIVE_TYPES_RESOLUTION;
+
+		static {
+			ID_TYPES = ImmutableSet.<Class<?>>builder()
+					.add(String.class)
+					.add(Long.class)
+					.build();
+
+			DATASTORE_NATIVE_TYPES_RESOLUTION = ImmutableList.<Class<?>>builder()
```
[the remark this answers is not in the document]
meltsufin: I agree with @ChengyuanZhao here, and in fact the iteration order of the immutable map is the same as the insertion order.
See: https://github.com/google/guava/blob/master/guava/src/com/google/common/collect/ImmutableMap.java#L365

dmitry-s:
@meltsufin  
Updated, please take a look.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/EntityPropertyValueProvider.java
```diff
@@ -33,8 +36,19 @@
 public class EntityPropertyValueProvider implements PropertyValueProvider<DatastorePersistentProperty> {
 	private Entity entity;
 
-	EntityPropertyValueProvider(Entity entity) {
+	private ConversionService conversionService;
+
+	private ConversionService internalConversionService;
+
+	private Function<Class<?>, DefaultDatastoreEntityConverter.TwoStepConversion> conversionGetter;
```
meltsufin: A reference to a function here seems odd. I think we need to think a bit more about this design.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/EntityPropertyValueProvider.java
```diff
@@ -51,14 +65,37 @@
...
 		if (val == null) {
 			return null;
 		}
-		if (ClassUtils.isAssignable(targetType, val.getClass())) {
-			return (T) val;
+
+		Class<?> targetType = persistentProperty.getType();
+		T result = null;
+
+		DefaultDatastoreEntityConverter.TwoStepConversion twoStepConversion = this.conversionGetter.apply(targetType);
```
meltsufin: Why don't we have a method/interface that simply takes an `Object` and a `targetType` and does the conversion?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/EntityPropertyValueProvider.java
```diff
@@ -33,8 +36,19 @@
 public class EntityPropertyValueProvider implements PropertyValueProvider<DatastorePersistentProperty> {
 	private Entity entity;
 
-	EntityPropertyValueProvider(Entity entity) {
+	private ConversionService conversionService;
```
meltsufin: This `conversionService` may be different from the one in `DefaultDatastoreEntityConverter` which seems to be a potential problem. I think that conversion logic for reading and for writing should be in the same class, not spread out between `EntityPropertyValueProvider` and `DefaultDatastoreEntityConverter`.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/EntityPropertyValueProvider.java
dmitry-s: mostly for symmetry with value write logic, but we can put all this in one class

dmitry-s:
@meltsufin PTAL

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/DatastoreCustomConversions.java
```diff
@@ -0,0 +1,60 @@
...
+import org.springframework.data.convert.ThreeTenBackPortConverters;
+
+/**
+ * Value object to capture custom conversion. {@link DatastoreCustomConversions}
+ *
+ * @author Dmitry Solomakha
+ *
+ * @since 1.1
+ */
+public class DatastoreCustomConversions extends org.springframework.data.convert.CustomConversions {
```
meltsufin: Import `org.springframework.data.convert.CustomConversions`.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/DatastoreSimpleTypes.java
```diff
@@ -0,0 +1,132 @@
...
+import org.springframework.data.mapping.model.SimpleTypeHolder;
+
+/**
+ * A class to manage Datastore-specific simple type conversions.
+ *
+ * @author Dmitry Solomakha
+ *
+ * @since 1.1
+ */
+class DatastoreSimpleTypes {
```
meltsufin: Is this class package-private on purpose?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/DatastoreSimpleTypes.java
```diff
@@ -0,0 +1,132 @@
...
+	static final Map<Class<?>, Function<?, Value<?>>> DATASTORE_TYPE_WRAPPERS;
+
+	static final Set<Class<?>> DATASTORE_NATIVE_TYPES;
```
meltsufin: Since the order matters, I would change it explicitly to a `List`.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/DatastoreSimpleTypes.java
```diff
@@ -0,0 +1,132 @@
...
+		if (wrapper != null) {
+			return (Value) wrapper.apply(propertyVal);
+		}
+		throw new DatastoreDataException("Unable to convert a property" +
+				" with name " + persistentProperty.getFieldName()
+				+ " to Datastore supported type. The property's type is " + propertyVal.getClass());
+	}
+
+
+	public Optional<Class<?>> getCustomWriteTarget(Class<?> sourceType) {
```
meltsufin: This one, and the method below are only used in `TwoStepsConversion`. Can we just move the two methods there?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/TwoStepsConversions.java
```diff
@@ -0,0 +1,150 @@
...
+ * In order to support {@link CustomConversions}, this class applies 2-step conversions.
+ * The first step produces one of {@link SimpleTypeHolder}'s simple types.
+ * The second step converts simple types to Datastore-native types.
+ * The second step is skipped if the first one produces a Datastore-native type.
+ *
+ * @author Dmitry Solomakha
+ *
+ * @since 1.1
+ */
+class TwoStepsConversions implements ReadWriteConversions {
```
meltsufin: Is this package-private on purpose?

PR #935, opened 2018-08-07
Remove mentions of the acknowledger in the refdoc

meltsufin:


artembilan:
The fix also must be back-ported to `1.0.x` branch.

PR #936, opened 2018-08-07
Bump dep versions to latest

meltsufin:
The new versions highlighted a bug in the `stackdriverSender` definition, which this change also fixes.


meltsufin:
@artembilan I'm also planning to merge this bump of versions into `1.0.x`. Is that OK, or should we stay on the older RC1 versions there?

PR #938, opened 2018-08-08
Fix property name for read bucket in GCS SI sample

meltsufin:


meltsufin:
Yes, that's why I'm doing this. I'll do a back-port PR as well.

PR #947, opened 2018-08-10
Datastore sample and readme

### LINKED ISSUE #678: Datastore sample module

ChengyuanZhao:
fixes #678


spring-cloud-gcp-samples/spring-cloud-gcp-data-datastore-sample/src/main/java/com/example/DatastoreRepositoryExample.java
```diff
@@ -0,0 +1,71 @@
...
+			Iterable<Singer> allSingers = this.singerRepository.findAll();
+			allSingers.forEach(System.out::println);
+
+			System.out.println("You can also retrieve by keys for strong consistency: ");
+
+			// Retrieving by keys or querying with a restriction to a single entity group
+			// / family is strongly consistent.
+			this.singerRepository
+					.findAllById(ImmutableList.of("singer1", "singer2", "singer3"))
+					.forEach(System.out::println);
```
dmitry-s: would be good to demonstrate non-key queries, like find by last name

ChengyuanZhao: Currently we don't have a method to submit arbitrary queries like this. I think we should include this eventually (perhaps when we support the later functions like Query Methods and paging and sorting), but at the moment I think it's not much different if the user just uses the autowired `Datastore` client lib service to directly send the query. 

As for the Repository class itself, the only find/read method is based on keys. When we implement Query Methods later is when we'll have the ability to query on other properties in the POJO.

spring-cloud-gcp-samples/spring-cloud-gcp-data-datastore-sample/src/main/java/com/example/DatastoreRepositoryExample.java
```diff
@@ -0,0 +1,71 @@
...
+			this.singerRepository.deleteAll();
+
+			this.singerRepository.save(new Singer("singer1", "John", "Doe"));
+			this.singerRepository
+					.saveAll(ImmutableList.of(new Singer("singer2", "Mary", "Jane"),
+							new Singer("singer3", "Scott", "Smith")));
+
+			// The following line uses count(), which is a global-query in Datastore. This
+			// has only eventual consistency.
+			Thread.sleep(3000);
```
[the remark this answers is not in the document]
ChengyuanZhao: This is a sample, and not really work-doing real code. I think a comment mentioning this is just a period of few seconds to wait is OK here. Doing a retry loop with perhaps a max-number-of-attempts constant is too much to read, and I don't think conveys anything more to the user than the simple message: "you need to wait usually because Datastore has eventual consistency here"

ChengyuanZhao:
@dmitry-s @meltsufin Addressed the PR comments, PTAL

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/datastore/GcpDatastoreAutoConfiguration.java
```diff
@@ -52,8 +52,7 @@
 @Configuration
 @AutoConfigureAfter(GcpContextAutoConfiguration.class)
 @ConditionalOnProperty(value = "spring.cloud.gcp.datastore.enabled", matchIfMissing = true)
-@ConditionalOnClass({ DatastoreMappingContext.class, DatastoreOperations.class,
-		ObjectToKeyFactory.class, DatastoreEntityConverter.class })
+@ConditionalOnClass({ DatastoreMappingContext.class, DatastoreOperations.class })
```
meltsufin: Actually, these two are also from the same package. We only need one of them - `DatastoreOprations.class`.
We also need to make sure that we're conditional on the presence of the Datastore client library by including `com.google.cloud.datastore.Datastore`.

ChengyuanZhao: added Datastore.class and removed DatastoreMappingContext.class

PR #955, opened 2018-08-13
Upgrade to Spring Boot and Spring Cloud Build v2.1

### LINKED ISSUE #951: Upgrade to Spring Cloud Build 2.1.0 to be aligned with the upcoming Spring Cloud Greenwich

meltsufin:
Fixes #951.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -6,7 +6,7 @@
 	<parent>
 		<groupId>org.springframework.cloud</groupId>
 		<artifactId>spring-cloud-build</artifactId>
-		<version>2.0.2.RELEASE</version>
+		<version>2.1.0.BUILD-SNAPSHOT</version>
 		<relativePath/><!-- lookup parent from repository -->
 	</parent>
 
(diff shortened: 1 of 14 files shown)


meltsufin:
@artembilan That's the issue with `protobuf-java:2.6`. The fix in Spring Boot didn't propagate yet. We'll re-run the build again tomorrow.

meltsufin:
Since we do not use the Spring Boot BOM, we still need to exclude the `protobuf-java` dep, until MySQL fixes the problem [reported here](https://bugs.mysql.com/bug.php?id=91999).

jlheard:
When switching from `1.1.0.M1` to `1.1.0.M2` we're seeing a missing dependency:
 
> Invocation of init method failed; nested exception is java.lang.NoClassDefFoundError: org/springframework/data/repository/query/QueryMethodEvaluationContextProvider

Do we need to add a dependency when switching to M2?

PR #963, opened 2018-08-17
add table names to the error messages

### LINKED ISSUE #959: Spanner data @Interleaved exception handling
For the attached exception stack traces it would be good to know the entity name (Parent & child entities) along with column names (if applicable) for which this exception is raised. 

image

image

**This is good . but column names would help in better triaging.** 

image

dmitry-s:
Fixes #959

### DIFF
--- a/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/mapping/SpannerPersistentEntityImpl.java
+++ b/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/mapping/SpannerPersistentEntityImpl.java
@@ -214,7 +214,9 @@ public class SpannerPersistentEntityImpl<T>
 				throw new SpannerDataException(
 						"A child table must contain the primary key columns of its "
 								+ "parent in the same order starting the first "
-								+ "column with additional" + " key columns after.");
+								+ "column with additional key columns after."
+								+ "The parent is " + getType().getSimpleName()
+								+ " and the child is " + childEntity.getType().getSimpleName() + ".");
 			}
 			for (int i = 0; i < primaryKeyProperties.size(); i++) {
 				SpannerPersistentProperty parentKey = primaryKeyProperties.get(i);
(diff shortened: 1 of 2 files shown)


spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/mapping/SpannerPersistentEntityImpl.java
```diff
@@ -225,7 +227,9 @@ private void verifyInterleavedProperties() {
 							"The child primary key column (" + childKey.getColumnName()
```
meltsufin: Can you just say `"The child primary key column (" +  childEntity.getType().getSimpleName() + "." + childKey.getColumnName() + ") ... does not match that of its parent (" + getType().getSimpleName() + "." + "parentKey.getColumnName() + " ..."`. I think it's easier to read than broken up into two sentences.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/mapping/SpannerPersistentEntityImpl.java
```diff
@@ -265,7 +269,7 @@ private void verifyPrimaryKeysConsecutive() {
 				throw new SpannerDataException(
 						"The primary key columns were not given a consecutive order. "
 								+ "There is no property annotated with order "
-								+ String.valueOf(i));
```
meltsufin: `String.valueOf(i)`???
You can just use `i`.

spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/core/mapping/SpannerPersistentEntityImplTests.java
```diff
@@ -80,7 +79,7 @@ public void testColumns() {
 				.columns(), containsInAnyOrder("id", "custom_col"));
 	}
 
-	@Test(expected = SpelEvaluationException.class)
+	@Test(expected = SpannerDataException.class)
```
meltsufin: You want to check more than just the exception type, but also the message.

PR #970, opened 2018-08-27
add collections support

### LINKED ISSUE #887: Datastore converter support collections

dmitry-s:
Fixes #887.


spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/DefaultDatastoreEntityConverter.java
```diff
@@ -79,21 +85,35 @@ public DefaultDatastoreEntityConverter(DatastoreMappingContext mappingContext,
...
-						writeProperty(sink, accessor, datastorePersistentProperty));
+						writeProperty(
+								sink, accessor.getProperty(datastorePersistentProperty), datastorePersistentProperty));
 	}
 
-	private void writeProperty(Entity.Builder sink, PersistentPropertyAccessor accessor,
+	private void writeProperty(Entity.Builder sink, Object propertyVal,
 			DatastorePersistentProperty persistentProperty) {
-		Object propertyVal = accessor.getProperty(persistentProperty);
+		boolean isArray = propertyVal != null && Object[].class.isAssignableFrom(propertyVal.getClass());
```
meltsufin: Maybe add a comment or turn the two lines into a utility method - `wrapArrayIntoList`.

meltsufin: We don't want to break `byte[]` up in to a list. Let's make sure that it gets converted to `Blob`.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/EntityPropertyValueProvider.java
```diff
@@ -41,32 +48,74 @@
...
+			List<?> elements = this.entity.getList(datastorePersistentProperty.getFieldName())
+					.stream()
+					.map(v -> readSingleWithConversion(v.get(), componentType, datastorePersistentProperty))
+					.collect(Collectors.toList());
+			try {
+				return (T) convertToCollection(elements, targetType);
+			}
+			catch (ConversionFailedException e) {
+				throw new DatastoreDataException(
+						"Unable to read property " + datastorePersistentProperty.getFieldName(), e);
```
meltsufin: Add entity name here.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/ReadWriteConversions.java
```diff
@@ -27,4 +27,6 @@
 	<T> T convertOnRead(Object val, Class<?> targetType);
 
 	<T> T convertOnWrite(Object obj);
+
+	<T> T convertCollection(Object collection, Class<?> aClass);
```
meltsufin: We shouldn't need this additional method. Read and write should already cover the purpose of this interface.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/EntityPropertyValueProvider.java
```diff
@@ -41,32 +48,74 @@
...
 	public <T> T getPropertyValue(DatastorePersistentProperty datastorePersistentProperty) {
 		String fieldName = datastorePersistentProperty.getFieldName();
 
 		if (!this.entity.contains(fieldName)) {
 			throw new DatastoreDataException("Field not found: " + fieldName);
 		}
 
-		return readSingleWithConversion(datastorePersistentProperty);
+		Class<?> targetType = datastorePersistentProperty.getTypeInformation().getType();
+		if (isNonNullCollection(datastorePersistentProperty)) {
```
meltsufin: Can we move this logic to `TwoStepsConversions`?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/TwoStepsConversions.java
```diff
@@ -49,11 +74,35 @@ public TwoStepsConversions(CustomConversions customConversions) {
...
+		this.internalConversionService.addConverter(BLOB_TO_BYTE_ARRAY_CONVERTER);
+
 	}
 
 	@Override
 	@SuppressWarnings("unchecked")
-	public <T> T convertOnRead(Object val, Class<?> targetType) {
+	public <T> T convertOnRead(Value val, Class<?> targetType, Class<?> componentType) {
+		Object unwrappedVal = val.get();
+		if (unwrappedVal instanceof List) {
```
ChengyuanZhao: Here we're relying on the client library giving back `List` for all of the singular types (string, number, blob, etc) to identify if we should be converting a single item or multiple items. Though i don't expect this to change (by that i mean what if they start returning Collection or Iterable in the future), I feel like it would be safer if we got this indication of List vs single item from the POJO side. By that I mean why don't we check `targetType` to see if it is a List or single item?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/DefaultDatastoreEntityConverter.java
```diff
@@ -67,39 +69,56 @@ public DefaultDatastoreEntityConverter(DatastoreMappingContext mappingContext,
...
-			ValueBuilder valueBuilder = val.toBuilder();
-			valueBuilder.setExcludeFromIndexes(true);
-			val = valueBuilder.build();
-		}
-		sink.set(persistentProperty.getFieldName(), val);
+				(DatastorePersistentProperty persistentProperty) -> {
+					try {
+						Object val = accessor.getProperty(persistentProperty);
+						//Check if property is a non-null array
+						if (val != null && persistentProperty.isArray() && val.getClass() != byte[].class) {
```
ChengyuanZhao: Since you've moved the list-vs-single conversion code to convertOnWrite, why not move this byte[] checking case in there as well?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/TwoStepsConversions.java
```diff
@@ -49,11 +74,35 @@ public TwoStepsConversions(CustomConversions customConversions) {
...
+			try {
+				List<?> elements = ((List<Value>) unwrappedVal)
+						.stream()
+						.map(v -> convertOnReadSingle(v.get(), componentType))
```
ChengyuanZhao: Would `componentType` be null or unavailable if `targetType` wasn't a list/collection? It seems like if the user had made a mistake and defined his POJO with a singular property but the corresponding datastore column was a List, then we would be on this line of code without a `componentType`. What would the error message be in that case?

spring-cloud-gcp-data-datastore/src/test/java/org/springframework/cloud/gcp/data/datastore/core/convert/TestDatastoreItemCollections.java
```diff
@@ -0,0 +1,83 @@
...
+ */
+public class TestDatastoreItemCollections {
+	private List<Integer> intList;
+
+	private ImmutableSet<Double> doubleSet;
+
+	private String[] stringArray;
+
+	private boolean[] boolArray;
+
```
ChengyuanZhao: Could you add a property like `byte[][]` or `List<byte[]>` just to make sure the converter (to Blob) is used for the component type and also that we don't try to treat `byte[]` like an array?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/ReadWriteConversions.java
```diff
@@ -24,7 +26,8 @@
  * @since 1.1
  */
 public interface ReadWriteConversions {
-	<T> T convertOnRead(Object val, Class<?> targetType);
+	<T> T convertOnRead(Value val, Class<?> targetType, Class<?> componentType);
```
meltsufin: Add javadoc here, especially to explain `componentType`.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/ReadWriteConversions.java
```diff
@@ -24,7 +26,8 @@
...
+
+	Value convertOnWrite(Object obj);
```
meltsufin: Javadoc here too.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/TwoStepsConversions.java
meltsufin: Good point. For increased robustness, I would change `List` to `Iterable` here.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/TwoStepsConversions.java
```diff
@@ -78,12 +127,29 @@ else if (typeTargets.getFirstStepTarget() != null && typeTargets.getSecondStepTa
...
+		}
+		else {
+			throw new DatastoreDataException("Unable to convert " + val.getClass() + " to " + targetType);
+		}
+
 	}
 
 	@Override
+	public Value convertOnWrite(Object propertyVal) {
+		if (propertyVal instanceof Collection) {
```
meltsufin: Why not `Iterable`?

PR #977, opened 2018-08-28
Fix PubSubChannelAdaptersIntegrationTests

### LINKED ISSUE #976: PubSubChannelAdaptersIntegrationTests broken
Root cause is that `GcpPubSubHeaders.ACKNOWLEDGEMENT` message header is set to `ConvertedBasicAcknowledgeablePubsubMessage`, which no longer implements `AckReplyConsumer`.

meltsufin:
Implement `AckReplyConsumer` using `ConvertedBasicAcknowledgeablePubsubMessage` and store it
in `GcpPubSubHeaders.ACKNOWLEDGEMENT`.

Fixes #976.

### DIFF
--- a/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapter.java
+++ b/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapter.java
@@ -18,6 +18,7 @@ package org.springframework.cloud.gcp.pubsub.integration.inbound;
 
 import java.util.Map;
 
+import com.google.cloud.pubsub.v1.AckReplyConsumer;
 import com.google.cloud.pubsub.v1.Subscriber;
 
 import org.springframework.cloud.gcp.pubsub.core.PubSubException;
@@ -120,7 +121,17 @@ public class PubSubInboundChannelAdapter extends MessageProducerSupport {
 
 		if (this.ackMode == AckMode.MANUAL) {
 			// Send the consumer downstream so user decides on when to ack/nack.
-			messageHeaders.put(GcpPubSubHeaders.ACKNOWLEDGEMENT, message);
+			messageHeaders.put(GcpPubSubHeaders.ACKNOWLEDGEMENT, new AckReplyConsumer() {
+				@Override
+				public void ack() {
+					message.ack();
+				}
+
+				@Override
+				public void nack() {
+					message.nack();
+				}
+			});
 		}
 
 		try {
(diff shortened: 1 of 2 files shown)


spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/GcpPubSubHeaders.java
```diff
@@ -29,7 +29,4 @@
 
 	public static final String TOPIC = PREFIX + "topic";
 
-	public static final String ACK_ID = PREFIX + "ackId";
```
meltsufin: These are not used anywhere.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapter.java
```diff
@@ -120,7 +121,17 @@ private void consumeMessage(ConvertedBasicAcknowledgeablePubsubMessage message)
 
 		if (this.ackMode == AckMode.MANUAL) {
 			// Send the consumer downstream so user decides on when to ack/nack.
-			messageHeaders.put(GcpPubSubHeaders.ACKNOWLEDGEMENT, message);
+			messageHeaders.put(GcpPubSubHeaders.ACKNOWLEDGEMENT, new AckReplyConsumer() {
```
artembilan: Well, what is the problem to stick here with the `BasicAcknowledgeablePubsubMessage` contract?
Looks like it covers all the `AckReplyConsumer` requirements and follows our abstractions.

More over that would be better to do seeing your removal in other class.
Looks like we are missing then those important headers otherwise.
So, with the `BasicAcknowledgeablePubsubMessage` we still continue to get access to the `ackId` and `subscrbiption`. Isn't it?

I mean wouldn't it be better to fix a test-case instead?

meltsufin: I thought of doing that too, but it requires fixes to refdoc and samples, and deprecation of `GcpPubSubHeaders.ACKNOWLEDGEMENT` as it would no longer be compatible with `1.0.0.RELEASE`.

If we were to add the `ConvertedBasicAcknowledgeablePubsubMessage` into the `Message` headers, we should probably call the header `GcpPubSubHeaders.MESSAGE`.
Maybe we can do it in a separate PR.

artembilan: Well, that's fine. Let's introduce a new header (`GcpPubSubHeaders.ORIGINAL_MESSAGE` ?) and deprecate an `ACKNOWLEDGEMENT` one.
Also there would be need to fix docs to mention how to (n)ack now.
And yes: we need to keep the deprecated header for a while, at least until `1.2`.

If you want, let's do that in the separate PR, but I definitely don't like how we are  missing `ack` and `subscription` headers...

meltsufin: I created #979 for that. Let's get this in to fix the CI build.

PR #978, opened 2018-08-28
Datastore Annotated query methods 

### LINKED ISSUE #899: Datastore annotated query-language query methods

ChengyuanZhao:
fixes #899


spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/DatastoreQueryLookupStrategy.java
```diff
@@ -0,0 +1,102 @@
...
+		Assert.notNull(evaluationContextProvider,
+				"A non-null EvaluationContextProvider is required.");
+		Assert.notNull(expressionParser, "A non-null SpelExpressionParser is required.");
+		this.datastoreMappingContext = datastoreMappingContext;
+		this.evaluationContextProvider = evaluationContextProvider;
+		this.datastoreOperations = datastoreOperations;
+		this.expressionParser = expressionParser;
+	}
+
+	Class<?> getEntityType(QueryMethod queryMethod) {
```
meltsufin: Should it be `private`? And the one below.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/DatastoreQueryLookupStrategy.java
```diff
@@ -0,0 +1,102 @@
...
+			String sql = queryMethod.getQueryAnnotation().value();
+			return createGqlDatastoreQuery(entityType, queryMethod, sql);
+		}
+		else if (namedQueries.hasQuery(queryMethod.getNamedQueryName())) {
+			String sql = namedQueries.getQuery(queryMethod.getNamedQueryName());
+			return createGqlDatastoreQuery(entityType, queryMethod, sql);
+		}
+
+		throw new DatastoreDataException(
+				"Only GQL-annotated and properties-file GQL Query Methods are currently supported.");
```
meltsufin: `@Query`-annotated probably more understandable... or just say what is *not* supported.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/DatastoreQueryLookupStrategy.java
```diff
@@ -0,0 +1,102 @@
...
+	}
+
+	<T> GqlDatastoreQuery<T> createGqlDatastoreQuery(Class<T> entityType,
```
meltsufin: `private`?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/DatastoreQueryMethod.java
```diff
@@ -0,0 +1,75 @@
...
+		this.method = method;
+	}
+
+	/**
+	 * Returns whether the method has an annotated query.
+	 *
+	 * @return True if this query method has annotation that holds the query string.
+	 */
+	public boolean hasAnnotatedQuery() {
+		return Optional.ofNullable(getQueryAnnotation()).map(AnnotationUtils::getValue)
```
meltsufin: Even if it's empty, let's not use properties named query if `@Query` is present.
```java
return findAnnotatedQuery().isPresent();
```
??

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/GqlDatastoreQuery.java
```diff
@@ -0,0 +1,327 @@
...
+	private static final Map<Class<?>, Function<Builder, BiFunction<String, Object, Builder>>>
+			GQL_PARAM_BINDING_FUNC_MAP;
+
+	// A small string that isn't used in GQL syntax
+	private static String ENTITY_CLASS_NAME_BOOKEND = "|";
+
+	static {
+		GQL_PARAM_BINDING_FUNC_MAP = ImmutableMap
+				.<Class<?>, Function<Builder, BiFunction<String, Object, Builder>>>builder()
+				.put(Cursor.class, builder -> (s, o) -> builder.setBinding(s, (Cursor) o))
```
meltsufin: Should we also map the array types like `String[]`?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/GqlDatastoreQuery.java
```diff
@@ -0,0 +1,327 @@
...
+		return rawResult.stream().map(result -> processRawObjectForProjection(result))
+				.collect(Collectors.toList());
+	}
+
+	@Override
+	public QueryMethod getQueryMethod() {
+		return this.queryMethod;
+	}
+
+	@VisibleForTesting
```
meltsufin: Still needed?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/GqlDatastoreQuery.java
```diff
@@ -0,0 +1,327 @@
...
+	Object processRawObjectForProjection(Object object) {
+		return this.queryMethod.getResultProcessor().processResult(object);
+	}
+
+	List<T> executeRawResult(Object[] parameters) {
+		List<Object> params = new ArrayList<>();
+
+		for (Object param : parameters) {
```
meltsufin: Use a utility method `Arrays.toList` instead of the loop.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/GqlDatastoreQuery.java
ChengyuanZhao: I'm thinking let's only support these singlular types directly right now (w/o conversion). I will make an issue for using conversion and arrays for a separate PR.

meltsufin: You wouldn't need to convert `String[]`.

ChengyuanZhao: @meltsufin changes made, PTAL

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/support/DatastoreRepositoryFactory.java
```diff
@@ -82,4 +103,39 @@ protected Object getTargetRepository(RepositoryInformation metadata) {
...
+				delegateContextProvider(evaluationContextProvider), EXPRESSION_PARSER));
+	}
+
+	@Override
+	public void setApplicationContext(ApplicationContext applicationContext)
+			throws BeansException {
+		this.applicationContext = applicationContext;
+	}
+
+	private QueryMethodEvaluationContextProvider delegateContextProvider(
```
meltsufin: This seems only need for SpEL.

ChengyuanZhao: My reasoning for keeping this one is:

1. This private method is needed by `getQueryLookupStrategy`, which is a function we must override to use query methods (the parent version returns nothing). 
2. In that function we are given the `QueryMethodEvaluationContextProvider` (which i infer is because the Spring Data folks expected SpEL to be used in query methods), so we might as well keep it and pass it down even if we aren't going to use it right now.
3. then this private function just adds this object's instance of the applicationContext to that  `QueryMethodEvaluationContextProvider`. Nothing specifically SpEL about doing this, and if we are going to not throw away that `QueryMethodEvaluationContextProvider` then we might as well also add this.applicationContext to it.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/DatastoreQueryLookupStrategy.java
```diff
@@ -0,0 +1,108 @@
...
+				this.datastoreMappingContext);
+	}
+
+	@VisibleForTesting
+	Class<?> getEntityType(QueryMethod queryMethod) {
+		return queryMethod.getResultProcessor().getReturnedType().getDomainType();
+	}
+
+	@VisibleForTesting
+	DatastoreQueryMethod createQueryMethod(Method method, RepositoryMetadata metadata,
```
meltsufin: What's the point of having this method? What does it add over just using `new DatastoreQueryMethod(method, metadata, factory)`?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/GqlDatastoreQuery.java
```diff
@@ -0,0 +1,202 @@
...
+	private SpelExpressionParser expressionParser;
+
+	/**
+	 * Constructor
+	 * @param type the underlying entity type
+	 * @param queryMethod the underlying query method to support.
+	 * @param datastoreOperations used for executing queries.
+	 * @param datastoreMappingContext used for getting metadata about entities.
+	 */
+	GqlDatastoreQuery(Class<T> type, QueryMethod queryMethod,
```
meltsufin: public/private?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/GqlDatastoreQuery.java
```diff
@@ -0,0 +1,202 @@
...
+		this.datastoreOperations = datastoreOperations;
+		this.datastoreMappingContext = datastoreMappingContext;
+		this.evaluationContextProvider = evaluationContextProvider;
+		this.expressionParser = expressionParser;
+		this.gql = StringUtils.trimTrailingCharacter(gql.trim(), ';');
+	}
+
+	@Override
+	public Object execute(Object[] parameters) {
+		List<T> rawResult = executeRawResult(parameters);
```
meltsufin: Are we going to support count queries? Cassandra implementation [supports](https://github.com/spring-projects/spring-data-cassandra/blob/c843293fa52cca77d6d0860845420167664903ee/spring-data-cassandra/src/main/java/org/springframework/data/cassandra/repository/query/AbstractCassandraQuery.java#L145) it for example. Also, Datastore has count query support.

spring-cloud-gcp-docs/src/main/asciidoc/datastore.adoc
```diff
@@ -386,15 +386,50 @@ This feature is not supported yet.
...
+Custom GQL queries can be mapped to repository methods in one of two ways:
+
+ * `namedQueries` properties file
+ * using the `@Query` annotation
+
+===== Query methods with annotation
+
+Using the `@Query` annotation:
+
+The names of the tags of the GQL correspond to the `@Param` annotated names of the method parameters.
```
meltsufin: Documentation should list which types are supported for the parameters.

spring-cloud-gcp-docs/src/main/asciidoc/datastore.adoc
```diff
@@ -386,15 +386,50 @@ This feature is not supported yet.
...
 
-This feature is not supported yet.
+Custom GQL queries can be mapped to repository methods in one of two ways:
+
+ * `namedQueries` properties file
+ * using the `@Query` annotation
+
+===== Query methods with annotation
+
+Using the `@Query` annotation:
```
meltsufin: I don't see any tests for this. Can we have a test with an actual repository?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/DatastoreQueryMethod.java
```diff
@@ -0,0 +1,71 @@
...
+
+	/**
+	 * Creates a new {@link QueryMethod} from the given parameters. Looks up the correct
+	 * query to use for following invocations of the method given.
+	 *
+	 * @param method must not be {@literal null}.
+	 * @param metadata must not be {@literal null}.
+	 * @param factory must not be {@literal null}.
+	 */
+	DatastoreQueryMethod(Method method, RepositoryMetadata metadata,
```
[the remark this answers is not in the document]
ChengyuanZhao: It can't be private because it's used elsewhere in the package. It doesn't need to be public because it's only used elsewhere in the package.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/DatastoreQueryMethod.java
```diff
@@ -0,0 +1,71 @@
...
+		super(method, metadata, factory);
+		this.method = method;
+	}
+
+	/**
+	 * Returns whether the method has an annotated query.
+	 *
+	 * @return True if this query method has annotation that holds the query string.
+	 */
+	boolean hasAnnotatedQuery() {
```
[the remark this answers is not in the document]
ChengyuanZhao: It can't be private because it's used elsewhere in the package. It doesn't need to be public because it's only used elsewhere in the package. package-private is appropriate.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/DatastoreQueryMethod.java
```diff
@@ -0,0 +1,71 @@
...
+	}
+
+	/**
+	 * Returns the {@link Query} annotation that is applied to the method or {@code null}
+	 * if none available.
+	 *
+	 * @return the query annotation that is applied.
+	 */
+	@Nullable
+	Query getQueryAnnotation() {
```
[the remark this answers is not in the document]
ChengyuanZhao: It can't be private because it's used elsewhere in the package. It doesn't need to be public because it's only used elsewhere in the package. package-private is appropriate.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/DatastoreQueryLookupStrategy.java
ChengyuanZhao: It's visible for testing as annotated. It's not practical to make a bunch of metadata that usually Spring Data generates from scanning a user's repo interface for a unit test.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/GqlDatastoreQuery.java
ChengyuanZhao: It can't be private because it's used elsewhere in the package. It doesn't need to be public because it's only used elsewhere in the package. package-private is appropriate.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/GqlDatastoreQuery.java
ChengyuanZhao: Yup, we support it with Spanner, too. You're looking at their "Abstract" query class that supports both annotated and also name-convention query methods (that's also how I do it in Spanner). The "count" ones only come from name-convention query methods, which we aren't supporting in this PR in this file.

spring-cloud-gcp-docs/src/main/asciidoc/datastore.adoc
ChengyuanZhao: This is tested in the integration test.

spring-cloud-gcp-docs/src/main/asciidoc/datastore.adoc
ChengyuanZhao: @meltsufin change made, PTAL.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/GqlDatastoreQuery.java
meltsufin: Why cant you have method with `@Query` that returns a count instead of the entities?

ChengyuanZhao: Unless you want to get into GQL parsing, I don't know how you could detect if a Query Method should return the domain type T or just a Long. 

The name-convention query methods' names are parsed by Spring Data into a `PartTree` for you and that will take care of checking if the name implies a `count` query method for you.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/GqlDatastoreQuery.java
```diff
@@ -0,0 +1,202 @@
...
+			return null;
+		}
+		return rawResult.stream().map(this::processRawObjectForProjection)
+				.collect(Collectors.toList());
+	}
+
+	private GqlQuery<Entity> bindArgsToGqlQuery(String gql, List<String> tags,
+			Object[] vals) {
+		Builder<Entity> builder = GqlQuery.newGqlQueryBuilder(ResultType.ENTITY, gql);
+		if (tags.size() != vals.length) {
```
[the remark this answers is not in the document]
ChengyuanZhao: this method is only called with the results of `getParamTags()` which takes care of the repeated usage of tags.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/GqlDatastoreQuery.java
meltsufin: [Cassandra](https://github.com/spring-projects/spring-data-cassandra/blob/c843293fa52cca77d6d0860845420167664903ee/spring-data-cassandra/src/main/java/org/springframework/data/cassandra/repository/Query.java) just has property on the `@Query`.
We can leave this to another PR of course.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/GqlDatastoreQuery.java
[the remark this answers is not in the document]
meltsufin: You throw an exception there, but maybe we shouldn't disallow it.
```java
				throw new DatastoreDataException(
						"More than one param has the same name: " + name);`
```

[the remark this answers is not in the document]
ChengyuanZhao: @meltsufin no this refers to something like
`myfunction(@Param("paramA") String blah, @Param("paramA") double x) ` 

The example you gave is allowed.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/GqlDatastoreQuery.java
meltsufin: Can you add this as a question for our next conversation with the Spring Data team?

ChengyuanZhao: Will make note.

PR #982, opened 2018-08-28
Spanner save all in single batch

### LINKED ISSUE #974: Batch Insert/Upsert
Hi!

I'm currently trying to optimise the upsert(ion) of multiple entities (resulting in 100k-500k rows). According to the [official Google docs](https://cloud.google.com/spanner/docs/getting-started/java/) I should crate a list of mutations (`List<Mutation>`) and handle these over to the dbClient (`dbClient.write(mutations);`), while making sure that the list of mutations and hence the resulting commit won't exceed the 5MB limit as suggested by [the docs](https://cloud.google.com/spanner/docs/bulk-loading). 

Currently the only way I found to upsert (or insert) into the database is to use the `SpannerTemplate.upsert(Object o)`method. Hence currently I can imagine something like this:

```java
 public void bulkUpsert(Iterable<Car> objectsToUpsert){

    carRepository.getSpannerTemplate().performReadWriteTransaction(spannerTemplate -> {
      for(Car car : objectsToUpsert){
        spannerTemplate.upsert(car);
      }
      return objectsToUpsert;
    });
  }
```

Is it really the "most I can get out of" `spring-cloug-gcp`? Does the library issue 100k commits towards the database, or is there optimisation in the background?

Or am I overcomplicating the things and I should simply issue a `carRepository.saveAll(objectsToUpsert);` ? 

Thank you for your effort in the library!

ChengyuanZhao:
fixes #974

### DIFF
--- a/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
+++ b/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
@@ -27,6 +27,7 @@ import java.util.StringJoiner;
 import java.util.function.BiFunction;
 import java.util.function.Function;
 import java.util.stream.Collectors;
+import java.util.stream.StreamSupport;
 
 import javax.annotation.Nullable;
 
@@ -212,7 +213,7 @@ public class SpannerTemplate implements SpannerOperations {
 	}
 
 	@Override
-	public void insertAll(List objects) {
+	public void insertAll(Iterable objects) {
 		applyMutationsUsingEntity(getListCollectionFunction(this.mutationFactory::insert),
 				objects);
 	}
@@ -223,7 +224,7 @@ public class SpannerTemplate implements SpannerOperations {
 	}
 
 	@Override
-	public void updateAll(List objects) {
+	public void updateAll(Iterable objects) {
 		applyMutationsUsingEntity(
 				getListCollectionFunction(x -> this.mutationFactory.update(x, null)),
 				objects);
@@ -247,7 +248,7 @@ public class SpannerTemplate implements SpannerOperations {
 	}
 
 	@Override
-	public void upsertAll(List objects) {
+	public void upsertAll(Iterable objects) {
 		applyMutationsUsingEntity(
 				getListCollectionFunction(x -> this.mutationFactory.upsert(x, null)),
 				objects);
@@ -268,6 +269,13 @@ public class SpannerTemplate implements SpannerOperations {
 	@Override
 	public void delete(Object entity) {
 		applyMutationUsingEntity(this.mutationFactory::delete, entity);
+	}
+
+	@Override
+	public void deleteAll(Iterable objects) {
+		this.databaseClient.write(
+				(Iterable<Mutation>) StreamSupport.stream(objects.spliterator(), false)
+						.map(this.mutationFactory::delete).collect(Collectors.toList()));
 	}
 
 	@Override
@@ -476,9 +484,9 @@ public class SpannerTemplate implements SpannerOperations {
 				});
 	}
 
-	private Function<List, Collection<Mutation>> getListCollectionFunction(
+	private Function<Iterable, Collection<Mutation>> getListCollectionFunction(
 			Function<Object, Collection<Mutation>> individualEntityMutationFunc) {
-		return list -> (Collection<Mutation>) list.stream()
+		return it -> (Collection<Mutation>) StreamSupport.stream(it.spliterator(), false)
 				.flatMap(x -> individualEntityMutationFunc.apply(x).stream())
 				.collect(Collectors.toList());
 	}
--- a/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
+++ b/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
@@ -214,7 +214,7 @@ public class SpannerTemplate implements SpannerOperations {
 	@Override
 	public void insertAll(Iterable objects) {
 		applyMutations(
-				getListCollectionFunction(this.mutationFactory::insert).apply(objects));
+				getListCollectionFunction(objects, this.mutationFactory::insert));
 	}
 
 	@Override
@@ -225,8 +225,8 @@ public class SpannerTemplate implements SpannerOperations {
 	@Override
 	public void updateAll(Iterable objects) {
 		applyMutations(
-				getListCollectionFunction(x -> this.mutationFactory.update(x, null))
-						.apply(objects));
+				getListCollectionFunction(objects,
+						x -> this.mutationFactory.update(x, null)));
 	}
 
 	@Override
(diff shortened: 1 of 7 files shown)


spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
```diff
@@ -463,4 +475,11 @@ private void resolveChildEntity(Object entity) {
 											childPersistentEntity)));
 				});
 	}
+
+	private Function<List, Collection<Mutation>> getListCollectionFunction(
```
meltsufin: Maybe it's just me, but this a bit confusing. Why don't we have a simple function that takes a list of objects and returns an `Iterable<Mutation>`? I think it all goes back to `applyMutationTwoArgs` which expects a function that generates the mutations rather than the straight forward list of mutations. Why was the `applyMutationTwoArgs` necessary instead of just `writeMutations(Iterable<Mutation>)`?

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
```diff
@@ -210,11 +211,24 @@ public void insert(Object object) {
 		applyMutationsUsingEntity(this.mutationFactory::insert, object);
 	}
 
+	@Override
+	public void insertAll(List objects) {
```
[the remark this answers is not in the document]
ChengyuanZhao: We're generally taking and returning `List`s from functions in the Spanner module (whereas it's mostly `Iterable` in the Datastore module). Honestly the `List`s are far more convenient to work with, and the other methods in this template all work with Lists, too.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
ChengyuanZhao: The reason is that there are 4 different types of mutation-generating functions, (insert, update, upsert, delete) and those mutation-generating functions are the variables when refactored into the way it currently is. 

The essence of this PRIVATE function is just the flatMap: each item to write individually is also a collection of mutations because each individual item can result in many mutations of children.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/support/SimpleSpannerRepository.java
```diff
@@ -85,9 +86,7 @@ public SpannerOperations getSpannerTemplate() {
 	@Override
 	public <S extends T> Iterable<S> saveAll(Iterable<S> entities) {
 		Assert.notNull(entities, "A non-null list of entities is required for saving.");
-		for (S entity : entities) {
-			save(entity);
-		}
+		this.spannerTemplate.upsertAll(Lists.newArrayList(entities));
```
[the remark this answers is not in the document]
ChengyuanZhao: @meltsufin changes made, PTAL

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
```diff
@@ -246,29 +268,29 @@ public void upsert(Object object, Optional<Set<String>> includeColumns) {
...
 	@Override
 	public void delete(Object entity) {
-		applyMutationUsingEntity(
-				this.mutationFactory::delete, entity);
+		applyMutationUsingEntity(this.mutationFactory::delete, entity);
+	}
+
+	@Override
+	public void deleteAll(Iterable objects) {
+		this.databaseClient.write(
```
meltsufin: Why in this situation you don't use `applyMutationTwoArgs`?
I don't mind it as I don't really like the `applyMutationTwoArgs` design, but just wondering.

ChengyuanZhao: I think it would look rather the same (in terms of number of lines and key functions called) in this situation because we need to do this extra stream-map call due to mutationFactory::delete returning a single mutation (since we rely on cascading delete) rather than Collection<mutation> (as with the other write operations). 

regarding that other delete function, I think you're right. I'll remove that one. When adding the deleteAll function just now I hadn't noticed we had one that acted similarly already. The new deleteAll is more consistent with the other "all" functions, and I'll keep that one.

ChengyuanZhao: @meltsufin changes made, PTAL.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
```diff
@@ -207,68 +208,87 @@ private StringBuilder wrapAsSubSelect(String sql) {
...
 	public void upsert(Object object) {
-		applyMutationsTwoArgs(this.mutationFactory::upsert, object, null);
+		applyMutations(this.mutationFactory.upsert(object, null));
+	}
+
+	@Override
+	public void upsertAll(Iterable objects) {
+		applyMutations(
+				getListCollectionFunction(x -> this.mutationFactory.upsert(x, null))
+						.apply(objects));
```
dmitry-s: You always immediately call apply(objects) on the result of getListCollectionFunction. I think it would be better to pass  the objects to getListCollectionFunction along with individualEntityMutationFunc and return a list of mutations back instead of returning a function.

ChengyuanZhao: change made, PTAL

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
```diff
@@ -207,68 +208,87 @@ private StringBuilder wrapAsSubSelect(String sql) {
...
 	}
 
 	@Override
-	public void delete(Class entityClass, Key key) {
-		applyMutationTwoArgs(
-				this.mutationFactory::delete,
-				entityClass, key);
+	public void deleteAll(Iterable objects) {
+		applyMutations(
+				(Collection<Mutation>) StreamSupport.stream(objects.spliterator(), false)
```
meltsufin: Can't use `getMutationsForMultipleObjects` here?

ChengyuanZhao: That method requires each element produces a collection of mutations, which isn't the case for `delete` because of relying on the cascade for delete. as a result, we would need to wrap each delete mutation into a singleton list, and the end result doesn't look better than directly making the list of mutations as shown here.

PR #992, opened 2018-09-04
Datastore run transaction function

### LINKED ISSUE #994: Datastore run in transaction
as requested ,created after-the-fact for #992

ChengyuanZhao:
Expose Datastore client lib's runInTransaction function.

as requested, fixes #994

### DIFF
--- a/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreOperations.java
+++ b/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreOperations.java
@@ -155,8 +155,8 @@ public interface DatastoreOperations {
 
 	/**
 	 * Performs multiple read and write operations in a single transaction.
-	 * @param operations the function representing the operations to perform using a
-	 * DatastoreOperations based on a single transaction.
+	 * @param operations the function that uses {@link DatastoreOperations}
+	 * to perform operations in a transaction.
 	 * @param <A> the final return type of the operations.
 	 * @return the final result of the transaction.
 	 */
--- a/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreOperations.java
+++ b/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreOperations.java
@@ -156,7 +156,7 @@ public interface DatastoreOperations {
 	/**
 	 * Performs multiple read and write operations in a single transaction.
 	 * @param operations the function representing the operations to perform using a
-	 * DatastoreRepository based on a single transaction.
+	 * DatastoreOperations based on a single transaction.
 	 * @param <A> the final return type of the operations.
 	 * @return the final result of the transaction.
 	 */
--- a/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreOperations.java
+++ b/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreOperations.java
@@ -16,6 +16,8 @@
 
 package org.springframework.cloud.gcp.data.datastore.core;
 
+import java.util.function.Function;
+
 import com.google.cloud.datastore.Entity;
 import com.google.cloud.datastore.GqlQuery;
 
@@ -150,4 +152,13 @@ public interface DatastoreOperations {
 	 * @return true if the given ID refers to an existing entity. False otherwise.
 	 */
 	<T> boolean existsById(Object id, Class<T> entityClass);
+
+	/**
+	 * Performs multiple read and write operations in a single transaction.
+	 * @param operations the function representing the operations to perform using a
+	 * DatastoreRepository based on a single transaction.
+	 * @param <A> the final return type of the operations.
+	 * @return the final result of the transaction.
+	 */
+	<A> A performTransaction(Function<DatastoreOperations, A> operations);
 }
--- a/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
+++ b/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
@@ -162,7 +162,8 @@ public class DatastoreTemplate implements DatastoreOperations {
 	public <A> A performTransaction(Function<DatastoreOperations, A> operations) {
 		if (!(this.datastore instanceof Datastore)) {
 			throw new DatastoreDataException(
-					"A Datastore object is required to run transactions.");
+					"This DatastoreReadWriter cannot be used to run transactions. A full Datastore service"
+							+ " object is required to run functions as transactions.");
 		}
 		return ((Datastore) this.datastore).runInTransaction(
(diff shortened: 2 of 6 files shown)


spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreOperations.java
```diff
@@ -150,4 +152,13 @@
 	 * @return true if the given ID refers to an existing entity. False otherwise.
 	 */
 	<T> boolean existsById(Object id, Class<T> entityClass);
+
+	/**
+	 * Performs multiple read and write operations in a single transaction.
+	 * @param operations the function representing the operations to perform using a
+	 * DatastoreOperations based on a single transaction.
```
meltsufin: `{@link DatastoreOperations}`

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
```diff
@@ -155,6 +158,20 @@ public long count(Class<?> entityClass) {
 		return findById(id, entityClass) != null;
 	}
 
+	@Override
+	public <A> A performTransaction(Function<DatastoreOperations, A> operations) {
+		if (!(this.datastore instanceof Datastore)) {
```
meltsufin: Why did you make the change from `Datastore` to `DatastoreReaderWriter`? Otherwise you wouldn't need this check.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreOperations.java
```diff
@@ -150,4 +152,13 @@
 	 * @return true if the given ID refers to an existing entity. False otherwise.
 	 */
 	<T> boolean existsById(Object id, Class<T> entityClass);
+
+	/**
+	 * Performs multiple read and write operations in a single transaction.
+	 * @param operations the function representing the operations to perform using a
```
meltsufin: "a function that uses `DatastoreOpeations` to perform a series of operations in a transaction"

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
ChengyuanZhao: The point is that the single-transaction-only DatastoreTemplate would only have a DatastoreReadWriter, and this check is meant to disallow transactions within transactions, which isn't allowed by Cloud Datastore.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreOperations.java
ChengyuanZhao: @meltsufin changes made, PTAL.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
meltsufin: There are a number of methods in the `Datastore` interface that we may need in the future for `DatastoreTemplate`. Switching to a smaller set of operations available in `DatastoreReaderWriter` might be limiting.

To do this properly without casts, we may need to reflect the type hierarchy of the Datastore library. In other words, we might need a `DatatastoreOperations`, and a `TransactionOperations` interfaces, both inheriting from `DatastoreReadWriteOperations`.

However, if we're confident that we will never need anything that `Datastore` interface adds on top of `DatastoreReaderWriter`, then it's probably not worth the trouble.

ChengyuanZhao:
@meltsufin yes there is an issue specifically for this PR though this PR it's more like a to-do left over from an earlier issue that was too large. The method being implemented actually already existed but was just throwing a not-implemented exception prior to this PR.

Regarding your comment, I don't expect DatastoreTemplate needing the extra methods Datastore adds to DatastoreReaderWriter. The extra methods are for running things in transaction (handled here) and for allocating keys, which we utilize outside of DatastoreTemplate. So, I don't think we should make that hierarchy of classes, because I think it is doubtful it will be necessary in the future (and also doubtful that it would be any more bothersome to add in the future if the functionality differences change (which is also doubtful)).

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
```diff
@@ -155,6 +158,20 @@ public long count(Class<?> entityClass) {
...
+			throw new DatastoreDataException(
+					"This DatastoreReadWriter cannot be used to run transactions. A full Datastore service"
+							+ " object is required to run functions as transactions.");
+		}
+		return ((Datastore) this.datastore).runInTransaction(
+				readerWriter -> operations.apply(new DatastoreTemplate(readerWriter,
```
dmitry-s: please add type specifier for readerWriter

spring-cloud-gcp-docs/src/main/asciidoc/datastore.adoc
```diff
@@ -380,6 +404,31 @@ public class MyApplication {
 
 ----
 
+==== Transactions
```
dmitry-s: I think it would make sense to consolidate all transaction-related documentation into one section instead of having two.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
ChengyuanZhao: Could you clarify what you mean? The anonymous function here is an implementation for the one-function interface TransactionCallable. The type of `readWriter` is determined by that `TransactionCallable` interface.

spring-cloud-gcp-docs/src/main/asciidoc/datastore.adoc
ChengyuanZhao: There was a concerted effort to lay out the structure of ref doc for our Spring Data modules back when we did Spanner, and we have "===" level sections Repository and Template since they are rather large concepts and then "====" sections for features under them (Transactions being such a section). 

If we were to consolidate move the Transactions section out of the Repository and Template sections, it would have to be a "===" section, and I don't think it's a concept that stands at the same level as those two. Furthermore, the transactions methods described specifically allow usage of Repository and Template objects within the anonymous function, and I think moving them away from those parent sections would be kind of orphaning those mentions within the Transactions section.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
dmitry-s: I meant `(DatastoreReaderWriter readerWriter) -> ...`

ChengyuanZhao: Change made, PTAL

PR #1001, opened 2018-09-06
Adds @Transactional support for Spanner

### LINKED ISSUE #845: FR - Spanner - Add More Transaction Support to support @Transactional
Low level transaction management feature was added to Spanner client library:
https://github.com/GoogleCloudPlatform/google-cloud-java/issues/1810

Spring Cloud GCP could benefit from this transaction manager and adopt it for JTA / Spring.

However, I can't tell whether the transaction isolation level is comparable to JTA.

alexkhimich:
Fixes #845.

Spring Cloud GCP Data Spanner

To use `@Transactional` in [declarative spring transaction management](https://docs.spring.io/spring/docs/current/spring-framework-reference/data-access.html#transaction-declarative)
`SpannerTransactionManager` has to be registered in Spring.
Example:
```
@Bean
public SpannerTransactionManager transactionManager() {
    return new SpannerTransactionManager(databaseClient)
}
```

`SpannerTemplate` respects SpannerTransactionManager and transaction boundaries set by `@Transactional` annotation.


saturnism:
For #845

meltsufin:
@alexkhimich Thanks for the contribution! It will take us a few days to review. Bear with us.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
```diff
@@ -305,40 +327,42 @@ public long count(Class entityClass) {
...
 				? this.databaseClient.readOnlyTransaction(
 						TimestampBound.ofReadTimestamp(options.getTimestamp()))
 				: this.databaseClient.readOnlyTransaction()) {
-			return operations.apply(new ReadOnlyTransactionSpannerTemplate(
-					SpannerTemplate.this.databaseClient,
-					SpannerTemplate.this.mappingContext,
-					SpannerTemplate.this.spannerEntityProcessor,
-					SpannerTemplate.this.mutationFactory,
-					SpannerTemplate.this.spannerSchemaUtils, readOnlyTransaction));
+			return operations.apply(
```
ChengyuanZhao: I'm totally new to Spring AOP and transactions so please forgive if I say something that doesn't make sense:

If somebody were to use `performReadOnlyTransaction` or `performReadWriteTransaction` from inside another method annotated with `@Transactional`, then I think that corresponds to the "NESTED" propagation setting : https://docs.spring.io/spring/docs/3.1.x/javadoc-api/org/springframework/transaction/annotation/Propagation.html#NESTED

However, currently in the PR we are getting a TransactionContext that is independent of the ongoing transaction.

alexkhimich:
Hey, totally agree on formatting other classes. However i found my `mvn compile` has been failing on exiting code lines i have not touched. Let me try to revert that.

artembilan:
Would be better to use a `Maven Wrapper` presented in the project.
Therefore: `mvnw clean package`.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
ChengyuanZhao: My recommendation: I think in this function we should be checking if we are already inside an ongoing `@Transactional` transaction and throwing an exception saying Spanner doesn't support transactions-within-transactions.

The 2 alternatives I don't think make sense:

1. current situation where it's an independent transaction. If the containing method got rolled back, then transaction run via the anonymous function inside wouldn't be rolled back, and that is counter-intuitive. 

2. merging the anonymous function into the ongoing transaction. I think this is also counter-intuitive because it would mean the user putting code into an anonymous function is no different than writing it plainly in the containing method's body. it would also make the very suggestive method name `performRead[only/write]Transaction` meaningless, which is unexpected.

meltsufin:
+1 on not reformatting lines that are not being modified. It's quiet distracting.

alexkhimich:
Fixed the formatting.

spring-cloud-gcp-data-spanner/pom.xml
```diff
@@ -1,47 +1,59 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-	xmlns="http://maven.apache.org/POM/4.0.0"
-	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+         xmlns="http://maven.apache.org/POM/4.0.0"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
 
-	<modelVersion>4.0.0</modelVersion>
```
meltsufin: @alexkhimich Thanks for reverting the formatting changes! This file is one you seem to have missed. Yeah I know tabs vs spaces. :smiley:

artembilan: Also I've pointed that I would like do not see Powermock dependency.
Please, follow standard Spring Boot set of Testing frameworks.

alexkhimich:
Well, there is limited we can do for static mocking. I will try but that is
way more then i expected and it will probably will take long time.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTransactionManager.java
```diff
@@ -0,0 +1,208 @@
...
+		logger.debug(tx + " create; state = " + tx.transactionManager.getState());
+		return tx;
+	}
+
+	@Override
+	protected void doBegin(Object transactionObject, TransactionDefinition transactionDefinition)
+			throws TransactionException {
+		Tx tx = (Tx) transactionObject;
+		final TransactionContext targetTransactionContext = tx.transactionManager.begin();
+		if (transactionDefinition.isReadOnly()) {
```
ChengyuanZhao: It looks like there's a function `DatabaseClient.readOnlyTransaction` can provide a read-only context object here.

alexkhimich: correct, however that is different class returned

spring-cloud-gcp-data-spanner/pom.xml
```diff
@@ -1,47 +1,47 @@
...
-			<groupId>org.springframework.cloud</groupId>
-			<artifactId>spring-cloud-gcp-core</artifactId>
-			<scope>test</scope>
-		</dependency>
-		<dependency>
-			<groupId>org.springframework</groupId>
-			<artifactId>spring-tx</artifactId>
-		</dependency>
-	</dependencies>
+    <modelVersion>4.0.0</modelVersion>
```
meltsufin: I think we need tabs here for a complete revert. :smiley:

alexkhimich:
Oh.. :))

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTransactionManager.java
ChengyuanZhao: Yes, but I think it should be accommodated because of the potentially large performance difference.

meltsufin: So, should this be a different issue?

spring-cloud-gcp-data-spanner/README.adoc
```diff
@@ -0,0 +1,13 @@
+= Spring Cloud GCP Data Spanner
```
ChengyuanZhao: We have a separate ref doc for the modules, and here's the  one for Spanner : https://github.com/spring-cloud/spring-cloud-gcp/blob/master/spring-cloud-gcp-docs/src/main/asciidoc/spanner.adoc

Perhaps mention the new transactional annotation in the Transaction sections.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTransactionManager.java
```diff
@@ -0,0 +1,217 @@
...
+							"TransactionDefinition.PROPAGATION_REQUIRED");
+		}
+		Tx tx = (Tx) transactionObject;
+		final TransactionContext targetTransactionContext = tx.transactionManager.begin();
+		if (transactionDefinition.isReadOnly()) {
+
+			tx.transactionContext = new TransactionContext() {
+				@Override
+				public void buffer(Mutation mutation) {
+					throw new IllegalStateException("Spanner transaction is in readonly more");
```
ChengyuanZhao: "mode" vs "more"

spring-cloud-gcp-data-spanner/README.adoc
meltsufin: Good catch! I didn't even realize it wasn't in the refdoc!

spring-cloud-gcp-docs/src/main/asciidoc/spanner.adoc
```diff
@@ -885,6 +885,20 @@ point in time of the transaction.
 Because read-only transactions are non-locking and can be performed on points in time in the past, these
 are recommended for functions that do not perform write operations.
 
+===== @Transactional and declerative transaction management
```
meltsufin: Spelling: "declarative".

I would title it
**Declarative Transaction Management and `@Transactional`**

spring-cloud-gcp-docs/src/main/asciidoc/spanner.adoc
```diff
@@ -885,6 +885,20 @@ point in time of the transaction.
...
+
+To use `@Transactional` in [declerative spring transaction management](https://docs.spring.io/spring/docs/current/spring-framework-reference/data-access.html#transaction-declarative)
```
meltsufin: declarative

spring-cloud-gcp-docs/src/main/asciidoc/spanner.adoc
```diff
@@ -885,6 +885,20 @@ point in time of the transaction.
...
+`SpannerTransactionManager` has to be registered in Spring.
+Example:
+```
+@Bean
+public SpannerTransactionManager transactionManager() {
+    return new SpannerTransactionManager(databaseClient)
```
meltsufin: I think we should add it to our auto-configuration in `GcpSpannerAutoConfiguration`.

alexkhimich: Hm.. There is already Spring's annotation `@EnableTransactionManagement` that controls `@Transactional` support. 

Are you sure every developer would like to have transaction manager setup always on `@ GcpSpannerAutoConfiguration` ? I think more flexible to keep this as explicit config or dedicated annotation.

alexkhimich:
Let me know if you want me to squash commit all my changes in single. After we get all the issues resolved.

alexkhimich:
@artembilan `powermock` dependency has been removed.

spring-cloud-gcp-docs/src/main/asciidoc/spanner.adoc
meltsufin: Ah fair enough. Forgot about the fact that we already integrate with `@EnableTransactionManagement`.

meltsufin: Or do we?

alexkhimich: `@EnableTransactionManagement` just enables `@Transactional` annotations. 
We don't add anything on top of `@EnableTransactionManagement`.

alexkhimich: I think not every consumer `@GcpSpannerAutoConfiguration` wants to have declarative transactions. In fact that aligned with Spring JDBC + Transaction management. 

In Spring JDBC you don't need to have transaction management - you can simply use JdbcTemplate. 

Transaction management enablement is covered by `@EnableTransactionManagement`and there is a spring doc section that covers requirements for transaction manager being registered to your datasource.

from developer experience i think we should keep it consistent. Developer tries to use Spanner - ok use SpannerTemplate, developer wants transactions - go read about spring transaction management and find there is `@transactional` annotation, developer annotates beans with it and realizes there is a requirement for transaction manager, developer goes to this repo and finds there is `SpannerTransactionManager`, developer registers it and now all works.

meltsufin: @olivergierke Can you guide us on this please?

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
```diff
@@ -101,11 +104,23 @@ public SpannerTemplate(DatabaseClient databaseClient,
...
+		}
+		else {
+			return this.databaseClient.singleUse();
+		}
 	}
 
 	protected ReadContext getReadContext(Timestamp timestamp) {
-		return this.databaseClient.singleUse(TimestampBound.ofReadTimestamp(timestamp));
+		Optional<TransactionContext> txContext = getTransactionContext();
+		if (txContext.isPresent()) {
```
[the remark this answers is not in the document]
alexkhimich: @olivergierke thanks for suggestion. This is a preference choice i would keep the code as is.

meltsufin:
@artembilan @ChengyuanZhao Does anyone still have any objections to merging this?

ChengyuanZhao:
@meltsufin no, nothing blocking

spring-cloud-gcp-docs/src/main/asciidoc/spanner.adoc
artembilan: Well, actually there is a `TransactionAutoConfiguration` which auto-configures for us the mentioned `@EnableTransactionManagement`. And this one depends on the `DataSourceTransactionManagerAutoConfiguration`, for example. This one, in turn, just register a `DataSourceTransactionManager`. The same happens in the `Neo4jDataAutoConfiguration`. 

So, for consistency with Spring Boot I would insist that we need to introduce a `SpannerTransactionManagerAutoConfiguration` - fully similar to that one `DataSourceTransactionManagerAutoConfiguration`. However we can do that in the separate issue.

Thanks.

artembilan:
We can merge this one now or it can wait until we fix compatibility with the latest Spring Cloud Stream:
```
[ERROR] /home/travis/build/spring-cloud/spring-cloud-gcp/spring-cloud-gcp-pubsub-stream-binder/src/main/java/org/springframework/cloud/gcp/stream/binder/pubsub/PubSubMessageChannelBinder.java:[40,8] org.springframework.cloud.gcp.stream.binder.pubsub.PubSubMessageChannelBinder is not abstract and does not override abstract method getExtendedPropertiesEntryClass() in org.springframework.cloud.stream.binder.ExtendedBindingProperties
```

spring-cloud-gcp-docs/src/main/asciidoc/spanner.adoc
meltsufin: Yep, in a separate #1032.

PR #1002, opened 2018-09-06
support @Embedded

### LINKED ISSUE #903: Support @Embedded entities

dmitry-s:
fixes #903


spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/EntityPropertyValueProvider.java
```diff
@@ -50,7 +53,10 @@
 			return null;
 		}
 		try {
-
+			if (persistentProperty.isEmbedded()) {
```
ChengyuanZhao: I see you added a  reference to the converter in  this class to do this `read(...)` call. However, when looping through the properties in the converter could you check for if the property is embedded there instead and do the recursive call right there? Perhaps it'll make the changes to the constructors go away, and might be simpler?

dmitry-s: That's not going to work, because it's not the only place where EntityPropertyValueProvider is used.

ChengyuanZhao: Other than in tests, Intellij is saying the only place `EntityPropertyValueProvider` is used is in `read(...)` in `DefaultDatastoreEntityConverter`. Doesn't that mean you can just call `read` directly from `read` in the Converter?

spring-cloud-gcp-data-datastore/src/test/java/org/springframework/cloud/gcp/data/datastore/core/convert/TestItemWithEmbeddedEntity.java
```diff
@@ -0,0 +1,81 @@
...
+import org.springframework.cloud.gcp.data.datastore.core.mapping.Embedded;
+
+/**
+ * @author Dmitry Solomakha
+ */
+public class TestItemWithEmbeddedEntity {
+	int intField;
+
+	@Embedded
+	EmbeddedEtity embeddedEntityField;
```
ChengyuanZhao: I think we need to support collections/arrays of embedded entities because that's a feature Datastore has, but currently it looks like on both the read and write side we do the recursive call before we go into the collections.

dmitry-s: What if it's a tree? :-)

A collection of embedded entities makes sense, I'll create a separate issue for that.

ChengyuanZhao: Good point! otherwise looks good to me

dmitry-s:
Added embedded collections support. @ChengyuanZhao @meltsufin PTAL

spring-cloud-gcp-docs/src/main/asciidoc/datastore.adoc
```diff
@@ -210,7 +210,62 @@ The `@Field` annotation optionally provides a different field name than that of
 
 ==== Embedded Objects
 
-This feature is not supported yet.
+Fields marked as `@Embedded` are converted to Datastore entities and stored inside the parent entity.
```
ChengyuanZhao: Great work! 

I think it's worth mentioning in the doc that the Embedded annotation here means something different than what it means for SQL style Spring Data modules: here it means it is a sub-struct/sub-row, but for SQL modules it just means using  the columns of the embedded object as columns of the parent. I think it would be good to prevent confusion.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/TwoStepsConversions.java
```diff
@@ -82,13 +94,20 @@ public TwoStepsConversions(CustomConversions customConversions) {
...
 	@SuppressWarnings("unchecked")
-	public <T> T convertOnRead(Value val, Class<?> targetType, Class<?> componentType) {
+	public <T> T convertOnRead(Value val, DatastorePersistentProperty persistentProperty) {
+		BiFunction<Object, Class<?>, T> readConverter = persistentProperty.isEmbedded()
+				? (value, targetClass) -> (T) this.datastoreEntityConverter.read(targetClass, (BaseEntity) value)
+				: this::convertOnReadSingle;
+
+		Class<?> targetType = persistentProperty.getType();
+
+		Class<?> componentType = persistentProperty.getComponentType();
```
ChengyuanZhao: If the user had not included the inner type in his POJO definition, what would the error here look like? 

For example if he had used `List myNumbers` instead of `List<Double> myNumbers`. Should we validate this early on when creating the DatastorePersistentEntity and throw an exception telling the user this error?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/TwoStepsConversions.java
```diff
@@ -137,8 +156,15 @@ else if (typeTargets.getFirstStepTarget() != null && typeTargets.getSecondStepTa
 	}
 
 	@Override
-	public Value convertOnWrite(Object proppertyVal) {
+	public Value convertOnWrite(Object proppertyVal, DatastorePersistentProperty persistentProperty) {
 		Object val = proppertyVal;
+
+		Function<Object, Value> writeConverter = this::convertOnWriteSingle;
```
ChengyuanZhao: No big deal, but why not use the ` statement ? trueResult : falseResult ; ` syntax from the read function here ?

spring-cloud-gcp-docs/src/main/asciidoc/datastore.adoc
```diff
@@ -210,7 +210,62 @@ The `@Field` annotation optionally provides a different field name than that of
 
 ==== Embedded Objects
```
meltsufin: s/Objects/Entities

spring-cloud-gcp-docs/src/main/asciidoc/datastore.adoc
```diff
@@ -210,7 +210,62 @@ The `@Field` annotation optionally provides a different field name than that of
...
+
+	public EmbeddableTreeNode(long value, EmbeddableTreeNode left, EmbeddableTreeNode right) {
+		this.value = value;
+		this.left = left;
+		this.right = right;
+	}
+}
+----
+
+Also, a collection of entities can be embedded; it will be converted to ListValue on write.
```
meltsufin: format as code: ListValue

spring-cloud-gcp-docs/src/main/asciidoc/datastore.adoc
```diff
@@ -210,7 +210,62 @@ The `@Field` annotation optionally provides a different field name than that of
...
+
+Example:
+
+Instead of a binary tree from the previous example, we would like to store a general tree
+(each node can have arbitrary number of children) in Cloud Datastore.
```
meltsufin: an arbitrary

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/TwoStepsConversions.java
```diff
@@ -148,11 +174,19 @@ public Value convertOnWrite(Object proppertyVal) {
...
+				values.add(writeConverter.apply(propEltValue));
 			}
 			return ListValue.of(values);
 		}
-		return convertOnWriteSingle(val);
+		return writeConverter.apply(val);
+	}
+
+	private EntityValue convertSingleEmbeddedEntity(Object val) {
+		IncompleteKey key = this.objectToKeyFactory.getIncompleteKey("embedded");
```
ChengyuanZhao: "embedded" is the kind name here for this fake key. I think you should make this a static final member somewhere. Also, the name "embedded" seems a little too common, and might be used for real as a kind name, don't you think? 

What if it was like "spring-data-cloud-datastore-embedded-dummy-kind-name" so that it will never be used. 

Finally , it can definitely work that the embedded class type is actually a real entity with a real type, but maybe we don't want to support this because it creates the opportunity for misuse and confusion.

spring-cloud-gcp-data-datastore/src/test/java/org/springframework/cloud/gcp/data/datastore/it/EmbeddableTreeNode.java
```diff
@@ -0,0 +1,71 @@
...
+import java.util.Objects;
+
+import org.springframework.cloud.gcp.data.datastore.core.mapping.Embedded;
+import org.springframework.data.annotation.Id;
+
+/**
+ * @author Dmitry Solomakha
+ */
+public class EmbeddableTreeNode {
+	@Id
```
ChengyuanZhao: Since this is an embedded object without a real Key, could we test this without it annotated ID ? Or perhaps just remove it from the example ? I wouldn't want a user to read that and thing that the ID annotation was required or that it had real meaning for an embedded entity.

dmitry-s: It is needed here, because it is used for the root node - which is the parent entity

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/ReadWriteConversions.java
```diff
@@ -44,5 +44,9 @@
 	 */
 	Value convertOnWrite(Object obj, DatastorePersistentProperty persistentProperty);
 
+	/**
+	 * Registers DatastoreEntityConverter to be used for embedded entities
```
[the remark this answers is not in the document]
artembilan: We can use `@link` as well if class is used here or if we specify fully qualified name.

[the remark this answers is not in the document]
meltsufin: Does it really work though when the class is in a different project? I would think that the actual javadoc can't be found automatically.

PR #1010, opened 2018-09-11
Fix pull-and-ack on empty list

### LINKED ISSUE #1009: Problem handling empty ack list in PubSubSubscriberTemplate.pullAndAck
In PubSubSubscriberTemplate, when there are no messages returned by a call to pullAndAck the method ends up throwing this exception

com.google.api.gax.rpc.InvalidArgumentException: io.grpc.StatusRuntimeException: INVALID_ARGUMENT: You have not specified an ack ID in the request.

When I would expect it to return an empty List<PubsubMessage>

We are using version 1.1.0.M1

meltsufin:
Stop acking the list of messages, if the pull resulted in an empty list.

Fixes #1009.

### DIFF
--- a/spring-cloud-gcp-pubsub/src/test/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberTemplateTests.java
+++ b/spring-cloud-gcp-pubsub/src/test/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberTemplateTests.java
@@ -57,12 +57,7 @@ import org.springframework.util.concurrent.ListenableFutureCallback;
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.mockito.ArgumentMatchers.any;
-import static org.mockito.Mockito.doAnswer;
-import static org.mockito.Mockito.doNothing;
-import static org.mockito.Mockito.reset;
-
-import static org.mockito.Mockito.verify;
-import static org.mockito.Mockito.when;
+import static org.mockito.Mockito.*;
 
 /**
  * Unit tests for {@link PubSubSubscriberTemplate}.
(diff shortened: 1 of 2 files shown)


spring-cloud-gcp-pubsub/src/test/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberTemplateTests.java
```diff
@@ -57,12 +57,7 @@
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.mockito.ArgumentMatchers.any;
-import static org.mockito.Mockito.doAnswer;
-import static org.mockito.Mockito.doNothing;
-import static org.mockito.Mockito.reset;
-
-import static org.mockito.Mockito.verify;
-import static org.mockito.Mockito.when;
+import static org.mockito.Mockito.*;
```
artembilan: No asterisk imports, please

spring-cloud-gcp-pubsub/src/test/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberTemplateTests.java
```diff
@@ -180,12 +175,12 @@ public void setUp() throws InterruptedException, TimeoutException, ExecutionExce
...
 
 		// create object under test
-		this.pubSubSubscriberTemplate = new PubSubSubscriberTemplate(this.subscriberFactory);
+		this.pubSubSubscriberTemplate = spy(new PubSubSubscriberTemplate(this.subscriberFactory));
 		this.pubSubSubscriberTemplate.setMessageConverter(this.messageConverter);
 	}
 
 	@Test
-	public void testSubscribeAndAck() throws InterruptedException, ExecutionException, TimeoutException {
+	public void testSubscribe_AndAck() throws InterruptedException, ExecutionException, TimeoutException {
```
artembilan: What drive you to do this renaming?
Why not `test_subscribe_and_ack()` then?

meltsufin: This all came out from the original `testPullAndAck` which wasn't testing the `pullAndAck` method, but testing `pull` followed by `ack`. What's the convention that we should follow?

artembilan: Well, for consistency it should not be with the `_` - only cameCase.
But that's my preference so far...
Feel free to leave it as is if it really makes life easier...

PR #1012, opened 2018-09-12
Update Stackdriver Logging documentation

saturnism:
To make the documentation more useable.

### DIFF
--- a/spring-cloud-gcp-autoconfigure/src/main/resources/org/springframework/cloud/gcp/autoconfigure/logging/logback-appender.xml
+++ b/spring-cloud-gcp-autoconfigure/src/main/resources/org/springframework/cloud/gcp/autoconfigure/logging/logback-appender.xml
@@ -5,7 +5,6 @@ Stackdriver appender logback configuration provided for import.
 -->
 
 <included>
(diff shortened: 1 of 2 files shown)


spring-cloud-gcp-autoconfigure/src/main/resources/org/springframework/cloud/gcp/autoconfigure/logging/logback-appender.xml
```diff
@@ -5,7 +5,7 @@ Stackdriver appender logback configuration provided for import.
 -->
 
 <included>
-  <property name="projectId" value="${projectId:-${GOOGLE_CLOUD_PROJECT}}"/>
+  <property name="projectId" value="${projectId:-${SPRING_CLOUD_GCP_LOGGING_PROJECT_ID}"/>
```
meltsufin: @saturnism Did you test this? I don't see where this `projectId` property is even used? It doesn't seem to be passed to `com.google.cloud.logging.logback.LoggingAppender`, let alone the fact that `com.google.cloud.logging.logback.LoggingAppender` doesn't have that property.

saturnism: good point :/ it was in the file so i assumed it would've worked :)

PR #1014, opened 2018-09-12
Add test rule for setting up Pub/Sub emulator

### LINKED ISSUE #435: Consider having binder integration tests start the emulator
The idea is to have something like the `PubSubEmulator` rule in https://github.com/spring-cloud/spring-cloud-gcp/pull/431/commits/d9dc51867227ac0452e872b602598882528d1f52#diff-6ff85717ce7e17b90d04f38c7b5cc8d0 so that the emulator doesn't need to be started before running the tests.

The rule should also be called with `@ClassRule` instead of `@Rule` so that the same instance of the emulator is used.

elefeint:
The new rule sets up an emulator, creates a binder for it, and kills the emulator process when the tests are done.

In the future, the rule should move to some place more easily accessible from other tests.

Fixes #435.

### DIFF
--- a/spring-cloud-gcp-pubsub-stream-binder/src/test/java/org/springframework/cloud/gcp/stream/binder/pubsub/PubSubEmulator.java
+++ b/spring-cloud-gcp-pubsub-stream-binder/src/test/java/org/springframework/cloud/gcp/stream/binder/pubsub/PubSubEmulator.java
@@ -1,0 +1,129 @@
+/*
+ *  Copyright 2018 original author or authors.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package org.springframework.cloud.gcp.stream.binder.pubsub;
+
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+
+import org.junit.rules.ExternalResource;
+
+public class PubSubEmulator extends ExternalResource {
+	private PubSubTestBinder binder;
+
+	// Needed for cleanup.
+	private Process emulatorProcess;
+
+	private String emulatorHost;
+
+	private String emulatorPort;
+
+	private final Path emulatorConfigPath = Paths.get(System.getProperty("user.home")).resolve(
+			Paths.get(".config", "gcloud", "emulators", "pubsub", "env.yaml"));
+
+	@Override
+	protected void before() throws Throwable {
+
+		this.emulatorProcess = new ProcessBuilder("gcloud", "beta", "emulators", "pubsub", "start").start();
+		assertConfigPresent();
+
+		Process envInitProcess = new ProcessBuilder("gcloud", "beta", "emulators", "pubsub", "env-init").start();
+		String envVariable = new BufferedReader(new InputStreamReader(envInitProcess.getInputStream())).readLine();
+		envInitProcess.waitFor();
+
+		String emulatorHostPort = envVariable.substring(envVariable.indexOf('=') + 1);
+		System.out.println("initialization completed; property = " + System.getenv("PUBSUB_EMULATOR_HOST")
+				+ "; property from output = " + envVariable + "; hostname = " + emulatorHostPort);
+
+		int portSeparatorIndex = emulatorHostPort.indexOf(":");
+		if (portSeparatorIndex < 0 || !envVariable.contains("localhost")) {
+			throw new RuntimeException("Malformed host, can't create emulator: " + envVariable);
+		}
+		this.emulatorHost = emulatorHostPort.substring(0, portSeparatorIndex);
+		this.emulatorPort = emulatorHostPort.substring(portSeparatorIndex + 1);
+
+		this.binder = new PubSubTestBinder(emulatorHostPort);
+	}
+
+	/**
+	 * Shuts down the two emulator processes.
+	 *
+	 * gcloud command is shut down through the direct process handle.
+	 * java process is identified and shut down through shell commands.
+	 */
+	@Override
+	protected void after() {
+		try {
+			String hostPortParams = String.format("--host=%s --port=%s", this.emulatorHost, this.emulatorPort);
+			Process psProcess = new ProcessBuilder("ps", "-v").start();
+			new BufferedReader(new InputStreamReader(psProcess.getInputStream()))
+					.lines()
+					.filter(psLine -> psLine.contains(hostPortParams))
+					.map(psLine -> psLine.substring(0, psLine.indexOf(' ')))
+					.forEach(pid -> {
+						this.killProcess(pid);
+					});
+		}
+		catch (IOException e) {
+			System.out.println("Failed to cleanup: ");
+			e.printStackTrace();
+		}
+		this.emulatorProcess.destroy();
+
+	}
+
+	/**
+	 * Returns an already-configured emulator binder when called from within a JUnit method.
+	 */
+	public PubSubTestBinder getBinder() {
+		return this.binder;
+	}
+
+	/**
+	 * Waits until a PubSub emulator configuration file is present.
+	 *
+	 * Fails if the file does not appear after 1 second.
+	 *
+	 * @throws InterruptedException which should interrupt the peaceful slumber and bubble up to fail the test.
+	 */
+	private void assertConfigPresent() throws InterruptedException {
+		int attempts = 10;
+		while (!Files.exists(this.emulatorConfigPath) && --attempts >= 0) {
+			Thread.sleep(100);
+		}
+		if (attempts < 0) {
+			throw new RuntimeException("Emulator could not be configured -- env.yaml not found");
+		}
+	}
+
+	/**
+	 * Attempts to kill a process on best effort basis.
+	 *
+	 * @param pid Presumably a valid PID. No checking done to validate.
+	 */
+	private void killProcess(String pid) {
+		try {
+			new ProcessBuilder("kill", pid).start();
+		}
+		catch (IOException e) {
+			System.err.println(String.format("Failed to clean up PID %s", pid));
+		}
+	}
+}
--- a/spring-cloud-gcp-pubsub-stream-binder/src/test/java/org/springframework/cloud/gcp/stream/binder/pubsub/PubSubEmulator.java
+++ b/spring-cloud-gcp-pubsub-stream-binder/src/test/java/org/springframework/cloud/gcp/stream/binder/pubsub/PubSubEmulator.java
@@ -23,21 +23,37 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
 import org.junit.rules.ExternalResource;
 
+/**
+ * Rule for instantiating and tearing down a Pub/Sub emulator instance.
+ */
 public class PubSubEmulator extends ExternalResource {
-	private PubSubTestBinder binder;
-
-	// Needed for cleanup.
-	private Process emulatorProcess;
-
-	private String emulatorHost;
-
-	private String emulatorPort;
-
-	private final Path emulatorConfigPath = Paths.get(System.getProperty("user.home")).resolve(
+	private static final Path emulatorConfigPath = Paths.get(System.getProperty("user.home")).resolve(
 			Paths.get(".config", "gcloud", "emulators", "pubsub", "env.yaml"));
 
+	private static final Log LOGGER = LogFactory.getLog(PubSubEmulator.class);
+
+
+	// Binder to Pub/Sub emulator for use in individual tests.
+	private PubSubTestBinder binder;
+
+	// Reference to emulator instance, for cleanup.
+	private Process emulatorProcess;
+
+	// Hostname for cleanup, should always be localhost.
+	private String emulatorHost;
+
+	// Port for cleanup.
+	private String emulatorPort;
+
+	/**
+	 * Launches an instance of pubsub simulator, creates a binder.
+	 *
+	 * @throws Throwable for any failed setup.
+	 */
 	@Override
 	protected void before() throws Throwable {
 
@@ -45,19 +61,13 @@ public class PubSubEmulator extends ExternalResource {
 		assertConfigPresent();
 
 		Process envInitProcess = new ProcessBuilder("gcloud", "beta", "emulators", "pubsub", "env-init").start();
-		String envVariable = new BufferedReader(new InputStreamReader(envInitProcess.getInputStream())).readLine();
+
+		// env-init output is a shell command of the form "export PUBSUB_EMULATOR_HOST=localhost:8085".
+		String emulatorInitString
+				= new BufferedReader(new InputStreamReader(envInitProcess.getInputStream())).readLine();
 		envInitProcess.waitFor();
-
-		String emulatorHostPort = envVariable.substring(envVariable.indexOf('=') + 1);
-		System.out.println("initialization completed; property = " + System.getenv("PUBSUB_EMULATOR_HOST")
-				+ "; property from output = " + envVariable + "; hostname = " + emulatorHostPort);
-
-		int portSeparatorIndex = emulatorHostPort.indexOf(":");
-		if (portSeparatorIndex < 0 || !envVariable.contains("localhost")) {
-			throw new RuntimeException("Malformed host, can't create emulator: " + envVariable);
-		}
-		this.emulatorHost = emulatorHostPort.substring(0, portSeparatorIndex);
-		this.emulatorPort = emulatorHostPort.substring(portSeparatorIndex + 1);
+		String emulatorHostPort = emulatorInitString.substring(emulatorInitString.indexOf('=') + 1);
+		extractTeardownParams(emulatorHostPort);
 
 		this.binder = new PubSubTestBinder(emulatorHostPort);
 	}
@@ -67,9 +77,13 @@ public class PubSubEmulator extends ExternalResource {
 	 *
 	 * gcloud command is shut down through the direct process handle.
 	 * java process is identified and shut down through shell commands.
+	 *
+	 * There should normally be only one process with that host/port combination, but if there are more, they will be
+	 * cleaned up as well.
 	 */
 	@Override
 	protected void after() {
+		this.emulatorProcess.destroy();
 		try {
 			String hostPortParams = String.format("--host=%s --port=%s", this.emulatorHost, this.emulatorPort);
 			Process psProcess = new ProcessBuilder("ps", "-v").start();
@@ -82,11 +96,8 @@ public class PubSubEmulator extends ExternalResource {
 					});
 		}
 		catch (IOException e) {
-			System.out.println("Failed to cleanup: ");
-			e.printStackTrace();
+			LOGGER.error("Failed to cleanup: ", e);
 		}
-		this.emulatorProcess.destroy();
-
 	}
 
 	/**
@@ -109,12 +120,32 @@ public class PubSubEmulator extends ExternalResource {
 			Thread.sleep(100);
 		}
 		if (attempts < 0) {
-			throw new RuntimeException("Emulator could not be configured -- env.yaml not found");
+			throw new RuntimeException(
+					"Emulator could not be configured due to missing env.yaml. Are PubSub and beta tools installed?");
 		}
 	}
 
 	/**
+	 * Remembers host/port combination for future process cleanup.
+	 *
+	 * Validates that localhost is the only valid value; this will stop the test if emulator decides to start with an
+	 * IPv6 ::1 address.
+	 *
+	 * @param emulatorHostPort typically "localhost:8085"
+	 */
+	private void extractTeardownParams(String emulatorHostPort) {
+		int portSeparatorIndex = emulatorHostPort.indexOf(":");
+		if (portSeparatorIndex < 0 || !emulatorHostPort.contains("localhost")) {
+			throw new RuntimeException("Malformed host, can't create emulator: " + emulatorHostPort);
+		}
+		this.emulatorHost = emulatorHostPort.substring(0, portSeparatorIndex);
+		this.emulatorPort = emulatorHostPort.substring(portSeparatorIndex + 1);
+	}
+
+	/**
 	 * Attempts to kill a process on best effort basis.
+	 *
+	 * Failure is logged and ignored.
 	 *
 	 * @param pid Presumably a valid PID. No checking done to validate.
 	 */
@@ -123,7 +154,7 @@ public class PubSubEmulator extends ExternalResource {
 			new ProcessBuilder("kill", pid).start();
 		}
 		catch (IOException e) {
-			System.err.println(String.format("Failed to clean up PID %s", pid));
+			LOGGER.error("Failed to clean up PID " + pid);
 		}
 	}
 }
--- a/spring-cloud-gcp-pubsub-stream-binder/src/test/java/org/springframework/cloud/gcp/stream/binder/pubsub/PubSubEmulator.java
+++ b/spring-cloud-gcp-pubsub-stream-binder/src/test/java/org/springframework/cloud/gcp/stream/binder/pubsub/PubSubEmulator.java
@@ -19,9 +19,15 @@ package org.springframework.cloud.gcp.stream.binder.pubsub;
 import java.io.BufferedReader;
 import java.io.IOException;
 import java.io.InputStreamReader;
+import java.nio.file.FileSystems;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.nio.file.StandardWatchEventKinds;
+import java.nio.file.WatchKey;
+import java.nio.file.WatchService;
+import java.util.Optional;
+import java.util.concurrent.TimeUnit;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
@@ -31,11 +37,14 @@ import org.junit.rules.ExternalResource;
  * Rule for instantiating and tearing down a Pub/Sub emulator instance.
  */
 public class PubSubEmulator extends ExternalResource {
-	private static final Path emulatorConfigPath = Paths.get(System.getProperty("user.home")).resolve(
-			Paths.get(".config", "gcloud", "emulators", "pubsub", "env.yaml"));
+	private static final Path EMULATOR_CONFIG_DIR = Paths.get(System.getProperty("user.home")).resolve(
+			Paths.get(".config", "gcloud", "emulators", "pubsub"));
+
+	public static final String ENV_FILE_NAME = "env.yaml";
+
+	private static final Path EMULATOR_CONFIG_PATH = EMULATOR_CONFIG_DIR.resolve(ENV_FILE_NAME);
 
 	private static final Log LOGGER = LogFactory.getLog(PubSubEmulator.class);
-
 
 	// Binder to Pub/Sub emulator for use in individual tests.
 	private PubSubTestBinder binder;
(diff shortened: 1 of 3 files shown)


spring-cloud-gcp-pubsub-stream-binder/src/test/java/org/springframework/cloud/gcp/stream/binder/pubsub/PubSubEmulator.java
```diff
@@ -0,0 +1,160 @@
...
+	/**
+	 * Waits until a PubSub emulator configuration file is present.
+	 *
+	 * Fails if the file does not appear after 1 second.
+	 *
+	 * @throws InterruptedException which should interrupt the peaceful slumber and bubble up to fail the test.
+	 */
+	private void assertConfigPresent() throws InterruptedException {
+		int attempts = 10;
+		while (!Files.exists(this.emulatorConfigPath) && --attempts >= 0) {
```
meltsufin: Is this basically verifying that the emulator started?
It seems somewhat fragile. For example, will it work on Windows?
Can we instead capture the stdout of the process to detect if it started successfully?

spring-cloud-gcp-pubsub-stream-binder/src/test/java/org/springframework/cloud/gcp/stream/binder/pubsub/PubSubEmulator.java
```diff
@@ -0,0 +1,160 @@
...
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.junit.rules.ExternalResource;
+
+/**
+ * Rule for instantiating and tearing down a Pub/Sub emulator instance.
+ */
+public class PubSubEmulator extends ExternalResource {
+	private static final Path emulatorConfigPath = Paths.get(System.getProperty("user.home")).resolve(
+			Paths.get(".config", "gcloud", "emulators", "pubsub", "env.yaml"));
```
meltsufin: Is this the same on other systems like Windows?

spring-cloud-gcp-pubsub-stream-binder/src/test/java/org/springframework/cloud/gcp/stream/binder/pubsub/PubSubEmulator.java
```diff
@@ -0,0 +1,160 @@
...
+
+	/**
+	 * Launches an instance of pubsub simulator, creates a binder.
+	 *
+	 * @throws Throwable for any failed setup.
+	 */
+	@Override
+	protected void before() throws Throwable {
+
+		this.emulatorProcess = new ProcessBuilder("gcloud", "beta", "emulators", "pubsub", "start").start();
```
meltsufin: Does it matter that an emulator might already be running or multiple are allowed to run at the same time?

spring-cloud-gcp-pubsub-stream-binder/src/test/java/org/springframework/cloud/gcp/stream/binder/pubsub/PubSubEmulator.java
elefeint: This file gets overwritten with the host/port combination every time the emulator starts. It looks like env-init command reads from this file to generate "export PUBSUB_EMULATOR_HOST=localhost:8085" configuration.

So checking the file's existence is insufficient anyway -- the important thing is that the file *updated*. I'll try catching the output instead; this should work at least as well.

In principle, it should work on windows -- this is why I used Paths instead of hardcoding the directory.

spring-cloud-gcp-pubsub-stream-binder/src/test/java/org/springframework/cloud/gcp/stream/binder/pubsub/PubSubEmulator.java
elefeint: Looking at the driving cloudsdk python files, yes.

spring-cloud-gcp-pubsub-stream-binder/src/test/java/org/springframework/cloud/gcp/stream/binder/pubsub/PubSubEmulator.java
elefeint: Multiples are allowed, but in local testing, after a couple of localhost:NNNN instances, it will go into IPv6 mode (localhost:::1), which the binder can't handle.

So I'd rather take them down once the test completes.

elefeint:
@meltsufin PTAL? I separated the config creation check (Travis will use this) from config update check (local test runs will use that), and removed the travis pre-test config that created an instance.

artembilan:
Please, don't merge yet.
Give me a chance to review it locally on my Windows.
Thanks

artembilan:
General comment.
@elefeint , please, consider in the future to raise PR against top-level branches and in the `upsrtream`, not `origin`.
This way we will be able to push commits for simple polishing from our side.
And definitely don't raise PRs against `master`. Now you will need to `rest hard` your local and origin masters.

Hope I am clear.

artembilan:
```
Tests in error:
  org.springframework.cloud.gcp.stream.binder.pubsub.PubSubMessageChannelBinderTests: Cannot run program "gcloud": CreateProcess error=2, Не удается найти указанный файл
```

Indeed I just don't have any `gcloud` software on my machine. I believe in this case the tests should be just skipped.

Thoughts?

spring-cloud-gcp-pubsub-stream-binder/src/test/java/org/springframework/cloud/gcp/stream/binder/pubsub/PubSubEmulator.java
```diff
@@ -0,0 +1,216 @@
...
+			Paths.get(".config", "gcloud", "emulators", "pubsub"));
+
+	public static final String ENV_FILE_NAME = "env.yaml";
+
+	private static final Path EMULATOR_CONFIG_PATH = EMULATOR_CONFIG_DIR.resolve(ENV_FILE_NAME);
+
+	private static final Log LOGGER = LogFactory.getLog(PubSubEmulator.class);
+
+	// Binder to Pub/Sub emulator for use in individual tests.
+	private PubSubTestBinder binder;
```
artembilan: This makes this API only tied with the SCSt Binder stuff eliminating more broader usage of such a rule in any our (and end-user) tests where we would like to deal with the Pub/Sub.

Is this really intentional or we can come up with the solution to decouple a Binder environment to another component, e.g. one more JUnit rule?..

spring-cloud-gcp-pubsub-stream-binder/src/test/java/org/springframework/cloud/gcp/stream/binder/pubsub/PubSubEmulator.java
```diff
@@ -0,0 +1,216 @@
...
+
+	/**
+	 * Launches an instance of pubsub simulator, creates a binder.
+	 *
+	 * @throws Throwable for any failed setup.
+	 */
+	@Override
+	protected void before() throws Throwable {
+
+		boolean configPresent = Files.exists(this.EMULATOR_CONFIG_PATH);
```
artembilan: No `this.` on `static` references.

spring-cloud-gcp-pubsub-stream-binder/src/test/java/org/springframework/cloud/gcp/stream/binder/pubsub/PubSubEmulator.java
```diff
@@ -0,0 +1,216 @@
...
+	 * to fail the test.
+	 */
+	private void waitForConfigUpdate(WatchService watchService) throws InterruptedException {
+		int attempts = 10;
+		while (--attempts >= 0) {
+			WatchKey key = watchService.poll(100, TimeUnit.MILLISECONDS);
+
+			if (key != null) {
+				Optional<Path> configFilePath = key.pollEvents().stream().filter(
+						event -> event.kind() == StandardWatchEventKinds.ENTRY_MODIFY)
```
artembilan: I believe we don't need filter here since we are subscribing only for the `StandardWatchEventKinds.ENTRY_MODIFY`

spring-cloud-gcp-pubsub-stream-binder/src/test/java/org/springframework/cloud/gcp/stream/binder/pubsub/PubSubEmulator.java
```diff
@@ -0,0 +1,216 @@
...
+	 *
+	 * @param emulatorHostPort typically "localhost:8085"
+	 */
+	private void extractTeardownParams(String emulatorHostPort) {
+		int portSeparatorIndex = emulatorHostPort.indexOf(":");
+		if (portSeparatorIndex < 0 || !emulatorHostPort.contains("localhost")) {
+			throw new RuntimeException("Malformed host, can't create emulator: " + emulatorHostPort);
+		}
+		this.emulatorHost = emulatorHostPort.substring(0, portSeparatorIndex);
+		this.emulatorPort = emulatorHostPort.substring(portSeparatorIndex + 1);
```
artembilan: Why can't we just use `split(":")` here and check for the size of the result array?

spring-cloud-gcp-pubsub-stream-binder/src/test/java/org/springframework/cloud/gcp/stream/binder/pubsub/PubSubMessageChannelBinderTests.java
```diff
@@ -25,34 +26,21 @@
...
 import org.springframework.cloud.stream.binder.Spy;
 
-import static org.assertj.core.api.Assumptions.assumeThat;
-
 /**
- * Integration tests that require the Pub/Sub emulator to be installed and only run if the
- * GCP_PUBSUB_EMULATOR envvar is present.
+ * Integration tests that require the Pub/Sub emulator to be installed.
  *
  * @author João André Martins
```
artembilan: Your `@author` name here, Elena.
And in the `PubSubEmulator` class with an appropriate `@since 1.1`

spring-cloud-gcp-pubsub-stream-binder/src/test/java/org/springframework/cloud/gcp/stream/binder/pubsub/PubSubEmulator.java
meltsufin: Good point! I think this rule should only provide the `host` of the emulator that the binder tests can use to create the `PubSubTestBinder`.

elefeint:
> General comment.
> @elefeint , please, consider in the future to raise PR against top-level branches and in the `upsrtream`, not `origin`.
> This way we will be able to push commits for simple polishing from our side.
> And definitely don't raise PRs against `master`. Now you will need to `rest hard` your local and origin masters.
> 
> Hope I am clear.



> ```
> Tests in error:
>   org.springframework.cloud.gcp.stream.binder.pubsub.PubSubMessageChannelBinderTests: Cannot run program "gcloud": CreateProcess error=2, Не удается найти указанный файл
> ```
> Indeed I just don't have any `gcloud` software on my machine. I believe in this case the tests should be just skipped.
> 
> Thoughts?
@artembilan 
Ah, I had assumed, incorrectly, that cloud tools was a prerequisite. Thanks for catching that!
Mike suggested introducing a new gating flag for light integration tests that should always run on Travis and could optionally be run locally.

elefeint:
> General comment.
> @elefeint , please, consider in the future to raise PR against top-level branches and in the `upsrtream`, not `origin`.
> This way we will be able to push commits for simple polishing from our side.
> And definitely don't raise PRs against `master`. Now you will need to `rest hard` your local and origin masters.
> 
> Hope I am clear.
@artembilan @meltsufin 
Would it be better to create a branch off my personal fork, or to create a branch off spring-cloud/spring-cloud-gcp?

artembilan:
You create branch locally, but push it not to `origin`, but `upstream`.

Anyway you are fine now. Please, come back to us if you have a problems with resetting your local `master`.
Nevertheless, that's later already after merging this one.

elefeint:
@artembilan @meltsufin PTAL?

spring-cloud-gcp-pubsub-stream-binder/src/test/java/org/springframework/cloud/gcp/stream/binder/pubsub/PubSubMessageChannelBinderTests.java
```diff
@@ -25,34 +26,30 @@
...
-	public PubSubMessageChannelBinderTests() {
-		this.binder = new PubSubTestBinder(System.getenv(EMULATOR_HOST_ENVVAR_NAME));
-	}
+	@ClassRule
+	public static PubSubEmulator emulator = new PubSubEmulator();
 
 	@BeforeClass
 	public static void enableTests() {
-		assumeThat(System.getenv(EMULATOR_HOST_ENVVAR_NAME)).isNotNull();
+		assumeTrue(
```
artembilan: This still doesn't work for me since the `@ClassRule` is applied first.

I think it doesn't matter if we are in `it` mode or not, since there is just no `gcloud` software.

Can we be protected against that?

Thanks

spring-cloud-gcp-pubsub-stream-binder/src/test/java/org/springframework/cloud/gcp/stream/binder/pubsub/PubSubMessageChannelBinderTests.java
```diff
@@ -25,34 +26,30 @@
...
+				"true".equals(System.getProperty("it.pubsub")) || "true".equals(System.getProperty("it.localdeps")));
 	}
 
 	@Override
 	protected PubSubTestBinder getBinder() throws Exception {
-		return this.binder;
+		return new PubSubTestBinder(this.emulator.getEmulatorHostPort());
```
artembilan: No `this.` on `static` members.

The IDEA shows me that as a warning yellowed code.

spring-cloud-gcp-pubsub-stream-binder/src/test/java/org/springframework/cloud/gcp/stream/binder/pubsub/PubSubMessageChannelBinderTests.java
```diff
@@ -43,13 +44,15 @@
 
 	@BeforeClass
 	public static void enableTests() {
-		assumeTrue(
+		assumeTrue("PubSub Binder tests are disabled. Please enable them with -Dit.pubsub or -Dit.localdeps",
 				"true".equals(System.getProperty("it.pubsub")) || "true".equals(System.getProperty("it.localdeps")));
+		assumeThat(emulator.getEmulatorHostPort())
+				.withFailMessage("No PubSub emulator host configured").isNotNull();
```
artembilan: Now it works for me on Windows without any `gcloud`.
Nevertheless I believe we do this assumptions in the `PubSubEmulator.before()`.
It's not going to be to useful if we need to copy/paste this `enableTests()` everywhere we would like to use a `PubSubEmulator`

spring-cloud-gcp-pubsub-stream-binder/src/test/java/org/springframework/cloud/gcp/stream/binder/pubsub/PubSubEmulator.java
```diff
@@ -0,0 +1,216 @@
...
+	 * to fail the test.
+	 */
+	private void waitForConfigCreation() throws InterruptedException {
+		int attempts = 10;
+		while (!Files.exists(EMULATOR_CONFIG_PATH) && --attempts >= 0) {
+			Thread.sleep(1000);
+		}
+		if (attempts < 0) {
+			throw new RuntimeException(
+					"Emulator could not be configured due to missing env.yaml. Are PubSub and beta tools installed?");
```
elefeint: @artembilan @meltsufin 
Philosophical question -- should this rule never fail, but instead disable the test for any misconfiguration?

Specifically here: if cloud SDK is installed, but cloud beta tools or the pubsub emulator are not installed, should the tests get skipped? Or is the current behavior of failing and telling the user to install beta tools/emulator correct?

artembilan: Well, logically it is not our tests responsibilities that outside software is not configured properly.
We are responsible for the `host/port` only and should fail if there is really no service listening to us.
All other failures of external configuration should only WARN and skip tests.

That's my feeling...

spring-cloud-gcp-pubsub-stream-binder/src/test/java/org/springframework/cloud/gcp/stream/binder/pubsub/PubSubMessageChannelBinderTests.java
```diff
@@ -44,14 +44,15 @@
...
-		assumeThat(emulator.getEmulatorHostPort())
-				.withFailMessage("No PubSub emulator host configured").isNotNull();
+		assumeTrue("PubSub Binder tests are disabled. Please enable them with -Dit.pubsub or -Dit.emulator",
+				"true".equals(System.getProperty("it.pubsub")) || "true".equals(System.getProperty("it.emulator")));
 	}
 
 	@Override
 	protected PubSubTestBinder getBinder() {
+		if (emulator.getEmulatorHostPort() == null) {
+			fail("No PubSub emulator host found.");
```
artembilan: I think the assertion is not a responsibility of this method.
More over if we go false positive assumption in the `before()`,. we never step into this method.

spring-cloud-gcp-pubsub-stream-binder/src/test/java/org/springframework/cloud/gcp/stream/binder/pubsub/PubSubMessageChannelBinderTests.java
```diff
@@ -44,14 +44,15 @@
 
 	@BeforeClass
 	public static void enableTests() {
-		assumeTrue("PubSub Binder tests are disabled. Please enable them with -Dit.pubsub or -Dit.localdeps",
-				"true".equals(System.getProperty("it.pubsub")) || "true".equals(System.getProperty("it.localdeps")));
-		assumeThat(emulator.getEmulatorHostPort())
-				.withFailMessage("No PubSub emulator host configured").isNotNull();
+		assumeTrue("PubSub Binder tests are disabled. Please enable them with -Dit.pubsub or -Dit.emulator",
```
artembilan: I still don't see reason in overhead with this method. 
i just don't want to copy/paste always together with the `PubSubEmulator` rule.
All the logic simply can be done in the rule generic way.
And it already doesn't matter what consumer test-case is for this rule.

spring-cloud-gcp-pubsub-stream-binder/src/test/java/org/springframework/cloud/gcp/stream/binder/pubsub/PubSubMessageChannelBinderTests.java
elefeint: I agree that it's not a getter method's responsibility to check preconditions, but it's a convenient way to force all tests to fail when we either run on Travis CI or locally with a flag set.

In my mind, there are three scenarios:
1) test runs on Travis -- in this case, we want any environment issues to trigger test failure, breaking the build. Otherwise we can have tests skipping silently if, for example, emulator behavior changes. This behavior is achieved by having it.pubsub/it.emulator turned on by default on Travis.
2) running tests with no flags, and without cloud tools installed -- in this case, the binder tests should be skipped, since they cannot possibly run. This is your typical workflow, and it's addressed by the gating flags it.pubsub and it.emulator being off by default. Some of the emulator rule setup logic attempts to run, but only until the emulator rule realizes the environment is bad.
3) running tests locally with cloud tools installed -- this is my workflow, and I can just enable the it.emulator flag to achieve it.

So, not getting to the emulator check in getBinder() is intentional for scenario 2).

spring-cloud-gcp-pubsub-stream-binder/src/test/java/org/springframework/cloud/gcp/stream/binder/pubsub/PubSubMessageChannelBinderTests.java
elefeint: Do you mean that the gating flag could be consumed by PubSubEmulator rule itself?

artembilan: Correct. Otherwise the rule just does not make sense. I feel them as together: if we don't set flag, we don't go to check any other options.
More over as you noticed the `@ClassRule` works before this local `@ClassRule`, but that's not we would like to have when we don't set the system flag.

Hope this facilitates a bit my other concern about `getBinder()`  body.

spring-cloud-gcp-pubsub-stream-binder/src/test/java/org/springframework/cloud/gcp/stream/binder/pubsub/PubSubMessageChannelBinderTests.java
```diff
@@ -42,17 +47,26 @@
 	@ClassRule
 	public static PubSubEmulator emulator = new PubSubEmulator();
 
+	@Rule
+	public TestRule failOnAbsentEmulator = new TestRule() {
```
artembilan: Why do we need this extra rule?
What if I will forget to copy/paste it everywhere?
Why just single `PubSubEmulator` can't serve us all the requirements?

spring-cloud-gcp-pubsub-stream-binder/src/test/java/org/springframework/cloud/gcp/stream/binder/pubsub/PubSubMessageChannelBinderTests.java
```diff
@@ -42,17 +47,26 @@
...
+			}
+			return statement;
+		}
+	};
+
 	@BeforeClass
 	public static void enableTests() {
-		assumeTrue("PubSub Binder tests are disabled. Please enable them with -Dit.pubsub or -Dit.emulator",
-				"true".equals(System.getProperty("it.pubsub")) || "true".equals(System.getProperty("it.emulator")));
+		assumeThat(System.getProperty("it.pubsub"))
```
artembilan: Why do we still have this `assumeThat()` here?
Why it doesn't work from the `PubSubEmulator.before()`?

Doesn't look like `PubSubEmulator` is going to help us enough: still too much boilerplate code to copy/paste in each test where I would like to use a Pub/Sub emulator...

spring-cloud-gcp-pubsub-stream-binder/src/test/java/org/springframework/cloud/gcp/stream/binder/pubsub/PubSubMessageChannelBinderTests.java
elefeint: The extra rule is there to provide a user-friendly error message. Otherwise you end up with whatever error the binder chooses to serve when faced with a null constructor parameter.
PubSubEmulator emits a warning, but it's hard to find if it's clobbered by a hundred lines of unrelated error messages.

spring-cloud-gcp-pubsub-stream-binder/src/test/java/org/springframework/cloud/gcp/stream/binder/pubsub/PubSubMessageChannelBinderTests.java
elefeint: Mike suggested doing away with skipping tests through failed assumptions, and to use Maven profiles instead. What do you think? That would make both the assume* statements and my new rule unnecessary, so it would reduce the boilerplate.

elefeint: Actually, I think I know what you mean. The PubSubEmulator rule can force all tests to be skipped or to be failed based on the value of it.emulator, right?

artembilan: Yes. That's what I mean. Sorry for my pure English.

As a sample we have in Spring AMQP this one: https://github.com/spring-projects/spring-amqp/blob/master/spring-rabbit-junit/src/main/java/org/springframework/amqp/rabbit/junit/BrokerRunning.java#L82

And here is its usage: https://github.com/spring-projects/spring-amqp/blob/master/spring-rabbit/src/test/java/org/springframework/amqp/rabbit/listener/ContainerShutDownTests.java#L44

So, our AMQP tests go if broker is available or they are skipped otherwise.

spring-cloud-gcp-pubsub-stream-binder/src/test/java/org/springframework/cloud/gcp/stream/binder/pubsub/PubSubEmulator.java
```diff
@@ -59,29 +62,34 @@
...
 
 	public PubSubEmulator() {
-		if (!"true".equals(System.getProperty("it.emulator"))) {
-			LOGGER.warn("PubSubEmulator rule disabled. Please enable with -Dit.emulator");
-			this.ruleDisabled = true;
+		if ("true".equals(System.getProperty("it.pubsub-emulator"))) {
+			this.enableTests = true;
+		}
+		else {
+			LOGGER.warn("PubSubEmulator rule disabled. Please enable with -Dit.pubsub-emulator");
```
meltsufin: `-Dit.pubsub-emulator=true` right?

elefeint: Maven will default properties to "true" by default, so either way, I think.
https://github.com/apache/maven/blob/864d180edf7b66b07e2e3fc08709b5cb1e4f051c/maven-embedder/src/main/java/org/apache/maven/cli/MavenCli.java#L1690

meltsufin: Ok, cool!

elefeint:
@artembilan We are clearly moving into the nits territory, so this is actually really encouraging 😂

spring-cloud-gcp-pubsub-stream-binder/src/test/java/org/springframework/cloud/gcp/stream/binder/pubsub/PubSubEmulator.java
```diff
@@ -0,0 +1,251 @@
...
+	 *
+	 * If the property is on, any setup failure will trigger test failure. Failures during teardown are merely logged.
+	 *
+	 * @throws IOException if config file creation or directory watcher on existing file fails.
+	 * @throws InterruptedException if process is stopped while waiting to retry.
+	 */
+	@Override
+	protected void before() throws IOException, InterruptedException {
+
+		assumeTrue(this.enableTests);
```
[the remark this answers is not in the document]
artembilan: OK. Let's leave that WARN in the ctor and add the message here!
As you see my code snippet, it works for me in IDEA. This is enough for false positive conclusion. 
I mean that we do this in many other projects with success and nobody complained.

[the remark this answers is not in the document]
elefeint: Done. 
I guess IDEA has its own custom runner -- I see the message there, but it's completely missing when running tests from command line.

spring-cloud-gcp-pubsub-stream-binder/src/test/java/org/springframework/cloud/gcp/stream/binder/pubsub/PubSubEmulator.java
```diff
@@ -0,0 +1,239 @@
...
+	public String getEmulatorHostPort() {
+		return this.emulatorHostPort;
+	}
+
+	private void startEmulator() throws IOException, InterruptedException {
+		boolean configPresent = Files.exists(EMULATOR_CONFIG_PATH);
+		WatchService watchService = null;
+
+		if (configPresent) {
+			watchService = FileSystems.getDefault().newWatchService();
```
artembilan: Don't we need to `close()` the `WatchService` in the end of our business?

spring-cloud-gcp-pubsub-stream-binder/src/test/java/org/springframework/cloud/gcp/stream/binder/pubsub/PubSubEmulator.java
```diff
@@ -0,0 +1,239 @@
...
+			String emulatorHost = this.emulatorHostPort.substring(0, portSeparatorIndex);
+			String emulatorPort = this.emulatorHostPort.substring(portSeparatorIndex + 1);
+
+			String hostPortParams = String.format("--host=%s --port=%s", emulatorHost, emulatorPort);
+			Process psProcess = new ProcessBuilder("ps", "-v").start();
+			new BufferedReader(new InputStreamReader(psProcess.getInputStream()))
+					.lines()
+					.filter(psLine -> psLine.contains(hostPortParams))
+					.map(psLine -> new StringTokenizer(psLine).nextToken())
+					.forEach(this::killProcess);
```
artembilan: Don't we need to close the `Stream` after using it?
I believe it is going to close all the `BufferedReader` and `InputStream` it is based on...

spring-cloud-gcp-pubsub-stream-binder/src/test/java/org/springframework/cloud/gcp/stream/binder/pubsub/PubSubEmulator.java
elefeint: Yes, thanks for catching that!

PR #1020, opened 2018-09-13
Datastore key queries for count and exists query methods

### LINKED ISSUE #995: Support querying for BaseEntity or ProjectionEntity in DatastoreTemplate
Currently we're not supporting projection queries (`ProjectionEntity`) at all in Datastore via the template or data repositories.

ChengyuanZhao:
fixes #995

### DIFF
--- a/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreOperations.java
+++ b/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreOperations.java
@@ -18,7 +18,8 @@ package org.springframework.cloud.gcp.data.datastore.core;
 
 import java.util.function.Function;
 
-import com.google.cloud.datastore.Entity;
+import com.google.cloud.datastore.Key;
+import com.google.cloud.datastore.KeyQuery;
 import com.google.cloud.datastore.Query;
 
 /**
@@ -131,7 +132,14 @@ public interface DatastoreOperations {
 	 * @return a list of the objects found. If no keys could be found the list will be
 	 * empty.
 	 */
-	<T> Iterable<T> query(Query<Entity> query, Class<T> entityClass);
+	<T> Iterable<T> query(Query query, Class<T> entityClass);
+
+	/**
+	 * Finds Cloud Datastore Keys by using a Cloud Datastore query.
+	 * @param query the query to execute that retrieves only Keys.
+	 * @return the list of keys found.
+	 */
+	Iterable<Key> queryKeys(KeyQuery query);
 
 	/**
 	 * Get all the entities of the given domain type.
--- a/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreOperations.java
+++ b/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreOperations.java
@@ -18,8 +18,8 @@ package org.springframework.cloud.gcp.data.datastore.core;
 
 import java.util.function.Function;
 
+import com.google.cloud.datastore.BaseEntity;
 import com.google.cloud.datastore.Key;
-import com.google.cloud.datastore.KeyQuery;
 import com.google.cloud.datastore.Query;
 
 /**
@@ -132,14 +132,14 @@ public interface DatastoreOperations {
 	 * @return a list of the objects found. If no keys could be found the list will be
 	 * empty.
 	 */
-	<T> Iterable<T> query(Query query, Class<T> entityClass);
+	<T> Iterable<T> query(Query<? extends BaseEntity> query, Class<T> entityClass);
 
 	/**
 	 * Finds Cloud Datastore Keys by using a Cloud Datastore query.
 	 * @param query the query to execute that retrieves only Keys.
 	 * @return the list of keys found.
 	 */
-	Iterable<Key> queryKeys(KeyQuery query);
+	Iterable<Key> queryKeys(Query<Key> query);
 
 	/**
 	 * Get all the entities of the given domain type.
--- a/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/PartTreeDatastoreQuery.java
+++ b/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/PartTreeDatastoreQuery.java
@@ -131,7 +131,7 @@ public class PartTreeDatastoreQuery<T> extends AbstractDatastoreQuery<T> {
 	private StructuredQuery getQuery(Object[] parameters) {
 		StructuredQuery.Builder structredQueryBuilder;
 
-		if (this.tree.isExistsProjection() && this.tree.isCountProjection()) {
+		if (this.tree.isExistsProjection() || this.tree.isCountProjection()) {
 			structredQueryBuilder = StructuredQuery.newKeyQueryBuilder();
 		}
 		else if (!this.runAsProjectionQuery && !this.tree.isDistinct()) {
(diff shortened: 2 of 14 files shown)


spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/PartTreeDatastoreQuery.java
```diff
@@ -203,10 +224,26 @@ private Filter getFilter(Object[] parameters) {
 								+ this.queryMethod.getName());
 			}
 		}).toArray(Filter[]::new);
-		return filters.length > 1
+
+		if (builder instanceof ProjectionEntityQuery.Builder) {
```
ChengyuanZhao: somehow need to insert the equality-compared values into the entities coming back from Datastore for conversion

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/Query.java
```diff
@@ -44,4 +44,12 @@
...
 	 */
 	String value() default "";
+
+	/**
+	 * Indicates if the annotated query method will utilize a Cloud Datastore Projection
+	 * query that is subject to the limitations of Projection queries.
+	 * @return {@code true} if the query method is intended to be run as a Projection
+	 * query. {@code false} otherwise.
+	 */
+	boolean runAsProjectionQuery() default false;
```
ChengyuanZhao: do a separate annotation RunAsProjection instead.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreOperations.java
```diff
@@ -131,7 +132,14 @@
 	 * @return a list of the objects found. If no keys could be found the list will be
 	 * empty.
 	 */
-	<T> Iterable<T> query(Query<Entity> query, Class<T> entityClass);
+	<T> Iterable<T> query(Query query, Class<T> entityClass);
```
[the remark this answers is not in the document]
ChengyuanZhao: We cannot be sure if the `query` is a key-query or that it returns entities. Technically it would need to be a `StructredQuery<Key>` , but we cannot check the inner type of the StructuredQuery to make sure. (note that `KeyQuery` mentioned in the next method `queryKeys` is a child type of `StructuredQuery<Key>`, but it's not necessarily the only child). 

Even if we could check that inner type, we would still be requiring the user to make sure he has the right type of query AND the right type of entityClass. I think in that case he might as well just use `queryKeys` where he doesn't need to provide the entityclass at all.

ChengyuanZhao:
@meltsufin PTAL. I removed the projection entity query for name-based methods but kept the key-queries for count and exists. added the ? extends BaseEntity where relevant.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/PartTreeDatastoreQuery.java
```diff
@@ -65,6 +68,8 @@
 	 * @param datastoreMappingContext used to provide metadata for mapping results to
 	 * objects.
 	 * @param entityType the result domain type.
+	 * Projection query. Regardless of this setting the method will be treated as a Projection query
```
meltsufin: No longer applies.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/PartTreeDatastoreQuery.java
```diff
@@ -203,10 +210,29 @@ private Filter getFilter(Object[] parameters) {
...
+
+		builder.setFilter(
+				filters.length > 1
 				? CompositeFilter.and(filters[0],
 						Arrays.copyOfRange(filters, 1, filters.length))
-				: filters[0];
+				: filters[0]);
 	}
 
+	private <A> A ifCountExistsQuery(Supplier<A> ifCount, Supplier<A> ifExists,
```
meltsufin: This is an odd code style. Why not have a simple method `boolean isCountOrExistsQuery`?

ChengyuanZhao: Then you can't do the 3-case with count, exists, or neither separately.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/PartTreeDatastoreQuery.java
ChengyuanZhao: fixed. PR title also changed.

ChengyuanZhao: @meltsufin  PTAL

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/PartTreeDatastoreQuery.java
```diff
@@ -103,36 +108,36 @@ else if (this.tree.isDistinct()) {
...
+		List results = executeRawResult(parameters);
+		return ifCountExistsQuery(results::size, () -> !results.isEmpty(),
+				() -> applyProjection(results));
 	}
 
 	@Override
-	protected List<T> executeRawResult(Object[] parameters) {
-		Iterable<T> found = this.datastoreOperations.query(getQuery(parameters),
-				this.entityType);
+	protected List executeRawResult(Object[] parameters) {
```
meltsufin: I'm not convinced you need this method at all. Why can't this logic just move into the method above?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/PartTreeDatastoreQuery.java
ChengyuanZhao: Hmmm, i thought I deleted those lines. I must've not pushed that commit. will fix.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/PartTreeDatastoreQuery.java
```diff
@@ -103,36 +108,36 @@ else if (this.tree.isDistinct()) {
...
-							+ "Please explicitly use a find-query and examine the result size.");
-		}
-		else if (this.tree.isExistsProjection()) {
-			return !results.isEmpty();
-		}
-		else {
-			return applyProjection(results);
-		}
+		List results = executeRawResult(parameters);
+		return ifCountExistsQuery(results::size, () -> !results.isEmpty(),
```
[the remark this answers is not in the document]
ChengyuanZhao: Then we lose having this logic in one place that we discussed in person. Thank you for the opinion, but I prefer the current way where the logic is just in one place.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/PartTreeDatastoreQuery.java
ChengyuanZhao: The reason is in this PR we are still not supporting count/exists queries for the query-string annotated query methods. There is then a difference in the `execute` method between part-tree and gql query method implementations. 

We have a separate issue for supporting count/exists queries, and I expect when implementing that one that executeRawResult and execute can be combined since there won't be a difference between GQL and part-tree query methods.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/PartTreeDatastoreQuery.java
```diff
@@ -103,36 +106,36 @@ else if (this.tree.isDistinct()) {
 
 	@Override
 	public Object execute(Object[] parameters) {
```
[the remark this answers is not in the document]
ChengyuanZhao: @meltsufin PTAL.

Issue #991, opened 2018-08-31
v1.0.0.RELEASE samples are referring to SNAPSHOT parents


saturnism:
See https://github.com/spring-cloud/spring-cloud-gcp/blob/v1.0.0.RELEASE/spring-cloud-gcp-samples/pom.xml and also the samples in that directory.

They are referring to snapshot parents, where it should be the release version.
```
	<parent>
		<artifactId>spring-cloud-gcp</artifactId>
		<groupId>org.springframework.cloud</groupId>
		<version>1.0.0.BUILD-SNAPSHOT</version>
	</parent>
```

saturnism:
@meltsufin i updated the branch and tested it out. looks good :)

meltsufin:
@saturnism It looks like you missed 2. Take a look here: 3232934183c94eadd0c35b7d56b58ba4fa6e7c35.

saturnism:
good catch!

meltsufin:
`v1.0.0.RELEASE` has been amended. Closing.

PR #1021, opened 2018-09-13
Improve releasing instructions

meltsufin:
Change order of steps to make sure that modules not being released
are also included in the version change.

Specify which commands to run to change and revert versions.

Related to #991.

### DIFF
--- a/RELEASING.adoc
+++ b/RELEASING.adoc
@@ -15,7 +15,14 @@ In order to simplify that process, here is a list of prerequisites to be perform
 
 . [For GA releases only.] Comment out any non-GA modules from the `pom.xml` file.
 
-. After the release scripts have been run by the release admin, uncomment the commented out modules.
+. Wait for the release scripts have been run by the release admin.
+
+. Add a release tag.
+
+    $ git tag v$NEW_VERSION
+    $ git push origin v$NEW_VERSION
+
+. Uncomment the commented out modules.
 
 . Revert the versions using the https://www.mojohaus.org/versions-maven-plugin/revert-mojo.html[mvn versions:revert] command.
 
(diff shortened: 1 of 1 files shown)


RELEASING.adoc
```diff
@@ -3,12 +3,29 @@
...
 
-1. Comment out the `spring-cloud-gcp-samples` module from the root `pom.xml` file, so that the samples aren't released.
+. Manually change the value of `TRACKING_HEADER_PROJECT_VERSION` in link:spring-cloud-gcp-core/src/main/java/org/springframework/cloud/gcp/core/UsageTrackingHeaderProvider.java[`UsageTrackingHeaderProvider`] to the version being released from the interim `BUILD-SNAPSHOT` value.
 
-2. Comment out any non-GA modules from the `pom.xml` file.
+. Change the value of the `<version>` elements of the `pom.xml` of all modules to the name of the version being released from the interim `BUILD-SNAPSHOT` values. This can be done using the https://www.mojohaus.org/versions-maven-plugin/set-mojo.html[mvn versions:set] command.
 
-3. Manually change the value of `TRACKING_HEADER_PROJECT_VERSION` in link:spring-cloud-gcp-core/src/main/java/org/springframework/cloud/gcp/core/UsageTrackingHeaderProvider.java[`UsageTrackingHeaderProvider`] to the version being released from the interim `BUILD-SNAPSHOT` value.
+    $ NEW_VERSION=[new-version]
+    $ ./mvnw versions:set -pl spring-cloud-gcp-dependencies -DnewVersion=$NEW_VERSION
```
meltsufin: I don't know why, but the `spring-cloud-gcp-dependencies` module doesn't get processed by `versions:set` if run from the parent. So, we have to do it in a separate step.
Interestingly, `versions:revert` works fine for all modules when run from the parent.

PR #1033, opened 2018-09-18
Spanner count exist query for annotated query methods

### LINKED ISSUE #1030: Spanner counts exists query for annotated query methods
It is assumed the user gives a useable query string and we determine if it is counts or exists based on integer or boolean return type for the query method.

### LINKED ISSUE #1058: Spanner convenience query method for top1 element singularly

ChengyuanZhao:
fixes #1030 
fixes #1058

### DIFF
--- a/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerOperations.java
+++ b/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerOperations.java
@@ -24,6 +24,7 @@ import java.util.function.Function;
 import com.google.cloud.spanner.Key;
 import com.google.cloud.spanner.KeySet;
 import com.google.cloud.spanner.Statement;
+import com.google.cloud.spanner.Struct;
 
 /**
  * Defines operations available to use with Spanner.
@@ -76,6 +77,21 @@ public interface SpannerOperations {
 	 * could be found the list will be empty.
 	 */
 	<T> List<T> read(Class<T> entityClass, KeySet keys);
+
+	/**
+	 * Executes a given query string with tags and parameters and applies a given function
+	 * to each row of the result.
+	 * @param rowFunc the function to apply to each row of the result.
+	 * @param entityClass the underlying domain type whose properties are used in sorts in the result.
+	 * @param sql the query string to run.
+	 * @param tags the tags that appear in the query.
+	 * @param params the value to bind to those tags.
+	 * @param options the options with which to run this query.
+	 * @param <T> the type to transform each row into.
+	 * @return a list of the rows each transformed with the given function.
+	 */
+	<A,T> List<A> query(Function<Struct, A> rowFunc, Class<T> entityClass, String sql, List<String> tags,
+			Object[] params, SpannerQueryOptions options);
 
 	/**
 	 * Finds objects by using an SQL statement.
--- a/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerOperations.java
+++ b/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerOperations.java
@@ -82,7 +82,8 @@ public interface SpannerOperations {
 	 * Executes a given query string with tags and parameters and applies a given function
 	 * to each row of the result.
 	 * @param rowFunc the function to apply to each row of the result.
-	 * @param entityClass the underlying domain type whose properties are used in sorts in the result.
+	 * @param entityClass the underlying domain type whose properties are used in sorts in
+	 * the result.
 	 * @param sql the query string to run.
 	 * @param tags the tags that appear in the query.
 	 * @param params the value to bind to those tags.
@@ -90,13 +91,14 @@ public interface SpannerOperations {
 	 * @param <T> the type to transform each row into.
 	 * @return a list of the rows each transformed with the given function.
 	 */
-	<A,T> List<A> query(Function<Struct, A> rowFunc, Class<T> entityClass, String sql, List<String> tags,
-			Object[] params, SpannerQueryOptions options);
+	<A, T> List<A> query(Function<Struct, A> rowFunc, Class<T> entityClass, String sql,
+			List<String> tags, Object[] params, SpannerQueryOptions options);
 
 	/**
 	 * Finds objects by using an SQL statement.
 	 * @param entityClass the type of object to retrieve.
-	 * @param sql the SQL string to execute. this string can have Cloud Spanner param tags.
+	 * @param sql the SQL string to execute. this string can have Cloud Spanner param
+	 * tags.
 	 * @param tags the names of the tags to use
 	 * @param params the values to attach those tags, in the same order.
 	 * @param options Cloud Spanner read options with which to conduct the read operation.
@@ -105,8 +107,7 @@ public interface SpannerOperations {
 	 * empty.
 	 */
 	<T> List<T> query(Class<T> entityClass, String sql, List<String> tags,
-			Object[] params,
-			SpannerQueryOptions options);
+			Object[] params, SpannerQueryOptions options);
 
 	/**
 	 * Finds objects by using an SQL statement.
@@ -140,7 +141,8 @@ public interface SpannerOperations {
 	/**
 	 * Finds all objects of the given type.
 	 * @param entityClass the type of the object to retrieve.
-	 * @param options Cloud Spanner query options with which to conduct the query operation.
+	 * @param options Cloud Spanner query options with which to conduct the query
+	 * operation.
 	 * @param <T> the type of the object to retrieve.
 	 * @return a list of all objects stored of the given type. If there are no objects an
 	 * empty list is returned.
@@ -208,9 +210,9 @@ public interface SpannerOperations {
 	/**
 	 * Update an object in storage.
 	 * @param object the object to update.
-	 * @param includeColumns the columns to update. If null or an empty Optional is given, then
-	 * all columns are used. Note that an Optional occupied by an empty Set means that no columns
-	 * will be used.
+	 * @param includeColumns the columns to update. If null or an empty Optional is given,
+	 * then all columns are used. Note that an Optional occupied by an empty Set means
+	 * that no columns will be used.
 	 */
 	void update(Object object, Optional<Set<String>> includeColumns);
 
@@ -237,9 +239,9 @@ public interface SpannerOperations {
 	/**
 	 * Update or insert an object into storage.
 	 * @param object the object to update or insert.
-	 * @param includeColumns the columns to upsert. If null or an empty Optional is given, then
-	 * all columns are used. Note that an Optional occupied by an empty Set means that no columns
-	 * will be used.
+	 * @param includeColumns the columns to upsert. If null or an empty Optional is given,
+	 * then all columns are used. Note that an Optional occupied by an empty Set means
+	 * that no columns will be used.
 	 */
 	void upsert(Object object, Optional<Set<String>> includeColumns);
 
--- a/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerOperations.java
+++ b/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerOperations.java
@@ -82,42 +82,24 @@ public interface SpannerOperations {
 	 * Executes a given query string with tags and parameters and applies a given function
 	 * to each row of the result.
 	 * @param rowFunc the function to apply to each row of the result.
-	 * @param entityClass the underlying domain type whose properties are used in sorts in
-	 * the result.
-	 * @param sql the query string to run.
-	 * @param tags the tags that appear in the query.
-	 * @param params the value to bind to those tags.
+	 * @param statement the SQL statement used to select the objects.
 	 * @param options the options with which to run this query.
-	 * @param <T> the type to transform each row into.
 	 * @return a list of the rows each transformed with the given function.
 	 */
-	<A, T> List<A> query(Function<Struct, A> rowFunc, Class<T> entityClass, String sql,
-			List<String> tags, Object[] params, SpannerQueryOptions options);
-
-	/**
-	 * Finds objects by using an SQL statement.
-	 * @param entityClass the type of object to retrieve.
-	 * @param sql the SQL string to execute. this string can have Cloud Spanner param
-	 * tags.
-	 * @param tags the names of the tags to use
-	 * @param params the values to attach those tags, in the same order.
-	 * @param options Cloud Spanner read options with which to conduct the read operation.
-	 * @param <T> the type of object to retrieve.
-	 * @return a list of the objects found. If no keys could be found the list will be
-	 * empty.
-	 */
-	<T> List<T> query(Class<T> entityClass, String sql, List<String> tags,
-			Object[] params, SpannerQueryOptions options);
+	<A> List<A> query(Function<Struct, A> rowFunc, Statement statement,
+			SpannerQueryOptions options);
 
 	/**
 	 * Finds objects by using an SQL statement.
 	 * @param entityClass the type of object to retrieve.
 	 * @param statement the SQL statement used to select the objects.
+	 * @param options Cloud Spanner read options with which to conduct the read operation.
 	 * @param <T> the type of object to retrieve.
 	 * @return a list of the objects found. If no keys could be found the list will be
 	 * empty.
 	 */
-	<T> List<T> query(Class<T> entityClass, Statement statement);
+	<T> List<T> query(Class<T> entityClass, Statement statement,
+			SpannerQueryOptions options);
 
 	/**
 	 * Finds all objects of the given type.
@@ -147,7 +129,7 @@ public interface SpannerOperations {
 	 * @return a list of all objects stored of the given type. If there are no objects an
 	 * empty list is returned.
 	 */
-	<T> List<T> queryAll(Class<T> entityClass, SpannerQueryOptions options);
+	<T> List<T> queryAll(Class<T> entityClass, SpannerSortPageQueryOptions options);
 
 	/**
 	 * Deletes an object based on a key.
--- a/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerOperations.java
+++ b/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerOperations.java
@@ -129,7 +129,7 @@ public interface SpannerOperations {
 	 * @return a list of all objects stored of the given type. If there are no objects an
 	 * empty list is returned.
 	 */
-	<T> List<T> queryAll(Class<T> entityClass, SpannerSortPageQueryOptions options);
+	<T> List<T> queryAll(Class<T> entityClass, SpannerPageableQueryOptions options);
 
 	/**
 	 * Deletes an object based on a key.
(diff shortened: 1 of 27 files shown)


ChengyuanZhao:
@meltsufin PTAL

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerOperations.java
```diff
@@ -77,10 +78,27 @@
...
+	 * @param entityClass the underlying domain type whose properties are used in sorts in
+	 * the result.
+	 * @param sql the query string to run.
+	 * @param tags the tags that appear in the query.
+	 * @param params the value to bind to those tags.
+	 * @param options the options with which to run this query.
+	 * @param <T> the type to transform each row into.
+	 * @return a list of the rows each transformed with the given function.
+	 */
+	<A, T> List<A> query(Function<Struct, A> rowFunc, Class<T> entityClass, String sql,
```
meltsufin: See if you can get rid of the `entityClass` as well as `SpannerQueryOptions`.
We should probably assume that `sql` is already constructed with the options that can go into it.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SqlSpannerQuery.java
```diff
@@ -218,12 +218,30 @@ else if (Pageable.class.isAssignableFrom(paramClass)) {
...
+				: this.spannerOperations.query(this.entityType,
 				resolveEntityClassNames(queryTagValue.sql), queryTagValue.tags,
 				queryTagValue.params.toArray(),
 				spannerQueryOptions);
 	}
 
+	@Override
+	protected boolean isCountQuery() {
+		Class returnedType = this.queryMethod.getReturnedObjectType();
+		return returnedType == long.class || returnedType == int.class
```
meltsufin: Just like in the Datastore case, we should not make this guess, but support mapping `ResultSet` to a single POJO as a general rule.
Can we support something like this:
```java
@Query("select plate from cars")
List<String> getPlates();
```
Does Spring Data in general support this kind of query methods?

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryExecutor.java
```diff
@@ -193,7 +214,13 @@ private static StringBuilder buildSelect(
 		if (tree.isDistinct()) {
 			stringBuilder.append("DISTINCT ");
 		}
-		stringBuilder.append(getColumnsStringForSelect(spannerPersistentEntity) + " ");
+		if (tree.isExistsProjection() || tree.isCountProjection()) {
+			stringBuilder.append("COUNT(1) ");
```
meltsufin: Should we do COUNT(DISTINCT *) instead. Please check other Spring Data implementations.

ChengyuanZhao:
@meltsufin PTAL

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerOperations.java
```diff
@@ -124,12 +123,13 @@
...
 	 * @param entityClass the type of the object to retrieve.
-	 * @param options Cloud Spanner query options with which to conduct the query operation.
+	 * @param options Cloud Spanner query options with which to conduct the query
+	 * operation.
 	 * @param <T> the type of the object to retrieve.
 	 * @return a list of all objects stored of the given type. If there are no objects an
 	 * empty list is returned.
 	 */
-	<T> List<T> queryAll(Class<T> entityClass, SpannerQueryOptions options);
+	<T> List<T> queryAll(Class<T> entityClass, SpannerSortPageQueryOptions options);
```
meltsufin: It's interesting that other query methods take `SpanneryQueryOptions` vs this one `SpannerSortPageQueryOptions`. At the least the javadoc should clarify the difference.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerOperations.java
```diff
@@ -192,9 +192,9 @@
 	/**
 	 * Update an object in storage.
 	 * @param object the object to update.
-	 * @param includeColumns the columns to update. If null or an empty Optional is given, then
-	 * all columns are used. Note that an Optional occupied by an empty Set means that no columns
-	 * will be used.
+	 * @param includeColumns the columns to update. If null or an empty Optional is given,
```
meltsufin: please use `{@code Optional}` and `{@code Set}` for javadoc formatting.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerOperations.java
```diff
@@ -192,9 +192,9 @@
...
+	 * then all columns are used. Note that an Optional occupied by an empty Set means
+	 * that no columns will be used.
 	 */
 	void update(Object object, Optional<Set<String>> includeColumns);
```
meltsufin: Generally `Optional` should not be used as a method parameter. See [discussion on SO](https://stackoverflow.com/questions/31922866/why-should-java-8s-optional-not-be-used-in-arguments).
Just allow it to be `null`.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerSortPageQueryOptions.java
```diff
@@ -0,0 +1,103 @@
...
+import org.springframework.util.Assert;
+
+/**
+ * Encapsulates Cloud Spanner query options with sort and paging.
+ *
+ * @author Chengyuan Zhao
+ *
+ * @since 1.1
+ */
+public class SpannerSortPageQueryOptions extends SpannerQueryOptions {
```
meltsufin: In Spring they use the term "Pageable" to include bot page and sort options.
So, maybe `SpannerPageableQueryOptions`.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
```diff
@@ -142,26 +142,22 @@ public SpannerEntityProcessor getSpannerEntityProcessor() {
...
+		while (resultSet.next()) {
+			result.add(rowFunc.apply(resultSet.getCurrentRowAsStruct()));
 		}
-		return mapToListAndResolveChildren(executeQuery(SpannerStatementQueryExecutor
-				.buildStatementFromSqlWithArgs(finalSql, tags, param -> {
-					Builder builder = Struct.newBuilder();
-					this.spannerEntityProcessor.write(param, builder::set);
-					return builder.build();
-				}, params), options), entityClass, Optional.empty(), allowPartialRead);
+		resultSet.close();
```
meltsufin: Shouldn't this be in a `finally`?

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryExecutor.java
```diff
@@ -182,18 +224,17 @@ public static String getColumnsStringForSelect(
 				.getPersistentProperty(o.getProperty()).getColumnName());
 		buildLimit(tree, stringBuilder);
 
-		stringBuilder.append(";");
-		return Pair.of(stringBuilder.toString(), tags);
+		String selectSql = stringBuilder.toString();
+		return Pair.of(tree.isExistsProjection() || tree.isCountProjection()
+				? "SELECT COUNT(1) FROM (" + selectSql + ")"
```
meltsufin: Didn't we say during the review that this might have to be `COUNT(DISTINCT *)` or something? What was the result of that investigation?

spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/test/domain/TradeRepository.java
```diff
@@ -35,6 +35,14 @@
 
 	int countByAction(String action);
```
meltsufin: We need a test for something like this, as discussed during the review. Or is this just not something that can be easily supported? Is it supported in MongoDB or Cassandra?
```java
@Query("select plate from cars")
List<String> getPlates();
```

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerOperations.java
ChengyuanZhao: Yes I will clarify in java doc. The reason is that shorter-named one is all options that don't depend on the SQL string that will be run. The longer-named one includes paging and sorting options, which are only doable in SQL. Any method that takes a SQL string only takes the shorter-named one, which is what we discussed is most intuitive.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryExecutor.java
ChengyuanZhao: I don't think that works. If you were to say countTop100DistinctByColor that would mean you want to select the top 100 and count how many distinct there are in those 100. This means the answer is 100 or less. 

If you do not wrap a count around the entire statement like I do and instead do what you suggest
`SELECT COUNT(DISTINCT *) FROM TABLE WHERE COLOR = 'red' LIMIT 100` then I think that means you limit the number of final result rows to 100 (the count result), and not what is actually counted.

spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/test/domain/TradeRepository.java
ChengyuanZhao: This is definitely not being supported (and will need to be determined if at all, I have a question noted for the Spring Data team) in this PR. This PR is just about counts and exists queries.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryExecutor.java
meltsufin: I see, so you expect the user to use `DISTINCT` explicitly. OK.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerOperations.java
ChengyuanZhao: made new issue 
https://github.com/spring-cloud/spring-cloud-gcp/issues/1057

spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/test/domain/TradeRepository.java
meltsufin: I bring this up because, theoretically, returning a non-collection should also be possible when the result set is of size 1.
For example.
```java
String findFirstNameByLastname(String lastname)
```
Do we support this?

ChengyuanZhao: No, but I will make an issue for it: 

https://github.com/spring-cloud/spring-cloud-gcp/issues/1058

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerOperations.java
meltsufin: I think we should be consistent across the query methods. It seems like a soon as I want to provide a query string, I can't take advantage of the paging and sorting options. Why? It seems like a useful functionality that is already there.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerSortPageQueryOptions.java
meltsufin: Would it be possible to just get rid of this class and use `org.springframework.data.domain.PageRequest` instead?
You would need to pass `SpannerQueryOptions` and `Pageable` separately in some places, but I think it's a cleaner separation of concerns.

ChengyuanZhao: I don't agree with that actually. I think it's a nice benefit to package them together.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerOperations.java
ChengyuanZhao: Because the SQL string already can already specify the sorting and paging. That is the only way to do so for all of spanner client lib.

meltsufin: I understand the potential conflict, but don't we already allow it via `@Query` + Pageable in data repositories?

ChengyuanZhao: Yes, that still works because I moved that paging-sorting string appending/prepending further up in the process so now the final SQL string arrives at the SpannerTemplate not needing the paging/sorting params that the user provided.

meltsufin: My point is that one can still do:
```java
@Query("select * from person order by last_name")
List<Person> getPeople(Sort sort);
```

ChengyuanZhao: Yes, that still works. That was intentionally supported. The SpannerPageableQueryOptions is only used when using SpannerTemplate. I believe it's a different matter.

meltsufin: If one is only using `SpannerTemplate` how can they include `SpannerPageableQueryOptions` and provide their own SQL string, without `order by`?

ChengyuanZhao: They can't. Intentionally they can't. 

All we're doing is appending an "order by" and "limit" to the string anyway. If the user is going to supply the packaged-up sort/pageable options as a java object then he might as well append it to the SQL himself. I generally think it's  good to manipulating a user's given query string as little as possible.

ChengyuanZhao:
@meltsufin I'm taking a look at `AbstractCassandraQuery` but I think it's too different and even contrary to what we discussed today.. It has explicit `isCountQuery` and `isExistsQuery`, which is not what we discussed today. I'll put a commit soon with what we discussed in person today.

meltsufin:
cc: @mp911de

spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/test/domain/TradeRepository.java
ChengyuanZhao: @meltsufin PTAL. incidentally #1058 is also addressed. Now the SQL queries have no concept of count or exists, and it works with `String` or `List<String>` return types.

spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/test/domain/TradeRepository.java
```diff
@@ -35,6 +35,23 @@
...
+			+ "where action = @action limit 1)")
+	boolean existsByActionQuery(@Param("action") String action);
+
+	@Query("select action from :org.springframework.cloud.gcp.data.spanner.test.domain.Trade: "
+			+ "where action = @action limit 1")
+	String getFirstString(@Param("action") String action);
+
+	@Query("select action from :org.springframework.cloud.gcp.data.spanner.test.domain.Trade: "
+			+ "where action = @action")
+	List<String> getFirstStringList(@Param("action") String action);
```
meltsufin: Can we also add one that returns a single `Trade`?

spring-cloud-gcp-docs/src/main/asciidoc/spanner.adoc
```diff
@@ -1056,22 +1056,37 @@ The names of the tags of the SQL correspond to the `@Param` annotated names of t
...
+  	int countByActionQuery(String action);
+
+  	@Query(" select count(1) from trades where action = @action")
+  	boolean existsByActionQuery(String action);
+
+  	@Query("select action from trades where action = @action limit 1")
+  	String getFirstString(@Param("action") String action);
+
+  	@Query("select action from trades where action = @action")
+  	List<String> getFirstStringList(@Param("action") String action);
```
meltsufin: Maybe also add some of these examples to the sample app?

spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/test/domain/TradeRepository.java
ChengyuanZhao: @meltsufin PTAL. I added an integration test method for just a single POJO and also added these to the sample.

spring-cloud-gcp-docs/src/main/asciidoc/spanner.adoc
```diff
@@ -1056,22 +1056,37 @@ The names of the tags of the SQL correspond to the `@Param` annotated names of t
...
 	List<Trade> customSortedTrades = tradeRepository.sortedTrades(PageRequest
   				.of(2, 2, org.springframework.data.domain.Sort.by(Order.asc("id"))));
 ----
 
 The results would be sorted by "id" in ascending order.
 
+Your query method can also return non-entity types:
+[source, java]
+----
+		@Query(" select count(1) from trades where action = @action")
```
meltsufin: Formatting.

ChengyuanZhao: ah yes fixed

ChengyuanZhao:
@meltsufin @dmitry-s PTAL if possible. I have a PR coming next that supports DELETE part-tree queries but it touches nearly all the same code as here and would need this to be merged first.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerOperations.java
```diff
@@ -78,29 +79,27 @@
...
+	 * Executes a given query string with tags and parameters and applies a given function
+	 * to each row of the result.
+	 * @param rowFunc the function to apply to each row of the result.
+	 * @param statement the SQL statement used to select the objects.
+	 * @param options the options with which to run this query.
+	 * @return a list of the rows each transformed with the given function.
 	 */
-	<T> List<T> query(Class<T> entityClass, String sql, List<String> tags,
-			Object[] params,
+	<A> List<A> query(Function<Struct, A> rowFunc, Statement statement,
```
mp911de: Have a look at `JdbcTemplate` (or `CqlTemplate` vs. `CassandraTemplate`) as these type of functions are typically applied on a lower abstraction level (`Statement` level vs. logical operations).

Issue #1034, opened 2018-09-18
Implement new methods in ExtendedBindingProperties


meltsufin:
In 2.1, two [new methods were added](
https://github.com/spring-cloud/spring-cloud-stream/commit/07b5a691d4fe390d1f2e92f11614e62c1a1a39b9#diff-0a2bfdd7dbaeb4ebf847742d1da4c03c) to the `ExtendedBindingProperties` interface.
We need to implement them.

@artembilan Do you know what this is about, and how we should go about implementing them?

meltsufin:
I see that there is already a `spring-cloud-stream` `2.1.0.M2` [release](https://github.com/spring-cloud/spring-cloud-stream/releases/tag/v2.1.0.M2). @artembilan Is this the version that will go into `Greenwich.M1`? In that case we don't need to worry about it yet.

sobychacko:
@meltsufin this change is added (merged upstream today) to address the use case of allowing default extended producer/consumer properties. See [this issue](https://github.com/spring-cloud/spring-cloud-stream/issues/1360) for more details. There are PR's in review in Kafka and Rabbit binders at the moment to address these changes.

PR #1036, opened 2018-09-20
GH-1034: Align Pub/Sub Binder impl with SCSt core

artembilan:
Fixes spring-cloud/spring-cloud-gcp#1034

### DIFF
--- a/spring-cloud-gcp-docs/src/main/asciidoc/spring-stream.adoc
+++ b/spring-cloud-gcp-docs/src/main/asciidoc/spring-stream.adoc
@@ -46,7 +46,8 @@ spring.cloud.stream.bindings.output.destination=myConsumer
 spring.cloud.stream.gcp.pubsub.bindings.output.consumer.auto-create-resources=true
 ----
 
+Starting with version 1.1, these properties can be configured globally for all the bindings, e.g. `spring.cloud.stream.gcp.pubsub.default.consumer.auto-create-resources`.
+
 If you are using Pub/Sub auto-configuration from the Spring Cloud GCP Pub/Sub Starter, you should refer to the <<pubsub-configuration,configuration>> section for other Pub/Sub parameters.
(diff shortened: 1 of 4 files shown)


meltsufin:
"Alight"? What does that mean?
Fixes #1034?
I was just starting the same fix last night. I will take a look soon.

artembilan:
@meltsufin ,

Thank you for the pointers! 😃 

Fixed commit and PR description.

I somehow totally missed that issue.

Hope now it is good.

artembilan:
Well, we really don't have those *extended* properties to test against.
And also there is no Docs changes on the matter there in Kafka Binder, neither in the Rabbit Binder: https://github.com/spring-cloud/spring-cloud-stream-binder-rabbit/commit/bf26977d7fe10be456e18c539421a94825dfbb65

Therefore I don't think that there left too much to do for this issue .

Thanks for consideration this change! Although would be great to have such a change in stream with fixes for all other binder.  😉 

/CC @sobychacko

sobychacko:
@meltsufin That's correct.

artembilan:
OK. What do you want from me to do here yet?
Is there any guidance how to proceed with your requirements?

I can extract a test-case on the matter from the mentioned PRs, but I definitely don't see there any docs changes. That's confusing... 🤷‍♂️ 

Thanks for understanding.

artembilan:
>  I think docs are changed in the core docs

Right, that's why I insist do not change docs here. Or change docs for all involved binder implementations.

meltsufin:
Where is this updated documentation? We should at least link to it then?

sobychacko:
@meltsufin updated docs are here: https://docs.spring.io/spring-cloud-stream/docs/Fishtown.M3/reference/htmlsingle/#binding-properties
(just mentioned briefly there, how extended default properties could be configured).

PR #1038, opened 2018-09-21
Add Google-hosted maven mirror

### LINKED ISSUE #1035: Add Google Cloud Storage Maven Central Mirror
Travis is sometimes unable to download artifacts from Maven Central.
Google hosts a mirror which we should configure in our parent `pom.xml`.
See: https://storage-download.googleapis.com/maven-central/index.html

elefeint:
This may be a little wasteful since for snapshots maven downloads metadata from
every listed repository, but it will prevent Travis builds from timing out.

Fixes #1035.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -260,7 +260,7 @@
 					<name>GCS Maven Mirror</name>
 					<url>https://maven-central.storage-download.googleapis.com/repos/central/data</url>
 					<snapshots>
-						<enabled>true</enabled>
+						<enabled>false</enabled>
 					</snapshots>
 					<releases>
 						<enabled>true</enabled>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -255,6 +255,17 @@
 						<enabled>false</enabled>
 					</snapshots>
 				</repository>
+				<repository>
+					<id>gcs-mirror</id>
+					<name>GCS Maven Mirror</name>
+					<url>https://maven-central.storage-download.googleapis.com/repos/central/data</url>
+					<snapshots>
+						<enabled>true</enabled>
```
meltsufin: Maven Central does not store snapshots AFAIK. Why `true`?

PR #1042, opened 2018-09-21
[WIP] Datastore GQL support count exists

### LINKED ISSUE #1040: Datastore support query methods that return List of Key
If the user needs a `List<Key>` return type can't tell what the inner type is because of type erasure. Instead we will need to have a new annotation '@KeyQuery` to simply mark the query method that it returns list of keys.

ChengyuanZhao:
fixes #1040

### DIFF
--- a/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/GqlDatastoreQuery.java
+++ b/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/GqlDatastoreQuery.java
@@ -42,7 +42,6 @@ import org.springframework.cloud.gcp.data.datastore.core.mapping.DatastoreDataEx
 import org.springframework.cloud.gcp.data.datastore.core.mapping.DatastoreMappingContext;
 import org.springframework.data.repository.query.Parameter;
 import org.springframework.data.repository.query.Parameters;
-import org.springframework.data.repository.query.QueryMethod;
 import org.springframework.data.repository.query.QueryMethodEvaluationContextProvider;
 import org.springframework.expression.spel.standard.SpelExpressionParser;
 import org.springframework.util.StringUtils;
@@ -99,7 +98,7 @@ public class GqlDatastoreQuery<T> extends AbstractDatastoreQuery<T> {
 	 * @param datastoreOperations used for executing queries.
 	 * @param datastoreMappingContext used for getting metadata about entities.
 	 */
-	public GqlDatastoreQuery(Class<T> type, QueryMethod queryMethod,
+	public GqlDatastoreQuery(Class<T> type, DatastoreQueryMethod queryMethod,
 			DatastoreOperations datastoreOperations, String gql,
 			QueryMethodEvaluationContextProvider evaluationContextProvider,
 			SpelExpressionParser expressionParser,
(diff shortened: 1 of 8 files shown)


spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/GqlDatastoreQuery.java
```diff
@@ -118,7 +117,28 @@ public Object execute(Object[] parameters) {
 		List<T> rawResult = found == null ? Collections.emptyList()
 				: StreamSupport.stream(found.spliterator(), false)
 						.collect(Collectors.toList());
-		return applyProjection(rawResult);
+		Object result;
+		if (isCountQuery()) {
+			result = rawResult.size();
```
meltsufin: Cloud Spanner has the `COUNT` aggregate function. So, I don't think we need to do anything special here.
If someone provides `SELECT count(*) from users`, this code will always return 1. That's not right. We should just allow mapping to a singe POJO and error out if the result set has more than 1 row.

See: https://cloud.google.com/spanner/docs/functions-and-operators#aggregate-functions

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/GqlDatastoreQuery.java
```diff
@@ -118,7 +117,28 @@ public Object execute(Object[] parameters) {
...
+		}
+		else if (isExistsQuery()) {
+			result = !rawResult.isEmpty();
```
meltsufin: `EXISTS` also seems to be supported. So, like in the comment above, I don't think we need to do anything special.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/GqlDatastoreQuery.java
ChengyuanZhao: This one is datastore though !

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/GqlDatastoreQuery.java
ChengyuanZhao: This is for datastore!

meltsufin: Oops! I thought I was looking at the Spanner one :)

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/GqlDatastoreQuery.java
meltsufin: Oops :)

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/GqlDatastoreQuery.java
```diff
@@ -118,7 +117,28 @@ public Object execute(Object[] parameters) {
...
+		}
+		else {
+			result = applyProjection(rawResult);
+		}
+		return result;
+	}
+
+	private boolean isCountQuery() {
+		Class returnType = this.queryMethod.getReturnedObjectType();
+		return returnType == int.class || returnType == Integer.class || returnType == long.class ||
```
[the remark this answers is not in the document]
ChengyuanZhao: I would really prefer to not depend on the name of the method at all. I think that's much less of a firm thing than the return type. 

I actually noted from our discussion with the Spring Data team that the return type would be a good way to identify this. Another thing to remember is that the user would need to write this integer type as the return type even if he does have to set the new annotation setting.

spring-cloud-gcp-data-datastore/src/test/java/org/springframework/cloud/gcp/data/datastore/it/TestEntityRepository.java
```diff
@@ -30,6 +30,12 @@
 	@Query("select * from  test_entities_ci where id = @id_val")
 	List<TestEntity> findEntitiesWithCustomQuery(@Param("id_val") String id);
 
+	@Query("select * from  test_entities_ci where id = @id_val")
```
meltsufin: Can we add a test where 
```
@Query("select age from people where last_name = @name")
Integer findAgeByLastname(String lastName);
```
?

ChengyuanZhao: @meltsufin Added this test case and added the annotation setting for counts/exists just like how it's done in cassandra. PTAL

PR #1046, opened 2018-09-25
Datastore key GQL queries

ChengyuanZhao:
fixes https://github.com/spring-cloud/spring-cloud-gcp/issues/1040

### DIFF
--- a/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreOperations.java
+++ b/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreOperations.java
@@ -18,7 +18,6 @@ package org.springframework.cloud.gcp.data.datastore.core;
 
 import java.util.function.Function;
 
-import com.google.cloud.datastore.BaseEntity;
 import com.google.cloud.datastore.Key;
 import com.google.cloud.datastore.Query;
 
@@ -128,18 +127,12 @@ public interface DatastoreOperations {
 	 * Finds objects by using a Cloud Datastore query.
 	 * @param query the query to execute.
 	 * @param entityClass the type of object to retrieve.
-	 * @param <T> the type of object to retrieve.
+	 * @param <T> the type of object to retrieve. If the query was a key-query, then
+	 * {@link Key}s are returned instead of the mapped entities.
 	 * @return a list of the objects found. If no keys could be found the list will be
 	 * empty.
 	 */
-	<T> Iterable<T> query(Query<? extends BaseEntity> query, Class<T> entityClass);
-
-	/**
-	 * Finds Cloud Datastore Keys by using a Cloud Datastore query.
-	 * @param query the query to execute that retrieves only Keys.
-	 * @return the list of keys found.
-	 */
-	Iterable<Key> queryKeys(Query<Key> query);
+	<T> Iterable<?> query(Query query, Class<T> entityClass);
 
 	/**
 	 * Get all the entities of the given domain type.
--- a/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreOperations.java
+++ b/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreOperations.java
@@ -132,7 +132,7 @@ public interface DatastoreOperations {
 	 * @return a list of the objects found. If no keys could be found the list will be
 	 * empty.
 	 */
-	<T> Iterable<?> query(Query<? extends BaseEntity> query, Class<T> entityClass);
+	<T> Iterable<T> query(Query<? extends BaseEntity> query, Class<T> entityClass);
 
 	/**
 	 * Finds Cloud Datastore Keys by using a Cloud Datastore query.
--- a/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreOperations.java
+++ b/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreOperations.java
@@ -135,6 +135,13 @@ public interface DatastoreOperations {
 	<T> Iterable<?> query(Query query, Class<T> entityClass);
 
 	/**
+	 * Finds Cloud Datastore Keys by using a Cloud Datastore query.
+	 * @param query the query to execute that retrieves only Keys.
+	 * @return the list of keys found.
+	 */
+	Iterable<Key> queryKeys(Query<Key> query);
+
+	/**
 	 * Get all the entities of the given domain type.
 	 * @param entityClass the domain type to get.
 	 * @param <T> the type param of the domain type.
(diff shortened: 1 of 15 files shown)


ChengyuanZhao:
@meltsufin PTAL

ChengyuanZhao:
@dmitry-s @meltsufin PTAL

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/DatastoreQueryMethod.java
```diff
@@ -78,6 +78,16 @@ boolean isCountQuery() {
 		return annotation != null && annotation.count();
 	}
 
+	/**
+	 * Return whether this method is an count query.
```
meltsufin: The comment here seems off.

spring-cloud-gcp-data-datastore/src/test/java/org/springframework/cloud/gcp/data/datastore/it/TestEntityRepository.java
```diff
@@ -33,12 +35,16 @@
...
 
-	@Query(value = "select * from  test_entities_ci where id = @id_val", exists = true)
+	@Query(value = "select __key__ from  test_entities_ci "
+			+ "where id = @id_val", exists = true, keyQuery = true)
 	boolean existsByEntitiesWithCustomQuery(@Param("id_val") long id);
 
 	@Query("select id from  test_entities_ci where id <= @id_val ")
 	List<TestEntity> findEntitiesWithCustomProjectionQuery(@Param("id_val") long id);
 
+	@Query(value = "select __key__ from test_entities_ci", keyQuery = true)
```
meltsufin: Why can't we determine that it's a key query based on the return type? I believe Mark mentioned that it's possible.

ChengyuanZhao: Though I think we can do this another way. i'll add a commit shortly.

ChengyuanZhao: @meltsufin PTAL. I removed this setting but had to do it a different way.

ChengyuanZhao: The issue before the most recent commit @meltsufin was that we didn't know if we should try to map the result or return it directly. However by losing some type-safety we can treat entity queries and key queries together. It is a tradeoff for type safety, but I think worth it.

meltsufin: Thanks! I think this is on the right track now.
What would happen if you allowed a Key query to go through the mapping machinery? Also, what if it's not a Key query, but something like a custom projection, think about the selecting plates from cars table and return `List<String>`?

ChengyuanZhao: - So right now we can tell if the query is a Key or not based on the results of the query and not the return type of the function. I think this is safest because there are no assumptions being made. If the query was a projection or full row query (it's the same to us), then we will map the objects and the query method will return those. At this point there will be a cast exception because the entities cannot be cast to Key (at runtime if the query method is called). 

- Regarding `List<String>`. There are 2 possibilities. If the user had given a Key query, then we will return the list of keys. However due to type erasure I don't think this will be a runtime cast error unless the user actually tries to access an object inside the list at which point the Key to String cast will fail.  On the other hand if the user had given a projection or full entity query then he will get back a List of POJOs and the same situation would happen (cast error when accessing an item). 

We can tackle supporting the general `List<SimpleType>` issue separately in the next PR and it wouldn't conflict with this PR.

meltsufin: My understanding is that we have access to both the method return type and query result set type. So, it should simply be a matter of running the result set through the conversion machinery with the conversion target being the return type of the method. What am I missing here? @dmitry-s can you please weigh in on this?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreOperations.java
```diff
@@ -128,18 +127,12 @@
...
 	 */
-	<T> Iterable<T> query(Query<? extends BaseEntity> query, Class<T> entityClass);
-
-	/**
-	 * Finds Cloud Datastore Keys by using a Cloud Datastore query.
-	 * @param query the query to execute that retrieves only Keys.
-	 * @return the list of keys found.
-	 */
-	Iterable<Key> queryKeys(Query<Key> query);
+	<T> Iterable<?> query(Query query, Class<T> entityClass);
```
meltsufin: Since a key query is special and very different from an entity query, can we go back to the two separate method signatures that reflect this and buy us the type-safety?

ChengyuanZhao: @meltsufin I added back `queryKeys` however I did not decide to use it the part-tree query method. The reason is that i felt it made the code more complicated to read to have that BiFunction there and have it reassigned. The `query` method must support key queries anyway for the GQL query methods, so I think we should still use it in part-tree to simplify the code.

meltsufin: Can we put the generics back in though?

ChengyuanZhao: @meltsufin what do you mean ? The reason they the `<? extends BaseEntity>` was removed was because Key is not an entity.

meltsufin: Don't we have `queryKeys` now for that purpose so that we can specialize `query` for entities only.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
```diff
@@ -136,11 +137,19 @@ public long count(Class<?> entityClass) {
 	}
 
 	@Override
-	public <T> Iterable<T> query(Query<? extends BaseEntity> query,
+	public <T> Iterable<?> query(Query<? extends BaseEntity> query,
```
meltsufin: `Iterable<T>` please

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
```diff
@@ -136,11 +137,19 @@ public long count(Class<?> entityClass) {
...
 			Class<T> entityClass) {
 		return convertEntities(this.datastore.run(query), entityClass);
 	}
 
+	public <T> Iterable<?> queryKeysOrEntities(Query query, Class<T> entityClass) {
```
meltsufin: Please add javadoc.

Issue #975, opened 2018-08-28
Automatic grouping of Pub/Sub messages by subscription name when acking


meltsufin:
Using `ack()`, `nack()`, and `modifyAckDeadline()` methods for acknowledging messages in batches is more efficient than acknowledging messages individually, but they *require* the collection of messages to be from the same project and subscription. This restriction can be lifted by automatically grouping messages by subscription name and processing those batches separately.

The feature was originally there, but removed in #859 when converting to asynchronous acknowledgement.

dhoard:
@meltsufin Since subscription names are only unique within a project, I don't understand how this could be implemented.  That's why I am grouping based on `ProjectSubscriptionName`

What am I missing?

given ...

message A from a subscription named "test-subscription" via `PubSubSubscriber` A / credentials A in project A.

message B from subscription named "test-subscription" via `PubSubSubscriber` B / credentials B in project B.

then ...

both have the same subscription name (`test-subscription`), but because they are from different projects, they can't necessarily be grouped together.


Also, in this scenario ...

message B can't be acknowledged by `PubSubSubscriber` A

message A can't be acknowledged by `PubSubSubscriber` B


If `PubSubSubscriber` A and B are both using the same credentials, then it would be possible for either `PubSubSubscriber` (A or B) to acknowledge the message.

meltsufin:
I forgot to mention that all messages still have to be from the same project, but not necessarily from the same subscription. Does that make sense?

meltsufin:
This would indeed be a problem, but I would consider it to be a user error because it would mean that those message came from different Pub/Sub templates, yet being submitted together for acking to one of the templates.

That being said, if I'm the only one who thinks this would be useful, I willing to drop this feature. 😄

dhoard:
I would be interested in other people's thoughts. @dmitry-s @artembilan @ChengyuanZhao

artembilan:
> message B can't be acknowledged by PubSubSubscriber A

Can you confirm this with your experience or test case?
If that really fail with an error, we are good just document restrictions and left everything else up to end-user to group messages properly.
Well, i mean we still can group them automatically here in the template: the way @meltsufin suggests, but fail if there is wrong subscription.

We still fail here with the:
```
Assert.isTrue(
					projectSubscriptionName.equals(acknowledgeablePubsubMessage.getProjectSubscriptionName()),
					"The project id and subscription of all messages must match.");
```
[...]

PR #1047, opened 2018-09-25
Make pubsub sample page more user-friendly.

elefeint:
It's not fancy, but makes it easier to visually separate
different functionality.

Next I'd like to add ability to publish messages to multiple
subscriptions at the same time, and to subscribe with batch acking, to
exercise behavior that will change due to #975.

Before:
pubsub_ui_before

After:
pubsub_ui_after


meltsufin:
@elefeint Feel free to merge.

elefeint:
@meltsufin need LGTM because I merged the child branch with #1051  in.

Issue #1032, opened 2018-09-18
Add SpannerTransactionManager auto-config


meltsufin:
Conditional on `spring-tx`.
Follow-up to #918.

meltsufin:
Per @artembilan , `SpannerTransactionManagerAutoConfiguration`, similar to `DataSourceTransactionManagerAutoConfiguration`.

Issue #973, opened 2018-08-28
`@ConditionalOnProperty` target property with no metadata


snicoll:
Several auto-configurations use a condition on a property that has no metadata. As a result, trying to use it to customize the auto-configuration lead to a warning in an IDE.

The report detected the following:

```
org.springframework.cloud.gcp.autoconfigure.logging.StackdriverLoggingAutoConfiguration [spring.cloud.gcp.logging.enabled]
org.springframework.cloud.gcp.autoconfigure.pubsub.GcpPubSubAutoConfiguration [spring.cloud.gcp.pubsub.enabled]
org.springframework.cloud.gcp.autoconfigure.sql.GcpCloudSqlAutoConfiguration [spring.cloud.gcp.sql.enabled]
org.springframework.cloud.gcp.autoconfigure.trace.StackdriverTraceAutoConfiguration [spring.cloud.gcp.trace.enabled]
  
```

artembilan:
The mentioned configuration properties just don't have a particular `@ConfigurationProperties` class representative.

@snicoll ,

any ideas what would you like to see as a fix?

Thanks

snicoll:
Thanks for the feedback.

> any ideas what would you like to see as a fix?

Yes, that's what manual metadata is for. When a property is not used by the code directly, it doesn't make much sense to fake a property by adding it to the model. For those cases, we can [document the keys manually](https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#configuration-metadata-additional-metadata).

artembilan:
Thanks, @snicoll , for the right direction!

Any pointers how to test or at least reproduce?
Is it just a content assistance issue in the IDE?

snicoll:
I can't provide you with a way to test it doesn't happen again I am afraid as this is some code from Jetbrains scanning the code and spotting those mistakes (See also https://youtrack.jetbrains.com/issue/IDEA-197990).

> Is it just a content assistance issue in the IDE?

Like any other property the project may define, yes.

PR #1048, opened 2018-09-25
Spanner transactional integration tests, autoconfig, readonly

ChengyuanZhao:
fixes https://github.com/spring-cloud/spring-cloud-gcp/issues/1026
fixes https://github.com/spring-cloud/spring-cloud-gcp/issues/1032


spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/spanner/DatabaseClientTransactionManagerAutoConfiguration.java
```diff
@@ -0,0 +1,63 @@
...
+/**
+ * Auto-configuration for {@link SpannerTransactionManager}.
+ *
+ * @author Chengyuan Zhao
+ *
+ * @since 1.1
+ */
+@Configuration
+@EnableTransactionManagement
+@ConditionalOnClass({ SpannerTemplate.class, SpannerTransactionManager.class })
```
meltsufin: Both of these classes are in the same module. You should only need one of them.
Can we also make it conditional on a class in `spring-tx`? Of course, `spring-tx` would need to be an `optional` dependency. I remember you saying something before that `spring-tx` was already needed for something non-transactional in our Spanner implementation. Can you look into that?

ChengyuanZhao: Will make the conditional-on-class changes as requested.

About spring-tx, it was added because our exception type `SpannerDataException` extends `org.springframework.dao.DataAccessException`

ChengyuanZhao: noted

ChengyuanZhao:
@meltsufin PTAL

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/spanner/DatabaseClientTransactionManagerAutoConfiguration.java
```diff
@@ -0,0 +1,62 @@
...
+ * Auto-configuration for {@link SpannerTransactionManager}.
+ *
+ * @author Chengyuan Zhao
+ *
+ * @since 1.1
+ */
+@Configuration
+@EnableTransactionManagement
+@ConditionalOnClass(SpannerTransactionManager.class)
+@AutoConfigureOrder(Ordered.LOWEST_PRECEDENCE)
```
meltsufin: Why?

spring-cloud-gcp-docs/src/main/asciidoc/spanner.adoc
```diff
@@ -885,20 +885,11 @@ point in time of the transaction.
 Because read-only transactions are non-locking and can be performed on points in time in the past, these
 are recommended for functions that do not perform write operations.
 
-===== Declarative transaction management
+===== @Transactional Annotation
```
meltsufin: Compromise: `Declarative transaction management with @Transactional`.

spring-cloud-gcp-docs/src/main/asciidoc/spanner.adoc
```diff
@@ -885,20 +885,11 @@ point in time of the transaction.
...
 
-To use `@Transactional` in [declarative spring transaction management](https://docs.spring.io/spring/docs/current/spring-framework-reference/data-access.html#transaction-declarative)
-`SpannerTransactionManager` has to be registered in Spring.
-
-Example:
-```
-@Bean
-public SpannerTransactionManager transactionManager() {
```
meltsufin: I don't think we necessarily need to remove this from documentation. Some people may not be using the starter. In any case, if you do remove it, at least we need to clarify that the auto-config/starter will create the `SpannerTransactionManager` needed to enable it.

spring-cloud-gcp-docs/src/main/asciidoc/spanner.adoc
```diff
@@ -885,20 +885,11 @@ point in time of the transaction.
 Because read-only transactions are non-locking and can be performed on points in time in the past, these
 are recommended for functions that do not perform write operations.
 
-===== Declarative transaction management
+===== @Transactional Annotation
 
-To use `@Transactional` in [declarative spring transaction management](https://docs.spring.io/spring/docs/current/spring-framework-reference/data-access.html#transaction-declarative)
```
meltsufin: Link to declarative transaction management docs should remain.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTransactionManager.java
```diff
@@ -91,8 +95,9 @@ protected void doBegin(Object transactionObject, TransactionDefinition transacti
 							"TransactionDefinition.PROPAGATION_REQUIRED");
 		}
 		Tx tx = (Tx) transactionObject;
-		final TransactionContext targetTransactionContext = tx.transactionManager.begin();
 		if (transactionDefinition.isReadOnly()) {
+			final ReadContext targetTransactionContext = this.databaseClient
+					.readOnlyTransaction();
 
 			tx.transactionContext = new TransactionContext() {
```
meltsufin: I thought we wouldn't need this whole wrapper below when using `readOnlyTransaction`??

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/spanner/DatabaseClientTransactionManagerAutoConfiguration.java
```diff
@@ -0,0 +1,62 @@
...
+/**
+ * Auto-configuration for {@link SpannerTransactionManager}.
+ *
+ * @author Chengyuan Zhao
+ *
+ * @since 1.1
+ */
+@Configuration
+@EnableTransactionManagement
+@ConditionalOnClass(SpannerTransactionManager.class)
```
meltsufin: Like I said before, we really want it to be conditional on something in `spring-tx`. I know there's an issue with that now, but are we tracking it?

ChengyuanZhao: I don't really understand... Isn't it only conditional if spring-tx is optional and it isn't optional right now? I already stated why we have spring-tx as non-optional (we are using org.springframework.dao.dataaccessexception). We can just not have our exception class extend that one and remove direct use of spring-tx altogether. 

https://github.com/spring-cloud/spring-cloud-gcp/issues/1059 Here is an issue for discussing that. 

Is there something so wrong with including spring-tx all the time? This seems like a core feature we're offering..

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/spanner/DatabaseClientTransactionManagerAutoConfiguration.java
ChengyuanZhao: This also appears in the JDBC example I looked at. It makes sense to configure this after everything else, no?

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTransactionManager.java
ChengyuanZhao: We need a wrapper because from Spring Data's   point of view both transactions are the same type. I am substituting in the read-only context and Spring Data is not supposed to be aware of the difference. This is why.

spring-cloud-gcp-docs/src/main/asciidoc/spanner.adoc
ChengyuanZhao: Added a sentence mentioning that this is provided by the starter at the top of this section.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/spanner/DatabaseClientTransactionManagerAutoConfiguration.java
meltsufin: That issue is what I was looking for. Thanks!

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/spanner/DatabaseClientTransactionManagerAutoConfiguration.java
```diff
@@ -0,0 +1,62 @@
...
+@AutoConfigureOrder(Ordered.LOWEST_PRECEDENCE)
+@EnableConfigurationProperties(GcpSpannerProperties.class)
+public class DatabaseClientTransactionManagerAutoConfiguration {
```
artembilan: Why it can't be called like `SpannerTransactionManagerAutoConfiguration`?
I understand that this is an auto-configuration and is not for intention to be used in the end-user `@Import`, but there might be some use-case when end-end needs to *exclude* this or just would like to read a source code on the matter. Therefore a concern about naming consistency.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTransactionManager.java
```diff
@@ -155,7 +160,7 @@ public void close() {
 			};
 		}
 		else {
-			tx.transactionContext = targetTransactionContext;
+			tx.transactionContext = tx.transactionManager.begin();
```
artembilan: Is it OK do not have `.begin()` in case of `readOnlyTransaction()` ?

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/spanner/DatabaseClientTransactionManagerAutoConfiguration.java
```diff
@@ -0,0 +1,62 @@
...
+
+/**
+ * Auto-configuration for {@link SpannerTransactionManager}.
+ *
+ * @author Chengyuan Zhao
+ *
+ * @since 1.1
+ */
+@Configuration
+@EnableTransactionManagement
```
meltsufin: Remove this, and make sure that `TransactionAutoConfiguration` is after this config.
Here's how: `@AutoConfigureBefore(TransactionAutoConfiguration.class)`

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/spanner/DatabaseClientTransactionManagerAutoConfiguration.java
```diff
@@ -0,0 +1,62 @@
...
+	static class DatabaseClientTransactionManagerConfiguration {
+
+		private final DatabaseClient databaseClient;
+
+		DatabaseClientTransactionManagerConfiguration(DatabaseClient databaseClient) {
+			this.databaseClient = databaseClient;
+		}
+
+		@Bean
+		@ConditionalOnMissingBean(SpannerTransactionManager.class)
```
meltsufin: Replace with:
`@ConditionalOnMissingBean(PlatformTransactionManager.class)`

ChengyuanZhao: @meltsufin PTAL

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/spanner/SpannerTransactionManagerAutoConfiguration.java
```diff
@@ -39,9 +40,11 @@
 @Configuration
 @EnableTransactionManagement
 @ConditionalOnClass(SpannerTransactionManager.class)
-@AutoConfigureOrder(Ordered.LOWEST_PRECEDENCE)
+@AutoConfigureBefore(PlatformTransactionManager.class)
```
meltsufin: TransactionAutoConfiguration.class

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/spanner/SpannerTransactionManagerAutoConfiguration.java
```diff
@@ -39,9 +40,11 @@
 @Configuration
 @EnableTransactionManagement
```
meltsufin: `@EnableTransactionManagement` should not be here.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/spanner/SpannerTransactionManagerAutoConfiguration.java
```diff
@@ -39,9 +40,11 @@
...
 @ConditionalOnClass(SpannerTransactionManager.class)
-@AutoConfigureOrder(Ordered.LOWEST_PRECEDENCE)
+@AutoConfigureBefore(PlatformTransactionManager.class)
+@ConditionalOnProperty(value = "spring.cloud.gcp.spanner.transactional-annotation.enabled",
```
meltsufin: `spring.cloud.gcp.spanner.transactions.enabled` because you can use the transaction manager without `@Transactional`?

spring-cloud-gcp-docs/src/main/asciidoc/spanner.adoc
```diff
@@ -90,6 +90,7 @@ https://developers.google.com/identity/protocols/googlescopes[OAuth2 scope] for
 Cloud Spanner credentials | No | https://www.googleapis.com/auth/spanner.data
 | `spring.cloud.gcp.spanner.createInterleavedTableDdlOnDeleteCascade` | If `true`, then schema statements generated by `SpannerSchemaUtils` for tables
 with interleaved parent-child relationships will be "ON DELETE CASCADE". The schema for the tables will be "ON DELETE NO ACTION" if `false`. | No | `true`
+| `spring.cloud.gcp.spanner.transactional-annotation.enabled` | Enables support for Spring Data @Transactional annotation | No | `true`
```
meltsufin: I would say, configures `SpannerTransactionaManager` for use with declarative transaction management and `@Transactional`.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/spanner/SpannerTransactionManagerAutoConfiguration.java
ChengyuanZhao: Yes the `@Transactional` is the only one affected. The modern style transactions-in-a-function is still supported.

artembilan: Why can't we live without this property? The fact is that no any other transaction management auto-configuration does something similar...

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/spanner/SpannerTransactionManagerAutoConfiguration.java
```diff
@@ -0,0 +1,65 @@
...
+
+		private final DatabaseClient databaseClient;
+
+		DatabaseClientTransactionManagerConfiguration(DatabaseClient databaseClient) {
+			this.databaseClient = databaseClient;
+		}
+
+		@Bean
+		@ConditionalOnMissingBean(PlatformTransactionManager.class)
+		public SpannerTransactionManager spannerTransactionManager() {
```
artembilan: There was also a discussion to have here a `TransactionManagerCustomizers` as well.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/spanner/SpannerTransactionManagerAutoConfiguration.java
ChengyuanZhao: I'm OK with it being on permanently actually. @meltsufin WDYT

meltsufin: The reason we need it, I believe, is because we can't control it via classpath. It was Mark's suggestion, and I agree with it. We can, of course, decide to enable it by default.

artembilan: Well, not sure. We don't control it in case of JDBC TX Manager either. There is just enough to have a `spring-jdbc` in classpath and that's all.
We don't do that for the mentioned `Neo4jTransactionManager`, too, nor for `JpaTransactionManager`.
But if insist, let it be. We can always remove it afterward.

Only the problem that this has to be included into some `@ConfigurationProperties` or added into the `additional-spring-configuration-metadata.json`.

meltsufin: Sure, if no other Spring Data project does it, I'm fine with not doing too.

ChengyuanZhao: I actually didn't know about `additional-spring-configuration-metadata.json` until just now. What is the significance of this file, and how come only a few options are listed in there?

ChengyuanZhao: removed.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/spanner/SpannerTransactionManagerAutoConfiguration.java
ChengyuanZhao: added

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/spanner/SpannerTransactionManagerAutoConfiguration.java
artembilan: > I actually didn't know about additional-spring-configuration-metadata.json until just now.

See this issue: https://github.com/spring-cloud/spring-cloud-gcp/issues/973

ChengyuanZhao:
@meltsufin @artembilan Are there more changes? PTAL

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/spanner/SpannerTransactionManagerAutoConfiguration.java
```diff
@@ -0,0 +1,73 @@
...
+ * Auto-configuration for {@link SpannerTransactionManager}.
+ *
+ * @author Chengyuan Zhao
+ *
+ * @since 1.1
+ */
+@Configuration
+@ConditionalOnClass(SpannerTransactionManager.class)
+@AutoConfigureBefore(TransactionAutoConfiguration.class)
+@EnableConfigurationProperties(GcpSpannerProperties.class)
```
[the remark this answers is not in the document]
ChengyuanZhao: added unit test. removed this line. PTAL

PR #1055, opened 2018-09-28
Group pubsub acks by subscription

elefeint:
The final ListenableFuture returned will complete when either all async calls complete successfully OR as soon as one fails.

Other changes:
* Added a second subscription field to the sample app to allow for testing of multiple-subscription acking.
* Cleaned up warnings in the touched files.

### DIFF
--- a/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberTemplate.java
+++ b/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberTemplate.java
@@ -310,9 +310,9 @@ public class PubSubSubscriberTemplate implements PubSubSubscriberOperations {
 		int numExpectedFutures = groupedMessages.size();
 		AtomicInteger numCompletedFutures = new AtomicInteger();
 
-		for (ProjectSubscriptionName psName : groupedMessages.keySet()) {
+		groupedMessages.forEach((ProjectSubscriptionName psName, List<String> ackIds) -> {
 
-			ApiFuture<Empty> ackApiFuture = asyncOperation.apply(psName.getSubscription(), groupedMessages.get(psName));
+			ApiFuture<Empty> ackApiFuture = asyncOperation.apply(psName.getSubscription(), ackIds);
 
 			ApiFutures.addCallback(ackApiFuture, new ApiFutureCallback<Empty>() {
 				@Override
@@ -334,8 +334,7 @@ public class PubSubSubscriberTemplate implements PubSubSubscriberOperations {
 					}
 				}
 			}, MoreExecutors.directExecutor());
-
-		}
+		});
 
 		return settableListenableFuture;
 	}
--- a/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberTemplate.java
+++ b/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberTemplate.java
@@ -64,6 +64,7 @@ import org.springframework.util.concurrent.SettableListenableFuture;
  * @author Mike Eltsufin
  * @author Chengyuan Zhao
  * @author Doug Hoard
+ * @author Elena Felder
  *
  * @since 1.1
  */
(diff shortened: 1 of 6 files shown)


spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberTemplate.java
```diff
@@ -315,6 +281,67 @@ public void onSuccess(Empty empty) {
 		return this.subscriberStub.modifyAckDeadlineCallable().futureCall(modifyAckDeadlineRequest);
 	}
 
+	/**
+	 * Perform Pub/Sub operations (ack/nack/modifyAckDeadline) in per-subscription batches.
+	 * <p>The returned {@link ListenableFuture} will complete when either all batches completes successfully or when at
+	 * least one fails.</p>
+	 * <p>
+	 * In case of multiple batch failures, which exception exception will be in the final {@link ListenableFuture} is
```
meltsufin: "exception exception"

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberTemplate.java
```diff
@@ -315,6 +281,67 @@ public void onSuccess(Empty empty) {
...
+	 * non-deterministic.
+	 * </p>
+	 * @param acknowledgeablePubsubMessages messages, could be from different subscriptions.
+	 * @param asyncOperation Pub/Sub operation to perform.
+	 * @return ListenableFuture indicating overall success or failure.
```
meltsufin: Even though it's private, for consistency: `{@link ListenableFuture}`.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberTemplate.java
```diff
@@ -315,6 +281,67 @@ public void onSuccess(Empty empty) {
...
+				.collect(
+						Collectors.groupingBy(
+								AcknowledgeablePubsubMessage::getProjectSubscriptionName,
+								Collectors.mapping(AcknowledgeablePubsubMessage::getAckId, Collectors.toList())));
+
+		Assert.state(groupedMessages.keySet().stream().map(ProjectSubscriptionName::getProject).distinct().count() == 1,
+				"The project id of all messages must match.");
+
+		SettableListenableFuture<Void> settableListenableFuture	= new SettableListenableFuture<>();
+		ExecutorService cachedThreadPoolExecutorService = Executors.newCachedThreadPool();
```
meltsufin: I'm not sure we should use a cached thread pool here, especially since we're not allowing it to be configured. I would just go with a direct executor. WDYT?

spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-sample/src/main/java/com/example/WebController.java
```diff
@@ -69,22 +70,33 @@ public RedirectView publish(@RequestParam("topicName") String topicName,
 	}
 
 	@GetMapping("/pull")
-	public RedirectView pull(@RequestParam("subscription") String subscriptionName) {
```
meltsufin: This is supposed to be a simple example for people to get started with. Now it's a lot more complicated. Maybe leave the original, and add another endpoint called `pullFromTwo` or something?

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberTemplate.java
elefeint: Direct is fine; it's consistent with previous behavior where the executor selection was delegated to ApiFutures.addCallback(). I just did not want to keep the deprecated version of addCallback with unspecified executor.

artembilan: That's correct. Only my concern that you create a new `Executors.newCachedThreadPool()` for each method call.
This is not acceptable. The thread executor should be global one for the whole instance and should be destroyed in the end of component lifecycle.
Also we need to provide a setter to allow to inject an externally managed executor.

That's nothing more.

spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-sample/src/main/java/com/example/WebController.java
elefeint: Split into two methods: pull and multipull.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberTemplate.java
elefeint: @artembilan Did you have in mind a separate thread pool per each PubSubSubscriberTemplate instance?

artembilan: Well, typically we have only a single `PubSubSubscriberTemplate` in our microservice application, so that's OK to have some default internal `ThreadPoolTaskExecutor` in the `PubSubSubscriberTemplate`.

With an extra setter the user has option to provide a managed executor from the application context.

elefeint:
That was supposed to be "closes #975".

Issue #1039, opened 2018-09-21
Add test for extended default properties


meltsufin:
Follow-up to #1036.

See [Kafka for example](https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/pull/447/files#diff-f1ac5bc7515374575e036db53ee5822f).

PR #1062, opened 2018-10-02
WIP: PubSub ExtendedPropertiesBindings doesn't seem to propagate default settings

dzou:
This is a followup to #1039 and #1036.
@meltsufin @artembilan 

It seems like the default extended properties binding support doesn't quite work yet; I have an example unit test in this PR to demo this. In the test, I set the default value of auto-create-resources to false and will now expect the 2 bindings to also have auto-create-resources=false, but they retain the value of true.

Run:
cd spring-cloud-gcp-pubsub-stream-binder/
mvn -Dtest=PubSubExtendedBindingsPropertiesTest test

To try fixing the problem, I modified [PubSubConsumerProperties](https://github.com/spring-cloud/spring-cloud-gcp/blob/master/spring-cloud-gcp-pubsub-stream-binder/src/main/java/org/springframework/cloud/gcp/stream/binder/pubsub/properties/PubSubConsumerProperties.java) to implement  [`MergableProperties`](https://docs.spring.io/spring-cloud-stream/docs/Fishtown.BUILD-SNAPSHOT/api/org/springframework/cloud/stream/config/MergableProperties.html). This follows what is done by Kafka in [KafkaConsumerProperties](https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/blob/5446485cbf129c4800033b35f3b364b9559ef19f/spring-cloud-stream-binder-kafka-core/src/main/java/org/springframework/cloud/stream/binder/kafka/properties/KafkaConsumerProperties.java).

It turns out that implementing `MergableProperties` does propagate the default values but with 2 caveats:
1. The interface is marked "Not intended for public use" in the API docs
2. While the default value is propagated, it doesn't seem like you can override the default values with additional properties being set.

Any suggestions on how to proceed here?

### DIFF
--- a/spring-cloud-gcp-pubsub-stream-binder/src/test/java/org/springframework/cloud/gcp/stream/binder/pubsub/properties/PubSubExtendedBindingsPropertiesTest.java
+++ b/spring-cloud-gcp-pubsub-stream-binder/src/test/java/org/springframework/cloud/gcp/stream/binder/pubsub/properties/PubSubExtendedBindingsPropertiesTest.java
@@ -1,5 +1,7 @@
 package org.springframework.cloud.gcp.stream.binder.pubsub.properties;
 
+import static org.assertj.core.api.Assertions.assertThat;
+
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -18,13 +20,10 @@ import org.springframework.messaging.SubscribableChannel;
 import org.springframework.test.context.junit4.SpringRunner;
 
 @RunWith(SpringRunner.class)
-@EnableAutoConfiguration
-@EnableConfigurationProperties(PubSubExtendedBindingProperties.class)
 @SpringBootTest(
     webEnvironment = SpringBootTest.WebEnvironment.NONE,
     properties = {
 				"spring.cloud.stream.gcp.pubsub.default.consumer.auto-create-resources=false",
-				"spring.cloud.stream.gcp.pubsub.bindings.input.consumer.auto-create-resources=false",
 		})
 public class PubSubExtendedBindingsPropertiesTest {
 
@@ -36,14 +35,19 @@ public class PubSubExtendedBindingsPropertiesTest {
 		BinderFactory binderFactory = context.getBeanFactory().getBean(BinderFactory.class);
 		PubSubMessageChannelBinder binder = (PubSubMessageChannelBinder) binderFactory.getBinder("pubsub", MessageChannel.class);
 
+		// Print out the default values; they should both be false.
 		System.out.println("default for INPUT: " + binder.getExtendedConsumerProperties("input").isAutoCreateResources());
 		System.out.println("default for CUSTOM INPUT: " + binder.getExtendedConsumerProperties("custom-in").isAutoCreateResources());
+
+		assertThat(binder.getExtendedConsumerProperties("input").isAutoCreateResources()).isFalse();
+		assertThat(binder.getExtendedConsumerProperties("custom-in").isAutoCreateResources()).isFalse();
 	}
 
+	@EnableAutoConfiguration
 	@EnableBinding(CustomTestSink.class)
 	public static class PubSubTestBindings {
 
-		@StreamListener(CustomTestSink.INPUT)
+		@StreamListener("input")
 		public void process(String payload) {
 			System.out.println(payload);
 		}
--- a/spring-cloud-gcp-pubsub-stream-binder/src/test/java/org/springframework/cloud/gcp/stream/binder/pubsub/properties/PubSubExtendedBindingsPropertiesTest.java
+++ b/spring-cloud-gcp-pubsub-stream-binder/src/test/java/org/springframework/cloud/gcp/stream/binder/pubsub/properties/PubSubExtendedBindingsPropertiesTest.java
@@ -1,0 +1,63 @@
+package org.springframework.cloud.gcp.stream.binder.pubsub.properties;
+
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.autoconfigure.EnableAutoConfiguration;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.cloud.gcp.stream.binder.pubsub.PubSubMessageChannelBinder;
+import org.springframework.cloud.stream.annotation.EnableBinding;
+import org.springframework.cloud.stream.annotation.Input;
+import org.springframework.cloud.stream.annotation.StreamListener;
+import org.springframework.cloud.stream.binder.BinderFactory;
+import org.springframework.cloud.stream.messaging.Sink;
+import org.springframework.context.ConfigurableApplicationContext;
+import org.springframework.messaging.MessageChannel;
+import org.springframework.messaging.SubscribableChannel;
+import org.springframework.test.context.junit4.SpringRunner;
+
+@RunWith(SpringRunner.class)
+@EnableAutoConfiguration
+@EnableConfigurationProperties(PubSubExtendedBindingProperties.class)
+@SpringBootTest(
+    webEnvironment = SpringBootTest.WebEnvironment.NONE,
+    properties = {
+				"spring.cloud.stream.gcp.pubsub.default.consumer.auto-create-resources=false",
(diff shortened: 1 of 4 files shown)


spring-cloud-gcp-pubsub-stream-binder/src/test/java/org/springframework/cloud/gcp/stream/binder/pubsub/properties/PubSubExtendedBindingsPropertiesTest.java
```diff
@@ -0,0 +1,69 @@
...
+import org.springframework.test.context.junit4.SpringRunner;
+
+@RunWith(SpringRunner.class)
+@EnableAutoConfiguration
+@SpringBootTest(
+    webEnvironment = SpringBootTest.WebEnvironment.NONE,
+    properties = {
+				"spring.cloud.stream.gcp.pubsub.default.consumer.auto-create-resources=false",
+		})
+public class PubSubExtendedBindingsPropertiesTest {
```
artembilan: Our convention to call test classes with the `...Tests` suffix.

spring-cloud-gcp-pubsub-stream-binder/src/test/java/org/springframework/cloud/gcp/stream/binder/pubsub/properties/PubSubExtendedBindingsPropertiesTest.java
```diff
@@ -0,0 +1,69 @@
...
+	@Autowired
+	private ConfigurableApplicationContext context;
+
+	@Test
+	public void testExtendedProperties() {
+		BinderFactory binderFactory = context.getBeanFactory().getBean(BinderFactory.class);
+		PubSubMessageChannelBinder binder = (PubSubMessageChannelBinder) binderFactory.getBinder("pubsub", MessageChannel.class);
+
+		// Print out the default values; they should both be false.
+		System.out.println("default for INPUT: "
```
artembilan: Please, no SOUT nowhere in the project, unless you override it somehow do not print into the console.

artembilan:
> While the default value is propagated, it doesn't seem like you can override the default values with additional properties being set.

Is your point that when we have a configuration like this:
```
spring.cloud.stream.gcp.pubsub.default.consumer.auto-create-resources=false
spring.cloud.stream.gcp.pubsub.bindings.input.consumer.auto-create-resources=true
```

The value is set to `false` anyway?

Looks like this test in Kafka Binder confirms that everything is OK: https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/blob/master/spring-cloud-stream-binder-kafka/src/test/java/org/springframework/cloud/stream/binder/kafka/integration/KafkaBinderExtendedPropertiesTest.java#L55

dzou:
> > While the default value is propagated, it doesn't seem like you can override the default values with additional properties being set.
> 
> Is your point that when we have a configuration like this:
> 
> ```
> spring.cloud.stream.gcp.pubsub.default.consumer.auto-create-resources=false
> spring.cloud.stream.gcp.pubsub.bindings.input.consumer.auto-create-resources=true
> ```
> The value is set to `false` anyway?
> 
> Looks like this test in Kafka Binder confirms that everything is OK: https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/blob/master/spring-cloud-stream-binder-kafka/src/test/java/org/springframework/cloud/stream/binder/kafka/integration/KafkaBinderExtendedPropertiesTest.java#L55

Yes that is correct. I saw that too and was unsure why this was not working for us. I was wondering if there was any other detail we may have missed which may cause this problem; similar to the detail where we were missing `MergableProperties`.

artembilan:
Hm. Let's ping @sobychacko as original contributor for this `default` feature!

spring-cloud-gcp-pubsub-stream-binder/src/test/java/org/springframework/cloud/gcp/stream/binder/pubsub/properties/PubSubExtendedBindingsPropertiesTest.java
dzou: Done.

spring-cloud-gcp-pubsub-stream-binder/src/test/java/org/springframework/cloud/gcp/stream/binder/pubsub/properties/PubSubExtendedBindingsPropertiesTests.java
```diff
@@ -36,14 +37,8 @@ public void testExtendedProperties() {
...
 
-		// Print out the default values; they should both be false.
-		System.out.println("default for INPUT: "
-				+ binder.getExtendedConsumerProperties("input").isAutoCreateResources());
-		System.out.println("default for CUSTOM INPUT: "
-				+ binder.getExtendedConsumerProperties("custom-in").isAutoCreateResources());
-
-		assertThat(binder.getExtendedConsumerProperties("input").isAutoCreateResources()).isFalse();
 		assertThat(binder.getExtendedConsumerProperties("custom-in").isAutoCreateResources()).isFalse();
+		assertThat(binder.getExtendedConsumerProperties("input").isAutoCreateResources()).isTrue();
```
artembilan: Wait. Sounds like everything works as expected now.

What are we missing then? 🤷‍♂️

dzou: Ah not quite, I've modified the test so it can reproduce the problem, however it is still broken.

dzou:
> Hm. Let's ping @sobychacko as original contributor for this `default` feature!

Sounds good. I have modified the test so it now reproduces the issue with the defaults not being overridden.

olegz:
Guys, just an FYI, we're looking at it from the core framework perspective. Don't have an answer at the moment but stay tuned. The issue did come up in other discussions so it has been prioritized

dzou:
> @dzou This issue is resolved in [core](https://github.com/spring-cloud/spring-cloud-stream/pull/1494). Please re-run your tests and verify the issue is fixed on your end.
> cc/ @artembilan

Thank you, will do.

spring-cloud-gcp-pubsub-stream-binder/src/test/java/org/springframework/cloud/gcp/stream/binder/pubsub/properties/PubSubExtendedBindingsPropertiesTests.java
```diff
@@ -0,0 +1,69 @@
...
+public class PubSubExtendedBindingsPropertiesTests {
+
+	private ApplicationContextRunner contextRunner = new ApplicationContextRunner()
+			.withPropertyValues(
+				"spring.cloud.stream.gcp.pubsub.bindings.input.consumer.auto-create-resources=true",
+					"spring.cloud.stream.gcp.pubsub.default.consumer.auto-create-resources=false")
+			.withConfiguration(AutoConfigurations.of(
+					GcpPubSubEmulatorConfiguration.class,
+					GcpContextAutoConfiguration.class,
+					GcpPubSubAutoConfiguration.class,
```
meltsufin: Do you really need these first three? It will try to configure credentials and such, which I don't think is necessary for this test.

dzou: I am investigating if these can be removed. The test seems to need the beans PubSubAdmin and pubSubChannelProvisioner. I think these are needed either by the BinderFactory or by the bindings created in PubSubBindingsTestConfiguration.

I won't be able to keep them in order for the test to run on Travis.

meltsufin: You should be able to just provide mocks for those. Put the mocks into the application context.

dzou: So it looks like there is a little bit more to it, turns out I also need the PubSubTemplate binding because it looks like my test bindings (i.e. the CustomTestSink) are being registered by the Binding service. Doing a plain mock here doesn't work since it needs to keep track of the channels that were created. I'll dig deeper and get back to you on a solution.

dzou:
Hi guys, I made the changes to make the test pass on Travis again. PTAL.

PR #1080, opened 2018-10-03
Add Spanner sample integration test and consolidate Spanner sample to 1 main class.

### LINKED ISSUE #1084: Add Spanner Sample app test to CI
The Spanner sample app does not have an integration test. Add a test for the Spanner Sample to be run in CI.

dzou:
Adding tests for #705. Fixes #1084 
* Added simple integration test for Spanner sample
* Split the Spanner sample demos to be executable from 2 different Maven profiles. Prior to this, it was not possible to run the examples easily because there are two Main classes in the package. You would have to comment out "@SpringBootApplication" in one of the main files to run the other.

Now either demo may be run with `mvn spring-boot:run -P spanner-repository-example` or `mvn spring-boot:run -P spanner-template-example`.


.travis.yml
```diff
@@ -21,7 +21,7 @@ before_install:
         openssl aes-256-cbc -K $encrypted_1ef8dfbdb114_key -iv $encrypted_1ef8dfbdb114_iv -in travis.tar.gz.enc -out travis.tar.gz -d;
         tar -xzf travis.tar.gz;
         INTEGRATION_TEST_FLAGS="${INTEGRATION_TEST_FLAGS} -Dit.spanner=true -Dit.storage=true -Dit.config=true -Dit.pubsub=true -Dit.logging=true
-            -Dit.cloudsql=true -Dit.datastore=true
+            -Dit.cloudsql=true -Dit.datastore=true -Dit.spanner=true
```
ChengyuanZhao: Hey this is already on the line right above

spring-cloud-gcp-samples/spring-cloud-gcp-data-spanner-sample/src/test/resources/application-test.properties
```diff
@@ -0,0 +1,2 @@
+spring.cloud.gcp.spanner.instance-id=integration-instance
+spring.cloud.gcp.spanner.database=spanner-sample-test-db
```
ChengyuanZhao: I think you might get an error because our draconian checkstyle file requires newlines at the end of files

spring-cloud-gcp-samples/spring-cloud-gcp-data-spanner-sample/src/test/java/com/example/SpannerRepositoryTests.java
```diff
@@ -0,0 +1,60 @@
...
+	@After
+	public void cleanupSpannerTables() {
+		this.tradeRepository.deleteAll();
+		this.traderRepository.deleteAll();
+	}
+
+	@Test
+	public void verifySpannerApplicationLoadsRows() {
+		// Verifies that there are 3 items in the Trader Repository table,
+		assertThat(traderRepository.count()).isEqualTo(3);
```
meltsufin: What about testing `Trades`?
Also, you probably want to test `http://localhost:8080/trades` and see if it returns what is expected.
Essentially, the test should validate as much as possible what the user would do and see in the description of [steps](https://github.com/spring-cloud/spring-cloud-gcp/tree/master/spring-cloud-gcp-samples/spring-cloud-gcp-data-spanner-sample).

spring-cloud-gcp-samples/spring-cloud-gcp-data-spanner-sample/README.adoc
```diff
@@ -23,6 +23,14 @@ link:src/main/resources/application.properties[application.properties] file and
 objects filled with some demo data and persists them to the Cloud Spanner table created earlier.
 These POJOs are then retrieved and their contents are printed in the browser.
 
+.. To run the SpannerRepositoryExample:
+
+  mvn spring-boot:run -P spanner-repository-example
```
meltsufin: I'm not a big fan of using profiles in this situation. We should rather control what runs based on a command-line argument passed to the main class. The is exactly what a `CommandLineRunner` is good for! :)
Alternatively, we can even split the sample into two modules, but I don't think that's worthwhile right now.

spring-cloud-gcp-samples/spring-cloud-gcp-data-spanner-sample/src/test/java/com/example/SpannerRepositoryTests.java
```diff
@@ -0,0 +1,60 @@
...
+import org.springframework.cloud.gcp.data.spanner.core.admin.SpannerSchemaUtils;
+import org.springframework.test.context.ActiveProfiles;
+import org.springframework.test.context.TestPropertySource;
+import org.springframework.test.context.junit4.SpringRunner;
+
+@RunWith(SpringRunner.class)
+@TestPropertySource("classpath:application-test.properties")
+@SpringBootTest(classes = {SpannerRepositoryExample.class})
+@ActiveProfiles(profiles = "repository-example")
+public class SpannerRepositoryTests {
```
meltsufin: What about `SpannerTemplateTests`?

spring-cloud-gcp-samples/spring-cloud-gcp-data-spanner-sample/src/test/java/com/example/SpannerRepositoryTests.java
ChengyuanZhao: @meltsufin  Regarding "What about testing Trades". We aren't testing if Spring Data Spanner works; that is heavily tested in its own module. We're just testing if the sample ran, and I think this accomplishes that.

spring-cloud-gcp-samples/spring-cloud-gcp-data-spanner-sample/README.adoc
ChengyuanZhao: Could you elaborate on why to not to use profiles? It seems like it could solve the problem here.

I don't think combining the two samples into a single main function is a good, because it would be a rather big sample program in that case and is contrary to the simplicity and the approachable feel of a sample.

I also don't think we should break this into two samples because nearly all of the config is shared between the two, and the repository even uses a template underneath.

spring-cloud-gcp-samples/spring-cloud-gcp-data-spanner-sample/src/test/java/com/example/SpannerRepositoryTests.java
meltsufin: I think redundancy in the tests is OK. Given that we already start up the application, it's not all that much more execution cost to add these tests. It will also ensure that we don't regress the sample code itself, like enabling Spring Data REST.

spring-cloud-gcp-samples/spring-cloud-gcp-data-spanner-sample/README.adoc
meltsufin: Having more than one main class in a Spring Boot application is not generally recommended. Even if you do insist on doing it here, the normal way to control which main class runs is to use `-Dstart-class`, not profiles.

spring-cloud-gcp-samples/spring-cloud-gcp-data-spanner-sample/src/test/java/com/example/SpannerRepositoryTests.java
dzou: Added.

spring-cloud-gcp-samples/spring-cloud-gcp-data-spanner-sample/README.adoc
dzou: Done. Switched to use command line args + consolidate to 1 main. Spring seemed to have problems with having 2 main classes annotated with `@SpringBootApplication`; this may have been necessary.

spring-cloud-gcp-samples/spring-cloud-gcp-data-spanner-sample/src/test/java/com/example/SpannerRepositoryTests.java
dzou: Done. I improved tests to verify the actions done in each demo.

dzou:
I have consolidated to 1 main method which accepts flags and also increased what the test verifies. PTAL.

spring-cloud-gcp-samples/spring-cloud-gcp-data-spanner-sample/src/test/resources/application-test.properties
```diff
@@ -0,0 +1,2 @@
+spring.cloud.gcp.spanner.instance-id=integration-instance
```
meltsufin: Can we stick to the same configuration as described in the actual sample README:
```
spring.cloud.gcp.spanner.instance-id=spring-demo
spring.cloud.gcp.spanner.database=trades
```

spring-cloud-gcp-samples/spring-cloud-gcp-data-spanner-sample/src/main/java/com/example/SpannerExampleDriver.java
```diff
@@ -0,0 +1,52 @@
...
+	@Autowired
+	private SpannerTemplateExample spannerTemplateExample;
+
+	public static void main(String[] args) {
+		System.out.println(Arrays.toString(args));
+		SpringApplication.run(SpannerExampleDriver.class, args);
+	}
+
+	@Bean
+	@Profile("!test")
```
meltsufin: What's this profile for?

spring-cloud-gcp-samples/spring-cloud-gcp-data-spanner-sample/src/test/java/com/example/SpannerRepositoryTests.java
```diff
@@ -0,0 +1,113 @@
...
+import org.springframework.test.context.ActiveProfiles;
+import org.springframework.test.context.TestPropertySource;
+import org.springframework.test.context.junit4.SpringRunner;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.hamcrest.Matchers.is;
+import static org.junit.Assume.assumeThat;
+
+@RunWith(SpringRunner.class)
+@ActiveProfiles("test")
```
meltsufin: Why is this needed?

spring-cloud-gcp-samples/spring-cloud-gcp-data-spanner-sample/src/test/java/com/example/SpannerRepositoryTests.java
```diff
@@ -0,0 +1,113 @@
...
+@TestPropertySource("classpath:application-test.properties")
```
meltsufin: I would stick to the same properties as in the main.

spring-cloud-gcp-samples/spring-cloud-gcp-data-spanner-sample/src/main/java/com/example/SpannerExampleDriver.java
dzou: The `@SpringBootTest` annotation picks up any ApplicationRunner or CommandLineRunner bean and automatically runs it. However, in test we want to control when it gets run; we want to run the sample after we made some assertions. This allows us to first assert the Spanner tables are empty, run the sample, then verify the rows were created. 

`@Profiles` lets us exclude the bean from the Test if the `Test` profile is set, so that the bean does not automatically get run.

Reference: https://stackoverflow.com/questions/29344313/prevent-application-commandlinerunner-classes-from-executing-during-junit-test

spring-cloud-gcp-samples/spring-cloud-gcp-data-spanner-sample/src/test/java/com/example/SpannerRepositoryTests.java
dzou: See previous comment about profiles.

PR #1081, opened 2018-10-04
Spanner @Column schema options

### LINKED ISSUE #1078: Spanner @Column annotation setting for NOT NULL column
This can't have an affect on the Java POJO it annotates, but merely for the schema-generation to mark those columns as not null.

ChengyuanZhao:
Fixes #1078. Spanner `@Column` schema settings


spring-cloud-gcp-docs/src/main/asciidoc/spanner.adoc
```diff
@@ -256,7 +256,11 @@ Read https://cloud.google.com/spanner/docs/schema-and-data-model#primary_keys[Sp
 
 All accessible properties on POJOs are automatically recognized as a Cloud Spanner column.
 Column naming is generated by the `PropertyNameFieldNamingStrategy` by default defined on the `SpannerMappingContext` bean.
-The `@Column` annotation optionally provides a different column name than that of the property.
+The `@Column` annotation optionally provides a different column name than that of the property and some other settings:
+- `name` is the optional name of the column
+- `generateSchemaMaxLength` specifies for `ARRAY` and `BYTES` columns the maximum length. This setting is only used when generating DDL schema statements based on domain types.
+- `nullable` specifies if the column is created as `NOT NULL`. This setting is only used when generating DDL schema statements based on domain types.
+- `columnTypeCode` is the Cloud Spanner column type you can optionally specify. If this is not specified then a compatible column type is inferred from the Java property type.
```
dmitry-s: maybe just name it columnType for brevity? It is clear that it should be TypeCode from parameter type

spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/core/convert/TestEntities.java
```diff
@@ -54,10 +55,14 @@
...
 		boolean booleanField;
 
+		// This long is forced to be stored as a String for testing
+		@Column(columnTypeCode = TypeCode.STRING)
 		long longField;
 
 		double doubleField;
 
+		// This double array is forced to be stored as a String for testing
+		@Column(columnTypeCode = TypeCode.STRING)
```
dmitry-s: What happens if user marks a property to be stored in incompatible column type? Like property of type double[] to be stored in TypeCode.BOOL
Should have a test for that?

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/mapping/Column.java
```diff
@@ -41,5 +43,31 @@
...
+	 */
+	boolean nullable() default true;
+
+	/**
+	 * Optionally directly specify the column type in Cloud Spanner. For ARRAY columns
+	 * this refers to type of the item the array holds. If this is not specified then it
+	 * is inferred from the Java property type.
+	 * @return The user-specified column item type.
+	 */
+	TypeCode columnTypeCode() default TypeCode.TYPE_CODE_UNSPECIFIED;
```
meltsufin: What about a simple `spannerType` for a name?

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/mapping/Column.java
```diff
@@ -41,5 +43,31 @@
...
+
+	/**
+	 * The maximum length of the column in Cloud Spanner terms. For example, for STRING
+	 * columns this refers to the number of characters. For BYTES columns this refers to
+	 * the number of bytes. This setting is only used when generating schema from an
+	 * entity class. A setting of 0 indicates an unlimited maximum length.
+	 *
+	 * @return the maximum length for the column
+	 */
+	long generateSchemaMaxLength() default 0;
```
meltsufin: `spannerTypeMaxLength`?
Maybe `-1` to represent unlimited length, or use `null` with `Long`.

spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/core/convert/TestEntities.java
ChengyuanZhao: We can't stop them from doing this, and what would happen is that during mapping/conversion we would get errors saying there isn't a converter provided that can handle the situation, and we do have explicit exceptions thrown with helpful messages for that case.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/mapping/Column.java
ChengyuanZhao: Unfortunately `null` isn't considered a constant value and cannot be used in enum defaults. I can use -1

spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/core/convert/TestEntities.java
ChengyuanZhao: added a test!

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/mapping/Column.java
meltsufin: OK, what about the attribute name suggestion?

ChengyuanZhao: The `spannerType` ? I made that change...

meltsufin: changing this one to `spannerTypeMaxLength`

ChengyuanZhao: @meltsufin ah ok, also removed the "generateSchema". PTAL

spring-cloud-gcp-docs/src/main/asciidoc/spanner.adoc
```diff
@@ -256,7 +256,11 @@ Read https://cloud.google.com/spanner/docs/schema-and-data-model#primary_keys[Sp
 
 All accessible properties on POJOs are automatically recognized as a Cloud Spanner column.
 Column naming is generated by the `PropertyNameFieldNamingStrategy` by default defined on the `SpannerMappingContext` bean.
-The `@Column` annotation optionally provides a different column name than that of the property.
+The `@Column` annotation optionally provides a different column name than that of the property and some other settings:
+- `name` is the optional name of the column
+- `spannerTypeMaxLength` specifies for `ARRAY` and `BYTES` columns the maximum length. This setting is only used when generating DDL schema statements based on domain types.
```
meltsufin: String too?

PR #1090, opened 2018-10-09
Update release process instructions

meltsufin:


RELEASING.adoc
```diff
@@ -1,31 +1,28 @@
...
-. Uncomment the commented out modules.
-
-. Revert the versions using the https://www.mojohaus.org/versions-maven-plugin/revert-mojo.html[mvn versions:revert] command.
-
-    $ ./mvnw versions:revert
+. Make sure that the release tag was automatically added by the release process.
 
 . Change the `TRACKING_HEADER_PROJECT_VERSION` in link:spring-cloud-gcp-core/src/main/java/org/springframework/cloud/gcp/core/UsageTrackingHeaderProvider.java[`UsageTrackingHeaderProvider`] back to `BUILD-SNAPSHOT`.
+
+. Make sure that the link:https://github.com/spring-io/initializr/blob/master/initializr-service/src/main/resources/application.yml[Spring Boot Initializr] is updated to support the new version of Spring Cloud GCP.
```
artembilan: I think since we are now the part of the release train this should be done automatically from the Spring Cloud BOM level.
Therefore this note is not necessary anymore.

/CC @spencergibb

RELEASING.adoc
```diff
@@ -1,31 +1,28 @@
 = Spring Cloud GCP release process
 
-The actual release process is triggered by the release admin.
+The actual release process is triggered by the release admin and happens together with the rest of the Spring Cloud release.
 In order to simplify that process, here is a list of prerequisites to be performed before asking the release admin to run the release scripts.
 
 . Manually change the value of `TRACKING_HEADER_PROJECT_VERSION` in link:spring-cloud-gcp-core/src/main/java/org/springframework/cloud/gcp/core/UsageTrackingHeaderProvider.java[`UsageTrackingHeaderProvider`] to the version being released from the interim `BUILD-SNAPSHOT` value.
 
-. Change the value of the `<version>` elements of the `pom.xml` of all modules to the name of the version being released from the interim `BUILD-SNAPSHOT` values. This can be done using the https://www.mojohaus.org/versions-maven-plugin/set-mojo.html[mvn versions:set] command.
+. Ensure that modules that shouldn't be released have the `maven-deploy-plugin` configuration set to `<skip>true</skip>`.
```
artembilan: I believe this has to be done in the boundaries of the `central` profile as @spencergibb mention on the call.

Or am I missing anything?

RELEASING.adoc
meltsufin: The problem is that our 1.0 release wasn't part of the Spring Cloud release train. So, for the time being, we still need the Initializr to include our BOM in the generated POM. It also makes it more convenient to change versions for user app. Nothing else in Spring Cloud really depends on Spring Cloud GCP. So, I think we're OK to keep it this way.

RELEASING.adoc
meltsufin: I guess that will allow us to make the code change now, instead of flipping back and forth for GA releases. I'll make the change on this PR.

PR #1095, opened 2018-10-11
Add tests for logging sample.

### LINKED ISSUE #1094: Create test for logging sample app.
Create a test for the logging sample app.

In reference to #705

dzou:
Add test for logging sample.
Fixes #1094
In reference to #705


spring-cloud-gcp-samples/spring-cloud-gcp-logging-sample/src/test/java/com.example/LoggingSampleApplicationTests.java
```diff
@@ -0,0 +1,97 @@
...
+	public void testLogRecordedInStackDriver() {
+		String url = String.format("http://localhost:%s/log", port);
+		String traceHeader = "gcp-logging-test-" + Instant.now().toEpochMilli();
+
+		HttpHeaders headers = new HttpHeaders();
+		headers.add("x-cloud-trace-context", traceHeader);
+		testRestTemplate.exchange(url, HttpMethod.GET, new HttpEntity<>(headers), String.class);
+
+		String logFilter = String.format(LOG_FILTER, traceHeader);
+		Page<LogEntry> logEntryPage = logClient.listLogEntries(
```
meltsufin: I'm surprised that this works immediately. I would expect that you would need to wait a little bit for the trace to appear in the service.

spring-cloud-gcp-samples/spring-cloud-gcp-logging-sample/src/test/java/com.example/LoggingSampleApplicationTests.java
```diff
@@ -0,0 +1,97 @@
...
+				Logging.EntryListOption.filter(logFilter));
+
+		ImmutableList<LogEntry> logEntries = ImmutableList.copyOf(logEntryPage.iterateAll());
+		List<String> logContents = logEntries.stream()
+				.map(logEntry -> ((StringPayload) logEntry.getPayload()).getData())
+				.collect(Collectors.toList());
+
+		assertThat(logContents).containsExactlyInAnyOrder(
+				"This line was written to the log.",
```
meltsufin: Can we also verify the log name?

dzou:
Regarding the Travis build - for some reason the test works locally but not on Travis. I was comparing the differences between Travis runs vs. local runs and I think I narrowed it down to a different Maven profile being used by #636 for the previous logging integration test.

I consolidated the old logging integration test to not require 2 maven profiles in PR #1099, maybe this will resolve the problem.

meltsufin:
@dzou Why did you rename the logging config file?

Also, unrelated, but we generally don't put a period at the end of PR and commit titles.

dzou:
> @dzou Why did you rename the logging config file?
> 
> Also, unrelated, but we generally don't put a period at the end of PR and commit titles.

This was done as part of investigating why the test works locally but not on Travis. I'm testing a theory to see if the naming convention affects the test. According to [this documentation](https://logback.qos.ch/manual/configuration.html), logback files must be named logback.xml or logback-test.xml to be automatically picked up on the class path. I think naming it logback-spring.xml might be causing me problems.

And noted, thanks for heads up on the period convention.

dzou:
Hmm still stumped on this one. I tried variations of testing the flush level theory such as setting `STACKDRIVER_LOG_FLUSH_LEVEL` or changing the sample app to log at error instead of info. Maybe it's time to move on :(

Would you be open to keeping this a manual test? Maybe we can check for a new property like `it.manual` for manually runned tests.

dzou:
> Definitely move on for now, but keep the PR open. This may benefit from another pair of eyes.

Sounds good. I have reverted the logback name change add made some minor changes that might help in making progress like increased timeout.

PR #1105, opened 2018-10-15
Datastore DELETE query methods part-tree

### LINKED ISSUE #996: Datastore support DELETE query methods
perhaps using performTransaction to do the get then delete for the user.

ChengyuanZhao:
fixes #996
Datastore queries must have an ancestor when used inside a transaction, and that is not compatible with the format of part-tree queries. This implementation does the query and delete in separate requests.


spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/PartTreeDatastoreQuery.java
```diff
@@ -81,12 +81,7 @@ public PartTreeDatastoreQuery(DatastoreQueryMethod queryMethod,
...
 
 	private void validateAndSetFilterParts() {
-		if (this.tree.isDelete()) {
-			throw new UnsupportedOperationException(
-					"Delete queries are not supported in Cloud Datastore: "
-							+ this.queryMethod.getName());
-		}
-		else if (this.tree.isDistinct()) {
+		if (this.tree.isDistinct()) {
 			throw new UnsupportedOperationException(
```
ChengyuanZhao: I distinctly remember doing something to support this?

spring-cloud-gcp-data-datastore/src/test/java/org/springframework/cloud/gcp/data/datastore/it/DatastoreIntegrationTests.java
```diff
@@ -187,4 +196,12 @@ public void embeddedCollectionTest() {
 		assertEquals(treeCollection, loaded);
 	}
 
+	private void waitUntilTrue(Supplier<Boolean> condition) throws InterruptedException {
```
ChengyuanZhao: awaitility ?

spring-cloud-gcp-docs/src/main/asciidoc/datastore.adoc
```diff
@@ -652,6 +652,13 @@ Cloud Datastore only supports filter components joined by AND, and the following
 
 After writing a custom repository interface specifying just the signatures of these methods, implementations are generated for you and can be used with an auto-wired instance of the repository.
 
+Delete queries are also supported. For example, query methods such as deleteByAction or removeByAction delete entities found by findByAction.
```
ChengyuanZhao: Add note that this won't work in a transaction.

ChengyuanZhao: make issue to note what actions in template can't work in transaction

ChengyuanZhao: mention that it happens as 2 requests , a key retrieval.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/PartTreeDatastoreQuery.java
ChengyuanZhao: I remember now. We decided to NOT support projection queries in part-tree because of how cumbersome it is to set up and use from the user's end. Without projection queries we can no longer set the distinct and it makes no sense because the Key is always selected and is always distinct.

ChengyuanZhao:
@meltsufin @dmitry-s PTAL, changes made from discussion

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/PartTreeDatastoreQuery.java
```diff
@@ -108,36 +103,51 @@ else if (this.tree.isDistinct()) {
 	public Object execute(Object[] parameters) {
 		Supplier<StructuredQuery.Builder<?>> queryBuilderSupplier = StructuredQuery::newKeyQueryBuilder;
 		Function<T, ?> mapper = Function.identity();
-		Collector<?, ?, ?> collector;
+		Collector<?, ?, ?> collector = Collectors.toList();
+		Class returnedType = this.queryMethod.getReturnedObjectType();
 
-		if (this.tree.isCountProjection()) {
+		boolean returnedTypeisNumber = Number.class.isAssignableFrom(returnedType)
```
meltsufin: returnedTypeIsNumber

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/PartTreeDatastoreQuery.java
```diff
@@ -108,36 +103,51 @@ else if (this.tree.isDistinct()) {
...
-				: StreamSupport.stream(results.spliterator(), false).map(mapper)
+		Object result = rawResults == null ? null
+				: StreamSupport.stream(rawResults.spliterator(), false).map(mapper)
 						.collect(collector);
 
-		return this.tree.isExistsProjection() || this.tree.isCountProjection()
-				|| results == null
-						? returned
+		if (this.tree.isDelete()) {
+			if (returnedTypeisNumber) {
```
meltsufin: Why does requested return type affect which `deleteAll*` method you call?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/PartTreeDatastoreQuery.java
```diff
@@ -108,36 +103,51 @@ else if (this.tree.isDistinct()) {
...
+				this.datastoreTemplate.deleteAllById(rawResults, this.entityType);
+			}
+			else {
+				this.datastoreTemplate.deleteAll(rawResults);
+			}
+		}
+
+		return this.tree.isExistsProjection() || isCountingQuery || result == null
```
meltsufin: I think there is way too much conditional logic in this method. There are like a dozen different if statements here. Can this method be broken up into smaller ones or reorganized to make it more readable and easier to maintain going forward?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/PartTreeDatastoreQuery.java
ChengyuanZhao: If the return type is a count, then only Keys are required in the original query. There is a delete for that.

If the return type is the entities deleted, then we must query the entire entities anyway. There is also a delete function for that.

ChengyuanZhao:
@meltsufin made requested changes you had just now. PTAL.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/PartTreeDatastoreQuery.java
```diff
@@ -108,38 +103,58 @@ else if (this.tree.isDistinct()) {
...
 						.collect(collector);
 
-		return this.tree.isExistsProjection() || this.tree.isCountProjection()
-				|| results == null
-						? returned
-						: this.datastoreTemplate.getDatastoreEntityConverter()
-								.getConversions().convertOnRead(returned,
-										this.queryMethod.getCollectionReturnType(),
-										this.queryMethod.getReturnedObjectType());
+		deleteFoundEntities(returnedTypeIsNumber, rawResults);
```
meltsufin: It's really odd to see this method call here because it makes it seem like you unconditionally always delete found entities. I know that's not the case, but maybe you should move the `this.tree.isDelete` here.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/PartTreeDatastoreQuery.java
```diff
@@ -108,38 +103,58 @@ else if (this.tree.isDistinct()) {
...
+				: convertResultCollection(result);
+	}
+
+	private Object convertResultCollection(Object result) {
+		return this.datastoreTemplate.getDatastoreEntityConverter().getConversions()
+				.convertOnRead(result, this.queryMethod.getCollectionReturnType(),
+						this.queryMethod.getReturnedObjectType());
+	}
+
+	private void deleteFoundEntities(boolean returnedTypeIsNumber, Iterable rawResults) {
```
meltsufin: I would rename `returnTypeIsNumber` here to something like `rawResultsAreKeys`.

ChengyuanZhao: I don't think that's a good idea, because this boolean is actually based on the return type. Yes this implies that queries are key-queries, but then again if you rename it as you propose then you're "reverse-implying" that the return type is a number, which I think is difficult to read.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/PartTreeDatastoreQuery.java
ChengyuanZhao: @meltsufin change made, PTAL

Issue #1092, opened 2018-10-11
Datastore @Reference column-stored key POJO relationships


ChengyuanZhao:
This is distinct from the other POJO relationship ticket we have that is based on Ancestors.

This one specifically is about using `@Reference` and is identical in concept to the Objectify version of relationships where the key[s] of the other object are stored as an explicit column in the Kind.

ChengyuanZhao:
This is the other POJO relationships issue that is based on the ancestors feature: https://github.com/spring-cloud/spring-cloud-gcp/issues/1007

ChengyuanZhao:
Because this code touches nearly all the same files as the descendants work ,this is awaiting that.

ChengyuanZhao:
Because this uses nearly all the same logic as `@Descendants` , this one can proceed in lockstep with #1007.

The parts can be implemented in lockstep with that other issue:

- [ ] read and unit tests

- [ ] write and unit tests

- [ ] integration tests

- [ ] ref docs

PR #1114, opened 2018-10-17
Datastore @Reference read

ChengyuanZhao:
related to #1092 

This is the read functionality and unit tests for `@Reference`.

### DIFF
--- a/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
+++ b/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
@@ -243,32 +243,12 @@ public class DatastoreTemplate implements DatastoreOperations {
 		T convertedObject = this.datastoreEntityConverter.read(entityClass, entity);
 		results.add(convertedObject);
 
-		datastorePersistentEntity
-				.doWithDescendantProperties(descendantPersistentProperty -> {
+		resolveDescendantProperties(datastorePersistentEntity, entity, convertedObject);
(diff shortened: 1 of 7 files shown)


ChengyuanZhao:
@meltsufin @dmitry-s PTAL. This one has relatively few changes because it is very similar to `@Descendant` read

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
```diff
@@ -228,12 +231,13 @@ private Entity convertToEntityForSave(Object entity) {
...
+				entityClass,
 				results, datastorePersistentEntity, entity));
 		return results;
 	}
 
-	private <T> void convertEntityResolveDescendants(Class<T> entityClass,
+	private <T> void convertEntityResolveDescendantsAndReferences(Class<T> entityClass,
 			List<T> results, DatastorePersistentEntity datastorePersistentEntity,
 			BaseEntity entity) {
 		T convertedObject = this.datastoreEntityConverter.read(entityClass, entity);
```
meltsufin: I would break this long method up into two private methods - one for descendants and one for references.

ChengyuanZhao:
@meltsufin please see the attached ticket. This is not going to be ready to use until all parts are implemented so doc updates were intentionally not done now and will wait for the end.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
ChengyuanZhao: @meltsufin change made. PTAL

PR #1117, opened 2018-10-17
datastore rest repository

dmitry-s:


spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
```diff
@@ -184,12 +186,28 @@ public long count(Class<?> entityClass) {
 
 	@Override
 	public <T> Collection<T> findAll(Class<T> entityClass) {
-		return convertEntitiesForRead(
-				this.datastore.run(Query.newEntityQueryBuilder()
-						.setKind(this.datastoreMappingContext
-								.getPersistentEntity(entityClass).kindName())
-						.build()),
-				entityClass);
+		return findAll(entityClass, -1, -1, null);
```
ChengyuanZhao: Is it ok to put special cases for these trick params (-1) in a public method on the interface? what about `OptionalLong` ?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
```diff
@@ -184,12 +186,28 @@ public long count(Class<?> entityClass) {
...
+	}
+
+	@Override
+	public <T> Collection<T> findAll(Class<T> entityClass, int limit, int offset,
+			Sort sort) {
+		DatastorePersistentEntity<?> persistentEntity = this.datastoreMappingContext.getPersistentEntity(entityClass);
+		EntityQuery.Builder builder = Query.newEntityQueryBuilder()
+				.setKind(persistentEntity.kindName());
+		if (limit > 0) {
```
ChengyuanZhao: Is there a way to use the code from `PartTreeDatastoreQuery` or to merge the code here in? That one already supports limiting and sorting, and I don't think we should have logic for building a query with sort and limits in 2 places.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreOperations.java
```diff
@@ -160,6 +163,17 @@
...
+	/**
+	 * Get all the entities of the given domain type applying limit, offset and sort
+	 * @param entityClass the domain type to get.
+	 * @param limit max number of entities to return, null means no limit
+	 * @param offset number of entities to skip, null means no offset
+	 * @param sort sort order and properties
+	 * @param <T> the type param of the domain type.
+	 * @return the entities that were found.
+	 */
+	<T> Collection<T> findAll(Class<T> entityClass, Integer limit, Integer offset, Sort sort);
```
meltsufin: Why not use `Pageable` instead of the 3 parameters?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/PartTreeDatastoreQuery.java
```diff
@@ -111,7 +110,7 @@ public Object execute(Object[] parameters) {
 		Collector<?, ?, ?> collector;
 
 		if (this.tree.isCountProjection()) {
-			collector = Collectors.reducing(0, e -> 1, Integer::sum);
+			collector = Collectors.counting();
```
meltsufin: These changes are going to clash with the other PR. @ChengyuanZhao

spring-cloud-gcp-samples/spring-cloud-gcp-data-datastore-sample/pom.xml
```diff
@@ -18,7 +18,17 @@
 			<groupId>org.springframework.cloud</groupId>
 			<artifactId>spring-cloud-gcp-starter-data-datastore</artifactId>
 		</dependency>
-  </dependencies>
+		<dependency>
+			<groupId>org.springframework.boot</groupId>
+			<artifactId>spring-boot-starter-data-rest</artifactId>
+		</dependency>
+        <dependency>
```
meltsufin: Spacing seems off.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/PartTreeDatastoreQuery.java
ChengyuanZhao: I think it should be OK if either one goes in first. It will take some manual merging but I don't think anything is in contradictory.

spring-cloud-gcp-samples/spring-cloud-gcp-data-datastore-sample/pom.xml
meltsufin: I think tabs is what Spring Team prefers, but I don't think they're too particular about it in the xml files.

PR #1123, opened 2018-10-19
Add Sample Integration Test for SQL sample app.

dzou:
Adds integration test for the SQL sample app.

### DIFF
--- a/spring-cloud-gcp-samples/spring-cloud-gcp-sql-sample/src/main/java/com/example/WebController.java
+++ b/spring-cloud-gcp-samples/spring-cloud-gcp-sql-sample/src/main/java/com/example/WebController.java
@@ -17,7 +17,6 @@
 package com.example;
 
 import java.util.List;
-import java.util.Map;
 import java.util.stream.Collectors;
 
 import org.springframework.jdbc.core.JdbcTemplate;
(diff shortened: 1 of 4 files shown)


dzou:
@meltsufin 

To follow up on why I added application-test.properties - I saw that the [JPA sample](https://github.com/spring-cloud/spring-cloud-gcp/blob/master/spring-cloud-gcp-samples/spring-cloud-gcp-data-jpa-sample/src/test/resources/application-test.properties) uses the `spring.cloud.gcp.sql.database-name` property so I thought setting it in Travis wouldn't work since it would override the JPA sample database name.

However, this might be an argument for us to move all SQL sample tests to use a single Database, and then the database name can be set once in the Travis flag. Let me know what you think.

dzou:
> @dzou Notice that in Travis we already pass in `-Dspring.cloud.gcp.sql.instance-connection-name=spring-cloud-gcp-ci:us-central1:testmysql`. So, your setting in `application-test.properties` (`spring.cloud.gcp.sql.instance-connection-name=spring-cloud-gcp-ci:us-central1:testmysql`) will be overwritten on Travis anyway. JPA sample doesn't set `instance-connection-name` in `application-test.properties`.
> Regarding `database-name` and `password` settings. I would also prefer those passed in through Travis. That will of course mean that both samples and all other integration tests should use the same Cloud SQL instance. That could be left to a different PR, but for this one, we should at least remove the `instance-connection-name` from `application-test.properties`.

Ok sounds good, I created #1125 in which I will do this clean up. For now I'll just remove the instance connection property.

meltsufin:
@dzou Feel free to merge.

PR #1124, opened 2018-10-19
Datastore embedded map properties

### LINKED ISSUE #980: Support write/read for Map types in Datastore

ChengyuanZhao:
Fixes #980 
Adds embedded maps that can hold collection and embedded collection value types.

### DIFF
--- a/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/DefaultDatastoreEntityConverter.java
+++ b/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/DefaultDatastoreEntityConverter.java
@@ -16,9 +16,14 @@
 
 package org.springframework.cloud.gcp.data.datastore.core.convert;
 
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+
 import com.google.cloud.datastore.BaseEntity;
 import com.google.cloud.datastore.Value;
 import com.google.cloud.datastore.ValueBuilder;
+import com.google.cloud.datastore.ValueType;
 
 import org.springframework.cloud.gcp.data.datastore.core.mapping.DatastoreDataException;
 import org.springframework.cloud.gcp.data.datastore.core.mapping.DatastoreMappingContext;
@@ -29,6 +34,7 @@ import org.springframework.data.convert.EntityInstantiators;
 import org.springframework.data.mapping.PersistentPropertyAccessor;
 import org.springframework.data.mapping.model.ParameterValueProvider;
 import org.springframework.data.mapping.model.PersistentEntityParameterValueProvider;
+import org.springframework.data.util.TypeInformation;
 
 /**
  * A class for object to entity and entity to object conversions
@@ -60,6 +66,25 @@ public class DefaultDatastoreEntityConverter implements DatastoreEntityConverter
 	@Override
 	public ReadWriteConversions getConversions() {
 		return this.conversions;
+	}
+
+	@Override
+	public <R> Map<String, R> readAsMap(TypeInformation<R> type, BaseEntity entity) {
+		if (entity == null) {
+			return null;
+		}
+		Map<String, R> result = new HashMap<>();
+		EntityPropertyValueProvider propertyValueProvider = new EntityPropertyValueProvider(
+				entity, this.conversions);
+		Set<String> fieldNames = entity.getNames();
+		for (String field : fieldNames) {
+			result.put(field,
+					propertyValueProvider.getPropertyValue(field,
+							entity.getValue(field).getType() == ValueType.ENTITY, false,
+							null, type.isCollectionLike() ? type.getType() : null,
+							type.getComponentType().getType()));
+		}
+		return result;
 	}
 
 	@Override
--- a/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/DefaultDatastoreEntityConverter.java
+++ b/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/DefaultDatastoreEntityConverter.java
@@ -23,7 +23,6 @@ import java.util.Set;
 import com.google.cloud.datastore.BaseEntity;
 import com.google.cloud.datastore.Value;
 import com.google.cloud.datastore.ValueBuilder;
-import com.google.cloud.datastore.ValueType;
 
 import org.springframework.cloud.gcp.data.datastore.core.mapping.DatastoreDataException;
 import org.springframework.cloud.gcp.data.datastore.core.mapping.DatastoreMappingContext;
@@ -69,7 +68,8 @@ public class DefaultDatastoreEntityConverter implements DatastoreEntityConverter
 	}
 
 	@Override
-	public <R> Map<String, R> readAsMap(TypeInformation<R> type, BaseEntity entity) {
+	public <R> Map<String, R> readAsMap(TypeInformation<R> type, BaseEntity entity,
+			boolean embeddedMapValueIsEmbedded) {
 		if (entity == null) {
 			return null;
 		}
@@ -80,9 +80,10 @@ public class DefaultDatastoreEntityConverter implements DatastoreEntityConverter
 		for (String field : fieldNames) {
 			result.put(field,
 					propertyValueProvider.getPropertyValue(field,
-							entity.getValue(field).getType() == ValueType.ENTITY,
+							embeddedMapValueIsEmbedded,
 							type.isCollectionLike() ? type.getType() : null,
-							type.getComponentType().getType()));
+							type.isCollectionLike() ? type.getComponentType().getType()
+									: type.getType()));
 		}
 		return result;
 	}
--- a/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/DefaultDatastoreEntityConverter.java
+++ b/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/DefaultDatastoreEntityConverter.java
@@ -80,8 +80,8 @@ public class DefaultDatastoreEntityConverter implements DatastoreEntityConverter
 		for (String field : fieldNames) {
 			result.put(field,
 					propertyValueProvider.getPropertyValue(field,
-							entity.getValue(field).getType() == ValueType.ENTITY, false,
-							null, type.isCollectionLike() ? type.getType() : null,
+							entity.getValue(field).getType() == ValueType.ENTITY,
+							type.isCollectionLike() ? type.getType() : null,
 							type.getComponentType().getType()));
 		}
 		return result;
--- a/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/EntityPropertyValueProvider.java
+++ b/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/EntityPropertyValueProvider.java
@@ -16,12 +16,15 @@
 
 package org.springframework.cloud.gcp.data.datastore.core.convert;
 
+import java.util.Map;
+
 import com.google.cloud.datastore.BaseEntity;
 
 import org.springframework.cloud.gcp.data.datastore.core.mapping.DatastoreDataException;
 import org.springframework.cloud.gcp.data.datastore.core.mapping.DatastorePersistentProperty;
 import org.springframework.core.convert.ConversionException;
 import org.springframework.data.mapping.model.PropertyValueProvider;
+import org.springframework.data.util.TypeInformation;
 
 /**
  * A {@link PropertyValueProvider} for Datastore entities
@@ -46,27 +49,57 @@ public class EntityPropertyValueProvider implements PropertyValueProvider<Datast
 	}
 
 	@Override
-	@SuppressWarnings("unchecked")
 	public <T> T getPropertyValue(DatastorePersistentProperty persistentProperty) {
-		String fieldName = persistentProperty.getFieldName();
+		Class singularType = persistentProperty.getComponentType();
+		Class collectionType = persistentProperty.getType();
+		if (singularType == null) {
+			singularType = collectionType;
+			collectionType = null;
+		}
+		return getPropertyValue(persistentProperty.getFieldName(),
+				persistentProperty.isEmbedded(),
+				Map.class.isAssignableFrom(persistentProperty.getType()),
+				persistentProperty.getEmbeddedMapValueType(), collectionType,
+				singularType);
+	}
 
+	/**
+	 * Get a property value from the entity.
+	 * @param fieldName the name of the field to get.
+	 * @param isEmbedded if the property is an embedded entity.
+	 * @param isEmbeddedMap if the property should be retrieved as a map representing the
+	 * entity value
+	 * @param embeddedMapValueType if the property is an embedded entity and also to be
+	 * read as a {@code Map}, the value type of the map.
+	 * @param collectionType the collection type if the property is not singular. null if
+	 * the property is singular.
+	 * @param componentType the singular item type.
+	 * @return the property converted from the entity.
+	 */
+	@SuppressWarnings("unchecked")
+	public <T> T getPropertyValue(String fieldName, boolean isEmbedded,
+			boolean isEmbeddedMap, TypeInformation embeddedMapValueType,
+			Class collectionType, Class componentType) {
 		if (!this.entity.contains(fieldName)) {
 			return null;
 		}
 		try {
-			return persistentProperty.isEmbedded()
-					? this.conversion.convertOnReadEmbedded(
-					this.entity.getValue(fieldName).get(),
-							persistentProperty.getType(),
-							persistentProperty.getComponentType())
-					: this.conversion.convertOnRead(this.entity.getValue(fieldName).get(),
-							persistentProperty.getType(),
-							persistentProperty.getComponentType());
+			if (isEmbedded) {
+				return isEmbeddedMap
+						? this.conversion.convertOnReadEmbeddedMap(
+								this.entity.getEntity(fieldName), embeddedMapValueType)
+						: this.conversion.convertOnReadEmbedded(
+								this.entity.getValue(fieldName).get(), collectionType,
+								componentType);
+			}
+			else {
+				return this.conversion.convertOnRead(
+						this.entity.getValue(fieldName).get(), collectionType,
+						componentType);
+			}
 		}
 		catch (ConversionException | DatastoreDataException e) {
 			throw new DatastoreDataException("Unable to read property " + fieldName, e);
 		}
-
 	}
-
 }
--- a/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/EntityPropertyValueProvider.java
+++ b/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/EntityPropertyValueProvider.java
@@ -24,6 +24,7 @@ import org.springframework.cloud.gcp.data.datastore.core.mapping.DatastoreDataEx
 import org.springframework.cloud.gcp.data.datastore.core.mapping.DatastorePersistentProperty;
 import org.springframework.core.convert.ConversionException;
 import org.springframework.data.mapping.model.PropertyValueProvider;
+import org.springframework.data.util.ClassTypeInformation;
 import org.springframework.data.util.TypeInformation;
 
 /**
@@ -50,16 +51,22 @@ public class EntityPropertyValueProvider implements PropertyValueProvider<Datast
 
 	@Override
 	public <T> T getPropertyValue(DatastorePersistentProperty persistentProperty) {
-		Class singularType = persistentProperty.getComponentType();
+		boolean isMap = Map.class.isAssignableFrom(persistentProperty.getType());
+		TypeInformation singularType;
+		if (isMap) {
+			singularType = persistentProperty.getEmbeddedMapValueType();
+		}
+		else {
+			singularType = persistentProperty.getComponentType() == null ? null
+					: ClassTypeInformation.from(persistentProperty.getComponentType());
+		}
 		Class collectionType = persistentProperty.getType();
 		if (singularType == null) {
-			singularType = collectionType;
+			singularType = ClassTypeInformation.from(collectionType);
 			collectionType = null;
 		}
 		return getPropertyValue(persistentProperty.getFieldName(),
-				persistentProperty.isEmbedded(),
-				Map.class.isAssignableFrom(persistentProperty.getType()),
-				persistentProperty.getEmbeddedMapValueType(), collectionType,
+				persistentProperty.isEmbedded(), isMap, collectionType,
 				singularType);
 	}
 
@@ -67,35 +74,37 @@ public class EntityPropertyValueProvider implements PropertyValueProvider<Datast
 	 * Get a property value from the entity.
 	 * @param fieldName the name of the field to get.
 	 * @param isEmbedded if the property is an embedded entity.
-	 * @param isEmbeddedMap if the property should be retrieved as a map representing the
-	 * entity value
-	 * @param embeddedMapValueType if the property is an embedded entity and also to be
-	 * read as a {@code Map}, the value type of the map.
 	 * @param collectionType the collection type if the property is not singular. null if
 	 * the property is singular.
 	 * @param componentType the singular item type.
 	 * @return the property converted from the entity.
 	 */
-	@SuppressWarnings("unchecked")
 	public <T> T getPropertyValue(String fieldName, boolean isEmbedded,
-			boolean isEmbeddedMap, TypeInformation embeddedMapValueType,
 			Class collectionType, Class componentType) {
+		return (T) getPropertyValue(fieldName, isEmbedded, false, collectionType,
+				ClassTypeInformation.from(componentType));
+	}
+
+	@SuppressWarnings("unchecked")
+	private <T> T getPropertyValue(String fieldName, boolean isEmbedded, boolean isMap,
+			Class collectionType, TypeInformation componentType) {
 		if (!this.entity.contains(fieldName)) {
 			return null;
 		}
 		try {
 			if (isEmbedded) {
-				return isEmbeddedMap
+				return isMap
 						? this.conversion.convertOnReadEmbeddedMap(
-								this.entity.getEntity(fieldName), embeddedMapValueType)
+								this.entity.getEntity(fieldName), collectionType,
+								componentType)
 						: this.conversion.convertOnReadEmbedded(
 								this.entity.getValue(fieldName).get(), collectionType,
-								componentType);
+								componentType.getType());
 			}
 			else {
 				return this.conversion.convertOnRead(
 						this.entity.getValue(fieldName).get(), collectionType,
-						componentType);
+						componentType.getType());
 			}
 		}
 		catch (ConversionException | DatastoreDataException e) {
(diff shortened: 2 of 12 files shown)


ChengyuanZhao:
@meltsufin @dmitry-s PTAL , 

I added support for `Map<String, ?>` embedded properties where the value type can be simple types, collection types, even further embedded types, and collections of embedded types. This all comes pretty naturally from our existing conversion/mapping stuff.

spring-cloud-gcp-docs/src/main/asciidoc/datastore.adoc
```diff
@@ -242,18 +251,31 @@ import org.springframework.cloud.gcp.data.datastore.core.mapping.Embedded;
...
-	public EmbeddableTreeNode(List<EmbeddableTreeNode> children;) {
-		this.children = children;
-	}
+  @Embedded
+  @EmbeddedComponents
+  Map<String, EmbeddableTreeNode> siblingNodes;
+
+  @Embedded
+  @EmbeddedComponents
+  Map<String, Set<EmbeddableTreeNode>> subNodeGroups;
```
ChengyuanZhao: Don't know if possible, but check if `Map<String, @Embedded Set< ... > > ` is possible

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/DefaultDatastoreEntityConverter.java
```diff
@@ -62,6 +67,27 @@ public ReadWriteConversions getConversions() {
...
+			return null;
+		}
+		Map<String, R> result = new HashMap<>();
+		EntityPropertyValueProvider propertyValueProvider = new EntityPropertyValueProvider(
+				entity, this.conversions);
+		Set<String> fieldNames = entity.getNames();
+		for (String field : fieldNames) {
+			result.put(field,
+					propertyValueProvider.getPropertyValue(field,
+							embeddedMapValueIsEmbedded, false,
```
ChengyuanZhao: Right now this is always false to mean no arbitrary level of embedded value types. but we can put an integer arg in the embedded annotation as that setting and remove this limitation.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/EntityPropertyValueProvider.java
```diff
@@ -46,27 +50,71 @@ public EntityPropertyValueProvider(BaseEntity entity, ReadWriteConversions readW
...
+			Class collectionType, Class componentType) {
+		return (T) getPropertyValue(fieldName, isEmbedded, false, isEmbeddedComponents,
+				collectionType,
+				ClassTypeInformation.from(componentType));
+	}
+
+	@SuppressWarnings("unchecked")
+	private <T> T getPropertyValue(String fieldName, boolean isEmbedded, boolean isMap,
+			boolean isEmbeddedComponents,
+			Class collectionType, TypeInformation componentType) {
```
ChengyuanZhao: Combine two collection and component type into a single type information for the entire thing. removes one arg.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/ReadWriteConversions.java
```diff
@@ -51,10 +53,22 @@
 	 * {@code null} if the property is a singular object.
 	 * @param targetComponentType the type of the property to convert. For collection-like
 	 * properties this refers to the individual items' type.
+	 * @param isEmbeddedComponents {@code true} if the property is collection-like and its
+	 * items are individually stored as embedded entities.
 	 * @return an object of a target type.
 	 */
 	<T> T convertOnReadEmbedded(Object val, Class targetCollectionType,
-			Class targetComponentType);
+			Class targetComponentType, boolean isEmbeddedComponents);
```
ChengyuanZhao: "Embedded" might cause some to think it is the JPA/sql style embedded.  We should rename the annotation to "EmbeddedEntity" to make that clear.

spring-cloud-gcp-docs/src/main/asciidoc/datastore.adoc
```diff
@@ -242,18 +251,31 @@ import org.springframework.cloud.gcp.data.datastore.core.mapping.Embedded;
...
 
 public class EmbeddableTreeNode {
-	@Id
-	long value;
+  @Id
+  long value;
 
-	@Embedded
-	List<EmbeddableTreeNode> children;
+  @EmbeddedComponents
```
ChengyuanZhao: rename this to `ContainsEmbeddedEntities`

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/TwoStepsConversions.java
```diff
@@ -93,33 +95,51 @@ public TwoStepsConversions(CustomConversions customConversions, ObjectToKeyFacto
...
+
+	@Override
+	public <T> T convertOnReadEmbeddedMap(BaseEntity val, Class targetCollectionType,
+			TypeInformation targetComponentType, boolean embeddedMapValueIsEmbedded) {
+		return convertOnRead(val, true, true, embeddedMapValueIsEmbedded, null,
+				targetComponentType);
 	}
 
 	@SuppressWarnings("unchecked")
-	private <T> T convertOnRead(Object val, boolean isEmbedded,
```
ChengyuanZhao: discuss wtih Spring data team the annotation names we chose

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/TwoStepsConversions.java
```diff
@@ -93,33 +95,51 @@ public TwoStepsConversions(CustomConversions customConversions, ObjectToKeyFacto
...
 		this.internalConversionService.addConverter(BYTE_ARRAY_TO_BLOB_CONVERTER);
 		this.internalConversionService.addConverter(BLOB_TO_BYTE_ARRAY_CONVERTER);
-
 	}
 
 	@Override
 	public <T> T convertOnRead(Object val, Class targetCollectionType,
 			Class targetComponentType) {
-		return convertOnRead(val, false, targetCollectionType, targetComponentType);
+		return (T) convertOnRead(val, false, false, false, targetCollectionType,
```
ChengyuanZhao: There are many arguments, we should encapsulate them somehow.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/TwoStepsConversions.java
```diff
@@ -93,33 +95,51 @@ public TwoStepsConversions(CustomConversions customConversions, ObjectToKeyFacto
...
+	public <T> T convertOnReadEmbeddedMap(BaseEntity val, Class targetCollectionType,
+			TypeInformation targetComponentType, boolean embeddedMapValueIsEmbedded) {
+		return convertOnRead(val, true, true, embeddedMapValueIsEmbedded, null,
+				targetComponentType);
 	}
 
 	@SuppressWarnings("unchecked")
-	private <T> T convertOnRead(Object val, boolean isEmbedded,
-			Class targetCollectionType, Class targetComponentType) {
+	private <T> T convertOnRead(Object val, boolean isEmbedded, boolean isEmbeddedMap,
```
ChengyuanZhao: Instead of the 3 booleans let's make an enum. 

also write a table to spell out the combinations in the ref doc.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/TwoStepsConversions.java
```diff
@@ -93,33 +95,51 @@ public TwoStepsConversions(CustomConversions customConversions, ObjectToKeyFacto
...
+		else if (isEmbedded ^ isEmbeddedComponents) {
+			readConverter = this::convertOnReadSingleEmbedded;
+		}
+		else {
+			readConverter = this::convertOnReadSingle;
+		}
 
 		if ((val instanceof Iterable || val.getClass().isArray())
-				&& targetComponentType != null) {
+				&& targetComponentType != null && targetCollectionType != null) {
```
ChengyuanZhao: Recall the specific case where this new condition is needed

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/TwoStepsConversions.java
```diff
@@ -205,11 +249,23 @@ public Value convertOnWrite(Object proppertyVal, DatastorePersistentProperty per
...
 	}
 
-	private EntityValue convertOnWriteSingleEmbedded(Object val, String kindName) {
+	private EntityValue convertOnWriteSingleEmbedded(Object val, String kindName,
+			boolean isEmbeddedComponents) {
 		IncompleteKey key = this.objectToKeyFactory.getIncompleteKey(kindName);
 
 		FullEntity.Builder<IncompleteKey> builder = FullEntity.newBuilder(key);
-		this.datastoreEntityConverter.write(val, builder);
+		// Write an embedded Map
```
ChengyuanZhao: move this out to its own function so we have symmetry of 3 and 3 converter functions for read and write

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/TwoStepsConversions.java
```diff
@@ -205,11 +249,23 @@ public Value convertOnWrite(Object proppertyVal, DatastorePersistentProperty per
...
+		if (val instanceof Map) {
+			Map<String, ?> map = (Map<String, ?>) val;
+			for (String field : map.keySet()) {
+				builder.set(field, convertOnWrite(map.get(field),
+						isEmbeddedComponents, false, field));
```
ChengyuanZhao: Move the entity annotation to the class level of the embedded thing so we resolve this depth problem. 

just check for the existing entity annotation on those sub things

spring-cloud-gcp-docs/src/main/asciidoc/datastore.adoc
ChengyuanZhao: no longer necessary since i am now using the entity annotation on the embedded object's class.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/DefaultDatastoreEntityConverter.java
ChengyuanZhao: no longer necessary since i am now using the entity annotation on the embedded object's class.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/ReadWriteConversions.java
ChengyuanZhao: no longer necessary since i am now using the entity annotation on the embedded object's class.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/TwoStepsConversions.java
ChengyuanZhao: no longer necessary since i am now using the entity annotation on the embedded object's class.

spring-cloud-gcp-docs/src/main/asciidoc/datastore.adoc
ChengyuanZhao: no longer necessary since i am now using the entity annotation on the embedded object's class.

ChengyuanZhao:
@meltsufin @dmitry-s PTAL. I made the changes we discussed today:

- using the entity annotation on the embedded pojo's class instead to detect embedded
- using an enum to indicate the various types of embedded situations instead of boolean

spring-cloud-gcp-docs/src/main/asciidoc/datastore.adoc
```diff
@@ -195,65 +195,86 @@ The `@Field` annotation optionally provides a different field name than that of
...
-	@Embedded
-	EmbeddableTreeNode right;
+  EmbeddableTreeNode right;
 
-	public EmbeddableTreeNode(long value, EmbeddableTreeNode left, EmbeddableTreeNode right) {
-		this.value = value;
-		this.left = left;
-		this.right = right;
-	}
+  @EmbeddedMap
```
ChengyuanZhao: change it to `@AsEntity`

ChengyuanZhao: actually remove this annotation and assume all maps are embedded entities.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/TwoStepsConversions.java
```diff
@@ -93,33 +98,61 @@ public TwoStepsConversions(CustomConversions customConversions, ObjectToKeyFacto
...
-		BiFunction<Object, Class<?>, T> readConverter = isEmbedded
-				? this::convertOnReadSingleEmbedded
-				: this::convertOnReadSingle;
+		BiFunction<Object, TypeInformation<?>, ?> readConverter;
+		switch (embeddedStatus.getEmbeddedType()) {
+		case EMBEDDED_MAP:
+			readConverter = (x, y) -> convertOnReadSingleEmbeddedMap(x, y,
+					embeddedStatus.getEmbeddedMapDepthAllowance());
+			break;
+		case SINGULAR_EMBEDDED:
```
ChengyuanZhao: the singular and collection embedded actually don't have different use cases so they can just be one enum value.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/TwoStepsConversions.java
```diff
@@ -93,33 +98,61 @@ public TwoStepsConversions(CustomConversions customConversions, ObjectToKeyFacto
...
+			readConverter = this::convertOnReadSingleEmbedded;
+			break;
+		case COLLECTION_EMBEDDED_ELEMENTS:
+			readConverter = this::convertOnReadSingleEmbedded;
+			break;
+		default:
```
ChengyuanZhao: add case for NOT_EMBEDDED and throw exception in default.

ChengyuanZhao:
@meltsufin @dmitry-s PTAL. changes made

spring-cloud-gcp-docs/src/main/asciidoc/datastore.adoc
```diff
@@ -195,65 +195,76 @@ The `@Field` annotation optionally provides a different field name than that of
...
+
+  public EmbeddableTreeNode(long value, EmbeddableTreeNode left, EmbeddableTreeNode right) {
+    this.value = value;
+    this.left = left;
+    this.right = right;
+  }
 }
 ----
 
+Maps will be stored as embedded as entities where the key values become the field names in the embedded entity.
```
```suggestion
Maps will be stored as embedded entities where the key values become the field names in the embedded entity.
```

spring-cloud-gcp-docs/src/main/asciidoc/datastore.adoc
```diff
@@ -195,65 +195,76 @@ The `@Field` annotation optionally provides a different field name than that of
...
+  Map<String, Set<EmbeddableTreeNode>> subNodeGroups;
+
+  public EmbeddableTreeNode(List<EmbeddableTreeNode> children;) {
+    this.children = children;
+  }
 }
 ----
 
+Embedded objects can be stored in the values of embedded Maps.
+Collections of embedded objects can also be stored as the values of the embedded Maps.
```
```suggestion
Collections of embedded objects can also be stored as the values of embedded Maps.
```

ChengyuanZhao: @meltsufin changes applied! this is a pretty cool new feature. could be really convenient for fixes like this in the future.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/TwoStepsConversions.java
```diff
@@ -135,43 +163,74 @@ public TwoStepsConversions(CustomConversions customConversions, ObjectToKeyFacto
...
+			return this.datastoreEntityConverter.readAsMap(keyType, targetComponentType,
+					(BaseEntity) value);
+		}
+		throw new DatastoreDataException(
+				"Embedded entity was expected, but " + value.getClass() + " found");
 	}
 
 	@SuppressWarnings("unchecked")
-	private <T> T convertOnReadSingleEmbedded(Object value, Class<?> targetClass) {
+	private <T> T convertOnReadSingleEmbedded(Object value,
```
dmitry-s: no need to split into 2 lines

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/TwoStepsConversions.java
```diff
@@ -135,43 +163,74 @@ public TwoStepsConversions(CustomConversions customConversions, ObjectToKeyFacto
...
+					(BaseEntity) value);
 		}
-		throw new DatastoreDataException("Embedded entity was expected, but " + value.getClass() + " found");
+		throw new DatastoreDataException("Embedded entity was expected, but "
+				+ value.getClass() + " found");
 	}
 
 	@SuppressWarnings("unchecked")
-	private <T> T convertOnReadSingle(Object val, Class<?> targetType) {
+	private <T> T convertOnReadSingle(Object val,
```
dmitry-s: no need to split into 2 lines

ChengyuanZhao:
@meltsufin @dmitry-s any further changes?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/TwoStepsConversions.java
```diff
@@ -135,43 +163,74 @@ public TwoStepsConversions(CustomConversions customConversions, ObjectToKeyFacto
...
+			TypeInformation<?> targetTypeInformation) {
+		Class targetType = targetTypeInformation.getType();
+		Class sourceType = val.getClass();
 		Assert.notNull(val, "Cannot convert a null value.");
 		Object result = null;
 		TypeTargets typeTargets = computeTypeTargets(targetType);
 
-		if (typeTargets.getFirstStepTarget() == null && typeTargets.getSecondStepTarget() == null
+		if (typeTargets.getFirstStepTarget() == null
```
dmitry-s: is reformatting necessary?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/TwoStepsConversions.java
```diff
@@ -135,43 +163,74 @@ public TwoStepsConversions(CustomConversions customConversions, ObjectToKeyFacto
...
+		else if (typeTargets.getFirstStepTarget() != null &&
+				typeTargets.getSecondStepTarget() != null) {
 			//both steps were applied
-			Object secondStepVal = this.internalConversionService.convert(val, typeTargets.getFirstStepTarget());
+			Object secondStepVal = this.internalConversionService.convert(val,
+					typeTargets.getFirstStepTarget());
 			result = this.conversionService.convert(secondStepVal, targetType);
 		}
+		// if the value can be directly converted
+		else if (DatastoreNativeTypes.isNativeType(sourceType)
```
dmitry-s: no need for that, computeTypeTargets and the previous conditions take care of that

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/TwoStepsConversions.java
```diff
@@ -205,14 +286,35 @@ public Value convertOnWrite(Object proppertyVal, DatastorePersistentProperty per
 		return writeConverter.apply(val);
 	}
 
-	private EntityValue convertOnWriteSingleEmbedded(Object val, String kindName) {
+	private EntityValue applyEntityValueBuilder(String kindName,
```
dmitry-s: Refactor this method to accept kind name and return builder. Refactor methods that use this one appropriately.

spring-cloud-gcp-data-datastore/src/test/java/org/springframework/cloud/gcp/data/datastore/core/convert/DefaultDatastoreEntityConverterTests.java
```diff
@@ -508,13 +522,39 @@ public void testUnindexedField() {
 
 	@Test
 	public void testEmbeddedEntity() {
-		List<TestItemWithEmbeddedEntity.EmbeddedEtity> embeddedEtities = Arrays.asList(
-				new TestItemWithEmbeddedEntity.EmbeddedEtity("item 0"),
-				new TestItemWithEmbeddedEntity.EmbeddedEtity("item 1"));
+		EmbeddedEntity embeddedEntityA = new EmbeddedEntity("item 0");
```
dmitry-s: do we test that the added exceptions appear when they expected and that they have the right messages?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/TwoStepsConversions.java
```diff
@@ -93,33 +98,56 @@ public TwoStepsConversions(CustomConversions customConversions, ObjectToKeyFacto
...
 		if (val == null) {
 			return null;
 		}
-		BiFunction<Object, Class<?>, T> readConverter = isEmbedded
-				? this::convertOnReadSingleEmbedded
-				: this::convertOnReadSingle;
+		BiFunction<Object, TypeInformation<?>, ?> readConverter;
+		switch (embeddedType) {
+		case EMBEDDED_MAP:
+			readConverter = (x, y) -> convertOnReadSingleEmbeddedMap(x,
```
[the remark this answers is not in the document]
ChengyuanZhao: There is no meaning to having names here. In fact the other methods in this switch-case don't even have named params at all. Some real functional languages actually just allow you to use special no-name-necessary param names for these params in situations just like this.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/TwoStepsConversions.java
ChengyuanZhao: For me it looks like it's just 1 character too long if I don't

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/TwoStepsConversions.java
ChengyuanZhao: For me it looks like it's just 1 character too long if I don't

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/TwoStepsConversions.java
ChengyuanZhao: It does not. 

For example it will not convert Boolean to String because the source type is already a datastore native type. This came up in my unit test for an embedded map with a non-string key type.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/TwoStepsConversions.java
ChengyuanZhao: No, this would cause functions that use this function to repeat `EntityValue.of(builder.build())`

dmitry-s: Exactly, that is much more readable!

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/TwoStepsConversions.java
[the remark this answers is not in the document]
dmitry-s: The other methods have parameter names in the definition. I don't argue that sometimes it makes sense to skip the names, but in this particular case it would be more readable.

spring-cloud-gcp-data-datastore/src/test/java/org/springframework/cloud/gcp/data/datastore/core/convert/DefaultDatastoreEntityConverterTests.java
ChengyuanZhao: I have read that it is not a good idea to encode the text of exception messages into tests because of the burden of maintaining code like that. 

There aren't any new exceptions being thrown because this is using nearly all the existing mapping/converting code. The exception message about unsupported embedded type is not testable, and is there to catch the case where we add a new type in that enum but forget to account for it elsewhere.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/TwoStepsConversions.java
ChengyuanZhao: I respectfully disagree :P

ChengyuanZhao:
@meltsufin @dmitry-s PTAL. are there any more changes? We've discussed this topic multiple times now in person and I'd like to wrap this one up.

spring-cloud-gcp-data-datastore/src/test/java/org/springframework/cloud/gcp/data/datastore/core/convert/DefaultDatastoreEntityConverterTests.java
meltsufin: You can use the JUnit 4 rule to easily test expected exception message: https://junit.org/junit4/javadoc/4.12/org/junit/rules/ExpectedException.html
Since we don't have a rich exception hierarchy, I think verifying the messages is useful.
I would be interested in reading a description of a best practice to the contrary.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/mapping/EmbeddedType.java
```diff
@@ -0,0 +1,65 @@
...
+	 * embedded entity or arrays of embedded entities in the field.
+	 */
+	EMBEDDED_ENTITY,
+
+	/**
+	 * These are {@code Map}s that are stored as a single embedded entity in the field.
+	 */
+	EMBEDDED_MAP;
+
+	public static EmbeddedType of(TypeInformation typeInformation) {
```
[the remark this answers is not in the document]
ChengyuanZhao: @meltsufin done. PTAL

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/TwoStepsConversions.java
ChengyuanZhao: @meltsufin what do you mean.I thought we discussed in person and all agreed to use maps as embedded entities always.

PR #1134, opened 2018-10-23
Spanner option to exclude interleaved properties from mapping

### LINKED ISSUE #1133: Spanner change includeColumns param to includeProperties to also apply to interleaved properties

ChengyuanZhao:
fixes #1133 

Allow users to set option to exclude mapping interleaved properties

### DIFF
--- a/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerMutationFactory.java
+++ b/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerMutationFactory.java
@@ -49,11 +49,11 @@ public interface SpannerMutationFactory {
 	 * If there are no interleaved children then the returned list will contain the single
 	 * mutation for the given object.
 	 * @param object The object to update or newly insert.
-	 * @param includeColumns The columns to use in the operation. if null then all columns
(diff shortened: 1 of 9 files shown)


ChengyuanZhao:
@meltsufin We don't have this this specifically listed in the refdoc, and that's intentional because it is in the SpannerReadOptions and SpannerQueryOptions objects with obvious getters and setters. The ref doc is getting pretty big and cluttered, and I didn't think this was a big structural feature (like relationships and transactions) to deserve a mention there. In any case, I think a user would quickly discover it if they needed it.

meltsufin:
@ChengyuanZhao What about my comment regarding exclude vs. include properties?

ChengyuanZhao:
@meltsufin I thought about it from the perspective of using one POJO for multiple different queries (as we have seen from some customers). In that scenario I think it's easier for a user to think in terms of what he or she  DOES need rather than what we DOESN'T need in each case.

meltsufin:
It seems like it would be difficult to use for someone who just wants to store all properties, which I imagine would be a common case.

ChengyuanZhao:
@meltsufin if the Set is null then of course we include everything.

meltsufin:
I see. That works I guess. It's a bit unusual that a null set is equivalent to the full set, rather than the empty set, but OK.

PR #1135, opened 2018-10-23
Move SQL sample tests to be under 1 database

### LINKED ISSUE #1125: Move cloud SQL sample tests to same Database
As a followup to #1123, it would be ideal to remove `application-test.properties` files and move all properties set by integration tests to travis.yml. This way we do not provide users with tests that do not work out of the box.

dzou:
This moves the SQL sample tests to a single database under the spring-cloud-gcp-ci project.

The purpose of doing this is to have the properties be set in travis.yml and have the tests inherit from the `application.properties` that users modify. This way allows the tests to work for users who want to try out the test.

Fixes #1125.

### DIFF
--- a/spring-cloud-gcp-samples/spring-cloud-gcp-sql-sample/src/test/java/com/example/SqlApplicationTests.java
+++ b/spring-cloud-gcp-samples/spring-cloud-gcp-sql-sample/src/test/java/com/example/SqlApplicationTests.java
@@ -28,11 +28,11 @@ import org.springframework.boot.test.context.SpringBootTest;
 import org.springframework.boot.test.context.SpringBootTest.WebEnvironment;
 import org.springframework.boot.test.web.client.TestRestTemplate;
 import org.springframework.boot.web.server.LocalServerPort;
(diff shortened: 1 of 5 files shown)


spring-cloud-gcp-samples/spring-cloud-gcp-sql-sample/src/test/java/com/example/SqlApplicationTests.java
```diff
@@ -28,28 +28,27 @@
...
- * Run with: mvn -Dit.cloudsql -Dspring.cloud.gcp.sql.database-name=[...]
- * -Dspring.cloud.gcp.sql.instance-connection-name=[...]
- * -Dspring.datasource.password=[...] test
+ * Run with: mvn -Dit.cloudsql test
  *
  * @author Daniel Zou
  */
 @RunWith(SpringRunner.class)
-@TestPropertySource("classpath:application-test.properties")
+@PropertySource("classpath:application.properties")
```
meltsufin: I think you can omit this line altogether.

Issue #1007, opened 2018-09-10
Datastore ancestor-based parent-child  POJO relationships


ChengyuanZhao:
This refers to the specific case of a POJO holding a reference to an object of a different POJO type. The child POJO also has a Kind. 

This does NOT refer to:
- child POJO stored as an embedded entity
- Datastore Key referring to the entity in the other Kind

ChengyuanZhao:
Update from discussion in-person:

- We can use a new annotation @Descendants to mark a collection in the parent POJO that refers to children.
the children are then either allocated new keys with the ancestor component from the parent on save.

- This contrasts with how Objectify does it. There the parent actually has a column in its kind that stores the collection of keys of the children. Instead the relationship in our case is contained in the key column of the child that it has anyway.

ChengyuanZhao:
Due to the size of the code changes I am expecting, I'll break this up into a few PRs with the steps listed below. (as we did for Spanner's Interleaved work)

- [ ] Read functionality and unit tests

- [ ] Write functionality and unit tests

- [ ] Integration tests. this comes after because this will be a large-ish PR, too. It also won't be possible until both read and write functionality is in

- [ ] ref doc updates.

PR #1136, opened 2018-10-23
Descendants write support

dmitry-s:
Part of #1007

### DIFF
--- a/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
+++ b/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
@@ -268,9 +268,10 @@ public class DatastoreTemplate implements DatastoreOperations {
 				validateKey(entity, keyToPathElement(ancestor));
 			}
 		}
-		Builder builder = Entity.newBuilder(getKey(entity, true, ancestors));
+		Key key = getKey(entity, true, ancestors);
+		Builder builder = Entity.newBuilder(key);
 		this.datastoreEntityConverter.write(entity, builder);
-		saveDescendents(entity, getKey(entity, true, ancestors));
+		saveDescendents(entity, key);
 		return builder.build();
 	}
 
@@ -282,21 +283,20 @@ public class DatastoreTemplate implements DatastoreOperations {
 					//Convert and write descendants, applying ancestor from parent entry
 					PersistentPropertyAccessor accessor = datastorePersistentEntity.getPropertyAccessor(entity);
 					Object val = accessor.getProperty(persistentProperty);
-
-					//Check if property is a non-null array
-					if (val != null && val.getClass().isArray() && val.getClass() != byte[].class) {
-						//if a propperty is an array, convert it to list
-						val = CollectionUtils.arrayToList(val);
-					}
-					if (val instanceof Iterable) {
-						PathElement ancestorPE = keyToPathElement(key);
-						((Iterable<?>) val).forEach(descendant -> validateKey(descendant, ancestorPE));
+					if (val != null) {
+						//we can be sure that the property is an array or an iterable,
+						//because we check it in isDescendant
+						if (val.getClass().isArray() && val.getClass() != byte[].class) {
+							//if a property is an array, convert it to list
+							val = CollectionUtils.arrayToList(val);
+						}
 						saveAll((Iterable<?>) val, key);
 					}
 				});
 	}
 
-	private PathElement keyToPathElement(Key key) {
+	public static PathElement keyToPathElement(Key key) {
+		Assert.notNull(key, "A non-null key is required");
 		return key.getName() != null
 				? PathElement.of(key.getKind(), key.getName())
 				: PathElement.of(key.getKind(), key.getId());
--- a/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
+++ b/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
@@ -26,12 +26,14 @@ import java.util.stream.Collectors;
 import java.util.stream.StreamSupport;
 
 import com.google.cloud.datastore.BaseEntity;
+import com.google.cloud.datastore.BaseKey;
 import com.google.cloud.datastore.Datastore;
 import com.google.cloud.datastore.DatastoreReaderWriter;
 import com.google.cloud.datastore.Entity;
 import com.google.cloud.datastore.Entity.Builder;
 import com.google.cloud.datastore.EntityQuery;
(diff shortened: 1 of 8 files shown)


spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/DatastoreServiceObjectToKeyFactory.java
```diff
@@ -91,17 +90,34 @@ public Key getKeyFromObject(Object entity,
...
 		Class idPropType = idProp.getType();
-		Object id = idPropType.equals(Long.class) ? allocatedKey.getId() : allocatedKey.getId().toString();
 
-		datastorePersistentEntity.getPropertyAccessor(entity).setProperty(idProp,
-				idPropType.equals(Key.class) ? allocatedKey : id);
+		Object value;
+		if (idPropType.equals(Key.class)) {
+			value = allocatedKey;
+		}
+		else if (idPropType.equals(Long.class)) {
```
ChengyuanZhao: Should we also disallow getting long and string id values if ancestors != null? we have this policy in the datastoretemplate to prevent loss of ancestor info, but I think that should apply here and throughout the project, too

spring-cloud-gcp-data-datastore/src/test/java/org/springframework/cloud/gcp/data/datastore/it/DatastoreIntegrationTests.java
```diff
@@ -196,8 +196,44 @@ public void embeddedCollectionTest() {
 		assertEquals(treeCollection, loaded);
 	}
 
-	private void waitUntilTrue(Supplier<Boolean> condition) throws InterruptedException {
-		Awaitility.await().atMost(QUERY_WAIT_INTERVAL_SECONDS, TimeUnit.SECONDS)
-				.untilAsserted(() -> assertTrue(condition.get()));
+	@Test
+	public void ancestorsTest() {
+		AncestorEntity.DescendatEntry descendatEntryA = new AncestorEntity.DescendatEntry("a");
```
meltsufin: decenda**n**t

spring-cloud-gcp-data-datastore/src/test/java/org/springframework/cloud/gcp/data/datastore/it/DatastoreIntegrationTests.java
```diff
@@ -196,8 +196,44 @@ public void embeddedCollectionTest() {
...
+		this.datastoreTemplate.save(ancestorEntity);
+
+		waitUntilTrue(() ->
+				this.datastoreTemplate.findAll(AncestorEntity.DescendatEntry.class)
+						.stream().allMatch(descendatEntry -> descendatEntry.name.contains("updated")));
+
+		AncestorEntity loadedEntityAfterUpdate =
+				this.datastoreTemplate.findById(ancestorEntity.id, AncestorEntity.class);
+
+		this.datastoreTemplate.deleteAll(AncestorEntity.class);
```
meltsufin: This cleanup might never run if the test fails in the middle.

spring-cloud-gcp-data-datastore/src/test/java/org/springframework/cloud/gcp/data/datastore/it/AncestorEntity.java
```diff
@@ -0,0 +1,91 @@
...
+	public boolean equals(Object o) {
+		if (this == o) {
+			return true;
+		}
+		if (o == null || getClass() != o.getClass()) {
+			return false;
+		}
+		AncestorEntity that = (AncestorEntity) o;
+		return Objects.equals(this.name, that.name) &&
+				new HashSet<>(this.descndants).equals(new HashSet<>(that.descndants));
```
meltsufin: descendants

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
```diff
@@ -256,12 +262,62 @@ public static void applyQueryOptions(StructuredQuery.Builder builder, DatastoreQ
...
+					Object val = accessor.getProperty(persistentProperty);
+
+					//Check if property is a non-null array
+					if (val != null && val.getClass().isArray() && val.getClass() != byte[].class) {
+						//if a propperty is an array, convert it to list
+						val = CollectionUtils.arrayToList(val);
+					}
+					if (val instanceof Iterable) {
+						PathElement ancestorPE = keyToPathElement(key);
+						((Iterable<?>) val).forEach(descendant -> validateKey(descendant, ancestorPE));
```
meltsufin: Don't need to call `validateKey` here since it will be called inside `save`.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/DatastoreServiceObjectToKeyFactory.java
```diff
@@ -91,17 +90,34 @@ public Key getKeyFromObject(Object entity,
...
+			value = allocatedKey.getId();
```
meltsufin: Check if updating without ancestor info causes a failure on Datastore.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
```diff
@@ -256,12 +262,62 @@ public static void applyQueryOptions(StructuredQuery.Builder builder, DatastoreQ
...
+					//Convert and write descendants, applying ancestor from parent entry
+					PersistentPropertyAccessor accessor = datastorePersistentEntity.getPropertyAccessor(entity);
+					Object val = accessor.getProperty(persistentProperty);
+
+					//Check if property is a non-null array
+					if (val != null && val.getClass().isArray() && val.getClass() != byte[].class) {
+						//if a propperty is an array, convert it to list
+						val = CollectionUtils.arrayToList(val);
+					}
+					if (val instanceof Iterable) {
```
meltsufin: Is it already verified that it's collection-like? If not, we should also have an `else` throwing an exception.

dmitry-s:
@ChengyuanZhao @meltsufin  PTAL

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/DatastoreServiceObjectToKeyFactory.java
ChengyuanZhao: @meltsufin @dmitry-s I just did this test manually. The parent is indeed part of the key, so you can have two entities with the same ID value but with different parents since those are 2 separate keys.

PR #1139, opened 2018-10-24
Fix top bugs found by SonarQube analysis

meltsufin:


spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryExecutor.java
```diff
@@ -101,10 +102,10 @@
 							.getPersistentProperty(o.getProperty());
 					return property == null ? o.getProperty() : property.getColumnName();
 				});
-		if (options.hasLimit()) {
+		if (options.getLimit() != null) {
 			sb.append(" LIMIT ").append(options.getLimit());
 		}
-		if (options.hasOffset()) {
+		if (options.getLimit() != null) {
```
ChengyuanZhao: getOffset isn't null?

meltsufin: Great catch!

```suggestion
		if (options.getOffset() != null) {
```

meltsufin:
@ChengyuanZhao @dmitry-s Can you do one more pass over this? Thanks!

PR #1140, opened 2018-10-24
Add sample to demo Pub/Sub JSON payloads

dzou:
This adds a sample and integration test to demo automatically serializing and deserializing JSON payloads in Pub/Sub.

Fixes #831.

PR #1141, opened 2018-10-25
Make PubSubSubscriberTemplate pickup message converter bean

dzou:
Makes the PubSubSubscriberTemplate in the autoconfiguration automatically pick up PubSubMessageConverter bean if it exists. This is equivalent to how PubSubPublisherTemplate does it.

This PR is motivated by comments in #1140:
https://github.com/spring-cloud/spring-cloud-gcp/pull/1140#discussion_r228202166

### DIFF
--- a/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
+++ b/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
@@ -67,6 +67,7 @@ import org.springframework.context.annotation.Configuration;
  * @author João André Martins
  * @author Mike Eltsufin
  * @author Chengyuan Zhao
+ * @author Daniel Zou
  */
 @Configuration
 @AutoConfigureAfter(GcpContextAutoConfiguration.class)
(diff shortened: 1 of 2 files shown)


spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
```diff
@@ -130,8 +130,11 @@ public PubSubPublisherTemplate pubSubPublisherTemplate(PublisherFactory publishe
 
 	@Bean
 	@ConditionalOnMissingBean
-	public PubSubSubscriberTemplate pubSubSubscriberTemplate(SubscriberFactory subscriberFactory) {
-		return new PubSubSubscriberTemplate(subscriberFactory);
+	public PubSubSubscriberTemplate pubSubSubscriberTemplate(SubscriberFactory subscriberFactory,
+			ObjectProvider<PubSubMessageConverter> pubSubMessageConverter) {
+		PubSubSubscriberTemplate pubSubSubscriberTemplate = new PubSubSubscriberTemplate(subscriberFactory);
+		pubSubMessageConverter.ifUnique(pubSubSubscriberTemplate::setMessageConverter);
```
artembilan: Any possible unit-test to cover this change?

And your name to the `@author` list 😄 

Thanks

dzou: Added author tag.

I tried writing a unit test but it seemed a little bit messy since I would have to create a bunch of mocks to do it properly. I thought this was a signal that the configuration would be better tested via Integration test. What are your thoughts on this?

Thanks

artembilan: I think you just need to add extra `@Configuration` into the `PubSubTemplateIntegrationTests` with the `PubSubMessageConverter` `@Bean` and verify in the existing (or new) test-case against that `this.contextRunner` to be sure that `PubSubTemplate` is supplied with your  `PubSubMessageConverter`.

PR #1142, opened 2018-10-25
Enable Codecov reporting from Travis builds

meltsufin:


meltsufin:
I think we need to add Jacoco as well.

artembilan:
How to see the report here in PR from that Codecov?

And looks like Snyk is a noise and not so useful here on PR.
Plus it comes from my account 😢 ...

meltsufin:
This PR is still WIP. (Partially) agreed about Snyk. We need to setup a bot account rather than yours. Let's discuss that separately.

meltsufin:
@artembilan Ready for review now. I don't think the Codecov reports will kick in until we merge into master.

PR #1145, opened 2018-10-26
Datastore descendants ref doc

### LINKED ISSUE #1007: Datastore ancestor-based parent-child  POJO relationships

ChengyuanZhao:
fixes #1007 

ref doc for ancestors and descendants

### DIFF
--- a/spring-cloud-gcp-docs/src/main/asciidoc/datastore.adoc
+++ b/spring-cloud-gcp-docs/src/main/asciidoc/datastore.adoc
@@ -267,7 +267,46 @@ Collections of embedded objects can also be stored as the values of embedded Map
 
 ==== Relationships
 
-This feature is not supported yet.
+Parent-child relationships are supported via the `@Descendants` annotation.
+
+[source, java]
+----
+import org.springframework.cloud.gcp.data.datastore.core.mapping.Descendants;
+import org.springframework.data.annotation.Id;
+
+@Entity
+public class ShoppingOrder {
+  @Id
+  long id;
+
+  @Descendants
+  List<Item> children;
+}
+
+@Entity
+public class Item {
+  @Id
+  Key purchasedItemKey;
+
+  String name;
+
+  Timestamp originalAcquisitionTime;
+}
+----
+
+The parent-child relationship structure of objects is stored in Cloud Datastore using Datastore's https://cloud.google.com/datastore/docs/concepts/entities#ancestor_paths[ancestor relationships].
+These relationships can be many levels deep.
+
+Properties holding child entities must be collection-like, but they can be any of the supported types inter-convertible collection-like types that are supported for regular properties such as `List`, arrays, `Set`, etc...
+Child items must have `Key` as their ID type because Cloud Datastore stores the ancestor relationship link inside the keys of the children.
+
+Reading or saving an entity automatically causes all subsequent levels of children under that entity to be read or saved, respectively.
+If a new child is created and added to a property annotated `@Descendants` and the key property is left null, then a new key will be allocated for that child.
+
+Children entities cannot be moved from the property of one parent to that of another unless the child's key property is set to null or a value that contains the new parent as an ancestor.
+Since Cloud Datastore entity keys can have multiple parents it is possible that a child entity appears in the property of multiple parent entities.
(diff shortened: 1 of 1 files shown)


ChengyuanZhao:
@meltsufin @dmitry-s we're finally there! final little bit of the ancestors/descendants work. PTAL.

spring-cloud-gcp-docs/src/main/asciidoc/datastore.adoc
```diff
@@ -267,7 +267,47 @@ Collections of embedded objects can also be stored as the values of embedded Map
...
+These relationships can be many levels deep.
+
+Properties holding child entities must be collection-like, but they can be any of the supported inter-convertible collection-like types that are supported for regular properties such as `List`, arrays, `Set`, etc...
+Child items must have `Key` as their ID type because Cloud Datastore stores the ancestor relationship link inside the keys of the children.
+
+Reading or saving an entity automatically causes all subsequent levels of children under that entity to be read or saved, respectively.
+If a new child is created and added to a property annotated `@Descendants` and the key property is left null, then a new key will be allocated for that child.
+The ordering of the retrieved children may not be the same as the ordering in the original property that was saved.
+
+Children entities cannot be moved from the property of one parent to that of another unless the child's key property is set to null or a value that contains the new parent as an ancestor.
```
```suggestion
Child entities cannot be moved from the property of one parent to that of another unless the child's key property is set to `null` or a value that contains the new parent as an ancestor.
```

spring-cloud-gcp-docs/src/main/asciidoc/datastore.adoc
```diff
@@ -267,7 +267,47 @@ Collections of embedded objects can also be stored as the values of embedded Map
...
+Since Cloud Datastore entity keys can have multiple parents it is possible that a child entity appears in the property of multiple parent entities.
```
```suggestion
Since Cloud Datastore entity keys can have multiple parents, it is possible that a child entity appears in the property of multiple parent entities.
```

spring-cloud-gcp-docs/src/main/asciidoc/datastore.adoc
```diff
@@ -267,7 +267,47 @@ Collections of embedded objects can also be stored as the values of embedded Map
...
+import org.springframework.cloud.gcp.data.datastore.core.mapping.Descendants;
+import org.springframework.data.annotation.Id;
+
+@Entity
+public class ShoppingOrder {
+  @Id
+  long id;
+
+  @Descendants
+  List<Item> children;
```
```suggestion
  List<Item> items;
```

spring-cloud-gcp-docs/src/main/asciidoc/datastore.adoc
```diff
@@ -267,7 +267,47 @@ Collections of embedded objects can also be stored as the values of embedded Map
...
+}
+
+@Entity
+public class Item {
+  @Id
+  Key purchasedItemKey;
+
+  String name;
+
+  Timestamp originalAcquisitionTime;
```
meltsufin: I was expecting something like price or count. What's that?
```suggestion
  BigDecimal price;
```

ChengyuanZhao: @meltsufin  We cannot put "price" here because it is contrary to the example. 

I specifically avoided doing that. 

Price is something "static" for all items of that price, and as you can see our item is specific to a specific shopping cart. "originalAcquisitionTime" is the time this specific physical item was acquired from the factory or supplier, and that is not static to every item of the same model or type.

meltsufin: Maybe I don't understand retail enough, but I never considered that an item in the shopping cart has some acquisition time attirbute. Maybe time of placement into the cart? Maybe a realistic example would be to have a reference to a product? I don't know, I'm confused by this example. Does it make sense to everyone else?

ChengyuanZhao: This really doesn't have anything to do with retail. It just needs to be an attribute specific to an instance of an item, and not just to the model of item. 

If it helps you understand, I can change this the time it was added to cart instead of time the item entered the warehouse.

Issue #1115, opened 2018-10-17
Datastore @Transactional integration test and sample


ChengyuanZhao:
This should be done after ancestors work is completed because those features probably need to be demonstrated.

ChengyuanZhao:
- [ ] Transactional integration tests 

- [ ] sample updated to use transactions

PR #1147, opened 2018-10-26
Datastore transactional integ tests

ChengyuanZhao:
related to #1115 
Transactional annotation integration test

### DIFF
--- a/spring-cloud-gcp-data-datastore/src/test/java/org/springframework/cloud/gcp/data/datastore/it/DatastoreIntegrationTestConfiguration.java
+++ b/spring-cloud-gcp-data-datastore/src/test/java/org/springframework/cloud/gcp/data/datastore/it/DatastoreIntegrationTestConfiguration.java
@@ -28,6 +28,7 @@ import org.springframework.cloud.gcp.core.DefaultCredentialsProvider;
 import org.springframework.cloud.gcp.core.DefaultGcpProjectIdProvider;
 import org.springframework.cloud.gcp.core.UsageTrackingHeaderProvider;
 import org.springframework.cloud.gcp.data.datastore.core.DatastoreTemplate;
+import org.springframework.cloud.gcp.data.datastore.core.DatastoreTransactionManager;
 import org.springframework.cloud.gcp.data.datastore.core.convert.DatastoreEntityConverter;
 import org.springframework.cloud.gcp.data.datastore.core.convert.DatastoreServiceObjectToKeyFactory;
 import org.springframework.cloud.gcp.data.datastore.core.convert.DefaultDatastoreEntityConverter;
@@ -37,6 +38,7 @@ import org.springframework.cloud.gcp.data.datastore.repository.config.EnableData
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.context.annotation.PropertySource;
+import org.springframework.transaction.annotation.EnableTransactionManagement;
 
 /**
  * @author Chengyuan Zhao
@@ -44,6 +46,7 @@ import org.springframework.context.annotation.PropertySource;
 @Configuration
 @PropertySource("application-test.properties")
 @EnableDatastoreRepositories
+@EnableTransactionManagement
 public class DatastoreIntegrationTestConfiguration {
 
 	private final String projectId = new DefaultGcpProjectIdProvider().getProjectId();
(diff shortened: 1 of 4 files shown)


ChengyuanZhao:
@meltsufin @dmitry-s PTAL.

spring-cloud-gcp-data-datastore/src/test/java/org/springframework/cloud/gcp/data/datastore/it/TransactionalTemplateService.java
```diff
@@ -0,0 +1,64 @@
...
+		for (TestEntity testEntity : testEntities) {
+			assertNull(this.datastoreTemplate.findById(testEntity.getId(),
+					TestEntity.class));
+		}
+
+		this.datastoreTemplate.saveAll(testEntities);
+
+		// Because these saved entities should NOT appear when we subsequently check, we
+		// must wait a period of time that would see a non-transactional save go through.
+		Thread.sleep(DatastoreIntegrationTests.QUERY_WAIT_INTERVAL_SECONDS * 1000);
```
meltsufin: That's a long period of waiting without trying. Maybe use awaitility like we do in other tests?

ChengyuanZhao: I could be wrong, but the reason I didn't use await utility is that there is nothing to wait for. We're not supposed to see a condition be fulfilled, so we would end up waiting the entire period.

meltsufin: I see. So we are waiting for something not to happen. How do you choose how long to wait for? Why not longer than 15 seconds and why not less? If we can't come up with a reason for the magic number, I would say it's essentially untestable.

ChengyuanZhao: @meltsufin @dmitry-s PTAL. I am now waiting a period based on how long the other awaits took.

ChengyuanZhao: @meltsufin PTAL, requested change made.

spring-cloud-gcp-data-datastore/src/test/java/org/springframework/cloud/gcp/data/datastore/it/DatastoreIntegrationTests.java
```diff
@@ -75,13 +83,13 @@ public static void checkToRun() {
...
 		this.datastoreTemplate.deleteAll(AncestorEntity.class);
-		this.datastoreTemplate.deleteAll(AncestorEntity.DescendatEntry.class);
+		this.datastoreTemplate.deleteAll(AncestorEntity.DescendantEntry.class);
 		this.datastoreTemplate.deleteAll(TreeCollection.class);
 		this.testEntityRepository.deleteAll();
 	}
 
 	@Test
-	public void testSaveAndDeleteRepository() {
+	public void testSaveAndDeleteRepository() throws InterruptedException {
```
meltsufin: This test is really a bunch of different, but related tests. Stylistically, my preference would be to break it up into multiple tests methods.

spring-cloud-gcp-data-datastore/src/test/java/org/springframework/cloud/gcp/data/datastore/it/DatastoreIntegrationTestConfiguration.java
```diff
@@ -57,6 +60,18 @@
 	public DatastoreIntegrationTestConfiguration() throws IOException {
 	}
 
+	@Bean
+	@ConditionalOnMissingBean
```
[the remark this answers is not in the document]
ChengyuanZhao: I suppose it would make as much sense without these. I will remove them all.

spring-cloud-gcp-data-datastore/src/test/java/org/springframework/cloud/gcp/data/datastore/it/DatastoreIntegrationTests.java
meltsufin: Please make an issue for it. Thanks!

ChengyuanZhao: https://github.com/spring-cloud/spring-cloud-gcp/issues/1154

PR #1150, opened 2018-10-29
Datastore ancestors key check fix

ChengyuanZhao:
Fixes bug where we try to make ancestors even if none are given.


ChengyuanZhao:
@dmitry-s PTAL

PR #1151, opened 2018-10-29
Spanner admin template doesn't check database exists as often

### LINKED ISSUE #1149: Cloud Spanner integration tests are broken due to quota exceeded
In Travis log:

```
[ERROR] Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 0.127 s <<< FAILURE! - in org.springframework.cloud.gcp.data.spanner.repository.it.SpannerRepositoryIntegrationTests
[ERROR] queryMethodsTest(org.springframework.cloud.gcp.data.spanner.repository.it.SpannerRepositoryIntegrationTests)  Time elapsed: 0.114 s  <<< ERROR!
com.google.cloud.spanner.SpannerException: 
RESOURCE_EXHAUSTED: io.grpc.StatusRuntimeException: RESOURCE_EXHAUSTED: Quota exceeded for quota metric 'spanner.googleapis.com/admin_method_quota' and limit 'AdminMethodQuotaPerMinutePerProject' of service 'spanner.googleapis.com' for consumer 'project_number:653574825147'.
links {
  description: "Google developer console API key"
  url: "https://console.developers.google.com/project/653574825147/apiui/credential"
}
Caused by: java.util.concurrent.ExecutionException: 
io.grpc.StatusRuntimeException: RESOURCE_EXHAUSTED: Quota exceeded for quota metric 'spanner.googleapis.com/admin_method_quota' and limit 'AdminMethodQuotaPerMinutePerProject' of service 'spanner.googleapis.com' for consumer 'project_number:653574825147'.
links {
  description: "Google developer console API key"
  url: "https://console.developers.google.com/project/653574825147/apiui/credential"
}
Caused by: io.grpc.StatusRuntimeException: 
RESOURCE_EXHAUSTED: Quota exceeded for quota metric 'spanner.googleapis.com/admin_method_quota' and limit 'AdminMethodQuotaPerMinutePerProject' of service 'spanner.googleapis.com' for consumer 'project_number:653574825147'.
links {
  description: "Google developer console API key"
  url: "https://console.developers.google.com/project/653574825147/apiui/credential"
}

```

ChengyuanZhao:
fixes #1149

Doesn't check for database existence as often to reduce calls to that API.

This reduces several usages of databaseExists and tableExists, which both need to get the list of databases.

### DIFF
--- a/spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/test/AbstractSpannerIntegrationTest.java
+++ b/spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/test/AbstractSpannerIntegrationTest.java
@@ -20,8 +20,7 @@ import java.util.List;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
-import org.junit.After;
-import org.junit.Before;
+import org.junit.AfterClass;
 import org.junit.BeforeClass;
 import org.junit.Test;
 
@@ -95,15 +94,12 @@ public abstract class AbstractSpannerIntegrationTest {
 	private boolean setupFailed;
 
 	@BeforeClass
-	public static void checkToRun() {
+	public void setup() {
 		assumeThat(
 				"Spanner integration tests are disabled. Please use '-Dit.spanner=true' "
 						+ "to enable them. ",
 				System.getProperty("it.spanner"), is("true"));
-	}
 
-	@Before
-	public void setup() {
 		try {
 			createDatabaseWithSchema();
 		}
@@ -150,7 +146,7 @@ public abstract class AbstractSpannerIntegrationTest {
 				.getDropTableDdlStringsForInterleavedHierarchy(Trade.class);
 	}
 
-	@After
+	@AfterClass
 	public void clean() {
 		try {
 			// this is to reduce duplicated errors reported by surefire plugin
--- a/spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/test/AbstractSpannerIntegrationTest.java
+++ b/spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/test/AbstractSpannerIntegrationTest.java
@@ -96,7 +96,7 @@ public abstract class AbstractSpannerIntegrationTest {
 
 	private static boolean tablesInitialized;
 
-	private static boolean tablesCleanedUp;
+	private static int initializeAttempts;
 
 	@BeforeClass
 	public static void checkToRun() {
@@ -109,6 +109,7 @@ public abstract class AbstractSpannerIntegrationTest {
 	@Before
 	public void setup() {
 		try {
+			initializeAttempts++;
 			if (tablesInitialized) {
 				return;
 			}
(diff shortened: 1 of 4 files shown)


spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/test/AbstractSpannerIntegrationTest.java
```diff
@@ -122,19 +121,20 @@ protected void createDatabaseWithSchema() {
 		String tableName = this.spannerMappingContext.getPersistentEntity(Trade.class)
 				.tableName();
 
+		List<String> createStatements = createSchemaStatements();
+
 		if (!this.spannerDatabaseAdminTemplate.databaseExists()) {
 			assertFalse(this.spannerDatabaseAdminTemplate.tableExists(tableName));
```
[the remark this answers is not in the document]
ChengyuanZhao: This is actually a helper method  (see the protected access), and it is called within methods annotated Test, so I think it qualifies as a test method.

ChengyuanZhao:
@meltsufin the difference is in the executeDdlStrings function. It no longer checks if the database exists. This is the main difference. 

Also I am not understanding what you mean by the before and after annotations.. Our integration tests for repo and template are each a single test method each, so this checking is really just done once anyway.

spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/test/AbstractSpannerIntegrationTest.java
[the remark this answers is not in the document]
meltsufin: Can we have a separate `testTableExists` test so that when it fails, we would know exactly what failed?

meltsufin:
If you only have one test per test class, then there is no difference indeed. However, once you add more than one test, that will no longer be the case and you will be duplicating database creation.

spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/test/AbstractSpannerIntegrationTest.java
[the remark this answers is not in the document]
ChengyuanZhao: Yes , moved to separate test.

spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/test/AbstractSpannerIntegrationTest.java
```diff
@@ -113,28 +113,31 @@ public void setup() {
 		}
 	}
 
+	@Test
```
meltsufin: Now it's really vital to make `setup()` `@BeforeClass`.

ChengyuanZhao: Oooh i understand now. Yes will do.

ChengyuanZhao: @meltsufin 

I switched to using `@BeforeClass and @AfterClass`, however the requirement is that these both be static methods. I went and changed all the things they need to also be static, but it doesn't seem like ApplicationContext is set by Spring before `@BeforeClass` can be called. I tried both using `@Autowire` application context and also by implementing ApplicationContextAware, but neither work.

I found somebody on SO had the same problem and he noted neither solution worked for him either: 
https://stackoverflow.com/questions/29496643/how-to-access-spring-applicationcontext-in-junit-beforeclass-static-method

Instead I found an alternative of using some static booleans that keep track of if the DB has been initialized and the regular non-static `@Before` set up method just checks that static boolean :
https://stackoverflow.com/questions/17505952/how-to-populate-database-only-once-before-test-methods-in-spring-test

meltsufin: The static boolean works I guess.

ChengyuanZhao: @meltsufin OK, changes are up. PTAL

PR #1152, opened 2018-10-29
Create Pub/Sub JSON Sample App w/ Spring Integration

dzou:
This adds a sample application demonstrating using Pub/Sub with JSON payload conversion and with Spring integration.

This is a rework of #1143; the application now utilizes Spring Integration to get more feature coverage.

### DIFF
--- a/spring-cloud-gcp-samples/spring-cloud-gcp-integration-pubsub-json-sample/src/test/java/com/example/PubSubJsonPayloadApplicationTest.java
+++ b/spring-cloud-gcp-samples/spring-cloud-gcp-integration-pubsub-json-sample/src/test/java/com/example/PubSubJsonPayloadApplicationTest.java
@@ -1,0 +1,97 @@
+/*
+ *  Copyright 2018 original author or authors.
+ *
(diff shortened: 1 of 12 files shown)


spring-cloud-gcp-samples/spring-cloud-gcp-integration-pubsub-json-sample/src/test/java/com/example/PubSubJsonPayloadApplicationTest.java
```diff
@@ -0,0 +1,99 @@
...
+import static org.junit.Assume.assumeThat;
+
+@RunWith(SpringRunner.class)
+@SpringBootTest(webEnvironment = WebEnvironment.RANDOM_PORT, classes = { PubSubJsonPayloadApplication.class })
+public class PubSubJsonPayloadApplicationTest {
+
+	private static final String SUBSCRIPTION_NAME = "exampleSubscription";
+
+	@LocalServerPort
+	private int port;
```
artembilan: This property is out of use.
Also would be better to follow conventions and name test classes with the `Tests` suffix.

PR #1153, opened 2018-10-29
Datastore @Reference write support and integration tests

dmitry-s:
part of #1092

### DIFF
--- a/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
+++ b/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
@@ -48,6 +48,7 @@ import org.springframework.cloud.gcp.data.datastore.core.mapping.DatastoreDataEx
 import org.springframework.cloud.gcp.data.datastore.core.mapping.DatastoreMappingContext;
 import org.springframework.cloud.gcp.data.datastore.core.mapping.DatastorePersistentEntity;
 import org.springframework.cloud.gcp.data.datastore.core.mapping.DatastorePersistentProperty;
+import org.springframework.cloud.gcp.data.datastore.core.util.ValueUtil;
 import org.springframework.data.domain.Sort;
 import org.springframework.data.mapping.PersistentProperty;
 import org.springframework.data.mapping.PersistentPropertyAccessor;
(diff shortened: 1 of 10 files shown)


spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
```diff
@@ -272,9 +274,40 @@ private Entity convertToEntityForSave(Object entity, Key... ancestors) {
...
+			}
+			Value<?> value;
+			if (persistentProperty.isCollectionLike()) {
+				if (val.getClass().isArray() && val.getClass() != byte[].class) {
+					//if a property is an array, convert it to list
+					val = CollectionUtils.arrayToList(val);
+				}
+				saveAll((Iterable<?>) val);
+				List<KeyValue> keyValues = StreamSupport.stream(((Iterable<?>) val).spliterator(), false)
+						.map(o -> KeyValue.of(this.getKey(o, false)))
```
ChengyuanZhao: Here we're not allocating new keys, but should we? I'm not thinking of any downsides right now to just allocating the keys for them, did you have some in mind?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
```diff
@@ -272,9 +274,40 @@ private Entity convertToEntityForSave(Object entity, Key... ancestors) {
...
+				.getPersistentEntity(entity.getClass());
+		datastorePersistentEntity.doWithReferenceProperties(persistentProperty -> {
+			PersistentPropertyAccessor accessor = datastorePersistentEntity.getPropertyAccessor(entity);
+			Object val = accessor.getProperty(persistentProperty);
+			if (val == null) {
+				return;
+			}
+			Value<?> value;
+			if (persistentProperty.isCollectionLike()) {
+				if (val.getClass().isArray() && val.getClass() != byte[].class) {
```
ChengyuanZhao: We're using this check so often  I think we should make this into a method for `DatastorePersistentProperty` called `isNonBytesArray`

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
dmitry-s: saveAll takes care of creating keys, so we don't need it to allocate them here

dmitry-s:
@ChengyuanZhao PTAL

PR #1155, opened 2018-10-30
Datastore ref doc for @reference

### LINKED ISSUE #1092: Datastore @Reference column-stored key POJO relationships

ChengyuanZhao:
fixes #1092 
Ref doc update for @Reference annotation


spring-cloud-gcp-docs/src/main/asciidoc/datastore.adoc
```diff
@@ -309,6 +311,47 @@ Since Cloud Datastore entity keys can have multiple parents, it is possible that
...
+}
+----
+
+Reference properties can either be singular or collection-like.
+These properties correspond to actual columns in the entity and Cloud Datastore Kind that hold the key values of the referenced entities.
+The referenced entities are full-fledged entities of other Kinds.
+
+Similar to the `@Descendants` relationships, reading or writing an entity will recursively read or write all of the referenced entities at all levels.
+If referenced entities have `null` ID values, then they will be saved as new entities and will have ID values allocated by Cloud Datastore.
+There are no requirements for relationships between the key of an entity and the keys that entity holds as references.
```
dmitry-s: mention that the order of the items in the referenced collection can be different on load

ChengyuanZhao: Yes good point. wil do

spring-cloud-gcp-docs/src/main/asciidoc/datastore.adoc
```diff
@@ -265,7 +265,7 @@ Embedded objects can be stored in the values of embedded Maps.
 Collections of embedded objects can also be stored as the values of embedded Maps.
 
 
-==== Relationships
```
meltsufin: I would keep a higher-level section called `Relationships` and just summarize all types of relationships that are supported. I believe we have three types: one-to-many via `@Descendants`, one-to-many or one-to-one using `@Reference`, and embedded entities. WDYT?

spring-cloud-gcp-docs/src/main/asciidoc/datastore.adoc
```diff
@@ -265,7 +265,7 @@ Embedded objects can be stored in the values of embedded Maps.
...
+==== Implicit Ancestor/Descendant Relationships
```
meltsufin: I'm not sure about the use of the term "implicit" here. Maybe just "Ancestor-descendant Relationships"?

spring-cloud-gcp-docs/src/main/asciidoc/datastore.adoc
```diff
@@ -309,6 +311,47 @@ Since Cloud Datastore entity keys can have multiple parents, it is possible that
 Because entity keys are immutable in Cloud Datastore, to change the key of a child you must delete the existing one and re-save it with the new key.
 
 
+==== Explicit Key-Column Reference Relationships
```
meltsufin: Again, maybe just "Key Reference Relationships"?

spring-cloud-gcp-docs/src/main/asciidoc/datastore.adoc
ChengyuanZhao: Makes sense.  I will reorganize.

ChengyuanZhao:
@meltsufin @dmitry-s PTAL, changes made

PR #1160, opened 2018-10-30
Datastore sample shows transactions, references, descendants. 

### LINKED ISSUE #1115: Datastore @Transactional integration test and sample

ChengyuanZhao:
fixes #1115 
sample shows transactions in 2 ways, references, descendants.

### DIFF
--- a/spring-cloud-gcp-samples/spring-cloud-gcp-data-datastore-sample/src/main/java/com/example/DatastoreRepositoryExample.java
+++ b/spring-cloud-gcp-samples/spring-cloud-gcp-data-datastore-sample/src/main/java/com/example/DatastoreRepositoryExample.java
@@ -61,54 +61,63 @@ public class DatastoreRepositoryExample {
 							new Album("b", LocalDate.of(2018, Month.FEBRUARY, 12))));
 			Singer scottSmith = new Singer("singer3", "Scott", "Smith", ImmutableSet
 					.of(new Album("c", LocalDate.of(2000, Month.AUGUST, 31))));
+
 			this.singerRepository.saveAll(ImmutableList.of(maryJane, scottSmith));
 
-			Band band1 = new Band("band1");
-			Band band2 = new Band("band2");
-			Band band3 = new Band("band3");
-
-			// Creates the related Band and Instrument entities and links them to a Singer
-			// in a single atomic transaction
-			this.transactionalRepositoryService
-					.createAndSaveSingerRelationshipsInTransaction(maryJane, band1,
-							Arrays.asList(band1, band2),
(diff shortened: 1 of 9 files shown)


ChengyuanZhao:
@meltsufin @dmitry-s This doesn't work until the `@Reference` write PR goes in. So I will ping you when that one goes in.

ChengyuanZhao:
@meltsufin @dmitry-s the `@Reference` write is in so this sample is working now. Please take a look and give it a try!

ChengyuanZhao:
@dmitry-s added to test! PTAL !

spring-cloud-gcp-samples/spring-cloud-gcp-data-datastore-sample/src/main/java/com/example/DatastoreRepositoryExample.java
```diff
@@ -49,16 +53,41 @@ public CommandLineRunner commandLineRunner() {
 			System.out.println("Remove all records from 'singers' kind");
 			this.singerRepository.deleteAll();
 
-			this.singerRepository.save(new Singer("singer1", "John", "Doe", ImmutableSet.of()));
```
meltsufin: This example has become one big kitchen sink, which is very difficult to follow for me, let alone for someone new to Datastore. Is there any way this big method can be broken out into smaller sub-examples that it calls?

ChengyuanZhao: The issue is with running these though. There has to be just one main method. 

Do you mean just moving some code into some private methods that get called in main like "demoTransactions" ?

ChengyuanZhao: @meltsufin Also I gotta say it's kind of by-design. We specifically wanted to show all these advanced features in a sample, which I think might be contrary to the nature and purpose of samples in general.

ChengyuanZhao: Moved some code into some smaller methods

meltsufin: Thanks! Looks better!

ChengyuanZhao:
@meltsufin i forgot that the retrieved child entities aren't always in the same order, so I had to update the test for that. 
PTAL

PR #1126, opened 2018-10-19
[WIP] Extract IAP identity from HTTP header

elefeint:
This PR will be split into smaller ones.

Resolves: #1016

Issue #1016, opened 2018-09-13
FR: Spring Security Autoconfiguration to validate IAP header and extract principle


saturnism:
See https://cloud.google.com/iap/docs/signed-headers-howto

meltsufin:
Note that Identity-Aware Proxy is currently in Beta. So, we can't yet include it in our GA releases.

meltsufin:
Disregard the previous comment. I guess it's only the case for App Engine Flex that it is in Beta.

PR #1165, opened 2018-10-31
Add Cloud IAP pre-authentication filter

elefeint:
Core module for extracting identity from [Cloud Identity-Aware Proxy (Cloud IAP)](https://cloud.google.com/iap/) header.

The header is extracted through `IapAuthenticationFilter`, which is enabled by factory-configured (see _spring.factories_) `IapHttpConfigurer`. The header is then:
* decoded
* its signature verified against a public key. The public keys are cached by `JwkRegistry`, which downloads them when a key is not found in the cache, provided that at least a minute has passed.
* minimal claims expected by IAP (expiration date, issue date, audience, presence of subject and e-mail) verified. Issue and expiration date verification is delegated to [nimbus-jose-jwt ](https://connect2id.com/products/nimbus-jose-jwt) library, which also handles all token and public key parsing.

`IapAuthenticationFilter` is an extension of Spring security standard `AbstractPreAuthenticatedProcessingFilter`; the overridden methods provide the identity/principal (`IapAuthentication`) and "credentials" (actually the full JWT token), provided that the JWT token passed signature and claims verification.

The included `IapAuthenticationDetailsSource` simply returns a default role (ROLE_USER or a constructor-injected alternative) as details for any request, since this is a pre-authenticated scenario.

This PR addresses most of #1016.

This PR does not yet (but will) include:
* tests
* documentation
* proper javadoc (I added a few snippets where intent was unclear).

This PR will never include:
* autoconfiguration
* starter
* sample

For reference, #1126 included everything, but at this point will not be updated in favor of smaller* PRs.


meltsufin:
@elefeint I think the size of the PR is fine. A lot of the lines are headers and such.

spring-cloud-gcp-security-iap/src/main/java/org/springframework/cloud/gcp/security/iap/jwt/JwtECSignatureVerifier.java
```diff
@@ -0,0 +1,86 @@
...
+			}
+		}
+
+		return false;
+	}
+
+	private boolean verifyAgainstPublicKey(SignedJWT signedJwt, ECPublicKey publicKey) {
+		JWSVerifier jwsVerifier = null;
+		try {
+			jwsVerifier = new ECDSAVerifier(publicKey);
```
meltsufin: Should this be a dependency instead?

spring-cloud-gcp-security-iap/src/main/java/org/springframework/cloud/gcp/security/iap/jwt/JwtTokenVerifier.java
```diff
@@ -0,0 +1,132 @@
...
+		this.claimVerifier = claimVerifier;
+	}
+
+	/**
+	 * Builds up a default verifier assuming ECDSA signed key, plus common-sense
+	 * org.springframework.cloud.gcp.security.iap.claims checks.
+	 */
+	public JwtTokenVerifier(URL jwkRegistryUrl) {
+		this(new JwtECSignatureVerifier(jwkRegistryUrl),
+				new CompositeClaimVerifier(
```
meltsufin: Can you create a verifier class that just composes this set?

spring-cloud-gcp-security-iap/pom.xml
```diff
@@ -0,0 +1,43 @@
...
+        </dependency>
+
+        <dependency>
+            <groupId>com.nimbusds</groupId>
+            <artifactId>nimbus-jose-jwt</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>javax.servlet</groupId>
+            <artifactId>javax.servlet-api</artifactId>
```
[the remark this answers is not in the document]
elefeint: I was hoping it was not, but AbstractPreAuthenticatedProcessingFilter calls the implementation of AuthenticationDetailsSource with a plain HttpServletRequest, so the dependency is needed for IapAuthenticationDetailsSource to compile.

spring-cloud-gcp-security-iap/src/main/java/org/springframework/cloud/gcp/security/iap/jwt/JwtECSignatureVerifier.java
elefeint: Looked at it some more, and no, because the verifier needs to be created from a public key (which theoretically may differ), it can't be a singleton. For now, I've sequestered all algorithm-specific logic in JwtECSignatureVerifier, so that if anyone wants a different algorithm, the verifier bean can be replaced by a custom one.
There is, I think, a valid case to be made that the per-key verifiers should be cached instead of just keys. I might try that today, after adding docs.

elefeint:
And another one gone, and another one gone, another one bites the dust!
Closing custom solution in favor of spring security oauth built-in.

PR #1169, opened 2018-11-01
Create Separate Subscription in Pub/Sub JSON Test

dzou:
This modifies the Pub/Sub JSON Payload test to create a separate test subscription which subscribes to the topic that is producing the messages.

I noticed that if we subscribe to the existing subscription used by the sample app, sometimes we would not receive a message in the test since the message is already acked by the subscriber in the sample application. So this will increase reliability of the test.

Followup to #1152.

### DIFF
--- a/spring-cloud-gcp-samples/spring-cloud-gcp-integration-pubsub-json-sample/src/main/java/com/example/Person.java
+++ b/spring-cloud-gcp-samples/spring-cloud-gcp-integration-pubsub-json-sample/src/main/java/com/example/Person.java
@@ -16,6 +16,8 @@
 
 package com.example;
 
+import java.util.Objects;
+
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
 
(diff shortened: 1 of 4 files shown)


artembilan:
> sometimes we would not receive a message in the test 

Wait. That sounds like your test doesn't really test the sample application, but something independent.
Shoudn't we rethink the test that way we get interaction with the provided application?

Thanks

dzou:
> > sometimes we would not receive a message in the test
> 
> Wait. That sounds like your test doesn't really test the sample application, but something independent.
> Shoudn't we rethink the test that way we get interaction with the provided application?
> 
> Thanks

I see what you mean. Right now the App just logs to console which is not easily testable behavior; I think I will add an additional rest endpoint to the sample app so that the test will directly exercise the app.

dzou:
Hi guys, I modified the sample to expose a get endpoint that returns all the JSON payloads received by the Pub/Sub subscriber in the sample app. Now the test will actually test the app rather than in the before case where I created a new subscriber in the test which listened to the same topic the sample app was listening to. PTAL.

PR #1173, opened 2018-11-02
Datastore Template Map Operations

### LINKED ISSUE #1138: Add saveMap and readMap to DatastoreOperations
Need to check if it's possible at all.

dmitry-s:
fixes #1138

### DIFF
--- a/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreOperations.java
+++ b/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreOperations.java
@@ -17,6 +17,7 @@
 package org.springframework.cloud.gcp.data.datastore.core;
 
 import java.util.Collection;
+import java.util.Map;
 import java.util.function.Function;
 
 import com.google.cloud.datastore.BaseEntity;
@@ -195,4 +196,29 @@ public interface DatastoreOperations {
 	 * @return the final result of the transaction.
 	 */
 	<A> A performTransaction(Function<DatastoreOperations, A> operations);
+
+	/**
+	 * Get a Datastore entity based on a id and convert it to a map.
+	 * @param key the key of the entity
+	 * @param valueType type values should be converted to
+	 * @return if an entity for a given key exists, returns the map representation of it,
+	 * {@code null} otherwise
+	 */
+	<T> Map<String, T> findByIdAsMap(Key key, Class<T> valueType);
+
+	/**
+	 * Save map as a Datastore entity
+	 * @param key the key for the entity
+	 * @param map type information that is used to convert a datastore entity to a map
+	 */
+	<V> void writeMap(Key key, Map<String, V> map);
+
+	/**
+	 * Create a {@link com.google.cloud.datastore.Key} from kind name and id.
+	 * @param kind the kind name
+	 * @param id object to be used as id; if it is a Long, the value is used, otherwise it is
+	 * converted to String
+	 * @return created key
+	 */
+	Key createKey(String kind, Object id);
 }
--- a/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreOperations.java
+++ b/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreOperations.java
@@ -215,7 +215,7 @@ public interface DatastoreOperations {
 
 	/**
 	 * Create a {@link com.google.cloud.datastore.Key} from kind name and id.
-	 * @param kind the kind name
+	 * @param kind the Cloud Datastore kind name
 	 * @param id object to be used as id; if it is a Long, the value is used, otherwise it is
 	 * converted to String
 	 * @return created key
--- a/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreOperations.java
+++ b/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreOperations.java
@@ -207,9 +207,9 @@ public interface DatastoreOperations {
 	<T> Map<String, T> findByIdAsMap(Key key, Class<T> valueType);
 
 	/**
-	 * Save map as a Datastore entity
+	 * Save a map as a Datastore entity
 	 * @param key the key for the entity
-	 * @param map type information that is used to convert a datastore entity to a map
+	 * @param map a map
 	 */
 	<V> void writeMap(Key key, Map<String, V> map);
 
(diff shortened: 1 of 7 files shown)


spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreOperations.java
```diff
@@ -195,4 +196,29 @@
...
+	 * @param valueType type values should be converted to
+	 * @return if an entity for a given key exists, returns the map representation of it,
+	 * {@code null} otherwise
+	 */
+	<T> Map<String, T> findByIdAsMap(Key key, Class<T> valueType);
+
+	/**
+	 * Save map as a Datastore entity
+	 * @param key the key for the entity
+	 * @param map type information that is used to convert a datastore entity to a map
```
ChengyuanZhao: I think the javadoc for this param is from an older version?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
```diff
@@ -255,6 +257,42 @@ public static void applyQueryOptions(StructuredQuery.Builder builder, DatastoreQ
...
+		return map;
+	}
+
+	@Override
+	public <V> void writeMap(Key datastoreKey, Map<String, V> map) {
+		Assert.notNull(datastoreKey, "A non-null Key is required.");
+		Assert.notNull(map, "A non-null map is required.");
+
+		Builder builder = Entity.newBuilder(datastoreKey);
+		map.forEach(
```
ChengyuanZhao: These two little anonymous functions calling `convertOnWriteSingle` are very simple, but would it be possible to use the existing `readAsMap` and `write` methods from `DefaultDatastoreEntityConverter` ?  

It seems like what we're doing here is just a small subset of what those bigger methods can handle, so maybe it would be possible to use those here?

dmitry-s: Makes sense for `readAsMap`, good catch! 
But it is not going to work with `write` because it uses  persistent properties.

dmitry-s:
@ChengyuanZhao @meltsufin PTAL

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreOperations.java
```diff
@@ -195,4 +196,29 @@
...
+	 * @param valueType type values should be converted to
+	 * @return if an entity for a given key exists, returns the map representation of it,
+	 * {@code null} otherwise
+	 */
+	<T> Map<String, T> findByIdAsMap(Key key, Class<T> valueType);
+
+	/**
+	 * Save a map as a Datastore entity
+	 * @param key the key for the entity
+	 * @param map a map
```
meltsufin: I think we can do better for javadoc.
Maybe something describing the fact that the map keys will be used as column names?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreOperations.java
```diff
@@ -195,4 +196,29 @@
...
+	 */
+	<V> void writeMap(Key key, Map<String, V> map);
+
+	/**
+	 * Create a {@link com.google.cloud.datastore.Key} from kind name and id.
+	 * @param kind the kind name
```
meltsufin:
```suggestion
	 * @param kind the Cloud Datastore kind name
```

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
```diff
@@ -255,6 +257,32 @@ public static void applyQueryOptions(StructuredQuery.Builder builder, DatastoreQ
...
+				(key, value) ->
+						builder.set(key, this.datastoreEntityConverter.getConversions().convertOnWriteSingle(value)));
+		Entity entity = builder.build();
+		getDatastoreReadWriter().put(entity);
+	}
+
+	public Key createKey(String kind, Object id) {
```
meltsufin: `@Override`?
```suggestion
	@Override
	public Key createKey(String kind, Object id) {
```

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/ReadWriteConversions.java
```diff
@@ -65,6 +65,13 @@
 	 */
 	Value convertOnWrite(Object obj, DatastorePersistentProperty persistentProperty);
 
+	/**
+	 * Converts an object to a Cloud Datastore {@link Value}, for non-collection objects
```
meltsufin: In the method above, you should probably then say that it supports collection objects as well.
I'm not clear on why we need both methods exposed though. Why can't the method above be used to convert either collection or single?

dmitry-s: We need this one because we don't have a persistent property for the map.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreOperations.java
```diff
@@ -195,4 +196,29 @@
...
+	 * Get a Datastore entity based on a id and convert it to a map.
+	 * @param key the key of the entity
+	 * @param valueType type values should be converted to
+	 * @return if an entity for a given key exists, returns the map representation of it,
+	 * {@code null} otherwise
+	 */
+	<T> Map<String, T> findByIdAsMap(Key key, Class<T> valueType);
+
+	/**
+	 * Save a map as a Datastore entity, using map keys as a field names
```
meltsufin:
```suggestion
	 * Save a map as a Datastore entity, using map keys as field names
```

spring-cloud-gcp-docs/src/main/asciidoc/datastore.adoc
```diff
@@ -671,6 +671,28 @@ This feature requires a bean of `DatastoreTransactionManager`, which is provided
 If a method annotated with `@Transactional` calls another method also annotated, then both methods will work within the same transaction.
 `performTransaction` cannot be used in `@Transactional` annotated methods because Cloud Datastore does not support transactions within transactions.
 
+==== Read write support for maps
```
```suggestion
==== Read-write support for maps
```

PR #1176, opened 2018-11-02
Datastore spel query methods

### LINKED ISSUE #986: Datastore SpEL in query methods support

ChengyuanZhao:
fixes #986
adding SpEL expression support to query methods

### DIFF
--- a/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/GqlDatastoreQuery.java
+++ b/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/GqlDatastoreQuery.java
@@ -59,6 +59,9 @@ public class GqlDatastoreQuery<T> extends AbstractDatastoreQuery<T> {
 	// A small string that isn't used in GQL syntax
 	private static final String ENTITY_CLASS_NAME_BOOKEND = "|";
 
+	private static final Pattern CLASS_NAME_PATTERN = Pattern.compile("\\" + ENTITY_CLASS_NAME_BOOKEND + "\\S+\\"
+			+ ENTITY_CLASS_NAME_BOOKEND + "");
+
 	private final String originalGql;
 
 	private String gqlResolvedEntityClassName;
@@ -107,10 +110,11 @@ public class GqlDatastoreQuery<T> extends AbstractDatastoreQuery<T> {
 	@Override
 	public Object execute(Object[] parameters) {
 
-		QueryTagsValues queryTagsValues = getQueryTagsValues(parameters);
+		ParsedQueryWithTagsAndValues parsedQueryWithTagsAndValues = new ParsedQueryWithTagsAndValues(
+				getOriginalParamTags(), parameters);
 
-		GqlQuery query = bindArgsToGqlQuery(queryTagsValues.finalGql,
-				queryTagsValues.tagsOrdered, queryTagsValues.params);
+		GqlQuery query = bindArgsToGqlQuery(parsedQueryWithTagsAndValues.finalGql,
+				parsedQueryWithTagsAndValues.tagsOrdered, parsedQueryWithTagsAndValues.params);
 
 		boolean returnTypeIsCollection = this.queryMethod.isCollectionQuery();
 		Class returnedItemType = this.queryMethod.getReturnedObjectType();
@@ -148,7 +152,29 @@ public class GqlDatastoreQuery<T> extends AbstractDatastoreQuery<T> {
 
 	private String getGqlResolvedEntityClassName() {
 		if (this.gqlResolvedEntityClassName == null) {
-			this.gqlResolvedEntityClassName = resolveEntityClassNames(this.originalGql);
+			Matcher matcher = CLASS_NAME_PATTERN.matcher(this.originalGql);
+			this.gqlResolvedEntityClassName = this.originalGql;
+			while (matcher.find()) {
+				String matched = matcher.group();
+				String className = matched.substring(1, matched.length() - 1);
+				try {
+					Class entityClass = Class.forName(className);
+					DatastorePersistentEntity datastorePersistentEntity = this.datastoreMappingContext
+							.getPersistentEntity(entityClass);
+					if (datastorePersistentEntity == null) {
+						throw new DatastoreDataException(
+								"The class used in the GQL statement is not a Cloud Datastore persistent entity: "
+										+ className);
+					}
+					this.gqlResolvedEntityClassName = this.gqlResolvedEntityClassName.replace(matched,
+							datastorePersistentEntity.kindName());
+				}
+				catch (ClassNotFoundException e) {
+					throw new DatastoreDataException(
+							"The class name does not refer to an available entity type: "
+									+ className);
+				}
+			}
 		}
 		return this.gqlResolvedEntityClassName;
 	}
@@ -221,32 +247,13 @@ public class GqlDatastoreQuery<T> extends AbstractDatastoreQuery<T> {
 		if (tags.size() != vals.size()) {
 			throw new DatastoreDataException("Annotated GQL Query Method "
 					+ this.queryMethod.getName() + " has " + tags.size()
-					+ " initialTags but a different number of parameter values: " + vals.size());
+					+ " tags but a different number of parameter values: " + vals.size());
 		}
 		for (int i = 0; i < tags.size(); i++) {
 			Object val = vals.get(i);
 			DatastoreNativeTypes.bindValueToGqlBuilder(builder, tags.get(i), val);
 		}
 		return builder.build();
-	}
-
-	private QueryTagsValues getQueryTagsValues(Object[] parameters) {
-		QueryTagsValues queryTagsValues = new QueryTagsValues(getOriginalParamTags(),
-				parameters);
-
-		SpelQueryContext.EvaluatingSpelQueryContext spelQueryContext = getEvaluatingSpelQueryContext();
-
-		SpelEvaluator spelEvaluator = spelQueryContext.parse(getGqlResolvedEntityClassName(),
-				this.queryMethod.getParameters());
-		Map<String, Object> results = spelEvaluator.evaluate(queryTagsValues.rawParams);
-		queryTagsValues.finalGql = spelEvaluator.getQueryString();
-
-		for (Map.Entry<String, Object> entry : results.entrySet()) {
-			queryTagsValues.params.add(entry.getValue());
-			// Cloud Datastore requires the tag name without the
-			queryTagsValues.tagsOrdered.add(entry.getKey().substring(1));
-		}
-		return queryTagsValues;
 	}
 
 	private SpelQueryContext.EvaluatingSpelQueryContext getEvaluatingSpelQueryContext() {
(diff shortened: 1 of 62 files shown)


ChengyuanZhao:
@meltsufin @dmitry-s PTAL

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/GqlDatastoreQuery.java
```diff
@@ -167,43 +206,156 @@ boolean isNonEntityReturnedType(Class returnedType) {
...
+		EvaluationContext evaluationContext = this.evaluationContextProvider
+				.getEvaluationContext(this.queryMethod.getParameters(),
+						queryTagValue.rawParams);
+		for (Expression expression : expressions) {
+			if (expression instanceof LiteralExpression) {
+				sb.append(expression.getValue(String.class));
+			}
+			else if (expression instanceof SpelExpression) {
+				Object value = expression.getValue(evaluationContext);
+				if (valueToTag.containsKey(value)) {
```
dmitry-s: use `computeIfAbsent()` instead of this `if`block

spring-cloud-gcp-data-datastore/src/test/java/org/springframework/cloud/gcp/data/datastore/repository/query/GqlDatastoreQueryTests.java
```diff
@@ -149,6 +153,10 @@ public void compoundNameConventionTest() {
 			assertEquals(params[5], paramMap.get("tag5").get());
 			assertEquals(params[6], paramMap.get("tag6").get());
 			assertEquals(params[7], paramMap.get("tag7").get());
+			assertEquals(-1 * (double) params[6], (double) paramMap.get("SpELtag1").get(),
+					0.00001);
+			assertEquals(-1 * (double) params[7], (double) paramMap.get("SpELtag2").get(),
+					0.00001);
```
dmitry-s: maybe add integration tests?

spring-cloud-gcp-docs/src/main/asciidoc/datastore.adoc
```diff
@@ -786,7 +786,7 @@ Custom GQL queries can be mapped to repository methods in one of two ways:
 
 Using the `@Query` annotation:
 
-The names of the tags of the GQL correspond to the `@Param` annotated names of the method parameters.
+The names of the initialTags of the GQL correspond to the `@Param` annotated names of the method parameters.
```
dmitry-s: why is it called initialTags now?

dmitry-s: unrelated to this PR, but in line 796 we are missing a `"`
`(@Param("trader_name) String traderName)`

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/GqlDatastoreQuery.java
```diff
@@ -167,43 +206,156 @@ boolean isNonEntityReturnedType(Class returnedType) {
...
+			return ((CompositeStringExpression) expression).getExpressions();
+		}
+		else {
+			throw new DatastoreDataException("Unexpected expression type. "
+					+ "Query can either contain no SpEL expressions or have only "
+					+ "literal SpEL expressions in the GQL.");
+		}
+	}
+
+	private String resolveEntityClassNames(String sql) {
```
dmitry-s: add comments or java doc

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/GqlDatastoreQuery.java
```diff
@@ -167,43 +206,156 @@ boolean isNonEntityReturnedType(Class returnedType) {
...
+				if (valueToTag.containsKey(value)) {
+					sb.append("@").append(valueToTag.get(value));
+				}
+				else {
+					String newTag;
+					do {
+						tagNum++;
+						newTag = "SpELtag" + tagNum;
+					}
+					while (queryTagValue.initialTags.contains(newTag));
```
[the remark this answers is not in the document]
ChengyuanZhao: Could you clarify which parts? I don't think it makes sense to make a little private method that just holds a small while loop to which I need to pass the current tagNum.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/GqlDatastoreQuery.java
ChengyuanZhao: Could you clarify? We want to do something different in both cases, not just only get-or-compute.

spring-cloud-gcp-docs/src/main/asciidoc/datastore.adoc
ChengyuanZhao: I think this was a refactor auto-rename that did something wrong., will revert.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/GqlDatastoreQuery.java
ChengyuanZhao: We don't javadoc private methods, but I can add a comment.

spring-cloud-gcp-docs/src/main/asciidoc/datastore.adoc
```diff
@@ -836,6 +836,27 @@ You can also query for non-entity types:
 	String getOneId(@Param("id_val") long id);
 ----
 
+Kind names can also be resolved from the `@Entity` annotation on domain classes.
```
meltsufin: What does the "also" refer to? I don't see any discussion about other ways to get kind names right before this paragraph.
In any case, how is this `|com.example.Trade|` syntax relevant to SpEL? I don't remember seeing anything like this in other Spring Data implementations.

ChengyuanZhao: Also implies there are at least 2 methods.

methods 1 is directly writing the name in the string.
method 2 is using the java class name. 

This is a feature from spring data JPA.

The reason SpEL is here is because the Kind name annotation for the class can have SpEL expressions  in it.

ChengyuanZhao:
@meltsufin I thought we were gonna wait until speaking to the Spring Data team due to the differences in syntax between their expression extractor and the rather specific-looking spel expression evaluator.

meltsufin:
@ChengyuanZhao Do we have a specific list of questions like that for them. I can try to setup a meeting for tomorrow with them then.

ChengyuanZhao:
@meltsufin As we discussed, the sole question is why in their implementation of the spel extractor they use a regex matching string that has a preceding colon or question mark instead of using the ready-made `org.springframework.expression.spel.standard.SpelExpressionParser`, which seems to have more more advanced expression parsing. As a result neither way looks like the clear standard way of doing it.

meltsufin:
Can we just file this question in `spring-data-commons` issue tracker, or do you want to talk to them about it on the a call?

ChengyuanZhao:
@meltsufin that project doesn't have a public issues tab, so I don't think I could make an issue there. Let's just try mentioning the team by name here: @mp911de

ChengyuanZhao:
@mp911de  Mark, thank you that is very helpful!

ChengyuanZhao:
@mp911de  Mark, do you mean that `SpelQueryContext` and its companion classes `SpelExtractor` and `SpelEvaluator`  supports `:foo` and `?0` style tags? I see in `SpelQueryContext` that the regex `SPEL_PATTERN_STRING = "([:?])#\\{([^}]+)}";` is used to match expressions, which means `#{}` must always be there, so those 2 examples wouldn't be picked up.

ChengyuanZhao:
@dmitry-s @meltsufin PTAL

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/GqlDatastoreQuery.java
```diff
@@ -132,6 +146,13 @@ public Object execute(Object[] parameters) {
 		return result;
 	}
 
+	private String getGqlResolvedEntityClassName() {
+		if (this.gqlResolvedEntityClassName == null) {
```
dmitry-s: maybe move caching into `resolveEntityClassNames(this.originalGql)` method and remove this one?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/GqlDatastoreQuery.java
```diff
@@ -167,43 +188,132 @@ boolean isNonEntityReturnedType(Class returnedType) {
...
-			Object[] vals) {
+			List vals) {
 		Builder builder = GqlQuery.newGqlQueryBuilder(gql);
 		builder.setAllowLiteral(true);
-		if (tags.size() != vals.length) {
+		if (tags.size() != vals.size()) {
 			throw new DatastoreDataException("Annotated GQL Query Method "
 					+ this.queryMethod.getName() + " has " + tags.size()
-					+ " tags but a different number of parameter values: " + vals.length);
+					+ " initialTags but a different number of parameter values: " + vals.size());
```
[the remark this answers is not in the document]
ChengyuanZhao: ah bad auto-rename.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/GqlDatastoreQuery.java
```diff
@@ -97,7 +106,12 @@ private static Object getNonEntityObjectFromRow(Object x) {
 
 	@Override
 	public Object execute(Object[] parameters) {
-		GqlQuery query = bindArgsToGqlQuery(this.gql, getParamTags(), parameters);
+
+		QueryTagsValues queryTagsValues = getQueryTagsValues(parameters);
+
+		GqlQuery query = bindArgsToGqlQuery(queryTagsValues.finalGql,
```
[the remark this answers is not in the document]
ChengyuanZhao: We don't need everything in queryTagsValues, but just those 3 parts.

ChengyuanZhao:
@dmitry-s changes made ! PTAL !

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/GqlDatastoreQuery.java
```diff
@@ -132,6 +147,35 @@ public Object execute(Object[] parameters) {
 		return result;
 	}
 
+	private String getGqlResolvedEntityClassName() {
```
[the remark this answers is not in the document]
ChengyuanZhao: done

spring-cloud-gcp-docs/src/main/asciidoc/datastore.adoc
```diff
@@ -870,6 +870,29 @@ You can also query for non-entity types:
 	String getOneId(@Param("id_val") long id);
 ----
 
+SpEL can be used to provide GQL parameters:
+
+[source, java]
+----
+@Query("SELECT * FROM |com.example.Trade| WHERE trades.action = @act
+  AND price > :#{#priceRadius * -1} AND price < :#{#priceRadius * 2}")
```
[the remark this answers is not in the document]
ChengyuanZhao: Yes, it is a side effect of using that alternate method because its regex requires a prefix.

PR #1177, opened 2018-11-02
Anonymous consumer groups in Pub/Sub Binder

### LINKED ISSUE #1172: Pub/Sub Binder does not handle anonymous consumer groups
According to [Cloud Stream Docs](https://docs.spring.io/spring-cloud-stream/docs/current/reference/htmlsingle/#consumer-groups):

>  By default, when a group is not specified, Spring Cloud Stream assigns the application to an anonymous and independent single-member consumer group that is in a publish-subscribe relationship with all other consumer groups.

However, in our implementation, we always use the same subscription for all applications that do not specify a consumer group. This causes all consumers to effectively belong to the same consumer group, which is the exact opposite of the specification.

Supersedes #1162

meltsufin:
When a group is not provided for the consumer destination, generate a
random anonymous one.

This will make the binder compatible with Spring Cloud Bus and its use
of anonymous groups.

Fixes #1172

### DIFF
--- a/spring-cloud-gcp-docs/src/main/asciidoc/spring-stream.adoc
+++ b/spring-cloud-gcp-docs/src/main/asciidoc/spring-stream.adoc
@@ -29,25 +29,45 @@ A https://github.com/spring-cloud/spring-cloud-gcp/tree/master/spring-cloud-gcp-
 
 This binder binds producers to Google Cloud Pub/Sub topics and consumers to subscriptions.
 
-NOTE: Partitioning and consumer groups are not currently supported by this binder.
+NOTE: Partitioning is currently not supported by this binder.
 
 === Configuration
 
-You can configure the Spring Cloud Stream Binder for Google Cloud Pub/Sub to automatically generate the underlying resources, like the Google Cloud Pub/Sub subscriptions for the consumers.
-For that, you can use the `spring.cloud.stream.gcp.pubsub.bindings.[CHANNEL-NAME].consumer.auto-create-resources` property, which is turned ON by default.
+You can configure the Spring Cloud Stream Binder for Google Cloud Pub/Sub to automatically generate the underlying resources, like the Google Cloud Pub/Sub topics and subscriptions for producers and consumers.
+For that, you can use the `spring.cloud.stream.gcp.pubsub.bindings.<channelName>.<consumer|producer>.auto-create-resources` property, which is turned ON by default.
 
-If automatic resource creation is turned ON and the subscription and the topic do not exist for a consumer, a subscription and a topic will be created with the same name.
-For example, for the following configuration, a topic and a subscription called `myConsumer` would be created.
-
-.application.properties
-----
-spring.cloud.stream.bindings.output.destination=myConsumer
-
-spring.cloud.stream.gcp.pubsub.bindings.output.consumer.auto-create-resources=true
-----
-
-Starting with version 1.1, these properties can be configured globally for all the bindings, e.g. `spring.cloud.stream.gcp.pubsub.default.consumer.auto-create-resources`.
+Starting with version 1.1, these and other binder properties can be configured globally for all the bindings, e.g. `spring.cloud.stream.gcp.pubsub.default.consumer.auto-create-resources`.
 
 If you are using Pub/Sub auto-configuration from the Spring Cloud GCP Pub/Sub Starter, you should refer to the <<pubsub-configuration,configuration>> section for other Pub/Sub parameters.
 
 NOTE: To use this binder with a https://cloud.google.com/pubsub/docs/emulator[running emulator], configure its host and port via `spring.cloud.gcp.pubsub.emulator-host`.
+
+==== Producer Destination Configuration
+
+If automatic resource creation is turned ON and the topic corresponding to the destination name does not exist, it will be created.
+
+For example, for the following configuration, a topic called `myEvents` would be created.
+
+.application.properties
+----
+spring.cloud.stream.bindings.events.destination=myEvents
+spring.cloud.stream.gcp.pubsub.bindings.events.producer.auto-create-resources=true
+----
+
+==== Consumer Destination Configuration
+
+If automatic resource creation is turned ON and the subscription and/or the topic do not exist for a consumer, a subscription and potentially a topic will be created.
+The topic name will be the same as the destination name, and the subscription name will be the same as the consumer group name.
+If the consumer group is not specified, an anonymous one will be created with the name `anonymous.<destinationName>.<randomUUID>`.
+
+For example, for the following configuration, a topic named `myEvents` and a subscription called `counsumerGroup1` would be created.
+If the consumer group is not specified, a subscription called `anonymous.myEvents.a6d83782-c5a3-4861-ac38-e6e2af15a7be` would be created.
+
+.application.properties
+----
+spring.cloud.stream.bindings.events.destination=myEvents
+spring.cloud.stream.gcp.pubsub.bindings.events.consumer.auto-create-resources=true
+
+# specify consumer group, and avoid anonymous consumer group generation
+spring.cloud.stream.bindings.events.group=consumerGroup1
+----
--- a/spring-cloud-gcp-docs/src/main/asciidoc/spring-stream.adoc
+++ b/spring-cloud-gcp-docs/src/main/asciidoc/spring-stream.adoc
@@ -57,10 +57,10 @@ spring.cloud.stream.gcp.pubsub.bindings.events.producer.auto-create-resources=tr
 ==== Consumer Destination Configuration
 
 If automatic resource creation is turned ON and the subscription and/or the topic do not exist for a consumer, a subscription and potentially a topic will be created.
-The topic name will be the same as the destination name, and the subscription name will be the same as the consumer group name.
+The topic name will be the same as the destination name, and the subscription name will be the destination name followed by the consumer group name.
 If the consumer group is not specified, an anonymous one will be created with the name `anonymous.<destinationName>.<randomUUID>`.
 
-For example, for the following configuration, a topic named `myEvents` and a subscription called `counsumerGroup1` would be created.
+For example, for the following configuration, a topic named `myEvents` and a subscription called `myEvents.counsumerGroup1` would be created.
 If the consumer group is not specified, a subscription called `anonymous.myEvents.a6d83782-c5a3-4861-ac38-e6e2af15a7be` would be created.
 
 .application.properties
--- a/spring-cloud-gcp-docs/src/main/asciidoc/spring-stream.adoc
+++ b/spring-cloud-gcp-docs/src/main/asciidoc/spring-stream.adoc
@@ -57,10 +57,10 @@ spring.cloud.stream.gcp.pubsub.bindings.events.producer.auto-create-resources=tr
 ==== Consumer Destination Configuration
 
 If automatic resource creation is turned ON and the subscription and/or the topic do not exist for a consumer, a subscription and potentially a topic will be created.
-The topic name will be the same as the destination name, and the subscription name will be the destination name followed by the consumer group name.
+The topic name will be the same as the destination name, and the subscription name will be the same as the consumer group name.
 If the consumer group is not specified, an anonymous one will be created with the name `anonymous.<destinationName>.<randomUUID>`.
 
-For example, for the following configuration, a topic named `myEvents` and a subscription called `myEvents.counsumerGroup1` would be created.
+For example, for the following configuration, a topic named `myEvents` and a subscription called `counsumerGroup1` would be created.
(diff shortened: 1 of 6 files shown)


spring-cloud-gcp-pubsub-stream-binder/src/main/java/org/springframework/cloud/gcp/stream/binder/pubsub/provisioning/PubSubChannelProvisioner.java
```diff
@@ -54,8 +58,10 @@ public ProducerDestination provisionProducerDestination(String name,
 	public ConsumerDestination provisionConsumerDestination(String name, String group,
 			ExtendedConsumerProperties<PubSubConsumerProperties> properties)
 			throws ProvisioningException {
+		// Use group name as subscription name
+		// Generate anonymous random group, if one not provided
```
meltsufin: @artembilan The concern I have here is that there doesn't seem to be anything in the binder to clean the subscriptions up when the application shuts down. Should we put in some mechanism for that?

spring-cloud-gcp-pubsub-stream-binder/src/main/java/org/springframework/cloud/gcp/stream/binder/pubsub/provisioning/PubSubChannelProvisioner.java
```diff
@@ -65,7 +71,7 @@ public ConsumerDestination provisionConsumerDestination(String name, String grou
 				this.pubSubAdmin.createSubscription(subscription, name);
```
dzou: minor nit: seems like the tabs are off in this block

spring-cloud-gcp-pubsub-stream-binder/src/main/java/org/springframework/cloud/gcp/stream/binder/pubsub/provisioning/PubSubChannelProvisioner.java
artembilan: I think so. When consumer with such an anonymous group is shouting down we definitely should remove its subscription.

Good catch!

Let's address it in the separate issue though!?

spring-cloud-gcp-pubsub-stream-binder/src/main/java/org/springframework/cloud/gcp/stream/binder/pubsub/provisioning/PubSubChannelProvisioner.java
```diff
@@ -54,8 +58,10 @@ public ProducerDestination provisionProducerDestination(String name,
 	public ConsumerDestination provisionConsumerDestination(String name, String group,
 			ExtendedConsumerProperties<PubSubConsumerProperties> properties)
 			throws ProvisioningException {
+		// Use group name as subscription name
+		// Generate anonymous random group, if one not provided
+		String subscription = !StringUtils.hasText(group) ? "anonymous." + UUID.randomUUID().toString() : group;
```
dzou: Just to clarify on the feature request, is there to be exactly 1 anonymous subscription for each topic that every subscriber of the anonymous group will use? Or will there be multiple anonymous subscriptions for each topic?

If only 1 anonymous subscription per topic, one option is you could not worry about deleting it and just leave it there; in subsequent application runs, you could check if the anonymous subscription already exists for the topic and reuse it if present or create if it does not exist.

artembilan: I would like to include a `name` into the `anonymous` subscriptions.
This way it is going to be easier to tracking them on the GCP or some other place.

spring-cloud-gcp-pubsub-stream-binder/src/test/java/org/springframework/cloud/gcp/stream/binder/pubsub/provisioning/PubSubChannelProvisionerTests.java
```diff
@@ -0,0 +1,87 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.eq;
+import static org.mockito.ArgumentMatchers.matches;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+/**
+ * @author Mike Eltsufin
```
artembilan: This is new class, so `@since 1.1`

spring-cloud-gcp-pubsub-stream-binder/src/main/java/org/springframework/cloud/gcp/stream/binder/pubsub/provisioning/PubSubChannelProvisioner.java
artembilan: I believe anonymous subscriptions have to be volatile and die together with their subscriber.

spring-cloud-gcp-pubsub-stream-binder/src/main/java/org/springframework/cloud/gcp/stream/binder/pubsub/provisioning/PubSubChannelProvisioner.java
dzou: Ah nevermind, some of the code was compressed by the UI, I missed that there were lines between `if (this.pubSubAdmin.getTopic(name) == null) {` and `this.pubSubAdmin.createSubscription(subscription, name);`

spring-cloud-gcp-pubsub-stream-binder/src/main/java/org/springframework/cloud/gcp/stream/binder/pubsub/provisioning/PubSubChannelProvisioner.java
```diff
@@ -54,8 +58,12 @@ public ProducerDestination provisionProducerDestination(String name,
 	public ConsumerDestination provisionConsumerDestination(String name, String group,
 			ExtendedConsumerProperties<PubSubConsumerProperties> properties)
 			throws ProvisioningException {
+		// Use group name as subscription name
+		// Generate anonymous random group, if one not provided
+		String subscription = !StringUtils.hasText(group) ?
+				"anonymous." + name + "." + UUID.randomUUID().toString()
```
artembilan: Why exactly not start from the topic name first and only then an anonymous suffix?

Just wonder if that would better: I can live with this as well!

meltsufin: I thought that if a topic name itself ends with ".anonymous", this can cause ambiguity for parsing, if one ever needs to do it.

meltsufin:
@artembilan Can you please review the doc updates? I'm not very experienced with the binder. So, I hope I got everything right there.

spring-cloud-gcp-pubsub-stream-binder/src/main/java/org/springframework/cloud/gcp/stream/binder/pubsub/provisioning/PubSubChannelProvisioner.java
artembilan: OK. Makes sense.

Closing conversation...

spring-cloud-gcp-docs/src/main/asciidoc/spring-stream.adoc
```diff
@@ -29,25 +29,45 @@ A https://github.com/spring-cloud/spring-cloud-gcp/tree/master/spring-cloud-gcp-
...
 
-You can configure the Spring Cloud Stream Binder for Google Cloud Pub/Sub to automatically generate the underlying resources, like the Google Cloud Pub/Sub subscriptions for the consumers.
-For that, you can use the `spring.cloud.stream.gcp.pubsub.bindings.[CHANNEL-NAME].consumer.auto-create-resources` property, which is turned ON by default.
+You can configure the Spring Cloud Stream Binder for Google Cloud Pub/Sub to automatically generate the underlying resources, like the Google Cloud Pub/Sub topics and subscriptions for producers and consumers.
+For that, you can use the `spring.cloud.stream.gcp.pubsub.bindings.<channelName>.<consumer|producer>.auto-create-resources` property, which is turned ON by default.
 
-If automatic resource creation is turned ON and the subscription and the topic do not exist for a consumer, a subscription and a topic will be created with the same name.
-For example, for the following configuration, a topic and a subscription called `myConsumer` would be created.
+Starting with version 1.1, these and other binder properties can be configured globally for all the bindings, e.g. `spring.cloud.stream.gcp.pubsub.default.consumer.auto-create-resources`.
```
artembilan: The Docs are great, although this option makes me think that with an anonymous group we just should ignore it and always create a subscription.

artembilan:
Any thoughts about failed Travis?

spring-cloud-gcp-docs/src/main/asciidoc/spring-stream.adoc
meltsufin: I think you're fight, because in that scenario, there is no way the user can have the subscription already created.

meltsufin:
@artembilan The test failure caught a real problem. :)

spring-cloud-gcp-docs/src/main/asciidoc/spring-stream.adoc
meltsufin: Created #1180

spring-cloud-gcp-pubsub-stream-binder/src/main/java/org/springframework/cloud/gcp/stream/binder/pubsub/provisioning/PubSubChannelProvisioner.java
```diff
@@ -62,7 +62,7 @@ public ConsumerDestination provisionConsumerDestination(String name, String grou
 		// Generate anonymous random group, if one not provided
 		String subscription = !StringUtils.hasText(group) ?
 				"anonymous." + name + "." + UUID.randomUUID().toString()
-				: group;
+				: name + "." + group;
```
artembilan: I think this is not good. I don't believe that you can make end-users to always create subscriptions with this pattern.
We only have to say that `consumerGroup` has to be supplied with the subscription name.
More over this your pattern is going to be included in just general `group` solution.
Otherwise you can't support any arbitrary names and your provisioner is going to create extra subscription when end-user would like to rely on existing one.

Does it make sense to you?

meltsufin: I can see that point of view too, but `AbstractBinderTests.testSendAndReceiveMultipleTopics` will fail without this because it uses the same consumer group name with multiple topics.

artembilan: Huh? So, with this nasty hack you are going to meet wrong test requirements? 😕 

Why can't you specify consumer group properties there with the desired `[TOPIC].[SUBSCRIPTION]` pattern?

meltsufin: Haha, it's not my test - `org.springframework.cloud.stream.binder.AbstractBinderTests.testSendAndReceiveMultipleTopics`. I looked at the test as the binder contract definition. The test implies that consumer group names are scoped to a destination, which we can only support by adding the prefix to the subscription name. If that's in fact, not the contract, the test in `spring-cloud-stream-binder-test` should be fixed. I can ignore that particular test for now.

meltsufin: OK sounds good.
In Pub/Sub subscription names must be unique to a project. I'm not sure why the names aren't scoped to a topic, but that's just how it is right now. @kir-titievsky?

artembilan: Hm. OK. So, this is a specific of Pub/Sub and we need take it into account, but still not fix the core because end-users aware of such a global perspective and just fix the test in really scope of this project and not SCSt core.

PR #1179, opened 2018-11-05
Datastore refdoc cleanup and update

### LINKED ISSUE #1175: Datastore Ref doc update to all current features

ChengyuanZhao:
fixes #1175

Updates refdoc to include all features we have implemented.

### DIFF
--- a/spring-cloud-gcp-docs/src/main/asciidoc/datastore.adoc
+++ b/spring-cloud-gcp-docs/src/main/asciidoc/datastore.adoc
@@ -565,7 +565,21 @@ The Template API provides convenience methods for:
 
 ==== GQL Query
 
-This feature is not supported yet.
+In addition to retrieving entities by their IDs, you can also submit queries.
+
+[source,java]
+----
+  <T> Iterable<T> query(Query<? extends BaseEntity> query, Class<T> entityClass);
+
+  <A, T> Iterable<T> query(Query<A> query, Function<A, T> entityFunc);
+
+  Iterable<Key> queryKeys(Query<Key> query);
+----
+
+These methods, respectively, allow querying for:
+* entities mapped by a given entity class using all the same mapping and converting features
+* arbitrary types produced by a given mapping function
+* only the Cloud Datastore keys of the entities found by the query
 
 ==== Find by ID(s)
 
@@ -603,13 +617,10 @@ public class ExampleItem {
 }
 ----
 
-===== Read with offsets and limits
+===== Read with offsets, limits, and sorting
 
-This feature is not supported yet.
-
-===== Sorting
-
-This feature is not supported yet.
+`DatastoreRepository` and custom defined entitiy repositories implement Spring Data's `PagingAndSortingRepository`, which supports offsets and limits using page numbers and page sizes.
+Paging and sorting options are also supported in `DatastoreTemplate` by supplying a `DatastoreQueryOptions`.
 
 ===== Partial read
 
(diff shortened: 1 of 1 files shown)


ChengyuanZhao:
@meltsufin @dmitry-s PTAL

spring-cloud-gcp-docs/src/main/asciidoc/datastore.adoc
```diff
@@ -860,7 +871,35 @@ public interface TraderRepository extends CrudRepository<Trader, String> {
...
+
+	List<TradeProjection> findByAction(String action);
+
+	@Query("SELECT action, symbol FROM trades WHERE action = @action")
+	List<TradeProjection> findByQuery(String action);
+}
+----
+
+Projections can be provided by name-convention-based query methods as well as by custom SQL queries.
+If using custom SQL queries, you can further restrict the fields retrieved from Datastore to just those required by the projection.
```
dmitry-s: Mention that the whole Datastore Entity is retrieved in case of queries based on method name.

spring-cloud-gcp-docs/src/main/asciidoc/datastore.adoc
```diff
@@ -603,13 +617,10 @@ public class ExampleItem {
...
 
-===== Read with offsets and limits
-
-This feature is not supported yet.
+===== Read with offsets, limits, and sorting
 
-===== Sorting
-
-This feature is not supported yet.
+`DatastoreRepository` and custom defined entitiy repositories implement Spring Data's `PagingAndSortingRepository`, which supports offsets and limits using page numbers and page sizes.
```
```suggestion
`DatastoreRepository` and custom-defined entitiy repositories implement the Spring Data `PagingAndSortingRepository`, which supports offsets and limits using page numbers and page sizes.
```

spring-cloud-gcp-docs/src/main/asciidoc/datastore.adoc
```diff
@@ -603,13 +617,10 @@ public class ExampleItem {
...
+Paging and sorting options are also supported in `DatastoreTemplate` by supplying a `DatastoreQueryOptions` to `findAll`.
```
```suggestion
Paging and sorting options are also supported in the `DatastoreTemplate` by supplying a `DatastoreQueryOptions` to `findAll`.
```

spring-cloud-gcp-docs/src/main/asciidoc/datastore.adoc
```diff
@@ -765,6 +776,7 @@ Cloud Datastore only supports filter components joined by AND, and the following
 * `is null`
 
 After writing a custom repository interface specifying just the signatures of these methods, implementations are generated for you and can be used with an auto-wired instance of the repository.
+Because of Cloud Datastore's requirement that explicitly selected fields must all appear in a composite index together, `find` name-based query methods are run as `SELECT *`.
```
```suggestion
Cloud Datastore requires that explicitly selected fields must all appear in a composite index together. To avoid this restriction, name-based query `find` methods are run as `SELECT *`.
```

spring-cloud-gcp-docs/src/main/asciidoc/datastore.adoc
```diff
@@ -860,7 +872,35 @@ public interface TraderRepository extends CrudRepository<Trader, String> {
...
+
+	List<TradeProjection> findByAction(String action);
+
+	@Query("SELECT action, symbol FROM trades WHERE action = @action")
+	List<TradeProjection> findByQuery(String action);
+}
+----
+
+Projections can be provided by name-convention-based query methods as well as by custom GQL queries.
+If using custom GQL queries, you can further restrict the fields retrieved from Datastore to just those required by the projection.
```
```suggestion
If using custom GQL queries, you can further restrict the fields retrieved from Cloud Datastore to just those required by the projection.
```

spring-cloud-gcp-docs/src/main/asciidoc/datastore.adoc
```diff
@@ -860,7 +872,35 @@ public interface TraderRepository extends CrudRepository<Trader, String> {
...
+However, custom select statements (those not using `SELECT *`) require composite indexes containing the selected fields.
+
+Properties of projection types defined using SpEL use the fixed name `target` for the underlying domain object.
+As a result accessing underlying properties take the form `target.<property-name>`.
```
```suggestion
As a result, accessing underlying properties take the form `target.<property-name>`.
```

spring-cloud-gcp-docs/src/main/asciidoc/datastore.adoc
ChengyuanZhao: Hmmm I'm not sure this is better.

spring-cloud-gcp-docs/src/main/asciidoc/datastore.adoc
meltsufin: I always think of the class, not an instance when referencing. "The DatastoreTemplate [class]" sounds ok to me.

ChengyuanZhao: However "class" isn't here, and I don't think it's necessary to say that it is a class. 

I've been most comfortable and I think it's most natural to refer to classes without "the". For example when we talk about `Object`, `String`, or `Long`, we never say "the String class".

PR #1181, opened 2018-11-06
Add Trace Sample Test

dzou:
Add a integration test for Trace Sample.

In reference to #705.


spring-cloud-gcp-samples/spring-cloud-gcp-trace-sample/src/test/java/com/example/ApplicationTests.java
```diff
@@ -0,0 +1,149 @@
...
+		assumeThat(
+				"Google Cloud Trace integration tests are disabled. "
+						+ "Please use '-Dit.trace=true' to enable them. ",
+				System.getProperty("it.trace"), is("true"));
+	}
+
+	@Before
+	public void setupTraceClient() throws IOException {
+		this.url = String.format("http://localhost:%d/", this.port);
+		this.testRestTemplate = new TestRestTemplate();
```
artembilan: One more time: you don't need to instantiate this object and do some hard formatting logic for the port.
There is just enough to `@Autowired` this object from the application context.
See `ApplicationTests` in the `spring-cloud-gcp-data-datastore-sample`.

meltsufin: That's how we had it before, but that `TestRestTemplate` was getting instrumented by Sleuth and interfered with the test.
@dzou, maybe add a comment above this line explaining this.

dzou: No, this is a special case with the Trace sample test. Due to the configuration of the sample app, the autowired instance of TestRestTemplate has its own built-in Trace instrumentation and will attach a random `xb-3-traceid` header automatically to all requests. The traces on Stackdriver will use this ID. However, in the test we want to set the traceid manually with `x-cloud-trace-context` so we can use the value to look it up and verify it on StackDriver.

So here I have to create a new rest template instance which does not have the configuration that automatically adds that xb-3 header.

artembilan: Uh! So, sounds like you want to overcome (mock ?) the real functionality with your own code.
Why just not test whatever is provided by the tracer automatically?

How it can be an integration test if we don't use real functionality?

Maybe we can just add some interceptor into the autowired `TestRestTemplate` to catch that generated header and then verify against it in the end of test?

Thanks for understanding

dzou: Hmm, so the only functionality we are changing is overriding the traceId set in the request with a non-random one. That lets us know the traceId value beforehand so we can check it was successfully registered in StackDriver.

So I think it is ok here for the integration test since we are not modifying the core logic of what is going on. If there were some way of determining the random ID generated for the requests by the rest template beforehand, we could eliminate doing this. This was the only way that Mike and I could figure out how to make the traceId known by the test.

artembilan: OK. Fair enough. No reason to argue anymore.

Please, add an in-line comment here to explain "Why?" and it is going to be good for merging!

dzou: Sounds good, thanks.

PR #1190, opened 2018-11-07
Enable Google Cloud IAP authentication through spring security

### LINKED ISSUE #1016: FR: Spring Security Autoconfiguration to validate IAP header and extract principle

elefeint:
This PR changes/creates 3 projects.

**1. spring-cloud-gcp-autoconfigure.**
New package o.s.c.g.a.security  containing customizations to autoconfigured OAuth Resource Server.

Uses Spring Security OAuth2 Resource Server to parse the IAP header, validate its signature and claims, and provide an authorization object.

There is no additional logic, but two IAP-specific beans are provided through autoconfiguration:
* A version of JwtDecoder that goes against IAP public key and validates the private key using ES256 algorithm. The first part could have been configured with a property, but encryption is not yet property-configurable. When spring-projects/spring-boot#15115 is resolved, this bean can be removed and replaced with the two properties.
* A BearerTokenResolver that parses `x-goog-iap-jwt-assertion `header instead of the standard OAuth `access_token`.

There is one unrelated change in this PR -- bringing spring-security as a transitive dependency made GcpStorageAutoConfigurationTests needlessly secure, so I excluded security autoconfiguration from the test.

**2. spring-cloud-gcp-starters/spring-cloud-gcp-starter-security-iap**
A new project that brings in autoconfiguration and a dependency on `spring-boot-starter-oauth2-resource-server.` 

**3. spring-cloud-gcp-samples/spring-cloud-gcp-security-iap-sample**
A new project containing a minimal demo app showing how to configure a secured URL with a custom `WebSecurityConfigurerAdapter`.

Fixes #1016.

### DIFF
--- a/spring-cloud-gcp-autoconfigure/pom.xml
+++ b/spring-cloud-gcp-autoconfigure/pom.xml
@@ -193,6 +193,13 @@
             <optional>true</optional>
         </dependency>
 
+        <!-- Security -->
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-oauth2-resource-server</artifactId>
+            <optional>true</optional>
+        </dependency>
+
         <!-- Test -->
         <dependency>
             <groupId>commons-io</groupId>
--- a/spring-cloud-gcp-autoconfigure/pom.xml
+++ b/spring-cloud-gcp-autoconfigure/pom.xml
@@ -195,8 +195,20 @@
 
         <!-- Security -->
         <dependency>
-            <groupId>org.springframework.boot</groupId>
-            <artifactId>spring-boot-starter-oauth2-resource-server</artifactId>
+            <groupId>org.springframework.security</groupId>
+            <artifactId>spring-security-config</artifactId>
+            <optional>true</optional>
+        </dependency>
+
+        <dependency>
+            <groupId>org.springframework.security</groupId>
+            <artifactId>spring-security-oauth2-jose</artifactId>
+            <optional>true</optional>
+        </dependency>
+
+        <dependency>
+            <groupId>org.springframework.security</groupId>
+            <artifactId>spring-security-oauth2-resource-server</artifactId>
             <optional>true</optional>
         </dependency>
 
--- a/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/security/IapAuthenticationAutoConfiguration.java
+++ b/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/security/IapAuthenticationAutoConfiguration.java
@@ -1,0 +1,59 @@
+/*
+ *  Copyright 2018 original author or authors.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package org.springframework.cloud.gcp.autoconfigure.security;
+
+import org.springframework.boot.autoconfigure.AutoConfigureBefore;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.boot.autoconfigure.security.oauth2.resource.servlet.OAuth2ResourceServerAutoConfiguration;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.security.oauth2.jwt.JwtDecoder;
+import org.springframework.security.oauth2.jwt.NimbusJwtDecoderJwkSupport;
+import org.springframework.security.oauth2.server.resource.web.BearerTokenResolver;
+
+/**
+ * Configures GCP IAP pre-authentication.
+ * If the property is missing, this auto configuration is skipped.
+ */
+@Configuration
+@ConditionalOnProperty(value = "spring.cloud.gcp.security.iap.enabled", matchIfMissing = false)
+@ConditionalOnClass({NimbusJwtDecoderJwkSupport.class})
+@AutoConfigureBefore(OAuth2ResourceServerAutoConfiguration.class)
+public class IapAuthenticationAutoConfiguration {
+
+	// TODO: externalize as properties.
+	private static final String PUBLIC_KEY_VERIFICATION_LINK = "https://www.gstatic.com/iap/verify/public_key-jwk";
+
+	private static final String ENCRYPTION_ALGORITHM = "ES256";
+
+	private static final String IAP_HEADER = "x-goog-iap-jwt-assertion";
+
+
+	@Bean
+	@ConditionalOnMissingBean
+	public JwtDecoder iapJwtDecoder() {
+		return new NimbusJwtDecoderJwkSupport(PUBLIC_KEY_VERIFICATION_LINK, ENCRYPTION_ALGORITHM);
+	}
+
+	@Bean
+	@ConditionalOnMissingBean
+	public BearerTokenResolver iatTokenResolver() {
+		return r -> r.getHeader(IAP_HEADER);
+	}
+}
--- a/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/security/IapAuthenticationAutoConfiguration.java
+++ b/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/security/IapAuthenticationAutoConfiguration.java
@@ -16,6 +16,8 @@
 
 package org.springframework.cloud.gcp.autoconfigure.security;
 
+import com.google.common.annotations.VisibleForTesting;
+
 import org.springframework.boot.autoconfigure.AutoConfigureBefore;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
@@ -54,12 +56,14 @@ import org.springframework.security.oauth2.server.resource.web.BearerTokenResolv
 @AutoConfigureBefore(OAuth2ResourceServerAutoConfiguration.class)
 public class IapAuthenticationAutoConfiguration {
 
-	private static final String PUBLIC_KEY_VERIFICATION_LINK = "https://www.gstatic.com/iap/verify/public_key-jwk";
+	@VisibleForTesting
+	static final String PUBLIC_KEY_VERIFICATION_LINK = "https://www.gstatic.com/iap/verify/public_key-jwk";
 
-	private static final String ENCRYPTION_ALGORITHM = "ES256";
+	@VisibleForTesting
+	static final String ENCRYPTION_ALGORITHM = "ES256";
 
-	private static final String IAP_HEADER = "x-goog-iap-jwt-assertion";
-
+	@VisibleForTesting
+	static final String IAP_HEADER = "x-goog-iap-jwt-assertion";
 
 	@Bean
 	@ConditionalOnMissingBean
--- a/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/security/IapAuthenticationAutoConfiguration.java
+++ b/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/security/IapAuthenticationAutoConfiguration.java
@@ -54,7 +54,6 @@ import org.springframework.security.oauth2.server.resource.web.BearerTokenResolv
 @AutoConfigureBefore(OAuth2ResourceServerAutoConfiguration.class)
 public class IapAuthenticationAutoConfiguration {
 
-	// TODO: externalize as properties.
 	private static final String PUBLIC_KEY_VERIFICATION_LINK = "https://www.gstatic.com/iap/verify/public_key-jwk";
 
 	private static final String ENCRYPTION_ALGORITHM = "ES256";
--- a/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/security/IapAuthenticationAutoConfiguration.java
+++ b/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/security/IapAuthenticationAutoConfiguration.java
@@ -16,8 +16,6 @@
 
 package org.springframework.cloud.gcp.autoconfigure.security;
 
-import com.google.common.annotations.VisibleForTesting;
-
 import org.springframework.boot.autoconfigure.AutoConfigureBefore;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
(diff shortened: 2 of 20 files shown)


artembilan:
Travis report:
```
[ERROR] Failed to execute goal on project spring-cloud-gcp-security-iap-sample: Could not resolve dependencies for project com.example:spring-cloud-gcp-security-iap-sample:jar:1.1.0.BUILD-SNAPSHOT: Could not find artifact org.springframework.cloud:spring-cloud-gcp-starter-security-iap:jar:1.1.0.BUILD-SNAPSHOT in sonatype-snapshots (https://oss.sonatype.org/content/repositories/snapshots/)
```
Missed something in poms? As a module for example?

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/security/IapAuthenticationAutoConfiguration.java
```diff
@@ -0,0 +1,76 @@
...
+ * <p>If a custom {@link WebSecurityConfigurerAdapter} is present, it must add {@code .oauth2ResourceServer().jwt()}
+ * customization to {@link org.springframework.security.config.annotation.web.builders.HttpSecurity} object. If no
+ * custom {@link WebSecurityConfigurerAdapter} is found,
+ * Spring Boot's default {@code OAuth2ResourceServerWebSecurityConfiguration} will add this customization.
+ *
+ * @author Elena Felder
+ * @since 1.1
+ */
+@Configuration
+@ConditionalOnProperty("spring.cloud.gcp.security.iap.enabled")
```
artembilan: Why no `matchIfMissing = true`  as we have in all other auto-configuration in this project?

This property also has to be added to the `additional-spring-configuration-metadata.json`, if you are not going to extract a fully blown `@ConfigurationProperties` class, as you mention with TODO below.

spring-cloud-gcp-samples/spring-cloud-gcp-security-iap-sample/README.adoc
```diff
@@ -0,0 +1,7 @@
+TODO
```
artembilan: TODO?

spring-cloud-gcp-samples/spring-cloud-gcp-security-iap-sample/README.adoc
```diff
@@ -0,0 +1,7 @@
...
+
+Note: this sample should be tested when deployed to AppEngine Flexible or Kubernetes environments.
+
+It is possible, in principle, to grab a recent JWK token from a deployed apps' /headers path, and to test locally, but this token will expire, causing authentication to be declined locally.
```
artembilan: I believe `/headers` should be wrapped to the code.

spring-cloud-gcp-samples/spring-cloud-gcp-security-iap-sample/README.adoc
```diff
@@ -0,0 +1,7 @@
...
+
+If you don't create your own implementation of WebSecurityConfigurerAdapter, Spring Boot will introduce its own, OAuth2ResourceServerWebSecurityConfiguration
```
artembilan: The classes are definitely code. so, please, proper wrapping.
Also every sentence should finish with period.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/security/IapAuthenticationAutoConfiguration.java
```diff
@@ -54,13 +56,14 @@
 @AutoConfigureBefore(OAuth2ResourceServerAutoConfiguration.class)
 public class IapAuthenticationAutoConfiguration {
 
-	// TODO: externalize as properties.
-	private static final String PUBLIC_KEY_VERIFICATION_LINK = "https://www.gstatic.com/iap/verify/public_key-jwk";
-
-	private static final String ENCRYPTION_ALGORITHM = "ES256";
+	@VisibleForTesting
```
artembilan: What is the problem to make all these constants `public`?
to be honest I'm fully against such a `@VisibleForTesting` hack in the production code.
This is not Spring way...

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/security/IapAuthenticationAutoConfiguration.java
artembilan: I don't see that you have addressed this one.

Do you see my review at all?

Thanks

elefeint:
@artembilan Yep, I've seen it, thank you! I haven't finished addressing the feedback yet. I pushed code because I wanted to see how codecov would react to tests.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/security/IapAuthenticationAutoConfiguration.java
elefeint: So, this one is interesting. It's true that our normal practice is to have` matchIfMissing = true`, but our normal practice is to, at the same time, add @ConditionalOnClass that will invalidate the autoconfig whenever a required dependency is missing. IAP, however, no longer has any unique required dependencies -- the only dependency it needs is `spring-boot-starter-oauth2-resource-server`, which could be present on the classpath for unrelated reasons (for example, an application uses OAuth and also brings in `spring-cloud-gcp-autoconfigure` with a datastore starter).

I considered adding matchIfMissing and having the starter turn on the property in its application.properties, but that does not seem to be done anywhere else in the project, either. Do you think this is a good alternative?

artembilan: OK. Fine with me.

Since we really don't have a strict dependency and you are definitely right with some side-effect, so let it it as `false` by default!

Then we need explain in the Docs that this feature must be turned on explicitly.
On the other hand is there a choice when we really know that we are in GCP, so that is fully OK to have this IAP enabled by default? Are we really going to use some other JWT mechanism there ?

Thanks for explanation!

elefeint: I could see someone hosting an application on GCP but using Facebook or Twitter for authentication. IAP is more of an enterprise authN/authZ mechanism, since someone needs to set up users and groups for it to be useful.

I am writing documentation now; apparently that takes me even longer than writing code :\

spring-cloud-gcp-docs/src/main/asciidoc/security-iap.adoc
```diff
@@ -0,0 +1,41 @@
+:spring-security-ref: https://docs.spring.io/spring-security/site/docs/current/reference/htmlsingle/
+:spring-security-javadoc: https://docs.spring.io/spring-security/site/docs/current/api/org/springframework/security/
+
+== Google Cloud IAP Authentication
+
+https://cloud.google.com/iap/[Cloud Identity-Aware Proxy (IAP)] provides a security layer over applications deployed to Google Cloud.
+
+The IAP starter uses
```
artembilan: "One sentence per line": https://asciidoctor.org/docs/asciidoc-recommended-practices/#one-sentence-per-line

spring-cloud-gcp-samples/spring-cloud-gcp-security-iap-sample/README.adoc
elefeint: I've been retelling this story in English for years. Finally, I can cite it "in the original Klingon" (well, except when I first encountered it, the difficulty was with getting a painting past government censors, and not a strict editor): https://koppel.pro/diary/sobacka-favorskogo-2443

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/security/IapAuthenticationAutoConfiguration.java
elefeint: Makes sense. It's a cultural difference; I am still adjusting -- in a giant monorepo, a convention to restrict visibility, unless there is a good reason to relax it, is reasonable. But in Spring, where every project's boundaries are clearly delineated, it makes sense to do the opposite.

spring-cloud-gcp-autoconfigure/pom.xml
```diff
@@ -193,6 +193,13 @@
             <optional>true</optional>
         </dependency>
 
+        <!-- Security -->
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-oauth2-resource-server</artifactId>
```
meltsufin: It's odd to see a starter imported here. This module is just a pom. Can we be more specific about which modules we need here? Is it `spring-security-oauth2-resource-server` and something else?

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/security/IapAuthenticationAutoConfiguration.java
```diff
@@ -0,0 +1,74 @@
...
+ * Spring Boot's default {@code OAuth2ResourceServerWebSecurityConfiguration} will add this customization.
+ *
+ * @author Elena Felder
+ * @since 1.1
+ */
+@Configuration
+@ConditionalOnProperty("spring.cloud.gcp.security.iap.enabled")
+@ConditionalOnClass({NimbusJwtDecoderJwkSupport.class})
+@AutoConfigureBefore(OAuth2ResourceServerAutoConfiguration.class)
+public class IapAuthenticationAutoConfiguration {
```
meltsufin: Does it make sense to add and enable `IapAuthenticationProperties`?
It could expose the configuration you currently have via constants below, and the "enabled" property.

spring-cloud-gcp-docs/src/main/asciidoc/security-iap.adoc
```diff
@@ -0,0 +1,40 @@
+:spring-security-ref: https://docs.spring.io/spring-security/site/docs/current/reference/htmlsingle/
+:spring-security-javadoc: https://docs.spring.io/spring-security/site/docs/current/api/org/springframework/security/
+
+== Google Cloud IAP Authentication
+
+https://cloud.google.com/iap/[Cloud Identity-Aware Proxy (IAP)] provides a security layer over applications deployed to Google Cloud.
+
+The IAP starter uses {spring-security-ref}#oauth2resourceserver[Spring Security OAuth 2.0 Resource Server] functionality to automatically extract user identity from the proxy-injected `x-goog-iap-jwt-assertion` header.
```
meltsufin:
```suggestion
The IAP starter uses {spring-security-ref}#oauth2resourceserver[Spring Security OAuth 2.0 Resource Server] functionality to automatically extract user identity from the proxy-injected `x-goog-iap-jwt-assertion` HTTP header.
```

spring-cloud-gcp-docs/src/main/asciidoc/security-iap.adoc
```diff
@@ -0,0 +1,40 @@
...
+
+To use IAP authentication, you need to:
```
meltsufin: I think you have to use the abbreviation "Cloud IAP" everywhere. We've been dinged for this before when we tried to refer to Cloud Spanner as just "Spanner".
Notice the abbreviation use in [official docs](https://cloud.google.com/iap/).
```suggestion
To use Cloud IAP authentication, you need to:
```

spring-cloud-gcp-starters/spring-cloud-gcp-starter-security-iap/pom.xml
```diff
@@ -0,0 +1,27 @@
...
+
+    <dependencies>
+        <dependency>
+            <groupId>org.springframework.cloud</groupId>
+            <artifactId>spring-cloud-gcp-starter</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-oauth2-resource-server</artifactId>
```
meltsufin: See my comment in the autoconfigure module. This would match whatever we have there as optional deps for this.

spring-cloud-gcp-autoconfigure/pom.xml
artembilan: Ah! Right. We must not use starters in the `autocofigure` module as dependency.
I don't remember "Why?", but that is what @snicoll recommended us to do.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/security/IapAuthenticationAutoConfiguration.java
artembilan: We have an issue on the matter:  https://github.com/spring-cloud/spring-cloud-gcp/issues/1192

spring-cloud-gcp-starters/spring-cloud-gcp-starter-security-iap/pom.xml
artembilan: I think refer to starters from our own starters is OK. 
See `spring-cloud-gcp-starter-sql-mysql` and `spring-cloud-gcp-starter-trace`.
Plus we can borrow some pattern from Spring Boot per se: https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-starters/spring-boot-starter-websocket/pom.xml

spring-cloud-gcp-starters/spring-cloud-gcp-starter-security-iap/pom.xml
```diff
@@ -20,8 +20,16 @@
         </dependency>
 
         <dependency>
-            <groupId>org.springframework.boot</groupId>
-            <artifactId>spring-boot-starter-oauth2-resource-server</artifactId>
+            <groupId>org.springframework.security</groupId>
```
artembilan: Huh? We have just discussed that it is OK to use starters as dependencies in starters.

elefeint: Might as well be consistent with autoconfig. And, I guess, for someone trying to troubleshoot why the starter works but their application does not, it would be nice to see all the dependencies together and not have to do a recursive hunt for them.

artembilan: OK. Let it be. I just wanted to follow the practice from Spring Boot.
Won't hart to have ti like this...

elefeint:
@meltsufin 
About GCP-specific claims -- we discussed whether to enforce the container-specific (AppEngine Flex vs Kubernetes) audience claim a couple of weeks ago, and the conclusion was that there was not a ton of value, given that the verification would need to be different depending on where the app is deployed. But we can reevaluate -- building in custom validators would be as simple as setting them while creating `NimbusJwtDecoderJwkSupport` in `IapAuthenticationAutoConfiguration`. If we choose to do it, I would do it in a different PR.

As far as USER_ROLE, this was necessitated by the specific details source (`PreAuthenticatedGrantedAuthoritiesUserDetailsService`) I chose in the previous implementation. Spring security resource server implementation uses a standard `WebAuthenticationDetailsSource` which does not require authorities claims.

wyaeld:
maybe not right place to ask, but @elefeint @meltsufin are you able to share when next milestone (or RC) which might include this will be?  Reason I ask is I'm leading an effort to get apps up on GCP and IAP is one of the very next things on my roadmap.  This repo isn't using the Milestone functionality much so its hard to know what the 1.1.0 GA and M3 roadmap looks like.

meltsufin:
@wyaeld We are following the Spring Cloud Greenwich release train now. So, [this is the tentative schedule](https://github.com/spring-cloud/spring-cloud-release/milestones) that would correspond to our 1.1.0 releases. Next milestone might be tomorrow if all goes well.

Issue #1193, opened 2018-11-08
Create a Spring Boot Starter and Template API for Cloud Vision


dzou:
Create a Spring Boot Starter and Template API for Cloud Vision.

meltsufin:
We can probably split this one into two, and the current PR would resolve one of them. 😉

PR #1194, opened 2018-11-08
Add Basic Starter AutoConfiguration for Cloud Vision API

dzou:
This adds the basic starter for the Cloud vision APIs.

In reference to #1193.

### DIFF
--- a/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/vision/CloudVisionAutoConfiguration.java
+++ b/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/vision/CloudVisionAutoConfiguration.java
@@ -1,0 +1,74 @@
+/*
+ *  Copyright 2018 original author or authors.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package org.springframework.cloud.gcp.autoconfigure.vision;
+
+import java.io.IOException;
+
+import com.google.api.gax.core.CredentialsProvider;
+import com.google.cloud.vision.v1.ImageAnnotatorClient;
+import com.google.cloud.vision.v1.ImageAnnotatorSettings;
+
+import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+/**
+ * Provides Spring Beans for using Cloud Vision API.
+ *
+ * @author Daniel Zou
+ */
+@Configuration
+@EnableConfigurationProperties(CloudVisionProperties.class)
+@ConditionalOnProperty(value = "spring.cloud.gcp.vision.enabled", matchIfMissing = true)
+public class CloudVisionAutoConfiguration {
+
+	private final CredentialsProvider credentialsProvider;
+
+	public CloudVisionAutoConfiguration(CredentialsProvider credentialsProvider) {
+		this.credentialsProvider = credentialsProvider;
+	}
+
+	/**
+	 * Configure the Cloud Vision API client {@link ImageAnnotatorClient}. The
+	 * spring-cloud-gcp-starter autowires a {@link CredentialsProvider} object that provides
+	 * the GCP credentials, required to authenticate and authorize Vision API calls.
+	 *
+	 * <p>
+	 * Cloud Vision API client implements {@link AutoCloseable}, which is automatically
+	 * honored by Spring bean lifecycle.
+	 *
+	 * <p>
+	 * Most of the Google Cloud API clients are thread-safe heavy objects. I.e., it's better
+	 * to produce a singleton and re-using the client object for multiple requests.
+	 *
+	 * @param credentialsProvider GCP Credential to use to access Cloud Vision API
+	 * @return A Cloud Vision API client
+	 * @throws IOException if an exception occurs creating the ImageAnnotatorClient
+	 */
+	@Bean
+	@ConditionalOnMissingBean
+	public ImageAnnotatorClient imageAnnotatorClient(
+			CredentialsProvider credentialsProvider) throws IOException {
+		ImageAnnotatorSettings clientSettings = ImageAnnotatorSettings.newBuilder()
+				.setCredentialsProvider(credentialsProvider)
+				.build();
+
+		return ImageAnnotatorClient.create(clientSettings);
+	}
+}
--- a/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/vision/CloudVisionProperties.java
+++ b/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/vision/CloudVisionProperties.java
@@ -1,0 +1,35 @@
+/*
+ *  Copyright 2018 original author or authors.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package org.springframework.cloud.gcp.autoconfigure.vision;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.boot.context.properties.NestedConfigurationProperty;
+import org.springframework.cloud.gcp.core.Credentials;
+import org.springframework.cloud.gcp.core.CredentialsSupplier;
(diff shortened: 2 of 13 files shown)


spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/vision/CloudVisionAutoConfiguration.java
```diff
@@ -0,0 +1,72 @@
...
+	public CloudVisionAutoConfiguration(CredentialsProvider credentialsProvider) {
+		this.credentialsProvider = credentialsProvider;
+	}
+
+	/**
+	 * Configure the Cloud Vision API client {@link ImageAnnotatorClient}. The
+	 * spring-cloud-gcp-starter autowires a {@link CredentialsProvider} object that provides
+	 * the GCP credentials, required to authenticate and authorize Vision API calls.
+	 *
+	 * <p>
```
artembilan: No blank lines in method Javadocs and `<p>` has to stick to the sentence: https://github.com/spring-projects/spring-framework/wiki/Code-Style#javadoc-formatting

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/vision/CloudVisionProperties.java
```diff
@@ -0,0 +1,35 @@
...
+
+package org.springframework.cloud.gcp.autoconfigure.vision;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.boot.context.properties.NestedConfigurationProperty;
+import org.springframework.cloud.gcp.core.Credentials;
+import org.springframework.cloud.gcp.core.CredentialsSupplier;
+import org.springframework.cloud.gcp.core.GcpScope;
+
+@ConfigurationProperties("spring.cloud.gcp.vision")
```
artembilan: `@author` and `@since`

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/vision/CloudVisionProperties.java
```diff
@@ -0,0 +1,35 @@
...
+public class CloudVisionProperties implements CredentialsSupplier {
+
+	/** Overrides the GCP OAuth2 credentials specified in the Core module. */
```
artembilan: The Javadoc must be really Javadocs, so, from new line:
```suggestion
	/** 
	* Overrides the GCP OAuth2 credentials specified in the Core module. 
	*/
```

spring-cloud-gcp-autoconfigure/src/main/resources/META-INF/additional-spring-configuration-metadata.json
```diff
@@ -29,6 +29,12 @@
       "type": "java.lang.Boolean",
       "description": "Auto-configure Google Cloud Storage components.",
       "defaultValue": true
+    },
+    {
+      "name": "spring.cloud.gcp.vision.enabled",
```
artembilan: Why this property can't go to the `CloudVisionProperties` ?

spring-cloud-gcp-starters/spring-cloud-gcp-starter-trace/pom.xml
```diff
@@ -5,6 +5,8 @@
 	<parent>
 		<artifactId>spring-cloud-gcp-starters</artifactId>
 		<groupId>org.springframework.cloud</groupId>
+
+		<!-- TODO: What version do I put here? -->
```
artembilan: Is it really a concern of a purpose of this PR?

This is `parent` version BTW, we can't live without it anyway...

meltsufin: @dzou What's the concern around using the current version `1.1.0.BUILD-SNAPSHOT` as is done one all other starters?

dzou: nevermind, I got confused originally; thought I had to increment the version number to accommodate a new starter.

spring-cloud-gcp-autoconfigure/src/main/resources/META-INF/additional-spring-configuration-metadata.json
artembilan: Have you missed this one?

Thanks

dzou: You could, but putting it there won't do anything because `spring.cloud.gcp.vision.enabled` is used within a `@ConditionalOnProperty` and these [are processed before `ConfigurationProperties` are parsed](https://github.com/spring-projects/spring-boot/issues/2282). Any default value set in CloudVisionProperties will be ignored. That is why all the `.enabled` properties have been put in here.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/vision/CloudVisionAutoConfiguration.java
```diff
@@ -0,0 +1,79 @@
...
+
+	private final CredentialsProvider credentialsProvider;
+
+	public CloudVisionAutoConfiguration(
+			CloudVisionProperties properties, CredentialsProvider credentialsProvider)
+			throws IOException {
+
+		if (properties.getCredentials().hasKey()) {
+			this.credentialsProvider = new DefaultCredentialsProvider(properties);
+		} else {
```
artembilan: Checkstyle violation: 
```
[ERROR] /home/travis/build/spring-cloud/spring-cloud-gcp/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/vision/CloudVisionAutoConfiguration.java:53:9: '}' at column 3 should be alone on a line. [RightCurly]
```

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/vision/CloudVisionProperties.java
dzou: Done, I downgraded to inline comment since I felt it was not significant enough to be javadoc.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/vision/CloudVisionAutoConfiguration.java
```diff
@@ -0,0 +1,79 @@
...
+/**
+ * Provides Spring Beans for using Cloud Vision API.
+ *
+ * @author Daniel Zou
+ * @since 1.1
+ */
+@Configuration
+@EnableConfigurationProperties(CloudVisionProperties.class)
+@ConditionalOnClass(ImageAnnotatorClient.class)
+@ConditionalOnProperty(value = "spring.cloud.gcp.vision.enabled", matchIfMissing = true)
```
artembilan: @snicoll ,

Confirm, please, that we can't move such a property into the `@ConfigurationProperties` class.

See our discussion below in the `additional-spring-configuration-metadata.json` changes: https://github.com/spring-cloud/spring-cloud-gcp/pull/1194/files#r232061185

Thanks

snicoll: Yes. If you're not using the property in the API, it's basically dead code that shouldn't be there.

artembilan: OK. That is slightly different aspect though, but the question was if it really isn't going to work if we just place this property to the existing `@ConfigurationProperties` class as @dzou commented mentioning some open issue in Spring Boot: https://github.com/spring-projects/spring-boot/issues/2282

Aren't we going to have a generated `spring-configuration-metadata.json` in the META-INF based on that class anyway, which is going to be use in this `@ConditionalOnProperty`?

I'm just a bit confused that having `@ConfigurationProperties` with the prefix `spring.cloud.gcp.vision` we still need to add an entry to the `additional-spring-configuration-metadata.json` manually...

dzou:
> We can merge it if @snicoll confirms that we need to stay away of moving enable property into the @ConfigurationProperties class: https://github.com/spring-cloud/spring-cloud-gcp/pull/1194/files#r232300344

Sounds good, I'll also make the relevant doc updates before doing so.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/vision/CloudVisionAutoConfiguration.java
snicoll: I don't think that issue is related to what we're discussing here. It's about a custom condition that imports a `@ConfigurationProperties`. What we're discussing here is where the metadata should be generated. If you add an `enabled` flag in code but don't use it, the metadata will be generated automatically but you'd have effectively added dead code. I don't recommend you doing that.

I don't understand what's confusing. That property is not used in the code so it shouldn't be declared in the API. But you need to document it and that's what the additional metadata is for.

Makes sense?

artembilan: OK. Fair enough.

So, @dzou , feel free to merge.

spring-cloud-gcp-samples/spring-cloud-gcp-vision-api-sample/pom.xml
```diff
@@ -15,7 +15,8 @@
 	<dependencies>
 		<dependency>
 			<groupId>org.springframework.cloud</groupId>
-			<artifactId>spring-cloud-gcp-starter</artifactId>
+			<artifactId>spring-cloud-gcp-starter-vision</artifactId>
+			<version>1.1.0.BUILD-SNAPSHOT</version>
```
elefeint: Since it's our own starter, I think the version will be derived from the current project version.

dzou: Gotcha. However I get this problem though when I try to run the sample after: 'dependencies.dependency.version' for org.springframework.cloud:spring-cloud-gcp-starter-vision:jar is missing.

Is there something missing in my pom.xml for the starter which prevents me from omitting the version?

PR #1197, opened 2018-11-09
Prefix subscription with destination in binder

### LINKED ISSUE #1184: Pub/Sub binder should prefix subscription name with destination
Part of resolving this issue should be re-enabling the `testSendAndReceiveMultipleTopics` test in the binder. The test implies that consumer group names are scoped to a destination, which we can only support by adding the prefix to the subscription name. That prefix should be the destination name.

Follow-up to #1177

meltsufin:
* Fixes handling of multiple topics with the same consumer group name

* Re-enable testSendAndReceiveMultipleTopics

Fixes #1184.

### DIFF
--- a/spring-cloud-gcp-docs/src/main/asciidoc/spring-stream.adoc
+++ b/spring-cloud-gcp-docs/src/main/asciidoc/spring-stream.adoc
@@ -63,6 +63,8 @@ If the consumer group is not specified, an anonymous one will be created with th
 For example, for the following configuration, a topic named `myEvents` and a subscription called `myEvents.counsumerGroup1` would be created.
 If the consumer group is not specified, a subscription called `anonymous.myEvents.a6d83782-c5a3-4861-ac38-e6e2af15a7be` would be created.
 
+IMPORTANT: If you are manually creating Pub/Sub subscriptions for consumers, make sure that they follow the naming convention of `<destinationName>.<consumerGroup>`.
+
 .application.properties
 ----
 spring.cloud.stream.bindings.events.destination=myEvents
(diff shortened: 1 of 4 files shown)


spring-cloud-gcp-docs/src/main/asciidoc/spring-stream.adoc
```diff
@@ -57,10 +57,10 @@ spring.cloud.stream.gcp.pubsub.bindings.events.producer.auto-create-resources=tr
 ==== Consumer Destination Configuration
 
 If automatic resource creation is turned ON and the subscription and/or the topic do not exist for a consumer, a subscription and potentially a topic will be created.
-The topic name will be the same as the destination name, and the subscription name will be the same as the consumer group name.
+The topic name will be the same as the destination name, and the subscription name will be the destination name followed by the consumer group name.
 If the consumer group is not specified, an anonymous one will be created with the name `anonymous.<destinationName>.<randomUUID>`.
 
-For example, for the following configuration, a topic named `myEvents` and a subscription called `counsumerGroup1` would be created.
+For example, for the following configuration, a topic named `myEvents` and a subscription called `myEvents.counsumerGroup1` would be created.
```
artembilan: I think we need to add some IMPORTANT note that created subscriptions in advance should follow this pattern otherwise a new one is going to be created to confuse end-user.
Or more over when we configure an `auto-create-resources` to `false` and end-user is going to use an existing subscription he/she is going to be surprised that binding doesn't work because there is no `[name].[group]` subscription.

meltsufin: Good point. Done.

PR #1217, opened 2018-11-12
Add IAP issuer claim verifier

### LINKED ISSUE #1205: Add Cloud IAP issuer claim verifier
1) Check what spring security oauth provides for issuer verification.

This is a follow-up to PR #1190 and issue #1016.

elefeint:
Spring-security conveniently already has an issuer verifier implementation.

Right now, if client apps need to skip issuer validation, they can provide an overridden JwtDecoder bean. 

The alternative is to make the issuer bean injectable based on whether the property is non-empty. This is more flexible but less secure (Google Cloud says issuer MUST be set to the specific value).

Fixes #1205.

### DIFF
--- a/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/security/IapAuthenticationAutoConfiguration.java
+++ b/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/security/IapAuthenticationAutoConfiguration.java
@@ -26,6 +26,7 @@ import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.security.config.annotation.web.configuration.WebSecurityConfigurerAdapter;
 import org.springframework.security.oauth2.jwt.JwtDecoder;
+import org.springframework.security.oauth2.jwt.JwtValidators;
 import org.springframework.security.oauth2.jwt.NimbusJwtDecoderJwkSupport;
 import org.springframework.security.oauth2.server.resource.web.BearerTokenResolver;
 
@@ -59,7 +60,10 @@ public class IapAuthenticationAutoConfiguration {
 	@Bean
 	@ConditionalOnMissingBean
 	public JwtDecoder iapJwtDecoder(IapAuthenticationProperties properties) {
-		return new NimbusJwtDecoderJwkSupport(properties.getRegistry(), properties.getAlgorithm());
+		NimbusJwtDecoderJwkSupport jwkSupport
+				= new NimbusJwtDecoderJwkSupport(properties.getRegistry(), properties.getAlgorithm());
+		jwkSupport.setJwtValidator(JwtValidators.createDefaultWithIssuer(properties.getIssuer()));
+		return jwkSupport;
 	}
 
 	@Bean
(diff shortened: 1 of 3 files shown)


meltsufin:
Would it make sense to add another property like `spring.cloud.gcp.security.iap.issuer.verify=true/false` to make it easier to skip issuer verification? Alternatively, the `spring.cloud.gcp.security.iap.issuer` property being set to `""` could be a marker to skip verification.

elefeint:
Issuer is required on GCP. I can't imagine a scenario in which someone needs IAP authentication on a non-GCP project.
Audience, though, I think we might have to provide ability to turn off.

meltsufin:
For testing though, maybe it would be valuable to be able to disable it by providing an empty string for issuer.

elefeint:
I mean, by the time you are creating a "real" JWT token for testing, spoofing an issuer is the easy part. But it's easy to implement.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/security/IapAuthenticationAutoConfiguration.java
```diff
@@ -59,7 +60,10 @@
 	@Bean
 	@ConditionalOnMissingBean
 	public JwtDecoder iapJwtDecoder(IapAuthenticationProperties properties) {
-		return new NimbusJwtDecoderJwkSupport(properties.getRegistry(), properties.getAlgorithm());
+		NimbusJwtDecoderJwkSupport jwkSupport
+				= new NimbusJwtDecoderJwkSupport(properties.getRegistry(), properties.getAlgorithm());
+		jwkSupport.setJwtValidator(JwtValidators.createDefaultWithIssuer(properties.getIssuer()));
```
artembilan: I think this `JwtValidator` has to be present as a bean, so you can provide something like `StubJwtValidator` for testing purpose to avoid such an issuer validation altogether as @meltsufin would like to do with an extra property.

elefeint: That's an interesting idea. So, I would write something like a NoopValidator, and optionally inject it from a sample app's configuration, to show how to do it? Or just provide NoopValidator without complicating the sample?
I am a bit concerned about allowing end users to make their web app insecure too easily. Perhaps in the scenario of testing, the end user could provide a stub/noop NimbusJwtDecoderJwkSupport instead of just the claims validator; then the end user app is completely responsible for all JWK validation/security themselves.

artembilan: That's good point too, having the fact that you already do `@ConditionalOnMissingBean`.
So, our default impl is based on the built-in `JwtValidators.createDefaultWithIssuer` based on the provided issuer.
And if you don't like it in your app, you are feel free to provide your own `JwtDecoder` bean.

That's all: nothing to discuss any more! 😄

elefeint:
Spring-security conveniently already has an issuer verifier implementation

If client apps need to skip issuer validation, they can provide an overridden JwtDecoder bean.

Fixes #1205.

PR #1218, opened 2018-11-12
Spanner template transactions DML

### LINKED ISSUE #1213: Add executeUpdate method to implementation of Spanner TransactionContext
Spanner API added a new method `executeUpdate` in googleapis/google-cloud-java#3781. As a result, we cannot upgrade to the latest `google-cloud-bom.version` (`0.70.0-alpha`) until we resolve the compilation issue in [`SpannerTransactionManager`](https://github.com/spring-cloud/spring-cloud-gcp/blob/d1c600affb904d45677c87ed0e6ad950822c2c49/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTransactionManager.java#L102).

### LINKED ISSUE #1214: Update Cloud Spanner operations to use returned OperationFuture instead of Operation
Cloud Spanner updated return value of several things (like DatabaseAdminClient) to return a different type in googleapis/google-cloud-java#3881.

As a result, we cannot upgrade to the latest google-cloud-bom.version (0.70.0-alpha) until we resolve the compilation issue in [`SpannerDatabaseAdminTemplate`](https://github.com/spring-cloud/spring-cloud-gcp/blob/c24c07d6a6e232a77ebdec5e92fa3cfbea146e9e/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/admin/SpannerDatabaseAdminTemplate.java#L100)

ChengyuanZhao:
fixes #1213 fixes #1214 

SpannerTemplate supports DML in transactions, 0.70 update.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -321,7 +321,7 @@
 				<plugins>
 					<plugin>
 						<artifactId>maven-surefire-plugin</artifactId>
-						<version>2.22.0</version>
+						<version>2.20</version>
 						<configuration>
 							<!-- need for JaCoCo -->
 							<argLine>@{argLine} -Xms512m -Xmx512m</argLine>
--- a/pom.xml
+++ b/pom.xml
@@ -164,6 +164,7 @@
 				<plugin>
 					<groupId>org.apache.maven.plugins</groupId>
 					<artifactId>maven-surefire-plugin</artifactId>
+					<version>2.20</version> <!-- version 2.22.0 brought in by BOMs crashes -->
 					<configuration>
 						<argLine>-Xms512m -Xmx512m</argLine>
 					</configuration>
--- a/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/admin/SpannerDatabaseAdminTemplate.java
+++ b/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/admin/SpannerDatabaseAdminTemplate.java
@@ -20,13 +20,13 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.ExecutionException;
 
 import com.google.cloud.spanner.Database;
 import com.google.cloud.spanner.DatabaseAdminClient;
 import com.google.cloud.spanner.DatabaseClient;
 import com.google.cloud.spanner.DatabaseId;
 import com.google.cloud.spanner.ResultSet;
-import com.google.cloud.spanner.SpannerException;
 import com.google.cloud.spanner.Statement;
 import com.google.cloud.spanner.Struct;
 
@@ -93,14 +93,17 @@ public class SpannerDatabaseAdminTemplate {
 			if (createDatabase) {
 				this.databaseAdminClient
 						.createDatabase(getInstanceId(), getDatabase(), ddlStrings)
-						.waitFor();
+						.get();
 			}
 			else {
 				this.databaseAdminClient.updateDatabaseDdl(getInstanceId(), getDatabase(),
-						ddlStrings, null).waitFor();
+						ddlStrings, null).get();
 			}
 		}
-		catch (SpannerException e) {
+		catch (InterruptedException e) {
+			throw new SpannerDataException("DDL execution was interrupted", e);
+		}
+		catch (ExecutionException e) {
 			throw new SpannerDataException("DDL could not be executed", e);
 		}
 	}
--- a/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/admin/SpannerDatabaseAdminTemplate.java
+++ b/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/admin/SpannerDatabaseAdminTemplate.java
@@ -88,7 +88,7 @@ public class SpannerDatabaseAdminTemplate {
 	 * already exist.
 	 */
 	public void executeDdlStrings(Iterable<String> ddlStrings,
-			boolean createDatabase) {
+			boolean createDatabase) throws InterruptedException {
 		try {
 			if (createDatabase) {
 				this.databaseAdminClient
@@ -99,9 +99,6 @@ public class SpannerDatabaseAdminTemplate {
 				this.databaseAdminClient.updateDatabaseDdl(getInstanceId(), getDatabase(),
 						ddlStrings, null).get();
 			}
-		}
-		catch (InterruptedException e) {
-			throw new SpannerDataException("DDL execution was interrupted", e);
 		}
 		catch (ExecutionException e) {
 			throw new SpannerDataException("DDL could not be executed", e);
--- a/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/admin/SpannerDatabaseAdminTemplate.java
+++ b/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/admin/SpannerDatabaseAdminTemplate.java
@@ -88,7 +88,7 @@ public class SpannerDatabaseAdminTemplate {
 	 * already exist.
 	 */
 	public void executeDdlStrings(Iterable<String> ddlStrings,
-			boolean createDatabase) throws InterruptedException {
+			boolean createDatabase) {
 		try {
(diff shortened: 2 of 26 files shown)


ChengyuanZhao:
@meltsufin PTAL. this is related to the breaking changes from spanner client lib

ChengyuanZhao:
Looks like there are some test errors in Autoconfig module from moving to 0.70

ChengyuanZhao:
It's some issue with maven-surefire-plugin 

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:2.22.0:test (default-test) on project spring-cloud-gcp-autoconfigure: There are test failures.


java.lang.NoSuchMethodError: org.apache.maven.surefire.report.RunListener.testSetStarting(Lorg/apache/maven/surefire/report/ReportEntry;)V
	at org.apache.maven.surefire.junit4.JUnit4Provider.executeTestSet(JUnit4Provider.java:235)
	at org.apache.maven.surefire.junit4.JUnit4Provider.invoke(JUnit4Provider.java:161)
	at org.apache.maven.surefire.booter.ForkedBooter.invokeProviderInSameClassLoader(ForkedBooter.java:383)
	at org.apache.maven.surefire.booter.ForkedBooter.runSuitesInProcess(ForkedBooter.java:344)
	at org.apache.maven.surefire.booter.ForkedBooter.execute(ForkedBooter.java:125)
	at org.apache.maven.surefire.booter.ForkedBooter.main(ForkedBooter.java:417)
```

pom.xml
```diff
@@ -164,6 +164,7 @@
 				<plugin>
 					<groupId>org.apache.maven.plugins</groupId>
 					<artifactId>maven-surefire-plugin</artifactId>
+					<version>2.20</version> <!-- version 2.22.0 brought in by BOMs crashes -->
```
meltsufin: Are we tracking this in an issue?

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/admin/SpannerDatabaseAdminTemplate.java
```diff
@@ -88,19 +88,19 @@ public SpannerDatabaseAdminTemplate(DatabaseAdminClient databaseAdminClient,
 	 * already exist.
 	 */
 	public void executeDdlStrings(Iterable<String> ddlStrings,
-			boolean createDatabase) {
+			boolean createDatabase) throws InterruptedException {
```
meltsufin: Can we just catch the `InterruptedException` and throw the `SpannerDataException`?

meltsufin: So, did you try to just re-interrupt in the catch block with `Thread.currentThread().interrupt()`?

ChengyuanZhao: No, I didn't think that was preferable to just throwing the exception. 

The spanner admin template isn't the place to interrupt the thread in my opinion. It should be just as blind to handling the interrupt as the DDL request it submitted to the client library.

meltsufin: I don't think it's very user-friendly. Why don't we just return the future then?

ChengyuanZhao: I think it's perfectly user-friendly. The user is totally free to ignore the exception because they aren't subject to sonarcube haha.

Returning a future also necessarily means they have to deal with the exception, too. Not to mention it's far far less user friendly. All the user wanted to do was run some DDL strings. Why should he have to worry about  a future??

ChengyuanZhao:
@dmitry-s PTAL :)

pom.xml
ChengyuanZhao: yes, it's #1222

spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplateTests.java
```diff
@@ -148,6 +159,22 @@ public void readOnlyTransactionTest() {
 		verify(readOnlyTransaction, times(2)).read(eq("custom_test_table"), any(), any());
 	}
 
+	@Test(expected = SpannerDataException.class)
```
dmitry-s: please check exception messages, not just exception classes

ChengyuanZhao: I think this is gonna need to be a separate issue. We have about 600 tests in repo and about 280 of them are in Spanner, and only recently have we started checking the messages in Datastore.

#1225

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/admin/SpannerDatabaseAdminTemplate.java
meltsufin: The future would allow them to not block the thread on the operation completion, unless they want to.

ChengyuanZhao: Yes that's certainly true (and they would need to handle the interrupt too), and they would have even more flexibility if they used the client lib directly. The purpose of this method is purely for convenience, and I like the clean logical flow of DDL string straight to tables being created without worry about intermediate Java things like futures and threads.

meltsufin: Sure, I buy into the concept of convenience at the template level, which for me also includes not rethrowing checked exceptions. AFIAK, Spring's convention is no checked exceptions in templates. @artembilan Can you comment on this please?

ChengyuanZhao: I don't think it's inconvenient for the user to  catch and ignore the exception. These types of admin operations happening over the internet are usually things a user would have a in a try-catch block anyway.

meltsufin:
@artembilan @mp911de How would you recommend handling the `InterruptedException` in the `SpannerTemplate` methods coming from the underlying client library that we're using. I believe there is a convention around not throwing checked exceptions in template methods. WDYT?

artembilan:
> How would you recommend handling the InterruptedException in the SpannerTemplate methods coming from the underlying client library that we're using. I believe there is a convention around not throwing checked exceptions in template methods. 

Something like this:
```
catch (InterruptedException ex) {
	Thread.currentThread().interrupt();
	throw new IllegalStateException("Interrupted by the client library", ex);
}
```

Or you can use some specific `DataAccessException` sub-class.

ChengyuanZhao:
@meltsufin @artembilan change made, PTAL

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/admin/SpannerDatabaseAdminTemplate.java
```diff
@@ -93,14 +93,18 @@ public void executeDdlStrings(Iterable<String> ddlStrings,
...
+						.get();
 			}
 			else {
 				this.databaseAdminClient.updateDatabaseDdl(getInstanceId(), getDatabase(),
-						ddlStrings, null).waitFor();
+						ddlStrings, null).get();
 			}
 		}
-		catch (SpannerException e) {
+		catch (InterruptedException e) {
```
artembilan: Not related to this change, just for the future consideration: there is a rule in the upstream Checkstyle config which disallows a single letter identifiers. So, would be better to keep that in mind and start follow such a practice and don't bump to the pain to fix all of them at once, when we come to the mentioned config.

Thanks for understanding.

PR #1219, opened 2018-11-13
Cleanup anonymous Pub/Sub subscriptions

### LINKED ISSUE #1178: Cleanup anonymous Pub/Sub subscriptions created by the binder
When a consumer group name is not provided for a consumer destination, the binder will optionally create anonymous subscriptions. These anonymous subscriptions should be deleted when the consumer destination shuts down.

meltsufin:
When a consumer group name is not provided for a consumer destination,
the binder will optionally create anonymous subscriptions. These
anonymous subscriptions will now be deleted when the consumer
destination shuts down and the `afterUnbindConsumer` method is called.

Fixes #1178.

### DIFF
--- a/spring-cloud-gcp-pubsub-stream-binder/src/main/java/org/springframework/cloud/gcp/stream/binder/pubsub/provisioning/PubSubChannelProvisioner.java
+++ b/spring-cloud-gcp-pubsub-stream-binder/src/main/java/org/springframework/cloud/gcp/stream/binder/pubsub/provisioning/PubSubChannelProvisioner.java
@@ -16,7 +16,13 @@
 
 package org.springframework.cloud.gcp.stream.binder.pubsub.provisioning;
 
+import java.util.HashSet;
+import java.util.Set;
 import java.util.UUID;
+
+import com.google.pubsub.v1.Subscription;
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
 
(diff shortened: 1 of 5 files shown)


spring-cloud-gcp-pubsub-stream-binder/src/main/java/org/springframework/cloud/gcp/stream/binder/pubsub/provisioning/PubSubChannelProvisioner.java
```diff
@@ -58,24 +68,51 @@ public ProducerDestination provisionProducerDestination(String name,
...
 		}
+		else if (!pubSubSubscription.getTopic().equals(name)) {
+			throw new ProvisioningException("Existing '" + subscription + "' subscription is for a different topic '"
+					+ pubSubSubscription.getTopic() + "'.");
+		}
 		return new PubSubConsumerDestination(subscription);
 	}
+
+	public void afterUnbindConsumer(ConsumerDestination destination) {
+		if (this.anonymousSubscriptions.contains(destination.getName())) {
```
artembilan: I think you just can use and atomic operation here: `this.anonymousSubscriptions.remove(destination.getName()) != null`.

meltsufin:
@artembilan I'll update the docs in the next PR where we make anonymous subscriptions always created.

PR #1220, opened 2018-11-13
Always create subscriptions for anonymous groups

### LINKED ISSUE #1180: Unconditionally create subscriptions for anonymous consumer groups in binder
Currently the binder would only create Pub/Sub subscriptions for anonymous consumer groups when resource auto-creation is turned on. However, since the subscription names are completely random for anonymous consumer groups, there is no way the user can have the subscription already existing. So, we should unconditionally create consumer group regardless of auto-creation setting.

The change would be in `PubSubChannelProvisioner.provisionConsumerDestination()`.

meltsufin:
Previously, the binder would only create Pub/Sub subscriptions for
anonymous consumer groups when resource auto-creation was turned on.
This change ensures that we uncodnitionally create subscriptions for
anonymous consumer groups regardless of the auto-creation setting.

Fixes #1180.

### DIFF
--- a/spring-cloud-gcp-docs/src/main/asciidoc/spring-stream.adoc
+++ b/spring-cloud-gcp-docs/src/main/asciidoc/spring-stream.adoc
@@ -58,10 +58,12 @@ spring.cloud.stream.gcp.pubsub.bindings.events.producer.auto-create-resources=tr
 
 If automatic resource creation is turned ON and the subscription and/or the topic do not exist for a consumer, a subscription and potentially a topic will be created.
 The topic name will be the same as the destination name, and the subscription name will be the destination name followed by the consumer group name.
-If the consumer group is not specified, an anonymous one will be created with the name `anonymous.<destinationName>.<randomUUID>`.
+
+Regardless of the automatic resource creation setting, if the consumer group is not specified, an anonymous one will be created with the name `anonymous.<destinationName>.<randomUUID>`.
+Then when the binder shuts down, all Pub/Sub subscriptions created for anonymous consumer groups will be automatically cleaned up.
 
 For example, for the following configuration, a topic named `myEvents` and a subscription called `myEvents.counsumerGroup1` would be created.
-If the consumer group is not specified, a subscription called `anonymous.myEvents.a6d83782-c5a3-4861-ac38-e6e2af15a7be` would be created.
+If the consumer group is not specified, a subscription called `anonymous.myEvents.a6d83782-c5a3-4861-ac38-e6e2af15a7be` would be created and later cleaned up.
 
 IMPORTANT: If you are manually creating Pub/Sub subscriptions for consumers, make sure that they follow the naming convention of `<destinationName>.<consumerGroup>`.
 
--- a/spring-cloud-gcp-docs/src/main/asciidoc/spring-stream.adoc
+++ b/spring-cloud-gcp-docs/src/main/asciidoc/spring-stream.adoc
@@ -59,7 +59,7 @@ spring.cloud.stream.gcp.pubsub.bindings.events.producer.auto-create-resources=tr
 If automatic resource creation is turned ON and the subscription and/or the topic do not exist for a consumer, a subscription and potentially a topic will be created.
 The topic name will be the same as the destination name, and the subscription name will be the destination name followed by the consumer group name.
 
-Regardless of the automatic resource creation setting, if the consumer group is not specified, an anonymous one will be created with the name `anonymous.<destinationName>.<randomUUID>`.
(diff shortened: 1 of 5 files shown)


spring-cloud-gcp-docs/src/main/asciidoc/spring-stream.adoc
```diff
@@ -58,10 +58,12 @@ spring.cloud.stream.gcp.pubsub.bindings.events.producer.auto-create-resources=tr
 
 If automatic resource creation is turned ON and the subscription and/or the topic do not exist for a consumer, a subscription and potentially a topic will be created.
 The topic name will be the same as the destination name, and the subscription name will be the destination name followed by the consumer group name.
-If the consumer group is not specified, an anonymous one will be created with the name `anonymous.<destinationName>.<randomUUID>`.
+
+Regardless of the automatic resource creation setting, if the consumer group is not specified, an anonymous one will be created with the name `anonymous.<destinationName>.<randomUUID>`.
```
ChengyuanZhao: Regarding "automatic resource creation setting": 

I think you need some hyphens in there. Without hyphens it can mean that the setting for resource creation is itself automatic, rather than it being a manual setting that makes resource creation automatic.

spring-cloud-gcp-pubsub-stream-binder/src/main/java/org/springframework/cloud/gcp/stream/binder/pubsub/properties/PubSubCommonProperties.java
```diff
@@ -0,0 +1,35 @@
...
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package org.springframework.cloud.gcp.stream.binder.pubsub.properties;
+
+/**
+ * Properties commonn to consumers and producers.
```
artembilan: Typo: double `n`

spring-cloud-gcp-pubsub-stream-binder/src/main/java/org/springframework/cloud/gcp/stream/binder/pubsub/properties/PubSubCommonProperties.java
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * @author Mike Eltsufin
+ * @since 1.1
+ */
+public class PubSubCommonProperties {
+	private boolean autoCreateResources = true;
```
artembilan: Blank line before property.

spring-cloud-gcp-pubsub-stream-binder/src/main/java/org/springframework/cloud/gcp/stream/binder/pubsub/provisioning/PubSubChannelProvisioner.java
```diff
@@ -47,65 +47,61 @@
 
 	private final PubSubAdmin pubSubAdmin;
 
-	private final Set<String> anonymousSubscriptions = new HashSet<>();
+	private final Set<String> anonymousSubscriptionNames = new HashSet<>();
```
ChengyuanZhao: Please consider renaming this. "anonymous" literally means without-name, so "anonymous subscription names" is like "subscription names without names".

artembilan: `anonymousGroupSubscriptionNames` ?

spring-cloud-gcp-docs/src/main/asciidoc/spring-stream.adoc
ChengyuanZhao: Why not use another randomly generated string instead of "anonymous" in the name? 

"anonymous" is a common word and might be used in a different context. (for example,  a user might have "anonymous" refer to the subject matter the program is about, like poems from known poets or anonymous ones). I don't think it should be reserved for use here.

artembilan: That's a terminology has chosen in Spring Cloud Stream: https://docs.spring.io/spring-cloud-stream/docs/current/reference/htmlsingle/#_common_binding_properties
> group

>    The consumer group of the channel. Applies only to inbound bindings. See Consumer Groups.

>    Default: null (indicating an anonymous consumer).

That's not our own call here.

If you have some concern on the matter, feel free to raise an issue against that project.

PR #1227, opened 2018-11-14
Spanner custom conversions used in composite key accessor

### LINKED ISSUE #1226: Spanner use custom converters when forming composite key from id properties
Spanner composite keys can only be constructed from set of allowed types. Currently in the few places where this is done there is no conversion done. We should use the same custom converters that are used for the other properties in those cases.

ChengyuanZhao:
Fixes #1226

Utilizes custom converters in the composite key id accessor. 
We had the logic all built, it just needed to be used.


spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/ConverterAwareMappingSpannerEntityWriter.java
```diff
@@ -49,7 +49,7 @@
  * @since 1.1
  */
 public class ConverterAwareMappingSpannerEntityWriter implements SpannerEntityWriter {
-	private static Set<Class> SPANNER_KEY_COMPATIBLE_TYPES = ImmutableSet
+	private final static Set<Class> SPANNER_KEY_COMPATIBLE_TYPES = ImmutableSet
```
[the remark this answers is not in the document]
ChengyuanZhao: you gotta be kidding me lol. 
will fix.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/SpannerKeyWriter.java
```diff
@@ -0,0 +1,44 @@
...
+	 * @return the Cloud Spanner key.
+	 */
+	Key writeToKey(Object key);
+
+	/**
+	 * Get the SpannerWriteConverter used to convert types into Cloud Spanner compatible
+	 * types.
+	 * @return a SpannerWriteConverter
+	 */
+	SpannerWriteConverter getSpannerWriteConverter();
```
meltsufin: What does this have to do with keys?

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/SpannerKeyWriter.java
```diff
@@ -0,0 +1,44 @@
...
+ */
+public interface SpannerKeyWriter {
+
+	/**
+	 * Writes a given object to a Cloud Spanner key.
+	 * @param key the object containing the key values. This can already be a Cloud Spanner
+	 *     key, a single key component, or an array of key components.
+	 * @return the Cloud Spanner key.
+	 */
+	Key writeToKey(Object key);
```
meltsufin: It seems more like a `convertToKey` method.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/SpannerEntityWriter.java
```diff
@@ -38,6 +36,4 @@
 	 * written.
 	 */
 	void write(Object source, MultipleValueBinder sink, Set<String> includeColumns);
-
-	Key writeToKey(Object key);
```
meltsufin: What's the benefit of pulling this out into the `SpannerKeyWriter` interface?

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/SpannerKeyWriter.java
```diff
@@ -0,0 +1,44 @@
...
+import com.google.cloud.spanner.Key;
+
+/**
+ * An interface that can create Cloud Spanner keys and utlizes conversion.
+ *
+ * @author Chengyuan Zhao
+ *
+ * @since 1.1
+ */
+public interface SpannerKeyWriter {
```
meltsufin: I'm not sure why this interface is useful vs. keeping these methods in `SpannerEntityWriter`.

ChengyuanZhao: The reason is this new one is what's added to the constructor of SpannerPersistentEntityImpl because it is needed to convert  the properties that comprise the composite key.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/SpannerEntityWriter.java
ChengyuanZhao: The reason is this new one is what's added to the constructor of SpannerPersistentEntityImpl because it is needed to convert  the properties that comprise the composite key. We don't need the write-to-entity functionality, but only the write-to-key functionality.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/SpannerKeyWriter.java
ChengyuanZhao: will do

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/SpannerKeyWriter.java
ChengyuanZhao: This converter is needed to convert the individual properties that comprise the composite key.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/SpannerKeyWriter.java
meltsufin: Why would giving `SpannerEntityWriter` to the `SpannerPersistentEntityImpl` be so bad? You never implement  just `SpannerKeyWriter`, and it looks like an internal class.

ChengyuanZhao: To be honest I personally don't think it's bad, but I did it because I thought if I gave it a whole entity-converter interface to something that only uses the key-converter functions would offend somebody with more tender java sensibilities. I can get rid of the interface.

ChengyuanZhao: done

PR #1231, opened 2018-11-15
Slice and Page methods support

### LINKED ISSUE #1167: Slice and Page return types are not supported in pagination
Could you please make possible use Slice and Page as return type.
[Spring Data Reference with supported types](https://docs.spring.io/spring-data/jpa/docs/current/reference/html/#repositories.special-parameters)

For example, we have a model:

```
Entity(name = "singers")
public class Singer {

    @Id
    @Field(name = "singer_id")
    private String singerId;

    @Field(name = "first_name")
    private String firstName;
}
```
Datastore repository has a simple search method (with pagination):
```
public interface SingerRepository extends DatastoreRepository<Singer, String> {
    Slice<Singer> findAllByFirstName(String firstName, Pageable pageable);
}
```
An exception occurs after executing this method. List as return type works fine. 

```
Caused by: org.springframework.cloud.gcp.data.datastore.core.mapping.DatastoreDataException: Unable to convert class java.util.ArrayList to class com.example.Singer
	at org.springframework.cloud.gcp.data.datastore.core.convert.TwoStepsConversions.convertOnReadSingle(TwoStepsConversions.java:231) ~[spring-cloud-gcp-data-datastore-1.1.0.BUILD-20181031.173949-140.jar:1.1.0.BUILD-SNAPSHOT]
	at org.springframework.cloud.gcp.data.datastore.core.convert.TwoStepsConversions.convertOnRead(TwoStepsConversions.java:164) ~[spring-cloud-gcp-data-datastore-1.1.0.BUILD-20181031.173949-140.jar:1.1.0.BUILD-SNAPSHOT]
	at org.springframework.cloud.gcp.data.datastore.core.convert.TwoStepsConversions.convertOnRead(TwoStepsConversions.java:105) ~[spring-cloud-gcp-data-datastore-1.1.0.BUILD-20181031.173949-140.jar:1.1.0.BUILD-SNAPSHOT]
	at org.springframework.cloud.gcp.data.datastore.repository.query.PartTreeDatastoreQuery.convertResultCollection(PartTreeDatastoreQuery.java:144) ~[spring-cloud-gcp-data-datastore-1.1.0.BUILD-20181031.173949-140.jar:1.1.0.BUILD-SNAPSHOT]
	at org.springframework.cloud.gcp.data.datastore.repository.query.PartTreeDatastoreQuery.execute(PartTreeDatastoreQuery.java:139) ~[spring-cloud-gcp-data-datastore-1.1.0.BUILD-20181031.173949-140.jar:1.1.0.BUILD-SNAPSHOT]
	at org.springframework.data.repository.core.support.RepositoryFactorySupport$QueryExecutorMethodInterceptor.doInvoke(RepositoryFactorySupport.java:605) ~[spring-data-commons-2.1.2.RELEASE.jar:2.1.2.RELEASE]
	at org.springframework.data.repository.core.support.RepositoryFactorySupport$QueryExecutorMethodInterceptor.lambda$invoke$3(RepositoryFactorySupport.java:595) ~[spring-data-commons-2.1.2.RELEASE.jar:2.1.2.RELEASE]
	at org.springframework.data.repository.core.support.RepositoryFactorySupport$QueryExecutorMethodInterceptor.invoke(RepositoryFactorySupport.java:595) ~[spring-data-commons-2.1.2.RELEASE.jar:2.1.2.RELEASE]
	at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:186) ~[spring-aop-5.1.2.RELEASE.jar:5.1.2.RELEASE]
	at org.springframework.data.projection.DefaultMethodInvokingMethodInterceptor.invoke(DefaultMethodInvokingMethodInterceptor.java:59) ~[spring-data-commons-2.1.2.RELEASE.jar:2.1.2.RELEASE]
	at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:186) ~[spring-aop-5.1.2.RELEASE.jar:5.1.2.RELEASE]
	at org.springframework.aop.interceptor.ExposeInvocationInterceptor.invoke(ExposeInvocationInterceptor.java:93) ~[spring-aop-5.1.2.RELEASE.jar:5.1.2.RELEASE]
	at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:186) ~[spring-aop-5.1.2.RELEASE.jar:5.1.2.RELEASE]
	at org.springframework.data.repository.core.support.SurroundingTransactionDetectorMethodInterceptor.invoke(SurroundingTransactionDetectorMethodInterceptor.java:61) ~[spring-data-commons-2.1.2.RELEASE.jar:2.1.2.RELEASE]
	at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:186) ~[spring-aop-5.1.2.RELEASE.jar:5.1.2.RELEASE]
	at org.springframework.aop.framework.JdkDynamicAopProxy.invoke(JdkDynamicAopProxy.java:212) ~[spring-aop-5.1.2.RELEASE.jar:5.1.2.RELEASE]
	at com.sun.proxy.$Proxy64.findAllByLastName(Unknown Source) ~[na:na]
	at com.example.DatastoreRepositoryExample.lambda$commandLineRunner$0(DatastoreRepositoryExample.java:64) [classes/:na]
	at org.springframework.boot.SpringApplication.callRunner(SpringApplication.java:813) [spring-boot-2.1.0.RELEASE.jar:2.1.0.RELEASE]
	... 5 common frames omitted

```

### LINKED ISSUE #1266: Support pagination for custom queries
From documentation:
`Paging and sorting options are also supported in DatastoreTemplate by supplying a DatastoreQueryOptions to findAll.`

Is it possible to add pagination for custom queries?

For example, we have a model:
```
Entity(name = "singers")
public class Singer { 

    @Id     
    @Field(name = "singer_id")
    private String singerId;

    @Field(name = "created_date")
    private Timestamp createdDate;

}
```
Datastore repository has a search method (with pagination):

` List<Singer> findAllByCreatedDateLessThan(Timestamp timestamp, Pageable pageable);`

I insert 3 singers with created date = 01/01/2019
Search date = 01/01/2020

`List<Singer> singers = singerRepository.findAllByCreatedDateLessThan(searchDate, ageRequest.of(0, 1));`

Actual result - all 3 records returned.
Expected result - 1 record should be returned.

dmitry-s:
fixes #1167 fixes #1266


spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/PartTreeDatastoreQuery.java
```diff
@@ -101,15 +118,50 @@ private void validateAndSetFilterParts() {
...
+		Object result;
+		if (isSliceQuery() || isPageQuery()) {
+			List<?> resultEntries = (List) execute(parameters, returnedObjectType, List.class, false);
+			Long totalCount = (Long) execute(parameters, Long.class, null, true);
+
+			ParameterAccessor paramAccessor =
+					new ParametersParameterAccessor(getQueryMethod().getParameters(), parameters);
+
+			Pageable pageable = paramAccessor.getPageable();
+			if (getQueryMethod().isSliceQuery()) {
```
ChengyuanZhao: Could you call the convenience method here too ? `isSliceQuery()` that you use in the outer if-statement.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/PartTreeDatastoreQuery.java
```diff
@@ -101,15 +118,50 @@ private void validateAndSetFilterParts() {
...
+				boolean hasNext = !pageable.equals(Pageable.unpaged()) && pageable.next().getOffset() < totalCount;
+				result = new SliceImpl<>(resultEntries, pageable, hasNext);
+			}
+			else {
+				result = new PageImpl<>(resultEntries, pageable, totalCount);
```
ChengyuanZhao: Since Page is a child of Slice, could you just use PageImpl for both slice and page?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/PartTreeDatastoreQuery.java
```diff
@@ -86,13 +94,22 @@ private void validateAndSetFilterParts() {
 					"Cloud Datastore structured queries do not support the Distinct keyword.");
 		}
 
-		List parts = this.tree.getParts().get().collect(Collectors.toList());
+		if ((!this.tree.getSort().equals(Sort.unsorted())) && (getQueryMethod().getParameters().hasPageableParameter()
+				|| getQueryMethod().getParameters().hasSortParameter())) {
+			throw new DatastoreDataException("Ambiguous sort options (method name and parameter).");
```
ChengyuanZhao: Is there a guard against the user providing more than one pageable or sort parameter?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/PartTreeDatastoreQuery.java
```diff
@@ -101,15 +118,50 @@ private void validateAndSetFilterParts() {
 
 	@Override
 	public Object execute(Object[] parameters) {
+		Class<?> returnedObjectType = getQueryMethod().getReturnedObjectType();
+		Object result;
+		if (isSliceQuery() || isPageQuery()) {
+			List<?> resultEntries = (List) execute(parameters, returnedObjectType, List.class, false);
+			Long totalCount = (Long) execute(parameters, Long.class, null, true);
```
ChengyuanZhao: I understand that there isn't another way to get the count other than querying again for it.
This work is definitely worth putting in since it's done, but I question the original premise that because the slice/param feature is useful in old-school SQL jpa spring data that it also makes sense here in Datastore.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/PartTreeDatastoreQuery.java
dmitry-s: Excellent point!

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/PartTreeDatastoreQuery.java
dmitry-s: I don't think it's our responsibility to check that. Instead, `Parameters` class should deal with that.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/PartTreeDatastoreQuery.java
meltsufin: Have you looked into [datastore statistics](https://cloud.google.com/datastore/docs/concepts/stats)?
At the very least we should clearly document this inefficiency in the refdoc.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/PartTreeDatastoreQuery.java
```diff
@@ -101,15 +107,40 @@ private void validateAndSetFilterParts() {
 
 	@Override
 	public Object execute(Object[] parameters) {
+		Class<?> returnedObjectType = getQueryMethod().getReturnedObjectType();
+		if (isSliceQuery() || isPageQuery()) {
```
meltsufin: Note that `Slice` specifically doesn't require the total count. So, it is an unnecessary inefficiency to do this expensive count of total records, when it's not actually required.

dmitry-s: It doesn't require the total count, but it requires a boolean that indicates if the next slice exists. And the only way to know is to actually count the total number of records and compare it to (current offset + slice size). That's exactly what `Page` does to implement `Slice`.

meltsufin: Don't you only need to query for {page size} + 1 records to know if there is another slice?

dmitry-s:
@meltsufin @ChengyuanZhao PTAL

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/PartTreeDatastoreQuery.java
ChengyuanZhao: noting this just so it's here for reference, but after looking into the stats tables it turns out they are not updated in real time (more like at a few points per day), so they cannot be used for our purposes.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/PartTreeDatastoreQuery.java
```diff
@@ -101,15 +109,49 @@ private void validateAndSetFilterParts() {
...
+			List<?> resultEntries = (List) execute(parameters, returnedObjectType, List.class, false);
+			Long totalCount = (Long) execute(parameters, Long.class, null, true);
+
+			ParameterAccessor paramAccessor =
+					new ParametersParameterAccessor(getQueryMethod().getParameters(), parameters);
+
+			return new PageImpl<>(resultEntries, paramAccessor.getPageable(), totalCount);
+		}
+		if (isSliceQuery()) {
+			DatastoreTemplate.LimitedResult result = (DatastoreTemplate.LimitedResult) execute(parameters,
```
ChengyuanZhao: It looks to me that `sliceQuery` and `LimitedResult` are only used here, and the purpose is only to see if the given limit count is bigger or smaller than the result. 

I think we should be doing that here instead of having a public `sliceQuery` method in DatstoreTemplate and a public inner class `LimitedResult`. The reason is I think those two things don't have a use to our users, and are really implementation things that we currently and foreseeable-y only need here, so why not move them here instead?

docs/src/main/asciidoc/datastore.adoc
```diff
@@ -809,6 +815,14 @@ Delete queries can have the following return types:
 * A collection of entities that were deleted
 * 'void'
 
+Methods can have `org.springframework.data.domain.Pageable` parameter to control pagination and sorting,
+or `org.springframework.data.domain.Sort` parameter to control sorting only.
+See https://docs.spring.io/spring-data/data-commons/docs/current/reference/html/#repositories.query-methods[Spring Data documentation] for details.
+
+For returning multiple items in a repository method, we support Java collections as well as `org.springframework.data.domain.Page` and `org.springframework.data.domain.Slice`.
+If a method's return type is `org.springframework.data.domain.Page`, the returned object will include current page,
```
meltsufin: Please put a warning here about using Page vs Slice, that Page requires retrieval of the keys for all entities of the kind and that Slice is much more efficient.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/PartTreeDatastoreQuery.java
```diff
@@ -125,50 +167,81 @@ else if (!returnedTypeIsNumber) {
...
+		Sort sort = this.tree.getSort();
+		if (getQueryMethod().getParameters().hasPageableParameter()) {
+			sort = sort.and(paramAccessor.getPageable().getSort());
+		}
+		if (getQueryMethod().getParameters().hasSortParameter()) {
+			sort = sort.and(paramAccessor.getSort());
+		}
+
+		if (paramAccessor.getPageable().isPaged() && !total) {
+			limit = paramAccessor.getPageable().getPageSize() + (isSliceQuery() ? 1 : 0);
```
ChengyuanZhao: the `? 1 : 0` is a cute little trick that should have a comment to explain it haha.

That `sliceQuery` for slice queries absolutely depends on this +1 extra result to set the limit boolean is another reason to get rid of that method and that inner class and have everything related to slices in this class here.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/PartTreeDatastoreQuery.java
```diff
@@ -227,4 +300,38 @@ private void applySelectWithFilter(Object[] parameters, Builder builder) {
 				: filters[0]);
 	}
 
+	private LimitedResult sliceQuery(Query query, Class<T> entityClass, Integer limit) {
```
meltsufin: Maybe rename limit to conversionLimit, if you need it at all. Perhaps it can be pulled from the query itself.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
```diff
@@ -373,7 +379,7 @@ private void validateKey(Object entity, PathElement ancestorPE) {
 		throw new DatastoreDataException("Descendant object has a key without current ancestor");
 	}
 
-	private <T> Collection<T> convertEntitiesForRead(
+	public <T> List<T> convertEntitiesForRead(
```
meltsufin: document this method

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/PartTreeDatastoreQuery.java
```diff
@@ -125,50 +164,92 @@ else if (!returnedTypeIsNumber) {
...
+			//the additional item will not be converted on read
+			exceedsLimit = items.size() > limit;
+			if (exceedsLimit) {
+				items = items.subList(0, limit);
+			}
+		}
+
+		ParameterAccessor paramAccessor = new ParametersParameterAccessor(getQueryMethod().getParameters(), parameters);
+		Pageable pageable = paramAccessor.getPageable();
+		List entities = (List) this.datastoreTemplate
```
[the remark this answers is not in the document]
dmitry-s: that's not going to work

PR #1243, opened 2018-11-19
Validate IAP security token audience

### LINKED ISSUE #1216: Add audience IAP claim verifier
Create a new verifier for "aud" (audience) -- this will have to be environment-specific. Use `org.springframework.cloud.gcp.autoconfigure.core.AppEngineCondition` to determine which text pattern to match.

For app engine: `"/projects/%s/apps/%s"`
For compute engine: `"/projects/%s/global/backendServices/%s"`

For reference: https://cloud.google.com/iap/docs/signed-headers-howto

This is a follow-up to PR #1190 and issue #1016. The first part of this followup is #1205.

elefeint:
## New project spring-cloud-gcp-security-iap
* Fixed-string audience validation `AudienceValidator`; the string comes from a constructor-injected AudienceProvider.
* Custom audience string providers for App-engine and Compute-engine(`AppEngineAudienceProvider`, `ComputeEngineAudienceProvider`).

## Changes in spring-cloud-gcp-core
* New utility to determine which environment an application runs on (`DefaultGcpEnvironmentProvider`, `GcpEnvironment`), with a mockable interface `GcpEnvironmentProvider`. The logic comes from  [google-cloud-java](https://github.com/googleapis/google-cloud-java/blob/27c90f1298bb27a618ff59b3cd17321c577f3a6b/google-cloud-clients/google-cloud-logging/src/main/java/com/google/cloud/logging/MonitoredResourceUtil.java#L169), but there it's a private implementation detail.
* New wrapper around static Google Java Client metadata server functionality (`DefaultGcpMetadataProvider`) with a mockable interface (`MetadataProvider`).

## Changes in spring-cloud-gcp-autoconfigure
* A new Conditional rule`OnGcpEnvironmentCondition` for determining which environment an application runs on, plus a custom annotation for enabling it `ConditionalOnGcpEnvironment`.
* `IapAuthenticationAutoConfiguration` now wires in an audience validator, where possible. Property-based fixed audience takes precedence, but if no property is provided, the runtime environment will determine whether AppEngine (either Flexible or Standard) or ComputeEngine (also applies to Kubernetes) format is used.
* IAP property is now on by default -- this is now possible because there is a dedicated project that can be verified in the `@ConditionalOnClass` check.

Fixes #1216


spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/core/environment/ConditionalOnGcpEnvironment.java
```diff
@@ -0,0 +1,37 @@
...
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import org.springframework.cloud.gcp.core.GcpEnvironment;
+import org.springframework.context.annotation.Conditional;
+
+@Target({ ElementType.TYPE, ElementType.METHOD })
```
artembilan: Some Javadocs would be great. This is public API and Spring tradition to be as helpful as possible.
Plus don't forget `@since` and `@author`

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/core/environment/OnGcpEnvironmentCondition.java
```diff
@@ -0,0 +1,55 @@
...
+import org.springframework.boot.autoconfigure.condition.ConditionOutcome;
+import org.springframework.boot.autoconfigure.condition.SpringBootCondition;
+import org.springframework.cloud.gcp.core.GcpEnvironment;
+import org.springframework.cloud.gcp.core.GcpEnvironmentProvider;
+import org.springframework.context.annotation.ConditionContext;
+import org.springframework.core.type.AnnotatedTypeMetadata;
+
+/**
+ * Determine which GCP environment the application is running on.
+ *
```
artembilan: `@since` is missed.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/security/IapAuthenticationAutoConfiguration.java
```diff
@@ -54,18 +71,73 @@
...
 @EnableConfigurationProperties(IapAuthenticationProperties.class)
 public class IapAuthenticationAutoConfiguration {
 
+	/**
+	 * Provides default implementation for determining GCP environment.
+	 * Can be overridden to avoid interacting with real environment.
+	 */
+	@Bean
+	@ConditionalOnMissingBean
+	public GcpEnvironmentProvider gcpEnvironmentProvider() {
```
artembilan: I think this one must be `static` to avoid some early `BeanFactory` access when we evaluate condition.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/security/IapAuthenticationAutoConfiguration.java
```diff
@@ -59,13 +73,57 @@
 
 	@Bean
 	@ConditionalOnMissingBean
-	public JwtDecoder iapJwtDecoder(IapAuthenticationProperties properties) {
+	public JwtDecoder iapJwtDecoder(IapAuthenticationProperties properties, GcpProjectIdProvider projectIdProvider,
+					ObjectProvider<AudienceValidator> audienceVerifier) {
+
+		System.out.println("============================= GCP Environment diagnostics; ");
```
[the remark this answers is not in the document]
artembilan: No `System.out` in the project at all, please.

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/security/IapAuthenticationAutoConfigurationTests.java
```diff
@@ -155,4 +162,14 @@ public BearerTokenResolver bearerTokenResolver() {
 			return httpServletRequest -> FAKE_USER_TOKEN;
 		}
 	}
+
+	@Configuration
+	static class TestConfiguration {
+		@Bean
+		GcpProjectIdProvider mockProvider() {
+			//when(mockProjectIdProvider.getProjectId()).thenReturn("fake-project-id");
```
artembilan: If we need it, uncomment, otherwise - remove

spring-cloud-gcp-core/src/main/java/org/springframework/cloud/gcp/core/GcpEnvironment.java
```diff
@@ -0,0 +1,46 @@
...
+
+import java.util.function.BooleanSupplier;
+
+import com.google.cloud.MetadataConfig;
+import com.google.cloud.ServiceOptions;
+
+/**
+ * {@code COMPUTE_ENGINE} matches instances of Compute Engine that are not also AppEngine Flexible.
+ * {@code ANY_APP_ENGINE} matches both, AppEngine Flexible and AppEngine Standard.
+ * {@code ANY_CONTAINER} matches both, Compute Engine and Kubernetes Engine.
```
artembilan: Why no comment about all others?
OTOH why just don't add appropriate Javadocs for each enum value?

`@since`, `@author`, general class level Javadoc

spring-cloud-gcp-core/src/main/java/org/springframework/cloud/gcp/core/GcpEnvironment.java
```diff
@@ -0,0 +1,46 @@
...
+ */
+public enum GcpEnvironment {
+	KUBERNETES_ENGINE(() -> System.getenv("KUBERNETES_SERVICE_HOST") != null),
```
artembilan: Please, be sure to wrap each class member with blank lines.

spring-cloud-gcp-security-iap/src/main/java/org/springframework/cloud/gcp/security/iap/AppEngineAudienceValidator.java
```diff
@@ -0,0 +1,35 @@
...
+import org.springframework.cloud.gcp.core.GcpProjectIdProvider;
+
+public class AppEngineAudienceValidator extends AudienceValidator {
+
+	public AppEngineAudienceValidator(GcpProjectIdProvider projectIdProvider) {
+		ResourceManager resourceManager = ResourceManagerOptions.getDefaultInstance().getService();
+		Project project = resourceManager.get(projectIdProvider.getProjectId());
+		System.out.println("============= project number = " + project.getProjectNumber());
+
+		this.setAudience(String.format(
```
artembilan: No `this.` prefix on method calls.
Plus also when we call methods from ctors they have to be `final`.
There is some SonarQube rule, not sure why it doesn't appear here as a smell...

elefeint:
Thanks! I have not done a polish pass over style yet; I'll do it after adding tests.
Thank you for the tip on static beans; I was trying to figure out how to get it to behave properly when it is in Core, but ended up moving it to Iap autoconfig.

elefeint:
@artembilan Now it's in a good state to review.

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/core/environment/OnGcpEnvironmentConditionTests.java
```diff
@@ -0,0 +1,110 @@
...
+import org.springframework.context.annotation.ConditionContext;
+import org.springframework.core.type.AnnotatedTypeMetadata;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+import static org.mockito.Mockito.when;
+
+/**
+ * @author Elena Felder
```
artembilan: Missed `@since`

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/security/IapAuthenticationAutoConfigurationTests.java
```diff
@@ -44,15 +58,15 @@
 
 /**
  * @author Elena Felder
- * @since 1.1
```
artembilan: Why remove?

spring-cloud-gcp-core/src/main/java/org/springframework/cloud/gcp/core/GcpEnvironment.java
```diff
@@ -0,0 +1,76 @@
...
+ * Enumeration of valid individual GCP environments.
+ *
+ * <p>Also includes meta-members for convenient evaluation of environment groups, such as {@code ANY_APP_ENGINE}.
+ *
+ * @author Elena Felder
+ *
+ * @since 1.1
+ */
+public enum GcpEnvironment {
+	/**
```
artembilan: Missed blank line before class member.

spring-cloud-gcp-core/src/main/java/org/springframework/cloud/gcp/core/GcpEnvironment.java
```diff
@@ -0,0 +1,76 @@
...
+	 * Matches both, AppEngine Flexible and AppEngine Standard.
+	 */
+	ANY_APP_ENGINE(() -> APP_ENGINE_FLEXIBLE.matches() || APP_ENGINE_STANDARD.matches()),
+
+	/**
+	 * Matches both, Compute Engine and Kubernetes Engine.
+	 */
+	ANY_CONTAINER(() ->  KUBERNETES_ENGINE.matches() || COMPUTE_ENGINE.matches());
+
+	private BooleanSupplier matchCondition;
```
artembilan: I guess this has to be `final`

spring-cloud-gcp-core/src/main/java/org/springframework/cloud/gcp/core/GcpEnvironmentProvider.java
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+ * Interface for determining whether the current GCP environment matches the passed in parameter.
+ *
+ * @author Elena Felder
+ *
+ * @since 1.1
+ */
+public interface GcpEnvironmentProvider {
+	boolean isCurrentEnvironment(GcpEnvironment candidate);
```
artembilan: Missed blank lines around the class member.

spring-cloud-gcp-docs/src/main/asciidoc/security-iap.adoc
```diff
@@ -7,11 +7,15 @@ https://cloud.google.com/iap/[Cloud Identity-Aware Proxy (IAP)] provides a secur
...
+The following claims are validated automatically:
 
- * Pull in the `org.springframework.cloud:spring-cloud-gcp-starter-security-iap` dependency.
- * Set `spring.cloud.gcp.security.iap.enabled` property to `true`.
- * If you create a custom {spring-security-javadoc}config/annotation/web/configuration/WebSecurityConfigurerAdapter.html[`WebSecurityConfigurerAdapter`], enable extracting user identity by adding `.oauth2ResourceServer().jwt()` configuration to the {spring-security-javadoc}config/annotation/web/builders/HttpSecurity.html[`HttpSecurity`] object.
+ * Issue time
+ * Expiration time
+ * Issuer
+
+Additionally, Audience (`"aud"`) validation is enabled when the application is running on AppEngine/ComputeEngine/Kubernetes, or when a custom audience is provided through `spring.cloud.gcp.security.iap.audience` property. The custom property overrides automatic environment-based checks.
```
artembilan: Sentence per line rule.

spring-cloud-gcp-autoconfigure/pom.xml
```diff
@@ -200,6 +200,12 @@
         </dependency>
 
         <!-- Security -->
+        <dependency>
+            <groupId>org.springframework.cloud</groupId>
+            <artifactId>spring-cloud-gcp-security-iap</artifactId>
+            <optional>true</optional>
+        </dependency>
+
         <dependency>
```
elefeint: remove now-duplicate dependencies

spring-cloud-gcp-security-iap/pom.xml
```diff
@@ -0,0 +1,41 @@
...
+	<dependencies>
+		<dependency>
+			<groupId>org.springframework.security</groupId>
+			<artifactId>spring-security-oauth2-core</artifactId>
+		</dependency>
+
+		<dependency>
+			<groupId>org.springframework.security</groupId>
+			<artifactId>spring-security-oauth2-jose</artifactId>
+			<optional>true</optional>
```
elefeint: Remove optional.

spring-cloud-gcp-security-iap/pom.xml
```diff
@@ -0,0 +1,41 @@
...
+	<artifactId>spring-cloud-gcp-security-iap</artifactId>
+	<name>Spring Cloud GCP IAP Authentication Module</name>
+	<description>Spring Cloud GCP IAP Authentication Module</description>
+
+	<dependencies>
+		<dependency>
+			<groupId>org.springframework.security</groupId>
+			<artifactId>spring-security-oauth2-core</artifactId>
+		</dependency>
+
```
elefeint: newline

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/security/IapAuthenticationAutoConfiguration.java
```diff
@@ -16,56 +16,120 @@
...
+import org.springframework.security.oauth2.jwt.JwtIssuerValidator;
+import org.springframework.security.oauth2.jwt.JwtTimestampValidator;
 import org.springframework.security.oauth2.jwt.NimbusJwtDecoderJwkSupport;
 import org.springframework.security.oauth2.server.resource.web.BearerTokenResolver;
 
 /**
  * Autoconfiguration for extracting pre-authenticated user identity from Google Cloud IAP header.
  *
- * <p>The following conditions must be present for this configuration to take effect:
+ * <p>Provides:
```
elefeint: Link to GCP doc

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/security/IapAuthenticationAutoConfiguration.java
```diff
@@ -16,56 +16,120 @@
...
+@ConditionalOnClass({AudienceValidator.class})
 @AutoConfigureBefore(OAuth2ResourceServerAutoConfiguration.class)
+@AutoConfigureAfter(GcpContextAutoConfiguration.class)
 @EnableConfigurationProperties(IapAuthenticationProperties.class)
 public class IapAuthenticationAutoConfiguration {
 
 	@Bean
 	@ConditionalOnMissingBean
-	public JwtDecoder iapJwtDecoder(IapAuthenticationProperties properties) {
+	public JwtDecoder iapJwtDecoder(IapAuthenticationProperties properties, GcpProjectIdProvider projectIdProvider,
```
elefeint: Remove unneeded dependency.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/security/IapAuthenticationAutoConfiguration.java
```diff
@@ -16,56 +16,120 @@
...
+			@Qualifier("jwtValidators") List<OAuth2TokenValidator<Jwt>> validators) {
+
 		NimbusJwtDecoderJwkSupport jwkSupport
 				= new NimbusJwtDecoderJwkSupport(properties.getRegistry(), properties.getAlgorithm());
-		jwkSupport.setJwtValidator(JwtValidators.createDefaultWithIssuer(properties.getIssuer()));
+		jwkSupport.setJwtValidator(new DelegatingOAuth2TokenValidator(validators));
```
elefeint: Inject DelegatingOAuth2TokenValidator instead of list.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/security/IapAuthenticationAutoConfiguration.java
```diff
@@ -16,56 +16,120 @@
...
+
 		return jwkSupport;
 	}
 
+	@Bean
+	@ConditionalOnMissingBean(name = "jwtValidators")
```
elefeint: make name more specific

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/security/IapAuthenticationAutoConfiguration.java
```diff
@@ -16,56 +16,120 @@
...
+	List<OAuth2TokenValidator<Jwt>> jwtValidators(IapAuthenticationProperties properties,
+			ObjectProvider<AudienceValidator> audienceVerifier) {
+		List<OAuth2TokenValidator<Jwt>> validators = new ArrayList();
+		validators.add(new JwtTimestampValidator());
+		validators.add(new JwtIssuerValidator(properties.getIssuer()));
+		audienceVerifier.ifAvailable(validators::add);
+
+		return validators;
+	}
+
```
elefeint: space

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/security/IapAuthenticationAutoConfiguration.java
```diff
@@ -16,56 +16,120 @@
...
+
+	@Bean
+	@ConditionalOnMissingBean
+	@ConditionalOnProperty ("spring.cloud.gcp.security.iap.audience")
```
elefeint: space

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/security/IapAuthenticationAutoConfiguration.java
```diff
@@ -16,56 +16,120 @@
...
+	@ConditionalOnMissingBean
+	@ConditionalOnGcpEnvironment(GcpEnvironment.ANY_APP_ENGINE)
+	AudienceValidator appEngineBasedAudienceValidator(GcpProjectIdProvider projectIdProvider,
+			ResourceManager resourceManager) {
+		return new AppEngineAudienceValidator(projectIdProvider, resourceManager);
+	}
+
+	@Bean
+	@ConditionalOnMissingBean
+	@ConditionalOnGcpEnvironment(GcpEnvironment.ANY_CONTAINER)
```
elefeint: GKE_OR_GCE

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/core/GcpContextAutoConfiguration.java
```diff
@@ -72,4 +79,35 @@ public GcpProjectIdProvider gcpProjectIdProvider() {
...
 	}
+
+	/**
+	 * Provides default implementation for determining GCP environment.
+	 * Can be overridden to avoid interacting with real environment.
+	 * @since 1.1
+	 */
+	@Bean
+	@ConditionalOnMissingBean
+	public static GcpEnvironmentProvider gcpEnvironmentProvider() {
```
elefeint: In core refdoc, mention both EnvironmentProvider,  ResourceManager and MetadataProvider.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/core/AppEngineCondition.java
```diff
@@ -23,7 +23,7 @@
 import org.springframework.util.StringUtils;
 
 /**
- * Determine whether this is running inside of an App Engine environment by looking into
+ * Determine whether this is running inside of an App Engine Standard environment by looking into
```
elefeint: File issue to remove this condition.

spring-cloud-gcp-core/src/main/java/org/springframework/cloud/gcp/core/DefaultGcpEnvironmentProvider.java
```diff
@@ -0,0 +1,35 @@
...
+ * <p>Delegates the decision to {@link GcpEnvironment}.
+ *
+ * @author Elena Felder
+ *
+ * @since 1.1
+ */
+public class DefaultGcpEnvironmentProvider implements GcpEnvironmentProvider {
+
+	@Override
+	public boolean isCurrentEnvironment(GcpEnvironment candidate) {
```
elefeint: replace or add getCurrentEnvironment; return enum.

elefeint: isOneOf() or isGceOrGke()

spring-cloud-gcp-core/src/main/java/org/springframework/cloud/gcp/core/GcpEnvironment.java
```diff
@@ -0,0 +1,76 @@
...
+ *
+ * @author Elena Felder
+ *
+ * @since 1.1
+ */
+public enum GcpEnvironment {
+	/**
+	 * Matches Kubernetes instances.
+	 */
+	KUBERNETES_ENGINE(() -> System.getenv("KUBERNETES_SERVICE_HOST") != null),
```
elefeint: this will move to provider; allows fallthrough; guarantees only one environment selected

spring-cloud-gcp-samples/spring-cloud-gcp-security-iap-sample/src/main/resources/application.properties
```diff
@@ -1 +1 @@
-spring.cloud.gcp.security.iap.enabled=true
+
```
elefeint: delete file

spring-cloud-gcp-security-iap/src/main/java/org/springframework/cloud/gcp/security/iap/AudienceValidator.java
```diff
@@ -0,0 +1,74 @@
...
+	private static final Log LOGGER = LogFactory.getLog(AudienceValidator.class);
+
+	private static final OAuth2Error INVALID_AUDIENCE = new OAuth2Error(
+			OAuth2ErrorCodes.INVALID_REQUEST,
+					"This aud claim is not equal to the configured audience",
+							"https://tools.ietf.org/html/rfc6750#section-3.1");
+
+	private String audience;
+
+	public AudienceValidator(String audience) {
```
elefeint: string -> AudienceProvider; invert; the two special purpose validators become providers.

spring-cloud-gcp-security-iap/src/main/java/org/springframework/cloud/gcp/security/iap/AudienceValidator.java
```diff
@@ -0,0 +1,74 @@
...
+	protected final void setAudience(String audience) {
+		Assert.notNull(audience, "Audience cannot be null");
+		this.audience = audience;
+	}
+
+	@Override
+	public OAuth2TokenValidatorResult validate(Jwt t) {
+		if (t.getAudience() != null && t.getAudience().contains(this.audience)) {
+			return OAuth2TokenValidatorResult.success();
+		}
```
elefeint: add else for ease of reading
set to warn level; this is not a normal, expected scenario

spring-cloud-gcp-security-iap/src/main/java/org/springframework/cloud/gcp/security/iap/ComputeEngineAudienceValidator.java
```diff
@@ -0,0 +1,49 @@
...
+public class ComputeEngineAudienceValidator extends AudienceValidator {
+
+	public ComputeEngineAudienceValidator(GcpProjectIdProvider projectIdProvider, ResourceManager resourceManager,
+			MetadataProvider metadataProvider) {
+		Assert.notNull(projectIdProvider, "GcpProjectIdProvider cannot be null.");
+		Assert.notNull(resourceManager, "ResourceManager cannot be null.");
+		Assert.notNull(metadataProvider, "MetadataProvider cannot be null.");
+
+		Project project = resourceManager.get(projectIdProvider.getProjectId());
+		String serviceId = metadataProvider.getAttribute("id");
```
elefeint: Assert all nulls

spring-cloud-gcp-security-iap/src/main/java/org/springframework/cloud/gcp/security/iap/AppEngineAudienceValidator.java
```diff
@@ -0,0 +1,44 @@
...
+public class AppEngineAudienceValidator extends AudienceValidator {
+
+	public AppEngineAudienceValidator(GcpProjectIdProvider projectIdProvider, ResourceManager resourceManager) {
+		Assert.notNull(projectIdProvider, "GcpProjectIdProvider cannot be null.");
+		Assert.notNull(resourceManager, "ResourceManager cannot be null.");
+
+		Project project = resourceManager.get(projectIdProvider.getProjectId());
+
+		setAudience(String.format(
+				"/projects/%s/apps/%s", project.getProjectNumber(), projectIdProvider.getProjectId()));
```
elefeint: make patterns constant

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/core/environment/OnGcpEnvironmentCondition.java
```diff
@@ -0,0 +1,57 @@
...
+		Map<String, Object> attributes = metadata.getAnnotationAttributes(ConditionalOnGcpEnvironment.class.getName());
+		GcpEnvironment targetEnvironment = (GcpEnvironment) attributes.get("value");
+
+		try {
+			GcpEnvironmentProvider environmentProvider = context.getBeanFactory().getBean(GcpEnvironmentProvider.class);
+			if (!environmentProvider.isCurrentEnvironment(targetEnvironment)) {
+				return new ConditionOutcome(false, "Application is not running on " + targetEnvironment);
+			}
+		}
+		catch (NoSuchBeanDefinitionException ex) {
```
elefeint: let exception bubble up

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/security/IapAuthenticationAutoConfigurationTests.java
elefeint: I noticed that the other tests don't add it. I'll put them back!

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/security/IapAuthenticationAutoConfiguration.java
```diff
@@ -16,56 +16,120 @@
...
  * Spring Boot's default {@code OAuth2ResourceServerWebSecurityConfiguration} will add this customization.
  *
  * @author Elena Felder
+ *
  * @since 1.1
  */
 @Configuration
-@ConditionalOnProperty("spring.cloud.gcp.security.iap.enabled")
-@ConditionalOnClass({NimbusJwtDecoderJwkSupport.class})
+@ConditionalOnProperty(value = "spring.cloud.gcp.security.iap.enabled", matchIfMissing = true)
```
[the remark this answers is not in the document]
elefeint: We now have a project to depend on, so we can match based on a real dependency! Don't need workarounds anymore, which is nice.

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/security/IapAuthenticationAutoConfigurationTests.java
artembilan: We don't use `@since` on those classes which were written in version `1.0` since the project start.

elefeint:
@artembilan Uh, you trust me more than I trust me 😅

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/core/AppEngineCondition.java
elefeint: Filed #1254

spring-cloud-gcp-core/src/main/java/org/springframework/cloud/gcp/core/GcpEnvironment.java
artembilan: Missed this one. 
Or do you think we can't make it `final` ?

spring-cloud-gcp-security-iap/src/main/java/org/springframework/cloud/gcp/security/iap/AppEngineAudienceValidator.java
```diff
@@ -32,13 +32,17 @@
  */
 public class AppEngineAudienceValidator extends AudienceValidator {
 
+	public static final String AUDIENCE_FORMAT = "/projects/%s/apps/%s";
```
artembilan: Does it make sense to make this as `public`?

Otherwise you need to add a Javadoc on it

spring-cloud-gcp-core/src/main/java/org/springframework/cloud/gcp/core/GcpEnvironment.java
elefeint: Removing it completely in the next push because the logic for determining environment is moving into DefaultGcpEnvironmentProvider.

spring-cloud-gcp-core/src/main/java/org/springframework/cloud/gcp/core/GcpEnvironment.java
```diff
@@ -35,43 +30,64 @@
...
-	public boolean matches() {
-		return this.matchCondition.getAsBoolean();
+	public boolean matchesSimpleEnvironment(GcpEnvironment env) {
+		validateSimpleEnvironment(env);
+		return this == env;
+	}
+
+	private static void validateSimpleEnvironment(GcpEnvironment env) {
+		if (env == ANY_APP_ENGINE || env == GKE_OR_GCE) {
+			throw new IllegalArgumentException("Composite environments are not allowed here.");
```
artembilan: I don't think `here` is so informative word.
I guess we need to be more specific to explain end-user what is wrong.

Thanks

spring-cloud-gcp-docs/src/main/asciidoc/core.adoc
```diff
@@ -135,6 +135,35 @@ and append the additional scopes you need to add.
...
+
+=== ResourceManager
+
+A default instance of Google Cloud `ResourceManager` is auto-configured by the Spring Cloud GCP starter.
+
+
+=== Environment
+
+`MetadataProvider` is another functional interface that provides access to Google Cloud Metadata Server attributes.
+The implementation auto-configured by the Spring Cloud GCP starter is a thin wrapper around static `MetadataConfig` utility, which is mainly helpful for testability.
```
artembilan: Why do we need to pollute production code with auto-configured beans for testing?

Can't we just live without it and really use only in tests if that?

spring-cloud-gcp-security-iap/src/main/java/org/springframework/cloud/gcp/security/iap/AppEngineAudienceSupplier.java
```diff
@@ -30,19 +32,33 @@
  *
  * @since 1.1
  */
-public class AppEngineAudienceValidator extends AudienceValidator {
+public class AppEngineAudienceSupplier implements Supplier<String> {
```
artembilan: This is not good. There is a Spring Cloud Function project (https://cloud.spring.io/spring-cloud-function) which detects all the `java.function` interfaces and do some stuff around them. Therefore this one being auto-configured is going to be at least a noise in the Function Registry. I don't tell already that there might be some endpoint exposure or some other unexpected behavior.

spring-cloud-gcp-security-iap/src/main/java/org/springframework/cloud/gcp/security/iap/AppEngineAudienceValidator.java
artembilan: One more time: why do we need to have it as `public`? Doesn't look like some useful API for end-users of the framework.

spring-cloud-gcp-security-iap/src/main/java/org/springframework/cloud/gcp/security/iap/AudienceValidator.java
```diff
@@ -44,19 +47,10 @@
...
-	 * Allows subclasses to delay setting the audience value after the object is constructed.
-	 */
-	protected AudienceValidator() { }
-
-	protected final void setAudience(String audience) {
-		Assert.notNull(audience, "Audience cannot be null");
-		this.audience = audience;
+	public AudienceValidator(Supplier<String> audienceSupplier) {
+		Assert.notNull(audienceSupplier, "Audience cannot be null");
+		this.audience = audienceSupplier.get();
```
artembilan: I don't think it is a good idea to evaluate supplier exactly here in the ctor.
On the other hand if we still do this, what is the point then request a supplier wrapper if we could just ask end-users to inject the target `audience` result?

If we still have some hard logic about  `audience` evaluation, I would suggest to introduce a new `AudienceProvider` interface instead of using `Supplier` to not confuse Spring Cloud Function (see my comment above), but still don't evaluate the target result here in the ctor: some low level resources interactions may present in the provider implementation.

spring-cloud-gcp-security-iap/src/main/java/org/springframework/cloud/gcp/security/iap/AppEngineAudienceValidator.java
elefeint: Ah, I took your original comment to mean that these constants should either be private or have useful javadoc. It's not a problem, I'll make them private.

spring-cloud-gcp-docs/src/main/asciidoc/core.adoc
elefeint: The client library utility for accessing metadata server is a static utility method on MetadataConfig class. I prefer to stay away from using static utilities, since they make the code using them inflexible and hard to test, so I made a wrapper that could be injected.

I see your point that a default could be instantiated and made overridable for testing only. The reason I added three new autoconfigured beans to `GcpContextAutoConfiguration` is for consistency, since these wrappers behave similarly to `GcpProjectIdProvider`. But if they are not useful for other spring-cloud-gcp projects, I can remove them in favor of overridable defaults.

spring-cloud-gcp-security-iap/src/main/java/org/springframework/cloud/gcp/security/iap/AppEngineAudienceSupplier.java
elefeint: The suppliers are not exposed as autoconfigured beans; they are directly instantiated and passed to AudienceValidator objects (which are exposed as beans in autoconfiguration) in `IapAuthenticationAutoConfiguration`.

I guess I could create a special-purpose interface for AudienceProvider instead of implementing Supplier<String> directly. Would that help?

spring-cloud-gcp-security-iap/src/main/java/org/springframework/cloud/gcp/security/iap/AudienceValidator.java
elefeint: aaaand I should have read all comments first. Thanks!

elefeint: Hmm, now that I think about it, it seems better to only evaluate the audience string in the constructor and set it forever. The constructor runs once, but JWT token validation could happen at high volume. It seems wasteful to keep re-request information about the project and keep concatenating what is essentially a static string over the lifetime of the app.

An alternative would be to cache the string in each AudienceProvider, so that it's only evaluated once.

artembilan: You can cache it in this `AudienceValidator` implementing the evaluation logic in the `afterPropertiesSet()`. See `InitializingBean`.

spring-cloud-gcp-docs/src/main/asciidoc/core.adoc
artembilan: You can use `@SpyBean` or even `@MockBean` for the `ComputeEngineAudienceProvider` and `AppEngineAudienceProvider`: https://docs.spring.io/spring-boot/docs/2.1.0.RELEASE/reference/htmlsingle/#boot-features-testing-spring-boot-applications-mocking-beans.


Otherwise I guess you can just override those beans in the test `@Configuration` with `@Primary`.

elefeint: The behavior change would need to happen after the specific AudienceProvider bean is created, but before AudienceValidator is initialized because `SpyPostProcessor` overrides `postProcessAfterInitialization`, so I think it would run too late in the lifecycle.
Both overriding in @Configuration and using @MockBean would change behavior of autoconfiguration under test.

Do you mind if I leave the current implementation?

artembilan: OK. If there is no other choice...

elefeint:
@artembilan @meltsufin The IAP Audience validator PR is ready for another round of reviews.

spring-cloud-gcp-security-iap/src/main/java/org/springframework/cloud/gcp/security/iap/ComputeEngineAudienceProvider.java
```diff
@@ -0,0 +1,70 @@
...
+ *
+ * @author Elena Felder
+ *
+ * @since 1.1
+ */
+public class ComputeEngineAudienceProvider implements AudienceProvider {
+
+	private static final String AUDIENCE_FORMAT = "/projects/%s/global/backendServices/%s";
+
+	private GcpProjectIdProvider projectIdProvider;
```
artembilan: Must be `final`

spring-cloud-gcp-security-iap/src/main/java/org/springframework/cloud/gcp/security/iap/ComputeEngineAudienceProvider.java
```diff
@@ -0,0 +1,70 @@
...
+		Assert.notNull(project.getProjectNumber(), "Project Number expected not to be null.");
+
+		String serviceId = this.metadataProvider.getAttribute("id");
+		Assert.notNull(project, "Service ID expected not to be null.");
+
+		return String.format(AUDIENCE_FORMAT, project.getProjectNumber(), serviceId);
+	}
+
+	public void setResourceManager(ResourceManager resourceManager) {
+		this.resourceManager = resourceManager;
```
artembilan: `Assert.notNull(resourceManager)`

spring-cloud-gcp-security-iap/src/main/java/org/springframework/cloud/gcp/security/iap/ComputeEngineAudienceProvider.java
```diff
@@ -0,0 +1,70 @@
...
+	}
+
+	public void setMetadataProvider(MetadataProvider metadataProvider) {
+		this.metadataProvider = metadataProvider;
```
artembilan: `Assert.notNull(resourceManager)`

spring-cloud-gcp-security-iap/src/main/java/org/springframework/cloud/gcp/security/iap/ComputeEngineAudienceProvider.java
```diff
@@ -0,0 +1,70 @@
...
+	}
```
artembilan: Every class member has to be wrapped with blank lines even before the end of class.

spring-cloud-gcp-security-iap/src/test/java/org/springframework/cloud/gcp/security/iap/ComputeEngineAudienceProviderTests.java
```diff
@@ -0,0 +1,82 @@
...
+import org.mockito.junit.MockitoJUnitRunner;
+
+import org.springframework.cloud.gcp.core.GcpProjectIdProvider;
+import org.springframework.cloud.gcp.core.MetadataProvider;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.when;
+
+/**
+ * @author Elena Felder
```
artembilan: Missed `@since 1.1`

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/security/IapAuthenticationAutoConfiguration.java
```diff
@@ -16,59 +16,142 @@
...
+
+	@Bean
+	@ConditionalOnBean(AudienceProvider.class)
+	public AudienceValidator environmentBasedAudienceValidator(AudienceProvider audienceProvider) {
+		return new AudienceValidator(audienceProvider);
+	}
+
+	@Bean
+	@ConditionalOnMissingBean(name = "iapJwtValidators")
+	public List<OAuth2TokenValidator<Jwt>> iapJwtValidators(IapAuthenticationProperties properties,
```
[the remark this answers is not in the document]
elefeint: I did it for testing (DelegatingOAuth2TokenValidator does not have a getter, so testing the intermediate collection was easier), but now I am thinking I'd rather test environment-based AudienceProvider beans separately from testing "given some AudienceValidator, it gets added". The two environment-based tests do too much right now.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/security/IapAuthenticationAutoConfiguration.java
```diff
@@ -16,59 +16,142 @@
...
+	@Bean
+	@ConditionalOnMissingBean
+	@ConditionalOnGcpEnvironment(GcpEnvironment.GKE_OR_GCE)
+	public AudienceProvider computeEngineBasedAudienceProvider(GcpProjectIdProvider projectIdProvider) {
+		return new ComputeEngineAudienceProvider(projectIdProvider);
+	}
+
+	@Bean
+	@ConditionalOnBean(AudienceProvider.class)
+	public AudienceValidator environmentBasedAudienceValidator(AudienceProvider audienceProvider) {
```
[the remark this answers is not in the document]
elefeint: Yes, good catch.
I might be able to suppress AudienceValidator creation when testing the specific audience providers, so that they are never tested together (and so the hard-to-override functionality is never triggered). I'll try that.

spring-cloud-gcp-core/src/main/java/org/springframework/cloud/gcp/core/GcpEnvironment.java
```diff
@@ -0,0 +1,96 @@
...
+	 * Matches instances of Compute Engine that are not also AppEngine Flexible.
+	 */
+	COMPUTE_ENGINE,
+
+	/**
+	 * Matches both, AppEngine Flexible and AppEngine Standard.
+	 */
+	ANY_APP_ENGINE {
+		@Override
+		public boolean matchesSimpleEnvironment(GcpEnvironment env) {
```
meltsufin: This seems a bit confusing. What does "simple" mean? I think that instead of `ANY_APP_ENGINE` and `GKE_OR_GCE`, it might be better to just have methods on the enum like `isAppEngine()` and `isGkeOrGce()`.

elefeint: I agree that it's confusing. But what would the rule look like then? Right now, it's
```
@ConditionalOnGcpEnvironment(GcpEnvironment.ANY_APP_ENGINE)
```

meltsufin: That's a good point, but enums just don't lend themselves well to overlapping sets.
Can the annotation instead allow an array? 
```
@ConditionalOnGcpEnvironment([GcpEnvironment.APP_ENGINE_STANDARD, GcpEnvironment.APP_ENGINE_STANDARD])
```

PR #1262, opened 2018-11-28
Add Vision API Template

dzou:
Add a Spring Template for the Cloud Vision APIs.

### DIFF
--- a/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/vision/CloudVisionAutoConfiguration.java
+++ b/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/vision/CloudVisionAutoConfiguration.java
@@ -28,6 +28,7 @@ import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.cloud.gcp.core.DefaultCredentialsProvider;
 import org.springframework.cloud.gcp.core.UsageTrackingHeaderProvider;
+import org.springframework.cloud.gcp.vision.CloudVisionTemplate;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 
@@ -78,5 +79,11 @@ public class CloudVisionAutoConfiguration {
 				.build();
 
 		return ImageAnnotatorClient.create(clientSettings);
+	}
+
+	@Bean
+	@ConditionalOnMissingBean
+	public CloudVisionTemplate cloudVisionTemplate(ImageAnnotatorClient imageAnnotatorClient) {
+		return new CloudVisionTemplate(imageAnnotatorClient);
 	}
 }
--- a/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/vision/CloudVisionAutoConfiguration.java
+++ b/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/vision/CloudVisionAutoConfiguration.java
@@ -40,7 +40,7 @@ import org.springframework.context.annotation.Configuration;
  */
 @Configuration
 @EnableConfigurationProperties(CloudVisionProperties.class)
-@ConditionalOnClass(ImageAnnotatorClient.class)
+@ConditionalOnClass(CloudVisionTemplate.class)
 @ConditionalOnProperty(value = "spring.cloud.gcp.vision.enabled", matchIfMissing = true)
 public class CloudVisionAutoConfiguration {
 
@@ -86,4 +86,5 @@ public class CloudVisionAutoConfiguration {
 	public CloudVisionTemplate cloudVisionTemplate(ImageAnnotatorClient imageAnnotatorClient) {
 		return new CloudVisionTemplate(imageAnnotatorClient);
 	}
+
 }
--- a/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/vision/CloudVisionAutoConfiguration.java
+++ b/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/vision/CloudVisionAutoConfiguration.java
@@ -66,7 +66,6 @@ public class CloudVisionAutoConfiguration {
 	 * honored by Spring bean lifecycle.
 	 * <p>Most of the Google Cloud API clients are thread-safe heavy objects. I.e., it's better
 	 * to produce a singleton and re-using the client object for multiple requests.
-	 *
 	 * @return A Cloud Vision API client
 	 * @throws IOException if an exception occurs creating the ImageAnnotatorClient
 	 */
--- a/spring-cloud-gcp-samples/spring-cloud-gcp-vision-api-sample/pom.xml
+++ b/spring-cloud-gcp-samples/spring-cloud-gcp-vision-api-sample/pom.xml
@@ -18,6 +18,7 @@
 			<artifactId>spring-cloud-gcp-starter-vision</artifactId>
 		</dependency>
 
+
 		<dependency>
 			<groupId>org.springframework.boot</groupId>
 			<artifactId>spring-boot-starter-web</artifactId>
--- a/spring-cloud-gcp-samples/spring-cloud-gcp-vision-api-sample/pom.xml
+++ b/spring-cloud-gcp-samples/spring-cloud-gcp-vision-api-sample/pom.xml
@@ -18,7 +18,6 @@
 			<artifactId>spring-cloud-gcp-starter-vision</artifactId>
 		</dependency>
 
-
 		<dependency>
 			<groupId>org.springframework.boot</groupId>
 			<artifactId>spring-boot-starter-web</artifactId>
--- a/spring-cloud-gcp-samples/spring-cloud-gcp-vision-api-sample/src/main/java/com/example/VisionController.java
+++ b/spring-cloud-gcp-samples/spring-cloud-gcp-vision-api-sample/src/main/java/com/example/VisionController.java
@@ -36,8 +36,7 @@ import org.springframework.web.servlet.ModelAndView;
  * Code sample that shows how Spring Cloud GCP can be leveraged to use Google Cloud Client
  * Libraries.
  *
- * <p>
- * This uses the Cloud Vision API with the {@link ImageAnnotatorClient}, which is
+ * <p>This uses the Cloud Vision API with the {@link ImageAnnotatorClient}, which is
  * configured and provided by the spring-cloud-gcp-starter-vision module.
  *
  * @author João André Martins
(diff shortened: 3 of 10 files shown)


spring-cloud-gcp-samples/spring-cloud-gcp-vision-api-sample/pom.xml
```diff
@@ -18,6 +18,7 @@
 			<artifactId>spring-cloud-gcp-starter-vision</artifactId>
 		</dependency>
 
+
```
artembilan: Remove this line, so we don't have this extra file for review here

spring-cloud-gcp-samples/spring-cloud-gcp-vision-api-sample/src/main/java/com/example/VisionController.java
```diff
@@ -65,30 +59,15 @@
...
+		AnnotateImageResponse response = cloudVisionTemplate.analyzeImage(
+				resourceLoader.getResource(imageUrl), Type.LABEL_DETECTION);
 
-			ImmutableMap.Builder<String, Float> annotations = ImmutableMap.builder();
-			for (EntityAnnotation annotation : response.getLabelAnnotationsList()) {
-				annotations.put(annotation.getDescription(), annotation.getScore());
-			}
-			map.addAttribute("annotations", annotations.build());
+		ImmutableMap.Builder<String, Float> annotations = ImmutableMap.builder();
+		for (EntityAnnotation annotation : response.getLabelAnnotationsList()) {
```
artembilan: You can process this with the Java `Stream` and produce `collect(Collectors.toMap())`

spring-cloud-gcp-vision/src/main/java/org/springframework/cloud/gcp/vision/CloudVisionTemplate.java
```diff
@@ -0,0 +1,107 @@
+/*
+ *  Copyright 2018 original author or authors.
```
artembilan: We need to have a fist year as well: the date a project has been started.

```suggestion
 *  Copyright 2017-2018 original author or authors.
```

spring-cloud-gcp-vision/src/main/java/org/springframework/cloud/gcp/vision/CloudVisionTemplate.java
```diff
@@ -0,0 +1,107 @@
...
+ * @author Daniel Zou
+ *
+ * @since 1.1
+ */
+public class CloudVisionTemplate {
+
+	private final ImageAnnotatorClient imageAnnotatorClient;
+
+	public CloudVisionTemplate(ImageAnnotatorClient imageAnnotatorClient) {
+		this.imageAnnotatorClient = imageAnnotatorClient;
```
artembilan: `Assert.notNull(imageAnnotatorClient)`

spring-cloud-gcp-vision/src/main/java/org/springframework/cloud/gcp/vision/CloudVisionTemplate.java
```diff
@@ -0,0 +1,107 @@
...
+	}
+
+	/**
+	 * Analyze an image and extract the feature of the image specified by {@code featureType}.
+	 *
```
artembilan: No blank lines in the method Javadocs.

spring-cloud-gcp-vision/src/main/java/org/springframework/cloud/gcp/vision/CloudVisionTemplate.java
```diff
@@ -0,0 +1,107 @@
...
+	 * <p>
```
artembilan: The `<p>` must be a first symbol in the paragraph. Only in method Javadocs: https://github.com/spring-projects/spring-framework/wiki/Code-Style#javadoc-formatting

spring-cloud-gcp-vision/src/main/java/org/springframework/cloud/gcp/vision/CloudVisionTemplate.java
```diff
@@ -0,0 +1,107 @@
...
+	 * A feature describes the kind of Cloud Vision analysis one wishes to perform on an
+	 * image, such as text detection, image labelling, facial detection, etc. A full list of
+	 * feature types can be found in {@link Feature.Type}.
+	 *
+	 * @param imageResource the image one wishes to analyze. The Cloud Vision APIs support
+	 *     image formats described here: https://cloud.google.com/vision/docs/supported-files
+	 * @param featureTypes the types of image analysis to perform on the image.
+	 * @return the results of image analyses
+	 */
+	public AnnotateImageResponse analyzeImage(Resource imageResource, Set<Feature.Type> featureTypes)
```
artembilan: Why can't we have a single method with vararg?

spring-cloud-gcp-vision/src/main/java/org/springframework/cloud/gcp/vision/CloudVisionTemplate.java
```diff
@@ -0,0 +1,107 @@
...
+				.addRequests(
+						AnnotateImageRequest.newBuilder()
+								.addAllFeatures(featureList)
+								.setImage(image))
+				.build();
+
+		BatchAnnotateImagesResponse batchResponse = imageAnnotatorClient.batchAnnotateImages(request);
+		List<AnnotateImageResponse> annotateImageResponses = batchResponse.getResponsesList();
+
+		return annotateImageResponses.get(0);
```
artembilan: Why do you get only the first result? How about all others?
Can't that `getResponsesList()` be empty?

And what is the point to perform here a `BatchAnnotateImagesRequest` if we have only a single `image`?

spring-cloud-gcp-vision/src/main/java/org/springframework/cloud/gcp/vision/CloudVisionTemplate.java
```diff
@@ -0,0 +1,107 @@
...
+	}
+}
```
artembilan: All class members have to be wrapped to blank lines.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/vision/CloudVisionAutoConfiguration.java
```diff
@@ -79,4 +80,10 @@ public ImageAnnotatorClient imageAnnotatorClient() throws IOException {
 
 		return ImageAnnotatorClient.create(clientSettings);
 	}
+
+	@Bean
+	@ConditionalOnMissingBean
+	public CloudVisionTemplate cloudVisionTemplate(ImageAnnotatorClient imageAnnotatorClient) {
```
meltsufin: The whole auto-configuration should probably now be conditional on `CloudVisionTemplate.class` rather than `ImageAnnotatorClient.class`.

spring-cloud-gcp-vision/src/main/java/org/springframework/cloud/gcp/vision/CloudVisionTemplate.java
meltsufin: You probably do want to verify that there is only one, but that's what is expected because only one image was submitted in the batch.

spring-cloud-gcp-vision/src/main/java/org/springframework/cloud/gcp/vision/CloudVisionTemplate.java
meltsufin: Instead of `Set`, does it make sense to use varargs?
```java
public AnnotateImageResponse analyzeImage(Resource imageResource, Feature.Type ... featureTypes)
```

dzou: Yes I am open to this. Right now this is just a sketch of the template API to be used as a point for discussion as we figure out what the template will look like. I will raise this point in the next discussion we have.

spring-cloud-gcp-vision/src/main/java/org/springframework/cloud/gcp/vision/CloudVisionTemplate.java
dzou: > Why do you get only the first result? How about all others? Can't that getResponsesList() be empty?

So this method will pass in exactly 1 image into the Cloud Vision APIs. Then, in the corresponding batch response, we are guaranteed to receive 1 response in the batch; there is a 1-to-1 correspondence between images you send to the service and your response objects you receive in the batch. If the image analysis fails for our image, we will still receive a batch of responses with size = 1; the response inside that batch will indicate the error that occurred for processing that image.

> And what is the point to perform here a BatchAnnotateImagesRequest if we have only a single image?

This is due to the design of the CloudVision APIs; it [does not offer a method for processing a Single image](https://cloud.google.com/vision/docs/reference/rpc/google.cloud.vision.v1#google.cloud.vision.v1.ImageAnnotator), only batch. From the perspective of their API, the single image request case collapses into the batch request case.

artembilan: Good. Can we consider then some much higher API instead of this low level `AnnotateImageResponse` and really throw an exception if that one failed?

spring-cloud-gcp-vision/src/main/java/org/springframework/cloud/gcp/vision/CloudVisionTemplate.java
```diff
@@ -46,17 +47,17 @@
 	private final ImageAnnotatorClient imageAnnotatorClient;
 
 	public CloudVisionTemplate(ImageAnnotatorClient imageAnnotatorClient) {
+		Assert.notNull(
+				imageAnnotatorClient,
+				"CloudVisionTemplate must be provided a non-null ImageAnnotatorClient.");
```
artembilan: We typically say shortly: `'imageAnnotatorClient' must not be null.`

dzou:
Sounds good, thanks for the reviews guys. I have updated the docs for cloud vision to now reference the template.

I have noted the following remaining concerns:
- Consider then some much higher-level API
- Checking for error cases: what happens if image analysis fails, response_size != 1

I will do more research to determine what the appropriate way to improve on these area is.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-vision/src/main/resources/META-INF/spring.provides
```diff
@@ -1 +1 @@
-provides: spring-cloud-gcp-starter
+provides: spring-cloud-gcp-starter, spring-cloud-gcp-vision
```
artembilan: According latest Spring Boot such `/META-INF/spring.provides` files can be removed: https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-2.1-Release-Notes#removal-of-springprovides-files

We can do that in the separate PR though, for the whole project.

dzou: Noted. Sounds good.

PR #1282, opened 2018-12-03
Spanner DML query annotation

### LINKED ISSUE #1215: Spanner Annotated SQL Query Method support DML
Cloud Spanner supports DML statements as of 0.70.0 alpha. We should allow `@Query` methods contain DML statements.

ChengyuanZhao:
fixes #1215 
Spanner annotated queries support DML


ChengyuanZhao:
@meltsufin let me make that doc update as part of the larger doc update from the in-person session with Ray (to prevent a stale-reapproval overhead).

PR #1286, opened 2018-12-04
Fix IAP issuer validation tests

elefeint:
With spring-security 5.1.2, JwtIssuerValidator no longer uses getIssuer() method.

### DIFF
--- a/spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/security/IapAuthenticationAutoConfigurationTests.java
+++ b/spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/security/IapAuthenticationAutoConfigurationTests.java
@@ -177,7 +177,6 @@ public class IapAuthenticationAutoConfigurationTests {
 				.run(context -> {
 					DelegatingOAuth2TokenValidator validator
 							= context.getBean("iapJwtDelegatingValidator", DelegatingOAuth2TokenValidator.class);
-
 					assertFalse(validator.validate(mockJwt).hasErrors());
 				});
 	}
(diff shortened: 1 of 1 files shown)


spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/security/IapAuthenticationAutoConfigurationTests.java
```diff
@@ -171,22 +170,24 @@ public void testCustomPropertyOverridesDefault() {
...
+		when(mockJwt.getClaimAsString("iss")).thenReturn("https://cloud.google.com/iap");
 		verify(mockJwt, never()).getAudience();
 
 		this.contextRunner
 				.run(context -> {
 					DelegatingOAuth2TokenValidator validator
 							= context.getBean("iapJwtDelegatingValidator", DelegatingOAuth2TokenValidator.class);
-					assertFalse(validator.validate(mockJwt).hasErrors());
+
+					System.out.println(validator.validate(mockJwt).getErrors());
```
artembilan: No, `System.out` for printing, please, at all, anywhere in the code.

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/security/IapAuthenticationAutoConfigurationTests.java
```diff
@@ -171,22 +170,24 @@ public void testCustomPropertyOverridesDefault() {
...
+				//	assertFalse(validator.validate(mockJwt).hasErrors());
```
artembilan: And this looks like artificial success when we don't have any assertion in test.

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/security/IapAuthenticationAutoConfigurationTests.java
elefeint: // shame cube

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/security/IapAuthenticationAutoConfigurationTests.java
elefeint: Yeah, I should not have commented this out at all.

PR #1288, opened 2018-12-04
Modify SQL sample test to cover MySQL and Postgres

### LINKED ISSUE #1279: Add an integration test for the SQL sample with Postgres
Add an integration test for the SQL sample with Postgres. This would be useful to spot breakages like: #1277.

dzou:
To get additional coverage over Postgres starter, this splits the SQL Sample App Test into 2 tests for MySQL and Postgres.

Fixes #1279.

### DIFF
--- a/spring-cloud-gcp-samples/spring-cloud-gcp-sql-sample/README.adoc
+++ b/spring-cloud-gcp-samples/spring-cloud-gcp-sql-sample/README.adoc
@@ -1,46 +1,43 @@
 = Spring Cloud GCP SQL Sample
 
-This code sample demonstrates how to connect to a Google Cloud SQL instance using either the
-link:../../spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql-mysql[Spring Cloud GCP MySQL Starter]
-or link:../../spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql-postgresql[Spring Cloud GCP PostgreSQL Starter].
+This code sample demonstrates how to connect to a Google Cloud SQL instance using either the link:../../spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql-mysql[Spring Cloud GCP MySQL Starter] or link:../../spring-cloud-gcp-starters/spring-cloud-gcp-starter-sql-postgresql[Spring Cloud GCP PostgreSQL Starter].
 
 You will create an instance, a database within the instance, populate the database and then query it.
 
-You can also explore the GCP MySQL Starter in the
-https://codelabs.developers.google.com/codelabs/cloud-spring-petclinic-cloudsql/index.html[Spring Pet Clinic using Cloud SQL] codelab.
+You can also explore the GCP MySQL Starter in the https://codelabs.developers.google.com/codelabs/cloud-spring-petclinic-cloudsql/index.html[Spring Pet Clinic using Cloud SQL] codelab.
 
 == Setup
 
-1. Follow https://cloud.google.com/sql/docs/mysql/quickstart[these instructions] to set up a Google
-Cloud Project with billing enabled.
+1. Follow https://cloud.google.com/sql/docs/mysql/quickstart[these instructions] to set up a Google Cloud Project with billing enabled.
 
 2. Create a Google Cloud SQL instance from the https://console.cloud.google.com/sql/instances[Google Cloud Console Cloud SQL page].
 You will be given a choice of either 'MySQL' or 'PostgreSQL' as your database engine.
 
-3. Still within the Google Cloud Console SQL page, create a new database in the instance from the
-"Databases" section. If you decided to set a root password for your instance footnoteref:[note, Google Cloud SQL allows for creating MySQL instances without a password, while still remaining safe, as the connection is using Cloud SDK authentication and encrypted communication: see docs on https://cloud.google.com/sql/docs/mysql/sql-proxy[Cloud SQL proxy] and https://cloud.google.com/sql/docs/mysql/connect-external-app#java[the java socketFactory], which is used in spring-cloud-gcp to create the connection!], then don't forget to take note of your root password, as you will need it in the next step!
+3. Still within the Google Cloud Console SQL page, create a new database in the instance from the "Databases" section.
+If you decided to set a root password for your instance footnoteref:[note, Google Cloud SQL allows for creating MySQL instances without a password, while still remaining safe, as the connection is using Cloud SDK authentication and encrypted communication: see docs on https://cloud.google.com/sql/docs/mysql/sql-proxy[Cloud SQL proxy] and https://cloud.google.com/sql/docs/mysql/connect-external-app#java[the java socketFactory], which is used in spring-cloud-gcp to create the connection!], then don't forget to take note of your root password, as you will need it in the next step!
 
 4. Update the pom.xml with the correct Spring Cloud GCP SQL starter dependency.
 Based on your choice of database engine, you must use the correct starter dependency for the project.
 
 - For MySQL backends, use:
-
-  ```
-  <dependency>
-    <groupId>org.springframework.cloud</groupId>
-    <artifactId>spring-cloud-gcp-starter-sql-mysql</artifactId>
-  </dependency>
-  ```
++
+[source, xml]
+----
+<dependency>
+  <groupId>org.springframework.cloud</groupId>
+  <artifactId>spring-cloud-gcp-starter-sql-mysql</artifactId>
+</dependency>
+----
 
 - For PostgreSQL backends, use:
-
-  ```
-  <dependency>
-    <groupId>org.springframework.cloud</groupId>
-    <artifactId>spring-cloud-gcp-starter-sql-postgresql</artifactId>
-  </dependency>
-  ```
-
++
+[source, xml]
+----
+<dependency>
+  <groupId>org.springframework.cloud</groupId>
+  <artifactId>spring-cloud-gcp-starter-sql-postgresql</artifactId>
+</dependency>
+----
 +
 In the current pom.xml, both dependencies are listed as a reference.
 Use the starter dependency corresponding to your database engine and remove the one that does not apply to your application.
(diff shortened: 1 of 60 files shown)


spring-cloud-gcp-samples/spring-cloud-gcp-sql-sample/README.adoc
```diff
@@ -29,7 +54,7 @@ If you set a root password, add the `spring.datasource.password` property with t
 +
 If you would like to use a different user, set the `spring.datasource.username` property appropriately.
 
-5. https://cloud.google.com/sdk/gcloud/reference/auth/application-default/login[If
+6. https://cloud.google.com/sdk/gcloud/reference/auth/application-default/login[If
```
artembilan: If you are here, please, consider to build a proper one sentence per line.

spring-cloud-gcp-samples/spring-cloud-gcp-sql-sample/src/test/java/com/example/CloudSqlApplicationTests.java
```diff
@@ -0,0 +1,61 @@
...
+import org.springframework.cloud.gcp.autoconfigure.sql.DefaultCloudSqlJdbcInfoProvider;
+import org.springframework.cloud.gcp.autoconfigure.sql.GcpCloudSqlProperties;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.test.context.junit4.SpringRunner;
+
+/**
+ * Runs the {@link SqlApplicationTestCase} on a Cloud SQL database backend.
+ *
+ * @author Daniel Zou
```
artembilan: `@since 1.1` ?

spring-cloud-gcp-samples/spring-cloud-gcp-sql-sample/src/test/java/com/example/SqlApplicationTestCase.java
```diff
@@ -21,34 +21,27 @@
...
  *
  * @author Daniel Zou
  */
-@RunWith(SpringRunner.class)
-@SpringBootTest(webEnvironment = WebEnvironment.RANDOM_PORT, classes = { SqlApplication.class })
-public class SqlApplicationTests {
+public abstract class SqlApplicationTestCase {
 
 	@LocalServerPort
 	private int port;
```
artembilan: We don't need this here at all. The `TestRestTemplate` is fully supplied with the based url including port.
In the test-case below you don't need that `String url` as well.
Then you can have it only like this:
```
 this.testRestTemplate.exchange("/getTuples", ...
```

spring-cloud-gcp-samples/spring-cloud-gcp-sql-sample/README.adoc
```diff
@@ -14,13 +14,38 @@ https://codelabs.developers.google.com/codelabs/cloud-spring-petclinic-cloudsql/
...
 3. Still within the Google Cloud Console SQL page, create a new database in the instance from the
 "Databases" section. If you decided to set a root password for your instance footnoteref:[note, Google Cloud SQL allows for creating MySQL instances without a password, while still remaining safe, as the connection is using Cloud SDK authentication and encrypted communication: see docs on https://cloud.google.com/sql/docs/mysql/sql-proxy[Cloud SQL proxy] and https://cloud.google.com/sql/docs/mysql/connect-external-app#java[the java socketFactory], which is used in spring-cloud-gcp to create the connection!], then don't forget to take note of your root password, as you will need it in the next step!
 
-4. Open the link:src/main/resources/application.properties[application.properties] file and replace
+4. Update the pom.xml with the correct Spring Cloud GCP SQL starter dependency.
+Based on your choice of database engine, you must use the correct starter dependency for the project.
+
+- For MySQL backends, use:
+
+  ```
```
meltsufin: Adoc is different from Markdown. Please use adoc formatting.
```suggestion
[source,xml]
----
```

spring-cloud-gcp-samples/spring-cloud-gcp-sql-sample/README.adoc
```diff
@@ -14,13 +14,38 @@ https://codelabs.developers.google.com/codelabs/cloud-spring-petclinic-cloudsql/
...
+  <dependency>
+    <groupId>org.springframework.cloud</groupId>
+    <artifactId>spring-cloud-gcp-starter-sql-mysql</artifactId>
+  </dependency>
+  ```
```
meltsufin:
```suggestion
----
```

spring-cloud-gcp-samples/spring-cloud-gcp-sql-sample/README.adoc
```diff
@@ -14,13 +14,38 @@ https://codelabs.developers.google.com/codelabs/cloud-spring-petclinic-cloudsql/
...
+
+- For PostgreSQL backends, use:
+
+  ```
```
meltsufin: Adoc formatting here as well.

spring-cloud-gcp-samples/spring-cloud-gcp-sql-sample/pom.xml
```diff
@@ -13,10 +13,20 @@
 	<name>Spring Cloud GCP SQL MySQL Starter Code Sample</name>
 
 	<dependencies>
+		<!-- You only need to include the starter dependency of your database engine. -->
```
meltsufin: Maybe explain why both are included in the sample and whether there is any harm in that?

spring-cloud-gcp-samples/spring-cloud-gcp-sql-sample/src/test/java/com/example/CloudSqlApplicationTests.java
```diff
@@ -0,0 +1,61 @@
...
+		})
+public class CloudSqlApplicationTests extends SqlApplicationTestCase {
+
+	@Configuration
+	static class CloudSqlApplicationTestConfiguration {
+		@Bean
+		public CloudSqlJdbcInfoProvider mySqlJdbcInfoProvider(
+				GcpCloudSqlProperties gcpCloudSqlProperties) {
+			CloudSqlJdbcInfoProvider provider = new DefaultCloudSqlJdbcInfoProvider(gcpCloudSqlProperties,
+					DatabaseType.MYSQL);
```
meltsufin: What if we made `DatabaseType` an optional property in `CloudSqlJdbcInfoProvider`? I think it would make this code a lot simpler, and it would enable users to more easily force the database type when both dependencies are present.
cc/ @artembilan

spring-cloud-gcp-samples/spring-cloud-gcp-sql-sample/src/test/java/com/example/PostgreSqlApplicationTests.java
```diff
@@ -0,0 +1,62 @@
...
+import org.springframework.cloud.gcp.autoconfigure.sql.DefaultCloudSqlJdbcInfoProvider;
+import org.springframework.cloud.gcp.autoconfigure.sql.GcpCloudSqlProperties;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.test.context.junit4.SpringRunner;
+
+/**
+ * Runs the {@link SqlApplicationTestCase} on a Postgres database backend.
+ *
+ * @author Daniel Zou
```
meltsufin: `@Since 1.1`

spring-cloud-gcp-samples/spring-cloud-gcp-sql-sample/src/test/java/com/example/CloudSqlApplicationTests.java
```diff
@@ -0,0 +1,61 @@
...
+		webEnvironment = WebEnvironment.RANDOM_PORT,
+		classes = { CloudSqlApplicationTestConfiguration.class, SqlApplication.class },
+		properties = {
+				"spring.cloud.gcp.sql.database-name=code_samples_test_db",
+				"spring.cloud.gcp.sql.instance-connection-name=spring-cloud-gcp-ci:us-central1:testmysql",
+				"spring.datasource.password=test",
+				"spring.datasource.continue-on-error=true",
+				"spring.datasource.initialization-mode=always"
+		})
+public class CloudSqlApplicationTests extends SqlApplicationTestCase {
```
artembilan: I think for consistency this must be called `MySqlApplicationTests`

spring-cloud-gcp-samples/spring-cloud-gcp-sql-sample/src/test/java/com/example/CloudSqlApplicationTests.java
artembilan: I think we really can expose a `dataBaseType` as a configuration property in the `GcpCloudSqlProperties`.
And it really should have an effect in the `GcpCloudSqlAutoConfiguration` when we have both dependencies on classpath.

However looking at this from the target application perspective, I don't think that this is going to have some value: you connect to MySql or PsotgreSql in cloud from your application. Nothing more. So, my position to stay here with what we have so far. Or create MySql and PsotgreSql as separate samples. The point of samples to be as simple as possible with one-stop-shop guide to use in the target project.

dzou: I see your point. I would be open to either keeping it this way (for the sake of reducing duplicate code) or writing a separate sample (for the sake of being simpler). One can argue that this sample is now confusing because the user will have to comment/uncomment the SQL starter from the pom.xml. So I am open to redoing this PR for a separate sample too; let me know what you guys think.

spring-cloud-gcp-samples/spring-cloud-gcp-sql-sample/pom.xml
dzou: So both are included actually just to get the Test to work on both postgres and mysql. From the app's perspective, It turns out that including both means it defaults to using the MySql starter when the app is run.

dzou:
Ok, given both of your comments I'll split into two then. Will be closing this PR and remaking.

Issue #1299, opened 2018-12-07
Create a Kotlin sample app for spring-cloud-gcp integrations


dzou:
Create a Kotlin sample app example integrating with several spring-cloud-gcp libraries.

dzou:
This is completed; codelab is done as well.

PR #1300, opened 2018-12-07
Setup project directory for Kotlin Samples

dzou:
This accomplishes the pom file setup for a Kotlin sample app directory in spring-cloud-gcp.

In reference to #1299.


spring-cloud-gcp-kotlin-samples/spring-cloud-gcp-kotlin-app-sample/src/main/kotlin/com/example/Controller.kt
```diff
@@ -0,0 +1,27 @@
...
+
+package com.example
+
+import org.springframework.web.bind.annotation.GetMapping
+import org.springframework.web.bind.annotation.RestController
+
+@RestController
+class Controller {
+
+  @GetMapping("/")
```
ChengyuanZhao: I know this is supposed to be a placeholder hello-world function, but since this is a setup PR would it make sense to also make a test class with just that one placeholder hello-world-test? That way subsequent real functionality can just be written in there.

dzou: Done.

spring-cloud-gcp-kotlin-samples/spring-cloud-gcp-kotlin-app-sample/src/test/kotlin/com/example/ApplicationTests.kt
```diff
@@ -0,0 +1,48 @@
...
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.boot.test.context.SpringBootTest
+import org.springframework.boot.test.web.client.TestRestTemplate
+import org.springframework.test.context.junit4.SpringRunner
+
+@RunWith(SpringRunner::class)
+@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)
+class ApplicationTests {
```
meltsufin: I think it should be `ApplicationIntegrationTests`, but I did notice that every other sample that has integration tests doesn't follow this either.
@artembilan Is there a convention for this in Spring?

artembilan: Well, we say "integration" for only those tests which involve interaction with an external resource. 
I would even say that testing against an embedded one is not an integration.
On the other hand having all the simple tools for testing it is hard now to distinguish what is "unit" and what is "integration".

My position if we do access to the external resource, e.g. Pub/Sub emulator is a integration test. Otherwise, with all those Spring Boot goodies for testing in isolation it is not an integration one.
So, for all of this our samples tests we are fully safe and can do all the testing without access to external resources. Therefore on need to confuse end-users and just name test classes as is for simple testing against an embedded servlet container.

meltsufin: Thanks for the clarification. I completely agree. The thing is that we do now have real integration tests in our samples that access external resources like Cloud SQL, Pub/Sub (not emulator), etc.
In this particular case, this app will also probably be using cloud resources eventually.

dzou: Any resolution on this? I am indifferent about the naming convention and can go either way. I am inclined to favor keeping it `ApplicationTests` just for consistency with other samples.

At the present point in time, since all the samples have very few tests, I don't think just calling them "Tests" would lead to confusion. I think only if our samples had both unit and integration tests would a distinction be useful.

meltsufin: I think it's OK to leave it as is for now. Note that there is a [difference](https://maven.apache.org/surefire/maven-failsafe-plugin/) between unit tests and integration tests for Maven. They are run by different plugins. So, it's more than just naming. Although that difference probably doesn't affect us currently.

dzou: Noted. We can certainly consider in the future a large-scale renaming PR for all the tests if necessary.

Issue #1209, opened 2018-11-12
Create a Template for Cloud Vision APIs


dzou:
Create a Template API for cloud vision which simplifies common actions and workflows.

Related to #1208.

dzou:
Marking as completed; VisionTemplate and DocumentOcrTemplate have now been provided.

PR #1307, opened 2018-12-10
Add extractText method for Vision Template

dzou:
This adds a convenience extractText method for the Vision Template.

In reference to: #1209.


spring-cloud-gcp-vision/src/main/java/org/springframework/cloud/gcp/vision/CloudVisionTemplate.java
```diff
@@ -50,6 +51,11 @@ public CloudVisionTemplate(ImageAnnotatorClient imageAnnotatorClient) {
 		this.imageAnnotatorClient = imageAnnotatorClient;
 	}
 
+	public String extractTextFromImage(Resource resource) throws IOException {
```
meltsufin: What's the source of the `IOException`? We probably need to wrap it into a custom runtime exception.
I believe the templates in Spring generally try to avoid checked exceptions. @artembilan WDYT?

spring-cloud-gcp-samples/spring-cloud-gcp-vision-api-sample/src/main/resources/static/index.html
```diff
@@ -5,10 +5,22 @@
     <title>Google Cloud Vision API with Spring code sample</title>
 </head>
 <body>
+
 <div>
-    <form action="/vision">
+    <h1>Image Label Annotations</h1>
+    <p>Returns labels classifying the content of the image:</p>
+    <form action="/extractLabels">
+        URL of image to be uploaded: <input type="text" name="imageUrl"/> <input type="submit"/>
```
meltsufin: Can we pre-fill or suggest using the images already on the classpath?

dzou: So the intention here was not to have users upload images from disk but to specify the URL of an image on the web; I figured this would be much more convenient than uploading/including things on classpath.

I've reworded the description to say "Web URL of image to analyze" to clarify the intention.

spring-cloud-gcp-vision/src/main/java/org/springframework/cloud/gcp/vision/CloudVisionTemplate.java
artembilan: Well, you can simply wrap it into the `UncheckedIOException` 😄

meltsufin: Maybe a `CloudVisionException` wrapping `IOException`?
This also reminds me that we wanted to check for error responses from Cloud Vision and wrap them into exceptions.

spring-cloud-gcp-vision/src/main/java/org/springframework/cloud/gcp/vision/CloudVisionException.java
```diff
@@ -0,0 +1,33 @@
...
+import java.io.UncheckedIOException;
+
+/**
+ * Describes error conditions that can occur when interfacing with Cloud Vision APIs.
+ *
+ * @author Daniel Zou
+ *
+ * @since 1.1
+ */
+final class CloudVisionException extends UncheckedIOException {
```
meltsufin: Why extend the `UncheckedIOException` instead of just `RuntimeException`?

spring-cloud-gcp-vision/src/main/java/org/springframework/cloud/gcp/vision/CloudVisionException.java
```diff
@@ -0,0 +1,33 @@
...
+	CloudVisionException(IOException e) {
```
meltsufin: We should use the version of the constructor that also takes a string message.

spring-cloud-gcp-vision/src/main/java/org/springframework/cloud/gcp/vision/CloudVisionTemplate.java
```diff
@@ -50,6 +51,16 @@ public CloudVisionTemplate(ImageAnnotatorClient imageAnnotatorClient) {
...
 	}
 
+	/**
+	 * Extract the text out of an image and return the result as a String.
+	 * @param imageResource the image one wishes to analyze.
+	 * @return the text extracted from the image aggregated to a String.
+	 */
+	public String extractTextFromImage(Resource imageResource) {
+		AnnotateImageResponse response = analyzeImage(imageResource, Type.TEXT_DETECTION);
+		return response.getFullTextAnnotation().getText();
```
meltsufin: Can we check if the response was an error and wrap into an exception?

spring-cloud-gcp-vision/src/main/java/org/springframework/cloud/gcp/vision/CloudVisionTemplate.java
```diff
@@ -61,10 +72,15 @@ public CloudVisionTemplate(ImageAnnotatorClient imageAnnotatorClient) {
...
 	 */
-	public AnnotateImageResponse analyzeImage(Resource imageResource, Feature.Type... featureTypes)
-			throws IOException {
+	public AnnotateImageResponse analyzeImage(Resource imageResource, Feature.Type... featureTypes) {
+		ByteString imgBytes;
+		try {
+			imgBytes = ByteString.readFrom(imageResource.getInputStream());
+		}
+		catch (IOException e) {
+			throw new CloudVisionException(e);
```
meltsufin: Should also provide a message like "Failed to read image bytes from provided resource: <path>".

spring-cloud-gcp-vision/src/main/java/org/springframework/cloud/gcp/vision/CloudVisionTemplate.java
```diff
@@ -97,7 +104,13 @@ public AnnotateImageResponse analyzeImage(Resource imageResource, Feature.Type..
 		BatchAnnotateImagesResponse batchResponse = this.imageAnnotatorClient.batchAnnotateImages(request);
 		List<AnnotateImageResponse> annotateImageResponses = batchResponse.getResponsesList();
 
-		return annotateImageResponses.get(0);
+		if (!annotateImageResponses.isEmpty()) {
+			return annotateImageResponses.get(0);
+		}
+		else {
+			throw new CloudVisionException(
```
meltsufin: Should we throw an exception for error responses here too?

spring-cloud-gcp-vision/src/main/java/org/springframework/cloud/gcp/vision/CloudVisionTemplate.java
```diff
@@ -58,7 +59,13 @@ public CloudVisionTemplate(ImageAnnotatorClient imageAnnotatorClient) {
 	 */
 	public String extractTextFromImage(Resource imageResource) {
 		AnnotateImageResponse response = analyzeImage(imageResource, Type.TEXT_DETECTION);
-		return response.getFullTextAnnotation().getText();
+
+		String result = response.getFullTextAnnotation().getText();
+		if (result.isEmpty() && response.getError().getCode() != Code.OK.getNumber()) {
+			throw new CloudVisionException(response.getError().getMessage());
```
meltsufin: Please document when the exception is thrown in the javadoc above.

spring-cloud-gcp-vision/src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker
```diff
@@ -0,0 +1 @@
+mock-maker-inline
```
meltsufin: What does this do?

spring-cloud-gcp-samples/spring-cloud-gcp-vision-api-sample/src/main/java/com/example/VisionController.java
```diff
@@ -58,8 +59,8 @@
 	 * @return a string with the list of labels and percentage of certainty
 	 * @throws Exception if the Vision API call produces an error
 	 */
-	@GetMapping("/vision")
-	public ModelAndView uploadImage(String imageUrl, ModelMap map) throws Exception {
+	@GetMapping("/extractLabels")
+	public ModelAndView extractLabels(String imageUrl, ModelMap map) throws IOException {
```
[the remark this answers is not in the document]
artembilan: Why don't you address Sonar recommendations?

spring-cloud-gcp-samples/spring-cloud-gcp-vision-api-sample/src/main/java/com/example/VisionController.java
```diff
@@ -74,4 +75,12 @@ public ModelAndView uploadImage(String imageUrl, ModelMap map) throws Exception
 
 		return new ModelAndView("result", map);
 	}
+
+	@GetMapping("/extractText")
+	public String extractText(String imageUrl) throws IOException {
```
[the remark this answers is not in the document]
artembilan: DITTO

spring-cloud-gcp-vision/src/main/java/org/springframework/cloud/gcp/vision/CloudVisionException.java
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * Describes error conditions that can occur when interfacing with Cloud Vision APIs.
+ *
+ * @author Daniel Zou
+ *
+ * @since 1.1
+ */
+public final class CloudVisionException extends NestedRuntimeException {
+
```
artembilan: You have to add a `serialVersionUID` property into this exception class: https://stackoverflow.com/questions/285793/what-is-a-serialversionuid-and-why-should-i-use-it

spring-cloud-gcp-vision/src/test/java/org/springframework/cloud/gcp/vision/CloudVisionTemplateTests.java
```diff
@@ -0,0 +1,115 @@
...
+import org.junit.rules.ExpectedException;
+import org.mockito.Mockito;
+
+import org.springframework.core.io.AbstractResource;
+import org.springframework.core.io.ByteArrayResource;
+import org.springframework.core.io.Resource;
+
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.when;
+
```
artembilan: Missed `@author` and `@since`. And also seems for me Checkstyle will ask you about some Javadoc sentence as well.

spring-cloud-gcp-vision/src/main/java/org/springframework/cloud/gcp/vision/CloudVisionTemplate.java
meltsufin: I think in this case, it's not really an exception. So, let's leave it to the user.

spring-cloud-gcp-vision/src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker
dzou: Allows you to mock final classes or final methods: https://github.com/mockito/mockito/wiki/What%27s-new-in-Mockito-2#mock-the-unmockable-opt-in-mocking-of-final-classesmethods

This is used to mock the ImageAnnotatorClient to unit test the exception handling.

spring-cloud-gcp-samples/spring-cloud-gcp-vision-api-sample/src/main/java/com/example/VisionController.java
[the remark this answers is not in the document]
dzou: Sorry missed it on the first pass. Done.

spring-cloud-gcp-samples/spring-cloud-gcp-vision-api-sample/src/main/java/com/example/VisionController.java
```diff
@@ -16,6 +16,7 @@
 
 package com.example;
 
+import java.io.IOException;
```
artembilan: This import is out of use. Not sure why it doesn't fail for you with Checkstyle...

spring-cloud-gcp-samples/spring-cloud-gcp-vision-api-sample/src/main/java/com/example/VisionController.java
```diff
@@ -58,8 +58,8 @@
 	 * @return a string with the list of labels and percentage of certainty
 	 * @throws Exception if the Vision API call produces an error
```
meltsufin:
```suggestion
	 * @throws {@link CloudVisionException} if the Vision API call produces an error
```

spring-cloud-gcp-vision/src/main/java/org/springframework/cloud/gcp/vision/CloudVisionTemplate.java
```diff
@@ -50,6 +52,23 @@ public CloudVisionTemplate(ImageAnnotatorClient imageAnnotatorClient) {
 		this.imageAnnotatorClient = imageAnnotatorClient;
 	}
 
+	/**
+	 * Extract the text out of an image and return the result as a String.
+	 * @param imageResource the image one wishes to analyze
+	 * @return the text extracted from the image aggregated to a String
+	 * @throws CloudVisionException if the image could not be read or if text extraction failed
```
meltsufin:
```suggestion
	 * @throws {@link CloudVisionException} if the image could not be read or if text extraction failed
```

spring-cloud-gcp-vision/src/main/java/org/springframework/cloud/gcp/vision/CloudVisionTemplate.java
```diff
@@ -58,13 +77,20 @@ public CloudVisionTemplate(ImageAnnotatorClient imageAnnotatorClient) {
 	 * feature types can be found in {@link Feature.Type}.
 	 * @param imageResource the image one wishes to analyze. The Cloud Vision APIs support
 	 *     image formats described here: https://cloud.google.com/vision/docs/supported-files
-	 * @param featureTypes the types of image analysis to perform on the image.
+	 * @param featureTypes the types of image analysis to perform on the image
 	 * @return the results of image analyses
+	 * @throws CloudVisionException if the image could not be read or if a malformed response
```
meltsufin:
```suggestion
	 * @throws {@link CloudVisionException} if the image could not be read or if a malformed response
```

dzou: Actually I don't think these suggestions work. When I add links I get the checkstyle error: 

> Javadoc comment at column 13 has parse error. Details: no viable alternative at input '{' while parsing JAVADOC_TAG [NonEmptyAtclauseDescription]

I think for exceptions you can omit the `@link`?

PR #1311, opened 2018-12-11
Move Checkstyle suppression config to root pom.xml

dzou:
Moves the Checkstyle suppression config to the project root pom.xml. This resolves issue with the checkstyle suppression not working when a test is run from a sample app directory (and not from project root).

I believe the cause of this is the relative path in the checkstyle.xml which doesn't seem to get translated correctly when checkstyle is not run in the root directory.


pom.xml
```diff
@@ -188,6 +188,7 @@
 						<configuration>
 							<!-- TODO: remove when it matches https://github.com/spring-cloud/spring-cloud-build/blob/master/spring-cloud-build-tools/src/main/resources/checkstyle.xml -->
 							<configLocation>src/checkstyle/checkstyle.xml</configLocation>
+							<suppressionsLocation>src/checkstyle/checkstyle-suppressions.xml</suppressionsLocation>
```
artembilan: Right. This was my point about this: https://github.com/spring-cloud/spring-cloud-build/pull/79
So, that one doesn't work as we talked before: https://github.com/spring-cloud/spring-cloud-gcp/pull/1241/files#r239236261

/CC @dmitry-s

src/checkstyle/checkstyle.xml
```diff
@@ -11,10 +11,6 @@
         <property name="headerFile" value="${checkstyle.header.file}"/>
         <property name="fileExtensions" value="java"/>
     </module>
-    <module name="SuppressionFilter">
-        <property name="file" value="src/checkstyle/checkstyle-suppressions.xml"/>
```
artembilan: I think we need to consider to use here this: https://maven.apache.org/plugins/maven-checkstyle-plugin/check-mojo.html#suppressionsLocation. I mean `checkstyle.suppressions.location` property.

And suggest it for the `spring-cloud-build` instead of that wrong fix...

dzou: Sorry could you clarify how to use the `checkstyle.suppressions.location` property? Do we no longer include it in the pom?

dzou: Got it. I tried briefly and it seemed that including it in the pom alone was sufficient, however for the sake of consistency I will include it here as well. It would be needed if the checkstyle was ever run outside the context of maven.

dzou: I discussed with @dmitry-s and decided to ultimately remove the placeholder and keep the suppression in the pom file. This is because the checkstyle.xml will eventually be removed so it is better to keep the configuration in the pom.

According to my research, defining suppression in the pom is [equivalent to defining it in the checkstyle](https://stackoverflow.com/questions/43871601/defining-suppressions-filter-in-checkstyle-xml-vs-pom-xml-maven-checkstyle-plu).

artembilan: So, we should consider to PR Spring Cloud Build to fix the parent `checkstyle.xml` similar way.
Although I'm not sure that we should specify `checkstyle-suppressions.xml` in their POM...

meltsufin: @artembilan Do we need to revert https://github.com/spring-cloud/spring-cloud-build/pull/79?

artembilan: Yes, please.

Thanks

meltsufin: @dmitry-s PTAL.

PR #1314, opened 2018-12-12
Adds support for spanId in API-based Logging

### LINKED ISSUE #1313: Add support for spanId in Logging
In addition to trace ID that we already support, Stackdriver Logging `LogEntry` now supports span ID.
It works for JSON logging too. See: https://cloud.google.com/logging/docs/agent/configuration#process-payload

meltsufin:
Fixes #1313.


spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/logging/TraceIdLoggingEnhancer.java
```diff
@@ -23,9 +23,11 @@
 import org.springframework.cloud.gcp.core.GcpProjectIdProvider;
 
 /**
- * Adds the trace ID to the logging entry, in its correct format to be displayed in the Logs viewer.
+ * Adds the trace ID and span ID to the logging entry, in its correct format to be
+ * displayed in the Logs viewer.
  *
  * @author João André Martins
+ * @author Mike Eltusfin
```
dzou: Typo here I think :)

PR #1317, opened 2018-12-13
Copy README.adoc to /docs

meltsufin:
The README.adoc in /docs/src/main/asciidoc/ should be the source of
truth for /README.adoc. Every time we run `./mvnw clean install -Pdocs`
it will be copied over to /README.adoc.

Also some minor refdoc fixes.

### DIFF
--- a/docs/src/main/asciidoc/README.adoc
+++ b/docs/src/main/asciidoc/README.adoc
@@ -3,9 +3,7 @@
 This project makes it easy for Spring users to run their applications on Google Cloud Platform.
 You can check our project website http://cloud.spring.io/spring-cloud-gcp[here].
 
-For a deep dive into the project, refer to the
-https://docs.spring.io/spring-cloud-gcp/docs/1.0.0.RELEASE/reference/htmlsingle/[Spring
-Cloud GCP 1.0 Reference Document].
+For a deep dive into the project, refer to the https://docs.spring.io/spring-cloud-gcp/docs/1.0.0.RELEASE/reference/htmlsingle/[Spring Cloud GCP 1.0 Reference Document].
 
 If you prefer to learn by doing, try https://codelabs.developers.google.com/spring[Spring on GCP codelabs].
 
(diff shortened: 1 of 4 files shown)


docs/src/main/asciidoc/spring-cloud-gcp.adoc
```diff
@@ -14,13 +14,13 @@ The Spring Cloud GCP project aims at making the Spring Framework a first-class c
...
 Currently, Spring Cloud GCP lets you leverage the power and simplicity of the Spring framework to:
 
-1. Publish and subscribe from Google Cloud Pub/Sub topics
-2. Configure Spring JDBC with a few properties to use Google Cloud SQL
-3. Write and read from Spring Resources backed up by Google Cloud Storage
-4. Exchange messages with Spring Integration using Google Cloud Pub/Sub on the background
-5. Trace the execution of your app with Spring Cloud Sleuth and Google Stackdriver Trace
-6. Configure your app with Spring Cloud Config, backed up by the Google Runtime Configuration API
-7. Consume and produce Google Cloud Storage data via Spring Integration GCS Channel Adapters
+. Publish and subscribe to Google Cloud Pub/Sub topics
```
artembilan: Since we are here, would you mind also fixing a `Spring Cloud GCP Reference Documentation` to something like just `Spring Cloud GCP`?
The `Reference Documentation` doesn't play good in the aggregated doc of Spring Cloud: https://cloud.spring.io/spring-cloud-static/Greenwich.RC1/multi/multi_spring-cloud.html

Pay attention to other projects links. They don't have such a noise in their names.

Thanks

meltsufin: Sure. Done!

docs/src/main/asciidoc/README.adoc
```diff
@@ -1,18 +1,106 @@
...
-* Cloud Foundry
-* MemoryStore
-* Security IAP
-* Vision
+This project makes it easy for Spring users to run their applications on Google Cloud Platform.
+You can check our project website http://cloud.spring.io/spring-cloud-gcp[here].
+
+For a deep dive into the project, refer to the
+https://docs.spring.io/spring-cloud-gcp/docs/1.0.0.RELEASE/reference/htmlsingle/[Spring
+Cloud GCP 1.0 Reference Document].
```
artembilan: What's wrong with one sentence per line?

meltsufin: I didn't modify this part. I just copy-pasted from what it was before.

artembilan: Good, but what's wrong with our code style rules and why just don't fix a violation since you are already here?
I'm OK do not fix anything we don't touch, but this is a change and it doesn't follow our rules.
Anyone then can say: "This is not my code why should I fix it?"
Then we really won't be able to reach a single solid code style since no one is going to pursue it.

Not sure why my point about pursuing a simplicity for the contribution because we have a single source of code style is so hard to follow...

Thanks for understanding

PR #1321, opened 2018-12-17
Add synchronous pull PubSubMessageSource

### LINKED ISSUE #1056: Replace Pub/Sub Subscriber-based implementation with Pull-based
Pub/Sub streaming subscriber API is currently not well suited for low QPS applications.

See #984 for more context.
See [docs](https://cloud.google.com/pubsub/docs/pull#dealing-with-large-backlogs-of-small-messages) for an explanation.
cc: @kir-titievsky

elefeint:
Other changes:
* Modified sample Pub/Sub Sender app to accept the number of times to
send the message and append sequential message number to each copy.
* Split sample Pub/Sub Receiver app into StreamingPull (original behavior) and
SyncPull (new implementation) configuration subclasses.

Fixes #1056 .


spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubMessageSource.java
```diff
@@ -0,0 +1,125 @@
...
+	}
+
+	public void setHeaderMapper(HeaderMapper<Map<String, String>> headerMapper) {
+		this.headerMapper = headerMapper;
+	}
+
+	@Override
+	protected Object doReceive() {
+		List<ConvertedAcknowledgeablePubsubMessage> messages
+				= this.pubSubSubscriberOperations.pullAndConvert(this.subscriptionName, 1, true, this.payloadType);
```
meltsufin: Any way we can avoid always requesting one message at a time? Would it make sense?
Also, `returnImmediately` should probably be a setting exposed to the user.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubMessageSource.java
```diff
@@ -0,0 +1,125 @@
...
+
+		if (messages.size() < 1) {
+			return null;
+		}
+
+		if (messages.size() > 1) {
+			LOGGER.warn("One message expected; " + messages.size() + " received. Returning the first one.");
```
meltsufin: Maybe throw an exception instead?

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubMessageSource.java
```diff
@@ -0,0 +1,125 @@
...
+		return this.headerMapper;
+	}
+
+	public void setHeaderMapper(HeaderMapper<Map<String, String>> headerMapper) {
+		this.headerMapper = headerMapper;
+	}
+
+	@Override
+	protected Object doReceive() {
+		List<ConvertedAcknowledgeablePubsubMessage> messages
```
meltsufin: You probably need a try-catch around the whole method and nack, if `AUTO` ack mode is on.

spring-cloud-gcp-samples/spring-cloud-gcp-integration-pubsub-sample/README.adoc
```diff
@@ -18,10 +18,41 @@ Alternatively, if you have the https://cloud.google.com/sdk/[Google Cloud SDK] i
...
-4. Go to http://localhost:8080, write a message in the text box and hit the `Submit` button.
+5. Go to http://localhost:8080, write a message in the text box and hit the `Submit` button.
 
-5. Verify that the receiver logged the message you wrote.
+6. Verify that the receiver logged the message you wrote.
 +
 `Message arrived! Payload: [message-entered]`
+
+
+=== Synchronous Pull Based Subscription.
```
meltsufin:
```suggestion
=== Synchronous Pull Based Subscription
```

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubMessageSource.java
```diff
@@ -0,0 +1,125 @@
...
+
+/**
+ * A <a href="https://cloud.google.com/pubsub/docs/pull#pubsub-pull-messages-sync-java">PubSub Synchronous pull</a>
+ * implementation of {@link AbstractMessageSource}.
+ *
+ * Enables more even load balancing with polling adapters, at the cost of efficiency gained from batched pulls.
+ *
+ * @author Elena Felder
+ *
+ * @since 1.1
```
meltsufin: Do we want these changes to go into 1.1?

spring-cloud-gcp-samples/spring-cloud-gcp-integration-pubsub-sample/README.adoc
```diff
@@ -18,10 +18,41 @@ Alternatively, if you have the https://cloud.google.com/sdk/[Google Cloud SDK] i
...
+
+1. Start the `spring-cloud-gcp-spring-integration-pubsub-sample-receiver` sample app.
+
+1. Open http://localhost:8080, write a message in the text box,  update the "times" text box to 40, and hit send. The messages will then be queued up, with no subscribers available to drain them yet.
+
+1. In two separate terminal windows, start the `spring-cloud-gcp-integration-pubsub-sample-receiver` sample application twice, both times introducing a delay:
+
+  mvn spring-boot:run -Dspring-boot.run.arguments=--delay
+
+1. Observe that only one receiver app greedily retrieved all 40 messages, while the second application is idle.
```
meltsufin: I'm not sure it will always be 100% reproducible. The streaming pull batch size is not guaranteed.
If we're planning to deprecate the streaming pull, I'm not sure adding all of this clarification is necessary. It may actually create more confusion for new users.

spring-cloud-gcp-samples/spring-cloud-gcp-integration-pubsub-sample/spring-cloud-gcp-integration-pubsub-sample-sender/src/main/resources/static/index.html
```diff
@@ -1,6 +1,6 @@
 <!DOCTYPE html>
 <!--
-  ~  Copyright 2017 original author or authors.
+  ~  Copyright 2017-2018 the original author or authors.
```
artembilan: No Copyright in the non-Java(Kotlin, Groovy) files.
In other words: no Copyright, please, in the XML and HTML files.

artembilan:
> It seems like you're trying to keep both channel adapters around. What's the motivation for that?

Wait. What is the motivation to remove a streaming channel adapter? The Binder impl is based on it and that it how we reach streaming solution.
Right now I am fully confused... Is Pub/Sub going to remove their subscriber solution? 🤷‍♂️

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubMessageSource.java
artembilan: Yes, it's indeed make sense. The state of the topic may change in between your `receive()` calls. And that is not going to be to friendly to keep messages in the internal queue because we decided do not pull for each of them.

However if pulled, but not ack'ed messages are not visible for other consumers, that really might have a value to consider a local queue cache for the pulled messages according `maxFetchSize` option. 
See `AbstractFetchLimitingMessageSource` for more info.

elefeint:
My thinking was that we would keep `PubSubInboundChannelAdapter`, even if/when we deprecate `PubSubTemplate.subscribe*()` calls. It seems useful for the high-volume scenario that GCP Pub/Sub team addressed, with the message pre-caching logic on streaming pull. It also seems useful (simple, efficient) for the basic scenario with a single subscriber.

The MessageSource implementation is nice and simple, but it does not allow the user to vary batch size. So if we deprecate `PubSubTemplate.subscribe*()` calls, it may be useful to let users batch messages by changing `PubSubInboundChannelAdapter` internals to use synchronous pulls with batch size specified as a property.   

In any case, even if we decide on more invasive changes, I would suggest merging my PR in (after I address technical feedback), and considering the overall architecture post-1.1 release. The pollable message source is useful, and addresses the issue with load balancing that we keep seeing in the wild.

> > It seems like you're trying to keep both channel adapters around. What's the motivation for that?
> 
> Wait. What is the motivation to remove a streaming channel adapter? The Binder impl is based on it and that it how we reach streaming solution.
> Right now I am fully confused... Is Pub/Sub going to remove their subscriber solution?

spring-cloud-gcp-samples/spring-cloud-gcp-integration-pubsub-sample/README.adoc
elefeint: It's 100% reproducible if the receivers are started after the messages are queued up on the topic, though. We don't have to explain this behavior, but then it may be difficult for the user to pick their message consumption strategy.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubMessageSource.java
meltsufin: If it's a "should never happen" scenario, then it should be an exception IMO.

meltsufin:
@artembilan PTAL at the issue #1056 and linked references regarding why the streaming pull is problematic. We've had many complaints about it because it pre-fetches large batches, and not playing well with multiple consumers.

artembilan:
Hm. Can't we just have an Idempotent Receiver then on the consumer side to deal with duplicate messages? 
I would say I don't care too much about balancing between consumers on the topic, but all what you say is scaring me in regards what we have with the Binder implementation...

Sounds like we fully should not recommend it for the production usage...

But even if we have some limitation on that side, it doesn't mean that the whole implementation should go away.
I don't see anything bad from the Spring perspective. Might be just a case for the Pub/Sub (or its client lib) to fix the problem somehow...

meltsufin:
Pub/Sub team doesn't really consider it a problem, because the StreamingPull is optimized for high throughput. It doesn't really care about slow consumers and doesn't provide a mechanism for controlling back-pressure.
So, the Binder is not fundamentally broken, but the theory is that an implementation based on the synchronous Pull will be more predictable and accommodate slow consumers better.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubMessageSource.java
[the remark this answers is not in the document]
elefeint: It's a safety net for "this should never happen". The pullAndConvert() contract is to return a list. We know that the current implementation will only return 1 when 1 is asked for. But what if there is a bug in the implementation? Or a change in underlying behavior?

I'll replace the log with an assertion to make expected behavior clear.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubMessageSource.java
```diff
@@ -91,14 +77,11 @@ protected Object doReceive() {
...
-		if (messages.size() < 1) {
+		if (messages.isEmpty()) {
 			return null;
 		}
 
-		if (messages.size() > 1) {
-			LOGGER.warn("One message expected; " + messages.size() + " received. Returning the first one.");
-		}
-
+		Assert.isTrue(messages.size() == 1, "Only one Pub/Sub message expected.");
```
artembilan: `isTrue()` doesn't fit our requirements. It throws an `IllegalArgumentException`, which doesn't reflect the situation. That's why I recommended to use a `state()` here.

elefeint:
I will now put this PR on ice until we can discuss the PubSub implementation direction. This will likely conveniently be after 1.1 release.

I'll write up a doc with options and send it around.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubMessageSource.java
```diff
@@ -0,0 +1,109 @@
...
+ * A <a href="https://cloud.google.com/pubsub/docs/pull#pubsub-pull-messages-sync-java">PubSub Synchronous pull</a>
+ * implementation of {@link AbstractMessageSource}.
+ *
+ * Enables more even load balancing with polling adapters, at the cost of efficiency gained from batched pulls.
+ *
+ * @author Elena Felder
+ *
+ * @since 1.1
+ */
+public class PubSubMessageSource extends AbstractMessageSource<Object> {
```
dzou: Just wanted to clarify on naming convention - does MessageSource imply "polling" and channelAdapter imply "stream" in Spring? If not I would suggest including 'sync' or 'polling' in the class name.

artembilan: That's correct. The `MessageSource` does nothing by itself. Only wrapping `SourcePollingChannelAdapter` does the polling magic.
I agree that `PubSubInboundChannelAdapter` is not so obvious name, but our convention to use such a name for those components which are doing the streaming via listening in the internal task.
Sometimes we call them `MessageDrivenEndpoint`, e.g. `JmsMessageDrivenEndpoint`.

spring-cloud-gcp-samples/spring-cloud-gcp-integration-pubsub-sample/spring-cloud-gcp-integration-pubsub-sample-receiver/src/main/java/com/example/ReceiverApplication.java
```diff
@@ -16,53 +16,118 @@
...
+			message.ack();
+		}
 	}
 
-	@ServiceActivator(inputChannel = "pubsubInputChannel")
-	public void messageReceiver(String payload,
-			@Header(GcpPubSubHeaders.ORIGINAL_MESSAGE) BasicAcknowledgeablePubsubMessage message) {
-		LOGGER.info("Message arrived! Payload: " + payload);
-		message.ack();
+	@ConditionalOnProperty("syncpull")
```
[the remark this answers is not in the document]
elefeint: Yep.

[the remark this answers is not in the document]
artembilan: Yeah... Agreed. A separate Spring Boot application would be much cleaner. Even if it is going to be in the same module, but in separate package tree.

artembilan:
> Updating the Cloud Stream Binder to use the sync pull

Well, the point is to still keep an async listening consumer as is and provide a `PollableConsumerBinder.bindPollableConsumer()` implementation for this newly introduced `PubSubMessageSource`: http://cloud.spring.io/spring-cloud-static/spring-cloud-stream/2.1.0.RELEASE/single/spring-cloud-stream.html#spring-cloud-streams-overview-using-polled-consumers

meltsufin:
> > Updating the Cloud Stream Binder to use the sync pull
> 
> Well, the point is to still keep an async listening consumer as is and provide a `PollableConsumerBinder.bindPollableConsumer()` implementation for this newly introduced `PubSubMessageSource`: http://cloud.spring.io/spring-cloud-static/spring-cloud-stream/2.1.0.RELEASE/single/spring-cloud-stream.html#spring-cloud-streams-overview-using-polled-consumers

Nice! @elefeint Can you please create an issue to that effect?

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubMessageSource.java
elefeint: I am not sure that's useful, since it does not tell us anything about what happened during message processing. A channel adapter, like PubSubInboundChannelAdapter, handles sending to output, so it can observe whether downstream processing failed. A message source like PubSubMessageSource merely provides the source data; it does not have visibility into what happens at destination. That's a concern of SourcePollingChannelAdapter, which we don't control.

I think users of pollable message source will just have to opt for manual acking if "ack when the message reached Spring Integration infrastructure" is not sufficient.

artembilan: You can take a look what we don in the `AmqpMessageSource`: https://github.com/spring-projects/spring-integration/blob/master/spring-integration-amqp/src/main/java/org/springframework/integration/amqp/inbound/AmqpMessageSource.java
or `KafkaMessageSource`: https://github.com/spring-projects/spring-integration-kafka/blob/master/src/main/java/org/springframework/integration/kafka/inbound/KafkaMessageSource.java

elefeint: Ah, so the class would add its own callback header. I'll do that, thank you!

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubMessageSource.java
elefeint: Should I throw a post-1.1 tag on it?

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubMessageSource.java
```diff
@@ -0,0 +1,162 @@
...
+	 * @return {@link Message} wrapper containing the original message.
+	 */
+	private Message<Object> processMessage(ConvertedAcknowledgeablePubsubMessage message) {
+
+		Map<String, Object> messageHeaders =
+				this.headerMapper.toHeaders(message.getPubsubMessage().getAttributesMap());
+
+		if (this.ackMode == AckMode.MANUAL) {
+			// Send the original message downstream so user decides on when to ack/nack.
+			messageHeaders.put(GcpPubSubHeaders.ORIGINAL_MESSAGE, message);
```
artembilan: I think we need to revise this logic only in favor of `AcknowledgmentCallback` in the headers. 
See the same `AmqpMessageSource` or `KafkaMessageSource`

docs/src/main/asciidoc/spring-integration-pubsub.adoc
```diff
@@ -104,6 +104,31 @@ public MessageHandler messageReceiver() {
 }
 ----
 
+==== Pollable Message Source (using Pub/Sub Synchronous Pull)
+
+While `PubSubInboundChannelAdapter`, through the underlying Asynchronous Pull Pub/Sub mechanism, provides the best performance for high-volume applications that receive a steady flow of messages, it can create load balancing anomalies due to message caching.
+This behavior is most obvious when publishing a large batch of small messages that take a long time to process individually. It manifests as one subscriber taking up most messages, even if multiple subscribers are available to take on the work.
+For a more detailed explanation of this scenario, see https://cloud.google.com/pubsub/docs/pull#dealing-with-large-backlogs-of-small-messages[GCP Pub/Sub documentation].
+
+In such a scenario, `PubSubMessageSource`, combined with a pollable channel, help spread the load between different subscribers more evenly.
```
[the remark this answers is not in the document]
artembilan: Well, actually a `MessageSource` can be configured with an XML, as well as Java DSL.
It is not going to be fair to assume that `PubSubMessageSource` is going to work only with the `@InboundChannelAdapter`.

I would say it more generic way:
```suggestion
In such a scenario, a `PubSubMessageSource` can help spread the load between different subscribers more evenly.
```
And it is already doesn't matter for us how it is configured or even used directly, e.g. from the `MessageSourcePollingTemplate`.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubMessageSource.java
```diff
@@ -0,0 +1,162 @@
...
+	private ArrayDeque<ConvertedAcknowledgeablePubsubMessage> cachedMessages = new ArrayDeque<>();
+
+	public PubSubMessageSource(PubSubSubscriberOperations pubSubSubscriberOperations,
+			String subscriptionName, int maxFetchSize) {
+		Assert.notNull(pubSubSubscriberOperations, "Pub/Sub subscriber template can't be null.");
+		Assert.notNull(subscriptionName, "Pub/Sub subscription name can't be null.");
+		Assert.isTrue(maxFetchSize > 0, "maxFetchSize has to be a positive integer.");
+		this.pubSubSubscriberOperations = pubSubSubscriberOperations;
+		this.subscriptionName = subscriptionName;
+		super.setMaxFetchSize(maxFetchSize);
```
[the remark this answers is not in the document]
artembilan: Default, "super" negative value means poll as much as possible until `null` returned. 
In other words - unlimited request.

Can't we propagate a `null` in that case though? To meet this logic:
```
if (maxMessages != null) {
	pullRequestBuilder.setMaxMessages(maxMessages);
}
```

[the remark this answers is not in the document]
elefeint: Apparently that's not supported (the error message is misleading; internally the validation is for a positive integer):
````
io.grpc.StatusRuntimeException: INVALID_ARGUMENT: A required argument is missing in the request: (argument="max_messages")
````
I'll file an issue with PubSub to make this restriction clear in GCP documentation.

In the meantime, how about a sensible, property-configurable default of, say, 10 messages per batch?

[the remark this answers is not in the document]
Comment (meltsufin) [reply spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubMessageSource.java]: How about a default of 1? Seems like it would at least avoid the surprises of uneven load distribution among nodes, and the slow consumers issue.

File: spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubAcknowledgmentCallback.java [hunk 14]
```diff
@@ -0,0 +1,73 @@
...
+	 */
+	@Override
+	public void acknowledge(Status status) {
+		if (status == AcknowledgmentCallback.Status.ACCEPT) {
+			this.message.ack();
+		}
+		else if (this.ackMode == AckMode.MANUAL || this.ackMode == AckMode.AUTO) {
+			this.message.nack();
+		}
+		this.acknowledged = true;
```
meltsufin: Nacked also means acknowledged?

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubMessageSource.java
```diff
@@ -0,0 +1,147 @@
...
+import org.springframework.integration.endpoint.AbstractMessageSource;
+import org.springframework.integration.mapping.HeaderMapper;
+import org.springframework.messaging.Message;
+import org.springframework.util.Assert;
+
+/**
+ * A <a href="https://cloud.google.com/pubsub/docs/pull#pubsub-pull-messages-sync-java">PubSub Synchronous pull</a>
+ * implementation of {@link AbstractMessageSource}.
+ *
+ * <p>Enables more even load balancing with polling adapters, at the cost of efficiency gained from streaming pull.
```
meltsufin: I don't know if it's entirely accurate to say that there is always an efficiency cost with sync pull.
If you only need one message, sync pulling should be more efficient because you're not transferring a whole batch of messages.
It might be more accurate to say that message receive throughput is lower with sync.
```suggestion
 * <p>Enables more even load balancing with polling adapters, at the cost of lower maximum message receive throughput than achieved by streaming pull.
```

docs/src/main/asciidoc/spring-integration-pubsub.adoc
```diff
@@ -104,6 +104,32 @@ public MessageHandler messageReceiver() {
...
+
+[source,java]
+----
+@Bean
+@InboundChannelAdapter(channel = "pubsubInputChannel", poller = @Poller(fixedDelay = "100"))
+public MessageSource<Object> pubsubAdapter(PubSubTemplate pubSubTemplate) {
+	PubSubMessageSource messageSource = new PubSubMessageSource(pubSubTemplate,  "exampleSubscription");
+	messageSource.setAckMode(AckMode.MANUAL);
+	messageSource.setPayloadType(String.class);
+	messageSource.blockOnPull();
```
meltsufin: What about an example an discussion of fetch size?

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubMessageSource.java
```diff
@@ -0,0 +1,147 @@
...
+
+	public void setHeaderMapper(HeaderMapper<Map<String, String>> headerMapper) {
+		Assert.notNull(headerMapper, "The header mapper can't be null.");
+		this.headerMapper = headerMapper;
+	}
+
+	/**
+	 * Instructs synchronous pull to wait until at least one message is available.
+	 */
+	public void blockOnPull() {
```
meltsufin: `setBlockOnPull(boolean)`

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubMessageSource.java
```diff
@@ -0,0 +1,147 @@
...
+	protected Object doReceive(int fetchSize) {
+		if (this.cachedMessages.isEmpty()) {
+			Integer maxMessages = (fetchSize > 0) ? fetchSize : 1;
+
+			List<ConvertedAcknowledgeablePubsubMessage> messages
+					= this.pubSubSubscriberOperations.pullAndConvert(this.subscriptionName, maxMessages, !this.blockOnPull, this.payloadType);
+			if (messages.isEmpty()) {
+				return null;
+			}
+			else if (messages.size() == 1) {
```
meltsufin: I would just skip this branch because the `else` adequately handles this case.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubMessageSource.java
```diff
@@ -0,0 +1,147 @@
...
+				this.cachedMessages.addAll(messages);
+			}
+		}
+
+		return processMessage(this.cachedMessages.pollFirst());
+	}
+
+	@Override
+	public String getComponentType() {
+		return "gcp-pubsub:message-source";
```
meltsufin: What is this based on?

spring-cloud-gcp-samples/spring-cloud-gcp-integration-pubsub-sample/README.adoc
```diff
@@ -17,10 +17,36 @@ Alternatively, if you have the https://cloud.google.com/sdk/[Google Cloud SDK] i
...
+1. Verify that the receiver logged the message(s):
++
+`Message arrived by Synchronous Pull! Payload: [message-entered] (0)`
+
+
+
+NOTE: While `PubSubInboundChannelAdapter`, through the underlying Asynchronous Pull Pub/Sub mechanism, provides the best performance for high-volume applications that receive a steady flow of messages, it can create load balancing anomalies due to message caching.
+You can reproduce this behavior by starting up multiple instances of either receiver sample app, and introducing a processing delay with
+ +
+ `mvn spring-boot:run -Dspring-boot.run.arguments=--delay`.
```
meltsufin: If you're adding this "delay" parameter, you might as well consider making it possible to specify the delay in ms: `--delay=1000`.

elefeint: Meh, I'll remove all of delays and documented description of how to reproduce the problem. PubSub client library #googleapis/google-cloud-java/issues/3152 has a good description of how to reproduce the issue; we can always refer people to it.
The more I have to re-review this extra code, the less I remember why I added it in the first place.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubAcknowledgmentCallback.java
elefeint: It means "processed" more than acknowledged. This flag is used to disable auto acking and [auto nacking](https://github.com/spring-projects/spring-integration/blob/23a3462d8bc20e8b27771a6442599d0252733b26/spring-integration-core/src/main/java/org/springframework/integration/acks/AckUtils.java#L65).

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubMessageSource.java
elefeint: I am removing this comment entirely. The docs go into enough detail about why you may use one over the other.

docs/src/main/asciidoc/spring-integration-pubsub.adoc
```diff
@@ -104,6 +104,32 @@ public MessageHandler messageReceiver() {
...
+While `PubSubInboundChannelAdapter`, through the underlying Asynchronous Pull Pub/Sub mechanism, provides the best performance for high-volume applications that receive a steady flow of messages, it can create load balancing anomalies due to message caching.
+This behavior is most obvious when publishing a large batch of small messages that take a long time to process individually.
+It manifests as one subscriber taking up most messages, even if multiple subscribers are available to take on the work.
+For a more detailed explanation of this scenario, see https://cloud.google.com/pubsub/docs/pull#dealing-with-large-backlogs-of-small-messages[GCP Pub/Sub documentation].
+
+In such a scenario, a `PubSubMessageSource` can help spread the load between different subscribers more evenly.
+
+As with the Inbound Channel Adapter, the message source has a configurable acknowledgement mode, payload type, and header mapping.
+The default behavior is to return from the synchronous pull operation immediately if no messages are present.
+This can be overridden by using `blockOnPull()` method to wait for at least one message to arrive.
```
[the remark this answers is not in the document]
elefeint: @meltsufin Perhaps we should not be exposing this option at all. MessageSource is used by a polling channel -- what is the usecase in which a single poll needs to block? If nothing is returned, that's fine -- the channel will poll again shortly.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubMessageSource.java
elefeint: I don't know; putting a message into a data structure (no matter how efficient) only to immediately take it out does not sit well with me in something that is as high processing volume as pubsub.

meltsufin: Probably negligible performance impact considering everything else that needs to happen up to this point, but you have a point.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubAcknowledgmentCallback.java
```diff
@@ -0,0 +1,73 @@
...
+
+	private AcknowledgeablePubsubMessage message;
+
+	private AckMode ackMode;
+
+	private boolean acknowledged;
+
+	public PubSubAcknowledgmentCallback(AcknowledgeablePubsubMessage message, AckMode ackMode) {
+		this.message = message;
+		this.ackMode = ackMode;
```
artembilan: Both these properties must be `final` and probably with `Assert.notNull()` here in the ctor...

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubMessageSource.java
```diff
@@ -0,0 +1,145 @@
...
+
+	private final String subscriptionName;
+
+	private final PubSubSubscriberOperations pubSubSubscriberOperations;
+
+	private AckMode ackMode = AckMode.MANUAL;
+
+	private HeaderMapper<Map<String, String>> headerMapper = new PubSubHeaderMapper();
+
+	private Class payloadType = byte[].class;
```
artembilan: Must be like `Class<?>`

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubMessageSource.java
```diff
@@ -0,0 +1,145 @@
...
+		this.pubSubSubscriberOperations = pubSubSubscriberOperations;
+		this.subscriptionName = subscriptionName;
+	}
+
+	public void setAckMode(AckMode ackMode) {
+		Assert.notNull(ackMode, "The acknowledgement mode can't be null.");
+		this.ackMode = ackMode;
+	}
+
+	public void setPayloadType(Class payloadType) {
```
artembilan: Must be like `Class<?>`

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubMessageSource.java
```diff
@@ -0,0 +1,145 @@
...
+	 * by {@code fetchSize}.
+	 * @param fetchSize number of messages to fetch from Pub/Sub.
+	 * @return {@link Message} wrapper containing the original message.
+	 */
+	@Override
+	protected Object doReceive(int fetchSize) {
+		if (this.cachedMessages.isEmpty()) {
+			Integer maxMessages = (fetchSize > 0) ? fetchSize : 1;
+
+			List<ConvertedAcknowledgeablePubsubMessage> messages
```
artembilan: Must be like `List<? extends ConvertedAcknowledgeablePubsubMessage<?>>` after changing the `payloadType` above to the `Class<?>`

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubMessageSource.java
```diff
@@ -0,0 +1,145 @@
...
+	public void setHeaderMapper(HeaderMapper<Map<String, String>> headerMapper) {
+		Assert.notNull(headerMapper, "The header mapper can't be null.");
+		this.headerMapper = headerMapper;
+	}
+
+	/**
+	 * Instructs synchronous pull to wait until at least one message is available.
+	 */
+	public void setBlockOnPull() {
+		this.blockOnPull = true;
```
artembilan: Would be better to make it really as a setter: `setBlockOnPull(boolean blockOnPull)`.
This way end-user won't need to have an `if...else` in case this option comes from some configuration.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubMessageSource.java
```diff
@@ -0,0 +1,145 @@
...
+			else if (messages.size() == 1) {
+				// don't bother storing.
+				return processMessage(messages.get(0));
+			}
+			else {
+				this.cachedMessages.addAll(messages);
+			}
+		}
+
+		return processMessage(this.cachedMessages.pollFirst());
```
artembilan: we have to check for `null` here anyway.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubMessageSource.java
```diff
@@ -0,0 +1,145 @@
...
+	}
+
+	/**
+	 * Applies header customizations and acknowledges the message, if necessary.
+	 * <p>{@link AckMode#AUTO} and {@link AckMode#AUTO_ACK} result in automatic acking on
+	 * success. {@link AckMode#AUTO} results in automatic nacking on failure.
+	 * @param message source Pub/Sub message.
+	 * @return {@link Message} wrapper containing the original message.
+	 */
+	private Message<Object> processMessage(ConvertedAcknowledgeablePubsubMessage message) {
```
artembilan: The contract must be like this:
```
private Message<?> processMessage(ConvertedAcknowledgeablePubsubMessage<?> message) {
```

After fixing `Class<?>`, of course

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubMessageSource.java
```diff
@@ -0,0 +1,145 @@
...
+				this.headerMapper.toHeaders(message.getPubsubMessage().getAttributesMap());
+
+		messageHeaders.put(GcpPubSubHeaders.ORIGINAL_MESSAGE, message);
+		messageHeaders.put(IntegrationMessageHeaderAccessor.ACKNOWLEDGMENT_CALLBACK,
+					new PubSubAcknowledgmentCallback(message, this.ackMode));
+
+		return getMessageBuilderFactory()
+				.withPayload(message.getPayload())
+				.copyHeaders(messageHeaders)
+				.build();
```
artembilan: Consider return an `AbstractIntegrationMessageBuilder` instead.
See `evaluateHeaders()` in the `AbstractMessageSource`. So, with additional header we won't be need to rebuilt into a new `Message`.

In other words: don't call this `build()`.

spring-cloud-gcp-pubsub/src/test/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubMessageSourceTests.java
```diff
@@ -0,0 +1,258 @@
...
+
+	@Test
+	public void blockOnPullSetsReturnImmediatelyToFalse() {
+
+		PubSubMessageSource pubSubMessageSource = new PubSubMessageSource(
+				this.mockPubSubSubscriberOperations, "sub1");
+		pubSubMessageSource.setMaxFetchSize(1);
+		pubSubMessageSource.setBlockOnPull();
+
+		Object message = pubSubMessageSource.doReceive(1);
```
artembilan: I guess we need to *assert* this for `null` at least.

And also, please, add `@SuppressWarning("unchecked")` for all the method where we have such a warning, if you can't fix generics, of course...

docs/src/main/asciidoc/spring-integration-pubsub.adoc
```diff
@@ -104,6 +104,53 @@ public MessageHandler messageReceiver() {
...
+@ServiceActivator(inputChannel = "pubsubInputChannel")
+public void messageReceiver(String payload,
+        @Header(GcpPubSubHeaders.ORIGINAL_MESSAGE) AcknowledgeablePubsubMessage message)
+            throws InterruptedException {
+    LOGGER.info("Message arrived by Synchronous Pull! Payload: " + payload);
+    message.ack();
+}
+----
+
+NOTE: `AcknowledgeablePubSubMessage` objects acquired by synchronous pull are aware of their own acknowledgement IDs. Streaming pull does not expose this information due to limitations of the underlying API, and returns `BasicAcknowledgeablePubsubMessage` objects that allow acking/nacking individual messages, but not extracking acknowledgement IDs for future processing.
```
artembilan: One sentence per line.
Also there is a type in the `extracking` word 😄

docs/src/main/asciidoc/spring-integration-pubsub.adoc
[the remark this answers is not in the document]
elefeint: Tested this; poller behaves sensibly and waits for the previous poll to complete, even with `fixedRate`. Leaving block on pull option in the code.

docs/src/main/asciidoc/spring-integration-pubsub.adoc
```diff
@@ -104,6 +104,53 @@ public MessageHandler messageReceiver() {
...
+	return messageSource;
+}
+----
+
+The `MessageSource` attaches a `AcknowledgeablePubSubMessage` object to the `Message` headers.
+Users can extract the `AcknowledgeablePubSubMessage` using the `GcpPubSubHeaders.ORIGINAL_MESSAGE` key and use it to (n)ack a message.
+
+[source,java]
+----
+@ServiceActivator(inputChannel = "pubsubInputChannel")
```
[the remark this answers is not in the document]
elefeint: The async example has it; but it's far below the adapter configuration -- after the ack mode discussion.

elefeint:
Stress-testing GitHub's databases!

Issue #1318, opened 2018-12-13
Review and update refdoc for 1.1 release


meltsufin:
This should be done about a week before the release.

ChengyuanZhao:
check when completing proof reading:

- [ ] cloud foundry
- [ ] config
- [ ] core
- [ ] datastore
- [ ] dependency management
- [ ] getting-started
- [ ] logging
- [ ] memorystore
- [ ] pubsub
- [ ] README 
- [ ] security IAP
- [ ] spanner
- [ ] spring-cloud-gcp / index file
- [ ] spring integration
- [ ] si pubsub
- [ ] si storage
- [ ] spring stream
- [ ] SQL
- [ ] storage
- [ ] trace
- [ ] vision

artembilan:
Re. Cloud Foundry though: https://github.com/spring-cloud/spring-cloud-connectors/issues/248.

The point is to remove remove a `In order for the Cloud SQL integration to work in Cloud Foundry, auto-reconfiguration must be disabled. ` paragraph altogether and don't advertise disabling build-packs.
There might be some other features in the target application, not only just Cloud SQL.

Let's hope that feature is going to be implemented in the Cloud Connectors in time for our 1.1 GA!

PR #1344, opened 2018-12-26
Core refdoc proof read and update

ChengyuanZhao:
related #1318 
core refdoc update and proof read

### DIFF
--- a/docs/src/main/asciidoc/core.adoc
+++ b/docs/src/main/asciidoc/core.adoc
@@ -1,7 +1,7 @@
 [#spring-cloud-gcp-core]
 == Spring Cloud GCP Core
 
-At the center of every Spring Cloud GCP module are the concepts of `GcpProjectIdProvider` and `CredentialsProvider`.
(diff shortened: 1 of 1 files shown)


docs/src/main/asciidoc/core.adoc
```diff
@@ -1,7 +1,7 @@
 [#spring-cloud-gcp-core]
 == Spring Cloud GCP Core
 
-At the center of every Spring Cloud GCP module are the concepts of `GcpProjectIdProvider` and `CredentialsProvider`.
+Each Spring Cloud GCP module uses `GcpProjectIdProvider` and `CredentialsProvider` to get the GCP project ID and credentials to access the Google Cloud.
```
elefeint: "to get the GCP project ID and access credentials"? Otherwise this sentence has both "GCP" and "Google Cloud", and looks like project ID is needed to access.

ChengyuanZhao: done

Issue #1198, opened 2018-11-09
Checkstyle: enable IllegalImportCheck


meltsufin:
Currently it disallows `com.google.common`.
Also, upstream is missing disallowing `org.slf4j`.

Part of #1195.

PR #1349, opened 2018-12-27
[to be reviewed in 2019] Replace direct executor with cached pool

### LINKED ISSUE #1063: Use multi-threaded executor for Pub/Sub acking
`PubSubSubscriberTemplate.ack/nack/modifyAckDeadline` should use  a more efficient executor, for example Executors.newCachedThreadPool().

From [ this comment ](https://github.com/spring-cloud/spring-cloud-gcp/pull/1055/files/165e950214981f42c01930ef10a8eb074db22250#r221975690) on #1055 :

* The thread executor should be global per PubSubSubscriberTemplate instance, and should be destroyed in the end of component lifecycle.
* Also we need to provide a setter to allow to inject an externally managed executor.

elefeint:
This is a task left over from batching acks by subscription (#1055). I ran a few tests again, and I can no longer reproduce the performance improvement, but at this point we need to remove `MoreExecutors.directExecutor()` anyway, as part of Guava removal in #1198.

Fixes #1063.

### DIFF
--- a/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberTemplate.java
+++ b/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberTemplate.java
@@ -20,6 +20,8 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.BiFunction;
 import java.util.function.Consumer;
@@ -32,7 +34,6 @@ import com.google.cloud.pubsub.v1.AckReplyConsumer;
 import com.google.cloud.pubsub.v1.MessageReceiver;
 import com.google.cloud.pubsub.v1.Subscriber;
 import com.google.cloud.pubsub.v1.stub.SubscriberStub;
-import com.google.common.util.concurrent.MoreExecutors;
 import com.google.protobuf.Empty;
 import com.google.pubsub.v1.AcknowledgeRequest;
 import com.google.pubsub.v1.ModifyAckDeadlineRequest;
@@ -41,6 +42,7 @@ import com.google.pubsub.v1.PubsubMessage;
 import com.google.pubsub.v1.PullRequest;
 import com.google.pubsub.v1.PullResponse;
 
+import org.springframework.beans.factory.DisposableBean;
 import org.springframework.cloud.gcp.pubsub.support.AcknowledgeablePubsubMessage;
 import org.springframework.cloud.gcp.pubsub.support.BasicAcknowledgeablePubsubMessage;
 import org.springframework.cloud.gcp.pubsub.support.SubscriberFactory;
@@ -67,13 +69,15 @@ import org.springframework.util.concurrent.SettableListenableFuture;
  *
  * @since 1.1
  */
-public class PubSubSubscriberTemplate implements PubSubSubscriberOperations {
+public class PubSubSubscriberTemplate implements PubSubSubscriberOperations, DisposableBean {
 
 	private final SubscriberFactory subscriberFactory;
 
 	private final SubscriberStub subscriberStub;
 
 	private PubSubMessageConverter pubSubMessageConverter = new SimplePubSubMessageConverter();
+
+	private ExecutorService ackExecutor = Executors.newCachedThreadPool();
 
 	/**
 	 * Default {@link PubSubSubscriberTemplate} constructor
@@ -96,6 +100,10 @@ public class PubSubSubscriberTemplate implements PubSubSubscriberOperations {
 		Assert.notNull(pubSubMessageConverter, "The pubSubMessageConverter can't be null.");
 
 		this.pubSubMessageConverter = pubSubMessageConverter;
+	}
+
+	public void setAckExecutor(ExecutorService ackExecutor) {
+		this.ackExecutor = ackExecutor;
 	}
 
 	@Override
@@ -261,6 +269,11 @@ public class PubSubSubscriberTemplate implements PubSubSubscriberOperations {
 						modifyAckDeadline(subscriptionName, ackIds, ackDeadlineSeconds));
 	}
 
+	@Override
+	public void destroy() throws Exception {
+		this.ackExecutor.shutdown();
+	}
+
 	private ApiFuture<Empty> ack(String subscriptionName, Collection<String> ackIds) {
 		AcknowledgeRequest acknowledgeRequest = AcknowledgeRequest.newBuilder()
 				.addAllAckIds(ackIds)
@@ -333,7 +346,7 @@ public class PubSubSubscriberTemplate implements PubSubSubscriberOperations {
 						settableListenableFuture.set(null);
 					}
 				}
-			}, MoreExecutors.directExecutor());
+			}, this.ackExecutor);
 		});
 
 		return settableListenableFuture;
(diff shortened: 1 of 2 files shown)


spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberTemplate.java
```diff
@@ -98,6 +102,10 @@ public void setMessageConverter(PubSubMessageConverter pubSubMessageConverter) {
 		this.pubSubMessageConverter = pubSubMessageConverter;
 	}
 
+	public void setAckExecutor(ExecutorService ackExecutor) {
```
artembilan: Please, consider to expect here an `Executor` instead.

You have to have here an `Assert.notNull()` as well.
And also a property like `boolean ackExecutorSetExplicitly` to be used in that `destroy()`.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberTemplate.java
```diff
@@ -261,6 +269,11 @@ public SubscriberFactory getSubscriberFactory() {
 						modifyAckDeadline(subscriptionName, ackIds, ackDeadlineSeconds));
 	}
 
+	@Override
+	public void destroy() throws Exception {
+		this.ackExecutor.shutdown();
```
artembilan: See my comment in the `setAckExecutor()` and you definitely can't call `shutdown()` on the external executor. That's just not your responsibility to control lifecycle of the injected objects. However with the `!this.ackExecutorSetExplicitly` you can destroy that default internal `Executors.newSingleThreadExecutor()`. Only the point that you have to cast since we would like to expect from end-users a generic `Executor`.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
```diff
@@ -135,6 +135,8 @@ public PubSubSubscriberTemplate pubSubSubscriberTemplate(SubscriberFactory subsc
 			ObjectProvider<PubSubMessageConverter> pubSubMessageConverter) {
 		PubSubSubscriberTemplate pubSubSubscriberTemplate = new PubSubSubscriberTemplate(subscriberFactory);
 		pubSubMessageConverter.ifUnique(pubSubSubscriberTemplate::setMessageConverter);
+		pubSubSubscriberTemplate.setAckExecutor(
+				Executors.newFixedThreadPool(this.gcpPubSubProperties.getSubscriber().getAcknowledgementThreads()));
```
artembilan: We prefer to use `ThreadPoolTaskExecutor` from Spring.
Also you have to consider to make it as a bean to let its lifecycle to work properly, including automatic `shutdown()`.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberTemplate.java
elefeint: I am looking at the `RedisLockRegistry` for inspiration, since it uses this pattern. It seems to me that when creating an executor as part of field initialization, there is a possibility of it never being shutdown [when a custom `Executor` is injected](https://github.com/spring-projects/spring-integration/blob/88c7646ed9eaea2fe755485e1676ad130eb391a9/spring-integration-redis/src/main/java/org/springframework/integration/redis/util/RedisLockRegistry.java#L151), although the object itself will be garbage collected.

I could shut down the pool in the custom executor setter, although that's not ideal since clients would expect setters to not do extra work:

````
	public void setAckExecutor(Executor ackExecutor) {
		Assert.notNull(ackExecutor, "ackExecutor can't be null.");
                ((ExecutorService)this.ackExecutor).shutdown();
		this.ackExecutor = ackExecutor;
		this.ackExecutorSetExplicitly = true;
	}
````

Another option is to remove field initialization, and check instead for whether the field is null in `doBatchedAsyncOperation()`:

````
 doBatchedAsyncOperation(...) {
  if (this.ackExecutor == null) {
    this.ackExecutor = Executors.newCachedThreadPool();
  }
  // use the executor here
}
````

A third possibility is to have a `defaultAckExecutor` object that always gets destroyed, and a `customAckExecutor` that can be user-injected. Then in destroy(), it is safe to always shut down `defaultAckExecutor`. That feels like the cleanest approach to me.

What do you think?

Comment (elefeint) [reply spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberTemplate.java]: I understand what you mean -- that in real world, it won't cause problems, even though it's in the ExecutorService contract that unused objects should be shut down.

On the other hand... we don't typically prevent users from setting template properties after some work has already been done, so getting a new executor after the single-threaded executor started work is weird, but not impossible.

I'll go with `defaultAckExecutor`.

File: spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java (hunk 3 above)
Comment (elefeint) [reply spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java]: I did not realize destroy method inference existed, with Spring automatically calling close()/shutdown() methods.This was cool! Now I don't have to write my own `@PreDestroy` in autoconfig.

I think the only place it's documented is `@Bean` `destroyMethod() `[javadoc](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/context/annotation/Bean.html#destroyMethod--).

Comment (artembilan) [reply spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java]: I don't think so: https://docs.spring.io/spring/docs/5.1.3.RELEASE/spring-framework-reference/core.html#beans-factory-lifecycle

But yes: to have it automatic you need to declare `ThreadPoolTaskExecutor` as a `@Bean`.

File: spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberTemplate.java [hunk 4]
```diff
@@ -69,15 +73,18 @@
...
+		implements PubSubSubscriberOperations, DisposableBean {
 
 	private final SubscriberFactory subscriberFactory;
 
 	private final SubscriberStub subscriberStub;
 
 	private PubSubMessageConverter pubSubMessageConverter = new SimplePubSubMessageConverter();
 
-	private ExecutorService ackExecutor = Executors.newSingleThreadExecutor();
+	private ExecutorService defaultAckExecutor = Executors.newSingleThreadExecutor();
```
artembilan: This definitely can be `final` now.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberTemplate.java
```diff
@@ -69,15 +73,18 @@
...
+
+	private Executor customAckExecutor;
```
artembilan: I'm fully believe that you can assign a `defaultAckExecutor` to this property by default and don't have that ternary logic in the `ack()` method.

elefeint: Yes! That's a nice idea.

PR #1363, opened 2019-01-07
Package info classes

### LINKED ISSUE #1191: Add package-info.java to all production packages in the project to allow better JavaDocs publication

ChengyuanZhao:
fixes #1191
Adds package-info.java for all package directories in production code modules


spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/config/package-info.java
```diff
@@ -0,0 +1,20 @@
+/*
+ * Copyright 2017-2018 the original author or authors.
```
artembilan: The point is the `package-info.java` files go without Copyright:
https://github.com/spring-projects/spring-framework/blob/master/spring-webflux/src/main/java/org/springframework/web/reactive/config/package-info.java

https://github.com/spring-projects/spring-framework/blob/master/spring-beans/src/main/java/org/springframework/beans/package-info.java

I'm not sure why Spring Boot does that...

In that case you have to update all of them into the current year anyway. 😄 

/CC @snicoll

spring-cloud-gcp-core/src/main/java/org/springframework/cloud/gcp/core/package-info.java
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Core support for Spring Cloud GCP that provides credentials, project IDs, and environment meta data.
```
elefeint: "metadata"?

spring-cloud-gcp-core/src/main/java/org/springframework/cloud/gcp/core/util/package-info.java
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Util classes for core support for Spring Cloud GCP.
```
elefeint: "Utilities shared across Spring Cloud GCP projects"?
Or "Core util classes for Spring Cloud GCP"?

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/config/package-info.java
ChengyuanZhao: Our checkstyle file seems to require this header though.. 

I will update the year in the meantime.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/package-info.java
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Core support for Cloud Datastore support in Spring Data.
```
elefeint: Remove second "support"?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/util/package-info.java
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Util classes for Spring Data Cloud Datastore support.
```
elefeint: remove "support"? Or move it as "Util classes supporting Spring Data Cloud Datastore"?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/package-info.java
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Classes support Spring Data Cloud Datastore custom-generated Query Methods.
```
elefeint: "supporting"?

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/config/package-info.java
artembilan: We have this:
```
<suppress files="package-info\.java" checks=".*" />
```

in the `checkstyle-suppressions.xml` if that is interesting...

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/integration/filters/package-info.java
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * File filter classes Google Cloud Storage module for Spring Integration.
```
elefeint: "Google Cloud Storage file filter classes for Spring Integration"?

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/integration/inbound/package-info.java
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Inbound integration classes Google Cloud Storage module for Spring Integration.
```
elefeint: "Google Cloud Storage inbound integration classes for Spring Integration"?

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/integration/outbound/package-info.java
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Outbound integration classes Google Cloud Storage module for Spring Integration.
```
elefeint: "Google Cloud Storage outbound integration classes for Spring Integration."?

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/config/package-info.java
snicoll: @artembilan  I am not sure what I can tell you besides we add a copyright header on _every_ source file. What we do in Spring Boot seems more consistent and the header should only be changed if you update the `package-info.java` content which shouldn't be frequent.

artembilan: OK. We can follow that policy here as well.

My point was that Spring Framework doesn't do that and we don't do that in our other projects which have been started before Spring Boot. So, we tried to follow a consistency dictated by Spring Framework.
If you think Spring Boot code style is much appropriate, I'll keep that in mind and will update my projects respectively when time permits.

Thanks for the feedback, Stephane!

philwebb: FWIW I don't think there was much thought around this other than it was easier to be consistent across all java files rather needing to tweak checkstyle rules. I'm biased, but I quite like the consistency. It's especially annoying in framework that you create a new package and need to remember to delete the copyright header if you've setup your IDE to add it.

BTW, framework also used to have a mix of copyright/non-copyright headers in their package-info files until [I fixed them](https://github.com/spring-projects/spring-framework/commit/e9d1b39aff5fd912ec692ca594d7e02e42318789#diff-26ea4dab5f979c7e596b66deb86905a2). They now have a checkstyle to enforce that they are never there.

artembilan: Good. Since code style on the matter is just to be as efficient as possible (no checkstyle rule, no manual fixes etc.), I would really stay with approach to add Copyright into `package-info.java` and forget.
No further reason to spend time for this discussion 😄

PR #1366, opened 2019-01-09
Stop hard-coding version in User-Agent for metrics

meltsufin:
We found the root cause of incorrect versions being reported, and it's
not automatic version detection. So, hard-coding the version is no
longer needed.

This PR also renames the class to `UserAgentHeaderProvider`.

### DIFF
--- a/spring-cloud-gcp-core/src/main/java/org/springframework/cloud/gcp/core/UserAgentHeaderProvider.java
+++ b/spring-cloud-gcp-core/src/main/java/org/springframework/cloud/gcp/core/UserAgentHeaderProvider.java
@@ -29,18 +29,13 @@ import com.google.api.gax.rpc.HeaderProvider;
  * @author Chengyuan Zhao
  * @author Mike Eltsufin
  */
-public class UsageTrackingHeaderProvider implements HeaderProvider {
-
-	/**
-	 * The hard-coded version string reported with usage info.
-	 */
-	public static final String TRACKING_HEADER_PROJECT_VERSION = "1.1.0.BUILD-SNAPSHOT";
+public class UserAgentHeaderProvider implements HeaderProvider {
 
 	private String userAgent;
 
 	private final Map<String, String> headers;
 
-	public UsageTrackingHeaderProvider(Class clazz) {
+	public UserAgentHeaderProvider(Class clazz) {
 		this.userAgent = computeUserAgent(clazz);
 		this.headers = Collections.singletonMap("User-Agent", this.userAgent);
 	}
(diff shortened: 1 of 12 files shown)


spring-cloud-gcp-core/src/main/java/org/springframework/cloud/gcp/core/UserAgentHeaderProvider.java
```diff
@@ -29,18 +29,13 @@
  * @author Chengyuan Zhao
  * @author Mike Eltsufin
  */
-public class UsageTrackingHeaderProvider implements HeaderProvider {
-
-	/**
-	 * The hard-coded version string reported with usage info.
-	 */
-	public static final String TRACKING_HEADER_PROJECT_VERSION = "1.1.0.BUILD-SNAPSHOT";
+public class UserAgentHeaderProvider implements HeaderProvider {
```
artembilan: Everything is OK, unless I don't like such a breaking change in the point release. 
Would be great to deprecate in the current version and mention in its Javadocs that it i going to be remove in the next `1.2`.

I don't know how Spring Cloud team manages similar situations, but we in our projects stay with the rule as less breaking changes for end-users as possible.

spring-cloud-gcp-core/src/main/java/org/springframework/cloud/gcp/core/UserAgentHeaderProvider.java
```diff
@@ -67,9 +62,9 @@ public String getUserAgent() {
 	private String computeUserAgent(Class clazz) {
 		String[] packageTokens = clazz.getPackage().getName().split("\\.");
 		String springLibrary = "spring-cloud-gcp-" + packageTokens[packageTokens.length - 1];
+		String version = this.getClass().getPackage().getImplementationVersion();
```
artembilan: As long as I remember you decided to stay away from this approach since someone may decide to repackage our jars without versions. See `org.springframework.core.SpringVersion` Javadocs

spring-cloud-gcp-core/src/main/java/org/springframework/cloud/gcp/core/UserAgentHeaderProvider.java
meltsufin: This is really an internal class. Do you think people are relying on it when they modify our auto-configuration? If that's the case, we can just leave the old name. It doesn't matter that much.

spring-cloud-gcp-core/src/main/java/org/springframework/cloud/gcp/core/UserAgentHeaderProvider.java
meltsufin: We thought that many people are repackaging, but that doesn't seem to be the case. Repackaging seems to be very rare, and dealing with it is not worth the maintenance of the hard-coded versions. For example, we forgot to change the hard-coded version for 1.1 RC2 release.

meltsufin: There was an issue in our data pipeline that detected the wrong version sometimes, whereas we thought it was all due to repackaging.

PR #1374, opened 2019-01-11
Remove guava from spanner module

elefeint:
Suggestions and fixes welcome, but don't bother approving.
Extra attention will be needed to formerly ImmutableMap instances -- in some cases, their immutability was not as important as their maintaining insert order.

### DIFF
--- a/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/ConverterAwareMappingSpannerEntityWriter.java
+++ b/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/ConverterAwareMappingSpannerEntityWriter.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017-2018 the original author or authors.
+ * Copyright 2017-2019 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/ConverterAwareMappingSpannerEntityWriter.java
+++ b/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/ConverterAwareMappingSpannerEntityWriter.java
@@ -17,6 +17,9 @@
 package org.springframework.cloud.gcp.data.spanner.core.convert;
 
 import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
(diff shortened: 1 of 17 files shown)


spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/ConverterAwareMappingSpannerEntityWriter.java
```diff
@@ -74,46 +69,41 @@
...
-
-		builder.put(Date.class, (BiFunction<ValueBinder, Date, ?>) ValueBinder::to);
-		builder.put(Boolean.class, (BiFunction<ValueBinder, Boolean, ?>) ValueBinder::to);
-		builder.put(Long.class, (BiFunction<ValueBinder, Long, ?>) ValueBinder::to);
-		builder.put(long.class, (BiFunction<ValueBinder, Long, ?>) ValueBinder::to);
-		builder.put(Double.class, (BiFunction<ValueBinder, Double, ?>) ValueBinder::to);
-		builder.put(double.class, (BiFunction<ValueBinder, Double, ?>) ValueBinder::to);
-		builder.put(String.class, (BiFunction<ValueBinder, String, ?>) ValueBinder::to);
-		builder.put(Timestamp.class,
+		Map<Class<?>, BiFunction<ValueBinder, ?, ?>> map = new LinkedHashMap<>();
```
ChengyuanZhao: Could this (and the other one in this class) now also use MapBuilder from our util class?

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/StructAccessor.java
```diff
@@ -46,74 +46,72 @@
...
+				.put(double[].class, AbstractStructReader::getDoubleArray)
+				.put(long[].class, AbstractStructReader::getLongArray)
+				.put(boolean[].class, AbstractStructReader::getBooleanArray)
+				// Note that Struct.class appears in this map. While we support
+				// converting structs into POJO fields of POJOs, the value in this map is for
+				// the case where the field within the POJO is Struct.
+				.put(Struct.class, Struct::getStruct)
+				.build();
+
+	// TODO: Why is this unused?
```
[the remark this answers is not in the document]
ChengyuanZhao: Ah! this one is safe to delete.

I added it a few months ago along with "singleItemReadMethodMappingIntCol" for a sense of symmetry, but I must've not realized there isn't a use for it (yet). We can add it back at any time, so feel free to delete it now.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/ConverterAwareMappingSpannerEntityWriter.java
elefeint: I think order matters in `singleItemTypeValueBinderMethodMap` -- `findFirstCompatibleSpannerSingleItemNativeType()`
does an ordered search on it:
````
singleItemTypeValueBinderMethodMap.keySet().stream().filter(testFunc)
				.findFirst();
````

ChengyuanZhao: I see!

elefeint:
> Doh! Forgot to request an update for all the Copyrights to 2019 .
> 
> I wonder if we can write some script to do that for us automatically...
> 
> At least I can do it in Gradle

With a pre-commit git hook maybe?

PR #1379, opened 2019-01-14
Exclude io.grpc transitive dependency from zipkin GCP

### LINKED ISSUE #1373: spring-cloud-gcp-starter-pubsub doesn't work with spring-cloud-gcp-starter-trace version higher than 1.0.0.RC
Hi, there

We have been struggling with spring-cloud-gcp-starter-trace.

If we include the dependency of spring-cloud-gcp-starter-trace version higher than 1.0.0.RC, the app does not receive any Cloud Pub/Sub messages.

Here is the minimum sample project that reproduce this problem. https://github.com/shiraji/demo-pubsub

This project does not work with Cloud Pub/Sub in version 1.1.0.RC2. However, once we remove the dependency of spring-cloud-gcp-starter-trace, this project works perfectly.
Also, if we use 1.0.0.RC for all spring-cloud-gcp-starter's, the project works nicely. 

It may be related to the problem that spring-cloud-gcp-starter-trace version 1.0.0.RELEASE does not report to StackDriver Trace. (1.0.0.RC1 works fine)

Thanks in advance.

dzou:
Excludes io.grpc inclusions to resolve #1373.

This is one such solution involving excluding io.grpc packages being brought in from zipkin gcp so that grpc version stay at 1.16.

### DIFF
--- a/spring-cloud-gcp-starters/spring-cloud-gcp-starter-trace/pom.xml
+++ b/spring-cloud-gcp-starters/spring-cloud-gcp-starter-trace/pom.xml
@@ -29,6 +29,14 @@
 					<groupId>com.google.protobuf</groupId>
 					<artifactId>protobuf-java</artifactId>
 				</exclusion>
+				<exclusion>
+					<groupId>io.grpc</groupId>
+					<artifactId>grpc-protobuf</artifactId>
+				</exclusion>
+				<exclusion>
+					<groupId>io.grpc</groupId>
+					<artifactId>grpc-core</artifactId>
+				</exclusion>
 			</exclusions>
 		</dependency>
 		<dependency>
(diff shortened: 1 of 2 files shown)


spring-cloud-gcp-starters/spring-cloud-gcp-starter-trace/pom.xml
```diff
@@ -29,6 +29,14 @@
 					<groupId>com.google.protobuf</groupId>
 					<artifactId>protobuf-java</artifactId>
 				</exclusion>
+				<exclusion>
+					<groupId>io.grpc</groupId>
+					<artifactId>grpc-protobuf</artifactId>
```
meltsufin: You probably should duplicate this in the `autoconfiguration` module.

PR #1380, opened 2019-01-14
unit tests in Datastore two step conversions

elefeint:
This covers the null check discussed in #1375, plus a few more basic read conversion tests.
The tests in `TestDatastoreItemCollections` are more comprehensive, but they are more integration tests that cover both read and write conversions together. 
The new `TwoStepsConversionsTests` is a good place to put smaller unit tests.


elefeint:
This covers the null check discussed in #1375, plus a few more basic read conversion tests.
The tests in TestDatastoreItemCollections are more comprehensive, but they are more integration tests that cover both read and write conversions together.
The new TwoStepsConversionsTests is a good place to put smaller unit tests.

PR #1383, opened 2019-01-15
Updated the auto-configuration to support multiple tracing systems

TYsewyn:
- `spring-cloud-gcp-trace` is no longer depending on `spring-cloud-sleuth-zipkin`.
- `spring.zipkin.messageTimeout` has been replaced by `spring.cloud.gcp.trace.messageTimeout`.

PR should be rerun/merged after https://github.com/spring-cloud/spring-cloud-sleuth/pull/1181.


spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/trace/StackdriverTraceAutoConfiguration.java
```diff
@@ -118,8 +132,19 @@ public ManagedChannel stackdriverSenderChannel() {
 				.build();
 	}
 
-	@Bean
-	@ConditionalOnMissingBean
```
artembilan: Any pointers why we are not conditional here any more?

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/trace/StackdriverTraceAutoConfiguration.java
```diff
@@ -118,8 +132,19 @@ public ManagedChannel stackdriverSenderChannel() {
...
+	@Bean(REPORTER_BEAN_NAME)
+	public Reporter<Span> stackdriverReporter(ReporterMetrics reporterMetrics,
+			GcpTraceProperties trace, @Qualifier(SENDER_BEAN_NAME) Sender sender) {
+		return AsyncReporter.builder(sender).queuedMaxSpans(1000) // historical
+																	// constraint. Note:
+																	// AsyncReporter
+																	// supports memory
+																	// bounds
```
artembilan: Suspicious indents. Isn't it?

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/trace/StackdriverTraceAutoConfiguration.java
TYsewyn: No you're correct, that needs to be `@ConditionalOnMissingBean(name = REPORTER_BEAN_NAME)` and `@ConditionalOnMissingBean(name = SENDER_BEAN_NAME)`.

So the idea behind this is that whenever you add a tracing system (GCP Trace, Zipkin or smth else) to your classpath you'll always need to have a `Reporter<Span>` and `Sender`.
Behind the scenes the `TraceAutoConfiguration` will create a composite so you can send your traces to multiple systems "at the same time". (quoted because it's async)
That's also the reason why those 2 beans need to have a fixed name: to be able to override the autoconfig, otherwise you'll send your traces twice to the same system.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/trace/StackdriverTraceAutoConfiguration.java
TYsewyn: Let me check what has happened here, I guess it's to auto formatting that changed the comment.

TYsewyn:
> LGTM. Not sure why do we have to wait for the PR in Sleuth since we don't depend on it any more...

In the current setup you're still depending on a bean that has been moved from `ZipkinAutoConfiguration` (sleuth-zipkin) to `TraceAutoConfiguration` (sleuth-core) in the other PR. If all goes well @marcingrzejszczak and I are discussing that change tomorrow.

artembilan:
Good, @TYsewyn. Thank you for clarification!

I still wonder if this make a right to be merged for the current GA milestone this Thursday...

TYsewyn:
The changes in spring-cloud-sleuth are marked for 2.1.0.RELEASE, so if we can review that and update the docs in both repos by tomorrow EOD I think we can manage that? 😅😬

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/trace/StackdriverTraceAutoConfiguration.java
```diff
@@ -165,12 +190,6 @@ public Sender stackdriverSender(GcpTraceProperties traceProperties,
 				.build();
 	}
 
-	@Bean
-	@ConditionalOnMissingBean
-	public BytesEncoder<Span> spanBytesEncoder() {
```
artembilan: What dose make you to remove this bean altogether?

TYsewyn: That bean was needed to create a `Reporter` (because it was heavily depending on sleuth-zipkin): https://github.com/spring-cloud/spring-cloud-sleuth/blob/master/spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin2/ZipkinAutoConfiguration.java#L77

In the new configuration there is a dedicated reporter for Stackdriver so we don't need to have this bean anymore.
What was previously `.metrics(reporterMetrics).build(spanBytesEncoder);` (sleuth-zipkin) is now `.metrics(reporterMetrics).build(StackdriverEncoder.V1);` (gcp-trace)

spring-cloud-gcp-autoconfigure/pom.xml
```diff
@@ -231,6 +225,13 @@
             <version>2.5</version>
             <scope>test</scope>
         </dependency>
+        <!--<dependency>-->
+            <!--<groupId>io.grpc</groupId>-->
+            <!--<artifactId>grpc-testing</artifactId>-->
+            <!--<version>1.16.1</version>-->
+            <!--<scope>test</scope>-->
+        <!--</dependency>-->
```
artembilan: Why is this commented out since you just say we can simply remove it?

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/trace/StackdriverTraceAutoConfiguration.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017-2018 the original author or authors.
+ * Copyright 2017-2010 the original author or authors.
```
artembilan: Something is wrong here...

TYsewyn:
My pleasure! Which release are you referring too though? That one of Spring Cloud Sleuth I hope?

meltsufin:
...And Spring Cloud GCP. We release together with the rest of Spring Cloud.

TYsewyn:
Problem is that you’ll need to merge one of the 2 PRs before you can release GCP.
This one adds extra functionality, the other one fixes your test failure.
Except when you skip your tests during your release build 😅

PR #1372, opened 2019-01-11
Remove guava from datastore

elefeint:

PR #1385, opened 2019-01-15
Simplify updating license year on modified files

elefeint:
New script that checks all added/modified/created files and updates license end year to the current year.


elefeint:
I'll make it not run automatically and instead be an invokable maven target -- @meltsufin pointed out that `bash` presence cannot be assumed.

meltsufin:
Separate from automatic fix (that can be run manually), I think we should just have a check that detects when the copyright header has the wrong date. The check can just run automatically on Travis build.

elefeint:
I removed the maven changes entirely. The script is useful, but can be run from command line just as well; does not need to be a maven target if it's not going to be run automatically.
Tested it on #1372, seems to work.

meltsufin:
@elefeint I still see changes in `pom.xml`.

meltsufin:
@elefeint Maybe drop a line about it in https://github.com/spring-cloud/spring-cloud-gcp/blob/master/CONTRIBUTING.adoc?

PR #1399, opened 2019-01-23
Datastore use original Reference annotation

### LINKED ISSUE #1391: Support `org.springframework.data.annotation.Reference` annotation
Currently we have our own `@Reference` annotation.
We should deprecate our own version and switch to the one in `spring-data-commons`.

ChengyuanZhao:
fixes #1391
uses spring data core reference annotation


ChengyuanZhao:
@dmitry-s @meltsufin PTAL, this one turned out to be a very straightforward replacement.

meltsufin:
Please fix the checkstyle though.

PR #1404, opened 2019-01-24
fix test

dmitry-s:


spring-cloud-gcp-samples/spring-cloud-gcp-data-datastore-sample/src/main/java/com/example/Album.java
```diff
@@ -27,7 +27,7 @@
  * @author Dmitry Solomakha
  */
 @Entity
-public class Album {
+public class Album implements Comparable<Album>{
```
elefeint: it might be simpler to have Singer take a list of albums instead of a set instead of trying to make the Set ordered.

dmitry-s: That would require a lot of changes.

PR #1412, opened 2019-01-29
Remove unnecessary readme file

ChengyuanZhao:
It looks like we can't link to the root README.adoc file to show its contents. This file was originally added here to fit the convention of other repos like spring-cloud-aws and spring-cloud-netflix, but since has had its contents moved to the root README.adoc file.

Removing this file's contents (**or having it present but with a broken link**) doesn't seem to break or impact the docs generation via `mvn clean install -P=docs` command.


ChengyuanZhao:
@dzou @artembilan My apologies, but the presence of a "README" file seems to be required for the build with -P=docs to succeed, so I added a regular clickable link to the root README file.

ChengyuanZhao:
Ah so that's why it seems to override that root README every time! 

The problem with that is that means we need to manually run -P=docs before we make the commit or our root page README will be out of date... @dzou @meltsufin

artembilan:
See https://github.com/spring-cloud/spring-cloud-gcp/pull/1413

@marcingrzejszczak , 

pay attention, please, to the comment above:
> The problem with that is that means we need to manually run -P=docs before we make the commit or our root page README shown on github.com will be out of date..

I also find this a bit frustrating to run Maven to have a root README up-to-date 😢

meltsufin:
Closing in favor of #1413.

PR #1414, opened 2019-01-29
Suppress `LineLength` Checkstyle rule

artembilan:
Looks like this rule doesn't work well to calculate a proper line length


artembilan:
See http://checkstyle.sourceforge.net/config_sizes.html#LineLength for more info.

> The calculation of the length of a line takes into account the number of expanded spaces for a tab character ('\t'). The default number of spaces is 8. To specify a different number of spaces, the user can set TreeWalker property tabWidth which applies to all Checks, including LineLength; or can set property tabWidth for LineLength alone. 

This is the cause: https://github.com/spring-cloud/spring-cloud-build/commit/0d5e152e7a38a51dba172771bb4548598c96c363

/CC @marcingrzejszczak

marcingrzejszczak:
Thanks for the analysis. I'll setup the tab width tomorrow.

PR #1415, opened 2019-01-29
Reference loop fix

### LINKED ISSUE #1389: Datastore bi-directional relationships cause infinite recursion
Is there any way to implement OneToMany or ManyToMany relationships in Google DataStore?

```
    @Entity
    public class Entity {
        @Id
        private Long id;
        @Reference
        private List<SubEntity> subEntity;
    
    
        public Building(List<SubEntity> subEntity) {
            this.subEntity = subEntity;
        }
    
        //getters and setters
    }
    
    @Entity
    public class SubEntity {
        @Id
        private Long id;
    
        @Reference
        private Entity entity;
    
        //getters and setters
    }
```

It's not working

dmitry-s:
fixes #1389


spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
```diff
@@ -608,4 +693,23 @@ private DatastoreReaderWriter getDatastoreReadWriter() {
 			throw new DatastoreDataException("Property matchers are not supported");
 		}
 	}
+
+	/**
+	 * Class to hold caches for read and conversion.
+	 *
+	 * @author Dmitry Solomakha
+	 */
+	class ReadContext {
```
meltsufin: Since you already have this class, why not make it a bit smarter?
For example:
```java
Object getConvertedEntity(Key key)
```
The method would automatically convert if needed, or return `null`.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
```diff
@@ -168,11 +194,39 @@ public long count(Class<?> entityClass) {
...
+
+		ArrayList<T> result = new ArrayList<>();
+		for (Key key : keys) {
+			T convertedEntity;
+			if (context.converted(key)) {
+				convertedEntity = (T) context.getConvertedEntity(key);
+			}
+			else {
+				Object entity = context.getReadEntity(key);
+				convertedEntity = convertEntitiesForRead(Collections.singletonList((BaseEntity) entity).iterator(), entityClass,
```
ChengyuanZhao: Because it doesn't look like you're adding to the context in each iteration of this for-loop, I don't think you need to create singleton lists for convertEntitiesForRead. The loop inside convertEntitiesForRead looks like it will work just fine if you gave it all the entities at the same time.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
```diff
@@ -113,21 +116,44 @@ public DatastoreEntityConverter getDatastoreEntityConverter() {
...
 	}
 
 	@Override
 	public <T> Iterable<T> saveAll(Iterable<T> entities, Key... ancestors) {
-		getDatastoreReadWriter().put(StreamSupport.stream(entities.spliterator(), false)
-				.map((entity) -> convertToEntityForSave(entity, ancestors)).toArray(Entity[]::new));
+		return saveAll(entities, new HashSet<>(), ancestors);
+	}
+
+	private <T> Iterable<T> saveAll(Iterable<T> entities, Set<Key> persisted, Key... ancestors) {
```
meltsufin: Should add a test that this method is only called once, when there are cycles in the object graph.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
```diff
@@ -168,11 +194,39 @@ public long count(Class<?> entityClass) {
...
+
+	private <T> Collection<T> findAllById(Iterable<?> ids, Class<T> entityClass, ReadContext context) {
+		List<Key> keys = StreamSupport.stream(ids.spliterator(), false).map((id) -> getKeyFromId(id, entityClass))
+				.collect(Collectors.toList());
+		List<Key> missingKeys = keys.stream().filter(context::notCached).collect(Collectors.toList());
+
+		if (!missingKeys.isEmpty()) {
+			List<Entity> entities = getDatastoreReadWriter().fetch(missingKeys.toArray(new Key[] {}));
+
+			for (int i = 0; i < missingKeys.size(); i++) {
```
meltsufin: Add an assertion to make sure that the two collections are of the same size.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
```diff
@@ -168,11 +194,39 @@ public long count(Class<?> entityClass) {
...
+				Key key = missingKeys.get(i);
+				context.putReadEntity(key, entities.get(i));
+			}
+		}
+
+		ArrayList<T> result = new ArrayList<>();
+		for (Key key : keys) {
```
meltsufin: Use stream here instead of loop?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
```diff
@@ -407,37 +461,65 @@ private void validateKey(Object entity, PathElement ancestorPE) {
...
 		return results;
 	}
 
-	private <T> void convertEntityResolveDescendantsAndReferences(Class<T> entityClass,
-			List<T> results, DatastorePersistentEntity datastorePersistentEntity,
-			BaseEntity entity) {
+	private <T> T convertEntityResolveDescendantsAndReferences(Class<T> entityClass,
+			DatastorePersistentEntity datastorePersistentEntity,
+			BaseEntity entity, ReadContext context) {
 		T convertedObject = this.datastoreEntityConverter.read(entityClass, entity);
```
meltsufin: Check if entity has a key, and then look it up in the cache first before doing conversion.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
```diff
@@ -460,12 +543,13 @@ else if (referencePersistentProperty.isCollectionLike()) {
...
 								findAllById(
 										keyValues.stream().map(Value::get).collect(Collectors.toList()),
-										referencedType),
+										referencedType, context),
 								referencePersistentProperty.getType(),
 								referencedType);
 			}
 			else {
-				referenced = findById(entity.getKey(fieldName), referencePersistentProperty.getType());
+				Key key = entity.getKey(fieldName);
```
meltsufin: You can do this inline, in the next line.

spring-cloud-gcp-data-datastore/src/test/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplateTests.java
```diff
@@ -363,18 +364,40 @@ public void findByIdTest() {
...
-		when(this.datastore.get(eq(this.key1), eq(this.key2)))
-				.thenReturn(Arrays.asList(this.e1, this.e2).iterator());
+		when(this.datastore.fetch(eq(this.key1), eq(this.key2)))
+				.thenReturn(Arrays.asList(this.e1, this.e2));
 		List<Key> keys = Arrays.asList(this.key1, this.key2);
 		assertThat(this.datastoreTemplate.findAllById(keys, TestEntity.class)).containsExactly(this.ob1, this.ob2);
 	}
 
+	@Test
+	public void findAllByIdReferenceConsistencyTest() {
```
meltsufin: Can we add more unit tests here that test the cycles as well?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
```diff
@@ -113,21 +117,42 @@ public DatastoreEntityConverter getDatastoreEntityConverter() {
...
+		for (T entity : entities) {
+			Key key = getKey(entity, true, ancestors);
+			if (persisted.contains(key)) {
+				continue;
+			}
+			persisted.add(key);
+			entitiesForSave.add(convertToEntityForSave(entity, persisted, ancestors));
+		}
+		if (!entitiesForSave.isEmpty()) {
+			getDatastoreReadWriter().put(entitiesForSave.toArray(new Entity[0]));
```
meltsufin: Why not avoid this whole conversion from List to array?
```java
		Entity[] entitiesForSave = StreamSupport.stream(entities.spliterator(), false)
				.filter(persisted::contains)
				.peek(entity -> persisted.add(getKey(entity, true, ancestors)))
				.map(entity -> convertToEntityForSave(entity, persisted, ancestors))
				.toArray(Entity[]::new);
		if (entitiesForSave.length > 0) {
			getDatastoreReadWriter().put(entitiesForSave);
		}
		return entities;
```

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
```diff
@@ -113,21 +117,42 @@ public DatastoreEntityConverter getDatastoreEntityConverter() {
...
+		}
 		return entities;
```
meltsufin: This might be going back a bit, but what's the point of returning the same variable that is sent in as the argument??
Wouldn't returning the actual saved `Entity[]` be more useful?

ChengyuanZhao: The Spring Data Repository's save method has a return type that is `T` (the original pojo type). The template save method's return type has always been the same to support the repo's save method.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
ChengyuanZhao: It looks like in both cases you need to do a `.toArray(Entity[]::new);` somewhere in there

meltsufin: I guess you then have to do something like this:
```java
	Entity[] entitiesForSave = StreamSupport.stream(entities.spliterator(), false)
				.map(entity -> new AbstractMap.SimpleEntry<Key, T>(getKey(entity, true, ancestors), entity))
				.filter(entry -> persisted.contains(entry.getKey()))
				.peek(entry -> persisted.add(entry.getKey()))
				.map(entry -> convertToEntityForSave(entry.getValue(), persisted, ancestors))
				.toArray(Entity[]::new);
		if (entitiesForSave.length > 0) {
			getDatastoreReadWriter().put(entitiesForSave);
		}
		return entities;
```
A bit more ugly, but yeah.

dmitry-s: In this case I think it looks less ugly with a loop. The overhead of using ArrayList is not big, probably comparable with the stream overhead.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
dmitry-s: True. Do you prefer making that method return nothing? Returning `Entity[]` isn't more useful.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
```diff
@@ -113,21 +117,42 @@ public DatastoreEntityConverter getDatastoreEntityConverter() {
...
 
 	@Override
 	public <T> T save(T instance, Key... ancestors) {
-		getDatastoreReadWriter().put(convertToEntityForSave(instance, ancestors));
-		return instance;
+		return save(instance, new HashSet<>(), ancestors);
+	}
+
+	private <T> T save(T instance, Set<Key> persisted, Key... ancestors) {
+		Iterable<T> iterable = saveAll(Collections.singletonList(instance), persisted, ancestors);
```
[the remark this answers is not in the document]
dmitry-s: True, we can just return `instance`.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
meltsufin: Returning Entity I think would be useful if the public method also returned Entity[] or List<Entity>

meltsufin: There might be some use-cases, but let's tackle this in a different PR. Can you please create an issue for this?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
meltsufin: It looks like 2x working memory overhead with the non-streaming version.

PR #1422, opened 2019-01-31
Test ORIGINAL_MESSAGE gets added in all ack modes

elefeint:
Added tests for `GcpPubSubHeaders.ORIGINAL_MESSAGE` as followup to #1409 and simplified existing tests.

I've made some stylistic choices here that are different than the bulk of _spring-cloud-gcp_ code. If anyone hates `@Mock` annotations or prefixing mock objects with _mock_, speak up on this PR. I like both because they make it easy to separate real objects from mocks visually, but I'll go with whatever style the team settles on.


elefeint:
Added tests for GcpPubSubHeaders.ORIGINAL_MESSAGE as followup to #1409 and simplified existing tests.

Issue #1283, opened 2018-12-04
Spanner entitylistener and auditing


ChengyuanZhao:
There have been user requests to provide something similar to JPA and Hibernate style auditing functionality.

ChengyuanZhao:
- [ ] make template publisher-aware
- [ ] events objects
- [ ] default listener implementation
- [ ] sample and refdocs
- [ ] integration test

PR #1433, opened 2019-02-04
Spanner mutation events

### LINKED ISSUE #1377: Spanner : Trasnactions + updates to table
Hi @ChengyuanZhao  

We have a usecase where we have to read the data and update values on the row. 
1)Can you confirm what kind of locking is done for Read transactions vs update transactions ?
2)Is it possibel to specify for an update to fail if the last updated value on teh current row to update has elapsed teh value the user can provider. It falls in sync with versioning column on db if we use

ChengyuanZhao:
related #1283
implements events for mutations 

fixes #1377

### DIFF
--- a/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
+++ b/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
@@ -56,8 +56,10 @@ import org.springframework.cloud.gcp.data.spanner.core.mapping.SpannerPersistent
 import org.springframework.cloud.gcp.data.spanner.core.mapping.SpannerPersistentProperty;
 import org.springframework.cloud.gcp.data.spanner.core.mapping.event.AfterDeleteEvent;
 import org.springframework.cloud.gcp.data.spanner.core.mapping.event.AfterExecuteDmlEvent;
+import org.springframework.cloud.gcp.data.spanner.core.mapping.event.AfterSaveEvent;
 import org.springframework.cloud.gcp.data.spanner.core.mapping.event.BeforeDeleteEvent;
 import org.springframework.cloud.gcp.data.spanner.core.mapping.event.BeforeExecuteDmlEvent;
+import org.springframework.cloud.gcp.data.spanner.core.mapping.event.BeforeSaveEvent;
 import org.springframework.cloud.gcp.data.spanner.repository.query.SpannerStatementQueryExecutor;
 import org.springframework.context.ApplicationEvent;
 import org.springframework.context.ApplicationEventPublisher;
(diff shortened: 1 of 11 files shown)


spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/mapping/event/AfterDeleteEvent.java
```diff
@@ -0,0 +1,45 @@
+/*
```
dzou: Just to clarify, what is the need for the 4 classes `BeforeSaveEvent`, `AfterSaveEvent`, etc. over just instantiating the parent class?

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/mapping/event/MutationEvent.java
```diff
@@ -0,0 +1,79 @@
...
+	 */
+	public List<Mutation> getMutations() {
+		return (List<Mutation>) getSource();
+	}
+
+	/**
+	 * Get the list of entities that needed to be deleted.
+	 * @return This may be {@code null} depending on the type of delete request.
+	 */
+	public Iterable getTargetEntities() {
```
dzou: Do you get a warning for using Iterable without a type? (like `Iterable<T>`)

Also without a type the user must cast their object during iteration if they call `getTargetEntities()`, which may not be ideal.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/mapping/event/AfterDeleteEvent.java
ChengyuanZhao: I suppose you could have a setting in `SaveEvent` to indicate if it was a "before" or an "after" event. However it looks like the Spring Data way is to use separate classes for the before and after. In future cases they might actually hold different data, too.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/mapping/event/MutationEvent.java
ChengyuanZhao: We wouldn't be able to put a type inside of it anyway because there is source that `T`. The places where we generate the events and where we consume the events have no idea what `T` is. the target entities are always just `Object`s from our perspective.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
```diff
@@ -219,87 +223,116 @@ public long executeDmlStatement(Statement statement) {
 
 	@Override
 	public void insert(Object object) {
-		applyMutations(this.mutationFactory.insert(object));
+		List<Mutation> mutations = this.mutationFactory.insert(object);
+		emitSaveEventsApplyMutations(mutations, Collections.singletonList(object), null);
```
meltsufin: What's the purpose of the local variable `mutations`? Can't you just do it inline?

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
```diff
@@ -219,87 +223,116 @@ public long executeDmlStatement(Statement statement) {
...
 	}
 
 	@Override
 	public void upsert(Object object, Set<String> includeProperties) {
-		applyMutations(this.mutationFactory.upsert(object, includeProperties));
+		List<Mutation> mutations = this.mutationFactory.upsert(object, includeProperties);
+		emitSaveEventsApplyMutations(mutations, Collections.singletonList(object), includeProperties);
+	}
+
+	private void emitSaveEventsApplyMutations(List<Mutation> mutations, Iterable entities,
```
meltsufin: How about just `applySaveMutations`. The fact that it emits the save event, is an implementation detail.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
```diff
@@ -219,87 +223,116 @@ public long executeDmlStatement(Statement statement) {
...
+			Set<String> includeProperties) {
+		maybeEmitEvent(new BeforeSaveEvent(mutations, entities, includeProperties));
+		applyMutations(mutations);
+		maybeEmitEvent(new AfterSaveEvent(mutations, entities, includeProperties));
 	}
 
 	@Override
 	public void delete(Object entity) {
-		applyMutations(Collections.singletonList(this.mutationFactory.delete(entity)));
+		List<Mutation> mutations = Collections.singletonList(this.mutationFactory.delete(entity));
```
meltsufin: Again, why not just do it inline?

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
```diff
@@ -219,87 +223,116 @@ public long executeDmlStatement(Statement statement) {
...
 	@Override
 	public void delete(Class entityClass, KeySet keys) {
-		applyMutations(Collections
-				.singletonList(this.mutationFactory.delete(entityClass, keys)));
+		List<Mutation> mutations = Collections
+				.singletonList(this.mutationFactory.delete(entityClass, keys));
+		emitDeleteEventsApplyMutations(entityClass, keys, mutations);
+	}
+
+	private void emitDeleteEventsApplyMutations(Class entityClass, KeySet keys, List<Mutation> mutations) {
```
meltsufin: `applyDeleteMutations`?

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
ChengyuanZhao: good point. moved inline (and other instances too).

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
ChengyuanZhao: renamed.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
ChengyuanZhao: Renamed.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
```diff
@@ -219,87 +223,116 @@ public long executeDmlStatement(Statement statement) {
...
+		KeySet keys = KeySet.newBuilder().addKey(key).build();
+		emitDeleteEventsApplyMutations(entityClass, keys, mutations);
 	}
 
 	@Override
 	public void delete(Class entityClass, KeySet keys) {
-		applyMutations(Collections
-				.singletonList(this.mutationFactory.delete(entityClass, keys)));
+		List<Mutation> mutations = Collections
+				.singletonList(this.mutationFactory.delete(entityClass, keys));
```
dmitry-s: maybe make `mutationFactory.delete` return list so you don't have to manually wrap it?

PR #1443, opened 2019-02-05
Spanner load events

ChengyuanZhao:
related #1283 
publishing events for load operations.

### DIFF
--- a/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
+++ b/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
@@ -56,6 +56,7 @@ import org.springframework.cloud.gcp.data.spanner.core.mapping.SpannerPersistent
 import org.springframework.cloud.gcp.data.spanner.core.mapping.SpannerPersistentProperty;
 import org.springframework.cloud.gcp.data.spanner.core.mapping.event.AfterDeleteEvent;
 import org.springframework.cloud.gcp.data.spanner.core.mapping.event.AfterExecuteDmlEvent;
+import org.springframework.cloud.gcp.data.spanner.core.mapping.event.AfterLoadEvent;
 import org.springframework.cloud.gcp.data.spanner.core.mapping.event.AfterSaveEvent;
 import org.springframework.cloud.gcp.data.spanner.core.mapping.event.BeforeDeleteEvent;
 import org.springframework.cloud.gcp.data.spanner.core.mapping.event.BeforeExecuteDmlEvent;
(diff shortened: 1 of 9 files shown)


spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
```diff
@@ -185,15 +188,16 @@ public long executeDmlStatement(Statement statement) {
...
 	}
 
 	@Override
 	public <T> List<T> query(Class<T> entityClass, Statement statement,
 			SpannerQueryOptions options) {
-		return mapToListAndResolveChildren(executeQuery(statement, options), entityClass,
-				(options != null) ? options.getIncludeProperties() : null,
-				options != null && options.isAllowPartialRead());
+		List<T> entitites = queryAndResolveChildren(entityClass, statement, options);
+		maybeEmitEvent(new AfterLoadEvent(entitites, statement, null, options, null));
```
meltsufin: For updates, we made a distinction between the ones made via queries (DML) and those done via the API. Why is this kind of distinction not relevant here?

ChengyuanZhao: I took a moment to think and i hadn't been thinking about it that way actually. I separated mutations-based writes and DML-based writes because they're so different in terms of what info we have about them. For the mutation-based writes we have nice and discrete per-row mutations, the entities that triggered them, and the relevant keys. On the other hand, DML is basically a black box; the user could've written anything in that statement and the actual event object itself has little in common with the mutation-based events. 

For the query and key-based reads, they still do have enough in common that I think they can belong in the same event type.

meltsufin: I think one key difference is the existence of the statement.

Compare read vs query invocation for the event emission:
```
maybeEmitEvent(new AfterLoadEvent(entities, null, options, null, keys));
```
```
maybeEmitEvent(new AfterLoadEvent(entitites, statement, null, options, null));
```

Based on the nulls it seems obvious that these are different events and we're trying to merge them into one class.

ChengyuanZhao: I see what you mean there. 
The query and the read don't have the statement or key or options in common, but they both do share the entities property and originally I felt that was enough to justify putting them together.
Actually the main thing I thought was from the user's perspective if they would want to receive separate events for query-based or key-based reads. 

I can separate them though :)

ChengyuanZhao: @meltsufin changes made!

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/mapping/event/AfterReadEvent.java
```diff
@@ -0,0 +1,94 @@
...
+
+import org.springframework.cloud.gcp.data.spanner.core.SpannerReadOptions;
+import org.springframework.context.ApplicationEvent;
+
+/**
+ * This event read operation on Cloud Spanner.
+ *
+ * @author Chengyuan Zhao
+ */
+public class AfterReadEvent extends ApplicationEvent {
```
meltsufin: Did you mean to extend `LoadEvent`?

ChengyuanZhao: @meltsufin fixed :) PTAL

PR #1445, opened 2019-02-05
Add support for using emulator with Datastore

### LINKED ISSUE #1430: Datastore emulator support
Please add a possibility to set emulator host for datastore (as it's implemented for PubSub). Also, if emulator host is set, instantiate NoCredentialsProvider instead of DefaultCredentialsProvider

dzou:
Add support for an application property allowing the user to connect to a Datastore emulator.

Fixes #1430.

### DIFF
--- a/docs/src/main/asciidoc/datastore.adoc
+++ b/docs/src/main/asciidoc/datastore.adoc
@@ -64,18 +64,12 @@ The following configuration options are available:
 |===
 | Name | Description | Required | Default value
 | `spring.cloud.gcp.datastore.enabled` | Enables the Cloud Datastore client | No | `true`
-| `spring.cloud.gcp.datastore.project-id` | GCP project ID where the Google Cloud Datastore API is hosted, if different from the one in the <<spring-cloud-gcp-core,Spring Cloud GCP Core Module>>
-| No |
-| `spring.cloud.gcp.datastore.credentials.location` | OAuth2 credentials for authenticating with the
-Google Cloud Datastore API, if different from the ones in the
-<<spring-cloud-gcp-core,Spring Cloud GCP Core Module>> | No |
-| `spring.cloud.gcp.datastore.credentials.encoded-key` | Base64-encoded OAuth2 credentials for authenticating with the
-Google Cloud Datastore API, if different from the ones in the
-<<spring-cloud-gcp-core,Spring Cloud GCP Core Module>> | No |
-| `spring.cloud.gcp.datastore.credentials.scopes` |
-https://developers.google.com/identity/protocols/googlescopes[OAuth2 scope] for Spring Cloud GCP
-Cloud Datastore credentials | No | https://www.googleapis.com/auth/datastore
+| `spring.cloud.gcp.datastore.project-id` | GCP project ID where the Google Cloud Datastore API is hosted, if different from the one in the <<spring-cloud-gcp-core,Spring Cloud GCP Core Module>> | No |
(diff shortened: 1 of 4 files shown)


spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/datastore/GcpDatastoreAutoConfiguration.java
```diff
@@ -65,6 +66,8 @@
 
 	private final Credentials credentials;
 
+	private final Optional<String> emulatorHost;
```
meltsufin: I think going with the plain old nullable `String` instead of `Optional` will make things less verbose.

docs/src/main/asciidoc/datastore.adoc
```diff
@@ -64,18 +64,12 @@ The following configuration options are available:
...
+| `spring.cloud.gcp.datastore.credentials.location` | OAuth2 credentials for authenticating with the Google Cloud Datastore API, if different from the ones in the <<spring-cloud-gcp-core,Spring Cloud GCP Core Module>> | No |
+| `spring.cloud.gcp.datastore.credentials.encoded-key` | Base64-encoded OAuth2 credentials for authenticating with the Google Cloud Datastore API, if different from the ones in the <<spring-cloud-gcp-core,Spring Cloud GCP Core Module>> | No |
+| `spring.cloud.gcp.datastore.credentials.scopes` | https://developers.google.com/identity/protocols/googlescopes[OAuth2 scope] for Spring Cloud GCP Cloud Datastore credentials | No | https://www.googleapis.com/auth/datastore
 | `spring.cloud.gcp.datastore.namespace` | The Cloud Datastore namespace to use | No | the Default namespace of Cloud Datastore in your GCP project
+| `spring.cloud.gcp.datastore.emulator-host` | The `hostname:port` of the https://cloud.google.com/datastore/docs/tools/datastore-emulator[Datastore Emulator] to connect to | No |
```
ChengyuanZhao: It looks like this setting is just for a "host" that will be used, but just so happens that we intend it to be used for an "emulator-host". Would it make sense to just call this setting "host"?

meltsufin: This crossed my mind too, but seeing that with Pub/Sub we have the same situation and we decided to name it emulator-host, I think we can keep this way here too for consistency. For now, there is no reason to set it other than to point to an emulator.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/datastore/GcpDatastoreAutoConfiguration.java
dzou: Done.

dzou:
Changed to a null check and also added to Datastore autoconfiguration tests.

Awaiting the verdict on #1430 of whether or not to provide the NoCredentials provider

lucasoares:
Hello. I have a question and I don't know if I should open another ticket.

Google libraries have a good class to start the emulator through java. For datastore this class is `LocalDatastoreHelper.java`.

With this new support for datastore-emulator the host must have to be defined statically in `application.properties`. The problem is the emulator helper starts search for a random free port to start the emulator and the datastore connection for spring data cannot connect to the datastore instance.

Do you guys have any kind of helper similar to `LocalDatastoreHelper.java` that uses the `spring.cloud.gcp.datastore.emulator-host` properties?

Issue #1437, opened 2019-02-05
Add missing code snippets in reference documentation


meltsufin:
Many developers prefer code snippets over reading documentation prose. Although, our reference documentation includes a lot of code snippets, many sections only have the prose describing classes and methods, but no examples. 

A few cases identified so far:
- [ ] Subscribing to a subscription
- [ ] Pulling messages from a subscription
- [ ] Tracing section
- [ ] Cloud Foundry
- [ ] Kotlin Support

PR #1450, opened 2019-02-06
improve documentation

dmitry-s:
part of #1437

### DIFF
--- a/docs/pom.xml
+++ b/docs/pom.xml
@@ -12,6 +12,16 @@
 	<artifactId>spring-cloud-gcp-docs</artifactId>
     <packaging>pom</packaging>
     <name>Spring Cloud GCP Documentation</name>
+	<dependencies>
+		<dependency>
+			<groupId>org.springframework.cloud</groupId>
+			<artifactId>spring-cloud-gcp-starter-pubsub</artifactId>
+		</dependency>
+		<dependency>
+			<groupId>org.springframework.boot</groupId>
+			<artifactId>spring-boot-starter-web</artifactId>
+		</dependency>
+	</dependencies>
 
 	<properties>
 		<docs.main>spring-cloud-gcp</docs.main>
--- a/docs/pom.xml
+++ b/docs/pom.xml
@@ -11,7 +11,7 @@
 
 	<artifactId>spring-cloud-gcp-docs</artifactId>
     <packaging>pom</packaging>
-    <name>Spring Cloud GCP Documentation</name>
+	<name>Spring Cloud GCP Documentation</name>
 	<dependencies>
 		<dependency>
 			<groupId>org.springframework.cloud</groupId>
--- a/docs/pom.xml
+++ b/docs/pom.xml
@@ -10,7 +10,7 @@
 	<modelVersion>4.0.0</modelVersion>
 
 	<artifactId>spring-cloud-gcp-docs</artifactId>
-    <packaging>pom</packaging>
+	<packaging>pom</packaging>
 	<name>Spring Cloud GCP Documentation</name>
 	<dependencies>
 		<dependency>
--- a/docs/src/main/asciidoc/pubsub.adoc
+++ b/docs/src/main/asciidoc/pubsub.adoc
@@ -44,7 +44,7 @@ Here is an example of how to publish a message to a Google Cloud Pub/Sub topic:
 
 [source,java,indent=0]
 ----
-include::../../../src/test/java/org/springframework/cloud/gcp/autoconfigure/pubsub/it/PubSubTemplateDocumentationTests.java[tag=publish]
+include::../../test/java/org/springframework/cloud/gcp/autoconfigure/pubsub/it/PubSubTemplateDocumentationTests.java[tag=publish]
 ----
 
 By default, the `SimplePubSubMessageConverter` is used to convert payloads of type `byte[]`, `ByteString`, `ByteBuffer`, and `String` to Pub/Sub messages.
@@ -60,7 +60,7 @@ asynchronously and passed to a user provided message handler.
 Subscribe to a subscription with a message handler:
 [source,java,indent=0]
 ----
-include::../../../src/test/java/org/springframework/cloud/gcp/autoconfigure/pubsub/it/PubSubTemplateDocumentationTests.java[tag=subscribe]
+include::../../test/java/org/springframework/cloud/gcp/autoconfigure/pubsub/it/PubSubTemplateDocumentationTests.java[tag=subscribe]
 ----
 ===== Subscribe methods
 `PubsubTemplate` provides the following subscribe methods:
@@ -83,7 +83,7 @@ This is different from subscribing to a subscription, in the sense that subscrib
 Pull up to 10 messages:
 [source,java,indent=0]
 ----
-include::../../../src/test/java/org/springframework/cloud/gcp/autoconfigure/pubsub/it/PubSubTemplateDocumentationTests.java[tag=pull]
+include::../../test/java/org/springframework/cloud/gcp/autoconfigure/pubsub/it/PubSubTemplateDocumentationTests.java[tag=pull]
 ----
 
 ===== Pull methods
@@ -121,7 +121,7 @@ For serialization and deserialization of POJOs using Jackson JSON, configure a `
 
 [source,java,indent=0]
 ----
-include::../../../src/test/java/org/springframework/cloud/gcp/autoconfigure/pubsub/it/PubSubTemplateDocumentationTests.java[tag=json_bean]
+include::../../test/java/org/springframework/cloud/gcp/autoconfigure/pubsub/it/PubSubTemplateDocumentationTests.java[tag=json_bean]
 ----
 
 NOTE: Alternatively, you can set it directly by calling the `setMessageConverter()` method on the `PubSubTemplate`.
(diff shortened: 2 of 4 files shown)


docs/src/main/asciidoc/pubsub.adoc
```diff
@@ -82,36 +77,68 @@ Please refer to our link:../spring-cloud-gcp-samples/spring-cloud-gcp-integratio
...
-`Subscriber` objects.
 When listening to a subscription, messages will be pulled from Google Cloud Pub/Sub
-asynchronously, at a certain interval.
+asynchronously and passed to a user provided message handler.
+
+===== Example
+Subscribe to a subscription with a message handler:
+[source,java,indent=0]
+----
+include::../../../../spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-sample/src/main/java/com/example/WebController.java[tag=subscription]
```
artembilan: M-m-m. Have you checked this with the generated docs?
I feel like this is not going to be available when we have the final HTML doc.

docs/src/main/asciidoc/pubsub.adoc
```diff
@@ -82,36 +77,68 @@ Please refer to our link:../spring-cloud-gcp-samples/spring-cloud-gcp-integratio
...
+|===
 
-All `ack()`, `nack()`, and `modifyAckDeadline()` methods on messages as well as `PubSubSubscriberTemplate` are implemented asynchronously, returning a `ListenableFuture<Void>` to be able to process the asynchronous execution.
+===== Acknowledging messages
 
-`PubSubTemplate` uses a special subscriber generated by its `SubscriberFactory` to synchronously pull messages.
+There are two ways to acknowledge messages.
+
+. To acknowledge multiple messages at once, you can use the `PubSubTemplate.ack()` method.
+You can also use the `PubSubTemplate.nack()` for negatively acknowledging messages. Using these methods for acknowledging messages in batches is more efficient than acknowledging messages individually, but they *require* the collection of messages to be from the same project.
```
artembilan: One sentence per line

spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-sample/src/main/java/com/example/WebController.java
```diff
@@ -84,17 +84,20 @@ public RedirectView publish(@RequestParam("topicName") String topicName,
 
 	@GetMapping("/pull")
 	public RedirectView pull(@RequestParam("subscription1") String subscriptionName) {
-
+		//tag::pull10[]
```
artembilan: I don't think it is a good idea to pollute samples code with some metadata.
This code is going to be used by end-users and they are going to ask questions what is this about and how to deal with it.

docs/src/main/asciidoc/pubsub.adoc
dmitry-s: Yes, I checked. It works as expected.

spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-sample/src/main/java/com/example/WebController.java
meltsufin: @dmitry-s Any way we can make the comment actually useful and still work for embedding in docs?

dmitry-s: We can add any text before the tag name (tag::pull10[])

dzou: I am also not quite a fan of using the asciidoc metadata. It seems like an error-prone process; even in your current draft it looks like [some of the code snippets did not render correctly](https://github.com/spring-cloud/spring-cloud-gcp/blob/8ccfac1a07771f8e3510f0a54f4f01b077876f9c/docs/src/main/asciidoc/pubsub.adoc). Once we get to the point where we have many snippets, it is going to get confusing to have many references between code and docs.

I would prefer if we just copy/pasted code snippets in the doc. Yes there is a risk they may go out of sync and break, but this risk would be reduced significantly if we used snippets sparingly and kept them small. For example, it is not necessary to have snippets for all Pub/Sub actions like `createTopic`, `deleteTopic`, etc. since these are self-explanatory. It's probably only necessary to have a few code snippets for `Beans` to include in autoconfig and maybe complicated things like how to convert PubsubMessage to String. Such snippets would be small enough that they would be robust to breakage.

artembilan: Well, why then just not link to the existing samples which always are going to be up-to-date.

In other our projects we really have burden to support snippets in the docs as is without any includes and links.
Right, we sometimes forget to update them over there, but in the end it is less work, then answering for questions in issues or StackOverflow.

dmitry-s: We do link the samples, but it turns out users prefer seeing code snippets in the doc (based on feedback).

I think I'll just create a separate test class for ref doc code snippets, from which they will be included. This way samples can stay clean of metadata.

dzou: Gotcha, I like this idea. Is your intention to include Java code in the current `docs/`module to reference? I think this would be a good idea; basically keeping the code snippets and refdoc in the same place.

docs/src/main/asciidoc/pubsub.adoc
```diff
@@ -25,94 +25,130 @@ dependencies {
...
+----
 
-If Pub/Sub message payload conversion is desired, you can use the `subscribeAndConvert()` method, which will use the converter configured in the template.
+===== Pull methods
 
-==== Pulling messages from a subscription
+`PubsubTemplate` provides the following pull methods:
+[cols="20,80"]
+|===
+h|pull(String subscription, Integer maxMessages,
```
dzou: Should the method names be written with code font?

docs/src/main/asciidoc/pubsub.adoc
```diff
@@ -25,94 +25,130 @@ dependencies {
...
+       			Boolean returnImmediately) | Pulls a number of messages from a subscription, allowing for the retry settings to be configured.
+       			Any messages received by `pull()` are not automatically acknowledged. See <<Acknowledging messages>>.
 
-Google Cloud Pub/Sub supports synchronous pulling of messages from a subscription.
-This is different from subscribing to a subscription, in the sense that subscribing is an asynchronous task which polls the subscription on a set interval.
+       			If `returnImmediately` is `true`, the system will respond immediately even if it there are no messages available to return in the `Pull` response. Otherwise, the system may wait (for a bounded amount of time) until at least one message is available, rather than returning no messages.
+h|pullAndAck | Works the same as the `pull` method and, additionally, acknowledges all received messages.
```
dzou: I think for consistency, the method names should either all have their methods parameters or all not have method parameters. Right now because `pull` has the method parameters, it feels like the other methods have the same method signature as `pull`; however this is not the case.

docs/src/main/asciidoc/pubsub.adoc
```diff
@@ -25,94 +25,130 @@ dependencies {
...
+And thatt's how you convert messages to objects on pull:
+[source,java,indent=0]
+----
+include::../../../src/test/java/org/springframework/cloud/gcp/autoconfigure/pubsub/it/PubSubTemplateDocumentationTests.java[tag=json_pull]
+----
 
-`PubSubTemplate` uses a special subscriber generated by its `SubscriberFactory` to synchronously pull messages.
 
+Please refer to our link:../spring-cloud-gcp-samples/spring-cloud-gcp-integration-pubsub-json-sample[Pub/Sub JSON Payload Sample App] as a reference for using this functionality.
 === Pub/Sub management
```
dzou: Might need to add a newline above; the formatting appears off.

docs/src/main/asciidoc/pubsub.adoc
```diff
@@ -25,94 +25,130 @@ dependencies {
...
+
+And thatt's how you convert messages to objects on pull:
+[source,java,indent=0]
+----
+include::../../../src/test/java/org/springframework/cloud/gcp/autoconfigure/pubsub/it/PubSubTemplateDocumentationTests.java[tag=json_pull]
+----
 
-`PubSubTemplate` uses a special subscriber generated by its `SubscriberFactory` to synchronously pull messages.
 
+Please refer to our link:../spring-cloud-gcp-samples/spring-cloud-gcp-integration-pubsub-json-sample[Pub/Sub JSON Payload Sample App] as a reference for using this functionality.
```
dzou: This looks like a mistake from the previous version of the doc; I think this link is going to be broken if you test it with `mvn install -P=docs`. You will have to change to github link instead.

docs/src/main/asciidoc/pubsub.adoc
```diff
@@ -325,6 +347,13 @@ After this amount of time has elapsed (counting from the first element added), t
 Enables batching.|No | false
 |===
 
-=== Sample
+=== Details
```
dzou: I think this Details section might not be necessary because they are very abstracted from the user; someone wanting to truly understand would look at the sources instead. For me, the details about the factories went over my head - I think newer users of this library might be in a similar boat.

docs/pom.xml
```diff
@@ -12,6 +12,16 @@
 	<artifactId>spring-cloud-gcp-docs</artifactId>
     <packaging>pom</packaging>
     <name>Spring Cloud GCP Documentation</name>
+	<dependencies>
```
dzou: Can you fix the formatting in this file? looks like it was half indented with tabs, half with spaces.

docs/src/main/asciidoc/pubsub.adoc
```diff
@@ -25,94 +25,130 @@ dependencies {
...
-asynchronously, at a certain interval.
+Google Cloud Pub/Sub supports synchronous pulling of messages from a subscription.
+This is different from subscribing to a subscription, in the sense that subscribing is an asynchronous task.
 
-The Spring Boot starter for Google Cloud Pub/Sub auto-configures a `SubscriberFactory`.
+===== Example
+Pull up to 10 messages:
+[source,java,indent=0]
+----
+include::../../../src/test/java/org/springframework/cloud/gcp/autoconfigure/pubsub/it/PubSubTemplateDocumentationTests.java[tag=pull]
```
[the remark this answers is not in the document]
dmitry-s: But they ARE in `spring-cloud-gcp-docs` module :-)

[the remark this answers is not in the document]
dzou: Oops you are right.

docs/src/test/java/org/springframework/cloud/gcp/autoconfigure/pubsub/it/PubSubTemplateDocumentationTests.java
```diff
@@ -0,0 +1,292 @@
...
+public class PubSubTemplateDocumentationTests {
+
+	private ApplicationContextRunner contextRunner = new ApplicationContextRunner()
+			.withPropertyValues("spring.cloud.gcp.pubsub.subscriber.max-ack-extension-period=0")
+			.withConfiguration(AutoConfigurations.of(GcpContextAutoConfiguration.class,
+					GcpPubSubAutoConfiguration.class));
+
+	@BeforeClass
+	public static void enableTests() {
+//			assumeThat(System.getProperty("it.pubsub")).isEqualTo("true");
```
dzou: Uncomment this?

dmitry-s: good catch!

PR #1451, opened 2019-02-07
Fix jdk11 compilation/test problems

elefeint:
_javac-with-errorprone_ was causing this type of errors: 
```
cannot access java.lang.Object
  bad class file: /modules/java.base/java/lang/Object.class
    class file has wrong version 55.0, should be 53.0
```

The `OnGcpEnvironmentConditionTests` exception messages was too specific; it ended up checking internal java implementation errors.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -138,34 +138,6 @@
 		<pluginManagement>
 			<plugins>
 				<plugin>
-					<artifactId>maven-compiler-plugin</artifactId>
-					<configuration>
-						<compilerId>javac-with-errorprone</compilerId>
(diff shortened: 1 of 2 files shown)


artembilan:
> Jenkins build is back to normal : spring-cloud-jdk11-spring-cloud-gcp-master-ci 

👍

meltsufin:
Cool, but now we don't have errorprone? :disappointed:

elefeint:
The recommended way to use errorprone is as a plugin, not a complier now (google/error-prone#860). I'll 
 try to re-incorporate it.

PR #1454, opened 2019-02-07
Spanner Auditing

### LINKED ISSUE #1283: Spanner entitylistener and auditing

ChengyuanZhao:
fixes #1283 
adds user and date auditing annotation support. 

We are not supporting the CreatedBy and CreatedDate columns because Spring Data's support for those assumes the existence of hidden-from-the-user versioning columns which we do not support in this module.

### DIFF
--- a/docs/src/main/asciidoc/spanner.adoc
+++ b/docs/src/main/asciidoc/spanner.adoc
@@ -1271,6 +1271,44 @@ Spring Data Cloud Spanner publishes events extending the Spring Framework's `App
 |===
 
 
+=== Auditing
+
+Spring Data Cloud Spanner supports the `@LastModifiedDate` and `@LastModifiedBy` auditing annotations for properties:
+
+[source,java]
+----
+@Table
+public class SimpleEntity {
+    @PrimaryKey
+    String id;
+
+    @LastModifiedBy
+    String lastUser;
+
+    @LastModifiedDate
+    DateTime lastTouched;
+}
+----
+
+Upon insert, update, or save, these properties will be set automatically by the framework before mutations are generated and saved to Cloud Spanner.
+
+To take advantage of these features, add the `@EnableSpannerAuditing` annotation to your configuration class and provide a bean for an `AuditorAware<A>` implementation where the type `A` is the desired property type annotated by `@LastModifiedBy`:
+
+[source,java]
+----
+@Configuration
+@EnableSpannerAuditing
+public class Config {
+
+    @Bean
+    public AuditorAware<String> auditorProvider() {
+        return () -> Optional.of("YOUR_USERNAME_HERE");
+    }
+}
+----
+
+You can also set a custom provider for properties annotated `@LastModifiedDate` by providing a bean for `DateTimeProvider` and providing the bean name to `@EnableSpannerAuditing(dateTimeProviderRef = "customDateTimeProviderBean")`.
+
 === Sample
 
 
--- a/docs/src/main/asciidoc/spanner.adoc
+++ b/docs/src/main/asciidoc/spanner.adoc
@@ -1307,6 +1307,24 @@ public class Config {
 }
 ----
 
+The `AuditorAware` interface contains a single method that supplies the value for fields annotated by `@LastModofiedBy` and can be of any type.
+One alternative is to use Spring Security's `User` type:
+
+[source,java]
+----
+class SpringSecurityAuditorAware implements AuditorAware<User> {
+
+  public Optional<User> getCurrentAuditor() {
+
+    return Optional.ofNullable(SecurityContextHolder.getContext())
+			  .map(SecurityContext::getAuthentication)
+			  .filter(Authentication::isAuthenticated)
+			  .map(Authentication::getPrincipal)
+			  .map(User.class::cast);
+  }
+}
+----
+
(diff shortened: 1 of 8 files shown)


spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplateAuditingTests.java
```diff
@@ -0,0 +1,121 @@
...
+public class SpannerTemplateAuditingTests {
+
+	private static final List<Mutation> UPSERT_MUTATION = Arrays
+			.asList(Mutation.newInsertOrUpdateBuilder("custom_test_table").build());
+
+	@Autowired
+	SpannerTemplate spannerTemplate;
+
+	@Test
+	public void testModifiedProperties() {
```
dmitry-s: can we also check that lastTouched and lastUser are updated after saving when they are not null?

ChengyuanZhao: done!

docs/src/main/asciidoc/spanner.adoc
```diff
@@ -1271,6 +1271,44 @@ Spring Data Cloud Spanner publishes events extending the Spring Framework's `App
 |===
 
 
+=== Auditing
+
+Spring Data Cloud Spanner supports the `@LastModifiedDate` and `@LastModifiedBy` auditing annotations for properties:
```
dzou: Maybe add a link to reference Spring Auditing documentation? This may be helpful to give more context on what AuditorAware is.

ChengyuanZhao: There isn't a central Spring Data ref doc (only the actual implementations have them) ,but I can write more in this section.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
```diff
@@ -228,65 +228,68 @@ public long executeDmlStatement(Statement statement) {
 
 	@Override
 	public void insert(Object object) {
-		applySaveMutations(this.mutationFactory.insert(object), Collections.singletonList(object), null);
```
meltsufin: Since objects are passed by reference, the fact that `insert` modifies them shouldn't be a problem.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
```diff
@@ -228,65 +228,68 @@ public long executeDmlStatement(Statement statement) {
...
+		applySaveMutations(() -> this.mutationFactory.upsert(object, includeProperties),
+				Collections.singletonList(object),
 				includeProperties);
 	}
 
-	private void applySaveMutations(List<Mutation> mutations, Iterable entities,
+	private void applySaveMutations(Supplier<List<Mutation>> mutationsSupplier, Iterable entities,
 			Set<String> includeProperties) {
-		maybeEmitEvent(new BeforeSaveEvent(mutations, entities, includeProperties));
+		maybeEmitEvent(new BeforeSaveEvent(entities, includeProperties));
```
meltsufin: Why not include the mutations in `BeforeSaveEvent` as well? You would just need to move the next line up.

ChengyuanZhao: The BeforeSaveEvent alters the entities from which the mutations are generated, so we can't pass mutations into the creation of the event.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
ChengyuanZhao: It's not that the entity objects are modified, but that the mutations generated from them would be out of date (they're created new from the entity objects for the save operation).

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
meltsufin: I see, OK.

ChengyuanZhao:
@meltsufin I took a closer look how spring-data-jdbc determines if the entity was just-created.

It goes like this:
1. does if the entity has a `@version` annotated property then use that. if not, then use its ID value.
2. if the value is NULL then the entity was just-created
3. if the value is a number-like value and its value is 0 then it was just-created (so you can't use 0-value keys even if they're from existing rows in Spanner).
4. otherwise it's not just-created.

ChengyuanZhao:
@meltsufin does that sound OK to you? I think we can support the created-events but it's large enough to be a little project by itself.

meltsufin:
@ChengyuanZhao Yes, let's keep it separate.

Issue #1448, opened 2019-02-06
Consider making `logging` a separate artifact


stevenschlansker:
Currently, if you want to use GCP Stackdriver logging, you must include the whole `autoconfigure` artifact due to https://github.com/spring-cloud/spring-cloud-gcp/pull/346.
It would be nice to be able to use the feature independently, and manually configured.  Forcing users to use autoconfigure and get all the features together is unfortunate.

meltsufin:
Are you using API or JSON-based logging?
You do have the option of disabling the auto-configuration with `spring.cloud.gcp.logging.enabled=false`.
In any case, I think it's a valid suggestion.

stevenschlansker:
I'm trying to do JSON based logging.  I was surprised that the out of the box Google libraries did not have a functioning setup -- the resource labels are not all filled in appropriately on GKE -- so I've been evaluating this code instead.

dsyer:
Autoconfiguration is always optional - you don't have to use it. It's a pretty fine balance, but we prefer usually to keep features together if they are related, even if they are optional. The discussion about programming models not involving autoconfiguration should take place elsewhere really (like with the Spring Boot team), but I'm sure it would be interesting to hear your analysis beyond the throwaway comment.

stevenschlansker:
I don't want to get dragged into a religious war 😸 just want to make sure features work independently as well as together.  For what it's worth, we use Jackson over Gson -- so I ended up just borrowing the code and adapting it directly to my needs.  So if nobody else wants it split out, I consider my needs fulfilled for the moment (although we'd gladly ditch our copy if possible down the road).

dsyer:
There’s nothing religious about autoconfiguration, I promise you. It has always been optional by design, even if it is what most people use in practice. There are a couple of active efforts looking for more convenient “a la carte” programming models even (see this recent [blog](https://spring.io/blog/2019/01/21/manual-bean-definitions-in-spring-boot) and the [Spring Fu](https://github.com/spring-projects/spring-fu) project for instance).

PR #1455, opened 2019-02-07
Making Logging a separate module outside of autoconfiguration

meltsufin:
Fixes: #1448.


meltsufin:
@artembilan I decided against the "stackdriver" module because there is nothing useful to put there for the trace stuff, only logging.

meltsufin:
@artembilan I see your point. I did expect this to come up because there are some serious breaking changes, the most important of which is the path change to logback config files. I am, however, curious about the breaking changes policy. I have seen some breakage in minor versions of Spring like from Boot 2.0 to 2.1 and Framework 5.0 to 5.1, and even more so in Spring Cloud projects.
I'll try to make this a bit more backwards compatible. Perhaps we can import the logging configs from one file to the other.

artembilan:
I'm not sure what you mean about Spring Framework, but we really had some breaking changes in Spring Boot from 2.0 to 2.1: https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-2.1-Release-Notes
We had a big team discussion on the matter and those changes were really reasonable because previous behavior has just hidden an issue from us.

But that's fully different story and should be revised privately if that.

Anyway let's just don't make this breaking changes story like blaming each other! Let's try to keep our code from the breakage as much as possible. 
This way we will eliminate many questions and issues from community.

I think it won't hurt to keep content of those logging configs as identical. Doesn't look like we change it too often...

meltsufin:
@artembilan I followed your suggestion to also keep the original classes and files as deprecated. Please take another look.

PR #1457, opened 2019-02-08
Data store health indicator changes

nsrags:
Data store health indicator changes


nsrags:
Build failed to existing storage test case ,Not related to new code. 

[ERROR] Tests run: 2, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 19.785 s <<< FAILURE! - in org.springframework.cloud.gcp.autoconfigure.storage.GcpStorageAutoConfigurationTests
[ERROR] testValidObject(org.springframework.cloud.gcp.autoconfigure.storage.GcpStorageAutoConfigurationTests)  Time elapsed:

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/health/DatastoreHealthIndicator.java
```diff
@@ -0,0 +1,79 @@
...
+ *
+ * @author Raghavan N S
+ * @author Srinivasa Meenavalli
+ */
+@Component
+public class DatastoreHealthIndicator extends AbstractHealthIndicator {
+
+	private static final Status DATASTORE_HEALTH = new Status("DATASTORE_HEALTH");
+
+	private final SimpleDatastoreRepository<HealthCheckEntity, String> simpleDatastoreRepository;
```
meltsufin: I think it's too much machinery to use a repository to do a simple liveliness check.
Perhaps querying some metadata or statistics like `__namespace__` or `__Stat_Total__` directly using the `Datastore` class would be more efficient.
You can even just query for a non-existing kind:

```java
@Autowired
Datastore datastore;
...
datastore.run(Query.newKeyQueryBuilder().setKind("RandomKindName").build());
```
If no exception is thrown, Datastore must be up.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/health/DatastoreHealthIndicator.java
```diff
@@ -0,0 +1,79 @@
...
+		Assert.notNull(datastoreTemplate, "DatastoreTemplate must not be null");
+		this.simpleDatastoreRepository = new SimpleDatastoreRepository<>(datastoreTemplate,	HealthCheckEntity.class);
+	}
+
+	@Override
+	protected void doHealthCheck(Health.Builder builder) throws Exception {
+		String id = UUID.randomUUID().toString();
+		try {
+			HealthCheckEntity entity = new HealthCheckEntity(id, "Data store Health Check");
+			simpleDatastoreRepository.save(entity);
```
meltsufin: Writes are expensive. I would keep it low-level and simple without any entities and repositories. See comment above.

ChengyuanZhao:
Notwithstanding my prior opinion:
- these classes need to be in the Autoconfiguration module, not the Datastore Module
- there needs to be tests for these classes
- this code seems to break a previously-working test, which needs to be fixed in the PR. 
- a health-check insert and check can be done with a client library call, and doesn't seem to have anything to do with Spring Data

meltsufin:
+1 on keeping it self-contained in the auto-configuration module.
Of course a test would be nice too. If you're not sure how to write one, we can help you.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/health/DatastoreHealthIndicator.java
nsrags: I agree. I will use statistics query

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/health/DatastoreHealthIndicator.java
nsrags: Agree

nsrags:
> Notwithstanding my prior opinion:
> 
> * these classes need to be in the Autoconfiguration module, not the Datastore Module
> * there needs to be tests for these classes
> * this code seems to break a previously-working test, which needs to be fixed in the PR.
> * a health-check insert and check can be done with a client library call, and doesn't seem to have anything to do with Spring Data

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/storage/Content.java
```diff
@@ -0,0 +1,43 @@
...
+ * @author Srinivasa Meenavalli
+ */
+
+package org.springframework.cloud.gcp.autoconfigure.storage;
+
+import java.io.Serializable;
+
+public class Content implements Serializable {
+	/**
+	 *
```
ChengyuanZhao: There shouldn't be empty javadoc.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/datastore/health/DatastoreHealthIndicator.java
```diff
@@ -0,0 +1,67 @@
...
+	public DatastoreHealthIndicator(final Datastore datastore) {
+		super("Datastore health check failed");
+		this.datastore = datastore;
+		Assert.notNull(datastore, "Datastore must not be null");
+	}
+
+	@Override
+	protected void doHealthCheck(Health.Builder builder) throws Exception {
+		try {
+			datastore.run(Query.newKeyQueryBuilder().setKind("__Stat_Total__").build());
```
ChengyuanZhao: Is there a way to just get a single key rather than all of the keys in the table?

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/datastore/health/DatastoreHealthIndicator.java
```diff
@@ -0,0 +1,67 @@
...
+			builder.status(DATASTORE_HEALTH);
+			builder.up();
+		}
+		catch (Exception ex) {
```
ChengyuanZhao: Could this exception type be more specific to something from the client library? we only want to catch cases where contact with Datastore fails and not other types of errors.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/datastore/health/DatastoreHealthIndicatorConfiguration.java
```diff
@@ -0,0 +1,53 @@
+/*
+ * Copyright 2017-2018 the original author or authors.
```
ChengyuanZhao: 2019

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/datastore/health/DatastoreHealthIndicatorConfiguration.java
```diff
@@ -0,0 +1,53 @@
...
+import org.springframework.context.annotation.Configuration;
+
+/**
+ * Configuration for {@link DatastoreHealthIndicator}.
+ *
+ * @author Raghavan N S
+ * @author Srinivasa Meenavalli
+ */
+@Configuration
+@ConditionalOnClass({DatastoreRepositoryFactory.class, DatastoreTemplate.class, Datastore.class})
```
ChengyuanZhao: Why depend on the Spring Data Datastore classes? The indicator seems to be directly querying using the client lib.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/datastore/health/DatastoreHealthIndicatorConfiguration.java
```diff
@@ -0,0 +1,53 @@
...
+@ConditionalOnBean({DatastoreRepositoryFactory.class, DatastoreTemplate.class, Datastore.class})
+class DatastoreHealthIndicatorConfiguration extends CompositeHealthIndicatorConfiguration<DatastoreHealthIndicator, Datastore>	{
+	@Autowired
+	Datastore datastore;
+	DatastoreHealthIndicatorConfiguration(final Datastore datastore) {
+		this.datastore = datastore;
+	}
+	@Bean
+	@ConditionalOnMissingBean(name = "datastoreHealthIndicator")
```
ChengyuanZhao: Why conditional on the name "datastoreHealthIndicator" instead of the class you created `DatastoreHealthIndicator`?

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/datastore/GcpDatastoreAutoConfigurationTests.java
```diff
@@ -41,7 +41,8 @@
 	private ApplicationContextRunner contextRunner = new ApplicationContextRunner()
 			.withConfiguration(AutoConfigurations.of(GcpDatastoreAutoConfiguration.class,
 					GcpContextAutoConfiguration.class, DatastoreTransactionManagerAutoConfiguration.class,
-					DatastoreRepositoriesAutoConfiguration.class))
+					DatastoreRepositoriesAutoConfiguration.class,
+					TestRepository.class))
```
ChengyuanZhao: Why did you add `TestRepository` here? That circumvents the test below which is supposed to create the TestRepository instance using the Spring Data Datastore Repo factory instead of pulling a bean from the context....

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/datastore/health/DatastoreHealthIndicatorTest.java
```diff
@@ -0,0 +1,103 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.mock;
+
+/**
+ * Test class for DatastoreHealthIndicatorConfiguration.
+ *
+ * @author Raghavan N S
+ * @author Srinivasa Meenavalli
+ */
+public class DatastoreHealthIndicatorTest {
```
ChengyuanZhao: please put tests be put into the other existing test class `GcpDatastoreAutoConfigurationTests `. We must be sure that all of the things we want can be created from the same context.

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/storage/Content.java
```diff
@@ -0,0 +1,43 @@
...
+ *
+ * @author Raghavan N S
+ * @author Srinivasa Meenavalli
+ */
+
+package org.springframework.cloud.gcp.autoconfigure.storage;
+
+import java.io.Serializable;
+
+public class Content implements Serializable {
```
ChengyuanZhao: Could you explain the purpose of this new class and why you used it in the storage test?

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/datastore/health/DatastoreHealthIndicatorConfiguration.java
nsrags: Google Data Store Health indicator functionality will be enabled if spring-cloud-gcp-data-datastore has been added in Spring dependency .Refer Other health indicator implementations in Spring boot actuator 
https://docs.spring.io/spring-boot/docs/1.4.0.RC1/api/org/springframework/boot/actuate/autoconfigure/HealthIndicatorAutoConfiguration.MongoHealthIndicatorConfiguration.html

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/datastore/health/DatastoreHealthIndicatorConfiguration.java
nsrags: Refer Other health indicator implementations in Spring boot actuator
https://docs.spring.io/spring-boot/docs/1.4.0.RC1/api/org/springframework/boot/actuate/autoconfigure/HealthIndicatorAutoConfiguration.MongoHealthIndicatorConfiguration.html

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/datastore/GcpDatastoreAutoConfigurationTests.java
nsrags: I observed no bean definition found exception while running mvn clean install with out TestRepository.class

ChengyuanZhao: That means this PR is faulty and it's a problem that needs to be fixed because this repository class was intentionally NOT added and it worked before.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/datastore/health/DatastoreHealthIndicatorConfiguration.java
ChengyuanZhao: In my assessment I don't think simply using Spring Data Datastore should automatically trigger this background polling. Could you add a separate setting to enable it rather than making it automatic with Spring Data Datastore?

With other Spring Data modules that rely on user-hosted servers like Mongo that may make sense. But Cloud Datastore is not something the vast majority of users want to personally poll to check that it is up...

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/datastore/health/DatastoreHealthIndicatorConfiguration.java
ChengyuanZhao: In my assessment I don't think simply using Spring Data Datastore should automatically trigger this background polling. Could you add a separate setting to enable it rather than making it automatic with Spring Data Datastore?

With other Spring Data modules that rely on user-hosted servers like Mongo that may make sense. But Cloud Datastore is not something the vast majority of users want to personally poll to check that it is up...

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/storage/Content.java
nsrags: GcpStorageAutoConfigurationTests.java 
Existing code 
		Long actual = testRestTemplate.getForObject("http://localhost:" + this.port + "/resource", Long.class);
		assertThat(actual).isEqualTo(4096);
Error : error: Cannot deserialize instance of `java.lang.Long` out of START_OBJECT token; nested exception is com.fasterxml.jackson.databind.exc.MismatchedInputException: Cannot deserialize instance of `java.lang.Long` out of START_OBJECT toke 

I have created a Pojo( Content) to resolve above issue

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/datastore/health/DatastoreHealthIndicatorTest.java
nsrags: fixed

nsrags:
Can you review  updates?

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/datastore/health/DatastoreHealthIndicator.java
nsrags: DatastoreHealthIndicator extends AbstractHealthIndicator
AbstractHealthIndicator throws generic exception always . Should i use DatastoreException ? Please confirm
Class: AbstractHealthIndicator
@Override
	public final Health health() {
		Health.Builder builder = new Health.Builder();
		try {
			doHealthCheck(builder);
		}
		catch (Exception ex) {

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/datastore/health/DatastoreHealthIndicator.java
nsrags: Let me check . __Stat_Total__ always provide single record only

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/datastore/health/DatastoreHealthIndicator.java
meltsufin: I think in this particular case, catching any failure is OK because there are only two states: UP and DOWN. If it's not UP, it's DOWN, regardless what the reason is.

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/storage/Content.java
ChengyuanZhao: But that test used to work right? Why did it break in this PR?

spring-cloud-gcp-autoconfigure/pom.xml
```diff
@@ -225,7 +225,13 @@
             <artifactId>spring-security-oauth2-resource-server</artifactId>
             <optional>true</optional>
         </dependency>
-
+        
+		<!-- Actuator -->
```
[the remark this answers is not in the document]
ChengyuanZhao: Was this fixed? It still looks more indented than the preceding dependency

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/datastore/health/DatastoreHealthIndicatorConfiguration.java
```diff
@@ -0,0 +1,53 @@
...
+/**
+ * Configuration for {@link DatastoreHealthIndicator}.
+ *
+ * @author Raghavan N S
+ * @author Srinivasa Meenavalli
+ */
+@Configuration
+@ConditionalOnClass({DatastoreRepositoryFactory.class, DatastoreTemplate.class, Datastore.class})
+@ConditionalOnBean({DatastoreRepositoryFactory.class, DatastoreTemplate.class, Datastore.class})
+public class DatastoreHealthIndicatorConfiguration extends CompositeHealthIndicatorConfiguration<DatastoreHealthIndicator, Datastore>	{
```
ChengyuanZhao: Could this class be combined with `DatastoreHealthIndicatorAutoConfiguration` ? (That class is just an empty class)

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/datastore/GcpDatastoreAutoConfigurationTests.java
```diff
@@ -35,16 +42,27 @@
...
+					DatastoreRepositoriesAutoConfiguration.class,
+					DatastoreHealthIndicatorAutoConfiguration.class,
+					DatastoreHealthIndicatorConfiguration.class,
+					DatastoreHealthIndicator.class))
 			.withUserConfiguration(TestConfiguration.class)
 			.withPropertyValues("spring.cloud.gcp.datastore.project-id=test-project",
 					"spring.cloud.gcp.datastore.namespace-id=testNamespace");
+	@Before
+	public void setUp() {
+		datastoreHealthIndicator = new DatastoreHealthIndicator(mock(Datastore.class));
```
ChengyuanZhao: What is the purpose of this health indicator instance? Doesn't the autoconfig class use its own `createHealthIndicator` method to create the indicators?

meltsufin:
@nsrags Do you need any help to continue make progress on this PR? Please let us know. We should be able to add commits to your branch to fix things up if you like us to do that. Thanks!

nsrags:
> @nsrags Do you need any help to continue make progress on this PR? Please let us know. We should be able to add commits to your branch to fix things up if you like us to do that. Thanks!

Yes. Thanks for providing helping.  I couldn't spend time due to my current assignments load  . Can you add commits to my branch to fix pending review comments. Thanks once again .

ChengyuanZhao:
> > @nsrags Do you need any help to continue make progress on this PR? Please let us know. We should be able to add commits to your branch to fix things up if you like us to do that. Thanks!
> 
> Yes. Thanks for providing helping. I couldn't spend time due to my current assignments load . Can you add commits to my branch to fix pending review comments. Thanks once again .

I know how that can be! We currently are fully occupied with projects but will be glad to keep this PR open in the meantime.

meltsufin:
@nsrags @ChengyuanZhao I tried to make some progress on this PR. Please take a look. We're still missing unit tests for `DatastoreHealthIndicator`.

meltsufin:
@artembilan Do you know anyone who works on the Spring Boot Actuator who can take a quick look at this?

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/datastore/health/DatastoreHealthIndicator.java
```diff
@@ -0,0 +1,63 @@
...
+ * @author Mike Eltsufin
+ *
+ * @since 1.2
+ */
+@Component
+public class DatastoreHealthIndicator extends AbstractHealthIndicator {
+
+	private static final Status DATASTORE_HEALTH = new Status("DATASTORE_HEALTH");
+
+	private Datastore datastore;
```
artembilan: Must be `final`

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/datastore/health/DatastoreHealthIndicatorAutoConfiguration.java
```diff
@@ -0,0 +1,57 @@
...
+ *
+ * @since 1.2
+ */
+@Configuration
+@ConditionalOnClass(Datastore.class)
+@ConditionalOnBean(Datastore.class)
+@ConditionalOnEnabledHealthIndicator("datastore")
+@AutoConfigureBefore(HealthIndicatorAutoConfiguration.class)
+@AutoConfigureAfter(GcpDatastoreAutoConfiguration.class)
+public class DatastoreHealthIndicatorAutoConfiguration {
```
artembilan: This must be added to the `META-INF/spring.factories`  as one more `EnableAutoConfiguration` entry.
See sample in the Spring Boot per se: https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/resources/META-INF/spring.factories

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/datastore/health/DatastoreHealthIndicatorAutoConfiguration.java
```diff
@@ -0,0 +1,57 @@
...
+ * @author Raghavan N S
+ * @author Srinivasa Meenavalli
+ * @author Mike Eltsufin
+ *
+ * @since 1.2
+ */
+@Configuration
+@ConditionalOnClass(Datastore.class)
+@ConditionalOnBean(Datastore.class)
+@ConditionalOnEnabledHealthIndicator("datastore")
```
artembilan: I'm not sure that such a simple, generic *indicator* is really enough to distinguish GCP Datastore from any others.
What is the problem to have really more explicit name, e.g. `gcp.datastore`, at least?..

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/datastore/health/DatastoreHealthIndicatorAutoConfigurationTests.java
```diff
@@ -0,0 +1,82 @@
...
+
+	@Test
+	public void testDatastoreHealthIndicatorCreated() {
+		this.contextRunner.run((context) -> assertThat(context.getBean(DatastoreHealthIndicator.class)).isNotNull());
+	}
+
+	/**
+	 * Spring Boot config for tests.
+	 */
+	@AutoConfigurationPackage
```
artembilan: I think this is wrong annotation to indicate a simple `@Configuration`.
What made you to use exactly this annotation instead?

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/datastore/health/DatastoreHealthIndicatorAutoConfigurationTests.java
```diff
@@ -0,0 +1,82 @@
...
+import org.springframework.cloud.gcp.data.datastore.core.DatastoreTransactionManager;
+import org.springframework.context.annotation.Bean;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.mock;
+
+/**
+ * Tests for Datastore health check auto-config.
+ *
+ * @author Raghavan N S
```
artembilan: Tell, me please, that this is really your real name.
Otherwise, please, fix it since we only accept official names in our docs.

Thanks for understanding

meltsufin:
@artembilan Your review is certainly sufficient, and much appreciated! I just noticed you weren't looking at it for a couple of days and decided that you might think yourself not qualified to review. :)

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/datastore/health/DatastoreHealthIndicatorAutoConfiguration.java
meltsufin: This would only be a thing if someone is using Spring Cloud GCP, and there is only one "Datastore" in GCP. We can also change it to the more complete official name "cloud-datastore". WDYT?

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/datastore/health/DatastoreHealthIndicatorAutoConfigurationTests.java
meltsufin: It works with `@Configuration`, but interestingly we also use it in other places like `GcpSpannerAutoConfigurationTests`, and plain old `@Configuration` doesn't work in that case. :confused:

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/datastore/health/DatastoreHealthIndicatorAutoConfigurationTests.java
meltsufin: That's actually his name as far as I can tell. Google it. :)

meltsufin:
@artembilan Ready for another pass. Thanks!

PR #1460, opened 2019-02-08
Spanner simple lazy interleaved

### LINKED ISSUE #1235: Spanner lazy retrieval of @Interleaved children
Spanner equivalent of #1101 for Datastore.

ChengyuanZhao:
fixes #1235 

adds support for lazy retrieved interleaved collections using proxies

### DIFF
--- a/docs/src/main/asciidoc/spanner.adoc
+++ b/docs/src/main/asciidoc/spanner.adoc
@@ -415,7 +415,7 @@ List<Album> albums;
 ----
 
 Lazily fetched interleaved properties are retrieved upon the first interaction with the property.
-If a lazy property is never retrieved, then it is also skipped when saving the parent entity.
+If a property marked for lazy fetching is never retrieved, then it is also skipped when saving the parent entity.
 
 If used inside a transaction, subsequent operations on lazy properties use the same transaction context as that of the original parent entity.
 
--- a/docs/src/main/asciidoc/spanner.adoc
+++ b/docs/src/main/asciidoc/spanner.adoc
@@ -417,7 +417,7 @@ List<Album> albums;
 Lazy interleaved properties are retrieved upon the first interaction with the property.
 If a lazy property is never retrieved, then it is also skipped when saving the parent entity.
 
-If used inside a transaction, subsequent operations on lazy properties use the same transaction context as the original parent entity's retrieval.
+If used inside a transaction, subsequent operations on lazy properties use the same transaction context as that of the original parent entity.
 
 
 ==== Supported Types
--- a/docs/src/main/asciidoc/spanner.adoc
+++ b/docs/src/main/asciidoc/spanner.adoc
@@ -414,10 +414,10 @@ On read, all of the interleaved child rows are also all read.
 List<Album> albums;
 ----
 
-Lazily fetched interleaved properties are retrieved upon the first interaction with the property.
+Lazily-fetched interleaved properties are retrieved upon the first interaction with the property.
 If a property marked for lazy fetching is never retrieved, then it is also skipped when saving the parent entity.
 
-If used inside a transaction, subsequent operations on lazy properties use the same transaction context as that of the original parent entity.
+If used inside a transaction, subsequent operations on lazily-fetched properties use the same transaction context as that of the original parent entity.
 
 
 ==== Supported Types
--- a/docs/src/main/asciidoc/spanner.adoc
+++ b/docs/src/main/asciidoc/spanner.adoc
@@ -404,6 +404,22 @@ A table may have multiple child tables.
 On updating or inserting an object to Cloud Spanner, all of its referenced children objects are also updated or inserted in the same request, respectively.
 On read, all of the interleaved child rows are also all read.
 
+===== Lazy Fetch
+
+`@Interleaved` properties are retrieved eagerly by default, but can be fetched lazily for performance in both read and write:
+
+[source,java]
+----
+@Interleaved(lazy = true)
+List<Album> albums;
+----
+
+Lazy interleaved properties are retrieved upon the first interaction with the property.
+If a lazy property is never retrieved, then it is also skipped when saving the parent entity.
+
+If used inside a transaction, subsequent operations on lazy properties use the same transaction context as the original parent entity's retrieval.
+
+
 ==== Supported Types
 
 Spring Data Cloud Spanner natively supports the following types for regular fields but also utilizes custom converters (detailed in following sections) and dozens of pre-defined Spring Data custom converters to handle other common Java types.
--- a/docs/src/main/asciidoc/spanner.adoc
+++ b/docs/src/main/asciidoc/spanner.adoc
@@ -414,7 +414,7 @@ On read, all of the interleaved child rows are also all read.
 List<Album> albums;
 ----
 
-Lazy interleaved properties are retrieved upon the first interaction with the property.
+Lazily fetched interleaved properties are retrieved upon the first interaction with the property.
 If a lazy property is never retrieved, then it is also skipped when saving the parent entity.
 
 If used inside a transaction, subsequent operations on lazy properties use the same transaction context as that of the original parent entity.
--- a/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
+++ b/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
@@ -541,7 +541,7 @@ public class SpannerTemplate implements SpannerOperations, ApplicationEventPubli
 					SpannerPersistentEntity childPersistentEntity = this.mappingContext
 							.getPersistentEntity(childType);
 
-					Supplier<List> childPropertyFunc = () -> queryAndResolveChildren(childType,
+					Supplier<List> getChildrenEntitiesFunc = () -> queryAndResolveChildren(childType,
 							SpannerStatementQueryExecutor.getChildrenRowsQuery(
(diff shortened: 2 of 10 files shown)


spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplateTests.java
```diff
@@ -641,6 +641,33 @@ public void resolveChildEntityTest() {
 		});
 	}
 
+	@Test
+	public void lazyFetchChildrenTest() {
+		ChildEntity c = new ChildEntity();
```
dzou: Any reason these ChildrenEntity's fields are not set through constructor?

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
```diff
@@ -539,12 +540,17 @@ private void resolveChildEntity(Object entity, Set<String> includeProperties) {
 					Class childType = spannerPersistentProperty.getColumnInnerType();
 					SpannerPersistentEntity childPersistentEntity = this.mappingContext
 							.getPersistentEntity(childType);
+
+					Supplier<List> childPropertyFunc = () -> queryAndResolveChildren(childType,
```
dzou: Maybe rename var to something like `getChildrenRowsFunc`? (Currently not 100% sure what it is doing).

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/ConversionUtils.java
```diff
@@ -54,4 +60,46 @@ public static boolean isIterableNonByteArrayType(Class propType) {
 		source.forEach((item) -> result.add(converter.convert(item, targetType)));
 		return result;
 	}
+
+	public static <T> T wrapSimpleLazyProxy(Supplier<T> supplierFunc, Class<T> type) {
+		return (T) Proxy.newProxyInstance(type.getClassLoader(), new Class[] { type },
+				new SimpleLazyDynamicInvocationHandler<>(supplierFunc));
+	}
+
+	public static boolean isUnevaluatedLazyProxy(Object object) {
```
dmitry-s: The result is negated the only time this method is used. Can we make it `isEvaluatedLazyProxy` and avoid negation?

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/ConversionUtils.java
```diff
@@ -54,4 +60,46 @@ public static boolean isIterableNonByteArrayType(Class propType) {
...
+		if (!Proxy.isProxyClass(object.getClass()) ||
```
dmitry-s: maybe simplify this method to 
```
public static boolean isUnevaluatedLazyProxy(Object object) {
if (Proxy.isProxyClass(object.getClass()) && (Proxy.getInvocationHandler(object) instanceof SimpleLazyDynamicInvocationHandler)) {
SimpleLazyDynamicInvocationHandler handler = (SimpleLazyDynamicInvocationHandler) Proxy
				.getInvocationHandler(object);
		return !handler.isEvaluated();
}
return false;
}
```
to make it more readable?

spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplateTests.java
ChengyuanZhao: yes :D such a constructor does not exist

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
ChengyuanZhao: done!

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/ConversionUtils.java
ChengyuanZhao: I had that originally but that is decidedly less useful. 

assuming we use "isEvaluatedLazyProxy" this would evaluate to FALSE if the property in question is not a proxy at all. 

instead, "isUnevaluatedLazyProxy" evaluating to true represents all the cases we DON'T want to save-entity, because it means it is NOT a real object and also NOT an evaluated-proxy

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/ConversionUtils.java
ChengyuanZhao: done.

docs/src/main/asciidoc/spanner.adoc
```diff
@@ -404,6 +404,22 @@ A table may have multiple child tables.
 On updating or inserting an object to Cloud Spanner, all of its referenced children objects are also updated or inserted in the same request, respectively.
 On read, all of the interleaved child rows are also all read.
 
+===== Lazy Fetch
+
+`@Interleaved` properties are retrieved eagerly by default, but can be fetched lazily for performance in both read and write:
+
+[source,java]
+----
+@Interleaved(lazy = true)
```
meltsufin: Are there some existing enums we can reuse, like `fetch = FetchType.LAZY`?

docs/src/main/asciidoc/spanner.adoc
```diff
@@ -404,6 +404,22 @@ A table may have multiple child tables.
...
+List<Album> albums;
+----
+
+Lazy interleaved properties are retrieved upon the first interaction with the property.
```
meltsufin:
```suggestion
Lazily fetched interleaved properties are retrieved upon the first interaction with the property.
```

docs/src/main/asciidoc/spanner.adoc
```diff
@@ -404,6 +404,22 @@ A table may have multiple child tables.
...
+If a lazy property is never retrieved, then it is also skipped when saving the parent entity.
```
meltsufin:
```suggestion
If a property marked for lazy fetching is never retrieved, then it is also skipped when saving the parent entity.
```

docs/src/main/asciidoc/spanner.adoc
```diff
@@ -404,6 +404,22 @@ A table may have multiple child tables.
...
+
+If used inside a transaction, subsequent operations on lazy properties use the same transaction context as that of the original parent entity.
```
meltsufin: I'm not a fan of the "lazy property" wording. I know it's more succinct, but it just doesn't sound as good as lazy fetch.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/mapping/Interleaved.java
```diff
@@ -34,4 +34,12 @@
 @Retention(RetentionPolicy.RUNTIME)
 public @interface Interleaved {
 
+	/**
+	 * Controls whether the referenced entity should be loaded lazily. This defaults to
+	 * {@literal false}.
+	 *
+	 * @return whether the interleaved property is retrieved lazily.
+	 */
+	boolean lazy() default false;
```
meltsufin: I would call the property `fetch` and the value be an enum with the option either Lazy or Eager.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/ConversionUtils.java
dmitry-s: I think double negation here looks confusing: `!ConversionUtils.isUnevaluatedLazyProxy(kids)` 
Maybe rename the method?

ChengyuanZhao: renamed.

docs/src/main/asciidoc/spanner.adoc
ChengyuanZhao: Yes, however they are not in any dependency we have. 

I think it is for this reason that spring data Mongodb uses a boolean like we do here.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/mapping/Interleaved.java
ChengyuanZhao: Spring Data MongoDB actually uses a boolean named "lazy()" for I think the same reason (that the dependency for javax.persistence.fetchtype isn't a dependency)

docs/src/main/asciidoc/spanner.adoc
ChengyuanZhao: @meltsufin fixed all such instances.

ChengyuanZhao:
@vineela-gampa FYI lazy-read has been added for interleaved children

PR #1461, opened 2019-02-08
Reactive support for Pub/Sub subscription

### LINKED ISSUE #186: Add reactive binder for PubSub
A placeholder issue to investigate a reactor+pubsub integration - if I can I will take a stab at it. But before then let’s see if it makes sense :)

elefeint:
The polling Pub/Sub flux is demand-sensitive, implementing the pull/push strategy.
Unbounded/bounded demand is treated differently:
* For unlimited demand, the Pub/Sub subscription will be polled regularly, at intervals determined by `pollingPeriodMs` parameter passed in when creating the `Flux`.
* For bounded demand, as many messages as possible (up to the requested number) are delivered immediately, with the remaining messages delivered as they become available.

Fixes #186.


spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/reactive/FluxSubscriber.java
```diff
@@ -0,0 +1,95 @@
...
+		@Override
+		public void run() {
+			long remainingDemand = demand;
+
+			while (remainingDemand > 0) {
+				int numToRequest = remainingDemand > Integer.MAX_VALUE ? Integer.MAX_VALUE : (int) remainingDemand;
+
+				LOGGER.info("  Current remaining demand: " + remainingDemand);
+				List<ConvertedAcknowledgeablePubsubMessage<T>> messages
+						= this.subscriberOperations.pullAndConvert("reactiveSubscription",
```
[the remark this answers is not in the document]
elefeint: Definitely not intentional; I'll move it into a parameter.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/reactive/FluxSubscriber.java
```diff
@@ -0,0 +1,95 @@
...
+	private static final Log LOGGER = LogFactory.getLog(FluxSubscriber.class);
+
+	public static <T> Flux<ConvertedAcknowledgeablePubsubMessage<T>> createPolledFlux(
+			PubSubSubscriberOperations subscriberOperations, Class<T> targetType) {
+		ExecutorService executorService = Executors.newSingleThreadExecutor();
+
+		return Flux.create(sink -> {
+			sink.onRequest((numRequested) -> {
+				// request must be non-blocking.
+				executorService.submit(new PubSubPullTask<T>(numRequested, sink, subscriberOperations, targetType));
```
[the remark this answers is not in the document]
elefeint: The problem is that I am making a blocking pull call here, so if I request on the same thread, the code will violate the reactive streams rule that[ request should be non-obstructing](https://github.com/reactive-streams/reactive-streams-jvm#user-content-3.4).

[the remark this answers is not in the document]
meltsufin: There might be a way to do the pull asynchronously without your own `ExecutorService`.
See here: https://github.com/googleapis/google-cloud-java/blob/master/google-cloud-examples/src/main/java/com/google/cloud/examples/pubsub/snippets/SubscriberSnippets.java#L213
```
subscriber.pullCallable().futureCall()
```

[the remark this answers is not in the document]
artembilan: It's hard for me to object, but I think your thread spawning per subscriber is still bad idea.
Consider to use somehow `Schedulers.elastic()`.

I will try to pull some one from our Reactor team to see what that think on the matter...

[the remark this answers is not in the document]
artembilan: Oh! `futureCall()`. that's an answer! Right., we need to add it into the `PubSubSubscriberOperations` API and use it from this `Flux` trick.
There is simply need to be have a `flatMap(e -> Mono.fromFuture())` and so on with `flatMapIterable()`.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/reactive/FluxSubscriber.java
```diff
@@ -0,0 +1,95 @@
...
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+
+/**
+ * Creates a Flux populated by using Pub/Sub Synchronous Pull.
+ *
+ * @author Elena Felder
+ * @since 1.2
+ */
+public class FluxSubscriber {
```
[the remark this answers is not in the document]
elefeint: Something like `PubSubSubscriberFactory`? 
Then we can move the .subscribe() methods into this, as we discussed before for the refactoring.

On the other hand, we already have a `SubscriberFactory`, so I could just add the Flux creation logic into the existing `DefaultSubscriberFactory`.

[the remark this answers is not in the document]
artembilan: You can if `reactor-core` is hard dependency, not optional.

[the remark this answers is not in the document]
elefeint: Hmm, good point -- it should probably be optional. so it should not sit in a core class. Reactive is not the most common usecase.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/reactive/FluxSubscriber.java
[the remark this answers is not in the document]
elefeint: futureCall() sounded good in theory, but now we have potentially multiple outstanding calls to PubSub with additional callbacks being spawned.
Flux making an additional request when the data is 75% or so processed, combined with PubSub synchronous pull timing out after 10 minutes makes for messy logic.

I'd almost prefer to take a "poll interval" parameter, accumulate total outstanding demand in the custom flux, and then make "return_immediately=true" requests at an interval.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/reactive/PubSubReactiveFactory.java
```diff
@@ -0,0 +1,84 @@
...
+		subscriberOperations.pullAsync(subscriptionName, demand, false).addCallback(
+				acknowledgeablePubsubMessages -> {
+
+					sink.next(acknowledgeablePubsubMessages);
+
+					int remainingDemand = demand - acknowledgeablePubsubMessages.size();
+
+					if (remainingDemand > 0) {
+						// Ensure requested messages get fulfilled eventually.
+						produceMessages(subscriptionName, remainingDemand, sink, subscriberOperations);
```
elefeint: Between the need to keep `onRequest()` non-blocking and attempting to fulfill outstanding demand, the code relying on gRPC's `futureCall()` became messy and with potentially too many simultaneous calls to Pub/Sub. And adding callbacks to `ApiFuture` requires a separate executor, anyway (I reused the former ackExecutor in the latest implementation).

I would prefer to go back to maintaining a thread pool (but a shared one, not one per flux) in `PubSubReactiveFactory`, and using non-blocking polling at a customizable interval to satisfy remaining demand.

@artembilan @meltsufin What do you think?

meltsufin: Yeah, the call stack can get huge with this recursive call. So, it might make sense to go back to the synchronous pull with a pooled executor. It doesn't look like we're able to get away from providing our own executor anyway. WDYT @artembilan?

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberTemplate.java
```diff
@@ -109,9 +109,9 @@ public void setMessageConverter(PubSubMessageConverter pubSubMessageConverter) {
 		this.pubSubMessageConverter = pubSubMessageConverter;
 	}
 
-	public void setAckExecutor(Executor ackExecutor) {
-		Assert.notNull(ackExecutor, "ackExecutor can't be null.");
-		this.ackExecutor = ackExecutor;
+	public void setCallbackExecutor(Executor callbackExecutor) {
```
artembilan: Do you realize that this is a breaking change? We just can't rename API just because.
Please, consider to deprecate exiting method in favor of a new one.
And let's keep it here at least for the next version!

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/reactive/PubSubReactiveFactory.java
artembilan: I would say that's not our problem to deal with `remained` here and die in the stack overflow.

Please, see `Subscription.request()` JavaDocs:
```
* A {@link Publisher} can send less than is requested if the stream ends but
* then must emit either {@link Subscriber#onError(Throwable)} or {@link Subscriber#onComplete()}.
```

so, our goal here is to produce exactly requested amount or less and then `onComplete()`: `FluxSink.complete()`.

Not sure why have you gone a bit different way...

Also the rule number 1 of Publisher specification: https://github.com/reactive-streams/reactive-streams-jvm/blob/v1.0.2/README.md#specification

> The total number of onNext´s signalled by a Publisher to a Subscriber MUST be less than or equal to the total number of elements requested by that Subscriber´s Subscription at all times.

I think you are missing the fact that our `createPolledPublisher()` is just *finite* and that's already out of our scope to keep it hot and constantly updating.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberTemplate.java
elefeint: Point well taken. If this is the direction in which this code will go, I'll add the proper deprecation (I hope it does not go there, though -- per discussion below, I would prefer to revert the changes to `PubSubSubscriberTemplate` and manage multi threading in `PubSubReactiveFactory` instead).

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/reactive/PubSubReactiveFactory.java
elefeint: Is a cold publisher useful in Pub/Sub context? If the requirement is "get up to N messages and complete", then the client code can just use generate:
````
Flux.<List<AcknowledgeablePubsubMessage>>generate(
  () -> subscriberOperations.pull(subscriptionName, demand, true)
).flatMapIterable(Function.identity());
````

It seems to me based on discussion in reactor/reactor-core#314 that `onRequest()` was added to support our exact usecase of an unlimited flux that still responds to backpressure.


Off-topic, if I start a metal band, "We Die in the Stack Overflow" will be its name.

Comment (artembilan) [reply spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/reactive/PubSubReactiveFactory.java]: Hm... Then I have no answer or my Reactive knowledge is limited...

I wonder if we can bring someone who is an expert in the area to give us some feedback...

Comment (elefeint) [reply spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/reactive/PubSubReactiveFactory.java]: Well, everything I learned about Reactive I learned in the last month, so I am hardly an expert. Could we maybe set up a meeting with Reactive folks?

Comment (meltsufin) [reply spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/reactive/PubSubReactiveFactory.java]: Thinking about this a bit more, I'm not sure what's wrong with just returning whatever we get from Pub/Sub with the return-immediately=true, and not trying to fulfill the full demand. If Pub/Sub returns less than the demand, we just call `onComplete` assuming there are no more messages for the time being. The client can always resume its requests.
What's our outstanding question for the experts? I think the specification is fairly precise.

Comment (meltsufin) [reply spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/reactive/PubSubReactiveFactory.java]: Ok, just talked to @elefeint. I see why it's bad to terminate the stream when the demand is unmet by Pub/Sub. It doesn't work for infinite streams. So, it looks like it's better to keep the subscription open knowing that Pub/Sub can eventually provide more messages. We just need to poll it on a regular basis. Ideally, Pub/Sub would support true reactive streaming itself, but for the time being we need to simulate it by polling.

Comment (smaldini) [reply spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/reactive/PubSubReactiveFactory.java]: Hey that'd be great to meet if you folks want to introduce us to this in a zoom or somewhere and we can pair-review together !

File: spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/reactive/PubSubReactiveFactory.java [hunk 6]
```diff
@@ -0,0 +1,93 @@
...
+		this.subscriberOperations = subscriberOperations;
+
+	}
+
+	public Publisher<AcknowledgeablePubsubMessage> createPolledPublisher(
+			String subscriptionName, int delayInMilliseconds) {
+
+		PubSubReactiveSubscription reactiveSubscription = new PubSubReactiveSubscription(subscriptionName);
+
+		return Flux.interval(Duration.ofMillis(delayInMilliseconds))
```
elefeint: onBackpressureDrop()

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/reactive/PubSubReactiveFactory.java
```diff
@@ -0,0 +1,93 @@
...
+				.doOnRequest(reactiveSubscription::addDemand)
+				.map(t -> reactiveSubscription.pullMessages())
```
elefeint: collapse map and flatmap into one operation

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/reactive/PubSubReactiveFactory.java
```diff
@@ -0,0 +1,93 @@
...
+
+	public PubSubReactiveFactory(PubSubSubscriberOperations subscriberOperations) {
+		this.subscriberOperations = subscriberOperations;
+
+	}
+
+	public Publisher<AcknowledgeablePubsubMessage> createPolledPublisher(
+			String subscriptionName, int delayInMilliseconds) {
+
+		PubSubReactiveSubscription reactiveSubscription = new PubSubReactiveSubscription(subscriptionName);
```
elefeint: avoid creating subscription object outside of flux creation.
bottom line: use Flux.create for more control over the process

elefeint:
@smaldini @artembilan @meltsufin I've changed the Pub/Sub flux creation implementation as we discussed last Monday (with `Flux.create` and special treatment of unlimited demand). Could you take another look?

I still owe tests.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
```diff
@@ -381,4 +383,12 @@ public SubscriptionAdminClient subscriptionAdminClient(
 	public TransportChannelProvider transportChannelProvider() {
 		return InstantiatingGrpcChannelProvider.newBuilder().build();
 	}
+
+	@Bean
+	@ConditionalOnMissingBean
+	@ConditionalOnClass(Flux.class)
```
artembilan: This must be moved to a separate inner `@Configuration` class, marked with this annotation.
Just on the `@Bean` method won't help us to avoid classpath issues.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/reactive/PubSubReactiveFactory.java
```diff
@@ -0,0 +1,153 @@
...
+
+	private PubSubSubscriberOperations subscriberOperations;
+
+	/**
+	 * Instantiates `PubSubReactiveFactory` capable of generating subscription-based streams.
+	 *
+	 * @param subscriberOperations template for interacting with GCP Pub/Sub subscriber operations.
+	 */
+	public PubSubReactiveFactory(PubSubSubscriberOperations subscriberOperations) {
+		this.subscriberOperations = subscriberOperations;
```
artembilan: `Assert.notNull()` ?

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/reactive/PubSubReactiveFactory.java
```diff
@@ -0,0 +1,153 @@
...
+	}
+
+	/**
+	 * Creates an infinite stream {@link Publisher} of {@link AcknowledgeablePubsubMessage} objects.
```
artembilan: Method JavaDocs must be in the *imperative style*: https://github.com/spring-projects/spring-framework/wiki/Code-Style#javadoc-formatting

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/reactive/PubSubReactiveFactory.java
```diff
@@ -0,0 +1,153 @@
...
+	 * <p>For unlimited demand, the underlying subscription will be polled at a regular interval.
+	 * <p>For specific demand, as many messages as are available will be returned immediately, with remaining demand being
+	 * fulfilled in the future. Pub/Sub timeout will cause a retry with the same demand.
+	 * @param subscriptionName subscription from which to retrieve messages.
+	 * @param pollingPeriod how frequently to poll the source subscription in case of unlimited demand.
+	 * @return infinite stream of {@link AcknowledgeablePubsubMessage} objects.
+	 */
+	public Publisher<AcknowledgeablePubsubMessage> createPolledPublisher(String subscriptionName, long pollingPeriod) {
```
artembilan: I think our (Spring) programming module to expect Reactive Streams interfaces as method arguments, but return just an explicit Reacor types: `Flux` and `Mono`.
No reason to narrow it if you deal internally with only Reactor.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/reactive/PubSubReactiveFactory.java
```diff
@@ -0,0 +1,153 @@
...
+			int numMessagesToPull = demand > Integer.MAX_VALUE ? Integer.MAX_VALUE : (int) demand;
+
+			List<AcknowledgeablePubsubMessage> messages =  PubSubReactiveFactory.this.subscriberOperations.pull(
+					this.subscriptionName, numMessagesToPull, !block);
+
+			if (messages.size() > 0) {
+				messages.forEach(sink::next);
+			}
+
+			return messages;
```
artembilan: I think we may end up with the contract just really return an amount we have just pulled, not the whole list of messages.

spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-reactive-sample/pom.xml
```diff
@@ -0,0 +1,44 @@
...
+
+		<dependency>
+			<groupId>org.springframework.boot</groupId>
+			<artifactId>spring-boot-starter-web</artifactId>
+		</dependency>
+
+		<dependency>
+			<groupId>io.projectreactor</groupId>
+			<artifactId>reactor-core</artifactId>
+			<optional>true</optional>
```
artembilan: I believe it can be non-optional here already.
This is end-user code, so that is fully OK to have it as a hard dependency.

spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-reactive-sample/src/main/java/com/example/ReactiveController.java
```diff
@@ -0,0 +1,59 @@
...
+	@GetMapping(value = "/getmessages", produces = MediaType.TEXT_EVENT_STREAM_VALUE)
+	public Flux<? super String> getMessages() {
+
+		Publisher<AcknowledgeablePubsubMessage> flux
+				= this.reactiveFactory.createPolledPublisher("exampleSubscription", 1000);
+
+		return Flux.from(flux)
+				.doOnNext(message -> {
+					System.out.println("Received a message: " + message.getPubsubMessage().getMessageId());
+					message.ack();
```
artembilan: I think we need to document this manual ack'ing or maybe consider an auto-ack mode as well...

elefeint: Customizable ack mode is a good idea. Would it make sense to move (properly, with deprecation) `AckMode` from `o.s.c.gcp.pubsub.integration` up a level, or to simply use a `boolean autoAck` parameter when creating the `Flux`?

artembilan: I think `boolean autoAck` is fully enough, but let's really consider this improvement when we are already done with basics in this PR!

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/reactive/PubSubReactiveFactory.java
```diff
@@ -0,0 +1,168 @@
...
+	 * Pub/Sub timeout will cause a retry with the same demand.
+	 * @param subscriptionName subscription from which to retrieve messages.
+	 * @param pollingPeriod how frequently to poll the source subscription in case of unlimited demand.
+	 * @return infinite stream of {@link AcknowledgeablePubsubMessage} objects.
+	 */
+	public Flux<AcknowledgeablePubsubMessage> createPolledFlux(String subscriptionName, long pollingPeriod) {
+
+		return Flux.create(sink -> {
+			sink.onRequest((numRequested) -> {
+				if (numRequested == Long.MAX_VALUE) {
```
meltsufin: I'm still trying to wrap my head around the idea that if someone requests `Long.MAX_VALUE-1`, the behavior is radically different than with `Long.MAX_VALUE`, but I guess `Long.MAX_VALUE` is just has a very special meaning for reactive.

elefeint: Yes, this was the only awkward thing about Reactor; I'd prefer a boolean indicator of unbounded demand. The rest of the framework is perfect elegance, though.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/reactive/PubSubReactiveFactory.java
```diff
@@ -0,0 +1,168 @@
...
+		}
+
+		/**
+		 * Retrieve up to a specified number of messages, sending them to the subscription.
+		 * @param demand maximum number of messages to retrieve
+		 * @param block whether to wait for the first message to become available
+		 * @return number of messages retrieved
+		 */
+		protected int pullToSink(long demand, boolean block) {
+			int numMessagesToPull = demand > Integer.MAX_VALUE ? Integer.MAX_VALUE : (int) demand;
```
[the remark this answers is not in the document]
elefeint: The demand comes from `FluxSink.onRequest()` parameter, which is a `Long`. Ultimately, there is tension between Reactor (which operates on `Long` demand) and PubSub (which operates on `int` demand), so it has to get resolved somewhere in my code.

Or did you mean to simply move the conversion from `PubSubBlockingLimitedDemandPullTask.run()` up to `PubSubReactiveFactory.createPolledFlux.createPolledFlux()`?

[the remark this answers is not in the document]
meltsufin: Right, hand the responsibility to the caller.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/reactive/PubSubReactiveFactory.java
```diff
@@ -0,0 +1,168 @@
...
+
+			List<AcknowledgeablePubsubMessage> messages =  PubSubReactiveFactory.this.subscriberOperations.pull(
+					this.subscriptionName, numMessagesToPull, !block);
+
+			if (!this.sink.isCancelled()) {
+				messages.forEach(sink::next);
```
artembilan: I wonder what is wrong with sending the whole `List<AcknowledgeablePubsubMessage>` batch to the sink as a single event and then use a `flatMapIterable(Function.identity())` before returning the `Flux` to end-user...

elefeint: `flatMapIterable` changes demand, which could look unintuitive to the consumer (if subscription requests 3 messages, `flatMapIterable()` will cause 256 to be requested instead).

Also (this is not a good reason, just relevant) when I was testing, I ran into an issue with additional demand not being propagated from HTTP request to the Pub/Sub flux even though the initial request was fulfilled. I'd like to reproduce a minimal case to understand whether this behaves as intended or is a bug.

artembilan: OK. Then let's deffer it to @smaldini .

I'm not sure how to be with this situation then 😄

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/reactive/PubSubReactiveFactory.java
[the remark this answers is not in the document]
elefeint: I looked at the code again -- the responsibility for converting from `long` to `int` belongs with the polling task, as close to the Pub/Sub polling operation is possible. Consider a request of `Integer.MAX_VALUE + 1,000`. If we convert to `int` before creating `PubSubBlockingLimitedDemandPullTask`, then only a demand of `Integer.MAX_VALUE` will be fulfilled, leaving the extra demand of 1,000 unmet.

Leaving the conversion responsibility with `PubSubBlockingLimitedDemandPullTask` allows it to think in terms of `long` internally, subtracting up to `Integer.MAX_VALUE` messages at each poll.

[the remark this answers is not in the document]
meltsufin: I think the conversion belongs in `createPolledFlux`. How would the demand of `Integer.MAX_VALUE + 1,000` be fulfilled, if you cap it at `Integer.MAX_VALUE` here?  Automatically changing `demand > Integer.MAX_VALUE` to `Integer.MAX_VALUE` seems a bug waiting to happen. If the flux ever requests a demand greater than `Integer.MAX_VALUE` and it's not the special `Long.MAX_VALUE` we should probably just throw an unsupported exception in `createPolledFlux`.

spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-reactive-sample/src/main/java/com/example/ReactiveController.java
elefeint: I created #1517 to track; it will be a small follow-up PR.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
```diff
@@ -381,4 +383,19 @@ public SubscriptionAdminClient subscriptionAdminClient(
...
+
+	/**
+	 * Configuration for reactive support.
+	 */
+	@Configuration
+	@ConditionalOnClass(Flux.class)
+	protected static class ReactivePubSubAutoconfiguration {
+
+		@Bean
+		@ConditionalOnMissingBean
```
elefeint: Should there be a property preventing this object from instantiating?

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/reactive/PubSubReactiveFactory.java
```diff
@@ -0,0 +1,168 @@
...
+	 */
+	public Flux<AcknowledgeablePubsubMessage> createPolledFlux(String subscriptionName, long pollingPeriodMs) {
+
+		return Flux.create(sink -> {
+			sink.onRequest((numRequested) -> {
+				if (numRequested == Long.MAX_VALUE) {
+					// unlimited demand
+					Disposable task = Schedulers.single().schedulePeriodically(
+							new PubSubNonBlockingUnlimitedDemandPullTask(subscriptionName, sink), 0, pollingPeriodMs, TimeUnit.MILLISECONDS);
+					sink.onCancel(task::dispose);
```
elefeint: pass task directly

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/reactive/PubSubReactiveFactory.java
```diff
@@ -0,0 +1,168 @@
...
+	 * <p>For unlimited demand, the underlying subscription will be polled at a regular interval,
+	 * requesting up to {@code Integer.MAX_VALUE} messages at each poll.
+	 * <p>For specific demand, as many messages as are available will be returned immediately,
+	 * with remaining demand being fulfilled in the future.
+	 * Pub/Sub timeout will cause a retry with the same demand.
+	 * @param subscriptionName subscription from which to retrieve messages.
+	 * @param pollingPeriodMs how frequently to poll the source subscription in case of unlimited demand, in milliseconds.
+	 * @return infinite stream of {@link AcknowledgeablePubsubMessage} objects.
+	 */
+	public Flux<AcknowledgeablePubsubMessage> createPolledFlux(String subscriptionName, long pollingPeriodMs) {
```
elefeint: rename to poll(subscriptionName, pollingPeriodMs)

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/reactive/PubSubReactiveFactory.java
```diff
@@ -0,0 +1,168 @@
...
+
+		return Flux.create(sink -> {
+			sink.onRequest((numRequested) -> {
+				if (numRequested == Long.MAX_VALUE) {
+					// unlimited demand
+					Disposable task = Schedulers.single().schedulePeriodically(
+							new PubSubNonBlockingUnlimitedDemandPullTask(subscriptionName, sink), 0, pollingPeriodMs, TimeUnit.MILLISECONDS);
```
elefeint: remove PubSub prefix

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/reactive/PubSubReactiveFactory.java
```diff
@@ -0,0 +1,168 @@
...
+			long demand = this.initialDemand;
+			List<AcknowledgeablePubsubMessage> messages;
+
+			while (demand > 0 && !this.sink.isCancelled()) {
+				try {
+					int intDemand = demand > Integer.MAX_VALUE ? Integer.MAX_VALUE : (int) demand;
+					demand -= pullToSink(intDemand, true);
+				}
+				catch (DeadlineExceededException e) {
+					LOGGER.trace("Blocking pull timed out due to empty subscription " + this.subscriptionName + "; retrying.");
```
elefeint: add isTraceEnabled()

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/reactive/PubSubReactiveFactory.java
```diff
@@ -0,0 +1,168 @@
...
+		return Flux.create(sink -> {
+			sink.onRequest((numRequested) -> {
+				if (numRequested == Long.MAX_VALUE) {
+					// unlimited demand
+					Disposable task = Schedulers.single().schedulePeriodically(
+							new PubSubNonBlockingUnlimitedDemandPullTask(subscriptionName, sink), 0, pollingPeriodMs, TimeUnit.MILLISECONDS);
+					sink.onCancel(task::dispose);
+				}
+				else {
+					Schedulers.single().schedule(new PubSubBlockingLimitedDemandPullTask(subscriptionName, numRequested, sink));
```
elefeint: use elastic(), create single worker (don't call scheduler() directly)
create worker before request

also, kill task when sink is cancelled -- 10 minutes is too long

elefeint: reuse the worker for both, unlimited and limited demand

elefeint: autoconfigure the scheduler

spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-reactive-sample/src/main/java/com/example/ReactiveController.java
```diff
@@ -0,0 +1,58 @@
...
+	@Autowired
+	PubSubReactiveFactory reactiveFactory;
+
+	@GetMapping(value = "/getmessages", produces = MediaType.TEXT_EVENT_STREAM_VALUE)
+	public Flux<String> getMessages() {
+
+		Flux<AcknowledgeablePubsubMessage> flux
+				= this.reactiveFactory.createPolledFlux("exampleSubscription", 1000);
+
+		return flux
```
elefeint: reintroduce WebFlux dependency

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
artembilan: For what reason? 
Doesn't look like this light object is some kind of overhead for the target application.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/reactive/PubSubReactiveFactory.java
artembilan: No reason: this is in the `catch()`  block, so only in case of exception.
It's not going to be a big deal to build a logging message in case of this exception.

meltsufin: This exception is actually expected. So, it may happen quiet often.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
elefeint: Right, it's not necessary to keep it off by default because it's just a wrapper around a `SubscriberTemplate`. 
But I am also going to add a default scheduler bean into autoconfig and extract reactive configuration into a separate file, at which point it will make sense to add an overridable `matchIfMissing=true` property `spring.cloud.gcp.pubsub.reactive.enabled`.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/reactive/PubSubReactiveFactory.java
elefeint: Every 10 minutes by default, which is negligible in computer speed. On the other hand, we might as well add `isTraceEnabled()` for consistency.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubReactiveAutoConfiguration.java
```diff
@@ -0,0 +1,60 @@
...
+ */
+@Configuration
+@AutoConfigureAfter(GcpPubSubAutoConfiguration.class)
+@ConditionalOnClass(Flux.class)
+@ConditionalOnProperty(value = "spring.cloud.gcp.pubsub.reactive.enabled", matchIfMissing = true)
+public class GcpPubSubReactiveAutoConfiguration {
+
+	@Bean
+	@ConditionalOnMissingBean(name = "pubSubReactiveScheduler")
+	Scheduler  pubSubReactiveScheduler() {
```
meltsufin:
```suggestion
	Scheduler pubSubReactiveScheduler() {
```

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubReactiveAutoConfiguration.java
```diff
@@ -0,0 +1,60 @@
...
+		return Schedulers.elastic();
+	}
+
```
meltsufin:
```suggestion
```

elefeint:
I'll add an integration test in a follow-up PR -- I also want to change the sample, so users can send messages through a web UI and not have to have gcloud installed or multi-task with a third command-line window.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/reactive/PubSubReactiveFactory.java
vy: Hi! Great work @elefeint! We were also thinking of implementing our own Reactor Pubsub client due to improper back-pressure semantics which is giving us quite a lot of headache and thankfully @bsideup pointed this ticket to us. About this specific `messages.forEach(sink::next)` line, I have a question: If the `demand` is a big number, wouldn't you be fetching a gigantic batch of messages from Pubsub and choking the consumer via over-production? I would rather cap it by `Math.min(demand, maxDemand)` (e.g., `maxDemand=100` by default) for each `request(demand)` and wait for any follow up `request()`s that the consumer will signal at its convenience. Otherwise my impression is that we will be periodically bombarding the consumer with messages regardless of its availability. Is there a detail I am missing?

elefeint: If the consumer is specifying the large demand, isn't it a signal that it can handle that number of messages? Certainly throttling can be implemented, but it may be unintuitive to consumers that do want a larger batch.

vy: Most of the time consumers do not specify anything and simple `subscribe()` just triggers a `request(Long.MAX_VALUE)`. Though in the case of a message queue poller, "*Apparently the consumer is asking for a cargo ship load of messages.*" does not sound like a practical assumption to me. Consider the following case:

1. There are 1e6 messages in the queue.
2. `request(unbounded)` triggers a continuous `messages.forEach(sink::next)` sequence.
3. Scheduler threads start working on the initial tasks and the rest quickly get piled up in the work queue of the `ScheduledExecutorService`.
4. After a while, majority of the messages timeout, work queue starts choking memory, and every task threads pick up turn out to be obsolete.

This is the problem I am dealing for the last couple of months at work. Let me know if there is a way I can work around this.

elefeint: @vy I understand now. As a practical matter, would adding[ limitRequest() ](https://projectreactor.io/docs/core/release/api/reactor/core/publisher/Flux.html#limitRequest-long-) or [limitRate()](https://projectreactor.io/docs/core/release/api/reactor/core/publisher/Flux.html#limitRate-int-) operators to the invocation chain work?

@smaldini The conversation about default behavior of subscribe() is an interesting one.
Would it make sense to open an issue on reactor-core? Unbounded demand is not the most common usecase, so perhaps a `.subscribeWithInitialDemand()` operator makes sense?

vy: @elefeint I suppose a combination of `limit{Request,Rate}()` would do the trick, though it still expects an explicit action from the consumer side. I would still favor shipping a solution with decent defaults minimizing any effort from user side.

I think `subscribeWithInitialDemand()` is sort of against the Reactor's fluent API encouraging composition. Isn't it just a shortcut for `.limitRequest(n).subscribe()`?

Nevertheless, @smaldini, we would really appreciate your insights on the general back-pressure issue. (@simonbasle, is this something you can help us out?)

bsideup: FYI `limitRate` only, not `limitRequest`. `limitRequest` will cut the requests after the threshold specified as an argument

elefeint: @vy Would you mind creating an issue in this project for the `maxMessagesPerPoll` property? I've been thinking about this some more, and for unbounded demand scenario it makes sense to provide a configurable limit, especially if it's combined with a configurable project-wide default for `pollingPeriodMs`. 
The firehose of messages would likely still be the default, but smoothing message arrival by getting N messages per poll seems valuable.

PR #1466, opened 2019-02-13
Spanner integration test uses lazy

ChengyuanZhao:
Utilize lazy-fetched interleaved in integration tests Spanner

### DIFF
--- a/spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/test/domain/SubTrade.java
+++ b/spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/test/domain/SubTrade.java
@@ -38,7 +38,7 @@ public class SubTrade {
 	@PrimaryKey(keyOrder = 2)
 	String subTradeId;
 
-	@Interleaved
+	@Interleaved(lazy = true)
 	List<SubTradeComponent> subTradeComponentList;
 
 	public SubTrade() {
--- a/spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/test/domain/Trade.java
+++ b/spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/test/domain/Trade.java
@@ -62,7 +62,7 @@ public class Trade {
 
 	private List<Instant> executionTimes;
 
-	@Interleaved(lazy = true)
+	@Interleaved
 	private List<SubTrade> subTrades;
 
 	public static Trade aTrade() {
--- a/spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/test/domain/Trade.java
+++ b/spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/test/domain/Trade.java
@@ -62,7 +62,7 @@ public class Trade {
 
 	private List<Instant> executionTimes;
 
-	@Interleaved
+	@Interleaved(lazy = true)
 	private List<SubTrade> subTrades;
 
 	public static Trade aTrade() {


spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/test/domain/Trade.java
```diff
@@ -62,7 +62,7 @@
 
 	private List<Instant> executionTimes;
 
-	@Interleaved
+	@Interleaved(lazy = true)
```
dmitry-s: can we keep one of them eager?

ChengyuanZhao: I wanted to keep them both lazy because this would be 2-levels deep of lazy and I wanted to specifically touch that :/

spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/test/domain/SubTrade.java
```diff
@@ -38,7 +38,7 @@
 	@PrimaryKey(keyOrder = 2)
 	String subTradeId;
 
-	@Interleaved
+	@Interleaved(lazy = true)
```
dzou: How do you verify that this is actually doing something? (i.e. like verify the retrieval is truly lazy versus non-lazy)

ChengyuanZhao: There are existing unit tests in a previous PR that tested for this on both read and write .

Look at 
- SpannerTemplateTests.lazyFetchChildrenTest
- SpannerMutationFactoryImpl.lazyWriteTest

dzou: Gotcha, however I was more asking if there is a way to verify it's doing something at the integration test level.

Right now it doesn't seem quite right because existence of `lazy = true` doesn't have any affect on the integration test.

ChengyuanZhao: Yea, I don't think that's the responsibility of an "integration" test. I think verifying how something is done is up to unit tests, but verifying what is done (being blind to what's going on behind the scenes) is all that integration tests are responsible for.

dzou: Sounds good.

spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/test/domain/Trade.java
dmitry-s: do we have any integration tests for eager interleaved children?

ChengyuanZhao: At the moment all 3 of the generations of these tables are being linked together by these 2 levels of lazy links. 

I thought that it would be very unlikely that lazy works but eager doesn't considering how it is implemented (and then that specific implementation of lazy basically using all the same calls and functions as eager is verified by several unit tests). 

What do you think? I could of course add another table in here to be linked in eagerly, but I didn't want to create more clutter (because there is already Trade -> subtrade -> subtradecomponent and it may get messy).

dmitry-s: I think we need to have a test for eager read as well. Maybe a separate test class for that?

ChengyuanZhao: fixed. ptal

spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/test/domain/SubTrade.java
khakha010: https://github.com/spring-cloud/spring-cloud-gcp/blob/a592c888d8ed06011ad165e4f4eb964e986ed36e/spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/test/domain/SubTrade.java

PR #1469, opened 2019-02-19
Cloud Vision Document OCR Template

dzou:
Implement template for running OCR on documents stored on GCS through Cloud Vision APIs.

Note: The main design change from past discussions is in `runOcrForDocument` which will produce one JSON output file for each page of the document (i.e. it is paginated). I decided to do this because the API [actually forces paginated output (default is 20 pages of document per JSON output)](https://cloud.google.com/vision/docs/reference/rest/v1p2beta1/files/asyncBatchAnnotate#OutputConfig). This was something that we overlooked in our team discussions. In my discussion with Vision team, they said that producing 1 json file per page of document was a common practice, so I shifted to this practice here.


spring-cloud-gcp-vision/src/main/java/org/springframework/cloud/gcp/vision/DocumentOcrTemplate.java
```diff
@@ -0,0 +1,250 @@
...
+			@Override
+			public void onFailure(Throwable throwable) {
+				result.setException(throwable);
+			}
+
+			@Override
+			public void onSuccess(
+					AsyncBatchAnnotateFilesResponse asyncBatchAnnotateFilesResponse) {
+
+				String outputFolderUri = asyncBatchAnnotateFilesResponse.getResponsesList().get(0)
```
elefeint: Is it possible to have ore than one response?

spring-cloud-gcp-vision/src/main/java/org/springframework/cloud/gcp/vision/DocumentOcrResultSet.java
```diff
@@ -0,0 +1,99 @@
...
+	 * page of a document. For more information on reading this object, see:
+	 * https://cloud.google.com/vision/docs/reference/rpc/google.cloud.vision.v1#google.cloud.vision.v1.TextAnnotation
+	 *
+	 * @param pageNumber the zero-indexed page number of the document
+	 * @return the {@link TextAnnotation} representing the page of the document
+	 * @throws InvalidProtocolBufferException if the OCR information for the page failed to be
+	 *     parsed
+	 */
+	public TextAnnotation getPage(int pageNumber) throws InvalidProtocolBufferException {
+		Blob pageBlob = this.pageBlobs.get(pageNumber);
```
dmitry-s: maybe check for `pageNumber < getPageCount()` and throw appropriate exception?

spring-cloud-gcp-vision/src/main/java/org/springframework/cloud/gcp/vision/DocumentOcrTemplate.java
```diff
@@ -0,0 +1,250 @@
...
+	}
+
+	static TextAnnotation parseJsonBlob(Blob blob) throws InvalidProtocolBufferException {
+		AnnotateFileResponse.Builder annotateFileResponseBuilder = AnnotateFileResponse.newBuilder();
+		String jsonContent = new String(blob.getContent());
+		JsonFormat.parser().merge(jsonContent, annotateFileResponseBuilder);
+
+		AnnotateFileResponse annotateFileResponse = annotateFileResponseBuilder.build();
+
+		return annotateFileResponse.getResponses(0).getFullTextAnnotation();
```
elefeint: Is it possible to have more than 1 response?

spring-cloud-gcp-vision/src/main/java/org/springframework/cloud/gcp/vision/DocumentOcrResultSet.java
```diff
@@ -0,0 +1,99 @@
...
+ */
+public class DocumentOcrResultSet {
+
+	private final Storage storageClient;
+
+	private final List<Blob> pageBlobs;
+
+	public DocumentOcrResultSet(List<Blob> pages, Storage storageClient) {
+		this.pageBlobs = pages;
+		this.storageClient = storageClient;
```
dmitry-s: Why do we have `storageClient` here? It seems it is not used anywhere.

spring-cloud-gcp-vision/src/test/java/org/springframework/cloud/gcp/vision/it/DocumentOcrTemplateIntegrationTests.java
```diff
@@ -0,0 +1,94 @@
...
+import org.springframework.cloud.gcp.vision.DocumentOcrTemplate;
+import org.springframework.test.context.ContextConfiguration;
+import org.springframework.test.context.junit4.SpringRunner;
+import org.springframework.util.concurrent.ListenableFuture;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assumptions.assumeThat;
+
+@RunWith(SpringRunner.class)
+@ContextConfiguration(classes = { VisionTestConfiguration.class })
```
elefeint: Is it possible to rely on real configuration? It's nice when integration tests verify the way starter will be used in real life.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/vision/CloudVisionAutoConfiguration.java
```diff
@@ -86,4 +88,11 @@ public CloudVisionTemplate cloudVisionTemplate(ImageAnnotatorClient imageAnnotat
 		return new CloudVisionTemplate(imageAnnotatorClient);
 	}
 
+	@Bean
+	@ConditionalOnMissingBean
+	public DocumentOcrTemplate documentOcrTemplate(
+			ImageAnnotatorClient imageAnnotatorClient, Storage storage) {
```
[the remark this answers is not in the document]
dzou: Yup, it's provided above!

spring-cloud-gcp-vision/src/main/java/org/springframework/cloud/gcp/vision/DocumentOcrResultSet.java
```diff
@@ -0,0 +1,99 @@
...
+	 * https://cloud.google.com/vision/docs/reference/rpc/google.cloud.vision.v1#google.cloud.vision.v1.TextAnnotation
+	 *
+	 * @param pageNumber the zero-indexed page number of the document
+	 * @return the {@link TextAnnotation} representing the page of the document
+	 * @throws InvalidProtocolBufferException if the OCR information for the page failed to be
+	 *     parsed
+	 */
+	public TextAnnotation getPage(int pageNumber) throws InvalidProtocolBufferException {
+		Blob pageBlob = this.pageBlobs.get(pageNumber);
+		return DocumentOcrTemplate.parseJsonBlob(pageBlob);
```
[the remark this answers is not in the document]
dzou: Do you mean the whole method or specifically `DocumentOcrTemplate.parseJsonBlob`?

Open to suggestions - to give some more context, the `parseJsonBlob` method converts a "blob" file on GCS to a TextAnnotation proto. It's package-private static helper and shared between this class and DocumentOcrTemplate.

spring-cloud-gcp-vision/src/main/java/org/springframework/cloud/gcp/vision/DocumentOcrResultSet.java
```diff
@@ -0,0 +1,99 @@
...
+
+			@Override
+			public boolean hasNext() {
+				return currentPage < getPageCount();
+			}
+
+			@Override
+			public TextAnnotation next() {
+				try {
+					TextAnnotation result = getPage(currentPage);
```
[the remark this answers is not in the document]
dzou: Done.

spring-cloud-gcp-vision/src/main/java/org/springframework/cloud/gcp/vision/DocumentOcrTemplate.java
```diff
@@ -0,0 +1,250 @@
...
+		if (!document.isFile()) {
+			throw new IllegalArgumentException(
+					"Provided document location is not a valid file location: " + document);
+		}
+
+		GcsSource gcsSource = GcsSource.newBuilder()
+				.setUri(document.uriString())
+				.build();
+
+		Blob documentBlob = this.storage.get(
```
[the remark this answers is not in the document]
dzou: Done.

spring-cloud-gcp-vision/src/main/java/org/springframework/cloud/gcp/vision/DocumentOcrTemplate.java
```diff
@@ -0,0 +1,250 @@
...
+		}
+
+		String contentType = documentBlob.getContentType();
+		InputConfig inputConfig = InputConfig.newBuilder()
+				.setMimeType(contentType)
+				.setGcsSource(gcsSource)
+				.build();
+
+		GcsDestination gcsDestination = GcsDestination.newBuilder()
+				.setUri(outputFilePathPrefix.uriString())
```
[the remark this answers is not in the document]
dzou: The outputFilePathPrefix as very little restrictions (it can describe a file/folder/bucket) - in this case we can skip it because we can defer the error handling to the Vision API which will do validation on the URI.

spring-cloud-gcp-vision/src/main/java/org/springframework/cloud/gcp/vision/DocumentOcrTemplate.java
```diff
@@ -0,0 +1,250 @@
...
+				.build();
+
+		OutputConfig outputConfig = OutputConfig.newBuilder()
+				.setGcsDestination(gcsDestination)
+				.setBatchSize(1)
```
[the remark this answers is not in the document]
dzou: Done. This field means it will produce 1 JSON file per page in the document. If unset, this will [default to 20 pages per JSON file](https://cloud.google.com/vision/docs/reference/rpc/google.cloud.vision.v1#google.cloud.vision.v1.OutputConfig). And the range of values is 1-100.

I set it to 1 (aka 1-page-per-json-file) because it seemed like an intuitive default for someone who doesn't want to think about it. However upon reinspection it might make sense to make it configurable by the user ? (I suppose someone with a 1000-page pdf wouldn't want to output 1000 json files.)

spring-cloud-gcp-vision/src/main/java/org/springframework/cloud/gcp/vision/DocumentOcrTemplate.java
```diff
@@ -0,0 +1,250 @@
...
+				.build();
+
+		Feature feature = Feature.newBuilder()
```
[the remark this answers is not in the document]
dzou: Done.

spring-cloud-gcp-vision/src/main/java/org/springframework/cloud/gcp/vision/DocumentOcrTemplate.java
dzou: Nope, this is safe - this is due to that 1 request per batch request that we discussed during our meetings. Only our code will be providing the response object here.

spring-cloud-gcp-vision/src/main/java/org/springframework/cloud/gcp/vision/DocumentOcrTemplate.java
dzou: Nope, this is also safe for the same reason as above. Only our code will be able to provide the response object (with 1 response inside).

spring-cloud-gcp-vision/src/test/java/org/springframework/cloud/gcp/vision/it/DocumentOcrTemplateIntegrationTests.java
dzou: Agreed, unfortunately I couldn't quite get this to work. Currently, the autoconfiguration project depends on the Vision project ([to provide beans of Vision templates](https://github.com/spring-cloud/spring-cloud-gcp/blob/master/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/vision/CloudVisionAutoConfiguration.java#L85)). If I reference the real configuration here, I would have to make the vision project depend on autoconfiguration project which produces a circular dependency.

Do you have any suggestions? Maybe I am missing something with how Spring dependency injection works... In the Guice world I would only have the vision autoconfiguration provide `ImageAnnotatorClient` and be able to delete the autoconfiguration's dependency on the vision module. However, it seems like in Spring, I must declare the beans for the templates explicitly which forces the dependency between autoconfiguration and vision.

spring-cloud-gcp-vision/src/main/java/org/springframework/cloud/gcp/vision/DocumentOcrResultSet.java
dzou: Done.

spring-cloud-gcp-vision/src/main/java/org/springframework/cloud/gcp/vision/DocumentOcrResultSet.java
dzou: Nice catch. Done.

spring-cloud-gcp-vision/src/main/java/org/springframework/cloud/gcp/vision/DocumentOcrTemplate.java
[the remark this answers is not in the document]
elefeint: Configurable would be nice as a follow-up non-urgent PR.

spring-cloud-gcp-vision/src/test/java/org/springframework/cloud/gcp/vision/it/DocumentOcrTemplateIntegrationTests.java
elefeint: Makes sense, thanks! This way we exercise both, the pure template usage, and the template+autoconfig usage in sample test.

dzou:
Hi guys, I've adjusted this PR based on your comments. Also modified it such that it can support user specifying how many document pages to include in each JSON OCR output file. PTAL.

spring-cloud-gcp-vision/src/main/java/org/springframework/cloud/gcp/vision/DocumentOcrTemplate.java
```diff
@@ -60,12 +61,17 @@
...
+			.setType(Type.DOCUMENT_TEXT_DETECTION)
+			.build();
+
 	private static final Pattern OUTPUT_PAGE_PATTERN = Pattern.compile("output-\\d+-to-\\d+\\.json");
 
 	private final ImageAnnotatorClient imageAnnotatorClient;
 
 	private final Storage storage;
 
+	@Autowired
```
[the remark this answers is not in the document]
dzou: Done

spring-cloud-gcp-vision/src/main/java/org/springframework/cloud/gcp/vision/DocumentOcrTemplate.java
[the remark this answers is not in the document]
dzou: I ultimately changed it to configurable b/c I realized this is an important feature, i.e. batching seems important because it will reduce the amount of I/O needed for large documents.

dzou:
Made changes as discussed in our meeting; please take another look.

spring-cloud-gcp-vision/src/main/java/org/springframework/cloud/gcp/vision/DocumentOcrResultSet.java
```diff
@@ -0,0 +1,198 @@
...
+	 *     parsed
+	 *
+	 */
+	public TextAnnotation getPage(int pageNumber) throws InvalidProtocolBufferException {
+		if (pageNumber < this.minPage || pageNumber > this.maxPage) {
+			throw new IndexOutOfBoundsException("Page number out of bounds: " + pageNumber);
+		}
+
+		Map.Entry<Integer, OcrPageRange> floorEntry = this.ocrPageRanges.floorEntry(pageNumber);
+		OcrPageRange pageRange = floorEntry.getValue();
```
meltsufin: Why don't you check the `pageNumber` against the `pageRange` here, before parsing?

spring-cloud-gcp-vision/src/main/java/org/springframework/cloud/gcp/vision/DocumentOcrResultSet.java
```diff
@@ -0,0 +1,198 @@
...
+		List<TextAnnotation> documentPages = DocumentOcrTemplate.parseJsonBlob(pageRange.blob);
```
meltsufin: I think we discussed caching the parsed json blobs. Can we do that?

spring-cloud-gcp-starters/spring-cloud-gcp-starter-vision/pom.xml
```diff
@@ -27,5 +27,10 @@
 			<groupId>com.google.cloud</groupId>
 			<artifactId>google-cloud-vision</artifactId>
 		</dependency>
+
+		<dependency>
+			<groupId>org.springframework.cloud</groupId>
+			<artifactId>spring-cloud-gcp-storage</artifactId>
```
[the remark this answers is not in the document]
meltsufin: Do you need this dep here at all? Since the starter doesn't have any code, I don't think it makes sense to have any "optional" dependencies in it.

spring-cloud-gcp-vision/src/main/java/org/springframework/cloud/gcp/vision/DocumentOcrResultSet.java
```diff
@@ -0,0 +1,198 @@
...
+import com.google.protobuf.InvalidProtocolBufferException;
+
+/**
+ * Represents the parsed OCR content for an document in the provided range of pages.
+ *
+ * @author Daniel Zou
+ */
+public class DocumentOcrResultSet {
+
+	private static final Pattern OUTPUT_PAGE_PATTERN = Pattern.compile("output-(\\d+)-to-(\\d+)\\.json");
```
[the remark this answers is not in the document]
dzou: For now I don't see a reason why the user would want to rename the output files after they have been generated, so I think it makes sense to just keep it simple for now.

spring-cloud-gcp-vision/src/main/java/org/springframework/cloud/gcp/vision/DocumentOcrResultSet.java
```diff
@@ -0,0 +1,198 @@
...
+
+	private final TreeMap<Integer, OcrPageRange> ocrPageRanges;
```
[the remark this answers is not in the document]
dzou: Tried to, however I am using `.floorEntry(...)` which appears to only be in TreeMap and not in SortedMap.

spring-cloud-gcp-vision/src/main/java/org/springframework/cloud/gcp/vision/DocumentOcrResultSet.java
```diff
@@ -0,0 +1,225 @@
...
+
+	private final TreeMap<Integer, List<TextAnnotation>> documentPagesCache;
```
meltsufin: Can we instead just use one simple `HashMap<Integer, OcrPageRange>`, where there is an entry for every single page? The `getTextAnnotations()` can move to the `OcrPageRange` class, and caching can very easily be done there. I think it would simplify your code quiet a bit.

spring-cloud-gcp-vision/src/main/java/org/springframework/cloud/gcp/vision/DocumentOcrResultSet.java
dzou: Done.

spring-cloud-gcp-vision/src/main/java/org/springframework/cloud/gcp/vision/DocumentOcrResultSet.java
dzou: Done.

spring-cloud-gcp-starters/spring-cloud-gcp-starter-vision/pom.xml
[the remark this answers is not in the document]
dzou: Removed.

spring-cloud-gcp-vision/src/main/java/org/springframework/cloud/gcp/vision/DocumentOcrResultSet.java
dzou: Partially done. I moved the caching into OcrPageRange which simplified the caching code.

However, I didn't switch from TreeMap to HashMap. I don't think using TreeMap here is that much more complex that HashMap, so might as well do things efficiently. For users processing with JSON `batchSize = 100` we would reduce the number of entries stored in map by factor of 100.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/vision/CloudVisionProperties.java
```diff
@@ -35,7 +35,27 @@
 	@NestedConfigurationProperty
 	private final Credentials credentials = new Credentials(GcpScope.CLOUD_VISION.getUrl());
 
+	private int executorThreadsCount = 1;
+
+	private int jsonOutputBatchSize = 20;
```
meltsufin: I think there is a javadoc requirement for these? Can you look at other `...Properties` classes for an example?

dzou: Gotcha, will add these with the docs. As reminder to add javadoc to properties, I created https://github.com/spring-cloud/spring-cloud-gcp/issues/1548 to track.

spring-cloud-gcp-vision/pom.xml
```diff
@@ -24,5 +24,18 @@
 			<groupId>org.springframework</groupId>
 			<artifactId>spring-core</artifactId>
 		</dependency>
+
+		<dependency>
+			<groupId>org.springframework.cloud</groupId>
+			<artifactId>spring-cloud-gcp-storage</artifactId>
+			<optional>true</optional>
```
elefeint: Is it optional? VisionTemplate depends on Storage in all cases.

meltsufin: Just for OCR, I believe.

dzou: Yeah I wasn't 100% sure to be honest; it's what I had before but Mike suggested it could be marked optional since only the OCR template uses storage but not the general one. I figure once I merge this with my sample I'll be able to figure it out.

Issue #1113, opened 2018-10-17
Add support @EnableDatastoreAuditing annotation


sergeymeshkov:
Please add support of @EnableDatastoreAuditing annotation which simplifies auditing while using Datastore repositories.

meltsufin:
Thank you for the issue submission. Can you please provide some more context on what this annotation would do? Thanks!

sergeymeshkov:
This annotation can be the same as [@EnableJpaAuditing](https://docs.spring.io/spring-data/data-jpa/docs/current/api/org/springframework/data/jpa/repository/config/EnableJpaAuditing.html) or [@EnableMongoAuditing](https://docs.spring.io/spring-data/mongodb/docs/current/api/org/springframework/data/mongodb/config/EnableMongoAuditing.html)

Annotation elements:
**auditorAwareRef** - Configures the AuditorAware bean to be used to lookup the current principal.
**dateTimeProviderRef** - Configures a DateTimeProvider bean name that allows customizing the org.joda.time.DateTime to be used for setting creation and modification dates.
**modifyOnCreate** - Configures whether the entity shall be marked as modified on creation.
**setDates** - Configures whether the creation and modification dates are set.

ChengyuanZhao:
We just implemented this for Spanner so this should be a walk in the park.

PR #1472, opened 2019-02-20
Datastore template publisher aware, delete events

ChengyuanZhao:
related #1113
datastore template is publisher aware. added delete events.

### DIFF
--- a/docs/src/main/asciidoc/datastore.adoc
+++ b/docs/src/main/asciidoc/datastore.adoc
@@ -1117,6 +1117,16 @@ You can also write trades using `curl -XPOST -H"Content-Type: application/json" 
 
 To delete trades, you can use `curl -XDELETE \http://<server>:<port>/trades/<trader_id>`
 
+=== Events
+
+Spring Data Cloud Datastore publishes events extending the Spring Framework's `ApplicationEvent` to the context that can be received by `ApplicationListener` beans you register.
+
+[options="header"]
+|===
+| Type| Description | Contents
+| `BeforeDeleteEvent`|Published immediately before delete operations are executed by `DatastoreTemplate`| The keys to be sent to Cloud Datastore. The target entities, ID values, or entity type originally specified for the delete operation.
+| `AfterDeleteEvent`|Published immediately after delete operations are executed by `DatastoreTemplate`| The keys sent to Cloud Datastore. The target entities, ID values, or entity type originally specified for the delete operation.
+|===
 
 
 === Sample
--- a/docs/src/main/asciidoc/datastore.adoc
+++ b/docs/src/main/asciidoc/datastore.adoc
@@ -1123,7 +1123,7 @@ Spring Data Cloud Datastore publishes events extending the Spring Framework's `A
 
 [options="header"]
 |===
-| Type| Description | Contents
+| Type | Description | Contents
 | `BeforeDeleteEvent`|Published immediately before delete operations are executed by `DatastoreTemplate`| The keys to be sent to Cloud Datastore. The target entities, ID values, or entity type originally specified for the delete operation.
 | `AfterDeleteEvent`|Published immediately after delete operations are executed by `DatastoreTemplate`| The keys sent to Cloud Datastore. The target entities, ID values, or entity type originally specified for the delete operation.
 |===
--- a/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
+++ b/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
@@ -59,6 +59,9 @@ import org.springframework.cloud.gcp.data.datastore.core.mapping.DatastoreMappin
 import org.springframework.cloud.gcp.data.datastore.core.mapping.DatastorePersistentEntity;
 import org.springframework.cloud.gcp.data.datastore.core.mapping.DatastorePersistentProperty;
 import org.springframework.cloud.gcp.data.datastore.core.util.ValueUtil;
+import org.springframework.context.ApplicationEvent;
+import org.springframework.context.ApplicationEventPublisher;
+import org.springframework.context.ApplicationEventPublisherAware;
 import org.springframework.data.domain.Example;
 import org.springframework.data.domain.ExampleMatcher;
 import org.springframework.data.domain.Sort;
@@ -67,6 +70,7 @@ import org.springframework.data.mapping.PersistentProperty;
 import org.springframework.data.mapping.PersistentPropertyAccessor;
 import org.springframework.data.support.ExampleMatcherAccessor;
 import org.springframework.data.util.ClassTypeInformation;
+import org.springframework.lang.Nullable;
 import org.springframework.transaction.interceptor.TransactionAspectSupport;
 import org.springframework.transaction.support.DefaultTransactionStatus;
 import org.springframework.transaction.support.TransactionSynchronizationManager;
@@ -80,7 +84,7 @@ import org.springframework.util.TypeUtils;
  *
  * @since 1.1
  */
-public class DatastoreTemplate implements DatastoreOperations {
+public class DatastoreTemplate implements DatastoreOperations, ApplicationEventPublisherAware {
 
 	private final DatastoreReaderWriter datastore;
 
@@ -89,6 +93,8 @@ public class DatastoreTemplate implements DatastoreOperations {
 	private final DatastoreMappingContext datastoreMappingContext;
 
 	private final ObjectToKeyFactory objectToKeyFactory;
+
+	private @Nullable ApplicationEventPublisher eventPublisher;
 
 	public DatastoreTemplate(DatastoreReaderWriter datastore,
 			DatastoreEntityConverter datastoreEntityConverter,
@@ -113,6 +119,11 @@ public class DatastoreTemplate implements DatastoreOperations {
 	 */
 	public DatastoreEntityConverter getDatastoreEntityConverter() {
 		return this.datastoreEntityConverter;
+	}
+
+	@Override
+	public void setApplicationEventPublisher(ApplicationEventPublisher applicationEventPublisher) {
+		this.eventPublisher = applicationEventPublisher;
 	}
 
 	@Override
@@ -678,6 +689,12 @@ public class DatastoreTemplate implements DatastoreOperations {
 		}
 		if (matcher.getPropertySpecifiers().hasValues()) {
 			throw new DatastoreDataException("Property matchers are not supported");
+		}
+	}
+
+	private void maybeEmitEvent(ApplicationEvent event) {
+		if (this.eventPublisher != null) {
+			this.eventPublisher.publishEvent(event);
 		}
 	}
 
--- a/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
+++ b/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
@@ -58,6 +58,8 @@ import org.springframework.cloud.gcp.data.datastore.core.mapping.DatastoreDataEx
 import org.springframework.cloud.gcp.data.datastore.core.mapping.DatastoreMappingContext;
 import org.springframework.cloud.gcp.data.datastore.core.mapping.DatastorePersistentEntity;
 import org.springframework.cloud.gcp.data.datastore.core.mapping.DatastorePersistentProperty;
+import org.springframework.cloud.gcp.data.datastore.core.mapping.event.AfterDeleteEvent;
+import org.springframework.cloud.gcp.data.datastore.core.mapping.event.BeforeDeleteEvent;
 import org.springframework.cloud.gcp.data.datastore.core.util.ValueUtil;
 import org.springframework.context.ApplicationEvent;
 import org.springframework.context.ApplicationEventPublisher;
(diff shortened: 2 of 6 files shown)


ChengyuanZhao:
@meltsufin @dmitry-s PTAL

docs/src/main/asciidoc/datastore.adoc
```diff
@@ -1117,6 +1117,16 @@ You can also write trades using `curl -XPOST -H"Content-Type: application/json"
 
 To delete trades, you can use `curl -XDELETE \http://<server>:<port>/trades/<trader_id>`
 
+=== Events
+
+Spring Data Cloud Datastore publishes events extending the Spring Framework's `ApplicationEvent` to the context that can be received by `ApplicationListener` beans you register.
+
+[options="header"]
+|===
+| Type| Description | Contents
```
meltsufin:
```suggestion
| Type | Description | Contents
```

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
```diff
@@ -90,6 +96,8 @@
 
 	private final ObjectToKeyFactory objectToKeyFactory;
 
+	private @Nullable ApplicationEventPublisher eventPublisher;
```
meltsufin: What's the purpose of `@Nullable` here?
Also, once the bean is initialized, it should not be null.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
```diff
@@ -157,34 +170,38 @@ public DatastoreEntityConverter getDatastoreEntityConverter() {
...
 
 	@Override
 	public long deleteAll(Class<?> entityClass) {
 		Key[] keysToDelete = findAllKeys(entityClass);
-		getDatastoreReadWriter().delete(keysToDelete);
+		performDelete(keysToDelete, null, null, entityClass);
 		return keysToDelete.length;
 	}
 
+	private void performDelete(Key[] keys, Iterable ids, Iterable entities, Class entityClass) {
```
meltsufin: Requiring `Key[]` is causing you all these conversion from `List`.
Also, the only thing common across all deletes is the availability of the key and entity class.
Can we just provide those two pieces of information in the event for now?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
```diff
@@ -157,34 +170,38 @@ public DatastoreEntityConverter getDatastoreEntityConverter() {
...
-		getDatastoreReadWriter().delete(getKey(entity, false));
+		performDelete(new Key[] { getKey(entity, false) }, null, Collections.singletonList(entity), entity.getClass());
 	}
 
 	@Override
 	public <T> void deleteAll(Iterable<T> entities) {
-		getDatastoreReadWriter()
-				.delete(StreamSupport.stream(entities.spliterator(), false)
-				.map((x) -> getKey(x, false)).toArray(Key[]::new));
+		performDelete(StreamSupport.stream(entities.spliterator(), false)
```
meltsufin: Here you should probably generate several events, one for each entityClass present in the collection.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/mapping/event/AfterDeleteEvent.java
```diff
@@ -0,0 +1,42 @@
...
+	 *
+	 * @param keysToDelete The keys that are deleted in this operation (never {@code null}).
+	 * @param targetEntityClass The target entity type deleted. This may be {@code null}
+	 *     depending on the specific delete operation.
+	 * @param targetIds The target entity ID values deleted. This may be {@code null}
+	 *     depending on the specific delete operation.
+	 * @param targetEntities The target entity objects deleted. This may be {@code null}
+	 *     depending on the specific
+	 */
+	public AfterDeleteEvent(Key[] keysToDelete, Class targetEntityClass, Iterable targetIds, Iterable targetEntities) {
```
meltsufin: I'd avoid any nullable fields and instead have separate event classes to represent different delete events.
Otherwise, let's just keep the delete event simple and restrict the data in it to keys and entityClass.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
ChengyuanZhao: The conversion to Key[] needs to happen anyway because the client lib can only take the array (not the List). 

Does it hurt to have the IDs or java POJOs that were used to delete? Those can be very useful to have because the user can directly inspect the things being deleted which is either impossible or difficult if you only have the Key and entity type. Furthermore some types of delete (like by several POJOs where the types differ ) don't offer the entity class at all.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
ChengyuanZhao: I saw this from the other Spring Data implementations. I think this is here because there isn't necessarily always a publisher in the context (like if the user just isn't handling events from Spring).

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
ChengyuanZhao: I don't think that would be appropriate. All of the deletes happen in one instant/timestamp for Datastore, which I think would be very confusing if that single operation corresponded to multiple events. It might be impossible to connect the separate events later on to realize they're from the same delete operation.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
dmitry-s: The only issue with that is creating Iterables of ids or entities, even in case they will not be used. I think it can be useful to have them, but I'm not sure about those extra objects.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/mapping/event/AfterDeleteEvent.java
ChengyuanZhao: There are several issues with breaking it up. The reason is there are many delete operations where only 1 of the 4 fields are null. Several of the fields appear populated in several different delete operations.

In other words the delete operations overlap each other in which fields are populated and there wouldn't be a good way to give meaningful names to the delete events.  The fields and delete operations they correspond to are NOT disjoint.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
ChengyuanZhao: What is your proposal ? @dmitry-s 

I wanted to avoid a zoo of delete events with both BEFORE and AFTER versions...

```
BeforeSingleEntityDeleteEvent
AfterSingleEntityDeleteEvent
BeforeMultiEntityDeleteEvent
AfterMultiEntityDeleteEvent
BeforeSingleIDDeleteEvent
AfterSingleIDDeleteEvent
BeforeByClassDeleteEvent
AfterByClassDeleteEvent
DeleteEventWithEntityClass
DeleteEventWithEntities
DeleteEventWithIDs
...
etc....
```

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
meltsufin: We then have a problem with the definition of an event where we provide the entityClass since it may not be always be there. Maybe make the event just have the keys then. We can always add events that are subclasses of that event to provide more fields where appropriate.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
meltsufin: From OOP perspective, you would probably have an event for each delete method if the fields vary in each. That being said, I think for now we can just make an event that only has the keys. We can always add others as needed that extend that event.

ChengyuanZhao: @meltsufin  the problem with just holding keys is that the receiver of the events can't do anything with them. the user cannot retrieve any information using those  keys because of course the keys no longer exist in Datastore.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
ChengyuanZhao: @meltsufin  Potentially null fields in events does occur in the other modules. 
In cassandra the table name for the delete event may be null https://github.com/spring-projects/spring-data-cassandra/blob/master/spring-data-cassandra/src/main/java/org/springframework/data/cassandra/core/mapping/event/AbstractDeleteEvent.java

in JDBC the base relational event has a nullable mutation field !
https://github.com/spring-projects/spring-data-jdbc/blob/master/spring-data-relational/src/main/java/org/springframework/data/relational/core/mapping/event/SimpleRelationalEvent.java

ChengyuanZhao: @meltsufin you may want to sit down for this one:
https://github.com/spring-projects/spring-data-jdbc/blob/master/spring-data-relational/src/main/java/org/springframework/data/relational/core/mapping/event/BeforeDeleteEvent.java

JDBC delete event has an OPTIONAL entity field!

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/mapping/event/AfterDeleteEvent.java
meltsufin: You shouldn't have more events than delete methods and using the delete method name to inform the name of the event seems to make sense.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
meltsufin: Thanks for the examples. Yes, they do have at most one nullable field. I'm not sure of their reasons for it. Sometimes it's a compromise, but I think in our case we have a few more. Maybe it would be more palatable if we just had one nullable field.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/mapping/event/AfterDeleteEvent.java
ChengyuanZhao: @meltsufin the other modules all have single pair of "BeforeDeleteEvent" and "AfterDeleteEvent". 
Also, because of the before/after pair, we would have TEN CLASSES for the 5 delete methods.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
ChengyuanZhao: While there are 3 fields that are sometimes null I must mention they're never all null at the same time. Like the examples usually just 1 field is null.  

Let me check with the Spring Data guys about this.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/mapping/event/AfterDeleteEvent.java
meltsufin: Agreed the number of classes doubling is inconvenient.
Let's at least start by moving the nullable fields out of the constructor. The constructor should only take required fields. Also, it might make sense to provide `getOptional....` methods for nullable fields.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
ChengyuanZhao: @mp911de 

We have here just one pair of Before/After DeleteEvent here based on looking at the JDBC and Cassandra modules. However based on the specific delete request sometimes 1 of the fields in the event is NULL. (sometimes the entity, sometimes ID, sometimes entity type). 
How should this be handled?  While there are 3 nullable fields here I must note that they are NEVER all optional. Usually only 1 of them is null.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/mapping/event/AfterDeleteEvent.java
ChengyuanZhao: Sounds good. Let me go ahead and do that.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
mp911de: There's not really standard across events. This is due to the various implementations of how stores work (enriching native driver methods with Spring functionality) and how they treat deletes. In some cases, we just propagate the id, sometimes a query and sometimes the entity itself.

MongoDB is quite convenient as a delete by id `Query`/delete by query `Query` is always a document and so we can pass on always the query document. In Cassandra, we pass the `Statement` object as `BeforeDeleteEvent`.
In JDBC, an entity is comprised potentially of multiple tables so a single delete statement does not work there.

Does this help?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/mapping/event/AfterDeleteEvent.java
ChengyuanZhao: Done, PTAL @meltsufin

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/mapping/event/DeleteEvent.java
```diff
@@ -0,0 +1,141 @@
...
+	 */
+	public Key[] getKeys() {
+		return (Key[]) getSource();
+	}
+
+	/**
+	 * Get the target entity type deleted.
+	 * @return This may be {@code null} depending on the specific delete operation.
+	 */
+	public Class getTargetEntityClassIfPresent() {
```
meltsufin: I was thinking more like what is done in Cassandra:
```
	public Optional<Object> getOptionalTargetEntity()
```
Otherwise, it's just a method rename.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/mapping/event/DeleteEvent.java
```diff
@@ -0,0 +1,141 @@
...
+	 */
+	public void setTargetIds(Iterable targetIds) {
+		this.targetIds = targetIds;
+	}
+
+	/**
+	 * Set the target entity class of the delete operation.
+	 * @param targetEntityClass the target entity type of the delete operation.
+	 */
+	public void setTargetEntityClass(Class targetEntityClass) {
```
meltsufin: Since you created the setters, can you use them instead of the constructor, and restrict the constructor to only non-optional fields?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/mapping/event/DeleteEvent.java
ChengyuanZhao: Understood. As discussed in person getters now give Optionals.

PR #1476, opened 2019-02-21
Spanner session leak fix and tests

### LINKED ISSUE #1474: SingleUse reads causes a session leak
Using the `findById` method of the spring data spanner project causes the app to block after the method has been called the same time as the maximum number of spanner sessions. Upon further investigations, it looks like the problem is in the `org.springframework.cloud.gcp.data.spanner.core.SpannerTemplate`,  where a session is opened on line 414 and another session is opened on line 417. Further method calls only use the second session, which is closed properly after the read is completed, but the first session leaks and is never closed, causing the app to exhaust the session pool.

To recreate the problem, set the `maxSessions` to 100 (or even lower) and call the `findById` 101 times

The spring-cloud-gcp-data-spanner version in use is 1.1.0.RELEASE

ChengyuanZhao:
fixes #1474 
typo fix. removes accidental call to method rather than using existing variable.

### DIFF
--- a/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
+++ b/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
@@ -447,7 +447,7 @@ public class SpannerTemplate implements SpannerOperations, ApplicationEventPubli
 				: getReadContext();
 
 		if (options == null) {
-			resultSet = getReadContext().read(tableName, keys, columns);
+			resultSet = readContext.read(tableName, keys, columns);
 		}
 		else if (options.getIndex() != null) {
 			resultSet = readContext.readUsingIndex(tableName, options.getIndex(), keys,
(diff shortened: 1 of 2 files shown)


spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
```diff
@@ -447,7 +447,7 @@ private ResultSet executeRead(String tableName, KeySet keys, Iterable<String> co
 				: getReadContext();
```
meltsufin: Should `getReadContext()` be renamed to `createReadContext()` to help avoid this kind of bug in the future?

ChengyuanZhao: It doesn't necessarily always create one. In that method we check the Spring AOP support in there to see if there is an existing transaction context available to use and create the single-use if there isn't.  @meltsufin

meltsufin: So why is the second call to `getReadContext()` not finding the transaction started by the previous call in this method?

ChengyuanZhao: @meltsufin  Unlike Hibernate not every single operation happens inside explicit transactions that are opened and closed by the user or the framework .

When the user uses the transaction-anonymous-function support or the Transactional annotation then yes there is no problem before or now because we get the same transaction context.

however if the user was using this outside of any transaction boundaries (the  2 options mentioned above) then we use a single-use read context.

PR #1478, opened 2019-02-21
Create a GoogleStorageLocation class for Storage module

dzou:
This creates a `GoogleStorageLocation` class for the Storage module.

I did this because I wanted to reuse the logic to specify and validate URIs/locations of blobs in GCS.

This also offers users an easier way to construct GCS locations (by bucket name + file path). From my research, it looks like people [have trouble finding the documentation for how to construct a GCS URI](https://stackoverflow.com/questions/25373467/how-do-i-identify-the-google-cloud-storage-uri-from-my-google-developers-console).

### DIFF
--- a/spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageLocation.java
+++ b/spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageLocation.java
@@ -26,6 +26,8 @@ import org.springframework.util.Assert;
  * Represents a Google Cloud Storage location provided by the user.
  *
  * @author Daniel Zou
+ *
+ * @since 1.2
  */
 public class GoogleStorageLocation {
 
@@ -82,12 +84,8 @@ public class GoogleStorageLocation {
 			this.bucketName = bucketName;
 			this.blobName = pathToFile;
 
-			if (pathToFile == null) {
-				this.uri = new URI(String.format(GCS_URI_FORMAT, bucketName, ""));
-			}
-			else {
-				this.uri = new URI(String.format(GCS_URI_FORMAT, bucketName, pathToFile));
-			}
+			pathToFile = (pathToFile == null) ? "" : pathToFile;
+			this.uri = new URI(String.format(GCS_URI_FORMAT, bucketName, pathToFile));
 		}
 		catch (URISyntaxException e) {
 			String errorMessage = String.format(
--- a/spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageLocation.java
+++ b/spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageLocation.java
@@ -1,0 +1,148 @@
+/*
+ * Copyright 2017-2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.gcp.storage;
+
+import java.net.URI;
+import java.net.URISyntaxException;
+
+import org.springframework.util.Assert;
+
+/**
+ * Represents a Google Cloud Storage location provided by the user.
+ *
+ * @author Daniel Zou
+ */
+public final class GoogleStorageLocation {
+
+	private static final String GCS_URI_FORMAT = "gs://%s/%s";
+
+	private final String bucketName;
+
+	private final String blobName;
+
+	private final URI uri;
+
+	public GoogleStorageLocation(String gcsLocationUriString) {
+		try {
+			URI locationUri = new URI(gcsLocationUriString);
+
+			Assert.isTrue(
+					gcsLocationUriString.startsWith("gs://"),
+					"A Google Storage URI must start with gs://");
+
+			Assert.isTrue(
+					locationUri.getAuthority() != null,
+					"No bucket specified in the location: " + gcsLocationUriString);
+
+			this.bucketName = locationUri.getAuthority();
+			this.blobName = getBlobPathFromUri(locationUri);
+
+			// ensure that if it's a bucket handle, location ends with a '/'
+			if (this.blobName == null && !gcsLocationUriString.endsWith("/")) {
+				locationUri = new URI(gcsLocationUriString + "/");
+			}
+			this.uri = locationUri;
+		}
+		catch (URISyntaxException e) {
+			throw new IllegalArgumentException(
+					"Invalid location: " + gcsLocationUriString);
+		}
+	}
+
+	/**
+	 * Returns the Google Storage bucket name.
+	 */
+	public String getBucketName() {
+		return bucketName;
+	}
+
+	/**
+	 * Returns the path to the blob/folder relative from the bucket root. Returns the empty
+	 * String if the {@link GoogleStorageLocation} specifies a bucket itself.
+	 */
+	public String getBlobName() {
+		return blobName;
+	}
+
+	/**
+	 * Returns the GCS URI of the location.
+	 */
+	public URI uri() {
+		return this.uri;
+	}
+
+	/**
+	 * Returns the Google Storage URI string for the location.
+	 */
+	public String uriString() {
+		String processedBlobName = (blobName == null) ? "" : blobName;
+		return String.format(GCS_URI_FORMAT, bucketName, processedBlobName);
+	}
+
+	/**
+	 * Returns a {@link GoogleStorageLocation} to a bucket.
+	 * @param bucketName name of the GCS bucket
+	 * @return the {@link GoogleStorageLocation} to the location.
+	 */
+	public static GoogleStorageLocation forBucket(String bucketName) {
+		String uri = String.format(GCS_URI_FORMAT, bucketName, "");
+		return new GoogleStorageLocation(uri);
+	}
+
+	/**
+	 * Returns a {@link GoogleStorageLocation} for a file within a bucket.
+	 * @param bucketName name of the GCS bucket
+	 * @param pathToFile path to the file within the bucket
+	 * @return the {@link GoogleStorageLocation} to the location.
+	 */
+	public static GoogleStorageLocation forFile(String bucketName, String pathToFile) {
+		String uri = String.format(GCS_URI_FORMAT, bucketName, pathToFile);
+		return new GoogleStorageLocation(uri);
+	}
+
+	/**
+	 * Returns a {@link GoogleStorageLocation} to a folder whose path is relative to the
+	 * bucket.
+	 * @param bucketName name of the GCS bucket.
+	 * @param pathToFolder path to the folder within the bucket.
+	 * @return the {@link GoogleStorageLocation} to the location.
+	 */
+	public static GoogleStorageLocation forFolder(String bucketName, String pathToFolder) {
+		if (!pathToFolder.endsWith("/")) {
+			pathToFolder += "/";
+		}
+
+		String uri = String.format(GCS_URI_FORMAT, bucketName, pathToFolder);
+		return new GoogleStorageLocation(uri);
+	}
+
+	@Override
+	public String toString() {
+		return uriString();
+	}
+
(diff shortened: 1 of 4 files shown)


spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResource.java
```diff
@@ -315,26 +306,26 @@ public OutputStream getOutputStream() throws IOException {
...
+		return this.location.getBucketName();
 	}
 
 	/**
 	 * Check if this resource references a bucket and not a blob.
 	 * @return if the resource is bucket
 	 */
 	public boolean isBucket() {
-		return this.blobName == null;
+		return this.location.getBlobName() == null;
```
meltsufin: Can you just move the whole method `isBucket` to the location class?

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResource.java
```diff
@@ -48,72 +47,62 @@
  * @author Mike Eltsufin
  * @author Chengyuan Zhao
  * @author João André Martins
+ * @author Daniel Zou
  */
 public class GoogleStorageResource implements WritableResource {
```
meltsufin: Can you just extend `GoogleStorageLocation`?

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageLocation.java
```diff
@@ -0,0 +1,154 @@
...
+
+	/**
+	 * Constructs a {@link GoogleStorageLocation} based on the provided Google Storage URI string.
+	 * The URI string is of the form: {@code gs://<BUCKET_NAME>/<PATH_TO_FILE>}
+	 *
+	 * @param gcsLocationUriString a Google Storage URI string to a bucket/folder/file.
+	 */
+	public GoogleStorageLocation(String gcsLocationUriString) {
+		try {
+			URI locationUri = new URI(gcsLocationUriString);
```
elefeint: Should this be done after the first assert? If `gcsLocationUriString` is sufficiently malformed, this will throw an exception without giving a hint as to what should be passed in.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageLocation.java
```diff
@@ -0,0 +1,154 @@
...
+
+	private final URI uri;
+
+	/**
+	 * Constructs a {@link GoogleStorageLocation} based on the provided Google Storage URI string.
+	 * The URI string is of the form: {@code gs://<BUCKET_NAME>/<PATH_TO_FILE>}
+	 *
+	 * @param gcsLocationUriString a Google Storage URI string to a bucket/folder/file.
+	 */
+	public GoogleStorageLocation(String gcsLocationUriString) {
```
elefeint: If people have trouble forming a GCS URI, I wonder if it would be helpful to have a constructor taking in a bucket name and a path to file separately, so that this class can construct the proper URI itself.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageLocation.java
```diff
@@ -0,0 +1,154 @@
...
+
+	/**
+	 * Returns the Google Storage bucket name.
+	 */
+	public String getBucketName() {
+		return bucketName;
+	}
+
+	/**
+	 * Returns the path to the blob/folder relative from the bucket root. Returns the empty
```
elefeint: It would return `null` and not an empty string, based on `getBlobPathFromUri()` behavior.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageLocation.java
```diff
@@ -0,0 +1,154 @@
...
+	 */
+	public URI uri() {
+		return this.uri;
+	}
+
+	/**
+	 * Returns the Google Storage URI string for the location.
+	 */
+	public String uriString() {
+		String processedBlobName = (blobName == null) ? "" : blobName;
```
elefeint: The logic for returning an empty string for a null blobName seems useful both here and in `getBlobName()`. Perhaps `blobName` could be stored as an empty string instead of null in the first place? Are there any times when a `null` value is in itself useful?

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageLocation.java
```diff
@@ -0,0 +1,154 @@
...
+	}
+
+	/**
+	 * Returns a {@link GoogleStorageLocation} to a bucket.
+	 * @param bucketName name of the GCS bucket
+	 * @return the {@link GoogleStorageLocation} to the location.
+	 */
+	public static GoogleStorageLocation forBucket(String bucketName) {
+		String uri = String.format(GCS_URI_FORMAT, bucketName, "");
+		return new GoogleStorageLocation(uri);
```
elefeint: Having a constructor that just takes a bucketName and a path would come in useful here. Otherwise we end up formatting a GS:// string, and then immediately parsing it.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageLocation.java
```diff
@@ -0,0 +1,154 @@
...
+
+	/**
+	 * Returns a {@link GoogleStorageLocation} for a file within a bucket.
+	 * @param bucketName name of the GCS bucket
+	 * @param pathToFile path to the file within the bucket
+	 * @return the {@link GoogleStorageLocation} to the location.
+	 */
+	public static GoogleStorageLocation forFile(String bucketName, String pathToFile) {
+		String uri = String.format(GCS_URI_FORMAT, bucketName, pathToFile);
+		return new GoogleStorageLocation(uri);
```
elefeint: same idea; `new GoogleStorageLocation(bucketName, pathToFile)` would be nicer.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageLocation.java
dzou: Agreed, I think empty string would be better too. I kept the `blobName` null because the `getBlobName` method in `GoogleStorageResource` returns null; I was worried if I changed it to empty string that I may accidentally break some user who expect it to return null.

@meltsufin - Thoughts on this? Would it be safe to change so that it returns the empty string?

meltsufin: Blob name being null is an indication that it is a bucket resource (think directory). So, I think we should keep it that way.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResource.java
dzou: Done.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResource.java
dzou: Done.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageLocation.java
dzou: Done.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageLocation.java
dzou: Done.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageLocation.java
dzou: Done.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageLocation.java
dzou: OK, will leave as null then. 

Just to clarify, is there a reason why the empty string could not be used to indicate a bucket resource?

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageLocation.java
dzou: Done.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageLocation.java
dzou: Done.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageLocation.java
meltsufin: It probably can, but I'm not sure if an empty string is a valid filename... probably not. Null seems to be more semantically appropriate as the indicator.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageLocation.java
```diff
@@ -0,0 +1,201 @@
...
+	 * @param pathToFile path to the file or folder in the bucket; set as null if
+	 *    the location is to the bucket itself.
+	 */
+	public GoogleStorageLocation(String bucketName, @Nullable String pathToFile) {
+		try {
+			this.bucketName = bucketName;
+			this.blobName = pathToFile;
+
+			if (pathToFile == null) {
+				this.uri = new URI(String.format(GCS_URI_FORMAT, bucketName, ""));
```
meltsufin: I guess you can use the `pathToFile == null ? "" : pathToFile` to save a few lines of code here.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageLocation.java
```diff
@@ -0,0 +1,201 @@
...
+import java.net.URI;
+import java.net.URISyntaxException;
+
+import org.springframework.lang.Nullable;
+import org.springframework.util.Assert;
+
+/**
+ * Represents a Google Cloud Storage location provided by the user.
+ *
+ * @author Daniel Zou
```
meltsufin: `@since 1.2`

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResource.java
```diff
@@ -48,72 +47,69 @@
...
+	 * @param storage the Google Cloud Storage client
+	 * @param bucketName the name of the Google Storage bucket
+	 * @param pathToFile the path to the folder/blob in the bucket; leave as empty string for the
+	 *    path to the bucket itself.
+	 * @param autoCreateFiles determines the auto-creation of the file in Google Cloud Storage
+	 *     if an operation that depends on its existence is triggered (e.g., getting the
+	 *     output stream of a file)
+	 * @throws IllegalArgumentException if the location is an invalid Google Storage location
+	 */
+	public GoogleStorageResource(
```
meltsufin: `@since 1.2`
Also, can you add a test for this new constructor?

spring-cloud-gcp-storage/src/test/java/org/springframework/cloud/gcp/storage/GoogleStorageLocationTests.java
```diff
@@ -0,0 +1,42 @@
...
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.gcp.storage;
+
+import org.junit.Test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class GoogleStorageLocationTests {
```
meltsufin: `@author`

Issue #1475, opened 2019-02-21
Inheritance hierarchy for a single kind


sergeymeshkov:
Is it possible to support inheritance strategy for a single kind which containing entities for two or more different classes in an inheritance hierarchy?

For example, we have types:
```

public class Shape {    
    private String id;
    //other fields
}

public class Circle extends Shape {}

public class Polygon extends Shape{}

```
I would like to get information about all the shapes which was saved.

```
var circe = new Circle();
var polygon = new Polygon();
shapeRepository.save(circle);
shapeRepository.save(polygon);
var shapes = shapeRepository.findAll();

```

Right now I receive a collection with 2 elements of 'Shape' type. Is it possible to add mechanisms, for example like [JPA discriminators](https://docs.oracle.com/javaee/7/api/javax/persistence/DiscriminatorColumn.html), which will make childs (which are in an inheritance hierarchy) into their real types?

ChengyuanZhao:
If my reading of the JPA discriminators feature is correct this means:
- entities of all 3 types (shape, polygon, circle) are stored in the SAME table
- entities of the 3 types are distinguished by a "DTYPE" column that is invisible to the user, but tells Spring Data which java type it should correspond to

I think this would be a pretty deep change to the module that touches many many parts. 

Furthermore, I think it's somewhat contrary to the idea that each java type gets its own table and how each java type gets its own repository. My opinion and recommendation is to not support this feature, and to recommend users to use separate repositories for the specific types they want retrieved. 

WDYT @dmitry-s @meltsufin

meltsufin:
@sergeymeshkov Can you just use a separate entity and repository for each type? You should still be able to map all of them to the same kind.

ChengyuanZhao:
@meltsufin I'm not sure that would work as expected  because you may be retrieving rows for other types and you wouldn't be able to tell. (especially if using the count() or findAllI() methods).

meltsufin:
Right, the burden of ensuring that would be on the user. It's not as good as what JPA provides, but an imperfect workaround.

sergeymeshkov:
[...]

PR #1481, opened 2019-02-22
Datastore persistent entity discrimination field support

ChengyuanZhao:
related #1475
Makes the persistent entity impl aware of discriminator fields and values.


ChengyuanZhao:
@dmitry-s PTAL

PR #1484, opened 2019-02-26
Datastore READ discrimination 

ChengyuanZhao:
related #1475 
Read discrimination support

### DIFF
--- a/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/DefaultDatastoreEntityConverter.java
+++ b/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/DefaultDatastoreEntityConverter.java
@@ -18,6 +18,7 @@ package org.springframework.cloud.gcp.data.datastore.core.convert;
 
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
 
 import com.google.cloud.datastore.BaseEntity;
@@ -135,21 +136,13 @@ public class DefaultDatastoreEntityConverter implements DatastoreEntityConverter
 		if (ostensibleEntity.getDiscriminationFieldName() == null) {
 			return ostensibleEntity;
 		}
-		DatastorePersistentEntity persistentEntity = null;
 
 		Set<Class> members = DatastoreMappingContext.getDiscriminationFamily(ostensibleEntity.getType());
+		Optional<DatastorePersistentEntity> persistentEntity = members == null ? Optional.empty()
+				: members.stream().map(x -> (DatastorePersistentEntity) this.mappingContext.getPersistentEntity(x))
+						.filter(x -> x != null && isDiscriminationFieldMatch(x, propertyValueProvider)).findFirst();
 
-		if (members != null) {
-			for (Class member : members) {
-				DatastorePersistentEntity memberEntity = this.mappingContext.getPersistentEntity(member);
-				if (memberEntity != null && isDiscriminationFieldMatch(memberEntity, propertyValueProvider)) {
-					persistentEntity = memberEntity;
-					break;
-				}
-			}
-		}
-
-		return persistentEntity == null ? ostensibleEntity : persistentEntity;
+		return persistentEntity.orElse(ostensibleEntity);
 	}
 
 	private boolean isDiscriminationFieldMatch(DatastorePersistentEntity entity,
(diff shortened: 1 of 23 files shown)


spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/DefaultDatastoreEntityConverter.java
```diff
@@ -93,27 +94,31 @@ public ReadWriteConversions getConversions() {
...
+		Object instance;
 		try {
 			instance = instantiator.createInstance(persistentEntity, parameterValueProvider);
 			PersistentPropertyAccessor accessor = persistentEntity.getPropertyAccessor(instance);
 			persistentEntity.doWithColumnBackedProperties((datastorePersistentProperty) -> {
 						// if a property is a constructor argument, it was already computed on instantiation
 						if (!persistentEntity.isConstructorArgument(datastorePersistentProperty)) {
-							Object value = propertyValueProvider.getPropertyValue(datastorePersistentProperty);
+					Object value = propertyValueProvider
+							.getPropertyValue((DatastorePersistentProperty) datastorePersistentProperty);
```
dmitry-s: if this cast necessary, can we just add type to lambda parameter?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/DefaultDatastoreEntityConverter.java
```diff
@@ -122,7 +127,39 @@ public ReadWriteConversions getConversions() {
...
+			EntityPropertyValueProvider propertyValueProvider) {
+		if (ostensibleEntity.getDiscriminationFieldName() == null) {
+			return ostensibleEntity;
+		}
+		DatastorePersistentEntity persistentEntity = null;
+
+		Set<Class> members = DatastoreMappingContext.getDiscriminationFamily(ostensibleEntity.getType());
+
+		if (members != null) {
+			for (Class member : members) {
```
dmitry-s: is it possible to replace this loop with a stream?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/DefaultDatastoreEntityConverter.java
```diff
@@ -122,7 +127,39 @@ public ReadWriteConversions getConversions() {
...
+				DatastorePersistentEntity memberEntity = this.mappingContext.getPersistentEntity(member);
+				if (memberEntity != null && isDiscriminationFieldMatch(memberEntity, propertyValueProvider)) {
+					if (persistentEntity != null) {
+						throw new DatastoreDataException(
```
dmitry-s: Can we validate that when DiscriminationFamily map is created? That would be more efficient.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/DefaultDatastoreEntityConverter.java
ChengyuanZhao: Done

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/mapping/DatastoreMappingContext.java
```diff
@@ -55,6 +64,48 @@ public void setApplicationContext(ApplicationContext applicationContext)
 		this.applicationContext = applicationContext;
 	}
 
+	/**
+	 * Registers in the DatastoreMappingContext that two classes are discriminated from the
+	 * same Datastore Kind.
+	 * @param classA one class.
+	 * @param classB another class.
+	 */
+	public static void addDiscriminationClassConnection(Class classA, Class classB) {
```
dmitry-s: we probably want to test this method

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/DefaultDatastoreEntityConverter.java
ChengyuanZhao: I tried but a stream using an anonymous function doesn't work here because we need to set the `persistentEntity` variable inside the loop and anon functions can't set anything outside (it requires variables coming in to be final).

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/mapping/DatastoreMappingContext.java
ChengyuanZhao: This one is tested in the method `testDiscriminationMetadata` in `DatastorePersistentEntityImplTests`. In that method I test that the sets returned as what we expect and they can only be that way because this method is being used.

dmitry-s: I see, makes sense.

PR #1491, opened 2019-02-27
Datastore discrim col list

ChengyuanZhao:
discrimination column is a list due to datastore limitations of not supporting OR

### DIFF
--- a/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/DefaultDatastoreEntityConverter.java
+++ b/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/DefaultDatastoreEntityConverter.java
@@ -147,8 +147,9 @@ public class DefaultDatastoreEntityConverter implements DatastoreEntityConverter
 
 	private boolean isDiscriminationFieldMatch(DatastorePersistentEntity entity,
 			EntityPropertyValueProvider propertyValueProvider) {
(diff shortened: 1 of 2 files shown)


spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/DefaultDatastoreEntityConverter.java
```diff
@@ -147,8 +147,9 @@ private DatastorePersistentEntity getDiscriminationPersistentEntity(DatastorePer
 
 	private boolean isDiscriminationFieldMatch(DatastorePersistentEntity entity,
 			EntityPropertyValueProvider propertyValueProvider) {
-		return propertyValueProvider.getPropertyValue(entity.getDiscriminationFieldName(), EmbeddedType.NOT_EMBEDDED,
-				ClassTypeInformation.from(String.class)).equals(entity.getDiscriminationValue());
+		return ((String[]) propertyValueProvider.getPropertyValue(entity.getDiscriminationFieldName(),
```
dmitry-s: should we check if it's null? what happens if the column is not there?

ChengyuanZhao: This function is only called on persistent entities from the discrimination family, and class types only get added to that family map/tree if they have the appropriate discrimination annotations. Because of that we should only be here if the entity has the column and has the values in the column (or it would have triggered errors elsewhere and earlier)

Issue #1493, opened 2019-02-28
Spring Security with GCP IAP on GAE Flex fails with "Project expected not to be null"


dvdgsng:
Deploying to GAE Flex with spring-cloud-gcp-starter-security-iap fails to automatically determine the GCP project.

> ConfigServletWebServerApplicationContext : Exception encountered during context initialization - cancelling refresh attempt:
> org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'audienceValidator' defined in class path resource [org/springframework/cloud/gcp/autoconfigure/security/IapAuthenticationAutoConfiguration.class]:
>Invocation of init method failed; nested exception is java.lang.IllegalArgumentException:
> **Project expected not to be null.**

Few lines earlier the log states:

> o.s.c.g.a.c.GcpContextAutoConfiguration  : The default project ID is MY_CORRECT_GCP_PROJECT_ID

The configuration is unset, meaning none of properties for `spring.cloud.gcp.security.iap.*` have been specified in the application.properties (registry, algorithm, header, issuer, audience). According to the docs, this should work since the defaults are used and audience is not required in GAE.

Related env vars are set properly by GAE and available to the application: GOOGLE_CLOUD_PROJECT, GCLOUD_PROJECT. Not sure if they are used, though.

Am I missing something?

Do you require more information?

elefeint:
Most likely, the account does not have GCP's `resourcemanager.projects.get` IAM permission, so IAP is unable to retrieve project details.

Could you check what permissions are available to the account being used?

dvdgsng:
Thanks for the quick reply.

I've checked the permissions. We've not configured any additional service accounts, so the default service accounts should be used to run our services.:

* The "App Engine default service account" (...@appspot.gserviceaccount.com) has the "Editor" role assigned, which includes `resourcemanager.projects.get` permissions.
* The "App Engine Flexible Environment Service Agent" (...@gae-api-prod.google.com.iam.gserviceaccount.com) had only "App Engine flexible environment Service Agent" assigned, which is not listed in IAM->Roles so I can not check what permissions it actually has (most likely because it's a Google managed SA). I've added the "Editor" role and redeployed the service, but unfrot. without any change, the error persists.

elefeint:
Could you check that [Cloud Resource Manager API](https://console.developers.google.com/apis/api/cloudresourcemanager.googleapis.com) is enabled?

dvdgsng:
[...]

PR #1494, opened 2019-02-28
Clarify GCP setup requirements for IAP audience autoconfiguration

elefeint:
Automatic audience configuration on AppEngine (both Flex and Standard) require access to resource manager API in GCP.

As described in #1493, not having access (either by API not being enabled or by missing permission) causes application startup failure.

### DIFF
--- a/spring-cloud-gcp-security-iap/src/main/java/org/springframework/cloud/gcp/security/iap/AppEngineAudienceProvider.java
+++ b/spring-cloud-gcp-security-iap/src/main/java/org/springframework/cloud/gcp/security/iap/AppEngineAudienceProvider.java
@@ -48,7 +48,7 @@ public class AppEngineAudienceProvider implements AudienceProvider {
 	@Override
 	public String getAudience() {
 		Project project = this.resourceManager.get(this.projectIdProvider.getProjectId());
-		Assert.notNull(project, "Project expected not to be null.");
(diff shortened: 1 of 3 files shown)


spring-cloud-gcp-security-iap/src/main/java/org/springframework/cloud/gcp/security/iap/AppEngineAudienceProvider.java
```diff
@@ -48,7 +48,7 @@ public AppEngineAudienceProvider(GcpProjectIdProvider projectIdProvider) {
 	@Override
 	public String getAudience() {
 		Project project = this.resourceManager.get(this.projectIdProvider.getProjectId());
-		Assert.notNull(project, "Project expected not to be null.");
+		Assert.notNull(project, "Project expected not to be null. Is Cloud Resource Manager API enabled?");
```
dzou: Maybe link to the view in Cloud Console that lets you enable the API?

elefeint: Added it in code. Messages like these are very common in gcloud CLI, but don't appear in client library code, for example. Still, it will be a common issue, since Resource Manager API is not enabled by default on new projects, so it's worth being more verbose to save people time.

PR #1496, opened 2019-02-28
Datastore discrimination ref docs and sample

### LINKED ISSUE #1475: Inheritance hierarchy for a single kind

ChengyuanZhao:
fixes #1475 
refdoc and sample update for discrimination

### DIFF
--- a/docs/src/main/asciidoc/datastore.adoc
+++ b/docs/src/main/asciidoc/datastore.adoc
@@ -351,6 +351,59 @@ public class ConverterConfiguration {
 ----
 
 
+==== Inheritance Hierarchies
+
+Java entity types related by inheritance can be stored in the same Kind.
+When reading and querying entities using `DatastoreRepository` or `DatastoreTemplate` with a superclass as the type parameter, you can receive instances of subclasses if you annotate the superclass and its subclasses with `DiscriminationField` and `DiscriminationValue`:
+
+[source, java]
+----
+@Entity(name = "pets")
+@DiscriminationField(field = "pet_type")
+@DiscriminationValue("pet")
+abstract class Pet {
+	@Id
+	Long id;
+
+	abstract String speak();
+}
+
+@DiscriminationValue("cat")
+class Cat extends Pet {
+	@Override
+	String speak() {
+		return "meow";
+	}
+}
+
+@DiscriminationValue("dog")
+class Dog extends Pet {
+	@Override
+	String speak() {
+		return "woof";
+	}
+}
+
+@DiscriminationValue("pug")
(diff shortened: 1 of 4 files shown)


docs/src/main/asciidoc/datastore.adoc
```diff
@@ -351,6 +351,59 @@ public class ConverterConfiguration {
...
+@DiscriminationValue("pug")
+class Pug extends Dog {
+	@Override
+	String speak() {
+		return "woof woof";
+	}
+}
+----
+
+Instances of all 4 types are stored in the `pets` Kind.
```
dmitry-s: Pet is an abstract class, so only 3 types are stored in the kind.

docs/src/main/asciidoc/datastore.adoc
```diff
@@ -351,6 +351,59 @@ public class ConverterConfiguration {
...
+Because a single Kind is used, all classes in the hierarchy must share the same ID property and no two instances of any type in the hierarchy can share the same ID value.
+
+Entity rows in Cloud Datastore store their respective types' `DiscriminationValue` in a field specified by the root superclass's `DiscriminationField` (`pet_type` in this case).
+Reads and queries using a given type parameter will match each entity with its specific type.
+For example, reading a `List<Pet>` will produce a list containing instances of all 4 types.
```
dmitry-s: same here

docs/src/main/asciidoc/datastore.adoc
ChengyuanZhao: fixed !

meltsufin:
@ChengyuanZhao @dmitry-s Can we please change `@DiscriminationField` and `@DiscriminationValue` to `@DiscriminatorField` and `@DiscriminatorValue` more in line with how these annotations are called in JPA?

PR #1500, opened 2019-03-01
Suggestion: Make getBlobName return empty string if GoogleStorageResource is a bucket

dzou:
This is a suggestion PR for when `getBlobName()` is called, GoogleStorageResource should return empty string instead of null.

Followup to discussion in #1478 - Elena suggested possibility of returning empty string instead of null.

On further inspection, I think `getBlobName()` should return an empty string instead of null when the resource is a bucket. This is because the empty string is more useful as a value; the main argument for this is usage of expression that pops up: 

```
blobPath = (blobPath == null) ? "" : pathToFile
```

Within storage module this is done twice and this was also done in my work with vision OCR. I begin to fell that it is inconvenient to have to check for null and then convert to empty string.

### DIFF
--- a/spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageLocation.java
+++ b/spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageLocation.java
@@ -79,6 +79,8 @@ public class GoogleStorageLocation {
 	 *    the location is to the bucket itself.
 	 */
 	public GoogleStorageLocation(String bucketName, String pathToFile) {
+		Assert.notNull(pathToFile, "The GCS path to a file/folder must not be null.");
+
 		try {
 			this.bucketName = bucketName;
 			this.blobName = pathToFile;
@@ -184,7 +186,7 @@ public class GoogleStorageLocation {
 
 	private static String getBlobPathFromUri(URI gcsUri) {
 		String uriPath = gcsUri.getPath();
-		if (uriPath.isEmpty() || uriPath.equals("/")) {
+		if (uriPath == null || uriPath.isEmpty() || uriPath.equals("/")) {
 			// This indicates that the path specifies the root of the bucket
 			return "";
 		}
--- a/spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageLocation.java
+++ b/spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageLocation.java
@@ -19,7 +19,6 @@ package org.springframework.cloud.gcp.storage;
 import java.net.URI;
 import java.net.URISyntaxException;
 
-import org.springframework.lang.Nullable;
 import org.springframework.util.Assert;
 
 /**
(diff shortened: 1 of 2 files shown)


spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageLocation.java
```diff
@@ -61,7 +60,7 @@ public GoogleStorageLocation(String gcsLocationUriString) {
 			this.blobName = getBlobPathFromUri(locationUri);
 
 			// ensure that if it's a bucket handle, location ends with a '/'
-			if (this.blobName == null && !gcsLocationUriString.endsWith("/")) {
```
elefeint: Now we probably have to defensively assert that blobby is not null.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageLocation.java
```diff
@@ -76,15 +75,13 @@ public GoogleStorageLocation(String gcsLocationUriString) {
 	 * {@code pathToFile}.
 	 *
 	 * @param bucketName name of the Google Storage bucket
-	 * @param pathToFile path to the file or folder in the bucket; set as null if
+	 * @param pathToFile path to the file or folder in the bucket; set as empty string if
 	 *    the location is to the bucket itself.
 	 */
-	public GoogleStorageLocation(String bucketName, @Nullable String pathToFile) {
+	public GoogleStorageLocation(String bucketName, String pathToFile) {
 		try {
```
elefeint: same, assert not null

meltsufin:
@dzou I think I raised my concern before that something not existing should generally be `null` not empty string. I don't remember what came of that discussion though. Note that if you want to test whether the resource is a bucket (i.e. no blob), there is the `isBucket` method.

elefeint:
Would it be fair to say that while "blob" does not exist, "blobName" is empty?

dzou:
> @dzou I think I raised my concern before that something not existing should generally be `null` not empty string. I don't remember what came of that discussion though. Note that if you want to test whether the resource is a bucket (i.e. no blob), there is the `isBucket` method.

Understood, in this case it feels different because the empty string is a much more useful value. I made this suggestion PR because in my code, I am repeating this line a lot: 

`blobPath = (blobPath == null) ? "" : pathToFile`

i.e. I check whether blobPath is null or not, and if it is, I will set it to empty string. This comes up a lot when using the Storage API ([here](https://github.com/spring-cloud/spring-cloud-gcp/pull/1469/files#diff-b78524a384776d7c9319cf95848bcbd4R165)) or in general when you are formatting a URI string `"gs://" + bucketName + "/" + blobPath`. In such cases one must convert blobPath to a empty string to use it.

If it is the case that representing a bucket's blobPath with `""` or `null` are roughly equal on principle, then I think the empty string wins out because it has the usability advantage.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageLocation.java
dzou: Done.

meltsufin:
Note that a GCS resource can either be a bucket or a blob. Once you know which one it is, you don't need code like `blobPath = (blobPath == null) ? "" : pathToFile`.

Futhemore, if you just need the path to the GCS resource, we already have the `GoogleStorageResource.getURI()` method. Have you tried using it?

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageLocation.java
```diff
@@ -188,9 +186,9 @@ public String toString() {
 
 	private static String getBlobPathFromUri(URI gcsUri) {
```
meltsufin: @dzou Thinking about this a bit more, why wouldn't one just do `gcsUri.toString()`?

PR #1504, opened 2019-03-05
Datastore read event

ChengyuanZhao:
related #1113 
datastore read and query events

### DIFF
--- a/docs/src/main/asciidoc/datastore.adoc
+++ b/docs/src/main/asciidoc/datastore.adoc
@@ -1191,6 +1191,8 @@ Spring Data Cloud Datastore publishes events extending the Spring Framework's `A
 [options="header"]
 |===
 | Type | Description | Contents
+| `AfterKeyReadEvent`|Published immediately after read by-key operations are executed by `DatastoreTemplate`| The entities read from Cloud Datastore and the original keys in the request.
+| `AfterQueryReadEvent`|Published immediately after read byquery operations are executed by `DatastoreTemplate`| The entities read from Cloud Datastore and the original query in the request.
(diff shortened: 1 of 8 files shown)


docs/src/main/asciidoc/datastore.adoc
```diff
@@ -1191,6 +1191,8 @@ Spring Data Cloud Datastore publishes events extending the Spring Framework's `A
 [options="header"]
 |===
 | Type | Description | Contents
+| `AfterKeyReadEvent`|Published immediately after read by-key operations are executed by `DatastoreTemplate`| The entities read from Cloud Datastore and the original keys in the request.
```
meltsufin: The name of the event sounds like you're reading a key.
Maybe `AfterReadByKeyEvent` or `AfterFindByKeyEvent`?

docs/src/main/asciidoc/datastore.adoc
```diff
@@ -1191,6 +1191,8 @@ Spring Data Cloud Datastore publishes events extending the Spring Framework's `A
...
+| `AfterQueryReadEvent`|Published immediately after read byquery operations are executed by `DatastoreTemplate`| The entities read from Cloud Datastore and the original query in the request.
```
meltsufin: Here also, I would consider calling it `AfterReadByQueryEvent` or just simply `AfterQueryEvent`.

ChengyuanZhao: Yea can do

docs/src/main/asciidoc/datastore.adoc
ChengyuanZhao: Can do

PR #1505, opened 2019-03-05
Make Pub/Sub thread pools Spring-managed beans

### LINKED ISSUE #1503: spring-cloud-gcp-starter-pubsub not shutting down
Hi,

I'm experiencing some issues using spring-cloud-gcp-starter-pubsub where the application hangs on shutdown. It sounds the same as this issue https://github.com/spring-cloud/spring-cloud-gcp/issues/60 which was closed when it wasn't able to be reproduced.

I have created a Spring CommandLineRunner Application that will publish messages to PubSub and then shutdown however it does not. I've provided the bare minimum to reproduce here [https://github.com/danneville/spring-gcp-pub-sub-issue](https://github.com/danneville/spring-gcp-pub-sub-issue) . Just the inclusion of the spring-cloud-gcp-starter-pubsub dependency causes the application to hang on shutdown. If you remove the dependency it shuts down immediately as expected.

This happens both when running the application in IntelliJ and on the command line.

I've provided the [thread dump](https://github.com/danneville/spring-gcp-pub-sub-issue/blob/master/thread-dump.txt) in the repo if this helps.

Please can you confirm if this is an issue or if I should be specifying something to gracefully shutdown?

Many thanks

elefeint:
The two thread pools are hardcoded in autoconfiguration, so they fall through the cracks when a Spring application shuts down:
* Spring does not manage them because they are not context-managed beans.
* gRPC stub does not manage them because `FixedExecutorProvider.shouldAutoClose()` is `false` (also, we don't call `SubscriberStub.shutdown()`, which we arguably should do, as well.

This does not manifest on the publisher side because that thread pool is never started, but the subscriber thread pool _is_ started by way of subscriber stub creation when `PubSubSubscriberTemplate` is instantiated.

Fixes #1503.
Addresses the Pub/Sub part of #1506.

### DIFF
--- a/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
+++ b/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
@@ -19,6 +19,7 @@ package org.springframework.cloud.gcp.autoconfigure.pubsub;
 import java.io.IOException;
 import java.util.concurrent.Executor;
 import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
 import java.util.function.Consumer;
 import java.util.function.Function;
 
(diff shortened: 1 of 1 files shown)


spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
```diff
@@ -110,18 +111,30 @@ public GcpPubSubAutoConfiguration(GcpPubSubProperties gcpPubSubProperties,
 		}
 	}
 
+	@Bean
+	@ConditionalOnMissingBean(name = "pubsubPublisherThreadPool")
+	public ScheduledExecutorService pubsubPublisherThreadPool() {
+		return Executors.newScheduledThreadPool(this.gcpPubSubProperties.getSubscriber().getExecutorThreads());
```
artembilan: What is wrong with Spring's `ThreadPoolTaskScheduler` ?

elefeint: Nothing :)

PR #1516, opened 2019-03-11
Fix build after upgrading to SB 2.2

meltsufin:
It appears that the `Envirnment` bean is not available anymore in the context during bootstrap configuration. So, switching `GcpConfigProperties` to `ApplicationContextAware`. Also, fixing the fact that if config module is disabled, we shouldn't even have an instance of `GcpConfigProperties` bean.


spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/config/GcpConfigProperties.java
```diff
@@ -130,4 +116,20 @@ public Credentials getCredentials() {
 		return this.credentials;
 	}
 
+	@Override
+	public void setApplicationContext(ApplicationContext applicationContext) throws BeansException {
+		if (this.profile == null) {
+			String[] profiles = applicationContext.getEnvironment().getActiveProfiles();
```
elefeint: Should we add `Assert.notNull()` on `applicationContext` and` applicationContext.getEnvironment()`, to simplify troubleshooting in the future?

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/config/GcpConfigBootstrapConfigurationTest.java
```diff
@@ -51,15 +54,23 @@ public void testConfigurationValuesAreCorrectlyLoaded() {
...
-					assertThat(config.isEnabled()).isFalse();
+					assertThat(config.isEnabled()).isTrue();
 					assertThat(config.getProjectId()).isEqualTo("pariah");
 				});
 	}
+
+	@Test
+	public void testConfigurationDisabled() {
+		this.contextRunner.run((context) -> {
+					assertThatThrownBy(() -> context.getBean(GcpConfigProperties.class)).isInstanceOf(
```
elefeint: Line wrapping is a bit odd on this.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/config/GcpConfigProperties.java
meltsufin: I believe the contract of `ApplicationContextAware` is that it provides a non-null context, and the context should always have an environment.

meltsufin:
It's odd that it fails on Travis because `mvn clean install' passed for me locally.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/config/GcpConfigProperties.java
artembilan: That's correct.
Although I think we can narrow the responsibility to the `EnvironmentAware`  hook: nor reason to deal with the whole `ApplicationContext` here.

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/config/GcpConfigBootstrapConfigurationTest.java
```diff
@@ -51,15 +54,23 @@ public void testConfigurationValuesAreCorrectlyLoaded() {
...
-					assertThat(config.isEnabled()).isFalse();
+					assertThat(config.isEnabled()).isTrue();
 					assertThat(config.getProjectId()).isEqualTo("pariah");
 				});
 	}
+
+	@Test
+	public void testConfigurationDisabled() {
+		this.contextRunner.run((context) -> {
+			assertThatThrownBy(() -> context.getBean(GcpConfigProperties.class)).isInstanceOf(
```
artembilan: Not related and minor, but we recently opened for ourselves that `assertThatExceptionOfType()` is much handy for this kind of verifications: https://github.com/spring-projects/spring-integration/issues/2760

meltsufin:
Getting a login screen in `GcpStorageAutoConfigurationTests` and just can't figure out which auto configuration to exclude.

meltsufin:
Merging, but adding #1521 to be worked on to unignore the failing test.

PR #1524, opened 2019-03-13
Upgrade to latest versions of dependencies

meltsufin:


spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberTemplate.java
```diff
@@ -111,6 +111,7 @@ public void setMessageConverter(PubSubMessageConverter pubSubMessageConverter) {
 
 	public void setAckExecutor(Executor ackExecutor) {
 		Assert.notNull(ackExecutor, "ackExecutor can't be null.");
+		this.defaultAckExecutor.shutdown();
```
elefeint: it's shut down in destroy().

meltsufin: Why wait until `destroy()`? Once another executor is set, the default one will never be used.

elefeint: This smells like a different PR :)
Make sure it does not get shut down twice, though (no harm, but confusing).

PR #1527, opened 2019-03-14
fix dependency and date parsing

elefeint:
Datastore sample `ApplicationTests` failed because the actuator dependency was no longer brought in transitively (I did not investigate where it used to come from).

Behind that, there was a `DateTime` parsing issue.

### DIFF
--- a/spring-cloud-gcp-starters/spring-cloud-gcp-starter-data-datastore/pom.xml
+++ b/spring-cloud-gcp-starters/spring-cloud-gcp-starter-data-datastore/pom.xml
@@ -26,6 +26,7 @@
 		<dependency>
 			<groupId>org.springframework.boot</groupId>
 			<artifactId>spring-boot-actuator</artifactId>
+			<optional>true</optional>
 		</dependency>
 	</dependencies>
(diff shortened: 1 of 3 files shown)


spring-cloud-gcp-starters/spring-cloud-gcp-starter-data-datastore/pom.xml
```diff
@@ -23,5 +23,9 @@
 			<groupId>org.springframework.cloud</groupId>
 			<artifactId>spring-cloud-gcp-data-datastore</artifactId>
 		</dependency>
+		<dependency>
+			<groupId>org.springframework.boot</groupId>
+			<artifactId>spring-boot-actuator</artifactId>
```
meltsufin: It needs to remain optional and up to the user to decide whether to include.

meltsufin: Why add it here at all? The sample doesn't use the starter doesn't use the actuator. I think this is masking an issue in the autoconfiguration.

Issue #1525, opened 2019-03-13
SpannerTemplate support choosing among multiple connections/instances


ChengyuanZhao:
There are a few steps planned below that go from more-general support to more and more specific features: 

1. `SpannerTemplate` to use a very general so-called `DatabaseClientProvider` bean. The user can implement and provide a bean of this interface in Spring config. This interface would provide various database client / database connections for each request. 
2. provide an  implementation of `DatabaseClientProvider` that reuses all of the other Spanner connection auto-config settings and only varies in the instance-id. 
3. expand the `DatabaseClientProvider` interface to accept `ApplicationEvent` (which we already publish for all operations (if I recall)) so that the instance selection for each request can now see the specifics of the each operation

meltsufin:
(Sorry, closed by mistake.)

During our design discussion, we talked about (3), and I thought the concensus was that we should hold off attempting to address that since there are some issues with the idea.
- What we already implemented in #1530 addresses the original user's request.
- Even if one is able to partition the data on write, it's essentially impossible to do on the read. This point is especially important. I think it makes the whole idea of multiplexing instances based on row content impossible with this approach.

ChengyuanZhao:
@meltsufin  

I don't think the concerns you refer to are actually issues anymore!

The reason is that the user can selectively override the functions he wants. 

For example. 

If he overrides the upsert and read-by-key methods, then he can choose the database client by the entity value. 

However, if he wants to query and submit string queries then he still use the databaseId provider config in the same application! He simply didn't need to override the query method using this method.

BOTH methods can be used at the same time!

meltsufin:
@ChengyuanZhao Can this be closed now, and the PR?

PR #1530, opened 2019-03-14
Spanner expose databaseId provider bean

ChengyuanZhao:
related #1525 
allows bean for Database ID provider to support multiple instances or databases


spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/admin/DatabaseUtilityProvider.java
```diff
@@ -0,0 +1,33 @@
...
+package org.springframework.cloud.gcp.data.spanner.core.admin;
+
+/**
+ * A provider of Cloud Spanner database-related objects.
+ *
+ * @author Chengyuan Zhao
+ *
+ * @since 1.2
+ */
+public interface DatabaseUtilityProvider<T> {
```
meltsufin: This looks identical to `java.util.function.Supplier` except for the name. Why not just use `Supplier` in its place?

docs/src/main/asciidoc/spanner.adoc
```diff
@@ -1343,6 +1343,22 @@ class SpringSecurityAuditorAware implements AuditorAware<User> {
...
 
+=== Multi-Instance Usage
+
+Your application can be configured to use multiple Cloud Spanner instances or databases by providing a custom bean for `DatabaseUtilityProvider<DatabaseId>`.
+The default bean uses the instance ID, database name, and project ID options you configured in `application.properties`.
+
+[source,java]
+----
+    @Bean
+    public DatabaseUtilityProvider<DatabaseId> databaseIdProvider() {
```
meltsufin: I thought we discussed and agreed on using something like `DatabaseClientProvider`. It's more flexible in a sense that it allows the app to connect to Spanner instances even in different projects.
You can always implement a `DatabaseClientProvider` that works with `DatabaseId`.

ChengyuanZhao: @meltsufin , so there is a provider bean for DatabaseClient exposed. However the way DatabaseClients receive their instance/database/project id is via the DatabaseId class. furthermore there is a DatabaseAdminClient that also uses the DatabaseId 
so it made sense to also make a provider bean available for DatabaseId
(emphasizing that DatabaseClient provider bean is also exposed)

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/admin/DatabaseUtilityProvider.java
ChengyuanZhao: Yea I actually prefer using those general "functional" classes. Will change.

docs/src/main/asciidoc/spanner.adoc
ChengyuanZhao: Updated the refdoc to emphasize that a bean for Supplier of DatabaseClient is also available and an option to set. 


Speaking of Suppliers, should we go back and remove CredentialProvider and ProjectIDProvider and just use Suppler of Credentials and Supplier of ProjectID ?

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/admin/CachingDatabaseUtilityProvider.java
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ * A provider of database-related objects that relies on another provider and caches
+ * provided results.
+ *
+ * @param <T> the type of objects this provider produces.
+ * @param <U> the type of objects this provider bases its products on.
+ * @author Chengyuan Zhao
+ */
+public class CachingDatabaseUtilityProvider<T, U> implements Supplier<T> {
```
meltsufin: I think this is too generic. The specificity of the name of the class does not reflect its function. What it does seems very generic and not specific to "Database".

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/admin/SpannerDatabaseAdminTemplate.java
```diff
@@ -54,32 +55,31 @@
...
+	 *     statements.
+	 * @param databaseClientProvider the client used to access schema information tables.
+	 * @param databaseIdProvider the combination of Cloud Spanner Instance Id and Database Id.
+	 *     While databases can be created automatically by this template, instances determine
+	 *     billing and are not created automatically.
 	 */
 	public SpannerDatabaseAdminTemplate(DatabaseAdminClient databaseAdminClient,
-			DatabaseClient databaseClient,
-			DatabaseId databaseId) {
+										Supplier<DatabaseClient> databaseClientProvider,
```
meltsufin: Why would you need this, if you can compute it from the `databaseIdProvider`?

ChengyuanZhao: If the user had chosen to provide his own DatabaseClient supplier then that wouldn't be right (for example if he sets things that aren't in DatabaseId). Also if we just use the DAtabaseIdProvider we would need to create new instances of DatabaseClient in here and also have a Spanner.class on hand.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/admin/CachingDatabaseUtilityProvider.java
ChengyuanZhao: removed a type parameter, PTAL

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/admin/CachingDatabaseClientProvider.java
```diff
@@ -0,0 +1,55 @@
...
+import com.google.cloud.spanner.DatabaseClient;
+
+/**
+ * A provider of database client objects that relies on another provider and caches
+ * provided results.
+ *
+ * @param <U> the type of objects this provider bases its products on.
+ * @author Chengyuan Zhao
+ */
+public class CachingDatabaseClientProvider<U> implements Supplier<DatabaseClient> {
```
meltsufin: I don't think there's much value in this `U` type parameter either.
I would just call this class `DatabaseIdBasedDatabaseClientProvider` and it doesn't need any generics at all. It just needs the `DatabaseIdProvider` and a `Spanner` instance.
Why does this class even need to be public?

ChengyuanZhao: I don't think there's a reason to make it that restrictive. There's no reason `U` can't be the `Spanner` object in another use-case or a just a string specifying the instance ID. 
I don't think we gain any simplification by being more restrictive here, actually.

meltsufin: In that case, you might be better off going back to the most generic design as you had before, and just renaming this class to `CachingSupplier<INPUT, OUTPUT> implements Supplier<OUTPUT>`.

ChengyuanZhao: sounds good! changes made

docs/src/main/asciidoc/spanner.adoc
```diff
@@ -1343,6 +1343,26 @@ class SpringSecurityAuditorAware implements AuditorAware<User> {
...
+[source,java]
+----
+    @Bean
+    public DatabaseIdProvider databaseIdProvider() {
+        // return custom connection options provider
+    }
+----
+
+The `DatabaseId` given by this provider is used as the target database name and instance of each operation Spring Data Cloud Spanner executes.
+By providing a custom implementation of this bean (for example, supplying a thread-local `DatabaseId`), you can direct your application to use multiple instances or databases because.
```
dzou: grammar error - delete "because" or add more to sentence?

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/ReadWriteTransactionSpannerTemplate.java
```diff
@@ -43,12 +44,12 @@
...
-	ReadWriteTransactionSpannerTemplate(DatabaseClient databaseClient,
-			SpannerMappingContext mappingContext,
-			SpannerEntityProcessor spannerEntityProcessor,
-			SpannerMutationFactory spannerMutationFactory,
-			SpannerSchemaUtils spannerSchemaUtils,
-			TransactionContext transactionContext) {
+	ReadWriteTransactionSpannerTemplate(Supplier<DatabaseClient> databaseClient,
+										SpannerMappingContext mappingContext,
+										SpannerEntityProcessor spannerEntityProcessor,
+										SpannerMutationFactory spannerMutationFactory,
```
dzou: revert indentation to previous style? (or was this intended?)

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/admin/CachingComposingSupplier.java
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ * A supplier of objects that relies on another supplier and caches
+ * provided results.
+ *
+ * @param <T> the type of objects this supplier produces.
+ * @param <U> the type of objects this supplier bases its products on.
+ * @author Chengyuan Zhao
+ */
+public class CachingComposingSupplier<T, U> implements Supplier<T> {
```
dzou: Make package-private? Looks like an implementation detail.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/admin/DatabaseIdProvider.java
```diff
@@ -0,0 +1,32 @@
...
+
+/**
+ * A provider of {@code DatabaseId} that is used to determine the target Cloud Spanner
+ * databases on which to act.
+ *
+ * @author Chengyuan Zhao
+ *
+ * @since 1.2
+ */
+public interface DatabaseIdProvider extends Supplier<DatabaseId> {
```
dzou: What was the reason for creating this wrapper around `Supplier<DatabaseId>`? Is it better to just use `Supplier<DatabaseId>` in the code?

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/admin/SpannerDatabaseAdminTemplate.java
```diff
@@ -54,32 +55,31 @@
...
+										Supplier<DatabaseId> databaseIdProvider) {
```
dzou: just in case it wasn't intentional - indentation looks off here too.

docs/src/main/asciidoc/spanner.adoc
ChengyuanZhao: LOL oh that's a typo.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/admin/DatabaseIdProvider.java
ChengyuanZhao: See a comment further up from Mike :P

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/admin/CachingComposingSupplier.java
ChengyuanZhao: This is used in the auto-config module, which is a different package

ChengyuanZhao:
@dmitry-s Done !! PTAL

Issue #1490, opened 2019-02-27
Get Cursor from DatastoreRepository Response


nabilnasri:
Hello, 

I have a custom `Query` on my Repository method. This is what it looks like
```
public interface PostReplyRepository extends DatastoreRepository<PostReply, Long> {
	@Query("SELECT * FROM postReply WHERE __key__ HAS ANCESTOR @parentPostKey")
	List<PostReply> getRepliesForPostWithId(@Param("parentPostKey") Key parentPostKey);
}
```
It works so far, but i would like to have the Cursor included in the Response. Am I using the wrong return value with a `List` or is it not supported?

dmitry-s:
@nabilnasri This feature is currently unsupported, but we might implement it in the future.

nabilnasri:
Thanks for the quick response! This feature would be very useful for us. Do you have any alternative idea on how to accomplish this?

dmitry-s:
Could you provide more details on that? Why do you need the cursor?

Thanks

nabilnasri:
The official Google Docs say that one should avoid using offset with integers.

> Although Datastore mode databases support integer offsets, you should avoid using them. Instead, use cursors. 

https://cloud.google.com/datastore/docs/concepts/queries?hl=en#datastore-datastore-cursor-paging-java

dmitry-s:
Thanks for clarification. I don't think that's achievable in current implementation. But it looks like it shouldn't be hard to implement - most likely with `Page` return type.

nabilnasri:
Sounds good! 
Maybe it is something for another issue, but when using pageable with a method annotated with @Query it throws an exception because the PageRequest can not be converted to a Datastore-compatible type.
I can provide a stacktrace once I am on my notebook again.

dmitry-s:
That's a limitation of `@query` methods. We can't modify the provided query. Page return types work only with methods that generate queries (e.g. `findByColor(..)`)

We'll need to invent something new for `@query` annotated methods.

meltsufin:
@mp911de It seems like there is a lot of interet in this feature. Can you recommend something? Thanks!

mp911de:
The typical scheme for paging is by returning `Page<T>` and accepting `PageRequest` as in:

```java
public interface MyRepository<T, ID>
  extends CrudRepository<T, ID> {

  Page<T> findAll(Pageable pageable);
}
```

Alternatively, use `Stream` for transparent paging:

```java
[...]

PR #1532, opened 2019-03-15
support cursor in pageable

dmitry-s:
Creates classes required for cursor support in annotation-based queries and fixes Pageable and Split methods to use cursor. Also optimizes Pageable queries to only run count query once.
part of #1490

### DIFF
--- a/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreOperations.java
+++ b/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreOperations.java
@@ -235,7 +235,7 @@ public interface DatastoreOperations {
 	 * @param queryOptions the query options
 	 * @return query results, converted to objects of class T
 	 */
-	<T> Iterable<T> queryByExample(Example<T> example, DatastoreQueryOptions queryOptions);
+	<T> DatastoreResultsIterable<T> queryByExample(Example<T> example, DatastoreQueryOptions queryOptions);
 
 	/**
 	 * Run key query by example.
--- a/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreOperations.java
+++ b/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreOperations.java
@@ -16,6 +16,7 @@
 
 package org.springframework.cloud.gcp.data.datastore.core;
 
+import java.util.Collection;
 import java.util.Map;
 import java.util.function.Function;
 
@@ -177,7 +178,7 @@ public interface DatastoreOperations {
 	 * @param <T> the type param of the domain type.
 	 * @return the entities that were found.
 	 */
-	<T> DatastoreResultsCollection<T> findAll(Class<T> entityClass, DatastoreQueryOptions queryOptions);
+	<T> Collection<T> findAll(Class<T> entityClass, DatastoreQueryOptions queryOptions);
 
 	/**
 	 * Check if the given ID belongs to an entity in Cloud Datastore. If this is actually
@@ -234,7 +235,7 @@ public interface DatastoreOperations {
 	 * @param queryOptions the query options
 	 * @return query results, converted to objects of class T
 	 */
-	<T> DatastoreResultsIterable<T> queryByExample(Example<T> example, DatastoreQueryOptions queryOptions);
+	<T> Iterable<T> queryByExample(Example<T> example, DatastoreQueryOptions queryOptions);
 
 	/**
 	 * Run key query by example.
--- a/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreOperations.java
+++ b/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreOperations.java
@@ -16,7 +16,6 @@
 
 package org.springframework.cloud.gcp.data.datastore.core;
 
-import java.util.Collection;
 import java.util.Map;
 import java.util.function.Function;
 
@@ -178,7 +177,7 @@ public interface DatastoreOperations {
 	 * @param <T> the type param of the domain type.
 	 * @return the entities that were found.
 	 */
-	<T> Collection<T> findAll(Class<T> entityClass, DatastoreQueryOptions queryOptions);
+	<T> DatastoreResultsCollection<T> findAll(Class<T> entityClass, DatastoreQueryOptions queryOptions);
 
 	/**
 	 * Check if the given ID belongs to an entity in Cloud Datastore. If this is actually
--- a/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreQueryOptions.java
+++ b/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreQueryOptions.java
@@ -23,11 +23,11 @@ import com.google.cloud.datastore.Cursor;
 import org.springframework.data.domain.Sort;
 
 /**
- * Encapsulates Cloud Memorystore query options.
+ * Encapsulates Cloud Datastore query options.
  *
  * @author Dmitry Solomakha
  */
-public class DatastoreQueryOptions {
+public final class DatastoreQueryOptions {
 
 	private Integer limit;
 
@@ -37,7 +37,7 @@ public class DatastoreQueryOptions {
 
 	private Cursor cursor;
 
-	public DatastoreQueryOptions(Integer limit, Integer offset, Sort sort, Cursor cursor) {
+	private DatastoreQueryOptions(Integer limit, Integer offset, Sort sort, Cursor cursor) {
 		this.limit = limit;
 		this.offset = offset;
 		this.sort = sort;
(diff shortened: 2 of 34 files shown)


meltsufin:
@dmitry-s please reference the issue in the description.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/PartTreeDatastoreQuery.java
```diff
@@ -252,22 +271,31 @@ else if (this.tree.isLimiting()) {
...
 		if (getQueryMethod().getParameters().hasSortParameter()) {
 			sort = sort.and(paramAccessor.getSort());
 		}
 
-		if (paramAccessor.getPageable().isPaged() && !total) {
-			//for slice queries we retrieve one additional item to check if the next slice exists
-			limit = paramAccessor.getPageable().getPageSize() + (isSliceQuery() ? 1 : 0);
-			offset = (int) paramAccessor.getPageable().getOffset();
+		if (pageable.isPaged() && !total) {
+			offset = (int) pageable.getOffset();
```
ChengyuanZhao: Forgive me if I'm confused and have it all wrong here: 

Is the purposde of the cursor so that we can jump efficiently to page 54 of page size 10 of some query's result set right ? Pageable's getOffset method does that multiplication for you so it'll  do 54 * 10 = 540 offset , but if we already have the cursor wouldn't that be doing it twice? 
(I see here we apply both the offset that takes into account the page number as well as applying the cursor which means we will be starting on that correct page already? )

dmitry-s: Not exactly. When we create the actual query we only add offset when cursor is not available.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreOperations.java
```diff
@@ -235,7 +235,7 @@
 	 * @param queryOptions the query options
 	 * @return query results, converted to objects of class T
 	 */
-	<T> Iterable<T> queryByExample(Example<T> example, DatastoreQueryOptions queryOptions);
+	<T> DatastoreResultsIterable<T> queryByExample(Example<T> example, DatastoreQueryOptions queryOptions);
```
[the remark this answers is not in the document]
dmitry-s: I don't think we should do that if it is not required. That would increase the complexity for no good reason.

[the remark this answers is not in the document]
ChengyuanZhao: But right now

`	public Page<T> findAll(Pageable pageable) ` in SimpleDatastoreRepository still returns the non-cursor-aware page, and it depends on DatastoreTemplate's findAll method that doesn't return the cursor-aware pageable.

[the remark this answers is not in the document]
dmitry-s: Oh I see. Sorry, I didn't understand what you meant. It makes sense now.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreResultsIterable.java
```diff
@@ -0,0 +1,47 @@
...
+package org.springframework.cloud.gcp.data.datastore.core;
+
+import java.util.Iterator;
+
+import com.google.cloud.datastore.Cursor;
+
+/**
+ * @author Dmitry Solomakha
+ */
+public class DatastoreResultsIterable<T> implements Iterable<T> {
```
ChengyuanZhao: Since Collection extends Iterable, I think just having DatastoreResultsCollection would be enough

dmitry-s: good catch!

dmitry-s: Actually we don't always want to have a collection and we need to return an iterator with cursor. So I think it make sense to keep both classes to avoid complex logic in case we want only an iterator.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreOperations.java
```diff
@@ -178,7 +177,7 @@
 	 * @param <T> the type param of the domain type.
 	 * @return the entities that were found.
 	 */
-	<T> Collection<T> findAll(Class<T> entityClass, DatastoreQueryOptions queryOptions);
+	<T> DatastoreResultsCollection<T> findAll(Class<T> entityClass, DatastoreQueryOptions queryOptions);
```
meltsufin: Try to keep `Collection` as the return type here.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreQueryOptions.java
```diff
@@ -33,10 +35,13 @@
 
 	private Sort sort;
 
-	public DatastoreQueryOptions(Integer limit, Integer offset, Sort sort) {
+	private Cursor cursor;
+
+	public DatastoreQueryOptions(Integer limit, Integer offset, Sort sort, Cursor cursor) {
```
meltsufin: Explain in javadoc what can and cannot be null.
Generally, nullable params should be set through setters, not constructor.
Consider a builder.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/CursorPageable.java
```diff
@@ -0,0 +1,57 @@
...
+
+import com.google.cloud.datastore.Cursor;
+
+import org.springframework.data.domain.PageRequest;
+import org.springframework.data.domain.Pageable;
+
+/**
+ * @author Dmitry Solomakha
+ */
+public class CursorPageable extends PageRequest {
```
meltsufin: `DatastorePageable`?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/PartTreeDatastoreQuery.java
```diff
@@ -178,47 +197,46 @@ else if (!returnedTypeIsNumber) {
...
-				.convertEntitiesForRead(items.iterator(), this.entityType).stream()
-				.map((o) -> this.processRawObjectForProjection((T) o)).collect(Collectors.toList());
-		return new SliceImpl(entities, pageable, exceedsLimit);
+
+		Pageable pageable = CursorPageable.from(paramAccessor.getPageable(), resultList.getCursor(), null);
+
+		EntityQuery.Builder builderNext = StructuredQuery.newEntityQueryBuilder()
+				.setKind(this.datastorePersistentEntity.kindName());
+		StructuredQuery queryNext = applyQueryBody(parameters, builderNext, false, true, resultList.getCursor());
+		List datastoreResultsList = this.datastoreTemplate.query(queryNext, x -> x);
```
meltsufin: Can be boolean instead of `List`.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreOperations.java
```diff
@@ -235,7 +234,7 @@
 	 * @param queryOptions the query options
 	 * @return query results, converted to objects of class T
 	 */
-	<T> Iterable<T> queryByExample(Example<T> example, DatastoreQueryOptions queryOptions);
+	<T> DatastoreResultsIterable<T> queryByExample(Example<T> example, DatastoreQueryOptions queryOptions);
```
meltsufin: Can you check if using `Collection` is viable here?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/PartTreeDatastoreQuery.java
dmitry-s: has to be a list, because `query` returns a list

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreQueryOptions.java
dmitry-s: added a builder

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/CursorPageable.java
dmitry-s: renamed

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreQueryOptions.java
```diff
@@ -79,4 +79,36 @@ public boolean equals(Object o) {
...
+			this.sort = sort;
+			return this;
+		}
+
+		public Builder setCursor(Cursor cursor) {
+			this.cursor = cursor;
+			return this;
+		}
+
+		public DatastoreQueryOptions createDatastoreQueryOptions() {
```
meltsufin: I usually see the method called `build()`.

PR #1535, opened 2019-03-15
Document Firestore in Datastore Mode support

### LINKED ISSUE #1508: Update docs to mention that we support Firestore in Datastore mode
Existing projects that use Datastore are starting to be automatically [upgraded](https://cloud.google.com/datastore/docs/upgrade-to-firestore) to Firestore in Datastore mode. We should update our documentation to mention that Spring Data Datastore supports Firestore in Datastore mode.

Related to question in #1507.

dmitry-s:
Fixes Datastore docs to mention Firestore
fixes #1508

### DIFF
--- a/docs/src/main/asciidoc/datastore.adoc
+++ b/docs/src/main/asciidoc/datastore.adoc
@@ -2,6 +2,8 @@
 
 == Spring Data Cloud Datastore
 
+NOTE: This integration is fully compatible with https://cloud.google.com/datastore/docs/[Firestore in Datastore Mode], but not with Firestore in Native Mode.
+
 http://projects.spring.io/spring-data/[Spring Data] is an abstraction for storing and retrieving POJOs in numerous storage technologies.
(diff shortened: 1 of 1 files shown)


docs/src/main/asciidoc/datastore.adoc
```diff
@@ -3,7 +3,7 @@
 == Spring Data Cloud Datastore
 
 http://projects.spring.io/spring-data/[Spring Data] is an abstraction for storing and retrieving POJOs in numerous storage technologies.
-Spring Cloud GCP adds Spring Data support for http://cloud.google.com/datastore/[Google Cloud Datastore].
+Spring Cloud GCP adds Spring Data support for https://cloud.google.com/firestore/[Google Cloud Firestore] in Datastore mode.
```
meltsufin: I don't think this is enough and people might miss it.
I think we need a note box right after the title, explaining that the integration is fully compatible with Firestore in Datastore Mode, but not with Firestore in Native Model.

Issue #1521, opened 2019-03-12
Re-enable GcpStorageAutoConfigurationTests and StackdriverLoggingIntegrationTests after fixing or refactoring them


meltsufin:
After upgrading to Spring Boot 2.2 the `GcpStorageAutoConfigurationTests` and `StackdriverLoggingIntegrationTests` tests fails because of the login screen put up in front of the REST controller. Ideally, the fix would allow us to remove all of the auto-configuration exclusions.

dzou:
What are your thoughts on just deleting the tests? Since we have integration coverage from integration tests in their modules as well as from samples, I think they may be redundant.

meltsufin:
The advantage of these is that they run as unit tests with each PR, rather than integration tests that only run post-merge on master.

dzou:
> The advantage of these is that they run as unit tests with each PR, rather than integration tests that only run post-merge on master.

Hmm, I don't agree that this is an advantage. The two tests listed are integration tests, and we could just as easily remove the Dit.storage/Dit.logging requirement on the integration tests in storage/logging if we wanted to do that.

Also the [Logging tests are not being run as a unit test](https://github.com/spring-cloud/spring-cloud-gcp/blob/master/spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/logging/it/StackdriverLoggingIntegrationTests.java#L88).

meltsufin:
What I meant to say is that they run locally without invoking GCP. For integration tests that are guarded with flags, we can only run them on master, (1) because they're slow, and (2) because we don't have the encrypted credentials in the case of external pull requests.

dzou:
I see, in that case the storage test is fine, but `StackdriverLoggingIntegrationTests` definitely invokes GCP.

dzou:
Actually I think I understand the full story now. Looks like the logging IT was put here before you split out a logging module. Perhaps the logical solution is to just move that test to the newly created logging project which does not yet have an integration test.

dzou:
Turns out this is resolved.

GcpStorageAutoConfigurationTests fixed by Elena in #1638.
Logging integration tests consolidated to the trace sample sample test in #1676.

PR #1547, opened 2019-03-18
Reenable Autoconfiguration Logging Tests

dzou:
Reenables the Logging autoconfiguration tests referenced in #1521.

This PR moves the logging tests to the logging module.


spring-cloud-gcp-logging/src/test/java/org/springframework/cloud/gcp/logging/it/StackdriverLoggingIntegrationTests.java
```diff
@@ -96,37 +97,32 @@ public void test() throws IOException {
...
 	 */
 	@RestController
-	@SpringBootApplication(exclude = {
-			GcpCloudSqlAutoConfiguration.class,
-			GcpStorageAutoConfiguration.class,
-			DataSourceAutoConfiguration.class,
-			SecurityAutoConfiguration.class,
-			IapAuthenticationAutoConfiguration.class
-	})
+	@SpringBootApplication
```
meltsufin: Since you're no longer using autoconfiguration, you'll probably need to manually configure the `TraceIdLoggingWebMvcInterceptor` bean. and the `TraceIdExtractor`.

dzou:
Putting on hold for now - I was having trouble reenabling these tests because it turns out the tests use trace + logging. Once #1608 is resolved, these can be reenabled.

meltsufin:
@dzou I think you can revisit this following the #1672 merge.

dzou:
> @dzou I think you can revisit this following the #1672 merge.

Sounds good, will do.

meltsufin:
Looks like there is still a problem:
```
[ERROR] test(org.springframework.cloud.gcp.logging.it.StackdriverLoggingIntegrationTests)  Time elapsed: 23.04 s  <<< ERROR!
com.google.cloud.logging.LoggingException: io.grpc.StatusRuntimeException: UNAVAILABLE: Credentials failed to obtain metadata
Caused by: com.google.api.gax.rpc.UnavailableException: io.grpc.StatusRuntimeException: UNAVAILABLE: Credentials failed to obtain metadata
Caused by: io.grpc.StatusRuntimeException: UNAVAILABLE: Credentials failed to obtain metadata
Caused by: java.io.IOException: Error code 404 trying to get security access token from Compute Engine metadata for the default service account. This may be because the virtual machine instance does not have permission scopes specified. It is possible to skip checking for Compute Engine metadata by specifying the environment  variable NO_GCE_CHECK=true.

```

dzou:
Will close this PR and propose an alternate approach.

PR #1553, opened 2019-03-20
Spanner events instance selection

### LINKED ISSUE #1525: SpannerTemplate support choosing among multiple connections/instances

ChengyuanZhao:
fixes #1525 
Adds support for SpannerTemplate parameter-based instance selection

@meltsufin I added a refdoc section and a sample integration test as an example of how this should be used. 

I decided against using the events as a way to pass call and operation information as the way to let the user determine databases/instances because I think that's an arbitrarily decided set of params and it's not always available. It also places a HUGE burden on the user to examine that event to even see exactly what was happening. That would mean lots of reflecting and `if-else` `instanceof` statements for the user and TONS of casting. 

Instead I think it's more natural to have the user provide a small stub method for any of the `SpannerTemplate` methods they would use. This way they have access to exactly the params they would have using that specific `SpannerTemplate` method. Then the small stub method they provide that determines the database client is used together with the actual implementation of that template method.


meltsufin:
Can you provide a bit more description? What is it trying to achieve and how? Thanks!

ChengyuanZhao:
@meltsufin yes I remember those issues specifically regarding queries since they don't have any row or key content. I explain what I see happening there in #1525 and also I show a mixed usage of both methods in the sample integration test.

PR #1554, opened 2019-03-21
Add the Vision OCR Sample

dzou:
Add the Vision OCR Sample demonstrating how to use `DocumentOcrTemplate`.

Intentionally omitted integration tests for this because they are gonna be hard to write given how this app works. So I would rather do this at a later point in time - Tracking #1555.


spring-cloud-gcp-samples/spring-cloud-gcp-vision-ocr-sample/README.adoc
```diff
@@ -0,0 +1,53 @@
...
+
+b. Go to http://localhost:8080 in your browser, and specify the URL of a PDF or TIFF document to process.
++
+The application allows you to specify documents to upload in two ways:
+
+* From a web URL.
++
+Example: `http://www.google.com/document.pdf`
+
+* From the `resources/` directory.
```
dmitry-s: I think it would be easier to use if users could just upload files through web page instead.

dzou:
> Overall, it seems very complicated for a sample. I think it would be better to limit functionality, for example to only search the pdf file in specific location and make the code simpler, maybe without rest controller.
> Also, would be good to have CI tests.

This is more complicated than other samples because the intention was to provide an "advertisement" for the Document OCR feature. We are walking into this with mindset that we want to show off something cool that would interest users to use this feature. Moreover, overall complexity of the sample is not very high - user just needs to create a bucket to use the sample, and the amount of Java code is low.

dmitry-s:
In that case should it be called a "demo" instead of "sample"? IMHO a sample should be short and only show relevant details.

dzou:
> In that case should it be called a "demo" instead of "sample"? IMHO a sample should be short and only show relevant details.

Sure, would be open to that - though would it be unusual to name this as "demo" when all the other apps in the directory are named "sample"?

spring-cloud-gcp-samples/spring-cloud-gcp-vision-ocr-sample/README.adoc
```diff
@@ -0,0 +1,53 @@
...
+
+== Running the sample
+
+This section describes how to run the sample and use it to extract text from some documents.
+
+=== 1. Setting up your GCP Project
+
+a. Create a Google Cloud Platform project with https://cloud.google.com/billing/docs/how-to/modify-project#enable-billing[billing enabled], if you don't have one already.
+
+b. Enable the Google Cloud Vision API from the "APIs & Services" menu of the Google Cloud Console.
```
meltsufin: I believe you can also enable it via a gcloud command. Can you provide it here as well? Thanks!

spring-cloud-gcp-samples/spring-cloud-gcp-vision-ocr-sample/README.adoc
meltsufin: @dzou Since you're including a sample pdf there, it should be OK. You can probably even prefill the form with that value in the UI. In any case, you should, however, suggest specifying a `file://` or `gs://` path to a PDF doc.

spring-cloud-gcp-samples/spring-cloud-gcp-vision-ocr-sample/pom.xml
```diff
@@ -0,0 +1,47 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
```
meltsufin: I think we're done with `http:`. I has to be `https:` across the board now. 😄

spring-cloud-gcp-samples/spring-cloud-gcp-vision-ocr-sample/src/main/resources/templates/index.html
```diff
@@ -0,0 +1,54 @@
...
+  <p>
+    All operations will be done your Google Storage Bucket specified in application.properties:
+    <a th:href="@{'https://console.cloud.google.com/storage/browser/' + ${ocrBucket}}">
+      <b>[[${ocrBucket}]]</b>
+    </a>
+  </p>
+
+  <form action="/submitDocument" method="post">
+    <b>Specify a document URL to upload and analyze:</b>
+    <input type="text" name="documentUrl"/>
```
meltsufin:
```suggestion
    <input type="text" name="documentUrl" value="classpath:test_pdfs/test.pdf"/>
```

spring-cloud-gcp-samples/spring-cloud-gcp-vision-ocr-sample/src/main/resources/templates/index.html
```diff
@@ -0,0 +1,54 @@
...
+
+  <p>
+    All operations will be done your Google Storage Bucket specified in application.properties:
+    <a th:href="@{'https://console.cloud.google.com/storage/browser/' + ${ocrBucket}}">
+      <b>[[${ocrBucket}]]</b>
+    </a>
+  </p>
+
+  <form action="/submitDocument" method="post">
+    <b>Specify a document URL to upload and analyze:</b>
```
meltsufin:
```suggestion
    <b>Specify a document URL to upload and analyze:</b><br/>
```

spring-cloud-gcp-samples/spring-cloud-gcp-vision-ocr-sample/src/main/resources/templates/viewDocument.html
```diff
@@ -0,0 +1,24 @@
...
+
+<p>
+  <a href="/status">Back to Status Page</a>
+</p>
+
+<p>
+  First 50 words of on page [[${pageNumber}]]:
+</p>
+
+<p>
```
meltsufin: Maybe specify a different font here to distinguish the text from the PDF.

spring-cloud-gcp-samples/spring-cloud-gcp-vision-ocr-sample/src/main/java/com/example/WebController.java
```diff
@@ -0,0 +1,142 @@
...
+	}
+
+	private static String getFileType(Resource documentResource) {
+		int extensionIdx = documentResource.getFilename().lastIndexOf(".");
+		String fileType = documentResource.getFilename().substring(extensionIdx);
+
+		switch (fileType) {
+			case ".tif":
+				return "image/tiff";
+			default:
```
elefeint: what if it's some random file extension? Will it default to PDF?

spring-cloud-gcp-samples/spring-cloud-gcp-vision-ocr-sample/pom.xml
dzou: Hmm, this one didn't quite work for me; I got a "URI not registered error" when trying https. I noticed that all the maven urls in the repo are still `http://`.

dzou:
> Perhaps it's time for a top-level "demos" folder, with a blog post on each?

Yeah I'd be open to that; would also provide the distinction in complexity of samples vs. demo that @dmitry-s  mentioned.

@meltsufin - Would you be open to a new top-level demos directory?

meltsufin:
> > Perhaps it's time for a top-level "demos" folder, with a blog post on each?
> 
> Yeah I'd be open to that; would also provide the distinction in complexity of samples vs. demo that @dmitry-s mentioned.
> 
> @meltsufin - Would you be open to a new top-level demos directory?

Sure, if everyone feels strongly about the distinction between samples and demos. I personally do not see the big deal. Every sample is a demo in some sense and vice versa. Currently, we point everyone to the samples directory everywhere. So, introducing another director might make it a bit confusing.

dzou:
> > > Perhaps it's time for a top-level "demos" folder, with a blog post on each?
> > 
> > 
> > Yeah I'd be open to that; would also provide the distinction in complexity of samples vs. demo that @dmitry-s mentioned.
> > @meltsufin - Would you be open to a new top-level demos directory?
> 
> Sure, if everyone feels strongly about the distinction between samples and demos. I personally do not see the big deal. Every sample is a demo in some sense and vice versa. Currently, we point everyone to the samples directory everywhere. So, introducing another director might make it a bit confusing.

Ok fair enough, yeah in the interest of simplicity lets just keep it all in one directory and I will rename this sample to demo. Maybe if one day we have many demos and many samples that overlap we can split them.

spring-cloud-gcp-samples/spring-cloud-gcp-vision-ocr-sample/src/main/java/com/example/WebController.java
dzou: Done. I made it throw an exception now if a user enters a file type other than PDF or TIFF.

dzou:
Hi, I have addressed your comments and also added CSS to make the app look a little better. PTAL.

spring-cloud-gcp-samples/spring-cloud-gcp-vision-ocr-demo/src/main/resources/templates/viewDocument.html
```diff
@@ -0,0 +1,25 @@
...
+  <title>View Document Text</title>
+  <meta http-equiv="Content-Type" content="text/html; charset=UTF-8"/>
+  <link type="text/css" href="css/style.css" rel="stylesheet"/>
+</head>
+
+<body>
+<h1>Text from [[${gcsDocumentUrl}]]</h1>
+
+<p>
+  <a href="/status">Back to Status Page</a>
```
meltsufin: Maybe you can make this page auto-refresh every 30 seconds or so, or tell the user to refresh manually.

spring-cloud-gcp-samples/spring-cloud-gcp-vision-ocr-demo/README.adoc
```diff
@@ -0,0 +1,107 @@
...
+  // Uploads the document to the GCS bucket
+  Resource documentResource = resourceLoader.getResource(documentUrl);
+  BlobId outputBlobId = BlobId.of(ocrBucket, documentResource.getFilename());
+  BlobInfo blobInfo =
+      BlobInfo.newBuilder(outputBlobId)
+          .setContentType(getFileType(documentResource))
+          .build();
+
+  try (WriteChannel writer = storage.writer(blobInfo)) {
+    ByteStreams.copy(documentResource.getInputStream(), Channels.newOutputStream(writer));
```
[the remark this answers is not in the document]
dzou: Yeah that does work; however currently our `GoogleStorageResource` class doesn't yet let you control the type of blob you create; like this:

```
BlobInfo.newBuilder(outputBlobId)
          .setContentType(getFileType(documentResource))
          .build();
```

Currently all files created through GoogleStorageResource defaults to type: "application/octet-stream". Unfortunately this information is needed b/c you otherwise get this error when submitting the document to vision API: 

```
INVALID_ARGUMENT: At this time, the only MIME types supported are 'application/pdf', 'image/gif' and 'image/tiff'.
```

spring-cloud-gcp-samples/spring-cloud-gcp-vision-ocr-demo/README.adoc
```diff
@@ -0,0 +1,107 @@
+= Spring GCP Cloud Vision OCR Demo Application
+
+This demonstrates how to run Google Cloud Vision https://cloud.google.com/vision/docs/ocr[Optical Character Recognition (OCR)] analysis on PDF and TIFF documents.
+OCR is a process which allows you to detect and extract the text from digital documents.
+
+This demo application allows you to upload documents to a specified https://cloud.google.com/storage/docs/key-terms#buckets[Google Cloud Storage bucket] and run OCR on those documents you uploaded.
+After your documents are uploaded and processed, the application will allow you to access the extracted text from each document.
+
+The main work of running OCR in the application occurs in the `submitDocument` method in `WebController.java`:
```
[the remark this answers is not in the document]
dzou: Done.

spring-cloud-gcp-samples/spring-cloud-gcp-vision-ocr-demo/README.adoc
```diff
@@ -0,0 +1,107 @@
...
++
+Example: `gs://your-gcs-bucket/test.pdf`
+
+* From the `resources/` directory.
+You may specify any document under the `resources/` subdirectory using the `classpath:` protocol.
+Specify its path relative to the `resources/` directory.
++
+Example: `classpath:test_pdfs/test.pdf` references the sample pdf provided at `resources/test_pdfs/test.pdf`.
+
+c. When you submit the document through the application, it will be uploaded to your Google Storage bucket, and OCR analysis will be run on it.
```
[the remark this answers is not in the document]
dzou: Done. Added to readme and also I throw an error from the app if their bucket does not already exist.

spring-cloud-gcp-samples/spring-cloud-gcp-vision-ocr-demo/src/main/resources/templates/viewDocument.html
dzou: Done, added a note for user to refresh manually.

spring-cloud-gcp-samples/spring-cloud-gcp-vision-ocr-demo/README.adoc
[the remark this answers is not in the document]
meltsufin: Can you file a bug regarding this so we can fix it going forward? Thanks!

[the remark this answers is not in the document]
dzou: Done: https://github.com/spring-cloud/spring-cloud-gcp/issues/1603

PR #1480, opened 2019-02-22
support non-collection return types

dmitry-s:
fixes #1471 
fixes #1350

PR #1575, opened 2019-03-25
Fix ci tests in spring-cloud-gcp-data-datastore

Bobsel:
* Fix CI-Test from changes of #1480


dmitry-s:
Thanks for an attempt to fix the issue, but the test is actually correct. The method should return `null` if there are no results if it is annotated as `@Nullable`.

meltsufin:
@Bobsel We are looking into a fix that maintains the test assertion. Thanks for the PR though.

PR #1607, opened 2019-03-27
Update dependencies for 1.1.x branch

meltsufin:


spring-cloud-gcp-dependencies/pom.xml
```diff
@@ -16,9 +16,9 @@
 	<packaging>pom</packaging>
 
 	<properties>
-		<google-cloud-bom.version>0.77.0-alpha</google-cloud-bom.version>
-		<google-auth-library-oauth2-http.version>0.12.0</google-auth-library-oauth2-http.version>
-		<mysql-socket-factory.version>1.0.11</mysql-socket-factory.version>
+		<google-cloud-bom.version>0.84.0-alpha</google-cloud-bom.version>
```
meltsufin: This breaks the trace sample. #1608

elefeint:
If we are on 2.1.5.BUILD-SNAPSHOT with rule suppression, I can roll back #1645; it should not be necessary anymore.

Issue #1310, opened 2018-12-11
Ability to provide additional fields to append to json format


mradzevicius:
# Description

Ability to log additional context variables provides ability to search logs sent to Stackdriver without using MDC context to do that. Something similar to `StructuredArguments` in `logstash-logback-encoder`.

ie:
```java
logger.info("job {} has failed", kv("job_id", job.getId(), kv("retry_count", job.getRetryCount()))
```

after logging fields would be appended to the json object which is sent to stdout: 

```json
{
  "job_id": "xxxx-yyy-zzzz",
  "retry_count": 3
}
```

Then those fields would be logged into stackdriver when using Logging view then we could filter out by additional parameters either `job_id` or `retry_count`. 

Maybe it is already possible and I just did not find a way how to do this?

meltsufin:
Have you looked at the `includeMDC` [property](https://docs.spring.io/spring-cloud-gcp/docs/1.1.0.BUILD-SNAPSHOT/reference/htmlsingle/#_log_via_console)? Does it address your requirements?

mradzevicius:
@meltsufin yes i've looked at it, unfortunately not in this case - it has its place for usage.

Also a bonus, would be ability to configure `serviceContext` property or an ability to add static attributes (that might be possible via out of the box logback):

```json
{
"serviceContext": {"service": "app-name", "version": "app-version" }
}
```

That would also benefit Stackdriver error reporting when exceptions are thrown and catched also filtering by app/version when you have multiple running. By specifying `serviceContext`, additional information appears in Error Reporting window in GCP with those values which are also useful, otherwise exception appears in GKE case just as a general "seen in gke-instances" versus "seen in app-name:app-version". 

I've managed to achieve the above logging and the serviceContext concerns with using `logstash-logback-encoder` using stackdriver logging output as a reference to craft it manually. ie:

logback config:
```
...
<appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
            <encoder class="net.logstash.logback.encoder.LoggingEventCompositeJsonEncoder">
                <providers>
                    <mdc/>
                    <provider class="logging.Sleuth"></provider>
                    <logstashMarkers/>
                    <arguments/>
                    <pattern>
                        <pattern>
                            {
                            "serviceContext": {"service": "${appArtifactId}", "version": "${appVersion}" },
                            "severity": "%level",
                            "thread": "%thread",
                            "logger": "%logger{36}.%M",
[...]

PR #1611, opened 2019-03-28
ability to provide additional fields to append to json format #1310

Bobsel:
We have the same issue as described in #1310. We need the option to configure serviceContext so we can filter error reports for service-name and version. Additionally it would be helpful to add custom json data to the log if needed.

### DIFF
--- a/docs/src/main/asciidoc/logging.adoc
+++ b/docs/src/main/asciidoc/logging.adoc
@@ -125,6 +125,8 @@ This setting is only evaluated if `includeFormattedMessage` is `true`
 | `includeContextName` | `true` | Should the logging context be included
 | `includeMessage` | `false` | Should the log message with blank placeholders be included
 | `includeException` | `false` | Should the stacktrace be included as a own field
+| `serviceContext` | none | Define the service context data.
+| `customJson` | none | Defines custom json data. Data will be added to the json output.
 |=======================================================================
 
 This is an example of such an Logback configuration:
@@ -150,6 +152,11 @@ This is an example of such an Logback configuration:
         <!--<includeContextName>true</includeContextName>-->
         <!--<includeMessage>false</includeMessage>-->
         <!--<includeException>false</includeException>-->
+        <!--<serviceContext>
+              <service>service-name</service>
+              <version>service-version</version>
+            </serviceContext>-->
+        <!--<customJson>{"custom-key": "custom-value"}</customJson>-->
       </layout>
     </encoder>
   </appender>
--- a/spring-cloud-gcp-logging/src/main/java/org/springframework/cloud/gcp/logging/StackdriverJsonLayout.java
+++ b/spring-cloud-gcp-logging/src/main/java/org/springframework/cloud/gcp/logging/StackdriverJsonLayout.java
@@ -40,6 +40,7 @@ import org.springframework.util.StringUtils;
  *
  * @author Andreas Berger
  * @author Chengyuan Zhao
+ * @author Stefan Dieringer
  */
 public class StackdriverJsonLayout extends JsonLayout {
 
@@ -55,6 +56,10 @@ public class StackdriverJsonLayout extends JsonLayout {
 	private boolean includeSpanId;
 
 	private boolean includeExceptionInMessage;
+
+	private StackdriverServiceContext serviceContext;
+
+	private Map<String, Object> customJson;
 
 	/**
 	 * creates a layout for a Logback appender compatible to the Stackdriver log format.
@@ -133,6 +138,24 @@ public class StackdriverJsonLayout extends JsonLayout {
 		this.includeExceptionInMessage = includeExceptionInMessage;
 	}
 
+	/**
+	 * set the service context for stackdriver.
+	 * @param serviceContext the service context
+	 */
+	public void setServiceContext(StackdriverServiceContext serviceContext) {
+		this.serviceContext = serviceContext;
+	}
+
+	/**
+	 * set custom json data to include in log output.
+	 * @param json json string
+	 */
+	public void setCustomJson(String json) {
+		Gson gson = new Gson();
+		this.customJson = gson.fromJson(json, Map.class);
+	}
+
+
 	@Override
 	public void start() {
 		super.start();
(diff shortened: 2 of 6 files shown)


Bobsel:
i'm not sure if you would accept this feature also in 1.1.x. if you agree, i can open another pr for 1.1.x. thanks

meltsufin:
@Bobsel Thanks for the PR submission. We can decide whether to add to 1.1.x after it's merged into master. Although I think it would be unlikely since it's not really a bug fix.

docs/src/main/asciidoc/logging.adoc
```diff
@@ -125,6 +125,8 @@ This setting is only evaluated if `includeFormattedMessage` is `true`
 | `includeContextName` | `true` | Should the logging context be included
 | `includeMessage` | `false` | Should the log message with blank placeholders be included
 | `includeException` | `false` | Should the stacktrace be included as a own field
+| `serviceContext` | none | Define the service context data.
```
meltsufin: Clarify that it's used for Stackdriver Error Reporting.

spring-cloud-gcp-logging/src/main/java/org/springframework/cloud/gcp/logging/StackdriverServiceContext.java
```diff
@@ -0,0 +1,56 @@
...
+ *
+ * Reference:  https://cloud.google.com/error-reporting/reference/rest/v1beta1/ServiceContext
+ *
+ * @author Stefan Dieringer
+ */
+public class StackdriverServiceContext {
+
+	private String service;
+
+	private String version;
```
meltsufin: What about `resourceType`?

spring-cloud-gcp-logging/src/main/java/org/springframework/cloud/gcp/logging/StackdriverServiceContext.java
```diff
@@ -0,0 +1,56 @@
...
+package org.springframework.cloud.gcp.logging;
+
+/**
+ * This class provides the service context data for stackdriver error reporting.
+ *
+ * Reference:  https://cloud.google.com/error-reporting/reference/rest/v1beta1/ServiceContext
+ *
+ * @author Stefan Dieringer
+ */
+public class StackdriverServiceContext {
```
meltsufin: `StackdriverErrorReportingServiceContext`?

docs/src/main/asciidoc/logging.adoc
```diff
@@ -125,6 +125,8 @@ This setting is only evaluated if `includeFormattedMessage` is `true`
...
+| `customJson` | none | Defines custom json data. Data will be added to the json output.
```
meltsufin: Can you explain the use-case for this? You already have the MDC to provide attributes.

spring-cloud-gcp-logging/src/main/java/org/springframework/cloud/gcp/logging/StackdriverServiceContext.java
Bobsel: As i understand from the documentation, the ressourceType is set automatically and must not be set by error reporting. https://cloud.google.com/error-reporting/reference/rest/v1beta1/ServiceContext.

meltsufin: Ok, fair enough.

docs/src/main/asciidoc/logging.adoc
Bobsel: I think the option to put some custom static data is helpful for some custom use cases. We use this for example  to add some environment variables from the docker container to the json payload. So we can search and filter for these metadata. To do this with MDC has disadvantages: you have to write code to just add some static json payload. With MDC you can only add String / String payload. With this new option, you only need to add the json to the logback configuration.

spring-cloud-gcp-logging/src/main/java/org/springframework/cloud/gcp/logging/StackdriverErrorReportingServiceContext.java
```diff
@@ -0,0 +1,56 @@
...
+ */
+
+package org.springframework.cloud.gcp.logging;
+
+/**
+ * This class provides the service context data for stackdriver error reporting.
+ *
+ * Reference:  https://cloud.google.com/error-reporting/reference/rest/v1beta1/ServiceContext
+ *
+ * @author Stefan Dieringer
```
meltsufin:
```suggestion
 * @author Stefan Dieringer
 * @since 1.2
```

spring-cloud-gcp-logging/src/main/java/org/springframework/cloud/gcp/logging/StackdriverJsonLayout.java
```diff
@@ -133,6 +138,24 @@ public void setIncludeExceptionInMessage(boolean includeExceptionInMessage) {
 		this.includeExceptionInMessage = includeExceptionInMessage;
 	}
 
+	/**
+	 * set the service context for stackdriver.
+	 * @param serviceContext the service context
```
meltsufin:
```suggestion
	 * @param serviceContext the service context
	 * @since 1.2
```

spring-cloud-gcp-logging/src/main/java/org/springframework/cloud/gcp/logging/StackdriverJsonLayout.java
```diff
@@ -133,6 +138,24 @@ public void setIncludeExceptionInMessage(boolean includeExceptionInMessage) {
...
+	 */
+	public void setServiceContext(StackdriverErrorReportingServiceContext serviceContext) {
+		this.serviceContext = serviceContext;
+	}
+
+	/**
+	 * set custom json data to include in log output.
+	 * @param json json string
```
meltsufin:
```suggestion
	 * @param json json string
	 * @since 1.2
```

spring-cloud-gcp-logging/src/main/java/org/springframework/cloud/gcp/logging/StackdriverTraceConstants.java
```diff
@@ -66,6 +66,11 @@
 	 */
 	String MDC_FIELD_SPAN_EXPORT = "X-Span-Export";
 
+	/**
+	 * The JSON field name for the service context.
```
meltsufin:
```suggestion
	 * The JSON field name for the service context.
	 * @since 1.2
```

PR #1614, opened 2019-03-28
Datastore multithread query method fix and test

### LINKED ISSUE #1610: Can't initialize GqlDatastoreQuery properly in multithread environment: Annotated GQL Query Method findByQuery has 18 tags but a different number of parameter values: 3
My Code:

```
public interface CountryCountRepository extends DatastoreRepository<CountryCountProjection, Key> {

    @Query("select count, country from TABLE where date = @date and type = @type and platform = @platform")
    List<CountryCountProjection> findByQuery(String date, BQDataType type, MobilePlatform platform);

}
```

Problem:
method inside **GqlDatastoreQuery** failed when it's initialized with several threads.

```
	private List<String> getOriginalParamTags() {
		if (this.originalParamTags == null) {
			this.originalParamTags = new ArrayList<>();
I got exception here:
		if (tags.size() != vals.size()) {
			throw new DatastoreDataException("Annotated GQL Query Method "
					+ this.queryMethod.getName() + " has " + tags.size()
					+ " tags but a different number of parameter values: " + vals.size());
		}
```
this code throw exception:

```
	private GqlQuery<? extends BaseEntity> bindArgsToGqlQuery(String gql,
			List<String> tags,
			List vals) {
		Builder builder = GqlQuery.newGqlQueryBuilder(gql);
		builder.setAllowLiteral(true);
		if (tags.size() != vals.size()) {
			throw new DatastoreDataException("Annotated GQL Query Method "
					+ this.queryMethod.getName() + " has " + tags.size()
					+ " tags but a different number of parameter values: " + vals.size());
		}

```

Because:
in debug i have:

```
tags = {ArrayList@10850}  size = 12
 0 = "date"
 1 = "date"
 2 = "date"
 3 = "date"
 4 = "type"
 5 = "type"
 6 = "type"
 7 = "type"
 8 = "platform"
 9 = "platform"
 10 = "platform"
 11 = "platform"
vals = {ArrayList@10851}  size = 3
 0 = "20190222"
 1 = {BQDataType@10899} "NEW_USERS"
 2 = {MobilePlatform@10900} "ANDROID"
```

instead of:
```
tags = {ArrayList@10539}  size = 3
 0 = "date"
 1 = "type"
 2 = "platform"
vals = {ArrayList@10540}  size = 3
 0 = "20190215"
 1 = {BQDataType@10581} "NEW_USERS"
 2 = {MobilePlatform@10582} "ANDROID"
```

ChengyuanZhao:
fixes #1610
fixes error with initilializing query method  and adds integration test

### DIFF
--- a/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/PartTreeDatastoreQuery.java
+++ b/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/PartTreeDatastoreQuery.java
@@ -182,7 +182,7 @@ public class PartTreeDatastoreQuery<T> extends AbstractDatastoreQuery<T> {
 				|| (this.tree.isDelete() && returnedTypeIsNumber) || total;
 
 		Collector<?, ?, ?> collector = Collectors.toList();
-		if (isCountingQuery) {
+		if (isCountingQuery && !this.tree.isDelete()) {
 			collector = Collectors.counting();
 		}
 		else if (this.tree.isExistsProjection()) {
@@ -205,12 +205,15 @@ public class PartTreeDatastoreQuery<T> extends AbstractDatastoreQuery<T> {
 		Object result = StreamSupport.stream(rawResults.spliterator(), false).map(mapper).collect(collector);
 
 		if (this.tree.isDelete()) {
-			deleteFoundEntities(returnedTypeIsNumber, rawResults);
+			deleteFoundEntities(returnedTypeIsNumber, (Iterable) result);
 		}
-		boolean countingOrExistsQuery = this.tree.isExistsProjection() || isCountingQuery;
-		if (!countingOrExistsQuery) {
+
+		if (!this.tree.isExistsProjection() && !isCountingQuery) {
 			this.cursor = rawResults.getCursor();
 			return convertResultCollection(result, collectionType);
+		}
+		else if (isCountingQuery && this.tree.isDelete()) {
+			return ((List) result).size();
 		}
 		else {
 			return result;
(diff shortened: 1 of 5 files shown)


spring-cloud-gcp-data-datastore/src/test/java/org/springframework/cloud/gcp/data/datastore/it/DatastoreIntegrationTests.java
```diff
@@ -71,10 +68,7 @@
  */
 @RunWith(SpringRunner.class)
 @ContextConfiguration(classes = { DatastoreIntegrationTestConfiguration.class })
-public class DatastoreIntegrationTests {
-
-	// queries are eventually consistent, so we may need to retry a few times.
-	private static final int QUERY_WAIT_INTERVAL_SECONDS = 15;
```
dmitry-s: can we keep it 15 seconds?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/PartTreeDatastoreQuery.java
```diff
@@ -182,7 +182,7 @@ private Object execute(Object[] parameters, Class returnedElementType, Class<?>
 				|| (this.tree.isDelete() && returnedTypeIsNumber) || total;
 
 		Collector<?, ?, ?> collector = Collectors.toList();
-		if (isCountingQuery) {
+		if (isCountingQuery && !this.tree.isDelete()) {
```
dmitry-s: could you clarify why did you change that class?

ChengyuanZhao: It turns out we had a situation that was not covered by a test. 

If the query method was a delete method but required the return of the count of deleted IDs, then we would use the Collectors.counting() collector to iterate through the raw result and then we pass that raw result to the delete function. however, the iterable had already been used once and is at its end , so the delete operation would delete no IDs from the iterable.

spring-cloud-gcp-data-datastore/src/test/java/org/springframework/cloud/gcp/data/datastore/it/DatastoreIntegrationTests.java
ChengyuanZhao: done

PR #1615, opened 2019-03-28
Spanner multiple thread repo integration test

ChengyuanZhao:
Added threaded use of spanner repo integration test

### DIFF
--- a/spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/repository/it/ParallelSpannerRepositoryIntegrationTests.java
+++ b/spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/repository/it/ParallelSpannerRepositoryIntegrationTests.java
@@ -1,0 +1,80 @@
+/*
+ * Copyright 2017-2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.gcp.data.spanner.repository.it;
+
+import java.util.List;
+import java.util.function.IntConsumer;
+import java.util.stream.IntStream;
+
+import com.google.cloud.spanner.KeySet;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.cloud.gcp.data.spanner.test.AbstractSpannerIntegrationTest;
+import org.springframework.cloud.gcp.data.spanner.test.domain.Trade;
+import org.springframework.cloud.gcp.data.spanner.test.domain.TradeRepository;
+import org.springframework.test.context.junit4.SpringRunner;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Tests multiple threads using a single repository instance.
+ *
+ * @author Chengyuan Zhao
+ */
+@RunWith(SpringRunner.class)
+public class ParallelSpannerRepositoryIntegrationTests extends AbstractSpannerIntegrationTest {
+
+	private final static int PARALLEL_OPERATIONS = 20;
+
+	@Autowired
+	TradeRepository tradeRepository;
+
+	@Test
+	public void testParallelOperations() {
+
+		this.spannerOperations.delete(Trade.class, KeySet.all());
+
+		this.tradeRepository.performReadWriteTransaction(repo -> {
+
+			doOperation(x -> {
+				repo.save(Trade.aTrade());
+
+				// all of the threads are using the same transaction at the same time, so they all still
+				// see empty table
+				assertThat(repo.count()).isZero();
+			});
+			assertThat(repo.count()).isZero();
+			return null;
+		});
+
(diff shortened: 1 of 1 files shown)


ChengyuanZhao:
PTAL @dmitry-s @meltsufin

spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/repository/it/ParallelSpannerRepositoryIntegrationTests.java
```diff
@@ -0,0 +1,80 @@
...
+		doOperation(x -> this.tradeRepository.updateActionTradeById(
+				((List<Trade>) this.tradeRepository.findAll()).get(x).getId(),
+				"SELL"));
+
+		doOperation(x -> assertThat(this.tradeRepository.countByAction("SELL")).isEqualTo(PARALLEL_OPERATIONS));
+
+		doOperation(x -> assertThat(this.tradeRepository.countByActionQuery("SELL")).isEqualTo(PARALLEL_OPERATIONS));
+	}
+
+	private void doOperation(IntConsumer function) {
```
meltsufin: s/doOperation/executeInParallel

spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/repository/it/ParallelSpannerRepositoryIntegrationTests.java
```diff
@@ -0,0 +1,80 @@
...
+
+/**
+ * Tests multiple threads using a single repository instance.
+ *
+ * @author Chengyuan Zhao
+ */
+@RunWith(SpringRunner.class)
+public class ParallelSpannerRepositoryIntegrationTests extends AbstractSpannerIntegrationTest {
+
+	private final static int PARALLEL_OPERATIONS = 20;
```
meltsufin: Why 20? Looks like a magic number.

spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/repository/it/ParallelSpannerRepositoryIntegrationTests.java
```diff
@@ -0,0 +1,80 @@
...
+	TradeRepository tradeRepository;
+
+	@Test
+	public void testParallelOperations() {
+
+		this.spannerOperations.delete(Trade.class, KeySet.all());
+
+		this.tradeRepository.performReadWriteTransaction(repo -> {
+
+			doOperation(x -> {
```
meltsufin: What is `x`? Maybe a better name - `index`?

spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/repository/it/ParallelSpannerRepositoryIntegrationTests.java
```diff
@@ -0,0 +1,80 @@
...
+				((List<Trade>) this.tradeRepository.findAll()).get(x).getId(),
+				"SELL"));
+
+		doOperation(x -> assertThat(this.tradeRepository.countByAction("SELL")).isEqualTo(PARALLEL_OPERATIONS));
+
+		doOperation(x -> assertThat(this.tradeRepository.countByActionQuery("SELL")).isEqualTo(PARALLEL_OPERATIONS));
+	}
+
+	private void doOperation(IntConsumer function) {
+		IntStream.range(0, PARALLEL_OPERATIONS).parallel().forEach(function);
```
meltsufin: If you actually want to execute 20 operations in parallel, this will not work because `parallel()` uses a thread pool dependent on the number of CPU cores.

spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/repository/it/ParallelSpannerRepositoryIntegrationTests.java
ChengyuanZhao: It's an arbitrary number, and I don't think one could make a real argument for any number here. I'm open to any number that's not too small :)

spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/repository/it/ParallelSpannerRepositoryIntegrationTests.java
ChengyuanZhao: As long as 20 (or any other number you suggest) causes some parallel execution then that is fine. 
These threads (even if it's just 4) are sure transactions simultaneously.

ChengyuanZhao:
@meltsufin these test that operations can be done using a single repo instance and on some lines a single transaction context with expected results. 

This is in response to on the datastore side we had a bug that a user discovered when doing simple things simultaneously with a repo.

meltsufin:
@ChengyuanZhao Would 2 threads be enough? Since the exact order of execution is not guaranteed, how do you know that this is testing thread safety reliably? What change in the code would cause this to fail?

ChengyuanZhao:
@meltsufin this one here doesn't depend on order of execution, and it's meant to be a wide catch-all. The insert and count operations here don't depend on the order, and only depend on the value of the index given. I think 2 threads would be fine :) 

I'm looking to catch errors in our query methods and repo classes that would cause a crash (as it did in the Datastore module). The issue in the past with Datastore module was that our implementation of some query method classes wasn't prepared to handle threaded usage.  (in that case it caused the number of method params to be double counted because initialization was happening on first-query and that was happening simultaneously in several threads) .

ChengyuanZhao:
@meltsufin this one actually has no non-test changes associated and nothing was broken (this is Spanner, not the Datastore issue). It is just a test for posterity

PR #1618, opened 2019-03-29
Add Nullability Annotations on Storage module API

dzou:
Adds Nullability annotations to the Storage module using [@NonNullApi](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/lang/NonNullApi.html).

NonNullApi is added to `package-info.java`; at that point all public methods of our API are marked as returning non-null values. Then the nullable methods must be explicitly marked.

### DIFF
--- a/spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageLocation.java
+++ b/spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageLocation.java
@@ -19,6 +19,7 @@ package org.springframework.cloud.gcp.storage;
 import java.net.URI;
 import java.net.URISyntaxException;
 
+import org.springframework.lang.Nullable;
 import org.springframework.util.Assert;
 
 /**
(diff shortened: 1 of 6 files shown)


dzou:
Still researching if there is some way to help "verify" that the annotation are correct.

meltsufin:
Wouldn't certain Kotlin code only work depending on the annotations being present? I'm really out of my depth here. @artembilan Any suggestions on who might be able to take a look at this for us?

dzou:
> Wouldn't certain Kotlin code only work depending on the annotations being present? I'm really out of my depth here. @artembilan Any suggestions on who might be able to take a look at this for us?

If I understand correctly, the purpose of the nullability annotations is to give Kotlin users a compile-time check against null pointer exceptions. Without the annotations, you can still get NPE at runtime.

But it seems like any Java code is callable from Kotlin even without annotations, based on what I read here: https://kotlinlang.org/docs/reference/java-interop.html#null-safety-and-platform-types

dzou:
Closing for now due to inactivity; there seems to be a lack of interest for this feature. Am happy to reopen if new interest emerges.

Issue #1604, opened 2019-03-27
How to customize trace attributes/tags?


rajivml:
I have created a question on stackoverflow tagging spring-cloud-gcp but I didn't got any reply, so that's why re-posting it here again.
https://stackoverflow.com/questions/55320822/gke-stack-driver-trace-reporting-by-cluster-by-environment-by-service-by-service

We have multiple spring boot apps running on top of GKE and for spring boot applications am using spring-cloud-gcp-starter-trace to log traces to stack driver so that I can debug those traces via the stack driver UI.

Am not able to figure out how to add labels like service_name, service_version and cluster_name so that I can filter out only those traces for reporting purposes because right now we have istio configured on one cluster and even with one percent sampling rate it's generating tons of telemetry data and with UN-availability of filters or am missing some configuration, the trace UI has almost become useless for me

I had a look at the documentation for spring-cloud-gcp-starter-trace, they don't have any properties through which I can set these fields/tags.

Does spring-cloud-gcp-tracing library and stack driver tracing service support adding such tags and if yes, what are the touch points where I should add these tags and what are the list of valid tags applicable for GKE environment

Some references on what am talking about:
https://github.com/googleapis/cloud-trace-nodejs/issues/455
https://github.com/googleapis/cloud-trace-nodejs/issues/306

Can some one please let me know how can I achieve this.

rajivml:
I just now went through the documentation and found this property to set the name of the service. 

Similarly, am expecting few of other tags to be associated with these traces and stack driver tracing UI should honor these tags i.e allow me to filter the data based on these tags

Already Present:
spring.zipkin.service.name: foo

Expected Tags:
spring.zipkin.service.environment: prod/dev/qa
spring.zipkin.service.version: v1/v2/v3

meltsufin:
Hi @rajivml. Thanks for the question. We will look into it shortly. Currently, we're blocked due to another issue with Trace. #1608 

cc/ @saturnism

meltsufin:
@rajivml You can add custom tags using the `brave.SpanCustomizer`. Just autowire it in as the bean already exists in the application context.
You can then add tags, like this:
```
@Autowired
SpanCustomizer spanCustomizer;

...

spanCustomizer.tag("my-tag", "my tag value");
```
These will turn into labels on you traces in Stackdriver Trace, on which you can [search](https://cloud.google.com/trace/docs/trace-filters).

rajivml:
[...]

PR #1619, opened 2019-03-29
Add demonstration of SpanCustomizer in Trace sample

meltsufin:
Fixes: #1604.


meltsufin:
> Is it worth adding some notes in the docs to mention this? Or is it outside scope of spring-cloud-gcp?

Good call! Added.

PR #1626, opened 2019-04-02
Cloud config + bus sample integration test

elefeint:
Test setup creates a directory and writes to a file in /tmp.
The test starts up a config server and a config client sample applications, and tests both to make sure the configuration from the filesystem file is ingested properly.

### DIFF
--- a/spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-bus-config-sample/spring-cloud-gcp-pubsub-bus-config-sample-test/pom.xml
+++ b/spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-bus-config-sample/spring-cloud-gcp-pubsub-bus-config-sample-test/pom.xml
@@ -33,7 +33,6 @@
       <artifactId>spring-cloud-config-monitor</artifactId>
     </dependency>
 
-
   </dependencies>
 
   <build>
@@ -51,7 +50,6 @@
         <groupId>org.codehaus.mojo</groupId>
         <artifactId>build-helper-maven-plugin</artifactId>
         <version>3.0.0</version>
-
         <executions>
           <execution>
             <phase>generate-sources</phase>
@@ -67,12 +65,14 @@
           </execution>
         </executions>
       </plugin>
+
       <plugin>
         <artifactId>maven-deploy-plugin</artifactId>
         <configuration>
           <skip>true</skip>
         </configuration>
       </plugin>
+
     </plugins>
   </build>
 
--- a/spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-bus-config-sample/spring-cloud-gcp-pubsub-bus-config-sample-test/pom.xml
+++ b/spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-bus-config-sample/spring-cloud-gcp-pubsub-bus-config-sample-test/pom.xml
@@ -18,6 +18,22 @@
       <artifactId>spring-boot-starter-web</artifactId>
     </dependency>
 
+    <dependency>
+      <groupId>org.springframework.cloud</groupId>
+      <artifactId>spring-cloud-gcp-starter-bus-pubsub</artifactId>
+    </dependency>
+
+    <dependency>
+      <groupId>org.springframework.cloud</groupId>
+      <artifactId>spring-cloud-config-server</artifactId>
+    </dependency>
+
+    <dependency>
+      <groupId>org.springframework.cloud</groupId>
+      <artifactId>spring-cloud-config-monitor</artifactId>
+    </dependency>
+
+
   </dependencies>
 
   <build>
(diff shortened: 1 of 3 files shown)


spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-bus-config-sample/spring-cloud-gcp-pubsub-bus-config-sample-test/src/test/java/com/example/LocalSampleAppIntegrationTest.java
```diff
@@ -0,0 +1,131 @@
...
+		assumeThat(
+			"PUB/SUB integration tests are disabled. Use '-Dit.pubsub=true' to enable.",
+			System.getProperty("it.pubsub"), is("true"));
+
+		systemOut = System.out;
+		baos = new ByteArrayOutputStream();
+		TeeOutputStream out = new TeeOutputStream(systemOut, baos);
+		System.setOut(new PrintStream(out));
+
+		Files.createDirectories(Paths.get("/tmp/config_pubsub_integration_test"));
```
dmitry-s: make the directory name and the file name variables?

spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-bus-config-sample/spring-cloud-gcp-pubsub-bus-config-sample-test/src/test/java/com/example/LocalSampleAppIntegrationTest.java
```diff
@@ -0,0 +1,131 @@
...
+		SpringApplicationBuilder configServer = new SpringApplicationBuilder(PubSubConfigServerApplication.class)
+			.properties("server.port=8888",
+				"spring.profiles.active=native",
+				"spring.cloud.config.server.native.searchLocations=file:/tmp/config_pubsub_integration_test/");
+		configServer.run();
+
+		Awaitility.await("config server begins watching directory")
+			.atMost(60, TimeUnit.SECONDS)
+			.until(() -> {
+				return baos.toString().contains("Monitoring for local config changes: [/tmp/config_pubsub_integration_test]");
```
dmitry-s: If we are monitoring for changes, do we want to test what happens when we change the file?

elefeint:
Thanks for the review!
I added logic to change the file and watch it propagage.
Also, the test now starts the server/client applications the way the user would start them. The previous approach that copied only the Java files meant that the server and the client have the same dependencies, which caused all kinds of configuration contortions to prevent the config client from behaving like a config server.

@dmitry-s PTAL tomorrow.

spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-bus-config-sample/spring-cloud-gcp-pubsub-bus-config-sample-test/pom.xml
```diff
@@ -0,0 +1,23 @@
...
+  </parent>
+  <modelVersion>4.0.0</modelVersion>
+
+  <artifactId>spring-cloud-gcp-pubsub-bus-config-sample-test</artifactId>
+
+  <dependencies>
+
+    <dependency>
+      <groupId>org.springframework.boot</groupId>
+      <artifactId>spring-boot-starter-web</artifactId>
```
meltsufin: scope: test?

spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-bus-config-sample/spring-cloud-gcp-pubsub-bus-config-sample-test/src/test/java/com/example/LocalSampleAppIntegrationTest.java
```diff
@@ -0,0 +1,202 @@
...
+		}
+
+		if (this.configClientProcess != null) {
+			this.configClientProcess.destroy();
+		}
+
+	}
+
+	private void startConfigServer() throws IOException {
+		ProcessBuilder serverBuilder = new ProcessBuilder("mvn", "spring-boot:run",
```
meltsufin: `spring-cloud-gcp-integration-pubsub-sample-test` uses an in-process approach. Anything wrong with that approach?

elefeint: It's fine for the Pub/Sub integration sample, but it's not ideal for the config server sample because config server and config client need different dependencies.
It is possible to make work with an in-process approach, and I did that, but it required using custom several properties to suppress config server functionality in the client, which means that instead of testing the sample, we are testing a completely different scenario (since most of the sample IS configuration).

Here is what the client startup looked like in-process. By the time I was done explaning what deviations from default behavior were needed, it was clear that the sample was not being tested as is.
```
SpringApplicationBuilder
	configClient = new SpringApplicationBuilder(PubSubConfigApplication.class)
	.properties("server.port=8081",
		// suppress config server behavior.
		"spring.cloud.config.server.bootstrap=false",
		// re-enable config client behavior.
		"spring.cloud.config.enabled=true",
		"spring.cloud.config.watch.enabled=true",
		// Suppress Git validation configured through spring.provides in config server module.
		"spring.profiles.active=native"
);

```

meltsufin: I see. Thanks for the clarification.

PR #1638, opened 2019-04-05
Replace opt-out with opt-in autoconfigurations in GcpStorageAutoConfigurationTests

### LINKED ISSUE #1635: GcpStorageAutoConfigurationTests break due to spring web
Two-factor problem:
1) Spring Data Rest is broken by spring-projects/spring-framework#22543.
2) The storage test brings in lots of unnecessary autoconfigurations, including spring data rest.

elefeint:
Because `GcpStorageAutoConfigurationTests` lives in` spring-cloud-gcp-autoconfiguration`, every possible dependency is present on the classpath. The test then breaks whenever any autoconfigured dependency breaks (as it did in #1635), and newly added autoconfigurations have to be excluded (as was the case with IAP).

At this point, importing storage + web autoconfigurations is even terser than the earlier opt-out.

Fixes #1635.
Contributes to #1521.


elefeint:
@artembilan Like so? Had to lose the web setup, but that's tested in the sample already.

PR #1645, opened 2019-04-15
Add custom license header with https link

elefeint:
The 1.1.x branch is on an oldish minor version of parent `org.springframework.cloud:spring-cloud-build` project -- version `2.1.3`. 
It cannot be upgraded to newer minor versions because that requires removing Guava from 1.1.x -- it was too invasive of a change to backport.
Creating a project copy of a modern license header with "https" will fix the build, which is currently failing because all java files' license headers got updated to "https" in all branches.


artembilan:
I think we can live with that in the maintenance version.
Would be great to fix them all on `master` though.

Thanks

PR #1646, opened 2019-04-15
Fix compatibility with latest Boot dependencies

artembilan:
* Spring Integration 5.2 has removed `throws Exception` from its method
according Sonar requirements.
Therefore fix `PubSubMessageHandler` do not throw an `Exception`, but
catch all the intermediate and wrap them into the appropriate
`MessageException`
* Fix `SimpleDatastoreRepositoryTests` and `DatastoreTemplateTests`
to use `Sort.by()` instead of deprecated `Sort` ctor
* Also fix some warnings in those tests
* `@Ignore` the `PubSubExtendedBindingsPropertiesTests` until
Spring Cloud Stream is compatible with Spring Integration 5.2


artembilan:
> Is there a tracking issue in Spring Cloud Stream?

Now it is: https://github.com/spring-cloud/spring-cloud-stream/pull/1688

Issue #1664, opened 2019-04-30
Datastore reactive sample and integration tests. and ref doc


ChengyuanZhao:

meltsufin:
Closing, since we decided not to pursue reactive datastore.

Issue #1663, opened 2019-04-30
Datastore reactive repo implementation


ChengyuanZhao:
an imply of ReactiveCrudRepo for datastore. will use the reactive datastore template.

meltsufin:
Closing, since we decided not to pursue reactive datastore.

Issue #1662, opened 2019-04-30
Datastore Reactive Crud Repo boilerplate registrar and factory classes


ChengyuanZhao:
these are mostly boilerplate registration and factory classes for the reactive repo. 

They are mirror images of the standard repo boilerplate. 

includes unit tests.

meltsufin:
The description sounds like for a PR. :smile:

meltsufin:
Closing, since we decided not to pursue reactive datastore.

PR #1667, opened 2019-04-30
Datastore reactive template

### LINKED ISSUE #1659: Datastore template and converter/mapper produces fluxes
In effort to provide reactive repositories for Datastore we need to modify the converter/mapper to produce a flux from result set instead of doing the row-by-row iteration itself. 

The standard template can then also use this converter and simply consumes the entire flux before returning.

ChengyuanZhao:
fixes #1659 

This is the initial reactive implementation of Datastore Template. 

The intention is to use this in following implementations of reactive repo for Spring Data and in ref docs and samples. 
see #1662, #1663 , #1664

### DIFF
--- a/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/ReactiveDatastoreTemplate.java
+++ b/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/ReactiveDatastoreTemplate.java
@@ -1,0 +1,205 @@
+/*
+ * Copyright 2017-2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.gcp.data.datastore.core;
+
+import java.util.Collections;
+import java.util.List;
+import java.util.Objects;
+import java.util.Set;
+import java.util.function.Function;
+import java.util.stream.StreamSupport;
+
+import com.google.cloud.datastore.BaseEntity;
+import com.google.cloud.datastore.BaseKey;
+import com.google.cloud.datastore.Entity;
+import com.google.cloud.datastore.EntityQuery;
+import com.google.cloud.datastore.Key;
+import com.google.cloud.datastore.Query;
+import reactor.core.publisher.Flux;
+import reactor.core.publisher.Mono;
+
+import org.springframework.cloud.gcp.data.datastore.core.mapping.DatastorePersistentEntity;
+import org.springframework.data.domain.Example;
+import org.springframework.util.Assert;
+
+/**
+ * A reactive template for Cloud Datastore operations.
+ *
+ * @author Chengyuan Zhao
+ *
+ * @since 1.2
+ */
+public class ReactiveDatastoreTemplate {
+
+	private final DatastoreTemplate datastoreTemplate;
+
+	/**
+	 * Constructor.
+	 * @param datastoreTemplate A standard synchronous Datastore Template used to resolve
+	 *     relationships.
+	 */
+	public ReactiveDatastoreTemplate(DatastoreTemplate datastoreTemplate) {
+		Assert.notNull(datastoreTemplate,
+				"A non-null DatastoreTemplate is required.");
+		this.datastoreTemplate = datastoreTemplate;
+	}
+
+	public <T> Mono<T> findById(Object id, Class<T> entityClass) {
+		return performFindByKey(Collections.singleton(id), entityClass).next();
+	}
+
+	public <T> Mono<T> save(T instance, Key... ancestors) {
+		return Mono.fromSupplier(() -> {
+			List<T> instances = Collections.singletonList(instance);
+			this.datastoreTemplate.saveEntities(instances, ancestors);
+			return instance;
+		});
+	}
+
+	public <T> Flux<T> saveAll(Iterable<T> entities, Key... ancestors) {
+		return Flux.defer(() -> {
+			this.datastoreTemplate.saveEntities((List<T>) entities, ancestors);
+			return Flux.fromIterable(entities);
+		});
+	}
+
+	public <T> Mono<Void> deleteById(Object id, Class<T> entityClass) {
+		return Mono.fromRunnable(() -> this.datastoreTemplate.performDelete(
+				new Key[] { this.datastoreTemplate.getKeyFromId(id, entityClass) },
+				Collections.singletonList(id), null, entityClass));
+	}
+
+	public <T> Mono<Void> deleteAllById(Iterable<?> ids, Class<T> entityClass) {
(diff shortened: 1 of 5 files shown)


spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/ReactiveDatastoreTemplate.java
```diff
@@ -0,0 +1,205 @@
...
+import org.springframework.util.Assert;
+
+/**
+ * A reactive template for Cloud Datastore operations.
+ *
+ * @author Chengyuan Zhao
+ *
+ * @since 1.2
+ */
+public class ReactiveDatastoreTemplate {
```
ChengyuanZhao: make a ReactiveOperations interface.,

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/ReactiveDatastoreTemplate.java
```diff
@@ -0,0 +1,205 @@
...
+			return Flux.empty();
+		}
+
+		/*
+		 * We mark each key that will be read as DONE in the read context with a null value. The
+		 * null is only a placeholder to prevent circular-dependencies for when the actual entity
+		 * is read by the Flux below.
+		 */
+		for (BaseKey mk : keys) {
+			context.putReadEntity(mk, null);
```
ChengyuanZhao: check if the context is threadsafe

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/ReactiveDatastoreTemplate.java
```diff
@@ -0,0 +1,205 @@
...
+	public <T> Flux<T> findAllById(Iterable<?> ids, Class<T> entityClass) {
+		return performFindByKey(ids, entityClass);
+	}
+
+	private <T> Flux<T> performFindByKey(Iterable<?> ids, Class<T> entityClass) {
+		Set<Key> keys = this.datastoreTemplate.getKeysFromIds(ids, entityClass);
+		return findAllById(keys, entityClass, new DatastoreTemplate.ReadContext());
+	}
+
+	private <T> Flux<T> findAllById(Set<Key> keys, Class<T> entityClass, DatastoreTemplate.ReadContext context) {
```
ChengyuanZhao: How about `publisher<keys>`

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/ReactiveDatastoreTemplate.java
```diff
@@ -0,0 +1,205 @@
...
+						.getPersistentEntity(entityClass).kindName())
+				.build());
+	}
+
+	private <A> Flux<A> convertAndResolveRelationshipsEagerly(Flux<Entity> entities, Class<A> entityClass,
+			DatastoreTemplate.ReadContext readContext) {
+		DatastorePersistentEntity datastorePersistentEntity = this.datastoreTemplate.getDatastoreMappingContext()
+				.getPersistentEntity(entityClass);
+		return entities.doOnNext(entity -> readContext.putReadEntity(entity.getKey(), entity))
+				.map(entity -> this.datastoreTemplate.convertEntityResolveDescendantsAndReferences(entityClass,
```
ChengyuanZhao: look at pubsub example, get another thread to do this single entity conversion. "executorservice". 
some sort of "zip" that combines parent and children when both are ready.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/ReactiveDatastoreTemplate.java
```diff
@@ -0,0 +1,205 @@
...
+	}
+
+	public <T> Mono<T> findById(Object id, Class<T> entityClass) {
+		return performFindByKey(Collections.singleton(id), entityClass).next();
+	}
+
+	public <T> Mono<T> save(T instance, Key... ancestors) {
+		return Mono.fromSupplier(() -> {
+			List<T> instances = Collections.singletonList(instance);
+			this.datastoreTemplate.saveEntities(instances, ancestors);
```
ChengyuanZhao: could just call same `save` function on regular template?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/ReactiveDatastoreTemplate.java
```diff
@@ -0,0 +1,205 @@
...
+	 *     relationships.
+	 */
+	public ReactiveDatastoreTemplate(DatastoreTemplate datastoreTemplate) {
+		Assert.notNull(datastoreTemplate,
+				"A non-null DatastoreTemplate is required.");
+		this.datastoreTemplate = datastoreTemplate;
+	}
+
+	public <T> Mono<T> findById(Object id, Class<T> entityClass) {
+		return performFindByKey(Collections.singleton(id), entityClass).next();
```
ChengyuanZhao: regarding event, for `mono` results, emitting result when action happens makes sense.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/ReactiveDatastoreTemplate.java
```diff
@@ -0,0 +1,205 @@
...
+	public <T> Flux<Key> keyQueryByExample(Example<T> example, DatastoreQueryOptions queryOptions) {
+		Query query = this.datastoreTemplate.exampleToQuery(example, queryOptions, true);
+		return Flux.fromIterable(() -> this.datastoreTemplate.getDatastoreReadWriter().run(query));
+	}
+
+	public <T> Mono<Boolean> existsById(Object id, Class<T> entityClass) {
+		return findById(id, entityClass).map(Objects::nonNull);
+	}
+
+	private Flux<Key> findAllKeys(Class entityClass) {
```
ChengyuanZhao: Since there is no usecase where we use keys one at a time we could remove tyhis and just use the old blocking function.

spring-cloud-gcp-data-datastore/src/test/java/org/springframework/cloud/gcp/data/datastore/core/ReactiveDatastoreTemplateTests.java
```diff
@@ -0,0 +1,542 @@
...
+		when(this.datastore.fetch(ArgumentMatchers.<Key[]>any())).thenReturn(Collections.singletonList(null));
+		verifyBeforeAndAfterEvents(null, new AfterFindByKeyEvent(Collections.emptyList(), Collections.singleton(null)),
+				() -> assertThat(this.datastoreTemplate.findById(createFakeKey("key0"),
+						AbstractDatastoreTemplateTests.TestEntity.class)).isNull(),
+				x -> {
+				});
+	}
+
+	@Test
+	public void findAllByIdTest() {
```
ChengyuanZhao: is there a way to reuse code from the DatastoreTemplateTests that just swap in the template and then add the subscribe boilerplate.

ChengyuanZhao: use .block() and you can just get the inner contents back.

Issue #1673, opened 2019-05-03
Datastore List property with NULL values throws error


oarshad:
An entity with a property list which has null values throwing error:

```
@Entity(name = "Player")
public class PlayerKind {
   private List<String> collection;
}
```
```
org.springframework.dao.InvalidDataAccessApiUsageException: Cannot convert a null value.; nested exception is java.lang.IllegalArgumentException: Cannot convert a null value.

	at org.springframework.orm.jpa.EntityManagerFactoryUtils.convertJpaAccessExceptionIfPossible(EntityManagerFactoryUtils.java:373)
	at org.springframework.orm.jpa.vendor.HibernateJpaDialect.translateExceptionIfPossible(HibernateJpaDialect.java:255)
	at org.springframework.orm.jpa.AbstractEntityManagerFactoryBean.translateExceptionIfPossible(AbstractEntityManagerFactoryBean.java:527)
	at org.springframework.dao.support.ChainedPersistenceExceptionTranslator.translateExceptionIfPossible(ChainedPersistenceExceptionTranslator.java:61)
	at org.springframework.dao.support.DataAccessUtils.translateIfNecessary(DataAccessUtils.java:242)
	at org.springframework.dao.support.PersistenceExceptionTranslationInterceptor.invoke(PersistenceExceptionTranslationInterceptor.java:153)
	at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:186)
	at org.springframework.aop.framework.JdkDynamicAopProxy.invoke(JdkDynamicAopProxy.java:212)
	at com.sun.proxy.$Proxy218.findAllByPlayerId(Unknown Source)
```

The value in the collection in datastore is as follow: `collection=[NullValue{valueType=NULL, excludeFromIndexes=false, meaning=0, value=null}, NullValue{valueType=NULL, excludeFromIndexes=false, meaning=0, value=null}, StringValue{valueType=STRING, excludeFromIndexes=true, meaning=0, value=DUMMY}]`

Without the null values it works, however in my situation I do need null values to be present in my list.

meltsufin:
@dmitry-s @ChengyuanZhao  PTAL

ChengyuanZhao:
@oarshad Thank you for reporting this issue! Which version of the module are you using? I'll try to replicate with our current build in the meantime.

oarshad:
[...]

PR #1674, opened 2019-05-03
datastore null property fix

ChengyuanZhao:
related #1673 
allows null values in arrays and test.


ChengyuanZhao:
@meltsufin  my bad, it turns out I had to remove an existing test that specifically checks for this situation haha... could you re-approve?

PR #1678, opened 2019-05-08
PubsubTemplate publish to topics in other projects

meltsufin:
This change to the DefaultPublisherFactory allows the project id of the
topic to be overwitten directly from the topic specification in the
topic string.

This allows use-cases like
pubSubTemplate.publish("projects/other-project/topics/the-topic", "payload").

This change applied across the board to allow publishing, subscribing, creating, and deleting topics using fully-qualified topic names.

Fixes #1678.

### DIFF
--- a/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/DefaultPublisherFactory.java
+++ b/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/DefaultPublisherFactory.java
@@ -128,7 +128,7 @@ public class DefaultPublisherFactory implements PublisherFactory {
 		return this.publishers.computeIfAbsent(topic, (key) -> {
 			try {
 				Publisher.Builder publisherBuilder =
-						Publisher.newBuilder(PubSubTopicUtils.parseTopic(this.projectId, topic));
+						Publisher.newBuilder(PubSubTopicUtils.toProjectTopicName(topic, this.projectId));
 
 				if (this.executorProvider != null) {
 					publisherBuilder.setExecutorProvider(this.executorProvider);
--- a/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/DefaultPublisherFactory.java
+++ b/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/DefaultPublisherFactory.java
@@ -65,7 +65,7 @@ public class DefaultPublisherFactory implements PublisherFactory {
 	/**
 	 * Create {@link DefaultPublisherFactory} instance based on the provided {@link GcpProjectIdProvider}.
 	 * <p>The {@link GcpProjectIdProvider} must not be null, neither provide an empty {@code projectId}.
-	 * @param projectIdProvider provides the GCP project ID
+	 * @param projectIdProvider provides the default GCP project ID for selecting the topic
 	 */
 	public DefaultPublisherFactory(GcpProjectIdProvider projectIdProvider) {
 		Assert.notNull(projectIdProvider, "The project ID provider can't be null.");
@@ -128,8 +128,15 @@ public class DefaultPublisherFactory implements PublisherFactory {
 	public Publisher createPublisher(String topic) {
 		return this.publishers.computeIfAbsent(topic, (key) -> {
 			try {
+				String finalProject = this.projectId;
+				String finalTopic = topic;
+				if (topic.contains("/")) {
+					int splitIndex = topic.indexOf("/");
+					finalProject = topic.substring(0, splitIndex);
+					finalTopic = topic.substring(splitIndex + 1);
+				}
 				Publisher.Builder publisherBuilder =
-						Publisher.newBuilder(ProjectTopicName.of(this.projectId, key));
+						Publisher.newBuilder(ProjectTopicName.of(finalProject, finalTopic));
 
 				if (this.executorProvider != null) {
 					publisherBuilder.setExecutorProvider(this.executorProvider);
--- a/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/DefaultPublisherFactory.java
+++ b/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/DefaultPublisherFactory.java
@@ -27,7 +27,6 @@ import com.google.api.gax.retrying.RetrySettings;
 import com.google.api.gax.rpc.HeaderProvider;
 import com.google.api.gax.rpc.TransportChannelProvider;
 import com.google.cloud.pubsub.v1.Publisher;
-import com.google.pubsub.v1.ProjectTopicName;
 
 import org.springframework.cloud.gcp.core.GcpProjectIdProvider;
 import org.springframework.cloud.gcp.pubsub.core.PubSubException;
@@ -128,15 +127,8 @@ public class DefaultPublisherFactory implements PublisherFactory {
 	public Publisher createPublisher(String topic) {
 		return this.publishers.computeIfAbsent(topic, (key) -> {
 			try {
-				String finalProject = this.projectId;
-				String finalTopic = topic;
-				if (topic.contains("/")) {
-					int splitIndex = topic.indexOf("/");
-					finalProject = topic.substring(0, splitIndex);
-					finalTopic = topic.substring(splitIndex + 1);
-				}
 				Publisher.Builder publisherBuilder =
-						Publisher.newBuilder(ProjectTopicName.of(finalProject, finalTopic));
+						Publisher.newBuilder(PubSubTopicUtils.parseTopic(this.projectId, topic));
 
 				if (this.executorProvider != null) {
 					publisherBuilder.setExecutorProvider(this.executorProvider);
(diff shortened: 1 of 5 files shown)


spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/DefaultPublisherFactory.java
```diff
@@ -128,8 +128,15 @@ public void setBatchingSettings(BatchingSettings batchingSettings) {
 	public Publisher createPublisher(String topic) {
 		return this.publishers.computeIfAbsent(topic, (key) -> {
 			try {
+				String finalProject = this.projectId;
+				String finalTopic = topic;
+				if (topic.contains("/")) {
```
artembilan: Why `split("/")` doesn't work instead?
There is going to be a single element in the result array if there is no `/`.

meltsufin: I thought of using `split('/")`, but what if there is more than one `/` used in the topic.

meltsufin:
@artembilan I appreciate you taking a look, but as indicated, this is just a draft and not ready for review. I opened it to show an idea for someone who asked about this over email.

Docs and tests are coming, if the idea is accepted.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/DefaultPublisherFactory.java
artembilan: There is an overloaded `split()` method then with `limit`:
```
 * <p> The {@code limit} parameter controls the number of times the
     * pattern is applied and therefore affects the length of the resulting
     * array.  If the limit <i>n</i> is greater than zero then the pattern
     * will be applied at most <i>n</i>&nbsp;-&nbsp;1 times, the array's
     * length will be no greater than <i>n</i>, and the array's last entry
     * will contain all input beyond the last matched delimiter.  If <i>n</i>
     * is non-positive then the pattern will be applied as many times as
     * possible and the array can have any length.  If <i>n</i> is zero then
     * the pattern will be applied as many times as possible, the array can
     * have any length, and trailing empty strings will be discarded.
```

meltsufin: Thanks, I didn't know about that.

spring-cloud-gcp-pubsub/src/test/java/org/springframework/cloud/gcp/pubsub/support/PubSubTopicUtilsTests.java
```diff
@@ -0,0 +1,97 @@
...
+ * Tests for {@link PubSubTopicUtils}.
+ *
+ * @author Mike Eltsufin
+ */
+public class PubSubTopicUtilsTests {
+
+	/**
+	 * used to check exception messages and types.
+	 */
+	@Rule
```
elefeint: Replace with `assertThatThrownBy()`? We've been pretty consistent about it recently.

spring-cloud-gcp-pubsub/src/test/java/org/springframework/cloud/gcp/pubsub/support/PubSubTopicUtilsTests.java
```diff
@@ -0,0 +1,97 @@
...
+		assertThat(parsedProjectTopicName).isEqualTo(ProjectTopicName.of(project, topic));
+		assertThat(parsedProjectTopicName.toString()).isEqualTo(fqn);
+	}
+
+	@Test
+	public void testParseTopic_no_topic() {
+		expectedEx.expect(IllegalArgumentException.class);
+		expectedEx.expectMessage("The topic can't be null.");
+
+		ProjectTopicName parsedProjectTopicName = PubSubTopicUtils.parseTopic("topicA", null);
```
elefeint: Don't really need this variable.

spring-cloud-gcp-pubsub/src/test/java/org/springframework/cloud/gcp/pubsub/support/PubSubTopicUtilsTests.java
```diff
@@ -0,0 +1,97 @@
...
+		expectedEx.expect(IllegalArgumentException.class);
+		expectedEx.expectMessage("The project ID can't be null when using canonical topic name.");
+
+		String project = "projectA";
+		String topic = "topicA";
+		String fqn = "projects/" + project + "/topics/" + topic;
+
+		ProjectTopicName parsedProjectTopicName = PubSubTopicUtils.parseTopic(null, topic);
+
+		assertThat(parsedProjectTopicName).isEqualTo(ProjectTopicName.of(project, topic));
```
elefeint: The test never gets to those asserts because `parseTopic` throws. Remove those, and also `project` and `fqn` variables.

spring-cloud-gcp-pubsub/src/test/java/org/springframework/cloud/gcp/pubsub/support/PubSubTopicUtilsTests.java
```diff
@@ -0,0 +1,97 @@
...
+	@Test
+	public void testParseTopic_fqn() {
+		String project = "projectA";
+		String topic = "topicA";
+		String fqn = "projects/" + project + "/topics/" + topic;
+
+		ProjectTopicName parsedProjectTopicName = PubSubTopicUtils.parseTopic(project, fqn);
+
+		assertThat(parsedProjectTopicName).isEqualTo(ProjectTopicName.of(project, topic));
+		assertThat(parsedProjectTopicName.toString()).isEqualTo("projects/" + project + "/topics/" + topic);
```
elefeint: `assertThat(parsedProjectTopicName.toString()).isEqualTo(fqn);`?

spring-cloud-gcp-pubsub/src/test/java/org/springframework/cloud/gcp/pubsub/support/PubSubTopicUtilsTests.java
```diff
@@ -0,0 +1,97 @@
...
+	@Test
+	public void testParseTopic_fqn_no_project() {
+		String project = "projectA";
+		String topic = "topicA";
+		String fqn = "projects/" + project + "/topics/" + topic;
+
+		ProjectTopicName parsedProjectTopicName = PubSubTopicUtils.parseTopic(null, fqn);
+
+		assertThat(parsedProjectTopicName).isEqualTo(ProjectTopicName.of(project, topic));
+		assertThat(parsedProjectTopicName.toString()).isEqualTo("projects/" + project + "/topics/" + topic);
```
elefeint: same, use `fqn` since it was already concatenated.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/PubSubTopicUtils.java
```diff
@@ -0,0 +1,59 @@
...
+	/**
+	 * Creates a {@link ProjectTopicName} based on a topic name within a project or the
+	 * fully-qualified topic name. If the specified topic is in the
+	 * projects/&lt;project_name&gt;/topics/&lt;topic_name&gt; format, then the {@code projectId} is
+	 * ignored}
+	 * @param projectId the project ID to use if the topic is not a fully-qualified name
+	 * @param topic the topic name in the project or the fully-qualified project name
+	 * @return the Pub/Sub object representing the topic name
+	 */
+	public static ProjectTopicName parseTopic(String projectId, String topic) {
```
artembilan: I would change the order since it looks like `projectId` is optional. We might name it as a `defaultProjectId`.
Also would be great to mark it with the `@org.springframework.lang.Nullable`

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/PubSubTopicUtils.java
```diff
@@ -0,0 +1,59 @@
+/*
+ * Copyright 2017-2018 the original author or authors.
```
artembilan: It's already far away `2019`

spring-cloud-gcp-pubsub/src/test/java/org/springframework/cloud/gcp/pubsub/support/PubSubTopicUtilsTests.java
artembilan: 👍

spring-cloud-gcp-pubsub/src/test/java/org/springframework/cloud/gcp/pubsub/support/PubSubTopicUtilsTests.java
artembilan: Looks like the test has not been finished.
Although I guess that is a style how `ExpectedException`.

Really would be better to move to `assertThatIllegalArgumentException()` for consistency and more cleaner code.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/PubSubTopicUtils.java
dzou: Not sure if `parseTopic` is the right name for this; while it does parse the topic name, it still accepts the projectId as input. Perhaps something like `toProjectTopicName` or something would seem more natural.

PR #1687, opened 2019-05-17
Fix GcpDatastoreEmulatorIntTests for projectId

artembilan:
The `projectId` is required for the `DatastoreOptions` and if we don't
specify it explicitly, the `ServiceOptions` super class falls back to
the `getDefaultProjectId()`, which eventually reads a projectId from the
`GOOGLE_APPLICATION_CREDENTIALS` files.

The `GcpDatastoreEmulatorIntegrationTests` fails on th CI server just
because there is no any credentials JSON file in that environment

* Specify `projectId` explicitly for the `DatastoreOptions.Builder`.
Use for this purpose a `GcpProjectIdProvider` bean from the context.


spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/datastore/it/GcpDatastoreEmulatorIntegrationTests.java
```diff
@@ -24,6 +24,7 @@
 import com.google.cloud.datastore.EntityQuery;
 import com.google.cloud.datastore.Query;
 import com.google.cloud.datastore.StructuredQuery;
+
```
meltsufin: checkstyle doesn't like this space.

meltsufin:
This one is still failing:
`PubSubMessageChannelBinderEmulatorTests>AbstractBinderTests.testSendPojoReceivePojoKryoWithStreamListener:503 » MessageDelivery`

Issue #1692, opened 2019-05-22
Datastore Sample failing


meltsufin:
```
Caused by: org.springframework.beans.factory.NoSuchBeanDefinitionException: No qualifying bean of type 'com.example.SingerRepository' available: expected at least 1 bean which qualifies as autowire candidate. Dependency annotations: {@org.springframework.beans.factory.annotation.Autowired(required=true)}
	at org.springframework.beans.factory.support.DefaultListableBeanFactory.raiseNoMatchingBeanFound(DefaultListableBeanFactory.java:1681)
	at org.springframework.beans.factory.support.DefaultListableBeanFactory.doResolveDependency(DefaultListableBeanFactory.java:1239)
	at org.springframework.beans.factory.support.DefaultListableBeanFactory.resolveDependency(DefaultListableBeanFactory.java:1193)
	at org.springframework.beans.factory.annotation.AutowiredAnnotationBeanPostProcessor$AutowiredFieldElement.inject(AutowiredAnnotationBeanPostProcessor.java:623)
```

@ChengyuanZhao @dmitry-s PTAL

ChengyuanZhao:
I've gone as far back as this commit on master that used to work but also no longer does

![image](https://user-images.githubusercontent.com/29288945/58272460-91515300-7d5c-11e9-893e-70f67766c2dd.png)


I'll try to use some older versions of dependencies...

elefeint:
Since my face made an appearance... Spring-boot bom upgraded to spring-data Moore-M4 recently: https://github.com/spring-projects/spring-boot/issues/16574

ChengyuanZhao:
hahahaha thank you Detective Elena !!

PR #1696, opened 2019-05-23
Spring boot related config change

ChengyuanZhao:
related #1692 
New version of Spring Boot dependency somehow caused that bean to not be detected.

### DIFF
--- a/spring-cloud-gcp-samples/spring-cloud-gcp-data-datastore-sample/src/main/java/com/example/DatastoreRepositoryExample.java
+++ b/spring-cloud-gcp-samples/spring-cloud-gcp-data-datastore-sample/src/main/java/com/example/DatastoreRepositoryExample.java
@@ -26,7 +26,6 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.CommandLineRunner;
 import org.springframework.boot.SpringApplication;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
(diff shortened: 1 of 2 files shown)


ChengyuanZhao:
@artembilan Hey Artem, this PR actually isn't Spring Data related, so I thought you should know. With new version of spring dependency coming in 10 days ago our sample integration test broke because that bean i moved in this PR wasn't being detected anymore (it used to work and was detected).

artembilan:
@snicoll ,

Any pointers why the bean defined in the `@SpringBootApplication` class with `main()` is not visible any more in the latest Boot 2.2?
Moving that bean into the separate `@Configuration` fixed the issue. 

Thanks

snicoll:
Nothing jumps at me but I can certainly see why it wouldn't work. It's not related for it being the main class or not. The repository is injected using field injection and, as far as I understand, the actual implementation for that repository requires the converter that's defined in the very same configuration class. 

In general, please don't do field injection and don't do it in particular on the main Spring Boot app: this is the root of the context and therefore anything you define there forces it to eagerly fetch whatever is required to complete the initialization of the bean. Perhaps moving all that code that a dedicated type and calling it from the command line runner would be more clear anyway?

That said, if the context managed this slightly unusual arrangement, I think it would be nice to at least understand why it's not handled anymore. If you have a sample you can share, we're more than happy to have a look to it (cc @jhoeller).

PR #1703, opened 2019-06-18
Spanner array param binding

### LINKED ISSUE #1701: Spring Data Spanner Repositories `In` clause queries support
Spring data spanner gives `java.lang.UnsupportedOperationException: The statement type: IN (1): [IsIn, In] is not supported.` when trying to use `IN` for selecting columns  part of a composite key in a SpannerRepository interface
`@Query("select * from <TableName> t where t.<columnName1>= @param1 AND t.columnName2 IN (@param2)")`
```Set<OnHandInventory> findBy<ColumnName1>And<ColumnName2>In(@Param("param1") String columnOneValue,@Param("param2") Set<Long> column2Values);```

Note: Exception being thrown at [SpannerStatementQueryExecutor.java](https://github.com/spring-cloud/spring-cloud-gcp/blob/72d746486fedd19b16ead3cdbbcfb48c3e90c3a9/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryExecutor.java#L399)

### PR DESCRIPTION
**Pull request by ChengyuanZhao**
fixes #1701 
array param binding in queries. and IN clause

### DIFF
--- a/docs/src/main/asciidoc/spanner.adoc
+++ b/docs/src/main/asciidoc/spanner.adoc
@@ -1038,6 +1038,8 @@ The following filter options are supported:
 * Not like a string
 * Contains a string
 * Not contains a string
+* In
+* Not in
 
 Note that the phrase `SymbolIgnoreCase` is translated to `LOWER(SYMBOL) = LOWER(?)` indicating a non-case-sensitive matching.
 The `IgnoreCase` phrase may only be appended to fields that correspond to columns of type STRING or BYTES.
--- a/docs/src/main/asciidoc/spanner.adoc
+++ b/docs/src/main/asciidoc/spanner.adoc
@@ -1059,6 +1059,8 @@ List<Trade> findBySymbolContains(String symbolFragment);
 ----
 The param `symbolFragment` is a https://cloud.google.com/spanner/docs/functions-and-operators#regexp_contains[regular expression] that is checked for occurrences.
 
+The `In` and `NotIn` keywords must be used with `Iterable` corresponding parameters.
+
 Delete queries are also supported.
 For example, query methods such as `deleteByAction` or `removeByAction` delete entities found by `findByAction`.
 The delete operation happens in a single transaction.
@@ -1178,6 +1180,9 @@ SpEL can also be used to provide SQL parameters:
   AND price > #{#priceRadius * -1} AND price < #{#priceRadius * 2}")
 List<Trade> fetchByActionNamedQuery(String action, Double priceRadius);
 ----
+
+When using the `IN` SQL clause, remember to use `IN UNNEST(@iterableParam)` to specify a single `Iterable` parameter.
+You can also use a fixed number of singular parameters such as `IN (@stringParam1, @stringParam2)`.
 
 ==== Projections
 Spring Data Spanner supports {spring-data-commons-ref}/#projections[projections].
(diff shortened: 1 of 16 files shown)

### CODE REVIEW

**Comment by ChengyuanZhao** [comment]
The Spanner module builds, but the Streams module fails..... 

@dmitry-s

**Comment by ChengyuanZhao** [comment]
@dmitry-s @meltsufin PTAL

File: spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryExecutor.java [hunk 1]
```diff
@@ -193,16 +216,45 @@ public static Statement buildStatementFromSqlWithArgs(String sql, List<String> t
...
+	private static void bindArrayParameter(ValueBinder<Statement.Builder> bind,
+			SpannerCustomConverter spannerCustomConverter,
+			Iterable originalParam, Parameter paramMetadata) {
+		Iterable param = originalParam;
+		Class innerType = (Class) ((ParameterizedType) paramMetadata.getParameterizedType())
+				.getActualTypeArguments()[0];
+		BiConsumer<ValueBinder<?>, Iterable> toMethod = getIterableValueBinderBiConsumer(innerType);
+
+		// attempt conversion of elements inside
+		if (toMethod == null) {
```
dmitry-s: should this logic be in `ConverterAwareMappingSpannerEntityWriter`?

docs/src/main/asciidoc/spanner.adoc
```diff
@@ -1038,6 +1038,8 @@ The following filter options are supported:
 * Not like a string
 * Contains a string
 * Not contains a string
+* In
+* Not in
```
dmitry-s: we should document `IN UNNEST` syntax and when to use it instead of `IN` for @Query annotated methods

ChengyuanZhao: done.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryExecutor.java
ChengyuanZhao: I tried doing this but I decided against it. I tried extracting a method but the signature needed to have several arguments and they were rather specialized. I felt that it was an uncomfortably specialized method that shouldn't be public on ConverterAwareMappingSpannerEntityWriter

ChengyuanZhao: @dmitry-s actually great idea. I refactored . PTAL!

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/ConverterAwareMappingSpannerEntityWriter.java
```diff
@@ -193,6 +193,81 @@ public SpannerWriteConverter getSpannerWriteConverter() {
 		return this.writeConverter;
 	}
 
+	/**
```
ChengyuanZhao: @dmitry-s  Don't be alarmed by these big new sections .They're just moved up because now they're public.

Issue #1698, opened 2019-06-10
pom file forcing external repositories into a private build system


abgm:
This is a transitive dependency of my project that is built with a private nexus.
Build output : 
```
Downloading from gcs-mirror: https://maven-central.storage-download.googleapis.com/repos/central/data/io/grpc/grpc-core/maven-metadata.xml
[WARNING] Could not transfer metadata io.grpc:grpc-core/maven-metadata.xml from/to spring-snapshots (https://repo.spring.io/libs-snapshot-local): Connect to repo.spring.io:443 [repo.spring.io/35.241.58.96] failed: Connection timed out: con
nect
[WARNING] Could not transfer metadata io.grpc:grpc-core/maven-metadata.xml from/to spring-milestones (https://repo.spring.io/libs-milestone-local): Connect to repo.spring.io:443 [repo.spring.io/35.241.58.96] failed: Connection timed out: c
onnect
[WARNING] Could not transfer metadata io.grpc:grpc-core/maven-metadata.xml from/to spring-releases (https://repo.spring.io/release): Connect to repo.spring.io:443 [repo.spring.io/35.241.58.96] failed: Connection timed out: connect
[WARNING] Could not transfer metadata io.grpc:grpc-core/maven-metadata.xml from/to gcs-mirror (https://maven-central.storage-download.googleapis.com/repos/central/data): Connect to maven-central.storage-download.googleapis.com:443 [maven-c
entral.storage-download.googleapis.com/74.125.136.128] failed: Connection timed out: connect
```
https://github.com/spring-cloud/spring-cloud-gcp/blob/18778f5e77690a7038dfac9436e1b4fec3e0b388/pom.xml#L231

I already tried to turn off the spring profile via maven option but without success.
`mvn -U -V clean compile -P !spring`

related to : https://github.com/spring-cloud/spring-cloud-gcp/issues/1522

meltsufin:
It looks like your private Nexus doesn't have some of the required dependencies for Spring Cloud GCP, like `io.grpc:grpc-core`.
Which version of Spring Cloud GCP are you using?

abgm:
I can build with success.
But because this repositories are defined on this pom, maven try to get the maven-metadata.xml for them... 
This repositories definitions shouldn't be done on the pom file but on the maven setting...

meltsufin:
The repositories are defined in the `spring` profile.
Try running `mvn package -P '!spring'`.

abgm:
[...]

PR #1717, opened 2019-07-01
Remove Spring Profile activeByDefault

dzou:
Removes the Spring profile `activeByDefault` setting from pom.xml.

Fixes: #1698

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -225,9 +225,6 @@
 	<profiles>
 		<profile>
 			<id>spring</id>
-			<activation>
-				<activeByDefault>true</activeByDefault>
-			</activation>
 			<repositories>
 				<repository>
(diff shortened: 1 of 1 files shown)


meltsufin:
How is Travis still passing without you explicitly enabling this profile? :confused:

dzou:
> How is Travis still passing without you explicitly enabling this profile?

Looks like it is actually enabled here on the `install` line: https://github.com/spring-cloud/spring-cloud-gcp/blob/master/.travis.yml#L34

meltsufin:
> > How is Travis still passing without you explicitly enabling this profile?
> 
> Looks like it is actually enabled here on the `install` line: https://github.com/spring-cloud/spring-cloud-gcp/blob/master/.travis.yml#L34

Oh, I see.

PR #1718, opened 2019-07-01
Add basic integration test for Vision OCR sample app

### LINKED ISSUE #1555: Add integration tests for Vision OCR sample
Add integration tests for Vision OCR sample. Follow up to #1554.

dzou:
Add a basic integration test for the Vision OCR sample to check for `200` success response when hitting the app.

Note that we already have integration tests for the vision ocr template via. `DocumentOcrTemplateIntegrationTests`; this test is to just verify that the application will startup correctly.

Fixes #1555.

### DIFF
--- a/spring-cloud-gcp-samples/spring-cloud-gcp-vision-ocr-demo/src/test/java/VisionOcrSampleTest.java
+++ b/spring-cloud-gcp-samples/spring-cloud-gcp-vision-ocr-demo/src/test/java/VisionOcrSampleTest.java
@@ -14,46 +14,47 @@
  * limitations under the License.
  */
 
-import static org.assertj.core.api.Assertions.assertThat;
-import static org.hamcrest.Matchers.is;
-import static org.junit.Assume.assumeThat;
-
 import com.example.Application;
 import org.junit.BeforeClass;
 import org.junit.Test;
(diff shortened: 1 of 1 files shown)


spring-cloud-gcp-samples/spring-cloud-gcp-vision-ocr-demo/src/test/java/VisionOcrSampleTest.java
```diff
@@ -0,0 +1,60 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.hamcrest.Matchers.is;
+import static org.junit.Assume.assumeThat;
+
+@RunWith(SpringRunner.class)
+@SpringBootTest(
+		webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT,
+		classes = { Application.class },
+		properties = {
```
elefeint: does not need to be an array, does it?

dzou: Done. Indeed, looks like it could be simplified to the variable itself if there is only 1 element.

PR #1719, opened 2019-07-03
Enable multi data modules

### LINKED ISSUE #1512: Spanner/Datastore - unable to use both Spring Data Spanner & Datastore in the same app
Added both Spanner & Datastore starters to the same application, and tried to use both repositories.
The application failed to start.

```
***************************
APPLICATION FAILED TO START
***************************

Description:

The bean 'personRepositiory', defined in null, could not be registered. A bean with that name has already been defined in null and overriding is disabled.

Action:

Consider renaming one of the beans or enabling overriding by setting spring.main.allow-bean-definition-overriding=true

```

ChengyuanZhao:
fixes #1512 

adds support and test for using Spring  Data Spanner and Datastore in single app.


spring-cloud-gcp-samples/spring-cloud-gcp-data-spanner-sample/pom.xml
```diff
@@ -18,6 +18,14 @@
 			<groupId>org.springframework.cloud</groupId>
 			<artifactId>spring-cloud-gcp-starter-data-spanner</artifactId>
 		</dependency>
+
+		<!-- Used to test that multiple Spring Data modules can be used together -->
```
meltsufin: Since people will used this sample as a starting point potentially, let's try to keep this out of the pom.xml.
If that means we're not testing for this scenario automatically, that's fine.

spring-cloud-gcp-samples/spring-cloud-gcp-data-spanner-sample/src/test/java/com/example/DatastorePersonRepository.java
```diff
@@ -0,0 +1,34 @@
...
+import org.springframework.cloud.gcp.data.datastore.repository.DatastoreRepository;
+import org.springframework.cloud.gcp.data.datastore.repository.query.Query;
+import org.springframework.data.repository.query.Param;
+
+/**
+ * Sample repository for testing.
+ *
+ * @author Chengyuan Zhao
+ */
+public interface DatastorePersonRepository extends DatastoreRepository<Person, Long> {
```
meltsufin: Again, I would avoid any Datastore deps in the Spanner sample.

spring-cloud-gcp-samples/spring-cloud-gcp-data-spanner-sample/pom.xml
dmitry-s: I think we need a separate sample to demonstrate and test multi-data modules support.

ChengyuanZhao: moved. done.

spring-cloud-gcp-samples/spring-cloud-gcp-data-spanner-sample/src/test/java/com/example/DatastorePersonRepository.java
ChengyuanZhao: done

ChengyuanZhao:
@dmitry-s @meltsufin it's now a separate module. PTAL

spring-cloud-gcp-samples/spring-cloud-gcp-data-spanner-sample/src/test/java/com/example/SpannerTemplateTests.java
```diff
@@ -45,7 +45,7 @@
  */
 @RunWith(SpringRunner.class)
 @ActiveProfiles("test")
-@TestPropertySource("classpath:application.properties")
+@TestPropertySource("classpath:application-test.properties")
```
meltsufin: Why are you modifying the spanner-sample?

spring-cloud-gcp-samples/spring-cloud-gcp-data-spanner-sample/src/test/java/com/example/SpannerRepositoryTests.java
```diff
@@ -52,7 +52,7 @@
  */
 @RunWith(SpringRunner.class)
 @ActiveProfiles("test")
-@TestPropertySource("classpath:application.properties")
+@TestPropertySource("classpath:application-test.properties")
```
meltsufin: Why are you modifying this?

spring-cloud-gcp-samples/spring-cloud-gcp-data-spanner-sample/src/test/resources/application-test.properties
```diff
@@ -0,0 +1,2 @@
+spring.cloud.gcp.spanner.instance-id=spring-demo
```
meltsufin: Ditto.

spring-cloud-gcp-samples/spring-cloud-gcp-data-spanner-sample/src/test/java/com/example/SpannerRepositoryMultiDatabaseTests.java
```diff
@@ -45,7 +45,7 @@
  * @author Chengyuan Zhao
  */
 @RunWith(SpringRunner.class)
-@TestPropertySource("classpath:application.properties")
+@TestPropertySource("classpath:application-test.properties")
```
meltsufin: Ditto.

spring-cloud-gcp-samples/spring-cloud-gcp-data-multi-sample/README.adoc
```diff
@@ -0,0 +1,23 @@
+= Spring Cloud GCP Cloud Cloud Spanner and Datastore Combined Usage Example
```
meltsufin: I'm open to just calling this sample a kitchen-sink sample and not describing it too much in the readme.
We might want to use it for other combinations as well like Logging and Trace which have been problematic in the past too.

spring-cloud-gcp-samples/spring-cloud-gcp-data-spanner-sample/src/test/java/com/example/SpannerTemplateTests.java
ChengyuanZhao: The other samples actually use their own test-properties files like this. I think it should've been this way originally. 

These are test files btw.

spring-cloud-gcp-samples/spring-cloud-gcp-data-spanner-sample/src/test/java/com/example/SpannerRepositoryTests.java
ChengyuanZhao: The other samples actually use their own test-properties files like this. I think it should've been this way originally. 

These are test files btw.

spring-cloud-gcp-samples/spring-cloud-gcp-data-spanner-sample/src/test/resources/application-test.properties
ChengyuanZhao: The other samples actually use their own test-properties files like this. I think it should've been this way originally. 

These are test files btw.

spring-cloud-gcp-samples/spring-cloud-gcp-data-multi-sample/README.adoc
ChengyuanZhao: Any suggestions on the name ?

ChengyuanZhao: I thought a little more and I want to keep this one just about multi-spring-data usage. I think that concept alone needs a clean slate to display, and if we add other junk in here it might be hard to read.

meltsufin: I guess where I'm coming from is that I feel like this is not a very common use-case that needs a sample. The primary reason that we have this sample is to have an integration test. If that's not the case, and we really think it's a useful example for learning how to use the framework, let's leave it this way.

meltsufin: For the name, I would literally call it `spring-cloud-gcp-kitchen-sink-sample`.

ChengyuanZhao: Hmm yea actually Ray talked to me in person about this a while ago. I expressed some surprise he wanted to demo this combined usage on stage and he said it was quite a draw actually. I think if he's using it on stage we ought to have a sample to back it up too.

meltsufin: I would be curious to see his demo. The one I saw he was just switching examples from Datastore to Spanner. Maybe he was forgetting to remove the dependency. @saturnism

spring-cloud-gcp-samples/spring-cloud-gcp-data-multi-sample/src/test/java/com/example/MultipleDataModuleTest.java
```diff
@@ -44,6 +47,14 @@
 	@Autowired
 	PersonRepository datastorePersonRepository;
 
+	@BeforeClass
+	public static void checkToRun() {
+		assumeThat(
+				"Spanner integration tests are disabled. "
+						+ "Please use '-Dit.spanner=true' to enable them. ",
+				System.getProperty("it.spanner"), is("true"));
```
meltsufin: Do you need a second assumption that `it.datastore` is enabled as well?

ChengyuanZhao: yes good point

PR #1727, opened 2019-07-08
Spanner added bounded staleness option and tests

### LINKED ISSUE #1723: Spanner Template :: Is there a way to read Bounded Staleness
In SpannerTemplate is there a way to read records with bounded staleness instead of exact staleness

ChengyuanZhao:
fixes #1723 
added bounded staleness support and tests and ref doc update

### DIFF
--- a/docs/src/main/asciidoc/spanner.adoc
+++ b/docs/src/main/asciidoc/spanner.adoc
@@ -669,6 +669,9 @@ SpannerQueryOptions spannerQueryOptions = new SpannerQueryOptions().setTimestamp
 List<Trade> trades = this.spannerTemplate.query(Trade.class, Statement.of("SELECT * FROM trades"), spannerQueryOptions);
 ----
 
+You can also read with https://cloud.google.com/spanner/docs/timestamp-bounds[bounded staleness] by setting `.setBoundedTimestamp(true)` on the query and read options objects.
+Bounded staleness lets Cloud Spanner choose any point in time later than or equal to the given timestamp, but it cannot be used inside transactions.
+
 
 ===== Read from a secondary index
 
@@ -864,7 +867,7 @@ The final returned value and type of the function is determined by the user.
 You can use this object just as you would a regular `SpannerOperations` with
 a few exceptions:
 
-- Its read functionality cannot perform stale reads, because all reads happen at the single point in time of the transaction.
+- Its read functionality cannot perform stale reads (other than the staleness set on the entire transaction), because all reads happen at the single point in time of the transaction.
 - It cannot perform sub-transactions via `performReadWriteTransaction` or `performReadOnlyTransaction`
 - It cannot perform any write operations.
 
--- a/docs/src/main/asciidoc/spanner.adoc
+++ b/docs/src/main/asciidoc/spanner.adoc
@@ -646,8 +646,8 @@ Main benefit of reads over queries is reading multiple rows of a certain pattern
 ===== Stale read
 
 All reads and queries are *strong reads* by default.
-A *strong read* is a read at a current timestamp and is guaranteed to see all data that has been committed up until the start of this read.
-A *stale read* on the other hand is read at a timestamp in the past.
+A *strong read* is a read at a current time and is guaranteed to see all data that has been committed up until the start of this read.
+An *exact staleness read* is read at a timestamp in the past.
 Cloud Spanner allows you to determine how current the data should be when you read data.
 With `SpannerTemplate` you can specify the `Timestamp` by setting it on `SpannerQueryOptions` or `SpannerReadOptions` to the appropriate read or query methods:
 
@@ -656,7 +656,7 @@ Reads:
 [source,java]
 ----
 // a read with options:
-SpannerReadOptions spannerReadOptions = new SpannerReadOptions().setTimestamp(Timestamp.now());
+SpannerReadOptions spannerReadOptions = new SpannerReadOptions().setTimestamp(myTimestamp);
 List<Trade> trades = this.spannerTemplate.readAll(Trade.class, spannerReadOptions);
 ----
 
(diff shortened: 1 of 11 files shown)


spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerPageableQueryOptions.java
```diff
@@ -68,4 +70,16 @@ public SpannerPageableQueryOptions setAllowPartialRead(boolean allowPartialRead)
 		super.setAllowPartialRead(allowPartialRead);
 		return this;
 	}
+
+	@Override
+	public SpannerPageableQueryOptions setTimestamp(Timestamp timestamp) {
```
dmitry-s: Can we use setTimestampBound instead of setTimestamp and setBoundedTimestamp?
Users could pass TimestampBound.ofMinReadTimestamp(timestamp) or TimestampBound.ofReadTimestamp(timestamp) in that case

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerPageableQueryOptions.java
```diff
@@ -68,4 +70,16 @@ public SpannerPageableQueryOptions setAllowPartialRead(boolean allowPartialRead)
...
+		super.setTimestamp(timestamp);
+		return this;
+	}
+
+	@Override
+	public SpannerPageableQueryOptions setBoundedTimestamp(boolean boundedTimestamp) {
```
meltsufin: I think changing from `boolean` to an actual `BoundedTimestamp` would be more intuitive.

ChengyuanZhao: @dmitry-s @meltsufin  done. PTAL

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerPageableQueryOptions.java
ChengyuanZhao: Done PTAL

docs/src/main/asciidoc/spanner.adoc
```diff
@@ -665,10 +665,13 @@ Queries:
 [source,java]
 ----
 // a query with options:
-SpannerQueryOptions spannerQueryOptions = new SpannerQueryOptions().setTimestamp(Timestamp.now());
+SpannerQueryOptions spannerQueryOptions = new SpannerQueryOptions().setTimestamp(myTimestamp);
 List<Trade> trades = this.spannerTemplate.query(Trade.class, Statement.of("SELECT * FROM trades"), spannerQueryOptions);
 ----
 
+You can also read with https://cloud.google.com/spanner/docs/timestamp-bounds[*bounded staleness*] by setting `.setTimestampBound(TimestampBound.ofMinReadTimestamp(myTimestamp))` on the query and read options objects.
```
dmitry-s: would be good to have code snippets to show different staleness modes

ChengyuanZhao: There are literally just those two. The snippet above shows `.setTimestamp` and this one here inline shows bounded staleness.

ChengyuanZhao: @dmitry-s

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerQueryOptions.java
```diff
@@ -67,13 +68,23 @@ public SpannerQueryOptions setIncludeProperties(Set<String> includeProperties) {
 		return this;
 	}
 
-	public Timestamp getTimestamp() {
```
meltsufin: Seems like a breaking change. Can we just keep both instance variables?
It could be a bit confusing that you set an exact timestamp but the bound is set.

ChengyuanZhao: added it back. The timestamp is inside the timestampBound too, so I am just extracting it from that.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/AbstractSpannerRequestOptions.java
```diff
@@ -0,0 +1,91 @@
...
+
+import com.google.cloud.Timestamp;
+import com.google.cloud.spanner.TimestampBound;
+
+/**
+ * Abstract class of common Read and Query request settings.
+ *
+ * @author Chengyuan Zhao
+ */
+public class AbstractSpannerRequestOptions<A, B extends AbstractSpannerRequestOptions> {
```
meltsufin: I don't think you need this second class parameter `B extends AbstractSpannerRequestOptions`. The cast to `B` is not really safe anyway.
Why not just return `AbstractSpannerRequestOptions`?

ChengyuanZhao: The `B` is for the inheriting children class to provide their own class type. The reason is we return `this` at the end of the setter methods so the user can chain them together. Without the `B` and returning just the abstract class the child classes' setters would return the parent abstract class on the final setter call so the following wouldn't compile because the final setter returns the parent abstract class type and not the child type: 

```
SpannerReadOptions = new SpannerReadOptions.setblah(xyz).setblahblah(abc).setblahblahblah(123);
```
The final setter call still just returns the abstract parent class.

ChengyuanZhao: @meltsufin  the cast is safe here because B must extend the abstract parent class in the class definition and `this` is clearly of that same type.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/AbstractSpannerRequestOptions.java
```diff
@@ -0,0 +1,92 @@
...
+/**
+ * Abstract class of common Read and Query request settings.
+ *
+ * @author Chengyuan Zhao
+ */
+public class AbstractSpannerRequestOptions<A> implements Serializable {
+
+	protected transient List<A> requestOptions = new ArrayList<>();
+
+	protected Class<A> requestOptionType;
```
meltsufin: Does this actually work? This member is never assigned? Would it not be `null`?

ChengyuanZhao: It's assigned in the constructors of the child classes. It's needed to create that array of ReadOption or QueryOption that the client lib requires.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerQueryOptions.java
```diff
@@ -73,4 +73,13 @@ public SpannerQueryOptions setAllowPartialRead(boolean allowPartialRead) {
 		return this;
 	}
 
+	/**
+	 * Deprecated. Please use {@code getOptions}.
```
meltsufin: Can you use `@deprecated` like we do in other parts of the code base?

PR #1728, opened 2019-07-08
Split samples from the project root parent pom.xml

### LINKED ISSUE #1643: Make samples standalone
The samples are in the same repo as the libraries...
If I check out the entire repo it is not clear how do I use a given sample with a given release (HEAD is not released)

The pom refers to the parent sample that refers to the parent...
This would not be the case for real customers apps, and they are struggling to adapt to a correct pom with the correct references to a released version of SpringBoot and the GCP integration.

dzou:
All samples will now refer to parent pom `spring-boot-starter-parent`.

fixes #1643.

### DIFF
--- a/spring-cloud-gcp-kotlin-samples/pom.xml
+++ b/spring-cloud-gcp-kotlin-samples/pom.xml
@@ -2,32 +2,25 @@
 <project xmlns="https://maven.apache.org/POM/4.0.0"
 		xmlns:xsi="https://www.w3.org/2001/XMLSchema-instance"
 		xsi:schemaLocation="https://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
+
 	<parent>
-		<artifactId>spring-cloud-gcp</artifactId>
-		<groupId>org.springframework.cloud</groupId>
-		<version>1.2.0.BUILD-SNAPSHOT</version>
+		<!-- Your own application should inherit from spring-boot-starter-parent -->
+		<groupId>org.springframework.boot</groupId>
+		<artifactId>spring-boot-starter-parent</artifactId>
+		<version>2.2.0.BUILD-SNAPSHOT</version>
+		<relativePath/>
 	</parent>
 
-	<modelVersion>4.0.0</modelVersion>
+	<groupId>org.springframework.cloud</groupId>
 	<artifactId>spring-cloud-gcp-kotlin-samples</artifactId>
+	<version>1.2.0.BUILD-SNAPSHOT</version>
+
+	<modelVersion>4.0.0</modelVersion>
 	<packaging>pom</packaging>
 
 	<modules>
 		<module>spring-cloud-gcp-kotlin-app-sample</module>
 	</modules>
-
-	<!-- The Spring Cloud GCP BOM will manage spring-cloud-gcp version numbers for you. -->
-	<dependencyManagement>
-		<dependencies>
-			<dependency>
-				<groupId>org.springframework.cloud</groupId>
-				<artifactId>spring-cloud-gcp-dependencies</artifactId>
-				<version>${project.version}</version>
-				<type>pom</type>
-				<scope>import</scope>
-			</dependency>
-		</dependencies>
-	</dependencyManagement>
 
 	<build>
 		<sourceDirectory>${project.basedir}/src/main/kotlin</sourceDirectory>
@@ -81,11 +74,4 @@
 		</plugins>
 	</build>
 
-	<dependencies>
-		<dependency>
-			<groupId>org.jetbrains.kotlin</groupId>
-			<artifactId>kotlin-stdlib</artifactId>
-			<version>${kotlin.version}</version>
-		</dependency>
-	</dependencies>
 </project>
--- a/spring-cloud-gcp-kotlin-samples/pom.xml
+++ b/spring-cloud-gcp-kotlin-samples/pom.xml
@@ -22,7 +22,7 @@
 			<dependency>
 				<groupId>org.springframework.cloud</groupId>
 				<artifactId>spring-cloud-gcp-dependencies</artifactId>
-				<version>1.2.0.BUILD-SNAPSHOT</version>
+				<version>${project.version}</version>
 				<type>pom</type>
 				<scope>import</scope>
 			</dependency>
@@ -70,11 +70,6 @@
 						<version>${kotlin.version}</version>
 					</dependency>
 				</dependencies>
-			</plugin>
-
-			<plugin>
-				<groupId>org.springframework.boot</groupId>
-				<artifactId>spring-boot-maven-plugin</artifactId>
 			</plugin>
 
 			<plugin>
--- a/spring-cloud-gcp-kotlin-samples/pom.xml
+++ b/spring-cloud-gcp-kotlin-samples/pom.xml
@@ -16,6 +16,19 @@
 		<module>spring-cloud-gcp-kotlin-app-sample</module>
 	</modules>
 
+	<!-- The Spring Cloud GCP BOM will manage spring-cloud-gcp version numbers for you. -->
+	<dependencyManagement>
+		<dependencies>
+			<dependency>
+				<groupId>org.springframework.cloud</groupId>
+				<artifactId>spring-cloud-gcp-dependencies</artifactId>
+				<version>1.2.0.BUILD-SNAPSHOT</version>
+				<type>pom</type>
+				<scope>import</scope>
+			</dependency>
+		</dependencies>
+	</dependencyManagement>
+
 	<build>
 		<sourceDirectory>${project.basedir}/src/main/kotlin</sourceDirectory>
 		<testSourceDirectory>${project.basedir}/src/test/kotlin</testSourceDirectory>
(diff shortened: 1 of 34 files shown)


dzou:
I think this should be good to go now. You are now able to take any sample and copy its directory to a location outside the project, then run `mvn spring-boot:run`. It should work.

spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-binder-sample/pom.xml
```diff
@@ -25,13 +39,66 @@
             <groupId>org.springframework.boot</groupId>
             <artifactId>spring-boot-starter-web</artifactId>
         </dependency>
+
+        <!-- Test-related dependencies. -->
+        <dependency>
+            <groupId>org.awaitility</groupId>
+            <artifactId>awaitility</artifactId>
+            <version>3.1.6</version>
```
ChengyuanZhao: is there a way to put this 3.1.6 awaitility and 2.5 commons-io test dependencies in the BOM ? they're repeated like a billion times too

dzou: Hmm, i'm not sure if adding it in our bom is appropriate since those deps are only needed in the tests of the samples. In practice users of spring-cloud-gcp may never use those libraries.

ChengyuanZhao: yea that's a good point. I suppose it's just unfortunate then :P

spring-cloud-gcp-kotlin-samples/pom.xml
```diff
@@ -16,6 +16,19 @@
 		<module>spring-cloud-gcp-kotlin-app-sample</module>
 	</modules>
 
+	<!-- The Spring Cloud GCP BOM will manage spring-cloud-gcp version numbers for you. -->
+	<dependencyManagement>
+		<dependencies>
+			<dependency>
+				<groupId>org.springframework.cloud</groupId>
+				<artifactId>spring-cloud-gcp-dependencies</artifactId>
+				<version>1.2.0.BUILD-SNAPSHOT</version>
```
meltsufin: Would `${project.version}` work here instead?

spring-cloud-gcp-samples/pom.xml
```diff
@@ -43,6 +43,18 @@
 		<module>spring-cloud-gcp-data-multi-sample</module>
 	</modules>
 
+	<dependencyManagement>
+		<dependencies>
+			<dependency>
+				<groupId>org.springframework.cloud</groupId>
+				<artifactId>spring-cloud-gcp-dependencies</artifactId>
+				<version>1.2.0.BUILD-SNAPSHOT</version>
```
meltsufin: `${project.version}`?

spring-cloud-gcp-samples/spring-cloud-gcp-config-sample/pom.xml
```diff
@@ -2,16 +2,31 @@
 <project xmlns="https://maven.apache.org/POM/4.0.0"
          xmlns:xsi="https://www.w3.org/2001/XMLSchema-instance"
          xsi:schemaLocation="https://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
+
+    <parent>
+        <groupId>org.springframework.boot</groupId>
+        <artifactId>spring-boot-starter-parent</artifactId>
```
meltsufin: If we use `spring-boot-starter-parent`, don't we lose the ability to enforce checkstyle rules, etc?

spring-cloud-gcp-samples/spring-cloud-gcp-config-sample/pom.xml
```diff
@@ -2,16 +2,31 @@
...
+        <version>2.2.0.BUILD-SNAPSHOT</version>
```
meltsufin: Maybe use a release version of Spring Boot instead?

spring-cloud-gcp-samples/pom.xml
```diff
@@ -43,6 +43,18 @@
...
+				<type>pom</type>
+				<scope>import</scope>
+			</dependency>
+		</dependencies>
+	</dependencyManagement>
+
 	<build>
 		<plugins>
 			<plugin>
```
meltsufin: I don't think you need this plugin here `spring-boot-maven-plugin`.

spring-cloud-gcp-samples/spring-cloud-gcp-config-sample/pom.xml
dzou: Looks like the version used in the project currently is `2.2.0.M4`. So should we use that instead? Or is snapshot good for helping them test their project?

spring-cloud-gcp-samples/pom.xml
dzou: Done.

spring-cloud-gcp-samples/pom.xml
dzou: Done.

spring-cloud-gcp-kotlin-samples/pom.xml
dzou: Done.

spring-cloud-gcp-samples/spring-cloud-gcp-config-sample/pom.xml
dzou: Yes unfortunately checkstyle is lost. Seems like this is the price for disconnecting from the parent.

dzou:
Ok I think we should be good now - I've put the checkstyle configuration in the parent-sample and all children samples correctly inherit from it: https://github.com/spring-cloud/spring-cloud-gcp/pull/1728/files#diff-50e7c6073821905e7b9c6c95f86067a1

The reason why I got stuck was because I realized we forgot to add a groupId and version on the samples parent, so the children samples were never inheriting correctly from the parent sample. But it should be good now; checkstyle works and users can also just copy a sample outside of the project to run it by changing its parent to `spring-boot-starter-parent`.

spring-cloud-gcp-kotlin-samples/pom.xml
```diff
@@ -16,6 +16,19 @@
 		<module>spring-cloud-gcp-kotlin-app-sample</module>
 	</modules>
 
+	<!-- The Spring Cloud GCP BOM will manage spring-cloud-gcp version numbers for you. -->
```
meltsufin: 10 lines up, do you want to add the same comment about
```xml
 <!-- Your own application should inherit from spring-boot-starter-parent -->
```
?

spring-cloud-gcp-samples/spring-cloud-gcp-integration-pubsub-json-sample/pom.xml
```diff
@@ -25,6 +40,32 @@
...
+			<artifactId>awaitility</artifactId>
+			<version>3.1.6</version>
+			<scope>test</scope>
+		</dependency>
+		<dependency>
+			<groupId>junit</groupId>
+			<artifactId>junit</artifactId>
+			<scope>test</scope>
+		</dependency>
+
```
meltsufin: Blank lines?

spring-cloud-gcp-samples/spring-cloud-gcp-logging-sample/pom.xml
```diff
@@ -24,6 +38,32 @@
...
+            <artifactId>awaitility</artifactId>
+            <version>3.1.6</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
+            <scope>test</scope>
+        </dependency>
+
```
meltsufin: Blank lines?

spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-binder-sample/pom.xml
```diff
@@ -25,13 +39,34 @@
...
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
+            <scope>test</scope>
+        </dependency>
+
```
meltsufin: Blank lines?

spring-cloud-gcp-kotlin-samples/pom.xml
dzou: Done.

spring-cloud-gcp-samples/spring-cloud-gcp-integration-pubsub-json-sample/pom.xml
dzou: Done.

spring-cloud-gcp-samples/spring-cloud-gcp-logging-sample/pom.xml
dzou: Done.

spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-binder-sample/pom.xml
dzou: Done.

PR #1731, opened 2019-07-09
Make Vision OCR configuration conditional

### LINKED ISSUE #1729: Caused by: java.lang.ClassNotFoundException: com.google.cloud.storage.Storage
I have to reopen this issue, I am not sure if Chengyuan Zhao can see my latest message in the closed issue, so, I will repeat the same messages.

I have the same problem. I have just compiled full project doing clean install in root folder, with a build success, I tried with spring-cloud-gcp-vision-api-sample project, provisioning with a json file from gcp, and then springboot-run phase.
I am getting the same error than Abhinav1990

[stacktrace](https://pastebin.com/p7E1skC4)

[Storage class definition](https://github.com/googleapis/google-cloud-java/blob/master/google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/Storage.java)

[dependency tree output](https://pastebin.com/12nkYw2G)

dzou:
Modifies the `CloudVisionAutoConfiguration` to include the `Storage` class used by the `DocumentOcrTemplate` only if the document OCR feature is used.

Fixes #1729

### DIFF
--- a/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/vision/CloudVisionAutoConfiguration.java
+++ b/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/vision/CloudVisionAutoConfiguration.java
@@ -108,7 +108,6 @@ public class CloudVisionAutoConfiguration {
 	/**
 	 * Configures the {@link DocumentOcrTemplate} class.
 	 */
-	@EnableConfigurationProperties(CloudVisionProperties.class)
 	@ConditionalOnClass(Storage.class)
 	@ConditionalOnMissingBean(DocumentOcrTemplate.class)
 	static class VisionOcrConfiguration {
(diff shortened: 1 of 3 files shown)


spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/vision/CloudVisionAutoConfiguration.java
```diff
@@ -117,4 +104,32 @@ public Executor cloudVisionExecutor() {
 		ackExecutor.setThreadNamePrefix("gcp-cloud-vision-ocr-executor");
 		return ackExecutor;
 	}
+
+	/**
+	 * Configures the {@link DocumentOcrTemplate} class.
+	 */
+	@EnableConfigurationProperties(CloudVisionProperties.class)
```
meltsufin: I think it's enough that this annotation exists on the parent class.

dzou: Done.

PR #1740, opened 2019-07-17
Added test to verify behaviour of ZipkinBackwardsCompatibilityAutoConfiguration

TYsewyn:
See #1705

PR #1744, opened 2019-07-17
Backported #1383

### LINKED ISSUE #1705: Spring Cloud Sleuth GCP: every time spans are sent, they are sent to localhost as well
There should be a way to turn off the default ZipkinRestTemplateSenderConfiguration when using Spring Cloud GCP. I would prefer if it was turned off by default.

Currently, every time spans are being sent, a completely unnecessary call to `http://localhost:9411/api/v2/spans`, and it fails 100% of the time because there is nothing there to answer.

The only reason I saw it was because I used `logging.level.org.springframework.cloud=TRACE`.

I tried to fix it by setting `spring.zipkin.enabled=false` but it crashes on startup because of `ZipkinBackwardsCompatibilityAutoConfiguration`.

If there already is a way to turn this off I would be very happy. Anyway my opinion is it should be off by default because I have already chosen `spring.cloud.gcp.trace.enabled=true`.

Originally posted as https://github.com/spring-cloud/spring-cloud-sleuth/issues/1377 but probably better to originate a fix in this repo instead.

Thanks.

TYsewyn:
Fixes #1705, closes #1740 and resolves spring-cloud/spring-cloud-sleuth#1400


spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/trace/ZipkinProperties.java
```diff
@@ -0,0 +1,45 @@
...
+/**
+ * Zipkin Properties.
+ *
+ * @author Tim Ysewyn
+ * @deprecated Replaced by {@link GcpTraceProperties} since 1.2.
+ * Added for backward & forward compatibility.
+ */
+@Deprecated
+@ConfigurationProperties("spring.zipkin")
+public class ZipkinProperties {
```
meltsufin: Why do we need this? It's not part of the original PR.
Also, I don't see where it would be used.
`GcpTraceProperties` already exists in 1.1.

TYsewyn:
Nice catch!

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/trace/ZipkinProperties.java
TYsewyn: Yes, but the property `spring.cloud.gcp.trace.messageTimeout` doesn't exist in 1.1.
Previously `spring.zipkin.messageTimeout` was used to change the timeout.
Since the property and reporter were both part of the zipkin dependency we need to have a way to let the existing configurations work with the new setup since the next service release should be a drop-in replacement.

That's also why I marked it as deprecated and already added the "new" `spring.cloud.gcp.trace.messageTimeout` property to this PR, to be able to move to 1.2.x without any issue in the future.

`spring.cloud.gcp.trace.messageTimeout` takes precedence over `spring.zipkin.messageTimeout` if you configure both.

TYsewyn: It's used here: https://github.com/spring-cloud/spring-cloud-gcp/pull/1744/files#diff-fc1b9a6100070e7eb2d92338e77686e5R140

meltsufin: I think I understand what you're saying. However, I would avoid introducing a properties class with a namespace that we do not own (`spring.zipkin`). You can instead auto-wire the `Environment` and read the value of `spring.zipkin.messageTimeout` directly from the environment.

TYsewyn:
You are very welcome! Don't forget to close #1705. I - again - used to wrong identifier in the PR message. 😅

PR #1748, opened 2019-07-18
support key fields in name-based query methods

### LINKED ISSUE #1726: Support find Id fields in named query methods
follow up for #1720

dmitry-s:
fixes #1726

### DIFF
--- a/docs/src/main/asciidoc/datastore.adoc
+++ b/docs/src/main/asciidoc/datastore.adoc
@@ -1080,12 +1080,6 @@ List<Key> getKeys();
 
 @Query(value = "SELECT __key__ from test_entities_ci limit 1")
 Key getKey();
-
-@Query("SELECT id FROM test_entities_ci WHERE size <= @size")
-List<String> getIds(@Param("size") long size);
-
-@Query("SELECT id FROM test_entities_ci WHERE size <= @size limit 1")
-String getOneId(@Param("size") long size);
 ----
 
 In order to use `@Id` annotated fields in custom queries, use `+++__key__+++` keyword for the field name. The parameter type should be of `Key`, as in the following example.
--- a/docs/src/main/asciidoc/datastore.adoc
+++ b/docs/src/main/asciidoc/datastore.adoc
@@ -740,9 +740,7 @@ These methods, respectively, allow querying for:
 
 ==== Find by ID(s)
 
-Datstore reading a single entity or multiple entities in a kind.
-
-Using `DatastoreTemplate` you can execute reads, for example:
+Using `DatastoreTemplate` you can find entities by id. For example:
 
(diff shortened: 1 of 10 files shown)


spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/PartTreeDatastoreQuery.java
```diff
@@ -354,6 +344,19 @@ private void applySelectWithFilter(Object[] parameters, Builder builder) {
 						: filters[0]);
 	}
 
+	private Value convertValue(Iterator it, Part part, DatastorePersistentProperty persistentProperty) {
```
meltsufin: It's a bit odd to pass in an iterator here rather than the actual value.
Also the check for `IS_NULL` is completely redundant since you do it in the switch statement.

spring-cloud-gcp-data-datastore/src/test/java/org/springframework/cloud/gcp/data/datastore/it/TestEntityRepository.java
```diff
@@ -58,6 +58,9 @@
 
 	int countBySize(long size);
 
+	@Nullable
+	List<TestEntity> getByColorAndIdGreaterThanEqualOrderById(String color, Long id);
```
ChengyuanZhao: Could a test where the id is already a full `Key` with ancestors in a test method too?

Also, I think I might have added in the refdoc that ID properties aren't allowed. We could delete that now.

ChengyuanZhao: Also in the ref doc I think it should be mentioned that if the user is using `@Query` then he needs to be writing `__key__` himself and that he must use `Key` type params and construct those keys himself even if the properties are long or string.

docs/src/main/asciidoc/datastore.adoc
```diff
@@ -1072,24 +1068,39 @@ With the exception of `Cursor`, array forms of each of the types are also suppor
...
-	List<Key> getKeys();
+@Query(value = "SELECT __key__ from test_entities_ci")
+List<Key> getKeys();
+
+@Query(value = "SELECT __key__ from test_entities_ci limit 1")
+Key getKey();
 
-	@Query(value = "SELECT __key__ from test_entities_ci limit 1")
-	Key getKey();
+@Query("SELECT id FROM test_entities_ci WHERE size <= @size")
```
meltsufin: Does this still work when we don't have the id column?

PR #1750, opened 2019-07-19
Add null check for trace probability property

### LINKED ISSUE #1749: Assert probability not null before instantiating ProbabilityBasedSampler.
(surfaced when looking into #1482)
Bringing in the trace starter without any configuration results in an unintuitive NPE when Sleuth's `ProbabilityBasedSampler` tries to autobox a null probability*. 

A null check is needed before instantiation in spring-cloud-gcp, and ideally `ProbabilityBasedSampler` constructor should check, too.

\* now I want to write a scifi short story named "Null Probability".

elefeint:
Adds assert before creating a default `ProbabilityBasedSampler`.

Fixes #1749.


elefeint:
Rendered obsolete by #1774.

Issue #1725, opened 2019-07-06
Gradle build issue with Greenwich.SR2 and spring-cloud-gcp-starter-trace


RobMaskell:
I had an issue with upgrading my main project so I tried to figure our what was wrong starting with a vanilla spring cloud gateway project from start.spring.io and bumped the spring cloud version up to `greenwich.SR2` all ok.
```
plugins {
	id 'org.springframework.boot' version '2.1.6.RELEASE'
	id 'java'
}

apply plugin: 'io.spring.dependency-management'

group = 'com.example'
version = '0.0.1-SNAPSHOT'
sourceCompatibility = '11'

repositories {
	mavenCentral()
}

ext {
	set('springCloudVersion', "Greenwich.SR2")
}

dependencies {
	implementation 'org.springframework.cloud:spring-cloud-starter-gateway'
	testImplementation 'org.springframework.boot:spring-boot-starter-test'
}

dependencyManagement {
	imports {
		mavenBom "org.springframework.cloud:spring-cloud-dependencies:${springCloudVersion}"
	}
}
```
when I add the following
```
implementation("org.springframework.cloud:spring-cloud-gcp-starter-trace")
```
it fails to build as I can't find `zipkin-tests.jar`
```
Warning:<i><b>root project 'demo': Unable to resolve additional project configuration.</b>
Details: org.gradle.api.artifacts.ResolveException: Could not resolve all dependencies for configuration ':compileClasspath'.
Caused by: org.gradle.internal.resolve.ArtifactNotFoundException: Could not find zipkin-tests.jar (io.zipkin.zipkin2:zipkin:2.14.2).
Searched in the following locations:
    https://repo.maven.apache.org/maven2/io/zipkin/zipkin2/zipkin/2.14.2/zipkin-2.14.2-tests.jar</i>
```
Flipping back to `greenwich.SR1` and it builds fine again.

dzou:
Looks like zipkin-tests is present on Maven Central for [Zipkin versions 2.12 and prior](https://repo.maven.apache.org/maven2/io/zipkin/zipkin2/zipkin/2.12.9/), however for [Zipkin version 2.14](https://repo.maven.apache.org/maven2/io/zipkin/zipkin2/zipkin/2.14.1/) it is no longer present. It looks like it is failing to find that jar for 2.14.

@meltsufin - Any ideas where the zipkin version numbers/locations are being controlled?

meltsufin:
[...]

PR #1756, opened 2019-07-23
Exclude io.zipkin.zipkin2:zipkin test-jar

meltsufin:
Fixes: #1725.

### DIFF
--- a/spring-cloud-gcp-autoconfigure/pom.xml
+++ b/spring-cloud-gcp-autoconfigure/pom.xml
@@ -187,6 +187,10 @@
                     <groupId>io.grpc</groupId>
                     <artifactId>grpc-core</artifactId>
                 </exclusion>
+                <exclusion>
+                    <groupId>io.zipkin.zipkin2</groupId>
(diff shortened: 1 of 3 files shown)


spring-cloud-gcp-autoconfigure/pom.xml
```diff
@@ -187,6 +187,10 @@
                     <groupId>io.grpc</groupId>
                     <artifactId>grpc-core</artifactId>
                 </exclusion>
+                <exclusion>
+                    <groupId>io.zipkin.zipkin2</groupId>
+                    <artifactId>zipkin</artifactId>
```
artembilan: Huh? Isn't that enough to just exclude on the `dependencyManagement` level in the root pom?

meltsufin: I thought that didn't always work. It might be dependent on version of Maven. I'll give it a try.

meltsufin: I think you're right! I dunno why we have the exclusions this way.

PR #1757, opened 2019-07-23
Exclude io.zipkin.zipkin2:zipkin test-jar (1.1.x)

meltsufin:
Fixes: #1725.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -95,6 +95,12 @@
 				<groupId>io.zipkin.gcp</groupId>
 				<artifactId>zipkin-sender-stackdriver</artifactId>
 				<version>${zipkin-gcp.version}</version>
+				<exclusions>
+					<exclusion>
+						<groupId>io.zipkin.zipkin2</groupId>
+						<artifactId>zipkin</artifactId>
+					</exclusion>
+				</exclusions>
 			</dependency>
 
 			<dependency>
--- a/spring-cloud-gcp-autoconfigure/pom.xml
+++ b/spring-cloud-gcp-autoconfigure/pom.xml
@@ -172,10 +172,6 @@
                     <groupId>io.grpc</groupId>
                     <artifactId>grpc-core</artifactId>
                 </exclusion>
-                <exclusion>
-                    <groupId>io.zipkin.zipkin2</groupId>
-                    <artifactId>zipkin</artifactId>
-                </exclusion>
             </exclusions>
         </dependency>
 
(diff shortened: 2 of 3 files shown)


pom.xml
```diff
@@ -95,6 +95,12 @@
 				<groupId>io.zipkin.gcp</groupId>
 				<artifactId>zipkin-sender-stackdriver</artifactId>
 				<version>${zipkin-gcp.version}</version>
+				<exclusions>
+					<exclusion>
+						<groupId>io.zipkin.zipkin2</groupId>
+						<artifactId>zipkin</artifactId>
+					</exclusion>
```
dzou: Do you also need to add the `<type>` here? Like `<type>test-jar</type>`?

meltsufin: Doesn't look like it.

dzou: Isn't there a core artifact also named 'zipkin' though? https://repo.maven.apache.org/maven2/io/zipkin/zipkin2/zipkin/2.12.4/zipkin-2.12.4.pom

How does it disambiguate between the 'zipkin' core and the zipkin test jar?

meltsufin: `<type>` is not allowed under `dependencyManagement`.
In any case, `zipkin-sender-stackdriver` doesn't depend on `zipkin` core.
So, we should be OK. Other deps will bring it in.

meltsufin:
@artembilan In this branch, I'm not sure exactly why, but it's not enough to exclude it in dependency management. It's still being pulled in.

artembilan:
>  It's still being pulled in.

Can you print `dependency:tree` to be sure who brings that dependency for us?

meltsufin:
I think I have to go back to excluding it everywhere. It's actually the case on both branches. My initial test must've been incorrect.
```
[INFO] |  +- io.zipkin.gcp:zipkin-sender-stackdriver:jar:0.9.0:compile
[INFO] |  |  +- io.zipkin.gcp:zipkin-translation-stackdriver:jar:0.9.0:compile
[INFO] |  |  +- io.zipkin.reporter2:zipkin-reporter:jar:2.7.13:compile
[INFO] |  |  +- com.google.api.grpc:grpc-google-devtools-cloudtrace-v1:jar:0.1.1:compile
[INFO] |  |  |  \- com.google.api.grpc:grpc-google-common-protos:jar:1.16.0:compile
[INFO] |  |  +- io.zipkin.zipkin2:zipkin:test-jar:tests:2.11.11:compile
```

PR #1762, opened 2019-07-23
Datastore multi namespace

### LINKED ISSUE #1289: Datastore , find some dynamic way to set namespace to support multi-tenant
I was speaking with Ray and he explained a use case based on REST repositories where the namespace (perhaps each customer is a tenant/namespace) is part of the request. 

Other than just for the REST example there could be other use cases. I've had at least one large user on Spanner asking for setting the namespace in a way that's more dynamic than just in application.properties. In spanner there isn't a namespace to set at all, but here at least there is one, so perhaps we should find a way to do it.

ChengyuanZhao:
fixes #1289 
adds multi-tenant support with namespace provider.


spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/datastore/GcpDatastoreAutoConfiguration.java
```diff
@@ -152,8 +162,16 @@ public DatastoreEntityConverter datastoreEntityConverter(DatastoreMappingContext
...
+	public DatastoreTemplate datastoreTemplate(Supplier<? extends DatastoreReaderWriter> datastore,
+			DatastoreMappingContext datastoreMappingContext,
 			DatastoreEntityConverter datastoreEntityConverter, ObjectToKeyFactory objectToKeyFactory) {
 		return new DatastoreTemplate(datastore, datastoreEntityConverter, datastoreMappingContext, objectToKeyFactory);
 	}
+
+	/**
+	 * This interface is the return type for the bean that provides namespaces.
+	 */
+	public interface NamespaceProvider extends Supplier<String> {
```
meltsufin: Maybe rename to `DatastoreNamespaceProvider` for clarity?

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/datastore/GcpDatastoreAutoConfiguration.java
```diff
@@ -102,20 +105,27 @@
...
-			builder.setHost(this.host);
-		}
+	public NamespaceProvider namespaceProvider() {
+		return () -> this.namespace;
+	}
 
-		return builder.build().getService();
+	@Bean
+	@ConditionalOnMissingBean(value = Datastore.class, parameterizedContainer = Supplier.class)
+	public Supplier<Datastore> datastore(Supplier<String> namespaceProvider) {
```
meltsufin: I think you need to use `NamespaceProvider` here instead of `Supplier<String>` to avoid ambiguity with other `Supplier<String>` beans that might be in the application context.
In fact, is there anything wrong with using `NamespaceProvider` everywhere instead of `Supplier<String>`?

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/datastore/GcpDatastoreAutoConfigurationTests.java
```diff
@@ -107,6 +111,12 @@ public void testDatastoreHealthIndicatorNotCreated() {
 				.isThrownBy(() -> context.getBean(DatastoreHealthIndicator.class)));
 	}
 
+	private Datastore getDatastoreBean(ApplicationContext context) {
```
meltsufin: Maybe this method should be in `DatastoreTemplate` for convenience, or is it too risky?

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/datastore/GcpDatastoreAutoConfiguration.java
```diff
@@ -102,20 +105,27 @@
 
 	@Bean
 	@ConditionalOnMissingBean
-	public Datastore datastore() {
```
meltsufin: I'm worried that we breaking backwards compatibility because users won't be able to simply autowire a `Datastore` bean anymore. Can we conditionally create a `Datastore` bean if the default namespace provider is used?

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/datastore/GcpDatastoreAutoConfigurationTests.java
ChengyuanZhao: This is just for tests actually (this is a test class)

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/datastore/GcpDatastoreAutoConfiguration.java
ChengyuanZhao: Yes good catch. I made that new type and forgot to come back and replace it here.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/datastore/GcpDatastoreAutoConfiguration.java
ChengyuanZhao: My thinking here was the user could append `() -> ` to the front of the `Datastore`  custom bean they had previously. 

But to your point, I think we could still accept a `Datastore` bean. But I don't think we can make any assumptions about the namespace provider since the user may totally ignore it or use it in his own Datastore object.

ChengyuanZhao: Made change and added test. PTAL @meltsufin

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/datastore/GcpDatastoreAutoConfiguration.java
ChengyuanZhao: changed

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/datastore/GcpDatastoreAutoConfigurationTests.java
meltsufin: Yeah, I'm just suggesting to add this utility method to `DatastoreTemplate`.

ChengyuanZhao: Oh... For what purpose?

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/datastore/GcpDatastoreAutoConfiguration.java
```diff
@@ -102,20 +106,31 @@
...
-		DatastoreOptions.Builder builder = DatastoreOptions.newBuilder()
-				.setProjectId(this.projectId)
-				.setHeaderProvider(new UserAgentHeaderProvider(this.getClass()))
-				.setCredentials(this.credentials);
-		if (this.namespace != null) {
-			builder.setNamespace(this.namespace);
-		}
+	public DatastoreNamespaceProvider namespaceProvider() {
```
meltsufin: Do we even need this default bean?

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/datastore/GcpDatastoreAutoConfiguration.java
```diff
@@ -102,20 +106,31 @@
...
+		return () -> this.namespace;
+	}
 
-		if (this.host != null) {
-			builder.setHost(this.host);
+	@Bean
+	@ConditionalOnMissingBean(value = Datastore.class, parameterizedContainer = Supplier.class)
```
meltsufin: Can we make this `@ConditionalOnBean` `namespaceProvider`?

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/datastore/GcpDatastoreAutoConfiguration.java
```diff
@@ -102,20 +106,31 @@
...
+	public Supplier<Datastore> datastoreSupplier(DatastoreNamespaceProvider namespaceProvider,
+			Optional<Datastore> datastore) {
```
meltsufin: I don't think `Optional` works the way you expect it here.
I think you need to use `org.springframework.beans.factory.ObjectProvider` instead.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/datastore/health/DatastoreHealthIndicator.java
```diff
@@ -32,28 +34,29 @@
...
  * @author Mike Eltsufin
+ * @author Chengyuan Zhao
  *
  * @since 1.2
  */
 @Component
 public class DatastoreHealthIndicator extends AbstractHealthIndicator {
 
-	private final Datastore datastore;
+	private final Supplier<Datastore> datastore;
```
meltsufin: This is a bit unfortunate because we will probably break other users' code in a similar way.

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/datastore/GcpDatastoreAutoConfigurationTests.java
meltsufin: To avoid dealing with the provider.

ChengyuanZhao: I think it's ok. It's hard to misuse the supplier because it is of the wrong type unless you call .get() on it, which is the only method it has.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/datastore/GcpDatastoreAutoConfiguration.java
ChengyuanZhao: I don't think I'm following. Could you explain what you would put instead?

meltsufin: Nothing, just deal with the situation when `DatastoreNamespaceProvider` is not present.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/datastore/GcpDatastoreAutoConfiguration.java
ChengyuanZhao: If we did that wouldn't that mean there would be an error if the user also tried to define a bean of `Supplier<Datastore>` return type ?

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/datastore/GcpDatastoreAutoConfiguration.java
ChengyuanZhao: Yes I can change, but I saw this post from STephan https://stackoverflow.com/questions/49044770/change-in-how-spring-5-handles-null-beans

He seems to be saying both can be used the same. I have some tests do confirm it's working as expected with Optional.

ChengyuanZhao: Just curious, what's the difference between the two?

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/datastore/health/DatastoreHealthIndicator.java
ChengyuanZhao: If they were expecting to get a bean of Datastore directly, then yes we will break them :(

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/datastore/GcpDatastoreAutoConfiguration.java
```diff
@@ -101,21 +105,28 @@
 	}
 
 	@Bean
-	@ConditionalOnMissingBean
-	public Datastore datastore() {
```
meltsufin: Can we keep this been conditional on `DatastoreNamespaceProvider` missing and `Supplier<Datastore>` missing?

.travis.yml
```diff
@@ -3,6 +3,7 @@ git:
   depth: 9999999
 jdk:
   - oraclejdk8
+dist: trusty
```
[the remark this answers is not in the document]
ChengyuanZhao: Yes I searched online and it seems the "dist" if you don't specify it is now "xenial" (or something spelled similar). People started running into the problem because that "dist" didn't work with oraclejdk8. People either  tried openjdk8 or said they changed the dist to "trusty".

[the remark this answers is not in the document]
ChengyuanZhao: image

docs/src/main/asciidoc/datastore.adoc
```diff
@@ -1299,6 +1299,23 @@ class SpringSecurityAuditorAware implements AuditorAware<User> {
...
+
+[source,java]
+----
+    @Bean
+    public NamespaceProvider namespaceProvider() {
+        // return custom Supplier of a namespace string.
+    }
+----
+
+The `NamespaceProvider` is a synonym for `Supplier<String>`.
```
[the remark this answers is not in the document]
ChengyuanZhao: fixed (in the code sample above too)

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/datastore/GcpDatastoreAutoConfiguration.java
```diff
@@ -152,8 +163,16 @@ public DatastoreEntityConverter datastoreEntityConverter(DatastoreMappingContext
...
+	public DatastoreTemplate datastoreTemplate(Supplier<? extends DatastoreReaderWriter> datastore,
+			DatastoreMappingContext datastoreMappingContext,
 			DatastoreEntityConverter datastoreEntityConverter, ObjectToKeyFactory objectToKeyFactory) {
 		return new DatastoreTemplate(datastore, datastoreEntityConverter, datastoreMappingContext, objectToKeyFactory);
 	}
+
+	/**
+	 * This interface is the return type for the bean that provides namespaces.
+	 */
+	public interface DatastoreNamespaceProvider extends Supplier<String> {
```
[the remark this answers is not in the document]
ChengyuanZhao: I felt it could go either way. changed.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/datastore/GcpDatastoreAutoConfiguration.java
```diff
@@ -101,21 +105,28 @@
...
-		if (this.namespace != null) {
-			builder.setNamespace(this.namespace);
-		}
-
-		if (this.host != null) {
-			builder.setHost(this.host);
+	@ConditionalOnMissingBean(value = Datastore.class, parameterizedContainer = Supplier.class)
+	public Supplier<Datastore> datastoreSupplier(ObjectProvider<DatastoreNamespaceProvider> namespaceProvider,
+			ObjectProvider<Datastore> datastore) {
+		if (datastore.getIfAvailable() != null) {
```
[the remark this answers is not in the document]
ChengyuanZhao: Would be nice right? but there is no such `isAvailable` function.

Also there does NOT need to be an exception if neither are configured. That's a totally valid usecase and is probably the most common: using Cloud Datastore's default namespace.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/datastore/GcpDatastoreAutoConfiguration.java
ChengyuanZhao: Based on the highlighting on github I'm not clear on what line you're referring to. 

There is no longer a bean for `Datastore` at all.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/datastore/GcpDatastoreAutoConfiguration.java
```diff
@@ -101,21 +106,34 @@
...
-				.setCredentials(this.credentials);
-		if (this.namespace != null) {
-			builder.setNamespace(this.namespace);
-		}
-
-		if (this.host != null) {
-			builder.setHost(this.host);
+	@ConditionalOnMissingBean(value = Datastore.class, parameterizedContainer = Supplier.class)
+	public Supplier<Datastore> datastoreSupplier(
+			ObjectProvider<org.springframework.cloud.gcp.autoconfigure.datastore.DatastoreNamespaceProvider> namespaceProvider,
```
dmitry-s: why the fully qualified name? is there another DatastoreNamespaceProvider?

ChengyuanZhao: fixed!

.travis.yml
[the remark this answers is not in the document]
meltsufin: Got it.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/datastore/GcpDatastoreAutoConfiguration.java
meltsufin: Why? Could it still be also conditional on `Supplier<Datastore>` missing?

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/datastore/GcpDatastoreAutoConfiguration.java
meltsufin: I'm suggesting to bring it back.

ChengyuanZhao: I don't understand.. why bring it back if we're not gonna return a real client in there? Isn't this the same exact situation as the default namespace provider?

ChengyuanZhao: I'm trying to get this to work but I'm running into this issue: 

Because the `Supplier<Datastore>` bean has a `Datastore` param, it causes whatever `Datastore` bean I make to be picked up. The problem is if the user defines his own NamespaceProvider, then I must throw an exception if he tries to request a Datastore instance (since it wouldn't use his namespaceprovider). However that exception gets thrown because the param of the `Supplier<Datastore>` causes that `Datastore` bean to get created. 

I also don't think there's a way to remove the `Datastore` param from the `Supplier<Datastore>` bean because we need to use it if the user defined a `Datastore` himself.

ChengyuanZhao: @meltsufin

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/util/CachingDatastoreProvider.java
```diff
@@ -0,0 +1,57 @@
...
+	private final Supplier<A> inputProvider;
+
+	private final Function<A, Datastore> producer;
+
+	/**
+	 * Constructor.
+	 * @param inputProvider the provider that gives inputs for each Datstore client .
+	 * @param producer the function that returns products of this provider given inputs.
+	 */
+	public CachingDatastoreProvider(Supplier<A> inputProvider, Function<A, Datastore> producer) {
```
meltsufin: I think this is too generic. Maybe just pass in everything you need to build a Datastore client: namespace provider, properties, etc.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/datastore/GcpDatastoreAutoConfiguration.java
```diff
@@ -101,21 +106,34 @@
...
-		return builder.build().getService();
+		return new CachingDatastoreProvider<>(
+				namespaceProvider.getIfAvailable() == null ? () -> this.namespace : namespaceProvider.getIfAvailable(),
+				namespace -> {
+			DatastoreOptions.Builder builder = DatastoreOptions.newBuilder()
+					.setProjectId(this.projectId)
+					.setHeaderProvider(new UserAgentHeaderProvider(this.getClass()))
+					.setCredentials(this.credentials);
+			if (namespace != null) {
+				builder.setNamespace(namespace);
```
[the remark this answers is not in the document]
ChengyuanZhao: I don't think so. Originally we allowed them to provide a Datastore client all on their own and that could completely ignore the properties. In fact we do this in every single module.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/datastore/GcpDatastoreAutoConfiguration.java
```diff
@@ -101,21 +106,34 @@
...
+								"Only one can be configured.");
+			});
+			return () -> datastore.getIfAvailable();
 		}
-
-		return builder.build().getService();
+		return new CachingDatastoreProvider<>(
+				namespaceProvider.getIfAvailable() == null ? () -> this.namespace : namespaceProvider.getIfAvailable(),
+				namespace -> {
+			DatastoreOptions.Builder builder = DatastoreOptions.newBuilder()
```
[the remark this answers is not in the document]
ChengyuanZhao: Not exactly true. This logic here *only deals with utilizing a custom namespace* whereas the `CachingDatastoreProvider` can be used with any combination of input settings. What nails `CachingDatastoreProvider` to Datastore is the `Datastore` type.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/util/CachingDatastoreProvider.java
ChengyuanZhao: Should I just name this `CachingDatastoreNamespaceProvider`? That logic that you mentioned above only deals with custom namespaces and not any other setting that could go into a Datastore client.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/datastore/GcpDatastoreAutoConfiguration.java
ChengyuanZhao: @meltsufin I don't follow. The issue isn't the return type of `Supplier<Datastore>`. The issue is that one of the params for that bean is a `Datastore`.

docs/src/main/asciidoc/datastore.adoc
```diff
@@ -1299,6 +1299,23 @@ class SpringSecurityAuditorAware implements AuditorAware<User> {
 
 You can also set a custom provider for properties annotated `@LastModifiedDate` by providing a bean for `DateTimeProvider` and providing the bean name to `@EnableDatastoreAuditing(dateTimeProviderRef = "customDateTimeProviderBean")`.
 
+=== Partitioning Data by Namespace
+
+You can https://cloud.google.com/datastore/docs/concepts/multitenancy[partition your data by using more than one namespace].
+This is the recommended method for multi-tenancy in Cloud Datastore.
```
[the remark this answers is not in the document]
ChengyuanZhao: Yes good point. Done.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/datastore/GcpDatastoreAutoConfiguration.java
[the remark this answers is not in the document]
ChengyuanZhao: I gave this try and there is a sticking point.

there are references to `this.projectid, this.credentials, this.host`. These three things felt strange to pass into the `CachingDatastoreProvider` since it lives in the Spring Data Datastore module instead of in autoconfig here. 

I don't want to  move `CachingDatastoreProvider` to the auto-config module because its behavior is something I want to verify in DatastoreTemplateTests (I have a unit test there that makes sure the correct clients are being used by template operations).

meltsufin:
Maybe something like this?
https://github.com/spring-cloud/spring-cloud-gcp/commit/76a313c3338353b174c710069c5a53ef6e3a39bb

ChengyuanZhao:
@meltsufin thank you for that hint! I didn't know `@ConditionalOnMissingBean` could actually take an array of other types (i thought it only meant the single return type to which it's annotated). 

I think it all works as expected now!

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/datastore/DatastoreTransactionManagerAutoConfiguration.java
```diff
@@ -47,11 +49,11 @@
 	@Configuration
 	static class DatastoreTransactionManagerConfiguration {
 
-		private final Datastore datastore;
+		private final Supplier<Datastore> datastore;
```
meltsufin: Since we now have `DatastoreProvider`, should we just use it everywhere in place of `Supplier<Datastore>`?

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/datastore/GcpDatastoreAutoConfiguration.java
```diff
@@ -102,20 +107,25 @@
...
+		if (datastoreProvider.getIfAvailable() != null) {
+			namespaceProvider.ifAvailable(unused -> {
+				throw new DatastoreDataException(
+						"A Datastore namespace provider and Datastore client were both configured. " +
+								"Only one can be configured.");
+			});
+			return datastoreProvider::getIfAvailable;
 		}
+		return getDatastoreProvider(
+				namespaceProvider.getIfAvailable() == null ? () -> this.namespace : namespaceProvider.getIfAvailable());
```
meltsufin: I think you can simplify this line, if you change the order of the beans like this:

```java
	@Bean
	@ConditionalOnMissingBean({ Datastore.class, DatastoreNamespaceProvider.class })
	public Datastore datastore() {
		return getDatastore(this.namespace);
	}

	@Bean
	@ConditionalOnMissingBean
	public DatastoreProvider datastoreSupplier(
			ObjectProvider<DatastoreNamespaceProvider> namespaceProvider,
			ObjectProvider<Datastore> datastoreProvider) {
		if (datastoreProvider.getIfAvailable() != null) {
			namespaceProvider.ifAvailable(unused -> {
				throw new DatastoreDataException(
						"A Datastore namespace provider and Datastore client were both configured. " +
								"Only one can be configured.");
			});
			return datastoreProvider::getIfAvailable;
		}
		return getDatastoreProvider(namespaceProvider.getIfAvailable());
	}
```

There would be no way to end up here with `namespaceProvider` not existing.

ChengyuanZhao: completed!

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/datastore/DatastoreTransactionManagerAutoConfiguration.java
ChengyuanZhao: Yes good point, fixed.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTransactionManager.java
```diff
@@ -45,9 +46,9 @@
  * @since 1.1
  */
 public class DatastoreTransactionManager extends AbstractPlatformTransactionManager {
-	private final Datastore datastore;
+	private final Supplier<Datastore> datastore;
```
meltsufin: `DatastoreProvider`?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTransactionManager.java
```diff
@@ -45,9 +46,9 @@
...
 
-	public DatastoreTransactionManager(final Datastore datastore) {
+	public DatastoreTransactionManager(final Supplier<Datastore> datastore) {
```
meltsufin: `DatastoreProvider`?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/DatastoreServiceObjectToKeyFactory.java
```diff
@@ -37,16 +39,16 @@
  */
 public class DatastoreServiceObjectToKeyFactory implements ObjectToKeyFactory {
 
-	private final Datastore datastore;
+	private final Supplier<Datastore> datastore;
```
meltsufin: `DatastoreProvider`

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/DatastoreServiceObjectToKeyFactory.java
```diff
@@ -37,16 +39,16 @@
...
 
-	public DatastoreServiceObjectToKeyFactory(Datastore datastore) {
+	public DatastoreServiceObjectToKeyFactory(Supplier<Datastore> datastore) {
```
meltsufin: `DatastoreProvider`

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/datastore/GcpDatastoreAutoConfiguration.java
```diff
@@ -101,21 +106,25 @@
...
-			builder.setNamespace(this.namespace);
-		}
+		return getDatastore(this.namespace);
+	}
 
-		if (this.host != null) {
-			builder.setHost(this.host);
+	@Bean
+	@ConditionalOnMissingBean
+	public DatastoreProvider datastoreSupplier(
```
meltsufin:
```suggestion
	public DatastoreProvider datastoreProvider(
```

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/datastore/GcpDatastoreAutoConfiguration.java
```diff
@@ -101,21 +106,25 @@
 	}
 
 	@Bean
-	@ConditionalOnMissingBean
+	@ConditionalOnMissingBean({ Datastore.class, DatastoreNamespaceProvider.class })
```
meltsufin: Oh, probably also need on missing `DatastoreProvider.class`.

ChengyuanZhao: Fixed

PR #1772, opened 2019-07-26
DatastorePageable uses url-safe-string cursor for JSON serializing

### LINKED ISSUE #1769: Datastore Pageable/Page Cursor JSON serializing
The issue is that the Cursor in our DatastorePageable can't automatically be serialized into JSON.

Two things we discussed:

1. looks like the Cursor is just a wrapper for some bytes and can be created from bytes easily. So if we store bytes inside the Pageable instead of Cursor then Jackson might automatically handle it. 
2. just annotate the cursor property to not be included in the JSON. this means when the user gives back the Page it has no cursor, but it still has the offset number of rows, which we handle too. It should just be a bit less efficient.

ChengyuanZhao:
fixes #1769 
stores url safe String cursor for json compatibility

### DIFF
--- a/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/DatastorePageable.java
+++ b/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/DatastorePageable.java
@@ -53,6 +53,10 @@ public class DatastorePageable extends PageRequest {
 		return new DatastorePageable(pageable, urlSafeCursor, totalCount);
 	}
 
+	public String getUrlSafeCursor() {
+		return this.urlSafeCursor;
+	}
+
 	@Override
 	public Pageable next() {
 		return from(super.next(), this.urlSafeCursor, this.totalCount);
--- a/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/DatastorePageable.java
+++ b/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/DatastorePageable.java
@@ -58,7 +58,7 @@ public class DatastorePageable extends PageRequest {
 		return from(super.next(), this.urlSafeCursor, this.totalCount);
 	}
 
-	public Cursor getCursor() {
+	public Cursor toCursor() {
 		return this.urlSafeCursor == null ? null : Cursor.fromUrlSafe(this.urlSafeCursor);
 	}
 
--- a/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/DatastorePageable.java
+++ b/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/DatastorePageable.java
@@ -22,33 +22,44 @@ import org.springframework.data.domain.PageRequest;
 import org.springframework.data.domain.Pageable;
 
 /**
+ * A pageable implementation for Cloud Datastore that uses the cursor for efficient reads.
+ *
  * @author Dmitry Solomakha
+ * @author Chengyuan Zhao
  */
 public class DatastorePageable extends PageRequest {
-	private final Cursor cursor;
+	private final String urlSafeCursor;
 
 	private final Long totalCount;
 
-	DatastorePageable(Pageable pageable, Cursor cursor, Long totalCount) {
+	private DatastorePageable(Pageable pageable, String urlSafeCursor, Long totalCount) {
 		super(pageable.getPageNumber(), pageable.getPageSize(), pageable.getSort());
-		this.cursor = cursor;
+		this.urlSafeCursor = urlSafeCursor;
 		this.totalCount = totalCount;
 	}
 
+	DatastorePageable(Pageable pageable, Cursor cursor, Long totalCount) {
+		this(pageable, cursor.toUrlSafe(), totalCount);
+	}
+
 	public static Pageable from(Pageable pageable, Cursor cursor, Long totalCount) {
+		return from(pageable, cursor == null ? null : cursor.toUrlSafe(), totalCount);
+	}
+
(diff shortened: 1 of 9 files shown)


spring-cloud-gcp-samples/spring-cloud-gcp-data-datastore-basic-sample/pom.xml
```diff
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="UTF-8"?>
+ <?xml version="1.0" encoding="UTF-8"?>
```
meltsufin: ??

spring-cloud-gcp-samples/spring-cloud-gcp-data-datastore-basic-sample/src/test/java/com/example/DatastoreBookshelfExampleTests.java
```diff
@@ -64,6 +77,17 @@ public static void checkToRun() {
...
 
+	@Test
+	public void testSerializedPage() {
+		Book book = new Book("Book1", "Author1", 2019);
+		book.id = 12345678L;
+		this.bookRepository.save(book);
+		Awaitility.await().atMost(15, TimeUnit.SECONDS)
+				.until(() -> this.bookRepository.count() == 1);
+		assertThat(sendRequest("/allbooksserialized", null, HttpMethod.GET))
+				.contains("content\":[{\"id\":12345678}],\"pageable\":");
```
meltsufin: Can you also verify the cursor?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/DatastorePageable.java
```diff
@@ -22,33 +22,44 @@
...
 
 /**
+ * A pageable implementation for Cloud Datastore that uses the cursor for efficient reads.
+ *
  * @author Dmitry Solomakha
+ * @author Chengyuan Zhao
  */
 public class DatastorePageable extends PageRequest {
-	private final Cursor cursor;
+	private final String urlSafeCursor;
```
meltsufin: Maybe add `getUrlSafeCursor` method?

spring-cloud-gcp-samples/spring-cloud-gcp-data-datastore-basic-sample/pom.xml
ChengyuanZhao: I think this must've been the auto-formating

spring-cloud-gcp-samples/spring-cloud-gcp-data-datastore-basic-sample/src/test/java/com/example/DatastoreBookshelfExampleTests.java
ChengyuanZhao: Unfortunately the cursor isn't actually part of the response. 
While the `DatastorePageable` type is what gets serialized into JSON, what gets written is just the base `Pageable` and doesn't have the cursor.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/DatastorePageable.java
ChengyuanZhao: They can retrieve the urlSafeString from the Cursor actually. In fact we don't have the string aspect of it visible at all to the outside. I prefer to keep it only in terms of Cursor externally.

ChengyuanZhao: Actually let me add that getter. Perhaps a public getter will cause the serializer to pick it up. Stay tuned

spring-cloud-gcp-samples/spring-cloud-gcp-data-datastore-basic-sample/pom.xml
ChengyuanZhao: fixed :)

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/DatastorePageable.java
ChengyuanZhao: fixed :)

spring-cloud-gcp-samples/spring-cloud-gcp-data-datastore-basic-sample/src/test/java/com/example/DatastoreBookshelfExampleTests.java
ChengyuanZhao: fixed :)

spring-cloud-gcp-samples/spring-cloud-gcp-data-datastore-basic-sample/src/test/java/com/example/DatastoreBookshelfExampleTests.java
```diff
@@ -64,6 +77,18 @@ public static void checkToRun() {
...
+	@Test
+	public void testSerializedPage() {
+		Book book = new Book("Book1", "Author1", 2019);
+		book.id = 12345678L;
+		this.bookRepository.save(book);
+		Awaitility.await().atMost(15, TimeUnit.SECONDS)
+				.until(() -> this.bookRepository.count() == 1);
+		String responseBody = sendRequest("/allbooksserialized", null, HttpMethod.GET);
+		assertThat(responseBody).contains("content\":[{\"id\":12345678}],\"pageable\":");
+		assertThat(responseBody).contains("\"urlSafeCursor\":");
```
meltsufin: Can you check that it has a value using regex?

ChengyuanZhao: added

PR #1784, opened 2019-07-30
Support Key ancestors in REST repo

### LINKED ISSUE #1773: Convertor for Key to String & vice versa
Problem: Unable to convert String to Key while using `spring-cloud-gcp-data-datastore-sample`

Basically I'm writing `InstrumentRepository` as below

```
public interface InstrumentRepository extends DatastoreRepository<Instrument, Key> {

}
```

which will allow datastore operations on Instrument Entity. 
https://github.com/spring-cloud/spring-cloud-gcp/blob/8f233f616e4647403177e49549e6aa081b108b8b/spring-cloud-gcp-samples/spring-cloud-gcp-data-datastore-sample/src/main/java/com/example/Instrument.java#L32-L35

But, Instrument entity has Key as type of `@Id` (shown above) since it is `@Descedent` of `Singer` Entity (Only Key types are allowed for descendants id)

So whenever I try to access/update using REST endpoint I get 

```
{
  "timestamp": "2019-07-27T10:03:44.846+0000",
  "status": 500,
  "error": "Internal Server Error",
  "message": "No converter found capable of converting from type [java.lang.String] to type [com.google.cloud.datastore.Key]",
  "path": "/api/instruments/Key%7BprojectId=demo-app,%20namespace=,%20path=%5BPathElement%7Bkind=instruments,%20id=5673082664517632,%20name=null%7D%5D%7D"
}
```

I thought of writing `CustomConvertor` which will implement `keyToString` & `stringToKey` methods from `com.google.appengine.api.datastore.KeyFactory` like below

```
    static final Converter<String, Key> STRING_TO_KEY_CONVERTER =
            new Converter<String, Key>() {
                @Override
                public Key convert(String s) {
                    return KeyFactory.stringToKey(s);
                }
            };
```

But then I come to know that `stringToKey` is not available in `com.google.cloud.datastore.KeyFactory`

Shall I add dependency for `com.google.appengine.api.datastore.KeyFactory` into my project? Will this be a good approach? 

Am I doing it wrong? Can someone suggest approach to get it done?

ChengyuanZhao:
fixes #1773 
Key converter for descendants Datastore REST

### DIFF
--- a/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/datastore/DatastoreKeyIdConverter.java
+++ b/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/datastore/DatastoreKeyIdConverter.java
@@ -1,0 +1,70 @@
+/*
+ * Copyright 2017-2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.gcp.autoconfigure.datastore;
+
+import java.io.Serializable;
+import java.io.UnsupportedEncodingException;
+import java.net.URLDecoder;
+
+import com.google.cloud.datastore.Key;
+
+import org.springframework.cloud.gcp.data.datastore.core.mapping.DatastoreDataException;
+import org.springframework.cloud.gcp.data.datastore.core.mapping.DatastoreMappingContext;
+import org.springframework.data.rest.webmvc.spi.BackendIdConverter;
+
+/**
+ * A key converter that parses Key JSON from REST requests.
+ *
+ * @author Chengyuan Zhao
+ *
+ * @since 1.2
+ */
+public class DatastoreKeyIdConverter implements BackendIdConverter {
+
+	private final DatastoreMappingContext datastoreMappingContext;
+
+	/**
+	 * Constructor.
+	 *
+	 * @param datastoreMappingContext the mapping context.
+	 */
+	public DatastoreKeyIdConverter(DatastoreMappingContext datastoreMappingContext) {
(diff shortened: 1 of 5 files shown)


spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/datastore/DatastoreKeyIdConverter.java
```diff
@@ -0,0 +1,70 @@
...
+		}
+	}
+
+	@Override
+	public String toRequestId(Serializable serializable, Class<?> aClass) {
+		return ((Key) serializable).toUrlSafe();
+	}
+
+	@Override
+	public boolean supports(Class<?> delimiter) {
```
dmitry-s: why is it called `delimiter`?

ChengyuanZhao: honestly I don't know :/ 

The interface this implements names it that. In reality it  is the entity type.

meltsufin: lol, can we call it `type` like in the Spanner one?

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/datastore/DatastoreKeyIdConverter.java
```diff
@@ -0,0 +1,70 @@
...
+	 * @param datastoreMappingContext the mapping context.
+	 */
+	public DatastoreKeyIdConverter(DatastoreMappingContext datastoreMappingContext) {
+		this.datastoreMappingContext = datastoreMappingContext;
+	}
+
+	@Override
+	public Serializable fromRequestId(String s, Class<?> aClass) {
+		try {
+			return Key.fromUrlSafe(URLDecoder.decode(s, "UTF-8"));
```
meltsufin: How do you know it will be `UTF-8`?
I would use `Charset.defaultCharset().name()` instead.

ChengyuanZhao: I didn't know that was an option; will do

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/datastore/DatastoreKeyIdConverter.java
ChengyuanZhao: done

Issue #1270, opened 2018-11-30
Add Usage Tracking to Logging


meltsufin:
Currently, we're not adding the necessary metadata to the API-style Logback appender that writes to Stackdriver Logging. We can fix this by providing a wrapper that adds the tool metadata to track that the API calls are coming from Spring Cloud GCP.

PR #1799, opened 2019-08-07
Add Usage Tracking to API Logging Appender

meltsufin:
Fixes: #1270.

### DIFF
--- a/spring-cloud-gcp-data-firestore/src/main/java/com/google/cloud/firestore/PublicClassMapper.java
+++ b/spring-cloud-gcp-data-firestore/src/main/java/com/google/cloud/firestore/PublicClassMapper.java
@@ -60,7 +60,8 @@ public final class PublicClassMapper {
 			decodedValue = convertToDateIfNecessary(decodedValue);
 			decodedFields.put(entry.getKey(), decodedValue);
 		}
-		return CustomClassMapper.convertToCustomClass(decodedFields, clazz);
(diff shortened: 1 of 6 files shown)


spring-cloud-gcp-data-firestore/src/main/java/com/google/cloud/firestore/PublicClassMapper.java
```diff
@@ -60,7 +60,8 @@ private PublicClassMapper() {
 			decodedValue = convertToDateIfNecessary(decodedValue);
 			decodedFields.put(entry.getKey(), decodedValue);
 		}
-		return CustomClassMapper.convertToCustomClass(decodedFields, clazz);
+		return CustomClassMapper.convertToCustomClass(decodedFields, clazz,
+				new DocumentReference(null, ResourcePath.create(doc.getName())));
```
elefeint: stray line from a different PR?

meltsufin: I upgraded to the latest Google Cloud Java BOM, and this line is needed to fix a compilation error. You can argue that the version update should be in its own PR, but c'mon. :smile:

elefeint: Ooh, that makes sense. I thought because firestore PR was also in flight that you were experimenting.

Issue #1779, opened 2019-07-29
Implement save/saveAll for firestoreTemplate


dmitry-s:

ChengyuanZhao:
Regarding saveAll there are actually 2 cases in ReactiveCrudRepository.

- saveAll of a publisher 
- saveAll of an iterable 

The first case of the publisher I think we should just be reusing the singular save method we already have since we have no idea how long we would need to buffer the incoming entities before saving. We might as well save as we go. 

The second case I think we should be using the write(WriteRequest) batch save. 

We should do these in separate PRs. 

@dmitry-s @meltsufin  FYI

meltsufin:
I think we should use `write(WriteRequest)` in both cases. It's just that for iterable, you'll only have to make a single call. For the publisher, you would send a series of batches, each one of size 1.
It's a streaming RPC. So, it should be pretty efficient.
```
  // Streams batches of document updates and deletes, in order.
  rpc Write(stream WriteRequest) returns (stream WriteResponse) {
    option (google.api.http) = {
      post: "/v1/{database=projects/*/databases/*}/documents:write"
      body: "*"
    };
  }
```

ChengyuanZhao:
sorry if this is a dumb question, but are you saying that calling `write` with WriteRequests that have an array of only 1 document (https://cloud.google.com/firestore/docs/reference/rpc/google.firestore.v1#google.firestore.v1.WriteRequest) repeatedly is automatically very efficient?

meltsufin:
Right, I don't see a problem with that. The reason it should be efficient is because it's streaming. The fact that there's only one document in the array should not matter much. The point is that the series of `WriteRequest`s is streamed.

ChengyuanZhao:
@meltsufin  For my curiosity: what does it mean that it's "streaming" ? 
From the docs it looks like they give you back a token/id you pass back and forth with each subsequent request. They look like just regular discrete requests physically and that the "streaming" part of it is more of abstract meaning given by one of the parameters (the stream token/id) .

meltsufin:
Although the explanations are not very deep, GRPC docs describe streaming here: https://grpc.io/docs/guides/concepts/.
The idea is that backpressure is supported and there is no need for a round-trip for each request.

dzou:
I believe this is now resolved.

PR #1800, opened 2019-08-07
Firestore publisher SaveAll

ChengyuanZhao:
related #1779
saveAll for publisher input

### DIFF
--- a/spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreReactiveOperations.java
+++ b/spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreReactiveOperations.java
@@ -37,7 +37,7 @@ public interface FirestoreReactiveOperations {
 	<T> Mono<T> save(T instance);
 
 	/**
-	 * Saves multiple objects to Cloud Firestore. Behaves as insert only (not update).
+	 * Saves multiple objects to Cloud Firestore. Behaves as insert or update.
 	 * @param instances the objects to save.
 	 * @param <T> the type of the objects to save.
 	 * @return a flux of the instances that were saved.
--- a/spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreReactiveOperations.java
+++ b/spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreReactiveOperations.java
@@ -16,6 +16,7 @@
 
 package org.springframework.cloud.gcp.data.firestore;
 
+import org.reactivestreams.Publisher;
 import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
 
@@ -27,13 +28,21 @@ import reactor.core.publisher.Mono;
  */
 public interface FirestoreReactiveOperations {
 	/**
-	 * Saves an instance of an object to Cloud Firestore.
-	 * Behaves as update or insert.
+	 * Saves an instance of an object to Cloud Firestore. Behaves as an insert only (not
+	 * update).
 	 * @param <T> the type of the object to save.
 	 * @param instance the instance to save.
 	 * @return the instance that was saved.
 	 */
 	<T> Mono<T> save(T instance);
+
+	/**
+	 * Saves multiple objects to Cloud Firestore. Behaves as insert only (not update).
+	 * @param instances the objects to save.
+	 * @param <T> the type of the objects to save.
+	 * @return a flux of the instances that were saved.
+	 */
+	<T> Flux<T> saveAll(Publisher<T> instances);
 
 	/**
 	 * Get all the entities of the given domain type.
--- a/spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreReactiveOperations.java
+++ b/spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreReactiveOperations.java
@@ -37,7 +37,7 @@ public interface FirestoreReactiveOperations {
 	<T> Mono<T> save(T instance);
 
 	/**
-	 * Saves multiple objects to Cloud Firestore. Behaves as insert or update.
+	 * Saves multiple objects to Cloud Firestore. Not atomic. Behaves as insert or update.
 	 * @param instances the objects to save.
 	 * @param <T> the type of the objects to save.
 	 * @return a flux of the instances that were saved.
--- a/spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreTemplate.java
+++ b/spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreTemplate.java
@@ -17,6 +17,7 @@
 
 package org.springframework.cloud.gcp.data.firestore;
 
+import com.google.protobuf.ByteString;
 import java.util.Map;
 
 import com.google.cloud.firestore.PublicClassMapper;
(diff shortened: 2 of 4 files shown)


spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreReactiveOperations.java
```diff
@@ -27,14 +28,22 @@
...
+	 * Saves an instance of an object to Cloud Firestore. Behaves as an insert only (not
+	 * update).
 	 * @param <T> the type of the object to save.
 	 * @param instance the instance to save.
 	 * @return the instance that was saved.
 	 */
 	<T> Mono<T> save(T instance);
 
+	/**
+	 * Saves multiple objects to Cloud Firestore. Behaves as insert only (not update).
```
dmitry-s: I think we should make it to work for both update and inserts.

ChengyuanZhao: done !

ChengyuanZhao:
@meltsufin @elefeint PTAL, it's now using the streaming API

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreTemplate.java
```diff
@@ -81,6 +92,39 @@ public FirestoreTemplate(FirestoreStub firestore, String parent) {
 		});
 	}
 
+	@Override
+	public <T> Flux<T> saveAll(Publisher<T> instances) {
+		Flux<T> input = Flux.from(instances);
+		return ObservableReactiveUtil.streamingBidirectionalCall(this::openWriteStreamObserver, input,
+				(T entity, Flux<WriteResponse> responses) -> {
+					Mono<WriteResponse> firstResponse = responses.next();
+					return firstResponse.map((WriteResponse streamIds) -> {
```
elefeint: I'd rename `streamIds` to `initialWriteResponse`.

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/util/ObservableReactiveUtil.java
```diff
@@ -68,6 +72,21 @@ private ObservableReactiveUtil() {
 		});
 	}
 
+	public static <ResponseT, RequestT, A> Flux<ResponseT> streamingBidirectionalCall(
```
elefeint: Maybe use `T` instead of `A` to represent the actual data type; it's more conventional.

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/util/ObservableReactiveUtil.java
```diff
@@ -68,6 +72,21 @@ private ObservableReactiveUtil() {
...
+			Function<StreamObserver<ResponseT>, StreamObserver<RequestT>> remoteCall,
+			Flux<A> inputs, BiFunction<A, Flux<ResponseT>, Mono<RequestT>> requestFunc) {
```
elefeint: I'd rename `inputs` to `entitiesToSave` or something else that more precisely reflects the purpose.

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreTemplate.java
```diff
@@ -81,6 +92,39 @@ public FirestoreTemplate(FirestoreStub firestore, String parent) {
...
+												.setName(this.parent + "/"
+														+ persistentEntity.collectionName() + "/"
+														+ idVal.toString())
+												.build())
+										.build())
+								.build();
+
+					});
+				})
+				.filter(response -> response.getWriteResultsCount() > 0).thenMany(input);
```
meltsufin: What is this doing? Is it basically ignoring the flux of responses, and just emitting the input flux?
What's the point of doing filter then?
If you we don't want to transform `WriteResponse`s to something meaningful just yet, then maybe just return a `Flux<Integer>`, which would just be the flux of `response.getWriteResultsCount()` from all responses.

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/util/ObservableReactiveUtil.java
```diff
@@ -68,6 +72,21 @@ private ObservableReactiveUtil() {
...
+
+		AtomicReference<StreamObserver<RequestT>> requestObserver = new AtomicReference<>();
+		Flux<ResponseT> responses = (Flux<ResponseT>) streamingCall(
+				obs -> requestObserver.set(remoteCall.apply((StreamObserver<ResponseT>) obs))).cache();
+
+		return inputs
+				.flatMap((A input) -> requestFunc.apply(input, responses)
+						.doOnNext(request -> requestObserver.get().onNext(request)))
+				.doOnComplete(() -> requestObserver.get().onCompleted())
+				.thenMany(responses);
```
meltsufin: I think this causes the two direction not to be asynchronous.
If I'm reading this correctly, you will not emit any responses until the requests stream is completed.

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreTemplate.java
ChengyuanZhao: The return type is dictated by the reactive crud repo interface we ultimately support.

The chaining is absolutely necessary as it causes the earlier steps of the chain to actually execute.

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/util/ObservableReactiveUtil.java
ChengyuanZhao: This is actually exactly the point.

Otherwise you will return before the final write has taken effect.

You still gain benefits because you are never blocked .the waiting is handled by the stream's own thread.

You absolutely must wait until the final write confirmation if you want strong consistency. Otherwise you have dirty eventual consistency like data store.

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreTemplate.java
ChengyuanZhao: The filter is how we distinguish write confirmations from other misc replies.

Only write confirmations fill that condition.

ChengyuanZhao: @meltsufin

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/util/ObservableReactiveUtil.java
ChengyuanZhao: Of course I could be mistaken :D

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreTemplate.java
meltsufin: Maybe I don't understand `thenMany()` correctly, but it seems to me regardless of what happens with the response flux, as soon as it completes, the whole input flux is emitted. Am I missing something?

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/util/ObservableReactiveUtil.java
meltsufin: Why couldn't you wait for the response flux to complete to make sure that all writes happened? The point is that you should be able to be notified that the first write has completed before the last write has completed or even published. That would be asynchronous.

meltsufin: https://github.com/spring-cloud/spring-cloud-gcp/issues/1813

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreTemplate.java
elefeint: Actually, that's right; I missed this. The response we need to filter out is the first one, so filtering it out won't affect when the Flux completes.
As far as whether to stream the output -- if Firestore sends back a stream of individual update statuses, then I agree with Mike that we should return the dynamic flux to the end user, who is then free to apply chaining operations like `thenMany`. But I am not sure how to interpret Firestore docs -- will the response actually be a stream of per-write responses, or will there be one WriteResponse containing an array of write statuses?

elefeint: Never mind the last part; I see that it's done on each signal from request flux.

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreTemplate.java
```diff
@@ -115,4 +137,43 @@ public FirestoreTemplate(FirestoreStub firestore, String parent) {
...
+		WriteRequest openStreamRequest = WriteRequest.newBuilder().setDatabase(this.databasePath).build();
+		StreamObserver<WriteRequest> requestStreamObserver = this.firestore.write(obs);
+		requestStreamObserver.onNext(openStreamRequest);
+		return requestStreamObserver;
+	}
+
+	private <T> Mono<WriteRequest> writeEntityStream(T entity, Flux<WriteResponse> responses) {
+		Mono<WriteResponse> firstResponse = responses.next();
+		return firstResponse.map(
+			initialResponse -> buildWriteRequest(initialResponse.getStreamId(), initialResponse.getStreamToken(), entity));
```
elefeint: `buildResourceName()` should not be done for every entity; it should be moved up to the top level where entity type is known. But it was awkward to drag the name through several method signatures, and this will get refactored again anyway.

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreTemplate.java
```diff
@@ -81,6 +93,16 @@ public FirestoreTemplate(FirestoreStub firestore, String parent) {
...
 	}
 
+	@Override
+	public <T> Flux<T> saveAll(Publisher<T> instances) {
+		Flux<T> input = Flux.from(instances);
+		return ObservableReactiveUtil.streamingBidirectionalCall(
+			this::openWriteStream,
+			input,
+			this::writeEntityStream
+		).filter(response -> response.getWriteResultsCount() > 0).thenMany(input);
```
elefeint: I kept the filter because it's in scope for the follow up issue #1813.

PR #1805, opened 2019-08-11
Make LoggingAppenderTests more robust

meltsufin:


meltsufin:
@spencergibb This should fix the Jenkins build. You also should no longer have to set the project id.

PR #1816, opened 2019-08-14
convert object parameters to keys

### LINKED ISSUE #1810: Convert entities to Key when they are used for @Query annotated methods
This will enable such query methods:
```java
class SingerRepository extends DatastoreRepository<Singer, String> {
    @Query("select * from singers where firstBand = @firstBand")
    filterSingersByfirstBand(@Param("firstBand") Band firstBand);
}
```
See #1806 for details

dmitry-s:
fixes #1810

### DIFF
--- a/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
+++ b/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
@@ -668,6 +668,10 @@ public class DatastoreTemplate implements DatastoreOperations, ApplicationEventP
 				this.datastoreMappingContext.getPersistentEntity(entityClass).kindName());
 	}
 
+	public Key getKey(Object entity) {
+		return getKey(entity, false);
+	}
+
 	private Key getKey(Object entity, boolean allocateKey, Key... ancestors) {
 		DatastorePersistentEntity datastorePersistentEntity = this.datastoreMappingContext
 				.getPersistentEntity(entity.getClass());
@@ -780,10 +784,6 @@ public class DatastoreTemplate implements DatastoreOperations, ApplicationEventP
 		if (this.eventPublisher != null) {
 			this.eventPublisher.publishEvent(event);
 		}
-	}
-
-	public ObjectToKeyFactory getObjectToKeyFactory() {
-		return this.objectToKeyFactory;
 	}
 
 	/**
--- a/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
+++ b/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
@@ -782,6 +782,10 @@ public class DatastoreTemplate implements DatastoreOperations, ApplicationEventP
 		}
 	}
 
+	public ObjectToKeyFactory getObjectToKeyFactory() {
+		return this.objectToKeyFactory;
+	}
+
 	/**
 	 * Class to hold caches for read and conversion.
 	 *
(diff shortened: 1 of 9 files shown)


spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/GqlDatastoreQuery.java
```diff
@@ -372,6 +375,15 @@ private void setEvaluatingSpelQueryContext() {
 			}
 		}
 
+		private Object convertEntitiesToKeys(Object o, DatastoreTemplate datastoreTemplate, DatastoreMappingContext datastoreMappingContext) {
+			Entity entity = AnnotationUtils.findAnnotation(o.getClass(), Entity.class);
+			if (entity != null) {
+				return datastoreTemplate.getObjectToKeyFactory()
+						.getKeyFromObject(o, datastoreMappingContext.getPersistentEntity(o.getClass()));
```
meltsufin: Can you do this logic in `AbstractDatastoreQuery.convertCollectionParamToCompatibleArray` so that it applies to all types of queries?

dmitry-s: This method is used to convert collection-like parameters to arrays of Datastore Values, so I don't think it's a good fit

meltsufin: I see. I still think that doing this conversion in `ParsedQueryWithTagsAndValues()` is not the best place. The rest of the parameter conversion is happening in `GqlDatastoreQuery.bindArgsToGqlQuery()`. Can you do it there?
Maybe it can even be done in `ReadWriteConversions.convertOnWriteSignle()`?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
```diff
@@ -668,6 +668,10 @@ private Key getKeyFromId(Object id, Class entityClass) {
 				this.datastoreMappingContext.getPersistentEntity(entityClass).kindName());
 	}
 
+	public Key getKey(Object entity) {
```
meltsufin: Java doc?

spring-cloud-gcp-data-datastore/src/test/java/org/springframework/cloud/gcp/data/datastore/repository/query/GqlDatastoreQueryTests.java
```diff
@@ -117,28 +120,38 @@ public void compoundNameConventionTest() {
...
-				"blahblah", 1.11, 2.22 };
+				"blahblah", 1.11, 2.22, trade };
 
 		String[] paramNames = new String[] { "tag0", "tag1", "tag2", "tag3", "tag4",
-				"tag5", "tag6", "tag7" };
+				"tag5", "tag6", "tag7", "tag8" };
 
 		buildParameters(paramVals, paramNames);
 
+				KeyFactory keyFactory = new KeyFactory("proj");
```
meltsufin: Spacing issue?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/GqlDatastoreQuery.java
```diff
@@ -423,6 +425,14 @@ private String addSort(String finalGql, Sort sort) {
 					.collect(Collectors.joining(", "));
 			return finalGql + ORDER_BY + orderString;
 		}
+
+		private Object convertEntitiesToKeys(Object o) {
+			Entity entity = AnnotationUtils.findAnnotation(o.getClass(), Entity.class);
```
meltsufin: Can you use `mappingContext.hasPersistentEntityFor(o.getClass())` instead?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
dmitry-s: exposed in `DatastoreOperations`

PR #1826, opened 2019-08-16
[do not commit] Firestore reactive vs. Client library performance comparison

dzou:
Client library can save 300 records in 2sec.
FirestoreTemplate saveAll() saves 300 records in 17 seconds.


dzou:
After batching is implemented, the client library and reactive impls have comparable performance.

The benchmarking code is in `FirestoreIntegrationTests.java`.

- `saveBulkWithReactor()` runs in 3475ms.

- `saveBulkWithClientLibraries()` runs in 3485ms

PR #1833, opened 2019-08-23
field-level unindex for maps

### LINKED ISSUE #1832: @Unindex with Map
Currently we don't support `@Unindexed` for Map.
`com.google.cloud.datastore.DatastoreException: Exclude from indexes cannot be set on a list value at org.springframework.web.servlet.FrameworkServlet.processRequest(FrameworkServlet.java:1013) ~[spring-webmvc-5.1.3.RELEASE.jar:5.1.3.RELEASE]`

In order to support that we need to mark every property in the map as unindexed.

dmitry-s:
fixes #1832

### DIFF
--- a/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/DefaultDatastoreEntityConverter.java
+++ b/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/DefaultDatastoreEntityConverter.java
@@ -24,6 +24,9 @@ import java.util.Set;
 import java.util.stream.Collectors;
 
 import com.google.cloud.datastore.BaseEntity;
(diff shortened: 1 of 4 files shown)


spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/convert/DefaultDatastoreEntityConverter.java
```diff
@@ -196,7 +199,15 @@ public void write(Object source, BaseEntity.Builder sink) {
 	private Value setExcludeFromIndexes(Value convertedVal) {
 		// ListValues must have its contents individually excluded instead.
 		// the entire list must NOT be excluded or there will be an exception.
-		if (convertedVal.getClass().equals(ListValue.class)) {
+		// Same for maps which are storedd as EntityValue.
```
elefeint: stored

PR #1837, opened 2019-08-27
Refactor Bidi Streaming Write to Buffer Entities before writing

### LINKED ISSUE #1813: Make ObservableReactiveUtil.streamingBidirectionalCall Asynchronous
See https://github.com/spring-cloud/spring-cloud-gcp/pull/1800/files/ff256cb1c085516f08691f307583dff30079cbef#r313129168

dzou:
Refactors the Bidi Streaming write (the saveAll() operation) to buffer the incoming entities before writing.

Fixes #1813.

Note that I wasn't able to generalize this solution to put it in ObservableReactiveUtils... It is getting complex and it felt to me that the generalization would be harder to understand than just keeping the code here.

### DIFF
--- a/spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreReactiveOperations.java
+++ b/spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreReactiveOperations.java
@@ -40,7 +40,7 @@ public interface FirestoreReactiveOperations {
 	 * Saves multiple objects to Cloud Firestore. Not atomic. Behaves as insert or update.
 	 * @param instances the objects to save.
 	 * @param <T> the type of the objects to save.
-	 * @return a flux of the instances that were saved
+	 * @return a {@link Flux} of the instances that were saved
 	 */
 	<T> Flux<T> saveAll(Publisher<T> instances);
 
--- a/spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreReactiveOperations.java
+++ b/spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreReactiveOperations.java
@@ -40,9 +40,9 @@ public interface FirestoreReactiveOperations {
 	 * Saves multiple objects to Cloud Firestore. Not atomic. Behaves as insert or update.
 	 * @param instances the objects to save.
 	 * @param <T> the type of the objects to save.
-	 * @return a flux of the instances that were saved.
+	 * @return a Mono indicating completion of the operation.
 	 */
-	<T> Flux<T> saveAll(Publisher<T> instances);
+	<T> Mono<Void> saveAll(Publisher<T> instances);
 
 	/**
 	 * Get all the entities of the given domain type.
--- a/spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreReactiveOperations.java
+++ b/spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreReactiveOperations.java
@@ -42,7 +42,7 @@ public interface FirestoreReactiveOperations {
 	 * @param <T> the type of the objects to save.
 	 * @return a Mono indicating completion of the operation.
 	 */
-	<T> Mono<Void> saveAll(Publisher<T> instances);
+	<T> Flux<T> saveAll(Publisher<T> instances);
 
 	/**
 	 * Get all the entities of the given domain type.
--- a/spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreReactiveOperations.java
+++ b/spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreReactiveOperations.java
@@ -40,7 +40,7 @@ public interface FirestoreReactiveOperations {
 	 * Saves multiple objects to Cloud Firestore. Not atomic. Behaves as insert or update.
 	 * @param instances the objects to save.
 	 * @param <T> the type of the objects to save.
-	 * @return a Mono indicating completion of the operation.
+	 * @return a flux of the instances that were saved
 	 */
 	<T> Flux<T> saveAll(Publisher<T> instances);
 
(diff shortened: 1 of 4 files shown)


dzou:
> Can you also ask Mark to take a look at this? Thanks!

@mp911de - Hey Mark, could we get your counsel on Reactive in this PR? We just specifically have a question about this method: [saveAll](https://github.com/spring-cloud/spring-cloud-gcp/blob/bfa7abbc1a19352deb6bf348c82bfe30c1b73595/spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreTemplate.java#L143).

It is basically us trying to wrap a bi-directional stream into Reactive patterns. We have a `requestStream` that we send elements into, and a `responseStream` that outputs responses for each batch of elements we submit to the requestStream. And from every response that we get from the responses stream, there is a stream token; we take the most recent stream token we got and attach it to the next request that we send to `requestStream`. We would appreciate your input!

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreReactiveOperations.java
```diff
@@ -40,9 +40,9 @@
 	 * Saves multiple objects to Cloud Firestore. Not atomic. Behaves as insert or update.
 	 * @param instances the objects to save.
 	 * @param <T> the type of the objects to save.
-	 * @return a flux of the instances that were saved.
+	 * @return a Mono indicating completion of the operation.
 	 */
-	<T> Flux<T> saveAll(Publisher<T> instances);
+	<T> Mono<Void> saveAll(Publisher<T> instances);
```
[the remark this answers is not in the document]
dzou: I was able to preserve the input flux as return type in the latest commit.

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/util/ObservableReactiveUtil.java
```diff
@@ -82,19 +83,26 @@ private ObservableReactiveUtil() {
...
-						.doOnNext(request -> requestObserver.get().onNext(request)))
-				.doOnComplete(() -> requestObserver.get().onCompleted())
-				.thenMany(responses);
+				.flatMap(input ->
+						responsesFlux.next()
+								.map(mostRecentResponse -> responseHandler.apply(input, mostRecentResponse))
+								.doOnNext(nextRequest -> requestStream.get().onNext(nextRequest))
+								.thenReturn(input))
+				.doOnComplete(() -> requestStream.get().onCompleted())
+				.delayUntil(input -> responsesFlux)
```
ChengyuanZhao: I don't think you want to use delayUntil here because it will repeatedly check if responsesFlux has completed (delayUntil checks this condition for every single element). 

You could just use thenMany here since it'll just wait until your inputs are all sent over and then it'll wait until responseFlux is done, and then it'll continue on to your flatMap on your next line.

dzou:
As discussed offline with Chengyuan, I decided to keep the bidi code in the ObservableUtils class since this is going to be reused by delete and other batch operations. Modified it so that it is able to return the original flux of inputted elements and also added setter for batch size. PTAL.

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/util/ObservableReactiveUtil.java
dzou: Right, thenMany would be ideal but then I lose my original `input` type.

I think ideally we'd have something like a `delayUntilMany`... or something. But I think delayUntil is still the best choice here.

ChengyuanZhao: you're completely right

mp911de:
I've been on vacation lately, so sorry for my late response. I'm not sure I understood your question.

In general, `saveAll` should mirror the store's (or driver's) semantics. If it is the most efficient way, to stream element by elements as save stream, then that is a great way to solve your requirement.

In MongoDB, we have also a batch insert but Mongo's approach is to take a materialized batch of documents and then insert them in a single command (see [`ReactiveMongoOperations.insertAll(Mono<? extends Collection<? extends T>>)`](https://github.com/spring-projects/spring-data-mongodb/blob/master/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/core/ReactiveMongoOperations.java#L1054).

For Cassandra, insert-all issues a series of individual commands.

Let me know if you need to revisit that topic.

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreReactiveOperations.java
```diff
@@ -40,7 +40,7 @@
 	 * Saves multiple objects to Cloud Firestore. Not atomic. Behaves as insert or update.
 	 * @param instances the objects to save.
 	 * @param <T> the type of the objects to save.
-	 * @return a flux of the instances that were saved.
+	 * @return a Mono indicating completion of the operation.
```
ChengyuanZhao: Hmmm, the return type is still Flux though

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreTemplate.java
```diff
@@ -110,12 +147,11 @@ public FirestoreTemplate(FirestoreStub firestore, String parent) {
 
 	@Override
 	public <T> Flux<T> saveAll(Publisher<T> instances) {
-		Flux<T> input = Flux.from(instances);
+		Flux<List<T>> inputs = Flux.from(instances).bufferTimeout(
+				FIRESTORE_WRITE_MAX_SIZE, bufferTimeout);
```
meltsufin: Shouldn't you use `firestoreBufferWriteSize` here?

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreTemplate.java
```diff
@@ -75,6 +82,36 @@ public FirestoreTemplate(FirestoreStub firestore, String parent) {
...
+	public Duration getBufferTimeoutDuration() {
+		return this.bufferTimeout;
+	}
+
+	/**
+	 * Sets how many entities to include in an insert/update/delete buffered operation.
+	 * <p>The maximum buffer size is 500. In most cases users should leave this at the maximum value.
+	 * @param firestoreBufferWriteSize the entity buffer size for buffered operations (default = 500)
+	 */
+	public void setFirestoreBufferWriteSize(int firestoreBufferWriteSize) {
```
meltsufin: The buffer settings are not named consistently. I would do:
`(get/set)SaveAllBufferTimeoutDuration` and `(get/set)SaveAllBufferTimeoutDuration`.
Also the `saveAll` method needs a javadoc explaining that these settings exist for it.

PR #1845, opened 2019-08-29
Spanner doesn't wrap query with no pageable

### LINKED ISSUE #1843: Custom query in SpannerRespository doesn't seem to honor ORDER BY without including a LIMIT clause

ChengyuanZhao:
fixes #1843

Cloud spanner doesn't preserve derived tables' orders, so we avoid wrapping the derived table if possible.

### DIFF
--- a/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryExecutor.java
+++ b/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryExecutor.java
@@ -137,6 +137,14 @@ public final class SpannerStatementQueryExecutor {
 			SpannerMappingContext mappingContext) {
 		SpannerPersistentEntity<?> persistentEntity = mappingContext
 				.getPersistentEntity(entityClass);
+
+		// Cloud Spanner does not preserve the order of derived tables so we must not wrap the
+		// derived table
+		// in SELECT * FROM () if there is no overriding pageable param.
+		if ((options.getSort() == null || options.getSort().isUnsorted()) && options.getLimit() == null
+				&& options.getOffset() == null) {
+			return sql;
+		}
 		StringBuilder sb = SpannerStatementQueryExecutor.applySort(options.getSort(),
 				new StringBuilder("SELECT * FROM (").append(sql).append(")"), (o) -> {
 					SpannerPersistentProperty property = persistentEntity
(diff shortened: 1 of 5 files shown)


spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/repository/it/SpannerRepositoryIntegrationTests.java
```diff
@@ -167,6 +167,10 @@ public void queryMethodsTest() {
 		List<Trade> buyTradesRetrieved = this.tradeRepository
 				.annotatedTradesByAction("BUY");
 		assertThat(buyTradesRetrieved).containsExactlyInAnyOrderElementsOf(trader1BuyTrades);
+		assertThat(buyTradesRetrieved.get(0).getId()
+				.compareTo(buyTradesRetrieved.get(1).getId())).isPositive();
```
dmitry-s: why isPositive relevant?

ChengyuanZhao: This is to check that the results came back in sorted order. IsPositive is checking that it came back in DESC order

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryExecutor.java
```diff
@@ -137,6 +137,14 @@ private SpannerStatementQueryExecutor() {
 			SpannerMappingContext mappingContext) {
 		SpannerPersistentEntity<?> persistentEntity = mappingContext
 				.getPersistentEntity(entityClass);
+
+		// Cloud Spanner does not preserve the order of derived tables so we must not wrap the
+		// derived table
+		// in SELECT * FROM () if there is no overriding pageable param.
+		if ((options.getSort() == null || options.getSort().isUnsorted()) && options.getLimit() == null
```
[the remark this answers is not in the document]
ChengyuanZhao: They do. Any sort of OFFSET or LIMIT needs to also go outside the parens. They happen subsequent to the SORT, so the SQL is already wrapped.

spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/repository/it/SpannerRepositoryIntegrationTests.java
dmitry-s: Oh I see. Maybe just use assertThat(..).isGreater{Less}Than(..)?

Poojak16:
Does spanner support pagination along with custom column like findByCOLUMNNAME(String column_name,Pageable pageable).@ChengyuanZhao

dzou:
@Poojak16 - Yeah I think that should work because the pageable parameter is processed separate from the query. Give it a try - if it doesn't work, please post an issue in our github issue tracker; we accommodate feature requests quickly.

Poojak16:
@dzou it is only working if write it using custom query ..
Not working when we using JPA convention of method naming

@Query("SELCT *. ...........)
Working --> getCountByColumnName(@Param ("xyz") String xyz , Pageable paging)

NOT WORKING --->findByColumnName(String columnanme , Pageable Paging)

with the later one it is throwing error as Tags are not equals to param.

meltsufin:
@Poojak16 I confirmed it's a real issue. See: #2390.

PR #1846, opened 2019-08-30
ReactiveCrudRepository implementation

### LINKED ISSUE #1831: Implement ReactiveRepository

dmitry-s:
fixes #1831

### DIFF
--- a/spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreReactiveOperations.java
+++ b/spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreReactiveOperations.java
@@ -32,7 +32,7 @@ public interface FirestoreReactiveOperations {
 	 * update).
 	 * @param <T> the type of the object to save.
 	 * @param instance the instance to save.
-	 * @return the instance that was saved.
+	 * @return {@link Mono} emitting the saved entity.
 	 */
 	<T> Mono<T> save(T instance);
 
@@ -40,7 +40,7 @@ public interface FirestoreReactiveOperations {
 	 * Saves multiple objects to Cloud Firestore. Not atomic. Behaves as insert or update.
 	 * @param instances the objects to save.
 	 * @param <T> the type of the objects to save.
-	 * @return a flux of the instances that were saved.
+	 * @return {@link Flux} emitting the saved entities.
 	 */
 	<T> Flux<T> saveAll(Publisher<T> instances);
 
@@ -48,7 +48,7 @@ public interface FirestoreReactiveOperations {
 	 * Get all the entities of the given domain type.
 	 * @param <T> the type param of the domain type.
 	 * @param entityClass the domain type to get.
-	 * @return the entities that were found.
+	 * @return {@link Mono} emitting the number of entities.
 	 */
 	<T> Flux<T> findAll(Class<T> entityClass);
 
@@ -56,7 +56,61 @@ public interface FirestoreReactiveOperations {
 	 * Delete all entities of a given domain type.
 	 * @param <T> the type param of the domain type.
 	 * @param entityClass the domain type to delete from Cloud Datastore.
-	 * @return the number of entities that were deleted.
+	 * @return {@link Mono} emitting the number of deleted entities.
 	 */
 	<T> Mono<Long> deleteAll(Class<T> entityClass);
+
+	/**
+	 * Test if the entity of the given domain type with a given id exists.
+	 * @param <T> the type param of the domain type.
+	 * @param entityClass the domain type of the entity.
+	 * @param idPublisher publisher that provides an id.
+	 * @return {@link Mono} emitting {@literal true} if an entity with the given id exists, {@literal false} otherwise.
+	 */
+	<T> Mono<Boolean> existsById(Publisher<String> idPublisher, Class<T> entityClass);
+
+	/**
+	 * Get an entity of the given domain type by id.
+	 * @param <T> the type param of the domain type.
+	 * @param entityClass the domain type of the entity.
+	 * @param idPublisher publisher that provides an id.
+	 * @return {@link Mono} emitting the found entity.
+	 */
+	<T> Mono<T> findById(Publisher idPublisher, Class<T> entityClass);
+
+	/**
+	 * Get an entity of the given domain type by id.
+	 * @param <T> the type param of the domain type.
+	 * @param entityClass the domain type of the entity.
+	 * @param idPublisher publisher that provides ids.
+	 * @return {@link Flux} emitting the found entities.
+	 */
+	<T> Flux<T> findAllById(Publisher idPublisher, Class<T> entityClass);
+
+	/**
+	 * Count entities of the given domain.
+	 * @param <T> the type param of the domain type.
+	 * @param entityClass the domain type of entities.
+	 * @return {@link Mono} emitting the number of entities.
+	 */
+	<T> Mono<Long> count(Class<T> entityClass);
+
+	/**
+	 * Delete entities provided by publisher.
+	 * @param <T> the type param of the domain type.
+	 * @param entityPublisher publisher that provides entities to be removed.
+	 * @return {@link Mono} signaling when operation has completed.
+	 */
+	<T> Mono<Void> delete(Publisher<T> entityPublisher);
+
+	/**
+	 * Delete all entities of a given domain type.
+	 * @param <T> the type param of the domain type.
+	 * @param entityClass the domain type of entities.
+	 * @param idPublisher publisher that provides ids of entities to be removed.
+	 * @return {@link Mono} signaling when operation has completed.
+	 */
+	<T> Mono<Void> deleteAllById(Publisher<T> idPublisher, Class entityClass);
+
+
 }
--- a/spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreReactiveOperations.java
+++ b/spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreReactiveOperations.java
@@ -48,7 +48,7 @@ public interface FirestoreReactiveOperations {
 	 * Get all the entities of the given domain type.
 	 * @param <T> the type param of the domain type.
 	 * @param entityClass the domain type to get.
-	 * @return {@link Mono} emitting the number of entities.
+	 * @return {@link Mono} emitting the found entities.
 	 */
 	<T> Flux<T> findAll(Class<T> entityClass);
 
@@ -62,33 +62,37 @@ public interface FirestoreReactiveOperations {
 
 	/**
 	 * Test if the entity of the given domain type with a given id exists.
+	 * Uses the first emitted element to perform the query.
 	 * @param <T> the type param of the domain type.
 	 * @param entityClass the domain type of the entity.
 	 * @param idPublisher publisher that provides an id.
-	 * @return {@link Mono} emitting {@literal true} if an entity with the given id exists, {@literal false} otherwise.
+	 * @return {@link Mono} emitting {@code true} if an entity with the given id exists, {@code false} otherwise.
 	 */
 	<T> Mono<Boolean> existsById(Publisher<String> idPublisher, Class<T> entityClass);
 
 	/**
 	 * Get an entity of the given domain type by id.
+	 * Uses the first emitted element to perform the query.
 	 * @param <T> the type param of the domain type.
-	 * @param entityClass the domain type of the entity.
 	 * @param idPublisher publisher that provides an id.
+	 * @param entityClass the domain type of the entity.
 	 * @return {@link Mono} emitting the found entity.
 	 */
-	<T> Mono<T> findById(Publisher idPublisher, Class<T> entityClass);
+	<T> Mono<T> findById(Publisher<String> idPublisher, Class<T> entityClass);
 
 	/**
 	 * Get an entity of the given domain type by id.
 	 * @param <T> the type param of the domain type.
-	 * @param entityClass the domain type of the entity.
 	 * @param idPublisher publisher that provides ids.
+	 * @param entityClass the domain type of the entity.
 	 * @return {@link Flux} emitting the found entities.
 	 */
-	<T> Flux<T> findAllById(Publisher idPublisher, Class<T> entityClass);
+	<T> Flux<T> findAllById(Publisher<String> idPublisher, Class<T> entityClass);
 
 	/**
 	 * Count entities of the given domain.
+	 * Note that Firestore doesn't support "count" operation natively,
+	 * so id query will be executed and all ids will be retrieved so they could be counted.
 	 * @param <T> the type param of the domain type.
 	 * @param entityClass the domain type of entities.
 	 * @return {@link Mono} emitting the number of entities.
@@ -104,13 +108,13 @@ public interface FirestoreReactiveOperations {
 	<T> Mono<Void> delete(Publisher<T> entityPublisher);
 
 	/**
-	 * Delete all entities of a given domain type.
+	 * Delete entities of a given domain type using ids published by producer.
 	 * @param <T> the type param of the domain type.
-	 * @param entityClass the domain type of entities.
 	 * @param idPublisher publisher that provides ids of entities to be removed.
+	 * @param entityClass the domain type of entities.
 	 * @return {@link Mono} signaling when operation has completed.
 	 */
-	<T> Mono<Void> deleteAllById(Publisher<T> idPublisher, Class entityClass);
+	<T> Mono<Void> deleteById(Publisher<String> idPublisher, Class entityClass);
 
 
 }
--- a/spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreReactiveRepository.java
+++ b/spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreReactiveRepository.java
@@ -27,7 +27,7 @@ import org.springframework.data.repository.reactive.ReactiveCrudRepository;
  *
  * @since 1.2
  */
-public class FirestoreReactiveRepository implements ReactiveCrudRepository {
+public class FirestoreReactiveRepository<T> implements ReactiveCrudRepository<T, String> {
 
 	private FirestoreTemplate firestoreTemplate;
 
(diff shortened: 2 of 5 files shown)


spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreReactiveOperations.java
```diff
@@ -32,31 +32,85 @@
...
+	<T> Mono<T> findById(Publisher idPublisher, Class<T> entityClass);
+
+	/**
+	 * Get an entity of the given domain type by id.
+	 * @param <T> the type param of the domain type.
+	 * @param entityClass the domain type of the entity.
+	 * @param idPublisher publisher that provides ids.
+	 * @return {@link Flux} emitting the found entities.
+	 */
+	<T> Flux<T> findAllById(Publisher idPublisher, Class<T> entityClass);
```
dzou: Same here - either make the idPublisher generic or make it `Publisher<String>`

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreReactiveOperations.java
```diff
@@ -32,31 +32,85 @@
...
+	<T> Mono<Void> delete(Publisher<T> entityPublisher);
+
+	/**
+	 * Delete all entities of a given domain type.
+	 * @param <T> the type param of the domain type.
+	 * @param entityClass the domain type of entities.
+	 * @param idPublisher publisher that provides ids of entities to be removed.
+	 * @return {@link Mono} signaling when operation has completed.
+	 */
+	<T> Mono<Void> deleteAllById(Publisher<T> idPublisher, Class entityClass);
```
dzou: I don't think this should be called deleteAllById since the input is `Publisher<T>`

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreReactiveRepository.java
```diff
@@ -0,0 +1,125 @@
...
+import reactor.core.publisher.Mono;
+
+import org.springframework.data.repository.reactive.ReactiveCrudRepository;
+
+/**
+ * @author Dmitry Solomakha
+ *
+ * @since 1.2
+ */
+public class FirestoreReactiveRepository implements ReactiveCrudRepository {
```
dzou: It looks like `FirestoreReactiveOperations` matches fairly closely with `ReactiveCrudRepository`; is it feasible for the template to implement `ReactiveCrudRepository` directly and remove `FirestoreReactiveOperations`? If there are enough differences then this is fine.

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreTemplate.java
```diff
@@ -75,32 +82,37 @@ public FirestoreTemplate(FirestoreStub firestore, String parent) {
...
-	public <T> Flux<T> findAllById(Publisher idPublisher, Class<T> aClass) {
-		return ((Flux<String>) Flux.from(idPublisher)).flatMap(id -> {
-			FirestorePersistentEntity<?> persistentEntity = this.mappingContext.getPersistentEntity(aClass);
-			GetDocumentRequest request = GetDocumentRequest.newBuilder()
-					.setName(this.parent + "/" + persistentEntity.collectionName() + "/" + id).build();
-			return ObservableReactiveUtil.<Document>unaryCall(obs -> this.firestore.getDocument(request, obs));
-		}).onErrorMap(throwable -> new FirestoreDataException("Unable to find an entry by id", throwable))
-				.map(document -> PublicClassMapper.convertToCustomClass(document, aClass));
+	public <T> Flux<T> findAllById(Publisher idPublisher, Class<T> entityClass) {
+		return ((Flux<String>) Flux.from(idPublisher)).flatMap(id -> getDocument(id, entityClass, null))
```
dzou: Move the `.flatMap` operator to its own line.

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreTemplate.java
```diff
@@ -75,32 +82,37 @@ public FirestoreTemplate(FirestoreStub firestore, String parent) {
...
+				.flatMap(t -> t == DocumentReadContext.EMPTY ? Mono.empty() : Mono.just(t))
```
dzou: It seems like you must check `DocumentReadContext.EMPTY` after every getDocument call; it might be better to just have `getDocument` return `Mono<Document>` and if the document doesn't exist, `getDocument` would return Mono.empty. Then you can use `Mono.switchOnEmpty()` to check if a document exists or not.

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreTemplate.java
```diff
@@ -124,34 +136,76 @@ public FirestoreTemplate(FirestoreStub firestore, String parent) {
...
+		GetDocumentRequest.Builder builder = GetDocumentRequest.newBuilder()
+				.setName(buildResourceName(persistentEntity, id));
+
+		if (documentMask != null) {
+			builder.setMask(documentMask);
+		}
+
+		return ObservableReactiveUtil.<Document>unaryCall(obs -> this.firestore.getDocument(builder.build(), obs))
+				.map(DocumentReadContext::new)
+				.onErrorReturn(throwable -> throwable.getMessage().startsWith("NOT_FOUND: Document"),
```
dzou: "NOT_FOUND: Document" should be a constant.

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreTemplate.java
```diff
@@ -187,7 +241,32 @@ public FirestoreTemplate(FirestoreStub firestore, String parent) {
...
+	}
+
+	public String getIdValue(Object entity, FirestorePersistentEntity persistentEntity) {
+		FirestorePersistentProperty idProperty = persistentEntity.getIdPropertyOrFail();
+		Object idVal = persistentEntity.getPropertyAccessor(entity).getProperty(idProperty);
+
+		return idVal.toString();
+	}
+
+	static private class DocumentReadContext {
```
dzou: If you take my above suggestion about returning `Mono<Document>` from getDocument, then it will allow you to remove this class.

dzou: Also btw, style guide is to move private before static `private static class`

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreTemplate.java
```diff
@@ -187,7 +241,32 @@ public FirestoreTemplate(FirestoreStub firestore, String parent) {
...
 
-		return this.parent + "/" + persistentEntity.collectionName() + "/" + idVal.toString();
+		return buildResourceName(persistentEntity, idVal.toString());
+	}
+
+	private String buildResourceName(FirestorePersistentEntity<?> persistentEntity, String s) {
+		return this.parent + "/" + persistentEntity.collectionName() + "/" + s;
+	}
+
+	public String getIdValue(Object entity, FirestorePersistentEntity persistentEntity) {
```
dzou: Can this also be private?

Otherwise this should be moved with other public methods and add javadoc.

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreTemplate.java
```diff
@@ -124,34 +136,76 @@ public FirestoreTemplate(FirestoreStub firestore, String parent) {
...
+
+	private Mono<Empty> callDelete(String docName) {
+		DeleteDocumentRequest deleteDocumentRequest = DeleteDocumentRequest.newBuilder().setName(docName)
 				.build();
 		return ObservableReactiveUtil.unaryCall(
 						obs -> this.firestore.deleteDocument(deleteDocumentRequest, obs));
 	}
 
+	private <T> Mono<Void> deleteAllByFullName(Publisher<T> idPublisher) {
+		return Flux.from(idPublisher).flatMap(id -> callDelete(id.toString())).then(Mono.empty());
```
dzou: You don't need to write `.then(Mono.empty())`; it can be shortened to adding a `.then()` at the end.

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreTemplate.java
```diff
@@ -124,34 +136,76 @@ public FirestoreTemplate(FirestoreStub firestore, String parent) {
...
+	}
+
+	private Mono<Empty> callDelete(String docName) {
+		DeleteDocumentRequest deleteDocumentRequest = DeleteDocumentRequest.newBuilder().setName(docName)
 				.build();
 		return ObservableReactiveUtil.unaryCall(
 						obs -> this.firestore.deleteDocument(deleteDocumentRequest, obs));
 	}
 
+	private <T> Mono<Void> deleteAllByFullName(Publisher<T> idPublisher) {
```
dzou: Can this be consolidated into `delete()`? that is the only caller of this helper method. Then all the public delete methods can just delegate to `callDelete`.

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreReactiveOperations.java
```diff
@@ -32,31 +32,85 @@
...
+	 * @param entityClass the domain type of the entity.
+	 * @param idPublisher publisher that provides an id.
+	 * @return {@link Mono} emitting {@literal true} if an entity with the given id exists, {@literal false} otherwise.
+	 */
+	<T> Mono<Boolean> existsById(Publisher<String> idPublisher, Class<T> entityClass);
+
+	/**
+	 * Get an entity of the given domain type by id.
+	 * @param <T> the type param of the domain type.
+	 * @param entityClass the domain type of the entity.
```
elefeint: Nit: switch order of javadoc params to be the same as method params.

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreReactiveRepository.java
```diff
@@ -0,0 +1,125 @@
...
+
+	private Class type;
+
+	public FirestoreReactiveRepository(FirestoreTemplate firestoreTemplate, Class type) {
+		this.firestoreTemplate = firestoreTemplate;
+		this.type = type;
+	}
+
+	@Override
+	public Mono save(Object entity) {
```
elefeint: All `Mono`/`Flux` should be parameterized with real types they are wrapping. Otherwise we'll be causing warnings in downstream projects.

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreTemplate.java
```diff
@@ -54,6 +55,12 @@
  */
 public class FirestoreTemplate implements FirestoreReactiveOperations {
 
+	private static final StructuredQuery.Projection ID_PROJECTION = StructuredQuery.Projection.newBuilder()
```
elefeint: I'd make `__name__` a public constant, at minimum in this class. Would it also make sense to have a class of public Firestore-related constants that's standalone? Then both reactive and non-reactive functionality can use the same basic constants.

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreTemplate.java
```diff
@@ -75,32 +82,37 @@ public FirestoreTemplate(FirestoreStub firestore, String parent) {
...
+		return Flux.from(idPublisher).next()
+				.flatMap(id -> getDocument(id, entityClass, NAME_ONLY_MASK))
+				.map(d -> d == DocumentReadContext.EMPTY ? Boolean.FALSE : Boolean.TRUE)
+				//.onErrorReturn(throwable -> throwable.getMessage().startsWith("NOT_FOUND: Document"), Boolean.FALSE)
+				.onErrorMap(
+						throwable -> new FirestoreDataException("Unable to test for document existence", throwable));
+	}
+
+	public <T> Mono<T> findById(Publisher idPublisher, Class<T> entityClass) {
+		return findAllById(idPublisher, entityClass).next();
```
elefeint: Should we raise an error if more than one entity is found?

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreTemplate.java
elefeint: Also, `(Flux<String>)` is only necessary because the parameter `Publisher` is not parameterized with actual type.

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreTemplate.java
```diff
@@ -124,34 +136,76 @@ public FirestoreTemplate(FirestoreStub firestore, String parent) {
 						.map(document -> PublicClassMapper.convertToCustomClass(document, clazz)));
 	}
 
+	public <T> Mono<Long> count(Class<T> entityClass) {
+		return findAllDocuments(entityClass, ID_PROJECTION).count();
```
elefeint: This is going to have a performance penalty relative to expected `count()` performance, so we need to document it very well.

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreReactiveOperations.java
```diff
@@ -32,31 +32,85 @@
...
+	 * @return {@link Flux} emitting the saved entities.
 	 */
 	<T> Flux<T> saveAll(Publisher<T> instances);
 
 	/**
 	 * Get all the entities of the given domain type.
 	 * @param <T> the type param of the domain type.
 	 * @param entityClass the domain type to get.
-	 * @return the entities that were found.
+	 * @return {@link Mono} emitting the number of entities.
```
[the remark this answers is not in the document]
dmitry-s: done

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreReactiveOperations.java
```diff
@@ -32,31 +32,85 @@
...
+	 * @return {@link Mono} emitting the number of deleted entities.
 	 */
 	<T> Mono<Long> deleteAll(Class<T> entityClass);
+
+	/**
+	 * Test if the entity of the given domain type with a given id exists.
+	 * @param <T> the type param of the domain type.
+	 * @param entityClass the domain type of the entity.
+	 * @param idPublisher publisher that provides an id.
+	 * @return {@link Mono} emitting {@literal true} if an entity with the given id exists, {@literal false} otherwise.
```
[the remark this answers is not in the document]
dmitry-s: done

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreReactiveOperations.java
```diff
@@ -32,31 +32,85 @@
...
+	<T> Mono<Boolean> existsById(Publisher<String> idPublisher, Class<T> entityClass);
+
+	/**
+	 * Get an entity of the given domain type by id.
+	 * @param <T> the type param of the domain type.
+	 * @param entityClass the domain type of the entity.
+	 * @param idPublisher publisher that provides an id.
+	 * @return {@link Mono} emitting the found entity.
+	 */
+	<T> Mono<T> findById(Publisher idPublisher, Class<T> entityClass);
```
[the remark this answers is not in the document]
dmitry-s: fixed

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreTemplate.java
```diff
@@ -54,6 +55,16 @@
...
 
+	private static final String NAME_FIELD = "__name__";
+
+	private static final StructuredQuery.Projection ID_PROJECTION = StructuredQuery.Projection.newBuilder()
+			.addFields(StructuredQuery.FieldReference.newBuilder().setFieldPath(NAME_FIELD).build())
+			.build();
+
+	private static final DocumentMask NAME_ONLY_MASK = DocumentMask.newBuilder().addFieldPaths(NAME_FIELD).build();
+
+	private static final String NOT_FOUND_DOCUMENT = "NOT_FOUND: Document";
```
dzou: Maybe rename to `NOT_FOUND_DOCUMENT_MESSAGE` to reflect what the constant is used for.

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreTemplate.java
```diff
@@ -75,32 +86,37 @@ public FirestoreTemplate(FirestoreStub firestore, String parent) {
 		this.databasePath = parent.substring(0, StringUtils.ordinalIndexOf(parent, "/", 4));
 	}
 
-	public <T> Mono<T> findById(Publisher idPublisher, Class<T> aClass) {
-		return findAllById(idPublisher, aClass).next();
+	public <T> Mono<Boolean> existsById(Publisher<String> idPublisher, Class<T> entityClass) {
+		return Flux.from(idPublisher).next()
+				.flatMap(id -> getDocument(id, entityClass, NAME_ONLY_MASK))
+				.map(d -> Boolean.TRUE)
```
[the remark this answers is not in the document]
dzou: Btw, this is where i am coming from on the primitives vs. non-primitives thing: https://stackoverflow.com/a/38837450/906497

the code will be more readable using primitives, and in practice I don't think there will be performance difference between using primitives/non-primitives in this specific situation.

Issue #1254, opened 2018-11-26
Remove AppEngineCondition


elefeint:
Replace usage by the new @ConditionalOnGcpEnvironment from #1216 .

PR #1850, opened 2019-09-01
Replace AppEngineCondition for ConditionalOnGcpEnvironment

eddumelendez:
`ConditionalOnGcpEnvironment` allows to validate against which GCP
environment the application is running, so the use of this annotation
with the proper values should be enough. Also, `AppEngineCondition` class is removed.

Fixes gh-1254


spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/sql/GcpCloudSqlAutoConfiguration.java
```diff
@@ -77,7 +78,7 @@
 	 * The AppEngine Configuration for the {@link AppEngineCloudSqlJdbcInfoProvider}.
 	 */
 	@ConditionalOnClass(com.google.cloud.sql.mysql.SocketFactory.class)
-	@Conditional(AppEngineCondition.class)
+	@ConditionalOnGcpEnvironment({GcpEnvironment.APP_ENGINE_FLEXIBLE, GcpEnvironment.APP_ENGINE_STANDARD})
```
meltsufin: The original condition would only match for App Engine Standard, not Flex.
Also, I believe that with the latest version of the Cloud SQL Socket Factory, you don't even need special configuration for App Engine Standard, but I haven't validated that.

eddumelendez: thanks @meltsufin still trying to resolve the issue I have with the usage of `@ConditionalOnGcpEnvironment`.

> Also, I believe that with the latest version of the Cloud SQL Socket Factory, you don't even need special configuration for App Engine Standard, but I haven't validated that.

Are there any way I can validate this?

Thanks in advance.

meltsufin: @kurtisvg Can you confirm that the JDBC connection string format is the same now for both App Engine Standard and Flex/GKE/etc.

kurtisvg: Correct - if you are using the SocketFactory, the JDBC url should be the same regardless of environment.

meltsufin:
Made changes and verified that it still works on App Engine Standard.

elefeint:
Master build is broken; let me look into that before merging.

Issue #1479, opened 2019-02-21
Consider to use java-cfenv for better responsibility distinguishing when we parse VCAP_SERVICES


artembilan:
The logic is handled in the `GcpCloudFoundryEnvironmentPostProcessor`.

See more info in the https://spring.io/blog/2019/02/15/introducing-java-cfenv-a-new-library-for-accessing-cloud-foundry-services

meltsufin:
This is nice!

PR #1852, opened 2019-09-01
Add support for java-cfenv

eddumelendez:
Fixes gh-1479

### DIFF
--- a/spring-cloud-gcp-autoconfigure/pom.xml
+++ b/spring-cloud-gcp-autoconfigure/pom.xml
@@ -117,6 +117,18 @@
             <scope>test</scope>
         </dependency>
 
+        <!-- Java-CFEnv -->
+        <dependency>
+            <groupId>io.pivotal.cfenv</groupId>
+            <artifactId>java-cfenv</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>io.pivotal.cfenv</groupId>
+            <artifactId>java-cfenv-test-support</artifactId>
+            <scope>test</scope>
+        </dependency>
+
         <!-- Storage -->
         <dependency>
             <groupId>org.springframework.cloud</groupId>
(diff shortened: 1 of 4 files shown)


spring-cloud-gcp-autoconfigure/pom.xml
```diff
@@ -117,6 +117,18 @@
             <scope>test</scope>
         </dependency>
 
+        <!-- Java-CFEnv -->
+        <dependency>
+            <groupId>io.pivotal.cfenv</groupId>
+            <artifactId>java-cfenv</artifactId>
```
meltsufin: This should be optional, and we should document that one must add this dependency if they're using CF, or maybe we need to create a starter for CF.

meltsufin:
@artembilan I think it's a little odd to add this non-core-spring dependency for everyone using the project, even if they're not using CF. Should we instead create a tiny CF starter for it? I think that would also advertise support for CF better because we would list it as one of the starters provided by the project. WDYT?

artembilan:
>  Should we instead create a tiny CF starter for it?

Well, let's ask @eddumelendez then!
I think he has some experience with it in other projects since he has come to our side for the fix.

eddumelendez:
I think create a starter for it make sense since starters always provide the right set of dependencies. Also, since no every `spring-cloud-gcp` project is over `cloudfoundry` this will bring an additional dependency which is not needed. For that reasons, I will be in favor of a new starter.

If you both agreed I can work on it in this PR. WDYT?

artembilan:
Good. Then Docs pointer would be great addition to explain what is need when in CF.
Thanks

meltsufin:
Since the creation of the CF module seems a little bit separate from the intents of this PR, how about we merge this one, and then work on the new module in a follow-up PR?

eddumelendez:
ok, cool! I will send another PR then

PR #1853, opened 2019-09-01
Add missing configuration metada

eddumelendez:


artembilan:
Great! LGTM, of course!

Thank you, @eddumelendez , for contribution!
Would you mind to share with us the trick you have used to spot these missing properties?

eddumelendez:
> Would you mind to share with us the trick you have used to spot these missing properties?

Not a big one. I used to look at the `*AutoConfiguration` classes and look for properties at `@ConditionalOnProperty` which are not so many. Then, look if they are all added in the `additional-spring-configuration-metadata.json`

PR #1860, opened 2019-09-05
Add BigQuery to CloudFoundry configuration

dzou:
Adds BigQuery to the CloudFoundry configuration.


spring-cloud-gcp-autoconfigure/src/test/resources/VCAP_SERVICES
```diff
@@ -122,6 +122,29 @@
      "volume_mounts": []
    }
   ],
+  "google-bigquery": [
```
meltsufin: What is this test based on? We need to sue the real format for BigQuery in CF.
I believe it should look something like this:
```
 "VCAP_SERVICES": {
  "google-bigquery": [
   {
    "credentials": {
     "Email": "pcf-binding-1234asdf@my-gcp-project.iam.gserviceaccount.com",
     "Name": "pcf-binding-1234asdf",
     "PrivateKeyData": "ewogICJ0eXBlIjogInNlcnZpY2VfYWNjb3VudCIsCiAgInByb2plY3RfaWQiOiAicGRlLW1==",
     "ProjectId": "my-gcp-project",
     "UniqueId": "105423123456789196414",
     "dataset_id": "pcf_sb_2_1518504199123456789"
    },
    "label": "google-bigquery",
    "name": "mydataset",
    "plan": "default",
    "provider": null,
    "syslog_drain_url": null,
    "tags": [
     "gcp",
     "bigquery"
    ],
    "volume_mounts": []
   }
  ],
```
Based on this comment: https://github.com/spring-cloud/spring-cloud-gcp/issues/255#issuecomment-365428049

You then need to make sure that dataset is extracted correctly, I believe.

elefeint: It's true what they say about US being an overly litigious society; we even sue configuration formats :rofl:

dzou: Hmm, I'm not quite sure how we use the dataset_id value for us. Is the idea that we parse the dataset_id, then configure a message handler in our autoconfiguration and pass our dataset_id into our message handler?

However I don't see any Spring Integration channel adapters/message handlers being configured for pub/sub or storage.

meltsufin: Maybe just add the `spring.cloud.gcp.bigquery.datasetId` property to the environment?

dzou:
> @dzou The CF starter has been merged. So, you can now rebase on top of that.

Will do - I'll revisit this PR once the BigQueryTemplate is in; this allows us to use the dataset_id setting passed in by CF.

PR #1866, opened 2019-09-10
Firestore repo factory

ChengyuanZhao:
Adds config boilerplate for Firestore repos


ChengyuanZhao:
@dmitry-s There is that test in the autoconfig module I included that requests a bean of the arbitrary `FirestoreTestRepo` class I defined. So that one does verify user-defined repos can be generated.

As for actual CI of query method functionality i don't think that's possible at the moment since the PartTree class in this PR is just a placeholder :/

dmitry-s:
@ChengyuanZhao The repository methods like count, findAll, save, etc. are already implemented in FirestoreReactiveRepository - can we test a couple of those?

ChengyuanZhao:
@dmitry-s yes can do.

ChengyuanZhao:
@dmitry-s PTAL added integration tests

ChengyuanZhao:
@meltsufin @dmitry-s merged !

PR #1868, opened 2019-09-10
Add configuration properties generation

### LINKED ISSUE #1863: Update docs to use absolute urls using asciidoc attributes
See contract commit spring-cloud/spring-cloud-sleuth@590ffa3

ryanjbaxter:
Fixes #1863

### DIFF
--- a/docs/src/main/asciidoc/_configprops.adoc
+++ b/docs/src/main/asciidoc/_configprops.adoc
@@ -1,0 +1,110 @@
+|===
+|Name | Default | Description
+
+|spring.cloud.gcp.bigquery.credentials.default-scopes |  | 
+|spring.cloud.gcp.bigquery.enabled | true | Auto-configure Google Cloud BigQuery components.
+|spring.cloud.gcp.bigquery.project-id |  | 
+|spring.cloud.gcp.config.credentials.default-scopes |  | 
+|spring.cloud.gcp.config.enabled | false | Enables Spring Cloud GCP Config.
+|spring.cloud.gcp.config.name |  | Name of the application.
+|spring.cloud.gcp.config.profile |  | Comma-delimited string of profiles under which the app is running. Gets its default value from the {@code spring.profiles.active} property, falling back on the {@code spring.profiles.default} property.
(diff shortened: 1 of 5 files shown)


docs/src/main/asciidoc/_configprops.adoc
```diff
@@ -0,0 +1,110 @@
+|===
+|Name | Default | Description
+
+|spring.cloud.gcp.bigquery.credentials.default-scopes |  | 
+|spring.cloud.gcp.bigquery.enabled | true | Auto-configure Google Cloud BigQuery components.
```
artembilan: Does this mean that such a file is generated from the source code?
And whenever we change the code this one is going to be regenerated.

My point is that, please, pay attention that `default-scopes` goes without description and I believe that has to be fixed.

I don't say that you have to do this in this PR: just want to be sure what is needed to be done in the future.

Thanks

docs/src/main/asciidoc/_configprops.adoc
```diff
@@ -0,0 +1,110 @@
...
+|spring.cloud.gcp.bigquery.project-id |  | 
+|spring.cloud.gcp.config.credentials.default-scopes |  | 
+|spring.cloud.gcp.config.enabled | false | Enables Spring Cloud GCP Config.
+|spring.cloud.gcp.config.name |  | Name of the application.
+|spring.cloud.gcp.config.profile |  | Comma-delimited string of profiles under which the app is running. Gets its default value from the {@code spring.profiles.active} property, falling back on the {@code spring.profiles.default} property.
```
artembilan: Don't we need to be as simple as possible in the configuration properties descriptions?
I mean we should avoid to use any JavaDocs tag: we can't be sure what  tool is going to render such a description.
So, tags like this (or that `<code>`) below could break the target tool.

docs/src/main/asciidoc/_configprops.adoc
ryanjbaxter: Yeah this PR is just to enable the auto generation of the file and adding a link to the docs for it.  If there are issues with the specific properties we should fix them separately

PR #1869, opened 2019-09-10
Modify FirestoreTemplate to use Batched Deletes

### LINKED ISSUE #1798: Use write(WriteRequest) for batch DeleteAll in Firestore
the current implementation issues multiple single-delete requests. There seems to be the generic write(WriteRequest) method for batch update/delete/save

dzou:
Modify the FirestoreTemplate to use batched deletes so that deletions of larger number of entities is efficient.

Fixes #1798.

### DIFF
--- a/spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreTemplate.java
+++ b/spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreTemplate.java
@@ -22,7 +22,6 @@ import java.util.Map;
 
 import com.google.cloud.firestore.PublicClassMapper;
 import com.google.firestore.v1.CreateDocumentRequest;
-import com.google.firestore.v1.DeleteDocumentRequest;
 import com.google.firestore.v1.Document;
 import com.google.firestore.v1.DocumentMask;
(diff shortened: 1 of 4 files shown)


spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreTemplate.java
```diff
@@ -194,30 +191,81 @@ public int getSaveAllBufferWriteSize() {
...
 		return Mono.defer(() -> {
-			FirestorePersistentEntity<?> persistentEntity = this.mappingContext.getPersistentEntity(entityClass);
-			return Flux.from(idPublisher).flatMap(id -> callDelete(buildResourceName(persistentEntity, id)))
+			FirestorePersistentEntity<?> persistentEntity =
+					this.mappingContext.getPersistentEntity(entityClass);
+
+			Flux<List<String>> documentBuffers =
+					Flux.from(idPublisher)
+							.map(id -> buildResourceName(persistentEntity, id))
+							.bufferTimeout(this.writeBufferSize, this.writeBufferTimeout);
```
dmitry-s: the following multiple lines are duplicated - maybe move that to a separate method?

dzou: Done.

PR #1870, opened 2019-09-10
Upgrade to zipkin-gcp 0.14

elefeint:
Upgrading to `zipkin-gcp` 0.14 automatically upgrades our Stackdriver API to V2, which is technically still in Beta.

I had to keep the brave version workaround because `zipkin-gcp` explicitly downgrades it: openzipkin/zipkin-gcp#146

### DIFF
--- a/spring-cloud-gcp-samples/spring-cloud-gcp-trace-sample/pom.xml
+++ b/spring-cloud-gcp-samples/spring-cloud-gcp-trace-sample/pom.xml
@@ -14,10 +14,6 @@
         <version>1.2.0.BUILD-SNAPSHOT</version>
     </parent>
 
-    <properties>
-        <grpc-google-cloud-trace.version>0.69.0</grpc-google-cloud-trace.version>
-    </properties>
-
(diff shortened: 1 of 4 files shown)


spring-cloud-gcp-samples/spring-cloud-gcp-trace-sample/pom.xml
```diff
@@ -14,6 +14,10 @@
         <version>1.2.0.BUILD-SNAPSHOT</version>
     </parent>
 
+    <properties>
+        <grpc-google-cloud-trace.version>0.69.0</grpc-google-cloud-trace.version>
```
meltsufin: Hmm, so unless you do this, everything breaks? How will our users know to do this downgrade? Can we just exclude the transitive import from `zipking-gcp` dependency?

elefeint: This is just a test dependency, so we can query Stackdriver (v2 does not have get/list yet). But actually, the version is not necessary at all -- it's brought in through cloud BOM by way of `spring-cloud-gcp-dependencies`. 
I'll remove the property.

PR #1874, opened 2019-09-11
Fix maven schema

eddumelendez:


eddumelendez:
Thanks @meltsufin. is this something just for `spring-cloud-gcp`? I looked into other spring projects and they are keeping it the very same way the PR is doing.

pom.xml
```diff
@@ -373,78 +373,78 @@
...
-									<artifactId>error_prone_core</artifactId>
-									<version>2.3.1</version>
-								</dependency>
-							</dependencies>
-						</plugin>
-					</plugins>
-				</pluginManagement>
-			</build>
-		</profile>
+<!--		<profile>-->
```
dzou: Should this section be uncommented?

meltsufin:
@eddumelendez You're right, I just noticed that other Spring Cloud projects have switched back to HTTP. @artembilan What's the current guidance?

pom.xml
eddumelendez: done!

spring-cloud-gcp-cloudfoundry/pom.xml
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project xmlns="http://maven.apache.org/POM/4.0.0"
          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
```
dzou: Should this one stay as http? :)

eddumelendez: No, it shouldn't. I am following other spring projects

PR #1875, opened 2019-09-11
Add Autowiring for FirestoreTemplate

### LINKED ISSUE #1780: Autowire FirestoreTemplate

dzou:
Autowire the `FirestoreTemplate` in autoconfiguration.

Fixes #1780.

### DIFF
--- a/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/firestore/GcpFirestoreAutoConfiguration.java
+++ b/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/firestore/GcpFirestoreAutoConfiguration.java
@@ -23,6 +23,10 @@ import com.google.api.gax.rpc.TransportChannelProvider;
 import com.google.cloud.TransportOptions;
 import com.google.cloud.firestore.Firestore;
 import com.google.cloud.firestore.FirestoreOptions;
+import com.google.firestore.v1.FirestoreGrpc;
+import io.grpc.ManagedChannel;
+import io.grpc.ManagedChannelBuilder;
+import io.grpc.auth.MoreCallCredentials;
 
 import org.springframework.beans.factory.ObjectProvider;
 import org.springframework.boot.autoconfigure.AutoConfigureAfter;
@@ -33,6 +37,7 @@ import org.springframework.boot.context.properties.EnableConfigurationProperties
 import org.springframework.cloud.gcp.autoconfigure.core.GcpContextAutoConfiguration;
 import org.springframework.cloud.gcp.core.DefaultCredentialsProvider;
 import org.springframework.cloud.gcp.core.GcpProjectIdProvider;
+import org.springframework.cloud.gcp.core.UserAgentHeaderProvider;
 import org.springframework.cloud.gcp.data.firestore.FirestoreTemplate;
 import org.springframework.cloud.gcp.data.firestore.mapping.FirestoreMappingContext;
 import org.springframework.context.annotation.Bean;
@@ -52,9 +57,16 @@ import org.springframework.context.annotation.Configuration;
 @EnableConfigurationProperties(GcpFirestoreProperties.class)
 public class GcpFirestoreAutoConfiguration {
 
+	private static final String ROOT_PATH_FORMAT = "projects/%s/databases/(default)/documents";
+
 	private final String projectId;
 
 	private final CredentialsProvider credentialsProvider;
+
+	private final String firestoreRootPath;
+
+	private final UserAgentHeaderProvider headerProvider =
+			new UserAgentHeaderProvider(GcpFirestoreAutoConfiguration.class);
 
 	GcpFirestoreAutoConfiguration(GcpFirestoreProperties gcpFirestoreProperties,
 			GcpProjectIdProvider projectIdProvider,
@@ -67,6 +79,8 @@ public class GcpFirestoreAutoConfiguration {
 		this.credentialsProvider = (gcpFirestoreProperties.getCredentials().hasKey()
 				? new DefaultCredentialsProvider(gcpFirestoreProperties)
 				: credentialsProvider);
+
+		this.firestoreRootPath = String.format(ROOT_PATH_FORMAT, this.projectId);
 	}
 
 	@Bean
@@ -75,7 +89,8 @@ public class GcpFirestoreAutoConfiguration {
 			ObjectProvider<TransportChannelProvider> transportChannelProviderObjectProvider) {
 		FirestoreOptions.Builder builder = FirestoreOptions.getDefaultInstance().toBuilder()
 				.setCredentialsProvider(this.credentialsProvider)
-				.setProjectId(this.projectId);
+				.setProjectId(this.projectId)
+				.setHeaderProvider(this.headerProvider);
 
 		return builder.build();
 	}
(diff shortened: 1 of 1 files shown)


spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/firestore/GcpFirestoreAutoConfiguration.java
```diff
@@ -52,10 +57,17 @@
 @EnableConfigurationProperties(GcpFirestoreProperties.class)
 public class GcpFirestoreAutoConfiguration {
 
+	private static final String ROOT_PATH_FORMAT = "projects/%s/databases/(default)/documents";
+
 	private final String projectId;
 
 	private final CredentialsProvider credentialsProvider;
 
+	private final String firestoreRootPath;
```
meltsufin: This should probably be configurable through `GcpFirestoreProperties`. WDYT?

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/firestore/GcpFirestoreAutoConfiguration.java
```diff
@@ -94,8 +109,17 @@ public FirestoreMappingContext firestoreMappingContext() {
 
 	@Bean
 	@ConditionalOnMissingBean
-	public FirestoreTemplate firestoreTemplate() {
-		// Placeholder for until auto-config is complete.
-		return new FirestoreTemplate(null, "a/b/c/d/e");
+	public FirestoreTemplate firestoreTemplate() throws IOException {
+		ManagedChannel channel = ManagedChannelBuilder
+				.forAddress("firestore.googleapis.com", 443)
```
meltsufin: This should be configurable via `GcpFirestoreProperties.host` just like in Datastore so that one can switch to an emulator.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/firestore/GcpFirestoreAutoConfiguration.java
```diff
@@ -94,8 +109,17 @@ public FirestoreMappingContext firestoreMappingContext() {
...
+				.userAgent(this.headerProvider.getUserAgent())
+				.build();
+
+		FirestoreGrpc.FirestoreStub firestoreStub =
```
meltsufin: In the interests of flexible configuration, this could be a separate bean.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/firestore/GcpFirestoreAutoConfiguration.java
```diff
@@ -94,8 +120,17 @@ public FirestoreMappingContext firestoreMappingContext() {
...
-		// Placeholder for until auto-config is complete.
-		return new FirestoreTemplate(null, "a/b/c/d/e");
+	public FirestoreTemplate firestoreTemplate(
+			FirestoreGrpc.FirestoreStub firestoreStub) throws IOException {
+		return new FirestoreTemplate(firestoreStub, this.firestoreRootPath);
+	}
+
+	@Bean
+	@ConditionalOnMissingBean
+	ManagedChannel firestoreManagedChannel() {
```
meltsufin:
```suggestion
	public ManagedChannel firestoreManagedChannel() {
```

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/firestore/GcpFirestoreAutoConfiguration.java
```diff
@@ -52,10 +57,17 @@
...
 
+	private static final String ROOT_PATH_FORMAT = "projects/%s/databases/(default)/documents";
+
 	private final String projectId;
 
 	private final CredentialsProvider credentialsProvider;
 
+	private final String firestoreRootPath;
+
+	private final UserAgentHeaderProvider headerProvider =
```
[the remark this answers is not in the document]
dzou: I made this into a constant for the class because it is used in 2 places

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/firestore/GcpFirestoreAutoConfiguration.java
dzou: Done. called it "hostPort" since it includes host:port.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/firestore/GcpFirestoreAutoConfiguration.java
dzou: Done.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/firestore/GcpFirestoreProperties.java
```diff
@@ -39,6 +39,11 @@
 
 	private String projectId;
 
+	/**
+	 * The host and port of the Firestore emulator service; can be overridden to specify an emulator
+	 */
+	private String hostPort = "firestore.googleapis.com:443";
```
meltsufin: In `GcpDatastoreProperties`, we just call it `host`. Can we stick with that convention please?

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/firestore/GcpFirestoreAutoConfiguration.java
```diff
@@ -67,17 +80,22 @@
 		this.credentialsProvider = (gcpFirestoreProperties.getCredentials().hasKey()
 				? new DefaultCredentialsProvider(gcpFirestoreProperties)
 				: credentialsProvider);
+
+		this.hostPort = gcpFirestoreProperties.getHostPort();
+		this.firestoreRootPath = String.format(ROOT_PATH_FORMAT, this.projectId);
```
meltsufin: Can `firestoreRootPath` optionally be overwritten by `GcpFirestoreProperties` or it doesn't make sense?
Maybe all you need the user to be able to overwrite is the `(default)` part.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/firestore/GcpFirestoreAutoConfiguration.java
dzou: Hmm one thing I was thinking was to have the autoconfiguration infer the root path with the form `projects/{PROJECT_ID}/databases/(default)/documents` and then extending our template to allow the user to specify a relative path from the root. I don't want the user to have to configure it because it is confusing to figure out the root path format pattern.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/firestore/GcpFirestoreProperties.java
meltsufin: Actually, maybe it's Datastore that needs to be fixed.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/firestore/GcpFirestoreAutoConfiguration.java
dzou: In the client libs for `FirestoreImpl` you can see that the root path is inferred based on the machine's environment: https://github.com/googleapis/google-cloud-java/blob/61a759a4cea68b5db662dbbcce4447824d53483a/google-cloud-clients/google-cloud-firestore/src/main/java/com/google/cloud/firestore/FirestoreImpl.java#L87

(they build the path from the projectId and databaseId which is always `(default)`).

So as discussed offline, it's like a "prefix" as you mentioned. We could make the prefix configurable, but I hesitate because it's hard to get it right, you have to figure out that the databaseId is *always* `(default)` which requires some digging... So I prefer just to let users do everything relative to the root to keep it simple.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/firestore/GcpFirestoreProperties.java
dzou: Hmm, so should we change this to `host` or leave as `hostPort`? Am fine either way; I guess changing datastore might break some people? :/

PR #1883, opened 2019-09-12
Add Health Check call for Stackdriver AsyncReporter

### LINKED ISSUE #1786: Call AsyncReporter.check() when creating stackdriver beans
spring-cloud-gcp does not rely on `spring-cloud-sleuth-zipkin` (historical reasons?), so after openzipkin/zipkin-gcp#131 is done, we'll have to implement our own version of spring-cloud/spring-cloud-sleuth#1411 when instantiating the [reporter bean](https://github.com/spring-cloud/spring-cloud-gcp/blob/2d33d096546cdf935c3ac3e5e701a9f167f2ef2a/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/trace/StackdriverTraceAutoConfiguration.java#L138)

Longer discussion is -- should we rely on `spring-cloud-sleuth-zipkin` in the first place?

This is blocked on #1862.

dzou:
Adds the Healthcheck `.check()` call for the Stackdriver `AsyncReporter`.

Fixes #1786.


dzou:
> If you say "elefeint" three times, a wild review appears.

Lol :smile:

PR #1888, opened 2019-09-13
Delegate FirestoreTemplate.save() to call .saveAll()

### LINKED ISSUE #1796: Firestore update/updateAll

dzou:
Delegates the FirestoreTemplate.saveAll() method to calling `.save()`. This is done because saveAll() acts as an Upsert and covers the update functionality described in #1796 whereas the former method `CreateDocumentRequest` is used for document creation only.

Fixes #1796.

I moved the save() test to be an integration test because it is difficult to construct mocks to properly unit test bidi streams in `saveAll()`.


dzou:
> Few things are better than a PR that deletes more code than it creates.

Lol, indeed.

Issue #1896, opened 2019-09-16
PubSubTemplateIntegrationTests.testPullAndAck integration test failing


meltsufin:
The ack'ed message seems to get redelivered as though ack'ing doesn't work.

PR #1898, opened 2019-09-16
Fix Pub/Sub batch acking

meltsufin:
Fixes: #1896.

### DIFF
--- a/spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/pubsub/it/PubSubTemplateIntegrationTests.java
+++ b/spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/pubsub/it/PubSubTemplateIntegrationTests.java
@@ -149,6 +149,7 @@ public class PubSubTemplateIntegrationTests {
 				}
 				catch (InterruptedException | ExecutionException | TimeoutException ex) {
 					LOGGER.error(ex);
+					Thread.currentThread().interrupt();
 				}
 			});
 
@@ -175,6 +176,7 @@ public class PubSubTemplateIntegrationTests {
 				}
 				catch (InterruptedException | ExecutionException ex) {
 					LOGGER.error(ex);
+					Thread.currentThread().interrupt();
 				}
 			});
 
--- a/spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/pubsub/it/PubSubTemplateIntegrationTests.java
+++ b/spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/pubsub/it/PubSubTemplateIntegrationTests.java
@@ -160,11 +160,16 @@ public class PubSubTemplateIntegrationTests {
 			assertThat(messagesSet.size()).as("check that we received all the messages").isEqualTo(3);
 
 			ackableMessages.forEach((message) -> {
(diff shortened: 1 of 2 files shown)


spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberTemplate.java
```diff
@@ -339,7 +339,7 @@ public void destroy() {
 
 		groupedMessages.forEach((ProjectSubscriptionName psName, List<String> ackIds) -> {
 
-			ApiFuture<Empty> ackApiFuture = asyncOperation.apply(psName.getSubscription(), ackIds);
```
meltsufin: This used to work before #1880 because there was another bug that was cancelling this one out.
We need to pass the fully-qualified subscription name, which you get by calling `toString()` on `ProjectSubscriptionName`.

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/pubsub/it/PubSubTemplateIntegrationTests.java
```diff
@@ -160,11 +160,16 @@ public void testPullAndAck() {
...
+						message.ack().get(); // sync call
+					}
+					else {
+						message.nack().get(); // sync call
+					}
 				}
-				else {
-					message.nack(); //sync call
+				catch (InterruptedException | ExecutionException e) {
+					e.printStackTrace();
```
artembilan: `e.printStackTrace();` is not good.
You should log error or re-throw.
Also never swallow `InterruptedException`. At least cal `Thread.currentThread().interrupt()` in this `catch` block.

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/pubsub/it/PubSubTemplateIntegrationTests.java
```diff
@@ -168,8 +173,8 @@ public void testPullAndAck() {
 						message.nack().get(); // sync call
 					}
 				}
-				catch (InterruptedException | ExecutionException e) {
-					e.printStackTrace();
+				catch (InterruptedException | ExecutionException ex) {
+					LOGGER.error(ex);
```
artembilan: But you still don't `interrupt()`...

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/pubsub/it/PubSubTemplateIntegrationTests.java
meltsufin: Done.

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/pubsub/it/PubSubTemplateIntegrationTests.java
artembilan: Although this is a test code, so never mind.
Let's merge!

meltsufin: sorry, forgot, it's set now

PR #1899, opened 2019-09-16
Fix FirestoreSampleAppTests

meltsufin:
Without this fix, we get:
```
Caused by: java.lang.ClassNotFoundException: com.google.firestore.v1.FirestoreGrpc$FirestoreStub
	at java.net.URLClassLoader.findClass(URLClassLoader.java:381)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:335)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
	... 84 common frames omitted
```

### DIFF
--- a/spring-cloud-gcp-autoconfigure/pom.xml
+++ b/spring-cloud-gcp-autoconfigure/pom.xml
@@ -176,6 +176,11 @@
             <optional>true</optional>
         </dependency>
 
+        <dependency>
+            <groupId>com.google.api.grpc</groupId>
+            <artifactId>grpc-google-cloud-firestore-v1</artifactId>
+            <optional>true</optional>
+        </dependency>
 
         <!-- BigQuery -->
         <dependency>
--- a/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/firestore/GcpFirestoreAutoConfiguration.java
+++ b/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/firestore/GcpFirestoreAutoConfiguration.java
@@ -109,7 +109,6 @@ public class GcpFirestoreAutoConfiguration {
 	 * The Firestore reactive template and data repositories support auto-configuration.
 	 */
 	@ConditionalOnClass({ FirestoreGrpc.FirestoreStub.class })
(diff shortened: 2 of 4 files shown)


spring-cloud-gcp-autoconfigure/pom.xml
```diff
@@ -176,6 +176,11 @@
             <optional>true</optional>
         </dependency>
 
+        <dependency>
+            <groupId>com.google.api.grpc</groupId>
+            <artifactId>grpc-google-cloud-firestore-v1</artifactId>
+            <optional>true</optional>
```
elefeint: This is not brought in transitively from `com.google.cloud:google-cloud-firestore`?

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/firestore/GcpFirestoreAutoConfiguration.java
```diff
@@ -104,32 +105,40 @@ public Firestore firestore(FirestoreOptions firestoreOptions) {
...
-	public ManagedChannel firestoreManagedChannel() {
-		return ManagedChannelBuilder
-				.forTarget(this.hostPort)
-				.userAgent(USER_AGENT_HEADER_PROVIDER.getUserAgent())
-				.build();
+	/**
+	 * The Firestore reactive template and data repositories support auto-configuration.
+	 */
+	@ConditionalOnClass({ FirestoreGrpc.FirestoreStub.class })
+	@ConditionalOnMissingBean(CloudSqlJdbcInfoProvider.class)
```
elefeint: Is `CloudSqlJdbcInfoProvider` the right bean?

dzou: Is doing conditional on missing bean `CloudSqlJdbcInfoProvider.class` intentional? Or is there a different bean you wanted to specify?

spring-cloud-gcp-autoconfigure/pom.xml
meltsufin: Good catch!

Issue #1894, opened 2019-09-16
Firestore tests failing on Jenkins


meltsufin:
We see an error like this:
```
Corrupted STDOUT by directly writing to native stream in forked JVM 1
```
[Example build](https://jenkins.spring.io/blue/organizations/jenkins/spring-cloud-gcp-master-ci/detail/spring-cloud-gcp-master-ci/2560/pipeline).

Currently, unable to reproduce locally or on Travis.

meltsufin:
It looks like `Corrupted STDOUT...` is just a red herring. The real problem is that the test simply hangs.

PR #1900, opened 2019-09-16
Fix failing tests on Jenkins

meltsufin:
Fixes: #1894.


meltsufin:
One key insight was turning on Surefire test output to the console rather than file. The error messages were all there. We just can't easily see them when the output goes into a file on Jenkins.

PR #1901, opened 2019-09-17
Improve Spanner readonly transactional test

ChengyuanZhao:
related #1897

makes fix and improves closing of readonly transactions.

Previously it was still located in a block that expected the transaction state to be STARTED, but this is not correct. Test improved to account for this.

### DIFF
--- a/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTransactionManager.java
+++ b/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTransactionManager.java
@@ -178,6 +178,7 @@ public class SpannerTransactionManager extends AbstractPlatformTransactionManage
 		}
 		else {
 			tx.transactionContext = tx.transactionManager.begin();
+			tx.isReadOnly = false;
 		}
 
 		logger.debug(tx + " begin; state = " + tx.transactionManager.getState());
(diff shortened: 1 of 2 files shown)


spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTransactionManager.java
```diff
@@ -100,6 +100,7 @@ protected void doBegin(Object transactionObject, TransactionDefinition transacti
 		if (transactionDefinition.isReadOnly()) {
 			final ReadContext targetTransactionContext = this.databaseClientProvider.get()
 					.readOnlyTransaction();
+			tx.isReadOnly = true;
```
meltsufin: Don't you need to set it to `false` in the `else`, since this is a thread-local variable and will be recycled.

ChengyuanZhao: Will do. I didn't know that about thread-local variables :)

ChengyuanZhao: Done.

srivraokp:
@meltsufin @ChengyuanZhao 
I debugged the code it opened 2 sessions for 1 request and one session has leak expection
Attached is the screen for variables in memory
SessionLeak

meltsufin:
@srivraokp Let's continue the conversation in issue #1904. Thanks!

PR #1906, opened 2019-09-17
leak fix read only spanner

### LINKED ISSUE #1904: Spring Data Spanner Leaking Sessions
Follow-up to: #1887, #1897, #1901, and specifically [this comment](https://github.com/spring-cloud/spring-cloud-gcp/pull/1901#issuecomment-532352484).

@ChengyuanZhao @srivraokp

ChengyuanZhao:
fixes #1904 

It turns out that getting a transaction manager counts as using a session (not just getting a transaction from the manager)

I added integration test that limits sessions to 10 and then tries 20 operations with timeline of 10 seconds to catch hangs.


srivraokp:
@ChengyuanZhao @meltsufin 
I took the code from spanner-leak-fix branch and it works fine, thanks for fixing the same

Can you please let us know when this will be in release version, we have a prod deployment on 10/16
Thanks in advance

srivraokp:
@ChengyuanZhao 
Any update on the release version for this Fix

As this is critical for us to move to Production -- Thanks

meltsufin:
1.2.0.M3 is schedule for Monday.

srivraokp:
Thanks

PR #1916, opened 2019-09-19
Add BigQueryTemplate to BigQuery Module

dzou:
This refactors the BigQueryFileMessageHandler to extract the complex work of writing data to tables into a template class.

The template class is now scoped to the BigQuery `datasetName`.

Note: In BigQuery, hierarchy of data is as follows:
GCP Project Account > Dataset Name > Table Name

### DIFF
--- a/spring-cloud-gcp-bigquery/src/main/java/org/springframework/cloud/gcp/bigquery/core/BigQueryOperations.java
+++ b/spring-cloud-gcp-bigquery/src/main/java/org/springframework/cloud/gcp/bigquery/core/BigQueryOperations.java
@@ -1,0 +1,47 @@
+/*
+ * Copyright 2017-2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.gcp.bigquery.core;
+
+import java.io.IOException;
+import java.io.InputStream;
+
+import com.google.cloud.bigquery.FormatOptions;
+import com.google.cloud.bigquery.Job;
+
+import org.springframework.util.concurrent.ListenableFuture;
+
+/**
+ * Defines operations for use with BigQuery.
+ *
+ * @author Daniel Zou
+ * @since 1.2
+ */
+public interface BigQueryOperations {
+
+	/**
+	 * Writes data to a specified BigQuery table.
+	 *
+	 * @param tableName name of the table to write to
+	 * @param inputStream input stream of the table data to write
+	 * @param dataFormatOptions the format of the data to write
+	 * @return {@link ListenableFuture} containing the BigQuery Job indicating completion of
+	 * operation
+	 * @throws IOException if the data fails to be loaded
+	 */
+	ListenableFuture<Job> writeDataToTable(
+			String tableName, InputStream inputStream, FormatOptions dataFormatOptions) throws IOException;
+}
--- a/spring-cloud-gcp-bigquery/src/main/java/org/springframework/cloud/gcp/bigquery/core/BigQueryOperations.java
+++ b/spring-cloud-gcp-bigquery/src/main/java/org/springframework/cloud/gcp/bigquery/core/BigQueryOperations.java
@@ -16,7 +16,6 @@
 
 package org.springframework.cloud.gcp.bigquery.core;
 
-import java.io.IOException;
 import java.io.InputStream;
 
 import com.google.cloud.bigquery.FormatOptions;
@@ -40,8 +39,9 @@ public interface BigQueryOperations {
 	 * @param dataFormatOptions the format of the data to write
 	 * @return {@link ListenableFuture} containing the BigQuery Job indicating completion of
 	 * operation
-	 * @throws IOException if the data fails to be loaded
+	 *
+	 * @throws BigQueryException if errors occur when loading data to the BigQuery table
 	 */
 	ListenableFuture<Job> writeDataToTable(
-			String tableName, InputStream inputStream, FormatOptions dataFormatOptions) throws IOException;
+			String tableName, InputStream inputStream, FormatOptions dataFormatOptions);
 }
--- a/spring-cloud-gcp-bigquery/src/main/java/org/springframework/cloud/gcp/bigquery/core/BigQueryTemplate.java
+++ b/spring-cloud-gcp-bigquery/src/main/java/org/springframework/cloud/gcp/bigquery/core/BigQueryTemplate.java
@@ -14,12 +14,8 @@
  * limitations under the License.
  */
 
-package org.springframework.cloud.gcp.bigquery.integration;
+package org.springframework.cloud.gcp.bigquery.core;
 
-import java.io.BufferedInputStream;
-import java.io.ByteArrayInputStream;
-import java.io.File;
-import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
@@ -45,8 +41,9 @@ import org.springframework.util.concurrent.SettableListenableFuture;
  * Helper class which simplifies common operations done in BigQuery.
  *
  * @author Daniel Zou
+ * @since 1.2
  */
-public class BigQueryTemplate {
+public class BigQueryTemplate implements BigQueryOperations {
 
 	private final BigQuery bigQuery;
 
@@ -60,6 +57,14 @@ public class BigQueryTemplate {
 
 	private Duration jobPollInterval = Duration.ofSeconds(2);
 
+	/**
+	 * Creates the {@link BigQuery} template..
+	 *
+	 * @param bigQuery the underlying client object used to interface with BigQuery
+	 * @param datasetName the name of the dataset in which all operations will take place
+	 * @param taskScheduler the {@link TaskScheduler} used to poll for the status of
+	 *     long-running BigQuery operations
+	 */
 	public BigQueryTemplate(BigQuery bigQuery, String datasetName, TaskScheduler taskScheduler) {
 		this.bigQuery = bigQuery;
 		this.datasetName = datasetName;
@@ -68,9 +73,10 @@ public class BigQueryTemplate {
 
 	/**
 	 * Sets whether BigQuery should attempt to autodetect the schema of the data when loading
-	 * data into an empty table for the first time.
-	 * @param autoDetectSchema whether data schema should be autodetected from the structure of
-	 * 		the data. Default is true.
+	 * data into an empty table for the first time. If set to false, the schema must be
+	 * defined explicitly for the table before load.
+	 * @param autoDetectSchema whether data schema should be autodetected from the structure
+	 *     of the data. Default is true.
 	 */
 	public void setAutoDetectSchema(boolean autoDetectSchema) {
 		this.autoDetectSchema = autoDetectSchema;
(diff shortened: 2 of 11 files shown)


spring-cloud-gcp-bigquery/src/main/java/org/springframework/cloud/gcp/bigquery/integration/BigQueryTemplate.java
```diff
@@ -0,0 +1,196 @@
...
+
+	private final TaskScheduler taskScheduler;
+
+	private boolean autoDetectSchema = true;
+
+	private WriteDisposition writeDisposition = WriteDisposition.WRITE_APPEND;
+
+	private Duration jobPollInterval = Duration.ofSeconds(2);
+
+	public BigQueryTemplate(BigQuery bigQuery, String datasetName, TaskScheduler taskScheduler) {
```
meltsufin: Javadoc?

spring-cloud-gcp-bigquery/src/main/java/org/springframework/cloud/gcp/bigquery/integration/BigQueryTemplate.java
```diff
@@ -0,0 +1,196 @@
...
+		this.bigQuery = bigQuery;
+		this.datasetName = datasetName;
+		this.taskScheduler = taskScheduler;
+	}
+
+	/**
+	 * Sets whether BigQuery should attempt to autodetect the schema of the data when loading
+	 * data into an empty table for the first time.
+	 * @param autoDetectSchema whether data schema should be autodetected from the structure of
```
meltsufin: What will happen if it's `false`?

spring-cloud-gcp-bigquery/src/main/java/org/springframework/cloud/gcp/bigquery/integration/BigQueryTemplate.java
```diff
@@ -0,0 +1,196 @@
...
+	 * 		the data. Default is true.
+	 */
+	public void setAutoDetectSchema(boolean autoDetectSchema) {
+		this.autoDetectSchema = autoDetectSchema;
+	}
+
+	/**
+	 * Sets the {@link WriteDisposition} which specifies how data should be inserted into
+	 * BigQuery tables.
+	 * @param writeDisposition whether data should be appended or truncated to the BigQuery table.
```
meltsufin: What does "truncated" mean? overwritten?

spring-cloud-gcp-bigquery/src/main/java/org/springframework/cloud/gcp/bigquery/integration/outbound/BigQueryFileMessageHandler.java
```diff
@@ -168,7 +130,7 @@ public void setFormatOptions(FormatOptions formatOptions) {
 	 */
 	public void setWriteDisposition(WriteDisposition writeDisposition) {
```
meltsufin: I don't think you need this method anymore and the two below, because they can just set it directly on the template. You can create a getter for the template if you want.

spring-cloud-gcp-bigquery/src/main/java/org/springframework/cloud/gcp/bigquery/integration/BigQueryTemplate.java
```diff
@@ -0,0 +1,196 @@
...
+import org.springframework.util.StreamUtils;
+import org.springframework.util.concurrent.ListenableFuture;
+import org.springframework.util.concurrent.SettableListenableFuture;
+
+/**
+ * Helper class which simplifies common operations done in BigQuery.
+ *
+ * @author Daniel Zou
+ */
+public class BigQueryTemplate {
```
meltsufin: Can you extract the interface to `BigQueryOperations`?
All javadocs of public methods would move there.

spring-cloud-gcp-bigquery/src/main/java/org/springframework/cloud/gcp/bigquery/integration/BigQueryTemplate.java
```diff
@@ -0,0 +1,196 @@
...
+	/**
+	 * Writes data to a specified BigQuery table.
+	 *
+	 * @param tableName name of the table to write to
+	 * @param dataFile data file to write to BigQuery
+	 * @param dataFormatOptions the format of the data in the data file
+	 * @return {@link ListenableFuture} containing the BigQuery Job indicating completion of operation
+	 * @throws IOException if the data fails to be loaded
+	 */
+	public ListenableFuture<Job> writeDataToTable(
```
meltsufin: Do we really need this convenience method and the one below?
It doesn't seem to be used yet.
You can also argue that accepting a Spring `Resource` might be useful, or more useful.

spring-cloud-gcp-bigquery/src/main/java/org/springframework/cloud/gcp/bigquery/integration/BigQueryTemplate.java
dzou: Added clarification. User will have to manually specify the schema.

spring-cloud-gcp-bigquery/src/main/java/org/springframework/cloud/gcp/bigquery/integration/BigQueryTemplate.java
dzou: Added clarification. It means "overwritten".

spring-cloud-gcp-bigquery/src/main/java/org/springframework/cloud/gcp/bigquery/integration/outbound/BigQueryFileMessageHandler.java
dzou: Done, removed.

spring-cloud-gcp-bigquery/src/main/java/org/springframework/cloud/gcp/bigquery/integration/BigQueryTemplate.java
dzou: Done. Removed the other helpers. I prefer `InputStream` though over Resource since `InputStream` is still more versatile and user just has to call `resource.getInputStream()` to go back from resource to input stream.

spring-cloud-gcp-bigquery/src/main/java/org/springframework/cloud/gcp/bigquery/core/BigQueryTemplate.java
```diff
@@ -0,0 +1,160 @@
...
+
+		TableDataWriteChannel writer = bigQuery.writer(writeChannelConfiguration);
+
+		try (OutputStream sink = Channels.newOutputStream(writer)) {
+			// Write data from data input file to BigQuery
+			StreamUtils.copy(inputStream, sink);
+		}
+
+		if (writer.getJob() == null) {
+			throw new RuntimeException(
```
meltsufin: I think we need to introduce `BigQueryException`

spring-cloud-gcp-bigquery/src/main/java/org/springframework/cloud/gcp/bigquery/core/BigQueryOperations.java
```diff
@@ -0,0 +1,47 @@
...
+	 *
+	 * @param tableName name of the table to write to
+	 * @param inputStream input stream of the table data to write
+	 * @param dataFormatOptions the format of the data to write
+	 * @return {@link ListenableFuture} containing the BigQuery Job indicating completion of
+	 * operation
+	 * @throws IOException if the data fails to be loaded
+	 */
+	ListenableFuture<Job> writeDataToTable(
+			String tableName, InputStream inputStream, FormatOptions dataFormatOptions) throws IOException;
```
meltsufin: We may need to replace this `IOException` with `BigQueryException` as well.
@artembilan WDYT?

spring-cloud-gcp-bigquery/src/main/java/org/springframework/cloud/gcp/bigquery/core/BigQueryTemplate.java
artembilan: :+1 

Since we talk here about some data interaction, it would be grate to consider to make this exception (or several?) as an extension of the `DataAccessException`.
See some its extensions throughout the Spring modules.

dzou: Done, added exception.

spring-cloud-gcp-bigquery/src/main/java/org/springframework/cloud/gcp/bigquery/core/BigQueryOperations.java
dzou: Done, added exception.

artembilan: Why haven't you replaced this one with the `BigQueryException` ?

spring-cloud-gcp-bigquery/src/main/java/org/springframework/cloud/gcp/bigquery/core/BigQueryTemplate.java
```diff
@@ -0,0 +1,161 @@
...
+
+	/**
+	 * Sets the {@link WriteDisposition} which specifies how data should be inserted into
+	 * BigQuery tables.
+	 * @param writeDisposition whether to append to or truncate (overwrite) data in the
+	 *     BigQuery table. Default is {@code WriteDisposition.WRITE_APPEND} to append data to
+	 *     a table.
+	 */
+	public void setWriteDisposition(WriteDisposition writeDisposition) {
+		this.writeDisposition = writeDisposition;
```
artembilan: `Assert.notNull()`

spring-cloud-gcp-bigquery/src/main/java/org/springframework/cloud/gcp/bigquery/core/BigQueryTemplate.java
```diff
@@ -0,0 +1,161 @@
...
+				.setWriteDisposition(this.writeDisposition)
+				.build();
+
+		TableDataWriteChannel writer = bigQuery.writer(writeChannelConfiguration);
+
+		try (OutputStream sink = Channels.newOutputStream(writer)) {
+			// Write data from data input file to BigQuery
+			StreamUtils.copy(inputStream, sink);
+		} catch (IOException e) {
+			throw new BigQueryException("Failed to write data to BigQuery tables.", e);
```
artembilan: I think we need to mentioned in JavaDocs that this exception is thrown in case of error.
This would increase end-user experience.

spring-cloud-gcp-bigquery/src/test/java/org/springframework/cloud/gcp/bigquery/integration/BigQueryTemplateIntegrationTests.java
```diff
@@ -0,0 +1,105 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.cloud.gcp.bigquery.core.BigQueryTemplate;
+import org.springframework.test.context.junit4.SpringRunner;
+import org.springframework.util.concurrent.ListenableFuture;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.springframework.cloud.gcp.bigquery.integration.BigQueryTestConfiguration.DATASET_NAME;
+
+@RunWith(SpringRunner.class)
```
artembilan: Missed JavaDocs. At least `@author` and `@since`

spring-cloud-gcp-bigquery/src/test/java/org/springframework/cloud/gcp/bigquery/integration/BigQueryTemplateIntegrationTests.java
```diff
@@ -0,0 +1,105 @@
...
+	@Autowired
+	BigQuery bigQuery;
+
+	@Autowired
+	BigQueryTemplate bigQueryTemplate;
+
+	@Before
+	public void setup() {
+		// Clear the previous dataset before beginning the test.
+		this.bigQuery.delete(TableId.of(DATASET_NAME, TABLE_NAME));
```
artembilan: I wonder how this test is going to work when I don't have access to GCP...

And why don't you clean DB after tests?..

spring-cloud-gcp-bigquery/src/test/java/org/springframework/cloud/gcp/bigquery/integration/BigQueryTemplateIntegrationTests.java
```diff
@@ -0,0 +1,105 @@
...
+	}
+
+	@Test
+	public void testLoadFile() throws IOException, ExecutionException, InterruptedException {
+		File file = new File("src/test/resources/data.csv");
```
artembilan: Consider to use a `@Value` for `Resource`.
It is really bad idea to try to get access to files from the tests via absolute or relative path.

spring-cloud-gcp-bigquery/src/main/java/org/springframework/cloud/gcp/bigquery/core/BigQueryTemplate.java
dzou: Done. Added to javadoc of `BigQueryOperations`.

spring-cloud-gcp-bigquery/src/test/java/org/springframework/cloud/gcp/bigquery/integration/BigQueryTemplateIntegrationTests.java
artembilan: OK. Let them be integration tests, but how can you guarantee with their configuration that they are not going to be run for me locally unconditionally?

spring-cloud-gcp-bigquery/src/main/java/org/springframework/cloud/gcp/bigquery/core/BigQueryTemplate.java
```diff
@@ -0,0 +1,180 @@
...
+			// Write data from data input file to BigQuery
+			StreamUtils.copy(inputStream, sink);
+		}
+		catch (IOException e) {
+			throw new BigQueryException("Failed to write data to BigQuery tables.", e);
+		}
+
+		if (writer.getJob() == null) {
+			throw new BigQueryException(
+					"Failed to initialize the BigQuery write job in message handler: " + this);
```
meltsufin: `this`? It will always be the template class.
```suggestion
					"Failed to initialize the BigQuery write job.");
```

dzou: oops, nice catch.

dzou:
Just fixed the error message in bigquery template, ptal :)

PR #1918, opened 2019-09-20
Provide a fake stream function definition to prevent autodiscovery

elefeint:
This prevents `pubSubReactiveScheduler` from being identified as a valid Spring Cloud Stream function provider.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,8 +45,7 @@ before_install:
           -Dgcs-resource-test-bucket=gcp-storage-resource-bucket-sample
           -Dgcs-read-bucket=gcp-storage-bucket-sample-input
           -Dgcs-write-bucket=gcp-storage-bucket-sample-output
-          -Dgcs-local-directory=/tmp/gcp_integration_tests/integration_storage_sample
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -45,7 +45,8 @@ before_install:
           -Dgcs-resource-test-bucket=gcp-storage-resource-bucket-sample
           -Dgcs-read-bucket=gcp-storage-bucket-sample-input
           -Dgcs-write-bucket=gcp-storage-bucket-sample-output
-          -Dgcs-local-directory=/tmp/gcp_integration_tests/integration_storage_sample";
+          -Dgcs-local-directory=/tmp/gcp_integration_tests/integration_storage_sample
+          -Dspring.cloud.stream.function.definition=none";
```
meltsufin: Can we just do it on the sample that is broken rather than for the whole project build on Travis?
Or I'm missing something?

elefeint: Sure; I'll add the property directly to sample.

meltsufin:
@elefeint Can you create a tracking issue on our side as well to revert this change?

PR #1921, opened 2019-09-20
Implement collection name inference for Firestore Entities

### LINKED ISSUE #1787: Firestore @Entitry default collection name should be the class name
Currently, we throw an exception if `collectionName` is not set for Firestore entity. Instead, we should use the class name as a collection name in that case.

dzou:
This makes it so that if the user omits the `collectionName` from the Entity annotation, we will infer the collection name as the lower-cased name of the class.

Fixes #1787.

### DIFF
--- a/spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/mapping/FirestorePersistentEntityImpl.java
+++ b/spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/mapping/FirestorePersistentEntityImpl.java
@@ -21,6 +21,7 @@ import org.springframework.cloud.gcp.data.firestore.FirestoreDataException;
 import org.springframework.core.annotation.AnnotationUtils;
 import org.springframework.data.mapping.model.BasicPersistentEntity;
 import org.springframework.data.util.TypeInformation;
+import org.springframework.util.StringUtils;
 
 /**
  * Metadata class for entities stored in Datastore.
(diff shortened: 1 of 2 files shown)


spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/mapping/FirestorePersistentEntityImpl.java
```diff
@@ -57,4 +55,17 @@ public FirestorePersistentProperty getIdPropertyOrFail() {
 		}
 		return getIdProperty();
 	}
+
+	private static <T> String getEntityCollectionName(TypeInformation<T> typeInformation) {
+		Entity entity = AnnotationUtils.findAnnotation(typeInformation.getType(), Entity.class);
+		String collectionName = (String) AnnotationUtils.getValue(entity, "collectionName");
+
+		if (collectionName == null || collectionName.isEmpty()) {
```
meltsufin: `org.springframework.util.StringUtils.isEmpty(String)` can be used here.

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/mapping/FirestorePersistentEntityImpl.java
```diff
@@ -57,4 +55,17 @@ public FirestorePersistentProperty getIdPropertyOrFail() {
...
+			// Infer the collection name as the lowercase entity name.
+			return StringUtils.uncapitalize(typeInformation.getType().getSimpleName());
```
meltsufin: Whats wrong with `String.toLowerCase()`?

dzou: This is better because if the user has an entity named `EmployeeRecord` it will only lowercase the first letter so it retains the camelcasing `employeeRecord`.

This is also the convention used in Spanner for inferring table name if the user does not explicitly set it for their entity.

meltsufin: Oh!

Issue #1885, opened 2019-09-12
Firestore - add pagination and sort support for PartTree methods


dmitry-s:

dzou:
Is this feature relevant for Reactive?

Apparently the paginating/sorting functionality is something that does not necessarily fit in with reactive patterns: https://stackoverflow.com/questions/46384618/how-apply-pagination-in-reactive-spring-data

PR #1924, opened 2019-09-23
Implement Firestore Pageable LIMIT and OFFSET

dzou:
Implements passing in LIMIT and OFFSET from `Pageable` for Firestore part tree methods.

In progress #1885.

### DIFF
--- a/spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/repository/query/PartTreeFirestoreQuery.java
+++ b/spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/repository/query/PartTreeFirestoreQuery.java
@@ -25,7 +25,6 @@ import java.util.stream.Collectors;
 import com.google.cloud.firestore.PublicClassMapper;
 import com.google.firestore.v1.StructuredQuery;
 import com.google.protobuf.Int32Value;
-import reactor.core.publisher.Flux;
 
 import org.springframework.cloud.gcp.core.util.MapBuilder;
 import org.springframework.cloud.gcp.data.firestore.FirestoreDataException;
--- a/spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/repository/query/PartTreeFirestoreQuery.java
+++ b/spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/repository/query/PartTreeFirestoreQuery.java
@@ -24,6 +24,7 @@ import java.util.stream.Collectors;
 
 import com.google.cloud.firestore.PublicClassMapper;
 import com.google.firestore.v1.StructuredQuery;
+import com.google.protobuf.Int32Value;
 import reactor.core.publisher.Flux;
 
 import org.springframework.cloud.gcp.core.util.MapBuilder;
@@ -32,6 +33,9 @@ import org.springframework.cloud.gcp.data.firestore.FirestoreReactiveOperations;
 import org.springframework.cloud.gcp.data.firestore.mapping.FirestoreMappingContext;
 import org.springframework.cloud.gcp.data.firestore.mapping.FirestorePersistentEntity;
 import org.springframework.cloud.gcp.data.firestore.mapping.FirestorePersistentProperty;
+import org.springframework.data.domain.Pageable;
+import org.springframework.data.repository.query.ParameterAccessor;
+import org.springframework.data.repository.query.ParametersParameterAccessor;
 import org.springframework.data.repository.query.QueryMethod;
 import org.springframework.data.repository.query.RepositoryQuery;
 import org.springframework.data.repository.query.ReturnedType;
@@ -53,13 +57,13 @@ import static org.springframework.data.repository.query.parser.Part.Type.SIMPLE_
 public class PartTreeFirestoreQuery implements RepositoryQuery {
 	private final PartTree tree;
 
-	private final QueryMethod queryMethod;
+	private final FirestoreQueryMethod queryMethod;
 
 	private final FirestoreReactiveOperations reactiveOperations;
 
 	private final FirestorePersistentEntity<?> persistentEntity;
 
-	private  static final Map<Part.Type, StructuredQuery.FieldFilter.Operator> PART_TO_FILTER_OP =
+	private static final Map<Part.Type, StructuredQuery.FieldFilter.Operator> PART_TO_FILTER_OP =
 			new MapBuilder<Part.Type, StructuredQuery.FieldFilter.Operator>()
 					.put(SIMPLE_PROPERTY, StructuredQuery.FieldFilter.Operator.EQUAL)
 					.put(GREATER_THAN_EQUAL, StructuredQuery.FieldFilter.Operator.GREATER_THAN_OR_EQUAL)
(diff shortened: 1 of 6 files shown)


spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/repository/query/PartTreeFirestoreQuery.java
```diff
@@ -86,14 +90,26 @@ public Object execute(Object[] parameters) {
 		}
 
 		StructuredQuery.Builder builder = createBuilderWithFilter(parameters);
-		Flux<?> result = this.reactiveOperations.execute(builder, this.persistentEntity.getType());
 
+		// Handle Pageable parameters.
+		if (!getQueryMethod().getParameters().isEmpty()) {
```
meltsufin: Can you just call `getQueryMethod().isPageQuery()` here?

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/repository/query/PartTreeFirestoreQuery.java
```diff
@@ -86,14 +90,26 @@ public Object execute(Object[] parameters) {
...
+			ParameterAccessor paramAccessor = new ParametersParameterAccessor(getQueryMethod().getParameters(),
+					parameters);
+			Pageable pageable = paramAccessor.getPageable();
+			if (pageable != null && pageable.isPaged()) {
```
meltsufin: What about Slice query support, like we have in Datastore? Does it make sense for Firestore too?

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/repository/query/PartTreeFirestoreQuery.java
dzou: No, the getQueryMethod().isPageQuery() determines if its a page query if the return type is a subclass of Page.

```
  public final boolean isPageQuery() {
    return org.springframework.util.ClassUtils.isAssignable(Page.class, this.unwrappedReturnType);
  }
```

However we don't want to deal with Page or Slice objects because they are not reactive; they store the contents of the query into a list and allow you to manipulate them further.

See Mark's comments here: https://stackoverflow.com/a/46391830/906497

So in this case we are modelling our behavior after Mongo reactive repository's pattern of allowing LIMIT and OFFSET and then transmitting the results in a flux.

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/repository/query/PartTreeFirestoreQuery.java
dzou: See above comments, I think slice/page support may not make sense for a reactive repository.

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/repository/query/PartTreeFirestoreQuery.java
dmitry-s: @dzou could you clarify why sorting is blocking? I was unable to find that in the linked page.

dzou: Yup, sorting is a blocking operation because one must wait for all the elements in the flux to arrive (wait for flux to complete) before performing sort and then returning the result to the user.

Reactive lets you sort the elements of a flux; you can see what I mean - Its return type is `Mono<List<T>>` not `Flux<T>`:
https://projectreactor.io/docs/core/release/api/reactor/core/publisher/Flux.html#collectSortedList--

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/repository/query/PartTreeFirestoreQuery.java
```diff
@@ -86,6 +91,19 @@ public Object execute(Object[] parameters) {
...
+				builder.setOffset((int) Math.min(Integer.MAX_VALUE, pageable.getOffset()));
+				builder.setLimit(Int32Value.newBuilder().setValue(pageable.getPageSize()));
+			}
+		}
+
+		Flux<?> result = this.reactiveOperations.execute(builder, this.persistentEntity.getType());
```
dmitry-s: This variable is never used.

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/repository/query/PartTreeFirestoreQuery.java
```diff
@@ -86,6 +91,19 @@ public Object execute(Object[] parameters) {
...
 		StructuredQuery.Builder builder = createBuilderWithFilter(parameters);
 
+
+		// Handle Pageable parameters.
+		if (!getQueryMethod().getParameters().isEmpty()) {
+			ParameterAccessor paramAccessor = new ParametersParameterAccessor(getQueryMethod().getParameters(),
+					parameters);
+			Pageable pageable = paramAccessor.getPageable();
+			if (pageable != null && pageable.isPaged()) {
+				builder.setOffset((int) Math.min(Integer.MAX_VALUE, pageable.getOffset()));
```
dmitry-s: This is interesting - `setOffset` accepts an `int`, but `setLimit` needs a `Int32Value`? :)

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/repository/query/PartTreeFirestoreQuery.java
dmitry-s: We are not going to sort locally, we would just send sorting options to firestore, where everything is indexed and can be efficiently streamed to us in the requested order.

dzou: Ohh I see, that makes sense.

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/repository/query/PartTreeFirestoreQuery.java
dzou: Nice catch, done.

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/repository/query/PartTreeFirestoreQuery.java
dzou: Lol beats me... Yeah that is what RunQueryRequest accepts :open_mouth: 

See the object definition: https://cloud.google.com/firestore/docs/reference/rpc/google.firestore.v1#structuredquery

dzou:
> Please create an issue to handle sorting. Otherwise looks good.

Done. https://github.com/spring-cloud/spring-cloud-gcp/issues/1940

Issue #1930, opened 2019-09-24
help request: datastore & spring batch: no transaction aspect-managed TransactionStatus in scope spring batch item reader reading datastore


jfinci1:
This is a help request.

Leveraging Datastore with Spring Batch

I've implemented a custom item reader which reads from an API. I intended to weave in data from Datastore during the read step- but as soon as I attempt to query the DatastoreRepository I get an error:

`org.springframework.transaction.NoTransactionException: No transaction aspect-managed TransactionStatus in scope`

Spring batch is separately using a SQL instance backing the batch execution metadata.

Any thoughts? Thank you!

dzou:
Hey @jfinci1, thanks for posting an issue. Do you have any code snippets or examples we could look at to help us understand more? (ideally a reproducible sample!) Our team is not too familiar with Spring batch.

jfinci1:
Sure, I've switched to a tasklet based approach, but here's some code:

the batch configuration, where i've configured a datasource (using cloud SQL) to back Spring Batch execution metadata

```
@Configuration
@EnableBatchProcessing
public class BatchConfiguration {

    //...

  @Bean
  DataSource getDataSource(SecurityService securitySvc) throws IOException {

    Objects.requireNonNull(cloudSqlDatabaseName);
    Objects.requireNonNull(cloudSqlUsername);
    Objects.requireNonNull(cloudSqlPassword);
    Objects.requireNonNull(cloudSqlConnectionName);

    HikariConfig config = new HikariConfig();
      config.setJdbcUrl(String.format("jdbc:mysql://google/%s", cloudSqlDatabaseName));
    config.setUsername(cloudSqlUsername);
    config.setPassword(securitySvc.decryptProperty(cloudSqlPassword));
    config.addDataSourceProperty("socketFactory", "com.google.cloud.sql.mysql.SocketFactory");
    config.addDataSourceProperty("cloudSqlInstance", cloudSqlConnectionName);
    config.addDataSourceProperty("useSSL", "false");

    return new HikariDataSource(config);
  }
}
```

The configuration for the batch job:

```
@Configuration
public class CustomersBatchJobConfiguration {

  public static final String CUSTOMERS_BATCH_JOB = "CUSTOMERS_BATCH_JOB";

  public static final String CUSTOMERS_BATCH_JOB_LOAD_FROM_ERP_STEP
      = "CUSTOMERS_BATCH_JOB_LOAD_FROM_ERP_STEP";

[...]

PR #1936, opened 2019-09-25
Fix Datastore transaction management

meltsufin:
Fixes: #1930.

### DIFF
--- a/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTransactionManager.java
+++ b/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTransactionManager.java
@@ -16,24 +16,16 @@
 
 package org.springframework.cloud.gcp.data.datastore.core;
 
-import java.util.Iterator;
-import java.util.List;
 import java.util.function.Supplier;
 
 import com.google.cloud.datastore.Datastore;
 import com.google.cloud.datastore.DatastoreException;
-import com.google.cloud.datastore.Entity;
-import com.google.cloud.datastore.FullEntity;
-import com.google.cloud.datastore.Key;
-import com.google.cloud.datastore.Query;
-import com.google.cloud.datastore.QueryResults;
 import com.google.cloud.datastore.Transaction;
-import com.google.protobuf.ByteString;
+import com.google.datastore.v1.TransactionOptions;
 
 import org.springframework.transaction.TransactionDefinition;
 import org.springframework.transaction.TransactionException;
 import org.springframework.transaction.TransactionSystemException;
-import org.springframework.transaction.interceptor.TransactionAspectSupport;
 import org.springframework.transaction.support.AbstractPlatformTransactionManager;
 import org.springframework.transaction.support.DefaultTransactionStatus;
 import org.springframework.transaction.support.TransactionSynchronizationManager;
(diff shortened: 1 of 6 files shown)


spring-cloud-gcp-data-datastore/src/test/java/org/springframework/cloud/gcp/data/datastore/it/DatastoreIntegrationTests.java
```diff
@@ -358,7 +364,7 @@ public void testSaveAndDeleteRepository() throws InterruptedException {
 		assertThat(
 				this.testEntityRepository.removeByColor("red").stream()
 						.map(TestEntity::getId).collect(Collectors.toList()))
-								.containsExactlyInAnyOrder(1L, 3L, 4L);
+				.containsExactlyInAnyOrder(1L, 3L, 4L);
```
dmitry-s: indentation?

spring-cloud-gcp-data-datastore/src/test/java/org/springframework/cloud/gcp/data/datastore/it/DatastoreIntegrationTests.java
```diff
@@ -388,7 +394,7 @@ public void testSaveAndDeleteRepository() throws InterruptedException {
 		assertThat(
 				this.testEntityRepository.findTop3BySizeAndColor(1, "red").stream()
 						.map(TestEntity::getId).collect(Collectors.toList()))
-								.containsExactlyInAnyOrder(1L, 3L, 4L);
+				.containsExactlyInAnyOrder(1L, 3L, 4L);
```
dmitry-s: same here

meltsufin:
@mp911de Can you please take a look at this PR? This is a follow up to #1930. Thanks!

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTransactionManager.java
```diff
@@ -46,26 +38,23 @@
...
-		return TransactionSynchronizationManager.isActualTransactionActive()
-				? (Tx) ((DefaultTransactionStatus) TransactionAspectSupport
-						.currentTransactionStatus()).getTransaction()
-				: null;
-	}
-
 	@Override
 	protected Object doGetTransaction() throws TransactionException {
-		Tx tx = getCurrentTX();
+		Tx tx = (Tx) TransactionSynchronizationManager.getResource(datastore.get());
```
mp911de: Typically, `getResource` accepts the resource factory instead of the resource itself. Let me explain this by using JDBC terminology: You'd call `getResource(…)` with a `DataSource` and not a `Connection`.

meltsufin: In our case, I think `Datastore` is the factory. It might be confusing because we use `Supplier<Datastore>`, but that's just to enable connections to other projects, think different Datastore instances. Most of the time that supplier will always return the same instance. Notice how we call `datastore.newTransaction()`, which essentially creates a new session. Does it make sense?

mp911de: Yeah, the `Supplier<…>` caused some confusion. I'd still recommend adding eventually a dedicated interface (e.g. `DatastoreFactory`, `DatastoreSupplier`) instead of using a raw `Supplier`.

meltsufin: Fair point. How about the rest of the `TransactionSynchronizationManager` usage. Does it look good to you?

mp911de: Yes, the rest looks good to me.

spring-cloud-gcp-data-datastore/src/test/java/org/springframework/cloud/gcp/data/datastore/it/DatastoreIntegrationTests.java
```diff
@@ -112,6 +115,12 @@
 		this.allTestEntities = Arrays.asList(this.testEntityA, this.testEntityB, this.testEntityC, this.testEntityD);
 	}
 
+	/**
+	 * Used to check exception types and messages.
+	 */
+	@Rule
+	public ExpectedException expectedException = ExpectedException.none();
```
dmitry-s: maybe use assertThatExceptionOfType instead of the rule since we use assertj?

PR #1951, opened 2019-10-01
Add error propagation for pubsub stream binder

### LINKED ISSUE #1938: Error handlers in Spring Cloud GCP Stream Binder for Google PubSub are not invoked
Hi,

I've created a small app based on [the Spring Cloud GCP Stream Binder for Pub/Sub code sample](https://github.com/spring-cloud/spring-cloud-gcp/tree/v1.1.3.RELEASE/spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-binder-sample) to play with error handling. I followed [the Spring Cloud Stream application error handling docs](https://cloud.spring.io/spring-cloud-static/spring-cloud-stream/2.1.4.RELEASE/single/spring-cloud-stream.html#spring-cloud-stream-overview-error-handling):

- `pom.xml`:

```xml
<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
    <modelVersion>4.0.0</modelVersion>
    <parent>
        <artifactId>spring-boot-starter-parent</artifactId>
        <version>2.1.8.RELEASE</version>
        <relativePath/> <!-- lookup parent from repository -->
    </parent>
    <groupId>com.example</groupId>
    <artifactId>demo</artifactId>
    <version>0.0.1-SNAPSHOT</version>
    <name>demo</name>

    <properties>
        <java.version>11</java.version>
        <spring-cloud.version>Greenwich.SR3</spring-cloud.version>
    </properties>

    <dependencies>
        <dependency>
            <artifactId>spring-cloud-gcp-starter-pubsub</artifactId>
        </dependency>
        <dependency>
            <artifactId>spring-cloud-gcp-pubsub-stream-binder</artifactId>
        </dependency>
        <dependency>
            <artifactId>spring-boot-starter-web</artifactId>
        </dependency>
    </dependencies>

    <dependencyManagement>
        <dependencies>
            <dependency>
                <artifactId>spring-cloud-dependencies</artifactId>
                <version>${spring-cloud.version}</version>
                <type>pom</type>
                <scope>import</scope>
            </dependency>
        </dependencies>
    </dependencyManagement>

    <build>
        <plugins>
            <plugin>
                <artifactId>spring-boot-maven-plugin</artifactId>
            </plugin>
        </plugins>
    </build>

</project>
```

- `application.properties`:

```
spring.cloud.stream.bindings.input.destination=myFooDestination
spring.cloud.stream.bindings.input.group=myGroup

spring.cloud.stream.bindings.output.destination=myFooDestination
```

- first approach:

```java
@EnableBinding(Sink.class)
public class SinkExample {

	private static final Log LOGGER = LogFactory.getLog(SinkExample.class);

	@StreamListener(Sink.INPUT)
	public void handleMessage(UserMessage userMessage) {
		LOGGER.info("New message received from " + userMessage.getUsername() + ": " + userMessage.getBody() + " at " + userMessage.getCreatedAt());
		LOGGER.info("*** Throwing exception now! ***");
		throw new RuntimeException("BOOM!");
	}

	@ServiceActivator(inputChannel = Processor.INPUT + ".myGroup.errors")
	public void error(Message<?> message) {
		LOGGER.error("Handling ERROR: " + message);
	}
}
```

- second approach:

```java
@EnableBinding(Sink.class)
public class SinkExample {

	private static final Log LOGGER = LogFactory.getLog(SinkExample.class);

	@StreamListener(Sink.INPUT)
	public void handleMessage(UserMessage userMessage) {
		LOGGER.info("New message received from " + userMessage.getUsername() + ": " + userMessage.getBody() + " at " + userMessage.getCreatedAt());
		LOGGER.info("*** Throwing exception now! ***");
		throw new RuntimeException("BOOM!");
	}

	@ServiceActivator(inputChannel = "myFooDestination.myGroup.errors")
	public void error(Message<?> message) {
		LOGGER.error("Handling ERROR: " + message);
	}
}
```

- third approach:

```java
@EnableBinding(Sink.class)
public class SinkExample {

	private static final Log LOGGER = LogFactory.getLog(SinkExample.class);

	@StreamListener(Sink.INPUT)
	public void handleMessage(UserMessage userMessage) {
		LOGGER.info("New message received from " + userMessage.getUsername() + ": " + userMessage.getBody() + " at " + userMessage.getCreatedAt());
		LOGGER.info("*** Throwing exception now! ***");
		throw new RuntimeException("BOOM!");
	}

	@StreamListener("errorChannel")
	public void error(Message<?> message) {
		LOGGER.error("Handling ERROR: " + message);
	}
}
```

Error handlers in each approach do not work - **methods are not invoked**. What I am missing here? I'd really appreciate some help.

[console_logs.txt](https://github.com/spring-cloud/spring-cloud-gcp/files/3662271/console_logs.txt)

dzou:
Adds error propagation for the pub/sub stream binder.

Note that with error handling, I modified the ack'ing logic of the `PubSubInboundChannelAdapter` to ack messages and forward them to an error channel if an errorChannel is initialized for the adapter. Spring typically initializes an error channel for the adapter if it is created through Spring autoconfiguration.

Details:

If AckMode.AUTO is set:

If a RuntimeException is thrown while the message is processed, then the message is handled in the following way:

* If an error channel is configured for the adapter (pubSubInboundChannelAdapater.getErrorChannel() != null) then the message will be acked and forwarded with the error to the error channel.
* If no error channel is configured for the adapter, then the message is n’acked.

If Automatic acking OK (AckMode.AUTO_ACK) is set:

* No change to existing behavior.

If Manually acking (AckMode.MANUAL) is set:

* No change to existing behavior.

Fixes #1938.

### DIFF
--- a/docs/src/main/asciidoc/spring-integration-pubsub.adoc
+++ b/docs/src/main/asciidoc/spring-integration-pubsub.adoc
@@ -76,7 +76,11 @@ The `PubSubInboundChannelAdapter` supports three acknowledgement modes, with `Ac
 Automatic acking (`AckMode.AUTO`)
 
 A message is acked with GCP Pub/Sub if the adapter sent it to the channel and no exceptions were thrown.
-If a `RuntimeException` is thrown while the message is processed, then the message is nacked.
+
+If a `RuntimeException` is thrown while the message is processed, then the message is handled in the following way:
+
+* If an error channel is configured for the adapter (`pubSubInboundChannelAdapater.getErrorChannel() != null`) then the message will be acked and forwarded with the error to the error channel.
+* If no error channel is configured for the adapter, then the message is n'acked.
 
 Automatic acking OK (`AckMode.AUTO_ACK`)
 
--- a/docs/src/main/asciidoc/spring-integration-pubsub.adoc
+++ b/docs/src/main/asciidoc/spring-integration-pubsub.adoc
@@ -80,7 +80,7 @@ A message is acked with GCP Pub/Sub if the adapter sent it to the channel and no
 If a `RuntimeException` is thrown while the message is processed, then the message is handled in the following way:
 
 * If an error channel is configured for the adapter (`pubSubInboundChannelAdapater.getErrorChannel() != null`) then the message will be acked and forwarded with the error to the error channel.
-* If no error channel is configured for the adapter, then the message is n'acked.
+* If no error channel is configured for the adapter, then the message is nacked.
 
 Automatic acking OK (`AckMode.AUTO_ACK`)
 
--- a/spring-cloud-gcp-pubsub-stream-binder/src/main/java/org/springframework/cloud/gcp/stream/binder/pubsub/PubSubMessageChannelBinder.java
+++ b/spring-cloud-gcp-pubsub-stream-binder/src/main/java/org/springframework/cloud/gcp/stream/binder/pubsub/PubSubMessageChannelBinder.java
@@ -80,7 +80,13 @@ public class PubSubMessageChannelBinder
 	protected MessageProducer createConsumerEndpoint(ConsumerDestination destination, String group,
 			ExtendedConsumerProperties<PubSubConsumerProperties> properties) {
 
-		return new PubSubInboundChannelAdapter(this.pubSubTemplate, destination.getName());
+		PubSubInboundChannelAdapter adapter = new PubSubInboundChannelAdapter(this.pubSubTemplate,
+				destination.getName());
+
+		ErrorInfrastructure errorInfrastructure = registerErrorInfrastructure(destination, group, properties);
+		adapter.setErrorChannel(errorInfrastructure.getErrorChannel());
+
+		return adapter;
 	}
 
 	@Override
--- a/spring-cloud-gcp-pubsub-stream-binder/src/main/java/org/springframework/cloud/gcp/stream/binder/pubsub/PubSubMessageChannelBinder.java
+++ b/spring-cloud-gcp-pubsub-stream-binder/src/main/java/org/springframework/cloud/gcp/stream/binder/pubsub/PubSubMessageChannelBinder.java
@@ -90,6 +90,12 @@ public class PubSubMessageChannelBinder
 	}
 
 	@Override
+	protected String errorsBaseName(
+			ConsumerDestination destination, String group, ExtendedConsumerProperties<PubSubConsumerProperties> properties) {
+		return destination.getName() + ".errors";
+	}
+
+	@Override
 	public PubSubConsumerProperties getExtendedConsumerProperties(String channelName) {
 		return this.pubSubExtendedBindingProperties.getExtendedConsumerProperties(channelName);
 	}
--- a/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapter.java
+++ b/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapter.java
@@ -145,6 +145,8 @@ public class PubSubInboundChannelAdapter extends MessageProducerSupport {
 			});
 		}
 
+		boolean messageNacked = false;
+
 		try {
 			sendMessage(getMessageBuilderFactory()
 					.withPayload(message.getPayload())
@@ -153,20 +155,17 @@ public class PubSubInboundChannelAdapter extends MessageProducerSupport {
 		}
 		catch (RuntimeException re) {
 			if (this.ackMode == AckMode.AUTO) {
-				if (getErrorChannel() != null) {
-					// If the error channel is present, acknowledge the message.
-					// It gets forwarded to the error channel once PubSubException is thrown.
-					message.ack();
-				}
-				else {
-					message.nack();
-				}
+				message.nack();
+				messageNacked = true;
 			}
 			throw new PubSubException("Sending Spring message failed.", re);
 		}
-
-		if ((this.ackMode == AckMode.AUTO) || (this.ackMode == AckMode.AUTO_ACK)) {
-			message.ack();
+		finally {
+			if (this.ackMode == AckMode.AUTO || this.ackMode == AckMode.AUTO_ACK) {
+				if (!messageNacked) {
+					message.ack();
+				}
+			}
 		}
 	}
 
--- a/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapter.java
+++ b/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapter.java
@@ -153,7 +153,14 @@ public class PubSubInboundChannelAdapter extends MessageProducerSupport {
 		}
 		catch (RuntimeException re) {
 			if (this.ackMode == AckMode.AUTO) {
-				message.nack();
+				if (getErrorChannel() != null) {
+					// If the error channel is present, acknowledge the message.
+					// It gets forwarded to the error channel once PubSubException is thrown.
+					message.ack();
+				}
+				else {
+					message.nack();
+				}
 			}
 			throw new PubSubException("Sending Spring message failed.", re);
 		}
--- a/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapter.java
+++ b/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapter.java
@@ -161,10 +161,8 @@ public class PubSubInboundChannelAdapter extends MessageProducerSupport {
 			throw new PubSubException("Sending Spring message failed.", re);
 		}
 		finally {
-			if (this.ackMode == AckMode.AUTO || this.ackMode == AckMode.AUTO_ACK) {
-				if (!messageNacked) {
-					message.ack();
-				}
+			if (!messageNacked && (this.ackMode == AckMode.AUTO || this.ackMode == AckMode.AUTO_ACK)) {
+				message.ack();
 			}
 		}
 	}
--- a/spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-binder-sample/src/main/java/com/example/SinkExample.java
+++ b/spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-binder-sample/src/main/java/com/example/SinkExample.java
@@ -48,7 +48,7 @@ public class SinkExample {
 
 	// Note that the error inputChannel is formatted as [Pub/Sub subscription name with group].[group name].errors
 	// If you change the topic name in application.properties, you also have to change the inputChannel below.
-	@ServiceActivator(inputChannel = "my-topic.my-group.my-group.errors")
+	@ServiceActivator(inputChannel = "my-topic.my-group.errors")
 	public void error(Message<MessagingException> message) {
 		LOGGER.error("The message that was sent is now processed by the error handler.");
 		LOGGER.error("Failed message: " + message.getPayload().getFailedMessage());
--- a/spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-binder-sample/src/main/java/com/example/SinkExample.java
+++ b/spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-binder-sample/src/main/java/com/example/SinkExample.java
@@ -22,6 +22,8 @@ import org.apache.commons.logging.LogFactory;
 import org.springframework.cloud.stream.annotation.EnableBinding;
 import org.springframework.cloud.stream.annotation.StreamListener;
 import org.springframework.cloud.stream.messaging.Sink;
+import org.springframework.messaging.Message;
+import org.springframework.messaging.MessagingException;
 
 /**
  * Example of a sink for the sample app.
@@ -35,7 +37,17 @@ public class SinkExample {
 
 	@StreamListener(Sink.INPUT)
 	public void handleMessage(UserMessage userMessage) {
+		if (userMessage.isThrowError()) {
+			throw new RuntimeException("An error was triggered in the message handler!");
+		}
+
 		LOGGER.info("New message received from " + userMessage.getUsername() + ": " + userMessage.getBody() +
 				" at " + userMessage.getCreatedAt());
+	}
+
+	@StreamListener("errorChannel")
+	public void handleErrorMessage(Message<MessagingException> message) {
+		LOGGER.error("The message that was sent is now processed by the error handler.");
+		LOGGER.error("Failed message: " + message.getPayload().getFailedMessage());
 	}
 }
--- a/spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-binder-sample/src/main/java/com/example/SinkExample.java
+++ b/spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-binder-sample/src/main/java/com/example/SinkExample.java
@@ -21,7 +21,9 @@ import org.apache.commons.logging.LogFactory;
 
 import org.springframework.cloud.stream.annotation.EnableBinding;
 import org.springframework.cloud.stream.annotation.StreamListener;
+import org.springframework.cloud.stream.messaging.Processor;
 import org.springframework.cloud.stream.messaging.Sink;
+import org.springframework.integration.annotation.ServiceActivator;
 import org.springframework.messaging.Message;
 import org.springframework.messaging.MessagingException;
 
@@ -45,8 +47,10 @@ public class SinkExample {
 				" at " + userMessage.getCreatedAt());
 	}
 
-	@StreamListener("errorChannel")
-	public void handleErrorMessage(Message<MessagingException> message) {
+	// Note that the error inputChannel is formatted as [Pub/Sub subscription name with group].[group name].errors
+	// If you change the topic name in application.properties, you also have to change the inputChannel below.
+	@ServiceActivator(inputChannel = "my-topic.my-group.my-group.errors")
+	public void error(Message<MessagingException> message) {
 		LOGGER.error("The message that was sent is now processed by the error handler.");
 		LOGGER.error("Failed message: " + message.getPayload().getFailedMessage());
 	}
--- a/spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-binder-sample/src/main/java/com/example/SinkExample.java
+++ b/spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-binder-sample/src/main/java/com/example/SinkExample.java
@@ -21,7 +21,6 @@ import org.apache.commons.logging.LogFactory;
 
 import org.springframework.cloud.stream.annotation.EnableBinding;
 import org.springframework.cloud.stream.annotation.StreamListener;
-import org.springframework.cloud.stream.messaging.Processor;
 import org.springframework.cloud.stream.messaging.Sink;
 import org.springframework.integration.annotation.ServiceActivator;
 import org.springframework.messaging.Message;
--- a/spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-binder-sample/src/main/resources/static/index.html
+++ b/spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-binder-sample/src/main/resources/static/index.html
@@ -12,7 +12,7 @@
 
         Throw Error? (triggers the error handler)
         <select name="throwError">
-            <option value="false">false</option>
+            <option value="false" selected>false</option>
             <option value="true">true</option>
         </select>
         <br/>
--- a/spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-binder-sample/src/main/resources/static/index.html
+++ b/spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-binder-sample/src/main/resources/static/index.html
@@ -9,6 +9,14 @@
     <form action="/newMessage" method="post">
         Message body: <input type="text" name="messageBody" /><br/>
         Username: <input type="text" name="username" /><br/>
+
+        Throw Error? (triggers the error handler)
+        <select name="throwError">
+            <option value="false">false</option>
+            <option value="true">true</option>
+        </select>
+        <br/>
+
         <input type="submit" value="Post it!"/>
     </form>
 </body>
--- a/spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-binder-sample/README.adoc
+++ b/spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-binder-sample/README.adoc
@@ -1,8 +1,12 @@
 = Spring Cloud GCP Stream Binder for Pub/Sub Code Sample
 
 This code sample demonstrates how to use the Spring Cloud Stream binder for Google Cloud Pub/Sub.
+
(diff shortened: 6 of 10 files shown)


spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-binder-sample/src/main/resources/static/index.html
```diff
@@ -9,6 +9,14 @@
     <form action="/newMessage" method="post">
         Message body: <input type="text" name="messageBody" /><br/>
         Username: <input type="text" name="username" /><br/>
+
+        Throw Error? (triggers the error handler)
+        <select name="throwError">
+            <option value="false">false</option>
```
elefeint: Consider adding a "selected" attribute to the false option, so the user does not have to click in the most common scenario.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapter.java
```diff
@@ -153,7 +153,14 @@ public void nack() {
 		}
 		catch (RuntimeException re) {
 			if (this.ackMode == AckMode.AUTO) {
-				message.nack();
+				if (getErrorChannel() != null) {
```
elefeint: Unit testing this logic would be great, but can be done in a follow up, so we can have the error handling in (today's?) release.

docs/src/main/asciidoc/spring-integration-pubsub.adoc
```diff
@@ -76,7 +76,11 @@ The `PubSubInboundChannelAdapter` supports three acknowledgement modes, with `Ac
 Automatic acking (`AckMode.AUTO`)
 
 A message is acked with GCP Pub/Sub if the adapter sent it to the channel and no exceptions were thrown.
-If a `RuntimeException` is thrown while the message is processed, then the message is nacked.
+
+If a `RuntimeException` is thrown while the message is processed, then the message is handled in the following way:
+
+* If an error channel is configured for the adapter (`pubSubInboundChannelAdapater.getErrorChannel() != null`) then the message will be acked and forwarded with the error to the error channel.
+* If no error channel is configured for the adapter, then the message is n'acked.
```
meltsufin:
```suggestion
* If no error channel is configured for the adapter, then the message is nacked.
```

spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-binder-sample/src/main/java/com/example/SinkExample.java
```diff
@@ -35,7 +37,17 @@
...
+		if (userMessage.isThrowError()) {
+			throw new RuntimeException("An error was triggered in the message handler!");
+		}
+
 		LOGGER.info("New message received from " + userMessage.getUsername() + ": " + userMessage.getBody() +
 				" at " + userMessage.getCreatedAt());
 	}
+
+	@StreamListener("errorChannel")
+	public void handleErrorMessage(Message<MessagingException> message) {
```
meltsufin: How does this `errorChannel` get hooked up into the binder? By convention of the name?
We might need to document it.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapter.java
```diff
@@ -153,7 +153,14 @@ public void nack() {
...
+					// If the error channel is present, acknowledge the message.
+					// It gets forwarded to the error channel once PubSubException is thrown.
+					message.ack();
+				}
+				else {
+					message.nack();
+				}
 			}
 			throw new PubSubException("Sending Spring message failed.", re);
```
artembilan: Not related, but I would change this anyway:

`"Message producing for "+message+"has failed in: " + this`

Also pay attention how `MessageProducerSupport.sendMessage()` works: if there is a success error sending, there is no an exception re-throwing. So, this your `catch` logic is not going to work, when we have an `errorChannel`.

I would say that we are good with the `	if ((this.ackMode == AckMode.AUTO) || (this.ackMode == AckMode.AUTO_ACK)) {` in the end of this method and the message is going to be acked in normal case and when we have an `errorChannel`. Exactly what we would like to achieve. You just revert all your changes in this method and we are good to go ! 😄

spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-binder-sample/src/main/java/com/example/SinkExample.java
artembilan: You don't. This one comes from Spring Integration: https://docs.spring.io/spring-integration/docs/current/reference/html/#channel-special-channels
And it is indeed is wired in Spring Cloud Stream for error handling.

I'm fully don't see reason to document anything in this project. Unless you don't have links to SCSt docs.

I somehow assume when you try to use Pub/Sub Binder, you are already familiar with SCSt core...

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapter.java
dzou: Hey Artem, for some reason this catch-block is still actually triggered by the runtime exception.

I see what you are saying with `MessageProducerSupport`, however when I run the code in the debugger, an exception is actually thrown when the message template attempts to send the error message to the channel - [this line](https://github.com/spring-projects/spring-integration/blob/master/spring-integration-core/src/main/java/org/springframework/integration/endpoint/MessageProducerSupport.java#L217).

Any ideas why you think that may be the case? It appears that in that call, the message handler (which threw the exception) somehow gets triggered again. Do we need to implement `ErrorMessageStrategy` or some additional class?

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapter.java
dzou: Sounds good, will implement shortly after clarification with Artem on whether we can actually revert this or not.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapter.java
artembilan: So, that is a different story and you have an error during sending error message.
It would be great to see what is an exception there and what are subscribers to this error channel.

On other other hand we really can't guarantee that there is no exceptions during error message sending. 
So, you should consider to move ack logic into `finally` block with some state to determine that it has not been nacked in the `catch` block.

dzou: Hmm, in that case maybe the error handling setup our stream binder code sample is wrong? I wonder if there is a way to make it so we do not throw an error when sending to `errorChannel`.

The problem is it looks like the error handling invokes the message handler which threw the error initially *again*.

These are the [subscriber config](https://github.com/spring-cloud/spring-cloud-gcp/blob/953f697a9804847d5fce75f8f170200b78e33a97/spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-binder-sample/src/main/java/com/example/SinkExample.java) for the sample that I using to test error handling.

And the stack trace below - the first two logged lines by `com.example.SinkExample` is logged by my error handler.

Then the line logged by `o.s.integration.handler.LoggingHandler` is the error that I get in the `catch` block. This is what threw the error during error message sending, causing me to enter the catch block. It looks like the error that get thrown during error handling is the same as the initial error?

stack trace:

```
2019-10-02 16:48:43.514 ERROR 18967 --- [sub-subscriber1] com.example.SinkExample                  : The message that was sent is now processed by the error handler.
2019-10-02 16:48:43.516 ERROR 18967 --- [sub-subscriber1] com.example.SinkExample                  : Failed message: GenericMessage [payload=byte[97], headers={id=45f2dc80-6b45-955b-4806-17725f4a9701, contentType=application/json, gcp_pubsub_original_message=PushedAcknowledgeablePubsubMessage{projectId='cowzow', subscriptionName='anonymous.input-topic.29f225e0-67a5-48ff-827f-8363e524fb58', message=data: "{\"body\":\"dfasdf\",\"username\":\"medeadlift\",\"createdAt\":\"2019-10-02T16:47:27.047\",\"throwError\":true}"
attributes {
  key: "contentType"
  value: "application/json"
}
message_id: "437695676167497"
publish_time {
  seconds: 1570049249
  nanos: 519000000
}
}, timestamp=1570049321197}]

2019-10-02 16:48:43.517 ERROR 18967 --- [sub-subscriber1] o.s.integration.handler.LoggingHandler   : org.springframework.messaging.MessagingException: Exception thrown while invoking SinkExample#handleMessage[1 args]; nested exception is java.lang.RuntimeException: An error was triggered in the message handler!, failedMessage=GenericMessage [payload=byte[97], headers={id=45f2dc80-6b45-955b-4806-17725f4a9701, contentType=application/json, gcp_pubsub_original_message=PushedAcknowledgeablePubsubMessage{projectId='cowzow', subscriptionName='anonymous.input-topic.29f225e0-67a5-48ff-827f-8363e524fb58', message=data: "{\"body\":\"dfasdf\",\"username\":\"medeadlift\",\"createdAt\":\"2019-10-02T16:47:27.047\",\"throwError\":true}"
attributes {
  key: "contentType"
  value: "application/json"
}
message_id: "437695676167497"
publish_time {
  seconds: 1570049249
  nanos: 519000000
}
}, timestamp=1570049321197}]
	at org.springframework.cloud.stream.binding.StreamListenerMessageHandler.handleRequestMessage(StreamListenerMessageHandler.java:64)
	at org.springframework.integration.handler.AbstractReplyProducingMessageHandler.handleMessageInternal(AbstractReplyProducingMessageHandler.java:127)
	at org.springframework.integration.handler.AbstractMessageHandler.handleMessage(AbstractMessageHandler.java:177)
	at org.springframework.integration.dispatcher.AbstractDispatcher.tryOptimizedDispatch(AbstractDispatcher.java:115)
	at org.springframework.integration.dispatcher.UnicastingDispatcher.doDispatch(UnicastingDispatcher.java:133)
	at org.springframework.integration.dispatcher.UnicastingDispatcher.dispatch(UnicastingDispatcher.java:106)
	at org.springframework.integration.channel.AbstractSubscribableChannel.doSend(AbstractSubscribableChannel.java:73)
	at org.springframework.integration.channel.AbstractMessageChannel.send(AbstractMessageChannel.java:453)
	at org.springframework.integration.channel.AbstractMessageChannel.send(AbstractMessageChannel.java:403)
	at org.springframework.messaging.core.GenericMessagingTemplate.doSend(GenericMessagingTemplate.java:187)
	at org.springframework.messaging.core.GenericMessagingTemplate.doSend(GenericMessagingTemplate.java:166)
	at org.springframework.messaging.core.GenericMessagingTemplate.doSend(GenericMessagingTemplate.java:47)
	at org.springframework.messaging.core.AbstractMessageSendingTemplate.send(AbstractMessageSendingTemplate.java:109)
	at org.springframework.integration.endpoint.MessageProducerSupport.sendMessage(MessageProducerSupport.java:198)
	at org.springframework.cloud.gcp.pubsub.integration.inbound.PubSubInboundChannelAdapter.consumeMessage(PubSubInboundChannelAdapter.java:151)
	at org.springframework.cloud.gcp.pubsub.core.subscriber.PubSubSubscriberTemplate.lambda$subscribeAndConvert$1(PubSubSubscriberTemplate.java:154)
	at com.google.cloud.pubsub.v1.MessageDispatcher$4.run(MessageDispatcher.java:379)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$201(ScheduledThreadPoolExecutor.java:180)
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:295)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
Caused by: java.lang.RuntimeException: An error was triggered in the message handler!
	at com.example.SinkExample.handleMessage(SinkExample.java:41)
	at org.springframework.messaging.handler.invocation.InvocableHandlerMethod.doInvoke(InvocableHandlerMethod.java:171)
	at org.springframework.messaging.handler.invocation.InvocableHandlerMethod.invoke(InvocableHandlerMethod.java:120)
	at org.springframework.cloud.stream.binding.StreamListenerMessageHandler.handleRequestMessage(StreamListenerMessageHandler.java:55)
	... 23 more
```

artembilan: i don't see an *exception* in the log you show. That is just an `ERROR` logging message.
It fully doesn't show us that your `RuntimeException` is re-thrown after sending into the `errorChannel`.

Please, check an exception in debug mode when it is thrown during sending into the `errorChannel`.
I mean place a break point in the `MessageProducerSupport.sendErrorMessageIfNecessary()` and look for an exception which is re-thrown from the `this.messagingTemplate.send(channel, buildErrorMessage(message, exception));`

artembilan: It might be the fact that you need to add a subscriber to the *destination specific error channel*, not a global one with that `errorChannel` bean name: https://cloud.spring.io/spring-cloud-static/spring-cloud-stream/2.2.1.RELEASE/spring-cloud-stream.html#spring-cloud-stream-overview-error-handling. 

I guess if you don't have one, there is something like *re-throwing subscriber* for the global `errorChannel`

dzou:
PTAL - I ultimately went with Artem's suggestion on the error handling with moving the ACK to a finally-block in the case where an error is thrown when the error is passed to a message channel.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapter.java
```diff
@@ -154,12 +156,16 @@ public void nack() {
 		catch (RuntimeException re) {
 			if (this.ackMode == AckMode.AUTO) {
 				message.nack();
+				messageNacked = true;
```
meltsufin: So, we would only get in here if the sending to the error channel fails? In that case we want to **nack**? Wouldn't that cause repeated re-delivery to no end?

artembilan: I think so. If you have a constant re-delivery, then it is indeed something is wrong in logic of your application.
Fix it - and you are good with acking on error!

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapter.java
```diff
@@ -154,12 +156,16 @@ public void nack() {
...
 			}
 			throw new PubSubException("Sending Spring message failed.", re);
 		}
-
-		if ((this.ackMode == AckMode.AUTO) || (this.ackMode == AckMode.AUTO_ACK)) {
-			message.ack();
+		finally {
+			if (this.ackMode == AckMode.AUTO || this.ackMode == AckMode.AUTO_ACK) {
+				if (!messageNacked) {
```
artembilan: You can combine this `if` with the first one

spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-binder-sample/src/main/java/com/example/SinkExample.java
```diff
@@ -35,7 +37,17 @@
...
 	public void handleMessage(UserMessage userMessage) {
+		if (userMessage.isThrowError()) {
+			throw new RuntimeException("An error was triggered in the message handler!");
+		}
+
 		LOGGER.info("New message received from " + userMessage.getUsername() + ": " + userMessage.getBody() +
 				" at " + userMessage.getCreatedAt());
 	}
+
+	@StreamListener("errorChannel")
```
artembilan: I think its time to demonstrate a *destination-specific error handling* instead.
See that SCSt doc I've mentioned before: https://cloud.spring.io/spring-cloud-static/spring-cloud-stream/2.2.1.RELEASE/spring-cloud-stream.html#spring-cloud-stream-overview-error-handling

The sample is:
```
@ServiceActivator(inputChannel = Processor.INPUT + ".myGroup.errors") //channel name 'input.myGroup.errors'
```

dzou: Hey Artem, so to get this to work I ultimately had to do something like this:

```
@ServiceActivator(inputChannel = "inputTopic.myGroup.myGroup.errors")
```

But is there a way to not hardcode `inputTopic` and `myGroup`? We were hoping to let users configure the topic and group name through `application.properties`; however it appears I can only put constant values in the `ServiceActivator` for `inputChannel`.

artembilan: Not sure about `myGroup.myGroup`, but yes: this is only the way now. An error channel is named really after a destination mapping (see that `Processor.INPUT` above) an a group assigned into this binding target.

spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-binder-sample/src/main/java/com/example/SinkExample.java
```diff
@@ -45,8 +47,10 @@ public void handleMessage(UserMessage userMessage) {
 				" at " + userMessage.getCreatedAt());
 	}
 
-	@StreamListener("errorChannel")
-	public void handleErrorMessage(Message<MessagingException> message) {
+	// Note that the error inputChannel is formatted as [Pub/Sub subscription name with group].[group name].errors
+	// If you change the topic name in application.properties, you also have to change the inputChannel below.
+	@ServiceActivator(inputChannel = "my-topic.my-group.my-group.errors")
```
artembilan: Why `my-group` is twice here ?

dzou: Looking into it deeper, in our code when we see the user configure a group with a pub/sub topic, we will name the destination `topic_name + '.' + group_name`, so that is why it gets duplicated. [Code here.](https://github.com/spring-cloud/spring-cloud-gcp/blob/72d746486fedd19b16ead3cdbbcfb48c3e90c3a9/spring-cloud-gcp-pubsub-stream-binder/src/main/java/org/springframework/cloud/gcp/stream/binder/pubsub/provisioning/PubSubChannelProvisioner.java#L78)

So then when the error channel is named `destination` + `group` + `error` = `topic.group.group.error`.

But there might be a deeper issue with that `myGroup.myGroup` thing. Maybe naming convention is messing up the bean shutdown stuff.
Because now I get this issue and the application hangs a little bit when the application terminates.

```
14:14:12.617 [main] WARN  o.s.c.s.DefaultLifecycleProcessor - Failed to stop bean 'inputBindingLifecycle'
org.springframework.beans.factory.NoSuchBeanDefinitionException: No bean named 'my-topic.my-group.my-group.errors' available
	at org.springframework.beans.factory.support.DefaultListableBeanFactory.removeBeanDefinition(DefaultListableBeanFactory.java:988)
	at org.springframework.context.support.GenericApplicationContext.removeBeanDefinition(GenericApplicationContext.java:328)
	at org.springframework.cloud.stream.binder.AbstractMessageChannelBinder.destroyBean(AbstractMessageChannelBinder.java:796)
	at org.springframework.cloud.stream.binder.AbstractMessageChannelBinder.destroyErrorInfrastructure(AbstractMessageChannelBinder.java:783)
```

artembilan: OK. And how does it work when you remove one of those `my-group` from the `inputChannel` name ?

dzou: The error handler code does not get called.

spring-cloud-gcp-pubsub-stream-binder/src/main/java/org/springframework/cloud/gcp/stream/binder/pubsub/PubSubMessageChannelBinder.java
```diff
@@ -80,7 +80,13 @@ protected MessageHandler createProducerMessageHandler(ProducerDestination destin
 	protected MessageProducer createConsumerEndpoint(ConsumerDestination destination, String group,
 			ExtendedConsumerProperties<PubSubConsumerProperties> properties) {
 
-		return new PubSubInboundChannelAdapter(this.pubSubTemplate, destination.getName());
+		PubSubInboundChannelAdapter adapter = new PubSubInboundChannelAdapter(this.pubSubTemplate,
+				destination.getName());
+
+		ErrorInfrastructure errorInfrastructure = registerErrorInfrastructure(destination, group, properties);
```
artembilan: I suggest you to override:
```
 protected String errorsBaseName(ConsumerDestination destination, String group, C consumerProperties) {
        return destination.getName() + "." + group + ".errors";
    }
```

to remove that `+ "." + group` from the name for error channel.

This way we won't have that extra `.my-group` in your sample and the name will be consistent with other binder implementations.

dzou: Thanks, nice suggestion.

renannprado: @dzou @artembilan would be very nice if I had the option to specify the subscription name that I want to use instead of getting a name which is always prefixed with the topic name. I've found this code review while looking for the answer for this in google... maybe you guys can help.

dzou: @renannprado - Yeah we can definitely look into this more. Could you create a new feature request (or question) in our issue tracker with some more details of what you're looking for? We can continue our discussion there.

renannprado: @dzou sure I'll open an issue to continue there. Thanks!

Issue #1974, opened 2019-10-07
update spring cloud stream sample to not use annotations


elefeint:
spring cloud stream is stream-lining.

PR #1990, opened 2019-10-15
Sample app for a Spring Cloud Stream binder, functional style

elefeint:
This is a new sample app, showing the functional style of Spring Cloud Stream with Cloud Pub/Sub. It is broken into two smaller apps: source and sink. Source takes input from the browser and throws it into the channel. Sink just listens to the channel and logs results.

I removed the error handling from annotation-based sample for now; will wait to see what spring-cloud/spring-cloud-stream#1818 outcome looks like.

`spring-boot-starter-json` dependency brings in `spring-web`, which seems excessive for a tiny microservice, so I've filed spring-projects/spring-boot#18639 and spring-projects/spring-framework#23830. I could have just configured the date parsing, but this is going to be a common problem for `spring-cloud-stream` apps, so it's worth a try.

The "odd behaviors" in spring-cloud-stream were not odd; I incorrectly assumed that "input" and "output" names would be default.

Addresses most of #1974; I still owe a test.


spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-stream-functional-binder-sample/sink/src/main/resources/application.properties
```diff
@@ -0,0 +1,10 @@
+spring.cloud.stream.bindings.convertUserMessageToString-in-0.destination=[topic]
+
+spring.cloud.function.definition=convertUserMessageToString
```
olegz: If this is the only function in your AC you don't need this property, but it helps readability

spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-stream-functional-binder-sample/sink/src/main/resources/application.properties
```diff
@@ -0,0 +1,10 @@
...
+
+# If group is specified, the Pub/Sub subscription name will be [PUBSUB_TOPIC_NAME].[PUBSUB_GROUP_NAME]
+spring.cloud.stream.bindings.input.group=[subscription]
```
olegz: Should this be `spring.cloud.stream.bindings.convertUserMessageToString-in-0.group`? not 'input'

spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-stream-functional-binder-sample/sink/src/main/resources/application.properties
elefeint: Funny story: because we have that reactive scheduler bean hanging around as part of our pubsub infrastructure, SCS assumes that there are at least two eligible functions. 
I'll probably submit a feature request for that exclusion annotation (a property would have to be set by the user application, so an annotation may work better in this scenario).

spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-stream-functional-binder-sample/sink/src/main/resources/application.properties
elefeint: Same reason. We don't get to use shortcut input/output names because of that zombie scheduler bean hanging around the context, making it look like two eligible functions.

spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-stream-functional-binder-sample/sink/src/main/resources/application.properties
olegz: Right, i could predict your response given the history ;)

spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-stream-functional-binder-sample/sink/src/main/resources/application.properties
olegz: Not following. What I am saying is that the binding name is `convertUserMessageToString-in-0` not 'input'. In other words the above property has no effect in the current config 
no matter wha the value is

elefeint: Oh, that's right; it's the wrong binding name, so anonymous group will be used. I'll fix it.
I thought you were asking why a simple `spring.cloud.stream.bindings.input.destination` could not be used instead of `spring.cloud.stream.bindings.convertUserMessageToString-in-0.destination`.

olegz: Also, if you do want to use more descriptive names you can use `spring.cloud.stream.function.bindings.convertUserMessageToString-in-0=input`. With that you just mapped generated name to `input` so from that point you can refer to binding name as `input`.

spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-stream-functional-binder-sample/pom.xml
```diff
@@ -0,0 +1,33 @@
...
+    <groupId>org.springframework.cloud</groupId>
+    <version>1.2.0.BUILD-SNAPSHOT</version>
+  </parent>
+  <modelVersion>4.0.0</modelVersion>
+
+  <artifactId>spring-cloud-gcp-pubsub-stream-functional-binder-sample</artifactId>
+  <packaging>pom</packaging>
+  <name>Spring Cloud GCP Stream Functional Style Pub/Sub Code Sample</name>
+  <modules>
+    <module>source</module>
```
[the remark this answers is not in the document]
elefeint: Yeah, I'll rename them to match the convention.

spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-stream-functional-binder-sample/source/pom.xml
```diff
@@ -0,0 +1,43 @@
...
+
+  <dependencies>
+    <dependency>
+      <groupId>org.springframework.cloud</groupId>
+      <artifactId>spring-cloud-gcp-pubsub-stream-binder</artifactId>
+    </dependency>
+
+    <dependency>
+      <groupId>org.springframework.boot</groupId>
+      <artifactId>spring-boot-starter-webflux</artifactId>
```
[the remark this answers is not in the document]
elefeint: For `WebSocket` support.

spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-stream-functional-binder-sample/source/src/main/java/com/example/Source.java
```diff
@@ -0,0 +1,49 @@
...
+@SpringBootApplication
+public class Source {
+
+	@Autowired
+	@Qualifier("postOffice")
+	private ConcurrentLinkedQueue<UserMessage> postOffice;
+
+	@Bean
+	Supplier<UserMessage> generateUserMessages() {
+		return () -> postOffice.poll();
```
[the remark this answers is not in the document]
elefeint: @olegz Actually, now that I think about it more, is "inputting data from UI and sending it into a stream" a valid Spring Cloud Stream usage scenario at all?
In real world, it would be something like files arriving into file system, which you'd hook up with a SCS `Processor`, so it would not be a `Source` at all.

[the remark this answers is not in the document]
olegz: Yes, very valid and very possible. Here is all the code you need
```
@SpringBootApplication
@Controller
@RequestMapping("/home")
public class DemoFunctionStreamWebApplication {

	public static void main(String[] args) {
		SpringApplication.run(DemoFunctionStreamWebApplication.class, args);
	}

	EmitterProcessor<String> processor = EmitterProcessor.create();

	@RequestMapping(value="/send")
	@ResponseBody
	public String foo(@RequestParam("name") String name){
		System.out.println("Hello " + name);
		processor.onNext(name);
		return name;
	}

	@Bean
	public Supplier<Flux<String>> supplier() {
		return () -> processor;
	}
}
```
Is that what you were asking about?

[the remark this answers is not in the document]
elefeint: Yes! I'll use this way instead of opening a socket connection; this is simpler.
I like `EmitterProcessor` better than the queue, too, so I'll use it in the example. I did not think to use reactor primitives.

spring-cloud-gcp-samples/pom.xml
```diff
@@ -49,7 +49,8 @@
 		<module>spring-cloud-gcp-data-multi-sample</module>
 		<module>spring-cloud-gcp-data-firestore-sample</module>
 		<module>spring-cloud-gcp-bigquery-sample</module>
-	</modules>
+		<module>spring-cloud-gcp-pubsub-stream-functional-binder-sample</module>
```
meltsufin: Looking at  `spring-cloud-gcp-pubsub-binder-sample` and `spring-cloud-gcp-pubsub-polling-binder-sample`,
Maybe it should be `spring-cloud-gcp-pubsub-functional-binder-sample` for consistency sake.

spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-stream-functional-binder-sample/spring-cloud-gcp-pubsub-stream-functional-binder-sample-sink/src/main/java/com/example/Sink.java
```diff
@@ -0,0 +1,49 @@
...
+
+import org.springframework.boot.SpringApplication;
+import org.springframework.boot.autoconfigure.SpringBootApplication;
+import org.springframework.context.annotation.Bean;
+
+/**
+ * All-in-one Spring Boot app with a Spring Cloud Stream sink.
+ */
+@SpringBootApplication
+public class Sink {
```
meltsufin: Maybe `FunctionalSinkApplication` so that the logs will be easier to attribute to this sample app.

spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-stream-functional-binder-sample/spring-cloud-gcp-pubsub-stream-functional-binder-sample-source/src/main/java/com/example/SourceApplication.java
```diff
@@ -0,0 +1,47 @@
...
+import org.springframework.context.annotation.Bean;
+
+/**
+ * Spring Boot application for running the Spring Cloud Stream source.
+ *
+ * <p>This class bootstraps the Spring Boot application and creates the {@link EmitterProcessor}
+ * bean that is used for communication between {@link FrontendController} and {@link Source}.
+ */
+@SpringBootApplication
+public class SourceApplication {
```
meltsufin: `FunctionalSourceApplication`?

spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-stream-functional-binder-sample/README.adoc
```diff
@@ -0,0 +1,30 @@
...
+In the link:src/main/resources/application.properties[application.properties] file, a topic and group name are already preconfigured for you (it's `my-topic` and `my-group`, so the subscription name used by the `Sink` is `my-topic.my-group`).
+The topic and subscription will be created in your account if they do not already exist.
+
+1. In separate terminal windows, start the `spring-cloud-gcp-pubsub-stream-functional-binder-sample-source` and `spring-cloud-gcp-pubsub-stream-functional-binder-sample-sink` applications with the `$ mvn spring-boot:run` command in the same folder as the apps' `pom.xml` files.
+
+5. Browse to the source application at `http://localhost:8080`, type in a message and a user name, and press the `Post it!` button.
+
+6. Verify in the `spring-cloud-gcp-pubsub-stream-functional-binder-sample-sink` app's logs that the message was received:
++
+`New message received from [USERNAME] at [TIME_MESSAGE_SENT]: [MESSAGE]`
```
meltsufin: Do we need a cleanup step?
Will those auto-created topics and subscriptions just hang around?

meltsufin:
Forgot to ask... do you plan to have the tests in a separate PR?

spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-stream-functional-binder-sample/README.adoc
```diff
@@ -0,0 +1,30 @@
+= Spring Cloud GCP Stream Binder (functional style) for Pub/Sub Code Sample
```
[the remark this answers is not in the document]
elefeint: "annotation-less"? "without annotations"? "v3.0 style"?

spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-stream-functional-binder-sample/spring-cloud-gcp-pubsub-stream-functional-binder-sample-sink/src/main/java/com/example/Sink.java
```diff
@@ -0,0 +1,49 @@
...
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
+import org.springframework.boot.SpringApplication;
+import org.springframework.boot.autoconfigure.SpringBootApplication;
+import org.springframework.context.annotation.Bean;
+
+/**
+ * All-in-one Spring Boot app with a Spring Cloud Stream sink.
+ */
```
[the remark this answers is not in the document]
elefeint: Oh yeah; I always forget to put a pawprint on these things.

spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-stream-functional-binder-sample/spring-cloud-gcp-pubsub-stream-functional-binder-sample-sink/src/main/java/com/example/Sink.java
```diff
@@ -0,0 +1,49 @@
...
+public class Sink {
+
+	private static final Log LOGGER = LogFactory.getLog(Sink.class);
+
+	@Bean
+	public Consumer<UserMessage> convertUserMessageToString() {
+		return userMessage -> {
+
+			LOGGER.info(String.format("New message received from %s: %s at %s",
+					userMessage.getUsername(), userMessage.getBody(), userMessage.getCreatedAt()));
```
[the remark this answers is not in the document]
elefeint: You mean I went through the trouble of copy and pasting it... I'd rather remove the `toString()` implementation; it's not useful, and I have two of those duplicated classes that might as well be as small as possible. Sonar gave me a side eye about them.

spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-stream-functional-binder-sample/spring-cloud-gcp-pubsub-stream-functional-binder-sample-sink/src/main/java/com/example/Sink.java
```diff
@@ -0,0 +1,49 @@
...
+/**
+ * All-in-one Spring Boot app with a Spring Cloud Stream sink.
+ */
+@SpringBootApplication
+public class Sink {
+
+	private static final Log LOGGER = LogFactory.getLog(Sink.class);
+
+	@Bean
+	public Consumer<UserMessage> convertUserMessageToString() {
```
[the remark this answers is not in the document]
elefeint: Sure. I had an elaborate setup there at some point with two chained functions, with one throwing exceptions. I'll probably bring it back some day when adding exception handling back.

spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-stream-functional-binder-sample/spring-cloud-gcp-pubsub-stream-functional-binder-sample-sink/src/main/resources/application.properties
```diff
@@ -0,0 +1,15 @@
+# Binding name is mapped to a simpler shortcut name "input".
+spring.cloud.stream.function.bindings.convertUserMessageToString-in-0=input
```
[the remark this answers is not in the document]
elefeint: Then you have to refer to `convertUserMessageToString-in-0` everywhere else. We don't have to do it, but it's a new feature in SCS, so it might be nice to leave as an option. I added it into the sink and not into the source because the source has zero other configuration, while the sink also has the group subscription name.

So, without aliasing, configuration looks like this:
```
spring.cloud.stream.bindings.logUserMessage-in-0.destination=my-topic
spring.cloud.stream.bindings.logUserMessage-in-0.group=my-group
```

And with aliasing, it looks like this:
```
spring.cloud.stream.function.bindings.logUserMessage-in-0=input
spring.cloud.stream.bindings.input.destination=my-topic
spring.cloud.stream.bindings.input.group=my-group
```

This looks more dramatic once your chain -- function name becomes something like `logUserMessage-in-0|processUserMessage-in-0`.

spring-cloud-gcp-samples/pom.xml
elefeint: I want to rename `spring-cloud-gcp-pubsub-binder-sample` and `spring-cloud-gcp-pubsub-polling-binder-sample`, actually. The original name of "pubsub binder" is confusing -- what is this binder we speak about?

The new names would be `spring-cloud-gcp-pubsub-stream-binder-sample` and `spring-cloud-gcp-pubsub-polling-stream-binder-sample`.

Or even ditch "binder" completely -- it's an implementation detail, completely irrelevant to end user.  So: `spring-cloud-gcp-pubsub-stream-sample` and `spring-cloud-gcp-pubsub-polling-stream-sample` and `spring-cloud-gcp-pubsub-stream-functional-sample`.

spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-stream-functional-binder-sample/README.adoc
elefeint: Auto-created anonymous subscriptions get cleaned up, but manually specified topics and topic.group subscriptions stick around. I'll add a cleanup step.

elefeint:
> Forgot to ask... do you plan to have the tests in a separate PR?

... and that's how I know that you check e-mails in chronological order.

spring-cloud-gcp-samples/pom.xml
meltsufin: I think as long as the actual module name that this is a sample for is called `spring-cloud-gcp-pubsub-stream-binder`, we should keep it as a base name for the samples, even if that means renaming some of the older ones.

BTW, as you do the renaming and adding new samples, please don't forget to add/edit the links in refdocs.

spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-stream-functional-binder-sample/README.adoc
```diff
@@ -23,8 +23,14 @@ The topic and subscription will be created in your account if they do not alread
...
 
-6. Verify in the `spring-cloud-gcp-pubsub-stream-functional-binder-sample-sink` app's logs that the message was received:
+3. Verify in the `spring-cloud-gcp-pubsub-stream-functional-binder-sample-sink` app's logs that the message was received:
 +
 `New message received from [USERNAME] at [TIME_MESSAGE_SENT]: [MESSAGE]`
+
+
+== Cleanup
+
+Delete the test topic (`my-topic`) and subscription (`my-topic.my-group`).
```
meltsufin: I feel like that if the sample creates these resources automatically, it should also clean them up automatically.

elefeint: This is problematic. There are two applications -- source and sink, which could both autocreate the named topic. If either app deletes the topic, the other app could still be running with the assumption that the topic exists.
I've stopped one of the two apps repeatedly as part of my own testing, and it was convenient that the other app ran without hiccups, waiting for me to start the first one again.

docs/src/main/asciidoc/spring-stream.adoc
```diff
@@ -76,6 +76,69 @@ spring.cloud.stream.gcp.pubsub.bindings.events.consumer.auto-create-resources=tr
...
+```
+@Bean
+Supplier<Flux<UserMessage>> generateUserMessages() {
+  return () -> /* flux creation logic */;
+}
+```
+
+A processor application works similarly to a source application, except it is triggered by presence of a `Function` bean.
+
+WARNING: When using the `spring-cloud-gcp-pubsub-stream-binder` dependency, configuring `spring.cloud.function.definition` property to indicate function bean is *required*, even though Spring Cloud Stream has ability to autodiscover functional beans.
```
meltsufin: This seems like a major headache and defeats the promised simplicity of auto-discovering the function.
How bad would it be to hide the `pubSubReactiveScheduler` bean from the application context by creating it directly inside the `pubSubReactiveFactory` bean definition?

```suggestion
WARNING: Even though Spring Cloud Stream is able to autodiscover functional beans, when using the `spring-cloud-gcp-pubsub-stream-binder` dependency, configuring the `spring.cloud.function.definition` property to indicate the function bean is *required*.
```

docs/src/main/asciidoc/spring-stream.adoc
```diff
@@ -118,4 +181,5 @@ public class SinkExample {
 Sample applications are available:
 
 * For https://github.com/spring-cloud/spring-cloud-gcp/tree/master/spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-binder-sample[streaming input].
```
meltsufin:
```suggestion
* For https://github.com/spring-cloud/spring-cloud-gcp/tree/master/spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-binder-sample[streaming input, annotation-based].
```

docs/src/main/asciidoc/spring-stream.adoc
elefeint: Need to figure out the lifecycle of reactor schedulers; the reason for having schedulers as beans is to have Spring close them on application shutdown.

I created #1997 to track. 
Once this sample is in, I'll also create an issue in SCS repo, pointing to it as an example of the problem.

docs/src/main/asciidoc/spring-stream.adoc
```diff
@@ -76,6 +76,69 @@ spring.cloud.stream.gcp.pubsub.bindings.events.consumer.auto-create-resources=tr
...
+@Bean
+Supplier<Flux<UserMessage>> generateUserMessages() {
+  return () -> /* flux creation logic */;
+}
+```
+
+A processor application works similarly to a source application, except it is triggered by presence of a `Function` bean.
+
+WARNING: Even though Spring Cloud Stream is able to autodiscover functional beans, when using the `spring-cloud-gcp-pubsub-stream-binder` dependency, configuring the `spring.cloud.function.definition` property to indicate the function bean is *required*.
+If this property is omitted, you will see a warning `Found more then one function beans in BeanFactory: [_your-bean-name_, pubSubReactiveScheduler]`, and the function returned by _your-bean-name_ will never be bound to a Cloud Pub/Sub topic or subscription.
```
artembilan: ??? Seems for me we have agreed that this is not a case any more.
See #1997

PR #1995, opened 2019-10-18
Test for the functional stream binder sample

### LINKED ISSUE #1974: update spring cloud stream sample to not use annotations

elefeint:
Because both source and sink application dependencies have `application.properties` files, these are excluded with a custom resource loader, and the properties are added back to each app programmatically.

Followup to #1990 .
Fixes #1974.


elefeint:
I changed the base to the parent branch, so it's better now. But yes, this can wait.

PR #2001, opened 2019-10-29
Fix the Datastore Ancestor Key Query

### LINKED ISSUE #1999: Recursive Relationships Are Not Supported by Descendants
If you have an object with Descendants of the same class type:

```
class Test {
    @Id 
    private Key id;
    @Descendants
    private List<Test> children;
}
```

When you try to retrieve this object using a Repository findById method, the List children will be populated with a single element: the parent itself.

dzou:
Fixes the Datastore ancestor key query to correctly use entity keys with the path set to the entity itself (omit the fully-qualified path in the key passed into the ancestor query).

Fixes #1999

### DIFF
--- a/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
+++ b/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
@@ -645,10 +645,20 @@ public class DatastoreTemplate implements DatastoreOperations, ApplicationEventP
 					Class descendantType = descendantPersistentProperty
 							.getComponentType();
 
+					Key entityKey = (Key) entity.getKey();
+
+					Key.Builder ancestorLookupKey;
+					if (entityKey.hasName()) {
+						ancestorLookupKey = Key.newBuilder(entityKey.getProjectId(), entityKey.getKind(), entityKey.getName());
+					} else {
+						ancestorLookupKey = Key.newBuilder(entityKey.getProjectId(), entityKey.getKind(), entityKey.getId());
+					}
+					ancestorLookupKey.setNamespace(entityKey.getNamespace());
+
 					EntityQuery descendantQuery = Query.newEntityQueryBuilder()
 							.setKind(this.datastoreMappingContext
 									.getPersistentEntity(descendantType).kindName())
-							.setFilter(PropertyFilter.hasAncestor((Key) entity.getKey()))
+							.setFilter(PropertyFilter.hasAncestor(ancestorLookupKey.build()))
 							.build();
 
 					List entities = convertEntitiesForRead(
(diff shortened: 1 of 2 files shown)


spring-cloud-gcp-data-datastore/src/test/java/org/springframework/cloud/gcp/data/datastore/it/DatastoreIntegrationTests.java
```diff
@@ -718,6 +718,20 @@ public void readOnlyCountTest() {
...
+	@Test
+	public void sameClassDescendantsTest() {
+		Student entity3 = new Student(Collections.EMPTY_LIST);
+		Student entity2 = new Student(Collections.EMPTY_LIST);
+		Student entity1 = new Student(Arrays.asList(entity2, entity3));
+		School school = new School(1L, Arrays.asList(entity1));
+		this.datastoreTemplate.save(school);
+
+		School readSchool = this.datastoreTemplate.findById(school.id, School.class);
+		Student child = readSchool.descendants.get(0);
```
dmitry-s: add assertion to verify that `readSchool.descendants` contain only `entity1`

dzou: Done.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
```diff
@@ -645,10 +645,20 @@ else if (referencePersistentProperty.isCollectionLike()) {
 					Class descendantType = descendantPersistentProperty
 							.getComponentType();
 
+					Key entityKey = (Key) entity.getKey();
```
meltsufin: Please refactor this into a helper method like `removeAncestorsFromKey(Key)`.

spring-cloud-gcp-data-datastore/src/test/java/org/springframework/cloud/gcp/data/datastore/it/DatastoreIntegrationTests.java
```diff
@@ -872,6 +886,65 @@ public int hashCode() {
...
+	}
+}
+
+@Entity
+class Student {
+	@Id
+	public Key id;
+
+	@Descendants
+	public List<Student> subordinates;
```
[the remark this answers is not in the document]
dzou: Done.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
dzou: Done.

PR #2010, opened 2019-11-04
Make all thread pools create daemon threads

### LINKED ISSUE #2005: gcp-pubsub-subscriber thread starts on PubSubTemplate creation; never stops
This is a continuation of #1503. There is also a similar (but unrelated, as Spring Cloud GCP creates its own `SubscriberStub`) issue in the client library googleapis/google-cloud-java#6570.

`PubSubSubscriberTemplate` constructor creates a `SubscriberStub`, which starts at least one thread. The `SubscriberStub` is never closed. For a web application, this is not noticeable, but for a terminal app that expects JVM to close normally when its work is done, the result is a hanging app. 

I thought we've fixed this in #1503, but now I don't know how that solution (making schedulers spring managed beans) ever worked. Spring's bean destruction happens when on JVM shutdown hook, and JVM won't shut down while there is still a non-daemon thread running.

There are solutions of varying invasiveness to this:
1) Document the need to shut down the executor manually, as in this [comment](https://github.com/spring-cloud/spring-cloud-gcp/issues/1503#issuecomment-470142129).
2) For sure at least call `this.subscriberStub.close()` in the `destroy()` method in `PubSubSubscriberTemplate`, so it cleans up properly for a web app.
3) (this one needs thought and analysis of how/when these threads are created) Work with the client library to avoid starting the threads unless .subscribe() is called. Are the threads created only for the streaming pull? Or are they also used for blocking operations?

elefeint:
Currently the thread pool for Pub/Sub's `SubscriberStub` keeps JVM from shutting down, causing non-web application to "hang".

This PR makes all Spring Cloud GCP threads daemon threads that do not prevent a Spring application from stopping.

This is a change in behavior for Pub/Sub-dependent command line applications that subscribe to a Pub/Sub topic with `PubSubTemplate.subscribe()`, and expect the application to run until externally terminated.

Fixes #2005.

### DIFF
--- a/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
+++ b/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
@@ -116,6 +116,7 @@ public class GcpPubSubAutoConfiguration {
 		ThreadPoolTaskScheduler scheduler = new ThreadPoolTaskScheduler();
 		scheduler.setPoolSize(this.gcpPubSubProperties.getPublisher().getExecutorThreads());
 		scheduler.setThreadNamePrefix("gcp-pubsub-publisher");
+		scheduler.setDaemon(true);
 		return scheduler;
 	}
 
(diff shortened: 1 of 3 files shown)


spring-cloud-gcp-pubsub/src/test/java/org/springframework/cloud/gcp/pubsub/core/PubSubSubscriberTemplateTests.java
```diff
@@ -0,0 +1,67 @@
...
+/**
+ * Tests for {@link PubSubSubscriberTemplate} for functionality not exposed in
+ * {@link PubSubTemplate}.
+ *
+ * @author Elena Felder
+ *
+ * @since 1.2
+ */
+@RunWith(MockitoJUnitRunner.class)
+public class PubSubSubscriberTemplateTests {
```
meltsufin: Wasn't this part of some other PR?

elefeint: ... and this is why it's a best practice to delete branches after merging to master ...
// shame cube

Issue #1978, opened 2019-10-09
Trace Sample generates extra traces


meltsufin:
When running the sample, we expect to just see one trace per user request, but instead we see many addition traces called `async` that seem to have nothing to do with the request.

![image](https://user-images.githubusercontent.com/17732492/66518509-0672ba00-eaab-11e9-8754-e7f6af37439a.png)

meltsufin:
@adriancole FYI

meltsufin:
One workaround is to exclude the `traceSenderThreadPool` bean from being traced by setting this property:
```
spring.sleuth.async.ignoredBeans=traceSenderThreadPool
```

meltsufin:
Ultimately we need to make sure that all of our thread pool beans are not instrumented by Sleuth.

meltsufin:
We need to add `spring.sleuth.scheduled.enabled=false` to our sample and also document this in refdocs.

PR #2011, opened 2019-11-04
Fix Trace sample not to generate 'async' traces

meltsufin:
Also document the use of `spring.sleuth.scheduled.enabled=false`, which
is used to disable automatic instrumentation of executors by Sleuth.

Fixes: #1978.

### DIFF
--- a/docs/src/main/asciidoc/trace.adoc
+++ b/docs/src/main/asciidoc/trace.adoc
@@ -91,7 +91,10 @@ For example, when you are testing to see the traces are going through, you can s
 ----
 spring.sleuth.sampler.probability=1                     # Send 100% of the request traces to Stackdriver.
 spring.sleuth.web.skipPattern=(^cleanup.*|.+favicon.*)  # Ignore some URL paths.
+spring.sleuth.scheduled.enabled=false                   # disable executor 'async' traces
 ----
+
+WARNING: By default, Spring Cloud Sleuth auto-configuration instruments executor beans, which causes many traces with the name `async` to appear in Stackdriver Trace. This is especially a problem because our starter comes with an executor. To avoid this noise, please disable automatic instrumentation of executors via `spring.sleuth.async.enabled=false` in your application configuration.
 
 Spring Cloud GCP Trace does override some Sleuth configurations:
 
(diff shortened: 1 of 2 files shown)


docs/src/main/asciidoc/trace.adoc
```diff
@@ -91,8 +91,11 @@ For example, when you are testing to see the traces are going through, you can s
 ----
 spring.sleuth.sampler.probability=1                     # Send 100% of the request traces to Stackdriver.
 spring.sleuth.web.skipPattern=(^cleanup.*|.+favicon.*)  # Ignore some URL paths.
+spring.sleuth.scheduled.enabled=false                   # disable executor 'async' traces
 ----
 
+WARNING: By default, Spring Cloud Sleuth auto-configuration instruments executor beans, which causes many traces with the name `async` to appear in Stackdriver Trace. This is especially a problem because our starter comes with an executor. To avoid this noise, please disable automatic instrumentation of executors via `spring.sleuth.async.enabled=false` in your application configuration.
+
```
dzou: Should `spring.sleuth.async.enabled=false` be `spring.sleuth.scheduled.enabled=false`?

docs/src/main/asciidoc/trace.adoc
```diff
@@ -91,8 +91,11 @@ For example, when you are testing to see the traces are going through, you can s
 ----
 spring.sleuth.sampler.probability=1                     # Send 100% of the request traces to Stackdriver.
 spring.sleuth.web.skipPattern=(^cleanup.*|.+favicon.*)  # Ignore some URL paths.
+spring.sleuth.scheduled.enabled=false                   # disable executor 'async' traces
 ----
 
+WARNING: By default, Spring Cloud Sleuth auto-configuration instruments executor beans, which causes many traces with the name `async` to appear in Stackdriver Trace. This is especially a problem because our starter comes with an executor. To avoid this noise, please disable automatic instrumentation of executors via `spring.sleuth.async.enabled=false` in your application configuration.
```
dzou: Should `false` be the default setting for the application?

meltsufin: It's not our property. So, I don't know if we should change the default for it.
@adriancole WDYT?

docs/src/main/asciidoc/trace.adoc
meltsufin: Nice catch man! I dunno how I ended up copy-pasting the wrong one.

meltsufin:
@elefeint The problem with a private field is that it becomes closed to overwriting.

PR #2013, opened 2019-11-05
hide default scheduler from application context

### LINKED ISSUE #1997: Consider hiding pubSubReactiveScheduler in Pub/Sub autoconfig
The scheduler is a cached scheduler, which implements a `Supplier`. Since Spring Cloud Stream autodiscovers functional beans, it always finds `pubSubReactiveScheduler`, meaning that any SCS app that uses the Cloud Pub/Sub binder will not be able to take advantage of autodiscovery and will have to specify `spring.cloud.function.definition` property to disambiguate.

elefeint:
This is what I had in mind for hiding scheduler beans from application context.

The problem this immediately solves is Spring Cloud Stream accidentally finding this scheduler (because the implementation is a `Supplier`), but we could also use this approach to keep schedulers from getting instrumented in Trace and Pub/Sub.

Fixes #1997.

### DIFF
--- a/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubReactiveAutoConfiguration.java
+++ b/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubReactiveAutoConfiguration.java
@@ -16,6 +16,10 @@
 
 package org.springframework.cloud.gcp.autoconfigure.pubsub;
 
+import java.util.Optional;
+
+import javax.annotation.PreDestroy;
+
 import reactor.core.publisher.Flux;
 import reactor.core.scheduler.Scheduler;
 import reactor.core.scheduler.Schedulers;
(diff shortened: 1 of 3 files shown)


spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubReactiveAutoConfiguration.java
```diff
@@ -43,17 +47,20 @@
...
-	}
+	private Scheduler defaultPubSubReactiveScheduler = Schedulers.newElastic("pubSubReactiveScheduler");
 
 	@Bean
 	@ConditionalOnMissingBean
 	public PubSubReactiveFactory pubSubReactiveFactory(
-		PubSubSubscriberTemplate subscriberTemplate, @Qualifier("pubSubReactiveScheduler") Scheduler scheduler) {
-		return new PubSubReactiveFactory(subscriberTemplate, scheduler);
+			PubSubSubscriberTemplate subscriberTemplate,
+			@Qualifier("pubSubReactiveScheduler") Optional<Scheduler> scheduler) {
```
meltsufin: Does `Optional<Scheduler>` work for this? I thought you needed to use `ObjectProvider`, but this is cool.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubReactiveAutoConfiguration.java
```diff
@@ -43,17 +47,20 @@
 @ConditionalOnProperty(value = "spring.cloud.gcp.pubsub.reactive.enabled", matchIfMissing = true)
 public class GcpPubSubReactiveAutoConfiguration {
 
-	@Bean
-	@ConditionalOnMissingBean(name = "pubSubReactiveScheduler")
-	Scheduler pubSubReactiveScheduler() {
-		return Schedulers.elastic();
-	}
+	private Scheduler defaultPubSubReactiveScheduler = Schedulers.newElastic("pubSubReactiveScheduler");
```
meltsufin: A slight problem here is that if the user provides their own scheduler, we still have this scheduler around consuming resources. Maybe create it only when known to be needed?

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubReactiveAutoConfiguration.java
elefeint: Yep, it works. Even declaring something `@Nullable` should work the same way.
(I actually just did it automatically because it would work in Guice, and... it worked.)

PR #2020, opened 2019-11-07
Adding original message to the throwble for Pub/Sub publish failures.

sateekot:
Adding original message to Future provided by Publisher if any failures to publish events to Pub/Sub.

### DIFF
--- a/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubDeliveryException.java
+++ b/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubDeliveryException.java
@@ -33,8 +33,8 @@ public class PubSubDeliveryException extends PubSubException {
 	@Nullable
 	private final PubsubMessage failedMessage;
 
-	public PubSubDeliveryException(PubsubMessage pubsubMessage, Throwable cause) {
-		super(null, cause);
+	public PubSubDeliveryException(PubsubMessage pubsubMessage, String description, Throwable cause) {
+		super(description, cause);
 		this.failedMessage = pubsubMessage;
 	}
 
@@ -42,10 +42,4 @@ public class PubSubDeliveryException extends PubSubException {
 		return this.failedMessage;
 	}
 
-	@Override
-	public String toString() {
-
-		return super.toString() + (this.failedMessage == null ? ""
-				: (", failedMessage=" + this.failedMessage));
-	}
 }
--- a/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubDeliveryException.java
+++ b/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubDeliveryException.java
@@ -1,0 +1,76 @@
+/*
+ * Copyright 2017-2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.gcp.pubsub.core;
+
+import com.google.pubsub.v1.PubsubMessage;
+
+import org.springframework.lang.Nullable;
+
+/**
+ * The Spring Google Cloud Pub/Sub specific {@link PubSubException}.
+ * Handles failures while publishing events to Pub/Sub.
+ *
+ * @author SateeshKumar Kota
+ *
+ * @since 1.2
+ */
+public class PubSubDeliveryException extends PubSubException {
+
+	@Nullable
+	private final PubsubMessage failedMessage;
+
+	public PubSubDeliveryException(PubsubMessage pubsubMessage) {
+		super(null, null);
+		this.failedMessage = pubsubMessage;
+	}
+
+	public PubSubDeliveryException(String description) {
+		super(description);
+		this.failedMessage = null;
+	}
+
+	public PubSubDeliveryException(@Nullable String description, @Nullable Throwable cause) {
+		super(description, cause);
+		this.failedMessage = null;
+	}
+
+	public PubSubDeliveryException(PubsubMessage pubsubMessage, String description) {
+		super(description);
+		this.failedMessage = pubsubMessage;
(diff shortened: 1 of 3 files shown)


pivotal-issuemaster:
@sateekot Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-gcp&pullRequestId=2020)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-cloud/spring-cloud-gcp&pullRequestId=2020) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@sateekot Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-gcp&pullRequestId=2020)!

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/publisher/PubSubPublisherTemplate.java
```diff
@@ -103,7 +106,10 @@ public void setMessageConverter(PubSubMessageConverter pubSubMessageConverter) {
 			@Override
 			public void onFailure(Throwable throwable) {
 				LOGGER.warn("Publishing to " + topic + " topic failed.", throwable);
-				settableFuture.setException(throwable);
+				Message<?> originalMessage = MessageBuilder.withPayload(pubsubMessage).build();
+				MessageHandlingException messageHandlingException = new MessageHandlingException(
```
meltsufin: Can you use `MessageDeliveryException` instead please?

artembilan: Well, I wouldn't do that.
I mean we are good to have a `PubSubDeliveryException` as an extension of the `PubSubException`.
My point is to not wrap a `PubsubMessage` into a `Message<?>`: an original request is not Spring Messaging.
So, let's not abuse non-related API, but we definitely can borrow an idea! 😄

sateekot: Thank you for your comments. I have updated the code accordingly. Can you please check the updated commit.

sateekot:
@artembilan I am not sure how to update this feature in docs, Can you please give me some pointers.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubDeliveryException.java
```diff
@@ -0,0 +1,76 @@
...
+		super(description, cause);
+		this.failedMessage = null;
+	}
+
+	public PubSubDeliveryException(PubsubMessage pubsubMessage, String description) {
+		super(description);
+		this.failedMessage = pubsubMessage;
+	}
+
+	public PubSubDeliveryException(PubsubMessage pubsubMessage, Throwable cause) {
```
meltsufin: Let's just keep the one constructor we currently use. We can always add more. Otherwise, we have all this code that's missing coverage.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubDeliveryException.java
```diff
@@ -0,0 +1,76 @@
...
+		super(description, cause);
+		this.failedMessage = pubsubMessage;
+	}
+
+	public PubsubMessage getFailedMessage() {
+		return this.failedMessage;
+	}
+
+	@Override
+	public String toString() {
```
meltsufin: We don't typically overwrite the `toString()` for exceptions. Who is going to call this method?

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubDeliveryException.java
sateekot: removed constructors.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/PubSubDeliveryException.java
sateekot: If we don't override toString(), the newly added test case is failing as we are checking the message in the exception.

spring-cloud-gcp-pubsub/src/test/java/org/springframework/cloud/gcp/pubsub/core/PubSubTemplateTests.java
```diff
@@ -201,6 +202,14 @@ public void testSend_onFailure() {
 				.hasMessageContaining("future failed.");
 	}
 
+	@Test
+	public void testPublish_onFailureWithPayload() {
+		ListenableFuture<String> future = this.pubSubTemplate.publish("testTopic", this.pubsubMessage);
+		this.settableApiFuture.setException(new Exception("Publish failed"));
+		assertThatThrownBy(() -> future.get()).isInstanceOf(ExecutionException.class).hasMessageContaining("permanating")
```
artembilan: I think we need to check in the test that we have already that new `PubSubDeliveryException`  with expected `this.pubsubMessage` populated. I believe that was a main goal of this fix.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/publisher/PubSubPublisherTemplate.java
```diff
@@ -103,7 +105,8 @@ public void setMessageConverter(PubSubMessageConverter pubSubMessageConverter) {
 			@Override
 			public void onFailure(Throwable throwable) {
 				LOGGER.warn("Publishing to " + topic + " topic failed.", throwable);
-				settableFuture.setException(throwable);
+				PubSubDeliveryException pubSubDeliveryException = new PubSubDeliveryException(pubsubMessage, throwable.getMessage(), throwable);
```
artembilan: No, I don't want to see the same `throwable.getMessage()` here again on top of stack trace.
I can get it in the stack trace anyway parsing a `cause`.
What I want to see is an explicit message indicating that exactly here in the `PubSubTemplate` we couldn't delivered a message.

meltsufin: +1
I think what you're logging should work here as well for the message:
`"Publishing to " + topic + " topic failed."`

sateekot: Updated the code.

PR #2008, opened 2019-11-01
move PublicClassMapper to our package

dmitry-s:
fixes #2007

PR #2022, opened 2019-11-07
Move class conversion from duplicate package name; make autoconfigurable

elefeint:
We may want to merge #2008 instead, if the dummy project is preferred.

### DIFF
--- a/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/firestore/GcpFirestoreAutoConfiguration.java
+++ b/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/firestore/GcpFirestoreAutoConfiguration.java
@@ -22,6 +22,7 @@ import com.google.api.gax.core.CredentialsProvider;
 import com.google.api.gax.grpc.InstantiatingGrpcChannelProvider;
 import com.google.cloud.firestore.Firestore;
 import com.google.cloud.firestore.FirestoreOptions;
+import com.google.cloud.firestore.spi.v1.FirestoreRpc;
 import com.google.firestore.v1.FirestoreGrpc;
 import io.grpc.ManagedChannel;
 import io.grpc.ManagedChannelBuilder;
@@ -37,6 +38,7 @@ import org.springframework.cloud.gcp.autoconfigure.core.GcpContextAutoConfigurat
 import org.springframework.cloud.gcp.core.DefaultCredentialsProvider;
 import org.springframework.cloud.gcp.core.GcpProjectIdProvider;
 import org.springframework.cloud.gcp.core.UserAgentHeaderProvider;
+import org.springframework.cloud.gcp.data.firestore.FirestoreClassMapper;
 import org.springframework.cloud.gcp.data.firestore.FirestoreTemplate;
 import org.springframework.cloud.gcp.data.firestore.mapping.FirestoreMappingContext;
 import org.springframework.cloud.gcp.data.firestore.transaction.ReactiveFirestoreTransactionManager;
@@ -122,14 +124,22 @@ public class GcpFirestoreAutoConfiguration {
 
 		@Bean
 		@ConditionalOnMissingBean
-		public FirestoreMappingContext firestoreMappingContext() {
-			return new FirestoreMappingContext();
+		public FirestoreMappingContext firestoreMappingContext(FirestoreClassMapper mapper) {
+			return new FirestoreMappingContext(mapper);
 		}
 
 		@Bean
 		@ConditionalOnMissingBean
-		public FirestoreTemplate firestoreTemplate(FirestoreGrpc.FirestoreStub firestoreStub) {
-			return new FirestoreTemplate(firestoreStub, GcpFirestoreAutoConfiguration.this.firestoreRootPath);
+		public FirestoreClassMapper firestoreClassMapper(FirestoreOptions firestoreOptions, Firestore firestore) {
+			return new FirestoreClassMapper(firestoreOptions, (FirestoreRpc) firestoreOptions.getRpc());
+		}
+
+		@Bean
+		@ConditionalOnMissingBean
+		public FirestoreTemplate firestoreTemplate(
+				FirestoreGrpc.FirestoreStub firestoreStub, FirestoreClassMapper classMapper) {
+			return new FirestoreTemplate(
+					firestoreStub, GcpFirestoreAutoConfiguration.this.firestoreRootPath, classMapper);
 		}
 
 		@Bean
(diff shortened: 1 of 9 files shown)


spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/firestore/GcpFirestoreAutoConfiguration.java
```diff
@@ -122,14 +124,22 @@ public Firestore firestore(FirestoreOptions firestoreOptions) {
...
+		public FirestoreMappingContext firestoreMappingContext(FirestoreClassMapper mapper) {
+			return new FirestoreMappingContext(mapper);
 		}
 
 		@Bean
 		@ConditionalOnMissingBean
-		public FirestoreTemplate firestoreTemplate(FirestoreGrpc.FirestoreStub firestoreStub) {
-			return new FirestoreTemplate(firestoreStub, GcpFirestoreAutoConfiguration.this.firestoreRootPath);
+		public FirestoreClassMapper firestoreClassMapper(FirestoreOptions firestoreOptions, Firestore firestore) {
+			return new FirestoreClassMapper(firestoreOptions, (FirestoreRpc) firestoreOptions.getRpc());
```
meltsufin: Can/should the `FirestoreRpc` itself be a bean? When you call `options.getRpc()` does it actually create resources or only returns an existing object?

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreClassMapper.java
```diff
@@ -14,48 +14,52 @@
...
+import com.google.cloud.firestore.spi.v1.FirestoreRpc;
 import com.google.firestore.v1.Document;
 import com.google.firestore.v1.Value;
 
 import org.springframework.cloud.gcp.core.util.MapBuilder;
 
 /**
  *
- * Temporary class to expose package-private methods, will be removed in the future.
+ * TBD.
```
meltsufin: @dmitry-s Please let advise how the javadoc should be updated.

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreTemplate.java
```diff
@@ -76,7 +75,9 @@
 
 	private final String databasePath;
 
-	private final FirestoreMappingContext mappingContext = new FirestoreMappingContext();
+	private final FirestoreClassMapper classMapper;
+
+	private final FirestoreMappingContext mappingContext;
```
meltsufin: It seems odd that you would need both `classMapper` and `mappingContext`. I think only one should be required.

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreTemplate.java
```diff
@@ -88,11 +89,14 @@
...
 	 *     projects/{project_id}/databases/{database_id}/documents/chatrooms/{chatroom_id}
+	 * @param mapper converter between Firestore document snapshots and Java objects
 	 */
-	public FirestoreTemplate(FirestoreStub firestore, String parent) {
+	public FirestoreTemplate(FirestoreStub firestore, String parent, FirestoreClassMapper mapper) {
 		this.firestore = firestore;
 		this.parent = parent;
 		this.databasePath = Util.extractDatabasePath(parent);
+		this.classMapper = mapper;
+		this.mappingContext = new FirestoreMappingContext(mapper);
```
meltsufin: Calling `new` in the constructor is a code-smell because now you cannot configure your own version of the `mappingContext`.
The constructor should probably be changed to accept a `FirestoreMappingContext` bean instead of the mapper.

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/mapping/FirestoreMappingContext.java
```diff
@@ -35,6 +36,12 @@
 		AbstractMappingContext<FirestorePersistentEntity<?>, FirestorePersistentProperty>
 		implements ApplicationContextAware {
 
+	private FirestoreClassMapper firestoreClassMapper;
```
meltsufin: It doesn't seem like this class is doing anything with the `firestoreClassMapper` aside from just holding it. So, why should it be here?

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/repository/query/PartTreeFirestoreQuery.java
```diff
@@ -78,6 +80,7 @@ public PartTreeFirestoreQuery(FirestoreQueryMethod queryMethod, FirestoreReactiv
 		ReturnedType returnedType = queryMethod.getResultProcessor().getReturnedType();
 		this.tree = new PartTree(queryMethod.getName(), returnedType.getDomainType());
 		this.persistentEntity = mappingContext.getPersistentEntity(returnedType.getDomainType());
+		this.classMapper = mappingContext.getClassMapper();
```
meltsufin: Again, maybe just accept a `classMapper` in the constructor directly rather than going through this `mappingContext`.

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreTemplate.java
elefeint: I thought about only having a mapping context, but it seemed awkward to add boilerplate of `this.mappingContext.getClassMapper().doWhatever` everywhere.
I have no problem with doing that, though.

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/mapping/FirestoreMappingContext.java
elefeint: there is a usage in PartTreeFirestoreQuery.

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/repository/query/PartTreeFirestoreQuery.java
elefeint: Keeping them separate is fine.

PR #2024, opened 2019-11-08
nit: should delete subscription first

anguillanneuf:
It's good practice to delete the subscription(s) before deleting the topic(s) that they belong to.


meltsufin:
@elefeint I wonder how the integration test even works since Pub/Sub won't allow you to delete a topic that has subscriptions.

elefeint:
It does; I've even deleted topics and left orphan subscriptions through GCP console UI. Maybe behavior changed?

meltsufin:
@elefeint Should we merge or close?

elefeint:
Close, as order of REST endpoints does not matter. I'll address the integration test separately.

PR #2029, opened 2019-11-12
keep sink app from exiting

### LINKED ISSUE #2015: Pub/Sub newly daemonized thread pools shut down too well
After #2010, the Pub/Sub subscriber thread pool no longer keeps an application alive. This is good, expected behavior for simple `PubSubTemplate` users, but it results in very unintuitive behavior for Spring Cloud Stream sinks and (haven't tested it, but pretty sure) Spring Integration apps with service activators. Basically, if a Spring application is not a web application, it starts up, looks around, finds nothing interesting to do, and shuts down.

@artembilan What is expected behavior of a non-web Spring Integration application that subscribes to a queue? Intuitively, it seems that such an application should stay up waiting for data to come in.

One possible solution is to provide a daemon thread pool by default, but to override it with a non-daemon thread pool if both Pub/Sub and Spring Integration are detected on the classpath.

@artembilan @meltsufin WDYT?

anguillanneuf:
Closes #2015 

Discussed with @elefeint offline -


anguillanneuf:
@elefeint This PR is failing coverage test. Any suggestions as to what to do there?

PR #2050, opened 2019-11-20
Create SECURITY.md

ryanjbaxter:


SECURITY.md
```diff
@@ -0,0 +1,6 @@
+
+# Security Policy
+
+## Reporting a Vulnerability
+
+To report security vulnerabilities, please go to https://pivotal.io/security.
```
meltsufin: What if it's a security vulnerability in a Google product that the code integrates with?

ryanjbaxter: Then why would they be reporting the issue to us and not Google?

meltsufin: Because this document is telling them? :smile: 
Maybe we should be more clear here to say that this is about reporting security vulnerabilities in the Spring Cloud? I don't know. Maybe I'm just overthinking this.

meltsufin:
obsolete

PR #2053, opened 2019-11-20
populate key fields when virtual key property is set

### LINKED ISSUE #2044: Spanner REST Repo PUT operation not supported
I think this is actually similar to the Datastore issue we encountered a few months ago: 
https://github.com/spring-cloud/spring-cloud-gcp/issues/1773

The source is this SO issue : https://stackoverflow.com/questions/58815607/setting-the-primary-key-directly-via-the-key-id-property-is-not-supported-span

In there the user is doing a PUT request to write a new entity to Spanner, but because the new ID is contained as a string in the URL, it's not converted properly to the Spanner Key.java type. So this needs to be supported.

dmitry-s:
fixes #2044

### DIFF
--- a/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/mapping/SpannerPersistentEntity.java
+++ b/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/mapping/SpannerPersistentEntity.java
@@ -71,9 +71,9 @@ public interface SpannerPersistentEntity<T> extends
 	SpannerMappingContext getSpannerMappingContext();
 
 	/**
-	 * Gets a SpannerEntityWriter that can be used to create Cloud Spanner keys from one or
+	 * Gets a SpannerEntityProcessor that can be used to create Cloud Spanner keys from one or
 	 * more persistent properties.
-	 * @return a SpannerEntityWriter.
+	 * @return a SpannerEntityProcessor.
 	 */
 	SpannerEntityWriter getSpannerEntityProcessor();
 
--- a/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/mapping/SpannerPersistentEntity.java
+++ b/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/mapping/SpannerPersistentEntity.java
@@ -75,7 +75,7 @@ public interface SpannerPersistentEntity<T> extends
 	 * more persistent properties.
 	 * @return a SpannerEntityWriter.
 	 */
-	SpannerEntityWriter getSpannerEntityWriter();
+	SpannerEntityWriter getSpannerEntityProcessor();
 
 	@Override
 	SpannerCompositeKeyProperty getIdProperty();
--- a/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/mapping/SpannerPersistentEntityImpl.java
+++ b/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/mapping/SpannerPersistentEntityImpl.java
@@ -387,7 +387,7 @@ public class SpannerPersistentEntityImpl<T>
 					SpannerPersistentProperty[] primaryKeyProperties = owner.getPrimaryKeyProperties();
 
 					Key keyValue = (Key) value;
-					if (keyValue != null && keyValue.size() != primaryKeyProperties.length) {
+					if (keyValue == null || keyValue.size() != primaryKeyProperties.length) {
 						throw new SpannerDataException(
 								"The number of key parts is not equal to the number of primary key properties");
 					}
--- a/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/mapping/SpannerPersistentEntityImpl.java
+++ b/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/mapping/SpannerPersistentEntityImpl.java
@@ -20,6 +20,7 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -30,8 +31,8 @@ import com.google.cloud.spanner.Key;
 import org.springframework.beans.BeansException;
 import org.springframework.cloud.gcp.data.spanner.core.convert.ConversionUtils;
 import org.springframework.cloud.gcp.data.spanner.core.convert.ConverterAwareMappingSpannerEntityProcessor;
+import org.springframework.cloud.gcp.data.spanner.core.convert.SpannerEntityProcessor;
 import org.springframework.cloud.gcp.data.spanner.core.convert.SpannerEntityWriter;
-import org.springframework.cloud.gcp.data.spanner.core.convert.SpannerReadConverter;
 import org.springframework.context.ApplicationContext;
 import org.springframework.context.expression.BeanFactoryAccessor;
 import org.springframework.context.expression.BeanFactoryResolver;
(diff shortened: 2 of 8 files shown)


spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/mapping/SpannerPersistentEntityImpl.java
```diff
@@ -379,9 +385,22 @@ public PersistentPropertyAccessor getPropertyAccessor(Object object) {
...
-					throw new SpannerDataException(
-							"Setting the primary key directly via the Key ID property is not supported. "
-									+ "Please set the underlying column properties.");
+					SpannerPersistentEntity owner = (SpannerPersistentEntity) property.getOwner();
+					SpannerPersistentProperty[] primaryKeyProperties = owner.getPrimaryKeyProperties();
+
+					Key keyValue = (Key) value;
+					ArrayList<Object> keyParts = new ArrayList<>();
+					keyValue.getParts().forEach(keyParts::add);
+					if (keyParts.size() != primaryKeyProperties.length) {
```
elefeint: Is it possible to check property names, too, not just the number?

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/mapping/SpannerPersistentEntityImpl.java
```diff
@@ -86,6 +89,9 @@
 
 	private String tableName;
 
+	private SpannerReadConverter readConverter = new SpannerReadConverter();
```
ChengyuanZhao: Could we pipe in the bean for this? It is kind of a pain, but I think necessary since some of these key properties might be types using custom converters that the user set up. 

I took care to pass in the spannerEntityWrtier in the constructor for this reason too.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/mapping/SpannerPersistentEntityImpl.java
dmitry-s: I don't think we can. Key parts don't have names.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/mapping/SpannerPersistentEntityImpl.java
meltsufin: Just auto-wire the `SpannerEntityProcessor` an call `getReadConverter` on it.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/mapping/SpannerPersistentEntityImpl.java
```diff
@@ -379,9 +385,22 @@ public PersistentPropertyAccessor getPropertyAccessor(Object object) {
...
 				if (property.isIdProperty()) {
-					throw new SpannerDataException(
-							"Setting the primary key directly via the Key ID property is not supported. "
-									+ "Please set the underlying column properties.");
+					SpannerPersistentEntity owner = (SpannerPersistentEntity) property.getOwner();
+					SpannerPersistentProperty[] primaryKeyProperties = owner.getPrimaryKeyProperties();
+
+					Key keyValue = (Key) value;
+					ArrayList<Object> keyParts = new ArrayList<>();
+					keyValue.getParts().forEach(keyParts::add);
```
meltsufin: `keyValue.size()` can be used instead.

spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/core/mapping/SpannerPersistentEntityImplTests.java
```diff
@@ -461,4 +477,16 @@ public void testParentChildPkNamesMismatch() {
...
 	}
+
+	private static class MultiIdsEntity {
+		@PrimaryKey(keyOrder = 1)
+		String id;
+
+		@PrimaryKey(keyOrder = 2)
+		Long id2;
+
+
```
meltsufin: Extra blank line here.

spring-cloud-gcp-samples/spring-cloud-gcp-data-spanner-sample/src/main/java/com/example/TraderRepository.java
```diff
@@ -26,6 +28,6 @@
  * @author Chengyuan Zhao
  */
 @RepositoryRestResource(collectionResourceRel = "traders", path = "traders")
-public interface TraderRepository extends SpannerRepository<Trader, String> {
+public interface TraderRepository extends SpannerRepository<Trader, Key> {
```
meltsufin: Please put a TODO here that it needs to be changed back to `String`.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/mapping/SpannerPersistentEntityImpl.java
ChengyuanZhao: We can't use autowire because this shouldn't depend on Spring Boot and they made a big point about that. We've worked hard to keep it that way too

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/mapping/SpannerPersistentEntity.java
```diff
@@ -75,7 +75,7 @@
 	 * more persistent properties.
 	 * @return a SpannerEntityWriter.
```
meltsufin: Please fix the javadoc.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/mapping/SpannerPersistentEntityImpl.java
```diff
@@ -100,20 +104,20 @@ public SpannerPersistentEntityImpl(TypeInformation<T> information) {
 	 * @param information type information about the underlying entity type.
 	 * @param spannerMappingContext a mapping context that can be used to create persistent
 	 *     entities from properties of this entity
-	 * @param spannerEntityWriter an entity writer used to create keys by converting and
+	 * @param spannerEntityProcessor an entity writer used to create keys by converting and
```
meltsufin: the description seems off now

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/mapping/SpannerPersistentEntityImpl.java
```diff
@@ -100,20 +104,20 @@ public SpannerPersistentEntityImpl(TypeInformation<T> information) {
...
 	 */
 	public SpannerPersistentEntityImpl(TypeInformation<T> information,
-			SpannerMappingContext spannerMappingContext, SpannerEntityWriter spannerEntityWriter) {
+			SpannerMappingContext spannerMappingContext, SpannerEntityProcessor spannerEntityProcessor) {
 		super(information);
 
 		Assert.notNull(spannerMappingContext,
 				"A non-null SpannerMappingContext is required.");
-		Assert.notNull(spannerEntityWriter, "A non-null SpannerEntityWriter is required.");
+		Assert.notNull(spannerEntityProcessor, "A non-null SpannerEntityWriter is required.");
```
meltsufin:
```suggestion
		Assert.notNull(spannerEntityProcessor, "A non-null SpannerEntityProcessor is required.");
```

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/mapping/SpannerPersistentEntity.java
dmitry-s:
```suggestion
	 * @return a SpannerEntityProcessor.
```

gopalgowda:
Hi @dmitry-s, @meltsufin ,
what is the version number I should be using the get this change ? I used 1.1.4 RELEASE which didn't work.
Thanks for your support!

PR #2060, opened 2019-11-26
make FirestoreClassMapper a Bean

### LINKED ISSUE #2055: Make PublicClassMapper a bean and have non-static methods

dmitry-s:
fixes #2055

### DIFF
--- a/spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreTemplate.java
+++ b/spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreTemplate.java
@@ -37,7 +37,6 @@ import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
 import reactor.util.context.Context;
 
-import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.cloud.gcp.data.firestore.mapping.FirestoreClassMapper;
 import org.springframework.cloud.gcp.data.firestore.mapping.FirestoreMappingContext;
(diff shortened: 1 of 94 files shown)


spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreTemplate.java
```diff
@@ -57,6 +58,8 @@
  * @since 1.2
  */
 public class FirestoreTemplate implements FirestoreReactiveOperations {
+	@Autowired
```
elefeint: Since FirestoreTemplate is autoconfigured, we should probably stick to adding fields through constructor or setter from in `GcpFirestoreAutoConfiguration`. This feels optional (although what happens if it's not set, an NPE? Might want to provide a default), so using `setClassMapper()` makes sense.

spring-cloud-gcp-data-firestore/src/test/java/org/springframework/cloud/gcp/data/firestore/repository/query/PartTreeFirestoreQueryTests.java
```diff
@@ -175,6 +170,6 @@ private PartTreeFirestoreQuery setUpPartTreeFirestoreQuery(String methodName) {
 		when(this.queryMethod.getResultProcessor()).thenReturn(resultProcessor);
 
 		return new PartTreeFirestoreQuery(this.queryMethod,
-				this.firestoreTemplate, new FirestoreMappingContext());
+				this.firestoreTemplate, new FirestoreMappingContext(), new FirestoreDefaultClassMapper());
```
elefeint: Reuse this.classMapper?

PR #2062, opened 2019-11-27
Relax ImportOrder checkstyle for sample apps

### LINKED ISSUE #2028: disable checkstyle for sample apps
Working with someone on setting up a sample app, and import order is tripping us up. Sample apps don't need to follow the strict spring style.

elefeint:
We have two different usecases when someone is trying to run a sample:
* people who want to try the sample apps, but don't intend to contribute any changes to the codebase. For them, checkstyle is just a distraction from what they are trying to get done -- learn how to use our integrations.
* actual contributors, who should be bound by checkstyle because our codebase should be consistent.

What this change does now is:
* guarantee that incorrect style does not make it into the codebase (because the new profile runs on Travis at install phase).
* make it possible to run full checkstyle locally (so the contributor does not get surprised by new checkstyle requirements after PR is created).
* for casual users, default to not enforcing checkstyle on sample apps .

Fixes #2028.


spring-cloud-gcp-samples/checkstyle-samples-suppressions.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0"?>
+<!DOCTYPE suppressions PUBLIC
+    "-//Checkstyle//DTD SuppressionFilter Configuration 1.2//EN"
+    "https://checkstyle.org/dtds/suppressions_1_2.dtd">
+<suppressions>
+  <suppress files=".*" checks="LineLength" />
+  <suppress files=".*\.java" checks="HideUtilityClassConstructorCheck" />
+
+  <!-- Sample apps should be usable by someone without correct import order IDE settings. -->
+  <suppress files=".*\.java" checks="ImportOrder"/>
```
meltsufin: I think the committers should still be bound by strict style checks, even in the samples.
If the concern is for folks checking out the code and playing with the samples, and getting tripped by checkstyle, we should move checkstyle to a Maven profile, and only enable it on Travis. This will ensure that casual users of the samples are not tripped by it, but the style is still enforced on commits.

CONTRIBUTING.adoc
```diff
@@ -14,6 +14,8 @@ Run `./mvnw clean test` in the root directory of the project to run the tests.
 The `./mvnw` is a self-contained https://maven.apache.org/[Maven] wrapper that allows you to build the project without having Maven installed on your local machine.
 +
 You can run the tests of a specific module by using the `-f` flag like this: `./mvnw clean test -f spring-cloud-gcp-pubsub`
++
+If you are contributing to a sample application, please use `full-checkstyle` maven profile to validate code style locally: `./mvnw clean test -P full-checkstyle`.
```
artembilan: I'm not sure what is the difference then: we still require Checkstyle rules for samples and request more work to do when contribute a new one or modify existing.
So, why just don't *ignore* `ImportOrder` for samples like you did for `HideUtilityClassConstructorCheck`?

Thanks

elefeint: We have two different usecases when someone is trying to run a sample:
* people who want to try the sample apps, but don't intend to contribute any changes to the codebase. For them, checkstyle is just a distraction from what they are trying to get done -- learn how to use our integrations.
* actual contributors, who should be bound by checkstyle because our codebase should be consistent.

What this change does now is:
* guarantee that incorrect style does not make it into the codebase (because the new profile runs on Travis at install phase).
* make it possible to run full checkstyle locally (so the contributor does not get surprised by new checkstyle requirements after PR is created).
* for casual users, default to not enforcing checkstyle on sample apps .

Issue #11, opened 2017-05-24
Generate base documentation for website


viniciusccarvalho:
Get the basics for our plugin to generate the website

PR #2065, opened 2019-12-02
fix timestamp conversion

### LINKED ISSUE #2064: Unable to convert java.sql.Timestamp to com.google.cloud.Timestamp in Spring Data Cloud Spanner
https://stackoverflow.com/questions/59136904/unable-to-convert-java-sql-timestamp-to-com-google-cloud-timestamp-in-spring-dat

dmitry-s:
fixes #2064 

We are looking for the first compatible type in the order in which they appear. String is compatible with Timestamp so it was selected. More specific types should appear first!

### DIFF
--- a/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/ConverterAwareMappingSpannerEntityWriter.java
+++ b/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/ConverterAwareMappingSpannerEntityWriter.java
@@ -91,11 +91,11 @@ public class ConverterAwareMappingSpannerEntityWriter implements SpannerEntityWr
 		map.put(long.class, (BiFunction<ValueBinder, Long, ?>) ValueBinder::to);
 		map.put(Double.class, (BiFunction<ValueBinder, Double, ?>) ValueBinder::to);
 		map.put(double.class, (BiFunction<ValueBinder, Double, ?>) ValueBinder::to);
-		map.put(String.class, (BiFunction<ValueBinder, String, ?>) ValueBinder::to);
 		map.put(Timestamp.class,
 				(BiFunction<ValueBinder, Timestamp, ?>) ValueBinder::to);
 		map.put(ByteArray.class,
 				(BiFunction<ValueBinder, ByteArray, ?>) ValueBinder::to);
+		map.put(String.class, (BiFunction<ValueBinder, String, ?>) ValueBinder::to);
 		map.put(double[].class,
 				(BiFunction<ValueBinder, double[], ?>) ValueBinder::toFloat64Array);
 		map.put(boolean[].class,
--- a/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/ConverterAwareMappingSpannerEntityWriter.java
+++ b/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/ConverterAwareMappingSpannerEntityWriter.java
@@ -21,7 +21,6 @@ import java.util.Collections;
 import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.Map;
-import java.util.Optional;
 import java.util.Set;
 import java.util.function.BiConsumer;
 import java.util.function.BiFunction;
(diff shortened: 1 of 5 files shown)


meltsufin:
Does it fix both of these issues?
https://stackoverflow.com/questions/59136904/unable-to-convert-java-sql-timestamp-to-com-google-cloud-timestamp-in-spring-dat

https://stackoverflow.com/questions/59071133/timestamp-converter-not-working-in-spring-data-rest-with-spanner

dmitry-s:
Yes, it should fix both of them.

ayancancode:
The issue is when we are iterating over complex target type, we first encounter com.google.cloud.Date and try to check if source type java.sql.Timestamp can be converted in com.google.cloud.Date type or not. While checking we check for entire source type hierarchy and as java.util.Date is parent of java.sql.Timestamp and we check if java.util.Date and com.google.cloud.Date are compatible or not, we return true for original source type java.sql.Timestamp as if java.sql.Timestamp is compatible with com.google.cloud.Date. Even though that is true, in database column type is TIMESTAMP which does not allow com.google.cloud.Date to pass through and throw the following exception.

Caused by: java.util.concurrent.ExecutionException: com.google.api.gax.rpc.FailedPreconditionException: io.grpc.StatusRuntimeException: FAILED_PRECONDITION: Invalid value for column CREATED_ON in table TEST: Expected TIMESTAMP.

ayancancode:
I tried this fix, it did not work in my project. Only if I put Timestamp converter before Date, it is working but it will fail in other scenarios.

meltsufin:
@ayancancode Thanks for testing the PR. Can you try now?

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/ConverterAwareMappingSpannerEntityWriter.java
```diff
@@ -71,12 +70,12 @@
 	@SuppressWarnings("unchecked")
 	private static Map<Class<?>, BiConsumer<ValueBinder<?>, Iterable>> createIterableTypeMapping() {
 		Map<Class<?>, BiConsumer<ValueBinder<?>, Iterable>> map = new LinkedHashMap<>();
-		map.put(com.google.cloud.Date.class, ValueBinder::toDateArray);
 		map.put(Boolean.class, ValueBinder::toBoolArray);
 		map.put(Long.class, ValueBinder::toInt64Array);
-		map.put(String.class, ValueBinder::toStringArray);
 		map.put(Double.class, ValueBinder::toFloat64Array);
 		map.put(Timestamp.class, ValueBinder::toTimestampArray);
+		map.put(com.google.cloud.Date.class, ValueBinder::toDateArray);
```
meltsufin: `Date` doesn't need to be fully qualified. It's imported.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/ConverterAwareMappingSpannerEntityWriter.java
```diff
@@ -71,12 +70,12 @@
...
+		map.put(String.class, ValueBinder::toStringArray);
```
meltsufin: Can you put `String` after `ByteArray` to be consistent with the below.

spring-cloud-gcp-samples/spring-cloud-gcp-data-spanner-sample/src/test/java/com/example/SpannerRepositoryTests.java
```diff
@@ -83,7 +86,7 @@ public static void checkToRun() {
 		assumeThat(
 				"Spanner integration tests are disabled. "
 						+ "Please use '-Dit.spanner=true' to enable them. ",
-				System.getProperty("it.spanner"), is("true"));
+				"true", is("true"));
```
meltsufin: oops

ayancancode:
> @ayancancode Thanks for testing the PR. Can you try now?
I will definitely try this and let you know. I raised a similar PR for this and tried testing your PR without sample code. I got failure in Unit Testing. Are you sure you are passing the unit test case #11 in KeyConversionTests?

ayancancode:
> @ayancancode Thanks for testing the PR. Can you try now?

I will definitely try this and let you know. I raised a similar PR for this and tried testing your PR without sample code. I got failure in Unit Testing. Are you sure you are passing the unit test case #11 in KeyConversionTests?

PR #2075, opened 2019-12-10
Fix acking behavior of PubSub stream binder

### LINKED ISSUE #2074: AckMode.AUTO_ACK acks the message even if the processing exception occurs
From the doc:
The framework acks the PubsubMessage after it is sent to the channel and processed successfully. The framework does not immediately nack the message when the exception occurs, and allows the eventual redelivery to take effect.

The actual behavior:
Processing exception occurs and the message is still acked.

You can see in code that messageNacked = true is only set for AckMode.AUTO and that causes this bug.
`boolean messageNacked = false;

		try {
			sendMessage(getMessageBuilderFactory()
					.withPayload(message.getPayload())
					.copyHeaders(messageHeaders)
					.build());
		}
		catch (RuntimeException re) {
				message.nack();
				messageNacked = true;
			}
			throw new PubSubException("Sending Spring message failed.", re);
		}
		finally {
			if (!messageNacked && (this.ackMode == AckMode.AUTO || this.ackMode == AckMode.AUTO_ACK)) {
				message.ack();
			}
		}`

dzou:
Fixes a regression in the Pub/Sub stream channel adapter AUTO_ACK mode in which messages were being ACKed even though an exception was thrown.

Contributes to #2074. 

cc/ @radivojen

### DIFF
--- a/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapter.java
+++ b/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapter.java
@@ -159,13 +159,7 @@ public class PubSubInboundChannelAdapter extends MessageProducerSupport {
 			if (this.ackMode == AckMode.AUTO) {
 				message.nack();
 			}
-
-			if (this.ackMode == AckMode.AUTO_ACK) {
-				LOGGER.warn("Sending Spring message failed.", re);
-			}
-			else {
-				throw new PubSubException("Sending Spring message failed.", re);
-			}
+			throw new PubSubException("Sending Spring message failed.", re);
 		}
 	}
 
--- a/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapter.java
+++ b/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapter.java
@@ -145,24 +145,26 @@ public class PubSubInboundChannelAdapter extends MessageProducerSupport {
 			});
 		}
 
-		boolean messageNacked = false;
-
 		try {
 			sendMessage(getMessageBuilderFactory()
 					.withPayload(message.getPayload())
 					.copyHeaders(messageHeaders)
 					.build());
+
+			if (this.ackMode == AckMode.AUTO_ACK || this.ackMode == AckMode.AUTO) {
+				message.ack();
+			}
 		}
 		catch (RuntimeException re) {
 			if (this.ackMode == AckMode.AUTO) {
 				message.nack();
-				messageNacked = true;
 			}
-			throw new PubSubException("Sending Spring message failed.", re);
-		}
-		finally {
-			if (!messageNacked && (this.ackMode == AckMode.AUTO || this.ackMode == AckMode.AUTO_ACK)) {
-				message.ack();
+
+			if (this.ackMode == AckMode.AUTO_ACK) {
+				LOGGER.warn("Sending Spring message failed.", re);
+			}
+			else {
+				throw new PubSubException("Sending Spring message failed.", re);
 			}
 		}
 	}
--- a/spring-cloud-gcp-pubsub/src/test/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapterTests.java
+++ b/spring-cloud-gcp-pubsub/src/test/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapterTests.java
@@ -123,8 +123,13 @@ public class PubSubInboundChannelAdapterTests {
 		when(this.mockMessageChannel.send(any())).thenThrow(new RuntimeException(EXCEPTION_MESSAGE));
 
 		this.adapter.setAckMode(AckMode.AUTO_ACK);
-		this.adapter.start();
 
+		assertThatThrownBy(() -> {
+			this.adapter.start();
+		}).hasMessageContaining(EXCEPTION_MESSAGE);
+
+		// When exception thrown, verify that neither ack() nor nack() is called.
+		verify(mockAcknowledgeableMessage, times(0)).ack();
 		verify(mockAcknowledgeableMessage, times(0)).nack();
 	}
 
--- a/spring-cloud-gcp-pubsub/src/test/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapterTests.java
+++ b/spring-cloud-gcp-pubsub/src/test/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapterTests.java
@@ -123,10 +123,7 @@ public class PubSubInboundChannelAdapterTests {
 		when(this.mockMessageChannel.send(any())).thenThrow(new RuntimeException(EXCEPTION_MESSAGE));
 
 		this.adapter.setAckMode(AckMode.AUTO_ACK);
-
-		assertThatThrownBy(() -> {
-			this.adapter.start();
-		}).hasMessageContaining(EXCEPTION_MESSAGE);
+		this.adapter.start();
 
 		verify(mockAcknowledgeableMessage, times(0)).nack();
 	}


spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapter.java
```diff
@@ -145,24 +145,26 @@ public void nack() {
...
-		boolean messageNacked = false;
-
 		try {
 			sendMessage(getMessageBuilderFactory()
 					.withPayload(message.getPayload())
 					.copyHeaders(messageHeaders)
 					.build());
+
+			if (this.ackMode == AckMode.AUTO_ACK || this.ackMode == AckMode.AUTO) {
+				message.ack();
```
meltsufin: Why should we move `ack()` outside of the finally?
Can we just fix the bug that was introduced in the last PR without further changes?

dzou: Well the `finally` actually contributes to the mistaken logic of the bug.

This code correctly implements what is specified in the documentation:

> Automatic acking (AckMode.AUTO)
> 
> A message is acked with GCP Pub/Sub if the adapter sent it to the channel and no exceptions were thrown.
> 
> If a RuntimeException is thrown while the message is processed, then the message is handled in the following way:
> 
> If an error channel is configured for the adapter (pubSubInboundChannelAdapater.getErrorChannel() != null) then the message will be acked and forwarded with the error to the error channel.
> 
> If no error channel is configured for the adapter, then the message is nacked.
> 
> Automatic acking OK (AckMode.AUTO_ACK)
> 
> A message is acked with GCP Pub/Sub if the adapter sent it to the channel and no exceptions were thrown. If a RuntimeException is thrown while the message is processed, then the message is neither acked / nor nacked.
> 
> This is useful when using the subscription’s ack deadline timeout as a retry delivery backoff mechanism.

https://github.com/spring-cloud/spring-cloud-gcp/blob/8e4fe4dcf56b51f6eba3cc8085dc5e9877fe7055/docs/src/main/asciidoc/spring-integration-pubsub.adoc

spring-cloud-gcp-pubsub/src/test/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapterTests.java
```diff
@@ -123,10 +123,7 @@ public void testAckModeAutoAck_nacksWhenDownstreamProcessingFails() {
 		when(this.mockMessageChannel.send(any())).thenThrow(new RuntimeException(EXCEPTION_MESSAGE));
 
 		this.adapter.setAckMode(AckMode.AUTO_ACK);
-
-		assertThatThrownBy(() -> {
-			this.adapter.start();
-		}).hasMessageContaining(EXCEPTION_MESSAGE);
+		this.adapter.start();
```
elefeint: Can you change the name of this method to reflect intended behavior? 

But I am also confused why the exception is no longer thrown. Code under test rethrows in AUTO_ACK mode.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapter.java
elefeint: I like the new way; it's cleaner and easier to analyze. Java `finally` behavior, especially where exceptions are concerned, can be tricky.

spring-cloud-gcp-pubsub/src/test/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapterTests.java
```diff
@@ -123,10 +123,7 @@ public void testAckModeAutoAck_nacksWhenDownstreamProcessingFails() {
...
 
 		verify(mockAcknowledgeableMessage, times(0)).nack();
```
elefeint: Also verify that it does not ack. Maybe even verifyZeroInteractions.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapter.java
```diff
@@ -145,24 +145,26 @@ public void nack() {
...
-				messageNacked = true;
 			}
-			throw new PubSubException("Sending Spring message failed.", re);
-		}
-		finally {
-			if (!messageNacked && (this.ackMode == AckMode.AUTO || this.ackMode == AckMode.AUTO_ACK)) {
-				message.ack();
+
+			if (this.ackMode == AckMode.AUTO_ACK) {
+				LOGGER.warn("Sending Spring message failed.", re);
```
elefeint: AUTO_ACK behavior needs to be the same as MANUAL when an exception is thrown -- rethrow, so the user can handle it.

meltsufin: +1, why would we not want to rethrow the exception for `AUTO_ACK`?

dzou: I went off the request made by @radivojen in the original issue that we should consider not throwing here -

> throwing PubSubException causes immediate redelivery by the google pubsub client

I'm not really sure what the Spring cloud stream specs call for here.

meltsufin: How is throwing the exception causing immediate redelivery? Pub/Sub will generally redeliver only after the  ack-deadline expires. However, it's possible for a duplicate message to be delivered sooner because Pub/Sub only guarantees at-least-once delivery.

radivojen: If the exception is propagated it has to be handled by the Google pubsub sdk.
Since the subscribe message flow is googlesdk -> spring-gcp
Here is the log I receive, handled by google sdk as WARN level:

> Radivojes-MacBook-Pro.local 2019-12-12 16:17:12,427 reqId= userAgent= WARN  com.google.cloud.pubsub.v1.MessageDispatcher ->  [gcp-pubsub-subscriber1] - MessageReceiver failed to processes ack ID: K0MLKlgRTgQhIT4wPkVTRFAGFixdRkhRNxkIaFEOT14jPzUgKEUSIysdFmNdNBsNaFxcdAEEW0gjK2cnbgkZAQhFe059ZBwNalRcdAJVCB53dWh2aHLOlrPUiv31agExUJTfjNhkLdDQ1cJEZh09WBJLLD5-LSxFQV5AEkwmGURJUw, the message will be nacked.

radivojen: And here is the exact method from the MessageDispatcher:
`@Override
    public void onFailure(Throwable t) {
      logger.log(
          Level.WARNING,
          "MessageReceiver failed to processes ack ID: " + ackId + ", the message will be nacked.",
          t);
      pendingNacks.add(ackId);
      forget();
    }`

On the publisher level it is of course fine to propagate the exception.
I hope I helped. Please test it yourselves.

meltsufin: You're right. Essentially, it ends up behaving just like the `AUTO` mode, which is not the intent.

spring-cloud-gcp-pubsub/src/test/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapterTests.java
dzou: Done.

spring-cloud-gcp-pubsub/src/test/java/org/springframework/cloud/gcp/pubsub/integration/inbound/PubSubInboundChannelAdapterTests.java
dzou: I reverted this part of the test since we decided we will implement this fix for the clear regression in this PR and then decide on how exception should be thrown in the next PR.

dzou:
@radivojen @elefeint @meltsufin - So we discussed offline more about the fix. We decided that we will first fix the regression in which AUTO_ACK was acking the messages even when an exception was thrown in this PR.

The second issue about how the exception should be thrown will be resolved in the next PR. We just want to see why we had this behavior in there for so long and ensure that we will not break anyone by switching to a log WARN.

dzou:
Sounds good. @elefeint, PTAL and let me know if you agree with our current strategem and I will merge this in.

PR #2078, opened 2019-12-12
Detailed docs on acking/nacking in Spring Integration

elefeint:
Corrects documentation about what happens if processing fails.
Adds more detail on acking/nacking and error handling.
Replaces straight Pub/Sub module with starter, which is more useful as it provides autoconfigured templates.

Follow up to #2075 and #2074 .

### DIFF
--- a/docs/src/main/asciidoc/spring-integration-pubsub.adoc
+++ b/docs/src/main/asciidoc/spring-integration-pubsub.adoc
@@ -71,6 +71,7 @@ Then, we declare a `PubSubInboundChannelAdapter` bean.
 It requires the channel we just created and a `SubscriberFactory`, which creates `Subscriber` objects from the Google Cloud Java Client for Pub/Sub.
 The Spring Boot starter for GCP Pub/Sub provides a configured `PubSubSubscriberOperations` object.
 
+===== Acknowledging messages and handling failures
 The `PubSubInboundChannelAdapter` supports three acknowledgement modes, with `AckMode.AUTO` being the default value:
 
 * Automatic acking and nacking (`AckMode.AUTO`)
@@ -79,7 +80,7 @@ The `PubSubInboundChannelAdapter` supports three acknowledgement modes, with `Ac
 
 ===== Acking
 
-In `AUTO` and `AUTO_ACK` modes, a message acknowledgement is automatically sent to Cloud Pub/Sub if the message was successfully sent into the Spring Integration channel and processed with no exceptions.
+In `AUTO` and `AUTO_ACK` modes, a message acknowledgement is automatically sent to Cloud Pub/Sub if the message was successfully processed with no exceptions.
 
 In `MANUAL` mode, your application logic needs to manually acknowledge each message.
 If a message is not acknowledged within the applicable `ackDeadline`, Cloud Pub/Sub will attempt redelivery.
--- a/docs/src/main/asciidoc/spring-integration-pubsub.adoc
+++ b/docs/src/main/asciidoc/spring-integration-pubsub.adoc
@@ -71,28 +71,21 @@ Then, we declare a `PubSubInboundChannelAdapter` bean.
 It requires the channel we just created and a `SubscriberFactory`, which creates `Subscriber` objects from the Google Cloud Java Client for Pub/Sub.
 The Spring Boot starter for GCP Pub/Sub provides a configured `PubSubSubscriberOperations` object.
 
-The `PubSubInboundChannelAdapter` supports three acknowledgement modes, with `AckMode.AUTO` being the default value;
+The `PubSubInboundChannelAdapter` supports three acknowledgement modes, with `AckMode.AUTO` being the default value:
 
-Automatic acking (`AckMode.AUTO`)
+* Automatic acking and nacking (`AckMode.AUTO`)
+* Automatic acking only (`AckMode.AUTO_ACK`)
+* Manual acking and nacking (`AckMode.MANUAL`)
 
-A message is acked with GCP Pub/Sub if the adapter sent it to the channel and no exceptions were thrown.
+===== Acking
 
-If a `RuntimeException` is thrown while the message is processed, then the message is handled in the following way:
+In `AUTO` and `AUTO_ACK` modes, a message acknowledgement is automatically sent to Cloud Pub/Sub if the message was successfully sent into the Spring Integration channel and processed with no exceptions.
 
-* If an error channel is configured for the adapter (`pubSubInboundChannelAdapater.getErrorChannel() != null`) then the message will be acked and forwarded with the error to the error channel.
-* If no error channel is configured for the adapter, then the message is nacked.
-
-Automatic acking OK (`AckMode.AUTO_ACK`)
-
-A message is acked with GCP Pub/Sub if the adapter sent it to the channel and no exceptions were thrown.
-If a `RuntimeException` is thrown while the message is processed, then the message is neither acked / nor nacked.
-
-This is useful when using the subscription's ack deadline timeout as a retry delivery backoff mechanism.
-
-Manually acking (`AckMode.MANUAL`)
+In `MANUAL` mode, your application logic needs to manually acknowledge each message.
+If a message is not acknowledged within the applicable `ackDeadline`, Cloud Pub/Sub will attempt redelivery.
 
 The adapter attaches a `BasicAcknowledgeablePubsubMessage` object to the `Message` headers.
-Users can extract the `BasicAcknowledgeablePubsubMessage` using the `GcpPubSubHeaders.ORIGINAL_MESSAGE` key and use it to (n)ack a message.
+Users can extract the `BasicAcknowledgeablePubsubMessage` using the `GcpPubSubHeaders.ORIGINAL_MESSAGE` key and use it to ack (or nack) a message.
 
 [source,java]
 ----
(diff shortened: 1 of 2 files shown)


docs/src/main/asciidoc/spring-integration-pubsub.adoc
```diff
@@ -71,28 +71,21 @@ Then, we declare a `PubSubInboundChannelAdapter` bean.
 It requires the channel we just created and a `SubscriberFactory`, which creates `Subscriber` objects from the Google Cloud Java Client for Pub/Sub.
 The Spring Boot starter for GCP Pub/Sub provides a configured `SubscriberFactory`.
 
-The `PubSubInboundChannelAdapter` supports three acknowledgement modes, with `AckMode.AUTO` being the default value;
+The `PubSubInboundChannelAdapter` supports three acknowledgement modes, with `AckMode.AUTO` being the default value:
```
meltsufin: Maybe put a header above this sentence.
Something like: "Acknowledging messages and handling failures"

docs/src/main/asciidoc/spring-integration-pubsub.adoc
```diff
@@ -71,28 +71,21 @@ Then, we declare a `PubSubInboundChannelAdapter` bean.
...
-Automatic acking (`AckMode.AUTO`)
+* Automatic acking and nacking (`AckMode.AUTO`)
+* Automatic acking only (`AckMode.AUTO_ACK`)
+* Manual acking and nacking (`AckMode.MANUAL`)
 
-A message is acked with GCP Pub/Sub if the adapter sent it to the channel and no exceptions were thrown.
+===== Acking
 
-If a `RuntimeException` is thrown while the message is processed, then the message is handled in the following way:
+In `AUTO` and `AUTO_ACK` modes, a message acknowledgement is automatically sent to Cloud Pub/Sub if the message was successfully sent into the Spring Integration channel and processed with no exceptions.
```
meltsufin:
```suggestion
In `AUTO` and `AUTO_ACK` modes, a message acknowledgement is automatically sent to Cloud Pub/Sub if the message was successfully processed with no exceptions.
```

docs/src/main/asciidoc/spring-integration-pubsub.adoc
```diff
@@ -71,28 +71,21 @@ Then, we declare a `PubSubInboundChannelAdapter` bean.
...
-Automatic acking OK (`AckMode.AUTO_ACK`)
-
-A message is acked with GCP Pub/Sub if the adapter sent it to the channel and no exceptions were thrown.
-If a `RuntimeException` is thrown while the message is processed, then the message is neither acked / nor nacked.
-
-This is useful when using the subscription's ack deadline timeout as a retry delivery backoff mechanism.
-
-Manually acking (`AckMode.MANUAL`)
+In `MANUAL` mode, your application logic needs to manually acknowledge each message.
+If a message is not acknowledged within the applicable `ackDeadline`, Cloud Pub/Sub will attempt redelivery.
```
meltsufin: Would it be possible to clarify what is this `ackDeadline`?

elefeint:
This is staying until the new year. Since Dan's fix, the original docs are mostly correct. 
The only discrepancy is that in AUTO_ACK and MANUAL modes, when an exception bubbles up, the message will be nacked in client library.
This will become correct if/when we stop rethrowing the exception, so that will be moot.
The error handling docs are useful, but I can reformat them back into the original sections... in 2020.

meltsufin:
Can we return to this PR now?

docs/src/main/asciidoc/spring-integration-pubsub.adoc
```diff
@@ -71,29 +71,32 @@ Then, we declare a `PubSubInboundChannelAdapter` bean.
...
-If a `RuntimeException` is thrown while the message is processed, then the message is neither acked / nor nacked.
-
-This is useful when using the subscription's ack deadline timeout as a retry delivery backoff mechanism.
-
-Manually acking (`AckMode.MANUAL`)
-
+===== Acknowledging messages and handling failures
+When working with Cloud Pub/Sub, it is important to understand the concept of `ackDeadline` -- the amount of time Cloud Pub/Sub will wait until attempting redelivery of an outstanding message.
+While acknowledging (acking) a message removes it from Pub/Sub's known outstanding messages, nacking a message simply means resetting its acknowledgement deadline to 0, forcing immediate redelivery.
+A global `ackDeadline` exists on each subscription. Additionally, the Cloud Pub/Sub client library can extend each streamed message's `ackDeadline` up to `spring.cloud.gcp.pubsub.subscriber.max-ack-extension-period` seconds (in `spring-cloud-gcp`, this is set to 0, which prevents automatic deadline extension, but in a standalone client library, the default is 60 minutes).
```
meltsufin, line 77: Do we need to say why decided to change the default?

docs/src/main/asciidoc/spring-integration-pubsub.adoc
```diff
@@ -71,29 +71,32 @@ Then, we declare a `PubSubInboundChannelAdapter` bean.
...
+
+The `PubSubInboundChannelAdapter` supports three acknowledgement modes, with `AckMode.AUTO` being the default value:
+
+* Automatic acking and nacking (`AckMode.AUTO`) +
+If the message was successfully processed with no exceptions, an acknowledgement is automatically sent to Cloud Pub/Sub.
+In case of processing failure, a nack (request for immediate redelivery) is automatically sent to Cloud Pub/Sub.
```
meltsufin, line 83:
```suggestion
In case of a processing failure, a nack (request for immediate redelivery) is automatically sent to Cloud Pub/Sub.
```

docs/src/main/asciidoc/spring-integration-pubsub.adoc
```diff
@@ -71,29 +71,32 @@ Then, we declare a `PubSubInboundChannelAdapter` bean.
...
+
+* Automatic acking only (`AckMode.AUTO_ACK`) +
+If the message was successfully processed with no exceptions, an acknowledgement is automatically sent to Cloud Pub/Sub.
+In case of processing failure, the adapter neither acks nor nacks the message, resulting in the same behavior as `MANUAL` mode.
```
meltsufin, line 87:
```suggestion
In case of a processing failure, the adapter neither acks nor nacks the message, resulting in the same behavior as `MANUAL` mode, where the the application logic can decide to nack manually or allow for eventual redelivery.
```

docs/src/main/asciidoc/spring-integration-pubsub.adoc
```diff
@@ -71,29 +71,32 @@ Then, we declare a `PubSubInboundChannelAdapter` bean.
...
+If a message is not acknowledged within the applicable `ackDeadline`, Cloud Pub/Sub will attempt redelivery.
+
+* Manual acking and nacking (`AckMode.MANUAL`) +
+Your application logic needs to manually acknowledge each message.
+If a message is not acknowledged within the applicable `ackDeadline`, Cloud Pub/Sub will attempt redelivery.
++
+NOTE: The Pub/Sub client library this adapter is based on will attempt to extend the acknowledgement deadline for an hour by default.
```
meltsufin, line 94:
```suggestion
NOTE: The Pub/Sub client library that this adapter is based on will attempt to extend the acknowledgement deadline for an hour by default.
```
But we override the default, right?

docs/src/main/asciidoc/spring-integration-pubsub.adoc
```diff
@@ -108,6 +111,74 @@ public MessageHandler messageReceiver() {
 }
 ----
 
+====== Nacking
+In the context of Cloud Pub/Sub, nacking simply means instructing the message to be redelivered immediately.
+This could be useful in a load balanced architecture, where one of the subscribers is having issues but others are available to process messages.
+
+In `AUTO` mode, a message is nacked if processing causes any unchecked exceptions.
```
meltsufin, line 118:
```suggestion
In `AUTO` mode, a message is nacked if processing causes any unhandled exceptions.
```

docs/src/main/asciidoc/spring-integration-pubsub.adoc
```diff
@@ -108,6 +111,74 @@ public MessageHandler messageReceiver() {
...
+
+In `AUTO_ACK` and `MANUAL` modes, behavior depends on whether an error channel was configured on the `PubSubInboundChannelAdapter`:
+
+* If an error channel is present (see <<#Error-Handling>> for how to configure one), the message is neither acked nor nacked.
```
meltsufin, line 122: Can all of these complicated scenarios be captured in a table where on one axis is Mode and another is Error/No-Error/Handled/etc...? And the value is whether it's nacked/acked/neither.

docs/src/main/asciidoc/spring-integration-pubsub.adoc
elefeint: Do you know why we changed the default in #742?

docs/src/main/asciidoc/spring-integration-pubsub.adoc
elefeint: yes; I wrote this and forgot to fix that.

docs/src/main/asciidoc/spring-integration-pubsub.adoc
meltsufin: If I had to guess, it was probably because we didn't agree with the default. :)
In any case, we have to live with that decision now and own this new default.

PR #2085, opened 2019-12-17
Explicitly add dns scheme to grpc channel addresses

ST-DDT:
Improves compatibility with libraries that provide additional grpc `NameResolver`s

See also: https://github.com/grpc/grpc-java/pull/6499

yidongnan/grpc-spring-boot-starter#268
yidongnan/grpc-spring-boot-starter#304


ST-DDT:
> I think you missed two other places in `GcpPubSubEmulatorAutoConfiguration` and `GcpFirestoreAutoConfiguration`.

Those parts are configured via properties so I didn't include them in the first run.
(I hardcoded them at the place they are used, so it might no longer be possible to configure a different scheme via the config, but I don't know whether this was possible in the first place)

> In tests I also see us as configuring the `ManagedChannel` using `forAddress` instead of `forTarget`.
> Which is more appropriate?

`forAddress` is fine for tests and static addresses, but it won't get updated and IIRC doesn't support load balancing.
`forTarget` also supports dynamic addresses such as DNS and also supports resolution of multiple matches and thus allows client side load balancing.

I changed the test to match the other locations.

ST-DDT:
> We might also want to apply this to the `1.2.x` branch.

(Not sure what this implicates)
Should I create a second PR which adds it to that branch
or will you do that yourself and that is just an information for me?

meltsufin:
@ST-DDT I can take care of the merge into `1.2.x`.

PR #2095, opened 2019-12-20
stream binder functional sample: mention that explicit functional bean configuration is optional

elefeint:
Follow up to #2013.

Explicit functional bean configuration is no longer required (since SCS will autodiscover when there is only one functional bean available).
Unfortunately, the integration test has both sink and source on its classpath, so it autodiscovers both beans, making explicit definition _in test_ necessary.

### DIFF
--- a/spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-stream-binder-functional-sample/spring-cloud-gcp-pubsub-stream-binder-functional-sample-sink/src/main/resources/application.properties
+++ b/spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-stream-binder-functional-sample/spring-cloud-gcp-pubsub-stream-binder-functional-sample-sink/src/main/resources/application.properties
@@ -5,7 +5,8 @@
 spring.cloud.stream.function.bindings.logUserMessage-in-0=input
 spring.cloud.stream.bindings.input.destination=my-topic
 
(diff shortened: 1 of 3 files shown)


spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-stream-binder-functional-sample/spring-cloud-gcp-pubsub-stream-binder-functional-sample-sink/src/main/resources/application.properties
```diff
@@ -5,7 +5,8 @@
 spring.cloud.stream.function.bindings.logUserMessage-in-0=input
 spring.cloud.stream.bindings.input.destination=my-topic
 
-spring.cloud.function.definition=logUserMessage
+# Optional, as Spring Cloud Stream will autodiscover the correct functional bean.
+#spring.cloud.function.definition=logUserMessage
```
dmitry-s: maybe we should just add comments saying this property is optional, but keep it configured?

elefeint: Yeah, I think I'll do that.

PR #2099, opened 2019-12-26
Add overloaded `createBlob()` method accepting initial file contents.

### LINKED ISSUE #2097: Storage integration should provide ability to provide initial file contents
To fit into Spring Resource idiom, `GoogleStorageResource` [creates a new file](https://github.com/spring-cloud/spring-cloud-gcp/blob/72d746486fedd19b16ead3cdbbcfb48c3e90c3a9/spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResource.java#L306) before returning the output stream to a now-existing file.

But GCP Storage treats the act of overwriting a file as requiring both a _delete_ and a _create_ permission, so if the service account only has the _create_ permission, it actually cannot use the output stream provided by `GoogleStorageResource`, failing with `[account] does not have OBJECTS_DELETE access to [bucket/file]` error.

We should provide ability to create and write a resource in one go, using the overloaded `Storage.create(BlobInfo, byte[])` method.

elefeint:
Fixes #2097.

### DIFF
--- a/spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResource.java
+++ b/spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResource.java
@@ -195,6 +195,7 @@ public class GoogleStorageResource implements WritableResource {
 	/**
 	 * Creates the blob that this {@link GoogleStorageResource} represents in Google Cloud
 	 * Storage and fills it with provided content.
+	 * @param contents the initial file contents to write
 	 * @return the created blob object
 	 * @throws StorageException if any errors during blob creation arise,
 	 * such as if the blob already exists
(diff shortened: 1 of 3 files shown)


elefeint:
I updated getOutputStream() to write to a blob without pre-creating it. 
At the same time, I think the new `write(byte[])` method is a convenient shortcut for writing small amounts of content, so I've kept that.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResource.java
```diff
@@ -295,19 +310,15 @@ public OutputStream getOutputStream() throws IOException {
...
-				if (!this.autoCreateFiles) {
-					throw new FileNotFoundException("The blob was not found: " + getURI());
-				}
-
-				blob = createBlob();
-			}
+		Blob blob = getBlob();
 
-			return Channels.newOutputStream(blob.writer());
+		if ((blob == null || !blob.exists()) && !this.autoCreateFiles) {
```
meltsufin: Isn't this the same as `!isWritable()`?

elefeint: Yes, although exists() also checks whether it's a bucket, and we know it's not by now.
Also, blob.exists() is not even necessary, since it's just a `getBlob() != null` check, but I am going to keep it for now for consistency with old code. We can simplify the code in the new year.

Issue #2094, opened 2019-12-20
Spring Cloud version for 1.2.x branch seems wrong


elefeint:
The Spring Cloud Build parent version [on 1.2.x branch](https://github.com/spring-cloud/spring-cloud-gcp/blob/b90faa3527d79f4c138532cc81e97e198199bf61/pom.xml#L9) is the same as on [master](https://github.com/spring-cloud/spring-cloud-gcp/blob/6c029b249a567f39063cb45fe947a6381b99eb6a/pom.xml#L9) -- `2.2.1.BUILD-SNAPSHOT`.

It should probably be Greenwich `spring-cloud-build:2.1.10.BUILD-SNAPSHOT` on the 1.2.x branch.

meltsufin:
Spring Cloud Build 2.2.x corresponds to the Hoxton release AFAIK, which is aligned with our 1.2.x release. We should probably just upgrade to 2.2.2 snapshots.

elefeint:
[spring-cloud-release wiki](https://github.com/spring-cloud/spring-cloud-release/wiki/Supported-Versions) says that Hoxton will support both Spring Boot 2.2.x and 2.3.x. So I spring-cloud is not doing a version bump...

Here is my current understanding:

|spring-cloud-gcp|spring cloud release train|spring-cloud-build|spring-boot
|---|---|---|---|
|1.2.x|Hoxton|2.2.1|2.2.x and 2.3.x
|1.3.x|Ilford|2.2.2.BUILD-SNAPSHOT|2.4.x

So the reason I was confused is that `spring-cloud-build` in `Hoxton` release train has nowhere to grow. I would have expected the current snapshot to have been updated to `2.3.0.BUILD-SNAPSHOT`.

But anyway, our versioning seems reasonable.

For future reference: https://spring.io/blog/2019/12/23/spring-cloud-roadmap-and-hoxton-and-greenwich-maintenance-and-eol-announcements

PR #2100, opened 2019-12-27
Bump version after Hoxton.SR1 release

meltsufin:
Fixes: #2094.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -6,12 +6,12 @@
 	<parent>
 		<groupId>org.springframework.cloud</groupId>
 		<artifactId>spring-cloud-build</artifactId>
-		<version>2.2.1.BUILD-SNAPSHOT</version>
+		<version>2.2.2.BUILD-SNAPSHOT</version>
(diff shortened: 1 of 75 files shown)


pom.xml
```diff
@@ -47,9 +47,9 @@
 	<properties>
 		<maven.compiler.source>1.8</maven.compiler.source>
 		<maven.compiler.target>1.8</maven.compiler.target>
-		<spring-cloud-commons.version>2.2.1.BUILD-SNAPSHOT</spring-cloud-commons.version>
+		<spring-cloud-commons.version>2.2.2.BUILD-SNAPSHOT</spring-cloud-commons.version>
```
elefeint: Is there a Spring Cloud BOM we can use for all spring-cloud dependencies? Bus, Sleuth, Stream, Commons...

meltsufin: Since we're included in Spring Cloud BOM, we can't use Spring Cloud BOM ourselves.

PR #2103, opened 2019-12-30
Allows the ACK_MODE to be configurable when using Stream Binders

### LINKED ISSUE #2079: Pub/Sub Spring Cloud Stream binder should support ackMode customization
**Is your feature request related to a problem? Please describe.**
Currently stream binder creates a `PubSubInboundChannelAdapter` in default AUTO acknowlegement mode. As discussed in #2074, this does not allow a message to expire naturally -- a message will be acked in normal execution or in exception scenario where an error handler is declared, and a message will be nacked if an unhandled error occurs.

**Describe the solution you'd like**
Provide ability to override the `PubSubInboundChannelAdapter` `ackMode` property.

**Describe alternatives you've considered**
Another option discussed in original issue was to avoid raising an exception in AUTO_ACK and MANUAL modes. However, this would reduce the error handling flexibility, as it would eat the exception and provide no ability to recover its cause.

dzou:
This allows the user to configure the ACK_MODE when using the Pub/Sub Spring Cloud Stream binder.

Fixes #2079

### DIFF
--- a/docs/src/main/asciidoc/spring-stream.adoc
+++ b/docs/src/main/asciidoc/spring-stream.adoc
@@ -55,6 +55,18 @@ spring.cloud.stream.gcp.pubsub.bindings.events.producer.auto-create-resources=tr
 
 ==== Consumer Destination Configuration
 
+A `PubSubInboundChannelAdapter` will be configured for your consumer endpoint.
+You may adjust the `ACK_MODE` of the consumer endpoint which controls how messages will be acknowledged when they are successfully received through the `ack-mode` property.
+The three possible options are: `AUTO` (default), `AUTO_ACK`, and `MANUAL`.
+These options are described in detail in the <<inbound-channel-adapter-using-pubsub-streaming-pull, Pub/Sub channel adapter documentation>>.
+
+.application.properties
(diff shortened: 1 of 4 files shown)


docs/src/main/asciidoc/spring-stream.adoc
```diff
@@ -55,6 +55,18 @@ spring.cloud.stream.gcp.pubsub.bindings.events.producer.auto-create-resources=tr
 
 ==== Consumer Destination Configuration
 
+A `PubSubInboundChannelAdapter` will be configured for your consumer endpoint.
+You may adjust the `ACK_MODE` of the consumer endpoint which controls how messages will be acknowledged when they are successfully received through the `ack-mode` property.
```
meltsufin:
```suggestion
You may adjust the ack mode of the consumer endpoint using the `ack-mode` property.
The ack mode controls how messages will be acknowledged when they are successfully received.
```

dzou:
> Oh! One thing -- bring "Copyright 2017-2018" into the roaring twenties.

Lol, omg we're in the roaring twenties.

PR #2106, opened 2020-01-03
Add overloaded `createBlob()` method accepting initial file contents.…

### LINKED ISSUE #2102: Backport Storage write-only fix
Backport #2099 into the 1.2.x branch.

elefeint:
Backporting the fix to 1.2.x branch.

* Add overloaded `createBlob()` method accepting initial file contents.
* changed getOutputStream() to not pre-create the file

Follow-up to #2097.
Fixes #2102.

### DIFF
--- a/spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResource.java
+++ b/spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResource.java
@@ -48,6 +48,7 @@ import org.springframework.util.Assert;
  * @author Chengyuan Zhao
  * @author João André Martins
  * @author Daniel Zou
+ * @author Elena Felder
  */
 public class GoogleStorageResource implements WritableResource {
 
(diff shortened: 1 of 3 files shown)


spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/GoogleStorageResource.java
```diff
@@ -191,6 +192,20 @@ public Blob createBlob() throws StorageException {
...
 
+	/**
+	 * Creates the blob that this {@link GoogleStorageResource} represents in Google Cloud
+	 * Storage and fills it with provided content.
+	 * @param contents the initial file contents to write
+	 * @return the created blob object
+	 * @throws StorageException if any errors during blob creation arise,
+	 * such as if the blob already exists
+	 * @throws IllegalStateException if the resource reference is to a bucket, and not a blob.
+	 * @since 1.3
```
meltsufin: Doesn't seem right if it's in the 1.2 branch.
Remove this method?

elefeint: Hmm. It's useful and does not introduce any breaking changes.
Perhaps I should remove `@since` in the master instead?

meltsufin: AFAIK, only bug-fixes should be introduced in service releases.

elefeint: @artembilan Happy New Year! Are Spring Cloud's service releases expected to contain only bugfixes, or are backwards-compatible new features okay?

This PR is backporting a 1.3.x change that really contains two changes addressing the same issue:
1) A bugfix: making resource writing a one-step operation when an output stream is acquired, as the previous behavior of creating the file and then acquiring the stream failed unless the user had a delete permission (because replacement is considered a removal and subsequent addition in GCP storage).
2) A new convenience feature allowing the object to be created with content in a single call, without having to acquire an output stream first.

So, we definitely need to backport (1), but are not sure about (2).

artembilan: We are OK with new feature in the services release as long as they don't break anything and existing application can upgrade without noticing new stuff.

Do not back port feature is just a choice with less pain during maintenance: such a new feature has to be fixed now in several versions 😄 (if there are going to be bugs of course...)

PR #2108, opened 2020-01-08
support lazy loading for one-to-one references

### LINKED ISSUE #2083: DatastoreTemplate#convertEntitiesForRead very slow (17 second delay)
**Version:** 1.2.0.RC2
I am making a call with the DatastoreTemplate that freezes for ~17 seconds every call.

EDIT: Removed the code, the issue is that I have a lot of recursive references in some of my table definitions. What I would like to know is if there is a way to tell Spring not to do a deep search for every reference on an entity. I haven't been able to find one and projection still seems to take a long time to execute (though I need to dig into this a little more).

Any help is greatly appreciated.

dmitry-s:
fixes #2083

### DIFF
--- a/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
+++ b/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
@@ -462,18 +462,17 @@ public class DatastoreTemplate implements DatastoreOperations, ApplicationEventP
 				return;
 			}
 			Value<?> value;
-			if (persistentProperty.isCollectionLike() || persistentProperty.getType() == Optional.class) {
-				if (LazyUtil.isLazyAndNotLoaded(val)) {
-					value = ListValue.of(LazyUtil.getKeys(val));
-				}
-				else {
-					Iterable<?> iterableVal = (Iterable<?>) ValueUtil.toListIfArray(val);
-					entitiesToSave.addAll(getEntitiesForSave(iterableVal, persistedEntities));
-					List<KeyValue> keyValues = StreamSupport.stream((iterableVal).spliterator(), false)
-							.map((o) -> KeyValue.of(this.getKey(o, false)))
-							.collect(Collectors.toList());
-					value = ListValue.of(keyValues);
-				}
+			if (LazyUtil.isLazyAndNotLoaded(val)) {
+				value = LazyUtil.getKeys(val);
+			}
+			else if (persistentProperty.isCollectionLike()) {
+				Iterable<?> iterableVal = (Iterable<?>) ValueUtil.toListIfArray(val);
+				entitiesToSave.addAll(getEntitiesForSave(iterableVal, persistedEntities));
+				List<KeyValue> keyValues = StreamSupport.stream((iterableVal).spliterator(), false)
+						.map((o) -> KeyValue.of(this.getKey(o, false)))
+						.collect(Collectors.toList());
+				value = ListValue.of(keyValues);
+
 			}
 			else {
 				entitiesToSave.addAll(getEntitiesForSave(Collections.singletonList(val), persistedEntities));
@@ -615,15 +614,14 @@ public class DatastoreTemplate implements DatastoreOperations, ApplicationEventP
 	private <T> T computeReferencedField(BaseEntity entity, ReadContext context,
 			DatastorePersistentProperty referenceProperty, String fieldName, Class<T> type) {
 		T referenced;
-		if (referenceProperty.isLazyLoaded() && referenceProperty.isCollectionLike()) {
-			List<Value<Key>> keyList = entity.getList(fieldName);
+		if (referenceProperty.isLazyLoaded()) {
 			DatastoreReaderWriter originalTx = getDatastoreReadWriter();
-			referenced = LazyUtil.wrapSimpleLazyProxy((List<Value<Key>> storedKeys) -> {
+			referenced = LazyUtil.wrapSimpleLazyProxy(() -> {
 				if (getDatastoreReadWriter() != originalTx) {
 					throw new DatastoreDataException("Lazy load should be invoked within the same transaction");
 				}
-				return (T) fetchReferenced(referenceProperty, context, valuesToKeys(storedKeys));
-			}, type, keyList);
+				return (T) findReferenced(entity, referenceProperty, context);
+			}, type, entity.getValue(fieldName));
 		}
 		else {
 			referenced = (T) findReferenced(entity, referenceProperty, context);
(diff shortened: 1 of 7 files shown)


spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
```diff
@@ -462,18 +462,17 @@ public Key createKey(Class aClass, Object id) {
...
-				}
-				else {
-					Iterable<?> iterableVal = (Iterable<?>) ValueUtil.toListIfArray(val);
-					entitiesToSave.addAll(getEntitiesForSave(iterableVal, persistedEntities));
-					List<KeyValue> keyValues = StreamSupport.stream((iterableVal).spliterator(), false)
-							.map((o) -> KeyValue.of(this.getKey(o, false)))
-							.collect(Collectors.toList());
-					value = ListValue.of(keyValues);
-				}
+			if (LazyUtil.isLazyAndNotLoaded(val)) {
```
meltsufin: I thought this was done in the previous PR, but I guess we missed it somehow.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/LazyUtil.java
```diff
@@ -37,12 +41,29 @@
...
 				new SimpleLazyDynamicInvocationHandler<T>(supplierFunc, keys));
+		}
+		Factory factory = (Factory) objenesis.newInstance(getEnhancedTypeFor(type));
+		factory.setCallbacks(new Callback[] { new SimpleLazyDynamicInvocationHandler<T>(supplierFunc, keys) });
+
+		return (T) factory;
+	}
+
+	private static Class<?> getEnhancedTypeFor(Class<?> type) {
+
```
meltsufin:
```suggestion
```

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/LazyUtil.java
```diff
@@ -79,23 +99,30 @@ private static SimpleLazyDynamicInvocationHandler getProxy(Object object) {
 			return (SimpleLazyDynamicInvocationHandler) Proxy
 					.getInvocationHandler(object);
 		}
+		else if (object instanceof Factory) {
+			Callback[] callbacks = ((Factory) object).getCallbacks();
+			if (callbacks != null && callbacks.length == 1
+					&& callbacks[0] instanceof SimpleLazyDynamicInvocationHandler) {
+				return (SimpleLazyDynamicInvocationHandler) callbacks[0];
+			}
```
elefeint: Should we log a warning if callbacks length is > 1? Is this a "this should never happen" scenario?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/LazyUtil.java
```diff
@@ -63,7 +83,7 @@ public static boolean isLazyAndNotLoaded(Object object) {
 	 * @param object a proxy object
 	 * @return list of keys if the object is a proxy, null otherwise
 	 */
```
elefeint: Is there anything we could say about the difference between treatment of collections, interfaces and objects in the javadoc?

spring-cloud-gcp-data-datastore/src/test/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplateTests.java
```diff
@@ -538,10 +537,20 @@ public void saveTest() {
...
-				(keys) -> Collections.singletonList(this.childEntity7), List.class,
-				Collections.singletonList(KeyValue.of(this.childKey7)));
+				() -> Collections.singletonList(this.childEntity7), List.class,
+				ListValue.of(KeyValue.of(this.childKey7)));
+		saveTestCommon(this.ob1, true);
+	}
+
+	@Test
+	public void saveTestNotInterfaceLazy() {
+		ArrayList<ChildEntity> arrayList = new ArrayList();
```
elefeint: Is this going to also go through the collection path (similar to saveTestCollectionLazy)?

spring-cloud-gcp-data-datastore/src/test/java/org/springframework/cloud/gcp/data/datastore/it/DatastoreIntegrationTests.java
```diff
@@ -536,18 +538,43 @@ public void referenceTest() {
...
+	public void lazyReferenceTest() throws InterruptedException {
+		LazyEntity lazyParentEntity = new LazyEntity(new LazyEntity(new LazyEntity()));
+		this.datastoreTemplate.save(lazyParentEntity);
+
+		LazyEntity loadedParent = this.datastoreTemplate.findById(lazyParentEntity.id, LazyEntity.class);
+
+		//Saving an entity with not loaded lazy field
+		this.datastoreTemplate.save(loadedParent);
+
+		loadedParent = this.datastoreTemplate.findById(loadedParent.id, LazyEntity.class);
```
elefeint: does this have the same problem that we had in previous PR where the equality does not necessarily prove anything?

spring-cloud-gcp-data-datastore/src/test/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplateTests.java
dmitry-s: no, ArrayList is not an interface, so it will be proxied using cglib

spring-cloud-gcp-data-datastore/src/test/java/org/springframework/cloud/gcp/data/datastore/it/DatastoreIntegrationTests.java
dmitry-s: it did prove everything we needed actually :) I only changed it for readability. 
But in this case I'm testing multi-level lazy loading.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/LazyUtil.java
dmitry-s: sure

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/LazyUtil.java
dmitry-s: This should never happen if it's our proxy.

PR #2116, opened 2020-01-13
GcsStreamingMessageSource with comparator

lukasGemela:
fixing https://github.com/spring-cloud/spring-cloud-gcp/issues/2117

### DIFF
--- a/spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/integration/inbound/GcsStreamingMessageSource.java
+++ b/spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/integration/inbound/GcsStreamingMessageSource.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017-2018 the original author or authors.
+ * Copyright 2017-2020 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -17,6 +17,7 @@
 package org.springframework.cloud.gcp.storage.integration.inbound;
 
 import java.util.Collection;
+import java.util.Comparator;
 import java.util.List;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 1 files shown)


dzou:
Hi @lukasGemela , thanks for the contribution, this certainly looks reasonable to me. Will check-in with another member of our team just to confirm.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/integration/inbound/GcsStreamingMessageSource.java
```diff
@@ -35,14 +36,19 @@
...
 public class GcsStreamingMessageSource extends AbstractRemoteFileStreamingMessageSource<BlobInfo> {
 
-	public GcsStreamingMessageSource(RemoteFileTemplate<BlobInfo> template) {
-		super(template, null);
+	public GcsStreamingMessageSource(RemoteFileTemplate<BlobInfo> template, Comparator<BlobInfo> comparator) {
+		super(template, comparator);
 		doSetFilter(new GcsPersistentAcceptOnceFileListFilter(new SimpleMetadataStore(), "gcsStreamingMessageSource"));
 	}
 
+	public GcsStreamingMessageSource(RemoteFileTemplate<BlobInfo> template) {
```
artembilan: For smooth code flow this ctor has to be first. Because exactly this one calls another.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/integration/inbound/GcsStreamingMessageSource.java
```diff
@@ -35,14 +36,19 @@
  * @author João André Martins
  * @author Mike Eltsufin
  * @author Chengyuan Zhao
+ * @author Lukas Gemela
  */
 public class GcsStreamingMessageSource extends AbstractRemoteFileStreamingMessageSource<BlobInfo> {
 
-	public GcsStreamingMessageSource(RemoteFileTemplate<BlobInfo> template) {
-		super(template, null);
+	public GcsStreamingMessageSource(RemoteFileTemplate<BlobInfo> template, Comparator<BlobInfo> comparator) {
```
artembilan: Since this ctor is new public API in the class it would be great to have some JavaDoc on it with an appropriate `@since`.

dzou:
> Looks like good catch though!
> 
> I wonder if we should have some simple unit test to cover this new API.
> Plus some mentioning in the Docs.
> 
> Thank you for contribution so far!

@artembilan @lukasGemela , I can take care of updating the docs and adding a unit test.

spring-cloud-gcp-storage/src/main/java/org/springframework/cloud/gcp/storage/integration/inbound/GcsStreamingMessageSource.java
lukasGemela: updated accordingly

dzou:
Cool, shall we merge this PR, i'll add a followup shortly for the doc + test changes?

meltsufin:
@dzou Merged, yes please add the doc + test, and then we'll want to cherry-pick both commits to 1.2.x.

Issue #2127, opened 2020-01-18
The method SimpleSpannerRepository#existsById fetches all fields and interleaved child objects


s13o:
The method SimpleSpannerRepository#existsById could be implemented by a simple query like `SELECT EXISTS (SELECT 1 FROM %s WHERE %s)` instead of fetching of too many fields and interleaved objects from the database.

dmitry-s:
@s13o I'm working on this issue

s13o:
I'm too. I did as You recommend
pls take a look on https://github.com/spring-cloud/spring-cloud-gcp/pull/2135

dmitry-s:
@s13o I decided to take a different approach. I'm using a read call, as it was done originally; but instead of retrieving all the fields, I only fetch one.
I understand that you put a lot of work in #2135, but I thought you were not going to work on that PR when you closed it.

Going forward we will coordinate our efforts!

s13o:
well, I did a force-push for a batter commit history and it closed the PR
Basically I'd planned to resolve the #2127 and then #2132 because I thought they depend on each other in a part of the building of sql-where for KeySet.
So, will see after the merge - probably I'll be able to suggest some improvements

PR #2128, opened 2020-01-18
The method SimpleSpannerRepository#existsById fetches all fields and interleaved child objects

s13o:
pls see #2127

### DIFF
--- a/docs/src/main/asciidoc/spring-integration-pubsub.adoc
+++ b/docs/src/main/asciidoc/spring-integration-pubsub.adoc
@@ -53,9 +53,9 @@ public MessageChannel pubsubInputChannel() {
 @Bean
 public PubSubInboundChannelAdapter messageChannelAdapter(
     @Qualifier("pubsubInputChannel") MessageChannel inputChannel,
-    SubscriberFactory subscriberFactory) {
+    PubSubSubscriberOperations subscriberOperations) {
     PubSubInboundChannelAdapter adapter =
-        new PubSubInboundChannelAdapter(subscriberFactory, "subscriptionName");
+        new PubSubInboundChannelAdapter(subscriberOperations, "subscriptionName");
     adapter.setOutputChannel(inputChannel);
     adapter.setAckMode(AckMode.MANUAL);
 


spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/support/SimpleSpannerRepository.java
```diff
@@ -105,7 +113,39 @@ public SpannerOperations getSpannerTemplate() {
...
+			final SpannerPersistentEntity<?> persistentEntity = this.spannerTemplate.getMappingContext()
+					.getPersistentEntity(this.entityType);
+
+			int tagNum = 0;
+			List<String> columns = new ArrayList<>();
+			List<String> tags = new ArrayList<>();
+			List<Object> keyParts = new ArrayList<>();
+			List<SpannerPersistentProperty> properties = persistentEntity.getFlattenedPrimaryKeyProperties();
+			Iterator parentKeyParts = k.getParts().iterator();
+			while (parentKeyParts.hasNext()) {
```
dmitry-s: This logic overlaps with `SpannerStatementQueryExecutor.getChildrenRowsQuery` significantly. Please refactor to avoid duplication. 
Also, I think the bulk of this method should be moved to `SpannerTemplate`. Feel free to introduce `SpannerTemplate.existsById` method.

spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/repository/support/SpannerRepositoryImplTests.java
```diff
@@ -66,8 +82,20 @@
 	public void setup() {
 		this.template = mock(SpannerTemplate.class);
 		this.entityProcessor = mock(SpannerEntityProcessor.class);
+		this.mappingContext = mock(SpannerMappingContext.class);
```
dmitry-s: Correspondingly, after SpannerTemplate.existsById method is implemented, these tests should be moved to `SpannerTemplateTests` class.

s13o:
Today I  think the better idea is to move the "existsById" logic to the
"SpannerStatementQueryExecutor"
It is very similar to the method
"SpannerStatementQueryExecutor#getChildrenRowsQuery" (that actually should
be better named like "getChildrenRowsStatement")

I'll do another PR tomorrow I believe.

вт, 21 янв. 2020 г. в 22:35, dmitry-s <notifications@github.com>:

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/support/SimpleSpannerRepository.java
s13o: Today I think the better idea is to move the "existsById" logic to the "SpannerStatementQueryExecutor"
It is very similar to the method "SpannerStatementQueryExecutor#getChildrenRowsQuery" (that actually should be better named like "getChildrenRowsStatement")

I'll do another PR tomorrow I believe.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
```diff
@@ -204,6 +204,7 @@ public long executePartitionedDmlStatement(Statement statement) {
 		return result;
 	}
 
+	//todo:tune it
```
dmitry-s: remove and create a github issue instead

PR #2132, opened 2020-01-23
optimized eager-loading for interleaved properties

### LINKED ISSUE #2110: @Interleaved(lazy=false) should use Joins for Spanner
Spanner Template always retries an interleaved properties of an entity by an additional query per each row instead of using joins. It causes huge query performance degradation and makes "@Interleaved(lazy=false)" useless - it actually always "more or less lazy" now.

dmitry-s:
fixes #2110

### DIFF
--- a/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
+++ b/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
@@ -181,12 +181,20 @@ public class SpannerTemplate implements SpannerOperations, ApplicationEventPubli
 	@Override
 	public <T> List<T> read(Class<T> entityClass, KeySet keys,
 			SpannerReadOptions options) {
-		SpannerPersistentEntity<?> persistentEntity = this.mappingContext
+		SpannerPersistentEntity<T> persistentEntity = (SpannerPersistentEntity<T>) this.mappingContext
 				.getPersistentEntity(entityClass);
-		List<T> entities = mapToListAndResolveChildren(executeRead(persistentEntity.tableName(), keys,
-				persistentEntity.columns(), options), entityClass,
-				(options != null) ? options.getIncludeProperties() : null,
-				options != null && options.isAllowPartialRead());
+		List<T> entities;
+
+		if (options == null &&
+				persistentEntity.hasEagerlyLoadedProperties() && !keys.getRanges().iterator().hasNext()) {
+			entities = executeReadQueryAndResolveChildren(keys, persistentEntity);
+		}
+		else {
+			entities = mapToListAndResolveChildren(executeRead(persistentEntity.tableName(), keys,
+					persistentEntity.columns(), options), entityClass,
+					(options != null) ? options.getIncludeProperties() : null,
+					options != null && options.isAllowPartialRead());
+		}
 		maybeEmitEvent(new AfterReadEvent(entities, keys, options));
 		return entities;
 	}
@@ -228,7 +236,7 @@ public class SpannerTemplate implements SpannerOperations, ApplicationEventPubli
 		SpannerPersistentEntity<?> persistentEntity = this.mappingContext
 				.getPersistentEntity(entityClass);
 		String sql = "SELECT " + SpannerStatementQueryExecutor.getColumnsStringForSelect(
-				persistentEntity) + " FROM " + persistentEntity.tableName();
+				persistentEntity, this.mappingContext) + " FROM " + persistentEntity.tableName();
 		return query(entityClass,
 				SpannerStatementQueryExecutor.buildStatementFromSqlWithArgs(
 						SpannerStatementQueryExecutor.applySortingPagingQueryOptions(
@@ -445,6 +453,14 @@ public class SpannerTemplate implements SpannerOperations, ApplicationEventPubli
 		return resultSet;
 	}
 
+	private <T> List<T> executeReadQueryAndResolveChildren(KeySet keys, SpannerPersistentEntity<T> persistentEntity) {
+		Statement statement = SpannerStatementQueryExecutor.getChildrenRowsQuery(keys, persistentEntity,
+				this.spannerEntityProcessor.getWriteConverter(),
+				this.mappingContext);
+
+		return resolveChildEntities(query(persistentEntity.getType(), statement, null), null);
+	}
+
 	private ResultSet executeRead(String tableName, KeySet keys, Iterable<String> columns,
 			SpannerReadOptions options) {
 
@@ -547,6 +563,11 @@ public class SpannerTemplate implements SpannerOperations, ApplicationEventPubli
 							.contains(spannerPersistentEntity.getName())) {
 						return;
 					}
+					Object propertyValue = accessor.getProperty(spannerPersistentProperty);
+					if (propertyValue != null) {
+						resolveChildEntities((List) propertyValue, null);
+						return;
+					}
 					Class childType = spannerPersistentProperty.getColumnInnerType();
 					SpannerPersistentEntity childPersistentEntity = this.mappingContext
 							.getPersistentEntity(childType);
@@ -554,7 +575,8 @@ public class SpannerTemplate implements SpannerOperations, ApplicationEventPubli
 					Supplier<List> getChildrenEntitiesFunc = () -> queryAndResolveChildren(childType,
 							SpannerStatementQueryExecutor.getChildrenRowsQuery(
 									this.spannerSchemaUtils.getKey(entity),
-									childPersistentEntity, this.spannerEntityProcessor.getWriteConverter()),
+									childPersistentEntity, this.spannerEntityProcessor.getWriteConverter(),
+									this.mappingContext),
 							null);
 
 					accessor.setProperty(spannerPersistentProperty,
--- a/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
+++ b/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
@@ -185,8 +185,7 @@ public class SpannerTemplate implements SpannerOperations, ApplicationEventPubli
 				.getPersistentEntity(entityClass);
 		List<T> entities;
 
-		if (options == null &&
-				persistentEntity.hasEagerlyLoadedProperties() && !keys.getRanges().iterator().hasNext()) {
+		if (isEligibleForEagerFetch(keys, options, persistentEntity)) {
 			entities = executeReadQueryAndResolveChildren(keys, persistentEntity);
 		}
 		else {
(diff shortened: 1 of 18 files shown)


spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/mapping/SpannerPersistentEntityImpl.java
```diff
@@ -157,6 +159,9 @@ public void addPersistentProperty(SpannerPersistentProperty property) {
 		else if (!property.isInterleaved()) {
 			this.columnNames.add(property.getColumnName());
 		}
+		else if (property.isInterleaved() && !property.isLazyInterleaved()) {
+			this.hasEagerlyLoadedProperties = true;
+		}
```
elefeint: Leave a comment about what should happen if the property is interleaved lazily (that it's skipped and not persistent)?

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryExecutor.java
```diff
@@ -169,32 +171,63 @@ private SpannerStatementQueryExecutor() {
...
-			sj.add(keyProp.getColumnName() + " = @" + tagName);
-			tags.add(tagName);
-			keyParts.add(parentKeyParts.next());
-			tagNum++;
+
+		for (Key key : keySet.getKeys()) {
+			StringJoiner andJoiner = new StringJoiner(" AND ", "(", ")");
+			Iterator parentKeyParts = key.getParts().iterator();
+			while (parentKeyParts.hasNext()) {
+				SpannerPersistentProperty keyProp = childKeyProperties.get(tagNum % childKeyProperties.size());
```
elefeint: I don't think I understand what happens if the number of parent key parts exceeds the number of child key parts. Is that even supposed to happen?

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryExecutor.java
```diff
@@ -169,32 +171,63 @@ private SpannerStatementQueryExecutor() {
...
+				String tagName = "tag" + tagNum;
+				andJoiner.add(keyProp.getColumnName() + " = @" + tagName);
+				tags.add(tagName);
+				keyParts.add(parentKeyParts.next());
+				tagNum++;
+			}
+			orJoiner.add(andJoiner.toString());
```
elefeint: Let's check if I understand: the "and" conditions make sure the parent and child key for a single row match. And the "or" conditions match multiple rows' worth of keys?

I would take my confusion as an opportunity to factor out some logic into helper methods. For example, the while loop could go into a method named "singleRowKeyMatchCondition)". Better yet, those could go into a helper class that could be unit tested.

The goal of this would be to make it trivial for someone troubleshooting a production issue to look at the code and understand what it's doing.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryExecutor.java
```diff
@@ -169,32 +171,63 @@ private SpannerStatementQueryExecutor() {
...
+		String tableName = childPersistentEntity.tableName();
+		List<SpannerPersistentProperty> parentKeyProperties = parentPersistentEntity
+				.getFlattenedPrimaryKeyProperties();
+		String condition = parentKeyProperties.stream()
+				.map(keyProp -> tableName + "." + keyProp.getColumnName()
+						+ " = "
+						+ parentPersistentEntity.tableName() + "." + keyProp.getColumnName())
+				.collect(Collectors.joining(" AND "));
+
+		return "ARRAY (SELECT AS STRUCT " + getColumnsStringForSelect(childPersistentEntity, mappingContext) + " FROM "
```
elefeint: I wonder if it makes sense to pass string builders around as the query gets constructed instead of returning strings that have to, in turn, be concatenated. For a high-volume system, all this string concatenation will add up.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryExecutor.java
s13o: No, it could never happen

elefeint: Thanks! Then the modulo operation can be removed.

s13o:
I follow the same direction in my research but this implementation support only "reads" (spring data find methods). Queries remained not tuned I believe, is not it?

The bigger problem I suspect I've found - it seems to me this implementation does not support when some eager interleaved  contains another  and so on (Spanner supports up to 7 levels of such embedding)
Can sone one help me with poor coverage for the PR #2135 (I'm not so familiar with the GitHub tools so can't find so fast what test should be added) then I will finish my implementation and we will compare them?

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryExecutor.java
dmitry-s: It is there for a reason. When you have multiple keys in the keySet, you will get more key parts, because you have multiple parents. So wee need to repeat children parts for each parent.

dmitry-s:
Hi @s13o!
 
This implementation provides multi-level support for eager loading. There are integration tests where we load parent->eagerChildren->eagerChildren structures.

> Queries remained not tuned I believe, is not it?

could you clarify your question?

Please wait on  #2135 , I will push my PR soon and it will include the tests you can reuse.

Thanks

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryExecutor.java
elefeint: Would moving `int tagNum = 0` to within the `for` loop scope work? Then it would start at 0 for every parent key.

dmitry-s: That's not going to work, because tagNums have to be distinct. They are used for parameters binding.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
```diff
@@ -181,12 +181,20 @@ public long executePartitionedDmlStatement(Statement statement) {
...
-		SpannerPersistentEntity<?> persistentEntity = this.mappingContext
+		SpannerPersistentEntity<T> persistentEntity = (SpannerPersistentEntity<T>) this.mappingContext
 				.getPersistentEntity(entityClass);
-		List<T> entities = mapToListAndResolveChildren(executeRead(persistentEntity.tableName(), keys,
-				persistentEntity.columns(), options), entityClass,
-				(options != null) ? options.getIncludeProperties() : null,
-				options != null && options.isAllowPartialRead());
+		List<T> entities;
+
+		if (options == null &&
```
meltsufin: Comment that when options are there the new strategy is n/a.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
```diff
@@ -547,14 +563,20 @@ private void resolveChildEntity(Object entity, Set<String> includeProperties) {
 							.contains(spannerPersistentEntity.getName())) {
 						return;
 					}
+					Object propertyValue = accessor.getProperty(spannerPersistentProperty);
```
meltsufin: Cast here to `List`.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryExecutor.java
```diff
@@ -169,32 +171,63 @@ private SpannerStatementQueryExecutor() {
...
-		Iterator parentKeyParts = parentKey.getParts().iterator();
-		while (parentKeyParts.hasNext()) {
-			SpannerPersistentProperty keyProp = childKeyProperties.get(tagNum);
-			String tagName = "tag" + tagNum;
-			sj.add(keyProp.getColumnName() + " = @" + tagName);
-			tags.add(tagName);
-			keyParts.add(parentKeyParts.next());
-			tagNum++;
+
+		for (Key key : keySet.getKeys()) {
```
meltsufin: Can you make the for loop and the while loop separate private methods for readability?

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryExecutor.java
```diff
@@ -260,9 +293,19 @@ private static void bindParameter(ValueBinder<Statement.Builder> bind,
...
 	}
 
-	public static String getColumnsStringForSelect(
-			SpannerPersistentEntity spannerPersistentEntity) {
-		return String.join(" , ", spannerPersistentEntity.columns());
+	public static String getColumnsStringForSelect(SpannerPersistentEntity spannerPersistentEntity, SpannerMappingContext mappingContext) {
+		StringJoiner joiner = new StringJoiner(", ");
+		spannerPersistentEntity.doWithInterleavedProperties(persistentProperty -> {
+			SpannerPersistentProperty spannerPersistentProperty = (SpannerPersistentProperty) persistentProperty;
+			if (!spannerPersistentProperty.isLazyInterleaved()) {
```
meltsufin: `isEagerlyInterleaved`

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/mapping/SpannerPersistentProperty.java
```diff
@@ -113,4 +113,11 @@
 	 * @return {@code true} if the property is lazily-fetched. {@code false} otherwise.
 	 */
 	boolean isLazyInterleaved();
+
+	/**
+	 * Return whether this property is an eagerly-fetched interleaved property.
+	 * @return {@code true} if the property is eagerly-fetched. {@code false} otherwise.
+	 */
+	boolean isEagerlyInterleaved();
```
meltsufin: Maybe `isEagerInterleaved()` to be consistent with `isLazyInterleaved()`?

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/mapping/SpannerPersistentPropertyImpl.java
```diff
@@ -149,6 +149,12 @@ public boolean isInterleaved() {
 		return findAnnotation(Interleaved.class) != null;
 	}
 
+	@Override
+	public boolean isEagerlyInterleaved() {
```
meltsufin: Can you move it closer to `isLazyInterleaved` definition?

Issue #2138, opened 2020-01-25
Dependencies bom manages dependencies that do not exist


wilkinsona:
The dependencies bom contains dependency management for `spring-cloud-gcp-trace` but the module no longer exists.

wilkinsona:
`spring-cloud-gcp-config` has the same problem.

meltsufin:
You're absolutely right about the `spring-cloud-gcp-trace` one because it's a module we removed at some point. However, `spring-cloud-gcp-config` and `spring-cloud-gcp-starter-config` are modules that we do not promote to Maven Central because we consider them as "beta", but we do want them to be published to snapshots and milestones.
I guess we just have to use the `central` profile for that.

meltsufin:
Actually, I'm not sure how we could use a Maven profile for this. I guess, we'll just have to tell the users of `spring-cloud-gcp-starter-config` to specify the version.
Oh, and `spring-cloud-gcp-config` doesn't exist.

PR #2140, opened 2020-01-27
Remove missing dependencies from BOM

meltsufin:
Removing `spring-cloud-gcp-trace` and `spring-cloud-gcp-config` from the
BOM because they don't exist.

Also, removing `spring-cloud-gcp-starter-config` because it's never
published to Maven Central.

Fixes: #2138.


meltsufin:
@elefeint I only tested the doc generation, but not the whole build. :flushed: .

Issue #2145, opened 2020-01-28
Datastore Projections should restrict query to contain only the necessary fields


dzou:
When a user creates a Datastore query using a Projection object, the underlying query should contain only the necessary fields to satisfy the projection. 

I think this should be done using the Projection Query: https://cloud.google.com/datastore/docs/concepts/queries#projection_queries

First determine if this being done right now (I think not..)? If not, we can investigate if this is possible.

Flexy-recruitment-backend:
Thanks for the merge!! how can I test the change without waiting for a release?

dzou:
No problem -

Just add the Spring Snapshots repo to your POM:

```
<repository>
	<id>spring-snapshots</id>
	<name>Spring Snapshots</name>
	<url>https://repo.spring.io/libs-snapshot-local</url>
	<snapshots>
		<enabled>true</enabled>
	</snapshots>
</repository>
```

Then you can depend on the `1.3.0.BUILD-SNAPSHOT` version of this project.

dzou:
So the solution of this should also allow users to not have to create object-only constructors for their entities; see this: https://github.com/spring-cloud/spring-cloud-gcp/issues/2139#issuecomment-581588585

dmitry-s:
That isn't so. Object-only constructors will be required anyway.

dzou:
@dmitry-s - I don't quite agree. I think it's possible to not have object-only constructors if we simply convert `nulls` to sensible values for primitives (like `false`, `-1`, etc.). These placeholder values will not matter since the projection proxy prevents users from accessing them.

dmitry-s:
@dzou what I meant is it is not related to  this issue. This issue is about querying only the fields used in the projection interface.

dmitry-s:
Since Datastore projection queries require all properties specified in the projection to be included in an index, this feature is not backwards compatible. 
We would have to introduce some way to switch it on and off.

meltsufin:
As you can see we downgraded the priority of this. If we ever decide to do this, we will want to make it a configurable option because setting up indexes for each projection could be quite unfriendly to the users.

meltsufin:
@dmitry-s What about the note in this issue regarding this not being backwards compatible??

dmitry-s:
@meltsufin while it is true, the error message is clear and suggests the index that needs to be created, so all you have to do is to copy the index definition and deploy it.

PR #2146, opened 2020-01-28
Enable usage of projections within Slice repository queries

### LINKED ISSUE #2133: Projection and Slice return type
Hello, 
I am trying to use projection in order to query an entity in the datastore. I don't manage to use Slice as return type with my projection. The same thing works fine with a Page return type :

My entity is called "Customer" and have a lot of fields.
I have created a projection interface called CustomerProjection  : 

`public interface CustomerProjection {
    String getLastName();
}`

And in my repository, I have the following request : 

`public interface CustomerRepository extends DatastoreRepository<Customer, Long> {
    Page<CustomerProjection> findByIsActiveAndIsDeleted(boolean isActive, boolean isDeleted, Pageable pageable);
}`

This request work well, but when I change the return type in "Slice", I constat at runTime that I obtain a `Slice<Customer>` in place of `Slice<CustomerProjection>`

Is that a bug? is this problem due to some mistake from me?
Regards
CG

dzou:
This fixes the typing issue with projections so you can now use projection objects within your repositories queries that use Slice. i.e.

```
Slice<TestEntityProjection> findByName(String name, Pageable pageable)
```

Fixes #2133.

I think there is a bigger enhancement to make here -- I found that right now there is no performance benefit from using projections since when we issue the query to datastore, we don't restrict the query to only contain the necessary fields of the projection (we still just get everything). Filed #2145 to track.

cc/ @Flexy-recruitment-backend

### DIFF
--- a/spring-cloud-gcp-data-datastore/src/test/java/org/springframework/cloud/gcp/data/datastore/it/DatastoreIntegrationTests.java
+++ b/spring-cloud-gcp-data-datastore/src/test/java/org/springframework/cloud/gcp/data/datastore/it/DatastoreIntegrationTests.java
@@ -796,12 +796,8 @@ public class DatastoreIntegrationTests extends AbstractDatastoreIntegrationTests
 
 	@Test
 	public void testSlicedEntityProjections() {
-		testEntityRepository.saveAll(this.allTestEntities);
-		testEntityRepository.save(new TestEntity(123L, "red", 1L, Shape.CIRCLE, null));
-		testEntityRepository.save(new TestEntity(456L, "blue", 2L, Shape.CIRCLE, null));
-
 		Slice<TestEntityProjection> testEntityProjectionSlice =
-				testEntityRepository.findBySize(1L, PageRequest.of(0, 1));
+				testEntityRepository.findBySize(2L, PageRequest.of(0, 1));
 
 		List<TestEntityProjection> testEntityProjections =
 				testEntityProjectionSlice.get().collect(Collectors.toList());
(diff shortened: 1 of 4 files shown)


spring-cloud-gcp-data-datastore/src/test/java/org/springframework/cloud/gcp/data/datastore/it/TestEntityRepository.java
```diff
@@ -107,6 +107,8 @@
 	@Query("select * from  test_entities_ci where size = @size")
 	TestEntityProjection getBySize(@Param("size") long size);
 
+	Slice<TestEntityProjection> findBySize(@Param("size") long size, Pageable pageable);
```
dmitry-s: remove `@Param` annotation

spring-cloud-gcp-data-datastore/src/test/java/org/springframework/cloud/gcp/data/datastore/it/DatastoreIntegrationTests.java
```diff
@@ -793,6 +793,23 @@ public void sameClassDescendantsTest() {
...
+		testEntityRepository.save(new TestEntity(456L, "blue", 2L, Shape.CIRCLE, null));
+
+		Slice<TestEntityProjection> testEntityProjectionSlice =
+				testEntityRepository.findBySize(1L, PageRequest.of(0, 1));
+
+		List<TestEntityProjection> testEntityProjections =
+				testEntityProjectionSlice.get().collect(Collectors.toList());
+
+		assertThat(testEntityProjections).hasSize(1);
+		assertThat(testEntityProjections.get(0)).isInstanceOf(TestEntityProjection.class);
```
meltsufin: Add assert on the property values.

dzou: Done.

spring-cloud-gcp-data-datastore/src/test/java/org/springframework/cloud/gcp/data/datastore/it/TestEntityRepository.java
dzou: Done.

Flexy-recruitment-backend:
The correction solve the problem.
Regards
CG

Issue #2147, opened 2020-01-29
Add @ConditionalOnProperty to GcpContextAutoConfiguration


elefeint:
**Is your feature request related to a problem? Please describe.**

@ssoydan-strands reported not being able to fully turn off pubsub starter on Gitter

> I have a java-maven project which has a dependency - "spring-cloud-gcp-starter-pubsub"
I want to disable/enable pubsub with a property (not with a profile) but even when I think it is disabled, I see "failed to detect whether we are running on Google Compute Engine" info log.
I tried different properties like below but none of them seems to work. Do you know if any exist? With the dependency added, it always tries to connect.

**Describe the solution you'd like**
We should add a flag to turn off either common concerns (`GcpContextAutoConfiguration`) or a kill switch to disable all `spring-cloud-gcp` integrations. The former is sufficient, but for someone using several of our modules it's a hassle to turn them off one-by-one.

**Describe alternatives you've considered**
There are workarounds for the issue
1) explicitly specify credentials through properties or by registering your own bean of type `CredentialsProvider` ([docs](https://github.com/spring-cloud/spring-cloud-gcp/blob/b2811f70908ecf54c897475ab0b2874e0079e861/docs/src/main/asciidoc/core.adoc#credentials ))
2) continue relying on autodiscovery, but specifically turn off the GCE check with an environment variable `NO_GCE_CHECK=true` ([docs](https://github.com/googleapis/google-auth-library-java/blob/master/README.md#application-default-credentials))

But neither of those is intuitive to someone seeing a GCE warning in the logs.

meltsufin:
Have you considered using the [built-in mechanism in Spring Boot to disable specific auto-configuration classes](https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#using-boot-disabling-specific-auto-configuration)?

```java
@EnableAutoConfiguration(exclude={GcpContextAutoConfiguration.class})
```
or in `application.properties`:
```
spring.autoconfigure.exclude=org.springframework.cloud.gcp.autoconfigure.core.GcpContextAutoConfiguration
```

elefeint:
We don't make our users exclude on per-class basis in our scenarios and instead provide a documented property; why is this different?

meltsufin:
Yes.
Let's add `spring.cloud.gcp.core.enabled`.

ssoydan-strands:
[...]

PR #2150, opened 2020-01-30
Add @ConditionalOnProperty to GcpContextAutoConfiguration

ssoydan-strands:
Add conditional property to GCP core auto configuration.
"spring.cloud.gcp.core.enabled" property enables/disables the GCP core auto configuration.

Related document is updated. Unit tests are added.
Implemented for #2147

### DIFF
--- a/docs/src/main/asciidoc/core.adoc
+++ b/docs/src/main/asciidoc/core.adoc
@@ -24,6 +24,15 @@ dependencies {
 }
 ----
 
+=== Configuration
+
+The following options may be configured with Spring Cloud core.
+
+|===========================================================================
+| Name | Description | Required | Default value
+| `spring.cloud.gcp.core.enabled` | Enables or disables GCP core auto configuration | No | `true`
+|===========================================================================
+
 === Project ID
 
 `GcpProjectIdProvider` is a functional interface that returns a GCP project ID string.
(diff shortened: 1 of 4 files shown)


ssoydan-strands:
> You changes look good!
> Just one thing -- could you revert unrelated formatting changes introduced by the IDE? They are sensible, but we try to keep formatting PRs separate from functionality PRs.

Done!

docs/src/main/asciidoc/core.adoc
```diff
@@ -24,6 +24,15 @@ dependencies {
 }
 ----
 
+=== Configuration
+
+The following options may be configured with Spring Cloud core.
+
+|===========================================================================
+| Name | Description | Required | Default value
+| `spring.cloud.gcp.core.enabled` | Enables or disables GCP core auto configuration | No | `true`
```
meltsufin: Since we're adding this table, (which we should've always had), we should probably also list `spring.cloud.gcp.credentials.location`, `spring.cloud.gcp.credentials.encoded-key`, `spring.cloud.gcp.project-id`, and `spring.cloud.gcp.credentials.scopes`.

I understand that's not the main focus of this PR. So, I would be OK with it being done in a separate PR.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/core/GcpContextAutoConfiguration.java
```diff
@@ -40,8 +41,10 @@
  * @author Mike Eltsufin
  * @author Elena Felder
  * @author Chengyuan Zhao
+ * @author Serhat Soydan
  */
 @Configuration
+@ConditionalOnProperty(name = "spring.cloud.gcp.core.enabled", havingValue = "true", matchIfMissing = true)
```
meltsufin: We should also add it to `spring-cloud-gcp/spring-cloud-gcp-autoconfigure/src/main/resources/META-INF/additional-spring-configuration-metadata.json`.

docs/src/main/asciidoc/core.adoc
elefeint: I filed a follow-up issue #2151 for this.

ssoydan-strands: Better to do it in a separate PR connected to the new issue #2151 , I guess.

PR #2156, opened 2020-01-31
set of small code-style improvements

s13o:


s13o:
Can someone explain to me how changes in javadoc and string concatenations can decrease code coverage?
Something completely wrong with codecov reports.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SqlSpannerQuery.java
```diff
@@ -245,8 +245,7 @@ private List executeReadSql(Pageable pageable, Sort sort, QueryTagValue queryTag
 
 		queryTagValue.sql = SpannerStatementQueryExecutor
 				.applySortingPagingQueryOptions(this.entityType, spannerQueryOptions,
-						resolveEntityClassNames(queryTagValue.sql),
```
dmitry-s: could you clarify why is the call to `resolveEntityClassNames` removed?

dmitry-s: Ok so `executeReadSql` is only called from `executeRawResult`, and `resolveEntityClassNames` is done there. 

I think the better way to fix that would be to call `resolveEntityClassNames` in the constructor, this way the `sql` field will always have resolved class names and there will be no need for any individual method to do that.

meltsufin:
@s13o Don't worry about code coverage. It has to do with the fact that integration tests don't run on PRs. We're able to merge it despite codecov failing.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SqlSpannerQuery.java
dmitry-s: it will be added to the PR

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SqlSpannerQuery.java
```diff
@@ -301,11 +301,10 @@ else if (expression instanceof CompositeStringExpression) {
 
 		String sql;
 
-		QueryTagValue(List<String> tags, Object[] rawParams, Object[] params, String sql,
-				SpannerMappingContext spannerMappingContext) {
+		QueryTagValue(List<String> tags, Object[] rawParams, Object[] params, String sql) {
 			this.tags = tags;
 			this.intialParams = params;
-			this.sql = resolveEntityClassNames(sql, spannerMappingContext);
+			this.sql = sql;
```
s13o: I knew that, but you requested to modify the constructor :)

dmitry-s: not that constructor :)

Issue #2148, opened 2020-01-29
Disabling auto-configuration of instrumenting executor beans reduces visibility


JBodkin-LH:
**Describe the bug**
The documentation for 1.2.1 recommends to disable `spring.sleuth.scheduled.enabled=false` to prevent many traces appearing with the name `async`. This is problematic for applications that use executors under the hood, which can be commonly used in reactive programming when running IO requests or running asynchronous code.

The workaround to exclude the thread pool bean created by this project would be a better recommendation rather than disabling the setting. Ideally as mentioned in the original issue, it would be best to automatically exclude the code in this project from being traced.

Reference: https://github.com/spring-cloud/spring-cloud-gcp/issues/1978

**Sample**

elefeint:
I opened #2016 a couple of months back to completely hide the internal scheduler pool from the spring context.
This might be a good first issue, actually, since #2013 provides an example of hiding a scheduler in another one of our autoconfigurations.

Would you be interested in contributing the code?

Issue #2016, opened 2019-11-05
Hide traceSenderThreadPool in trace autoconfiguration


elefeint:
* Make changes similar to #2013.
* Check documentation from #2011; it may need to change.

PR #2158, opened 2020-01-31
Hide trace scheduler from spring sleuth

JBodkin:
Fixes https://github.com/spring-cloud/spring-cloud-gcp/issues/2148, #2016.

### DIFF
--- a/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/trace/StackdriverTraceAutoConfiguration.java
+++ b/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/trace/StackdriverTraceAutoConfiguration.java
@@ -17,7 +17,10 @@
 package org.springframework.cloud.gcp.autoconfigure.trace;
 
 import java.io.IOException;
+import java.util.Optional;
 import java.util.concurrent.TimeUnit;
+
+import javax.annotation.PreDestroy;
 
 import brave.http.HttpClientParser;
 import brave.http.HttpServerParser;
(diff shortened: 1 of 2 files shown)


spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/trace/StackdriverTraceAutoConfiguration.java
```diff
@@ -121,19 +126,21 @@ public SleuthProperties stackdriverSleuthProperties(SleuthProperties sleuthPrope
...
-		scheduler.setPoolSize(traceProperties.getNumExecutorThreads());
-		scheduler.setThreadNamePrefix("gcp-trace-sender");
-		scheduler.setDaemon(true);
-		return scheduler;
-	}
-
 	@Bean
 	@ConditionalOnMissingBean(name = "traceExecutorProvider")
-	public ExecutorProvider traceExecutorProvider(@Qualifier("traceSenderThreadPool") ThreadPoolTaskScheduler scheduler) {
+	public ExecutorProvider traceExecutorProvider(GcpTraceProperties traceProperties, @Qualifier("traceSenderThreadPool") Optional<ThreadPoolTaskScheduler> userProvidedScheduler) {
```
meltsufin: `Optional` works here? I thought we have to use `ObjectProvider`.
cc/ @artembilan

artembilan: I think it's OK to stick with `Optional` for simple use-cases.

PR #2160, opened 2020-01-31
Nit: fix an error in the comment for `inputChannel`

anguillanneuf:
Unless the code is wrong, the comment has a mistake in it. 

The error `inputChannel` should be`[Pub/Sub topic name].[group name].error` = `[Pub/Sub subscription name].error`, but **not** `[Pub/Sub subscription name with group name].[group name].error` because the Pub/Sub subscription name is the Pub/Sub topic name dot the group name, if one is specified.


dzou:
> Argh, checkstyle:
> 
> ```
> /home/travis/build/spring-cloud/spring-cloud-gcp/spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-binder-sample/src/main/java/com/example/SinkExample.java:49: Line matches the illegal pattern 'Trailing whitespace'. [Regexp]
> ```

I will fix this and merge -

dzou:
Thanks !

Issue #2115, opened 2020-01-13
Add support for Google Cloud Secret Manager


rohanliston:
**Is your feature request related to a problem? Please describe.**

Google Cloud have recently added a service called [Secret Manager](https://cloud.google.com/secret-manager), which provides secure, centralised storage for application-level secrets such as API keys, DB credentials, and so on. This service is comparable to [AWS Secrets Manager](https://aws.amazon.com/secrets-manager/).

It is currently in beta, but will apparently be [officially announced](https://www.reddit.com/r/googlecloud/comments/ebmd5z/google_clouds_new_secrets_manager_secret_manager/fb7srv4/) sometime this month (Jan 2020).

Using a managed service like this would be much more convenient for some users than setting up Vault instances or managing Spring Cloud Config Server. 

**Describe the solution you'd like**

Add support for using Google Cloud Secret Manager as a bootstrap property source, similar to Spring Cloud AWS Secrets Manager.

**Describe alternatives you've considered**

N/A

**Additional context**

N/A

meltsufin:
@rohanliston Thanks for the feature request. I'ts been on our radar and we're planning to start working to implement Spring Cloud Vault support for the GCP Secret Manager.

sethvargo:
Hi @meltsufin - let us (Secret Manager) know if we can help. I just published some Java samples at: https://github.com/GoogleCloudPlatform/java-docs-samples/tree/master/secretmanager

eddumelendez:
Hi, I will be glad to help too :)

meltsufin:
This is great! Thanks for volunteering to help. We can definitely benefit from some suggestions and code reviews from the community.
cc/ @dzou

dzou:
Hey thanks for the interest and support all, I'll be starting work on this shortly and keep you updated on progress along the way.

meltsufin:
Question for those of you asking for this support:
Are you looking to use HashiCorp’s Vault with GCP Secrets Manager as a backend or do you just need a bootstrap property source that will load secrets into the `Environment`?

rohanliston:
@meltsufin We would prefer to use GCP Secret Manager as a bootstrap property source directly.

mikkelbd:
Same use case here. Would be nice to use GCP Secret Manager directly to load secrets into the `Environment` without having to setup and manage Vault.

dzou:
[...]

PR #2168, opened 2020-02-02
Add initial secret manager property source implementation

dzou:
This is the initial scaffolding for setting up Secret Manager as a bootstrap property source.

In progress #2115.

### DIFF
--- a/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/GcpSecretManagerBootstrapConfiguration.java
+++ b/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/GcpSecretManagerBootstrapConfiguration.java
@@ -1,0 +1,72 @@
+/*
+ * Copyright 2017-2020 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.gcp.autoconfigure.secretmanager;
+
+import java.io.IOException;
+
+import com.google.api.gax.core.CredentialsProvider;
+import com.google.cloud.secretmanager.v1beta1.SecretManagerServiceClient;
+import com.google.cloud.secretmanager.v1beta1.SecretManagerServiceSettings;
+
+import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.cloud.bootstrap.config.PropertySourceLocator;
+import org.springframework.cloud.gcp.core.DefaultCredentialsProvider;
+import org.springframework.cloud.gcp.core.DefaultGcpProjectIdProvider;
+import org.springframework.cloud.gcp.core.GcpProjectIdProvider;
+import org.springframework.cloud.gcp.core.UserAgentHeaderProvider;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+@EnableConfigurationProperties(GcpSecretManagerProperties.class)
+@ConditionalOnProperty("spring.cloud.gcp.secretmanager.enabled")
+public class GcpSecretManagerBootstrapConfiguration {
+
+	private final GcpSecretManagerProperties properties;
+
+	private final CredentialsProvider credentialsProvider;
+
+	private final GcpProjectIdProvider gcpProjectIdProvider;
+
+	public GcpSecretManagerBootstrapConfiguration(GcpSecretManagerProperties properties)
+			throws IOException {
+
+		this.properties = properties;
+		this.credentialsProvider = new DefaultCredentialsProvider(properties);
+		this.gcpProjectIdProvider = properties.getProjectId() != null
+				? properties::getProjectId
+				: new DefaultGcpProjectIdProvider();
+	}
+
+	@Bean
+	@ConditionalOnMissingBean
+	public SecretManagerServiceClient secretManagerClient() throws IOException {
+		SecretManagerServiceSettings settings = SecretManagerServiceSettings.newBuilder()
+				.setCredentialsProvider(this.credentialsProvider)
+				.setHeaderProvider(new UserAgentHeaderProvider(GcpSecretManagerBootstrapConfiguration.class))
+				.build();
+
+		return SecretManagerServiceClient.create(settings);
+	}
+
+	@Bean
+	public PropertySourceLocator secretManagerPropertySourceLocator(SecretManagerServiceClient client) {
+		return new SecretManagerPropertySourceLocator(client, this.gcpProjectIdProvider);
+	}
+}
--- a/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/GcpSecretManagerBootstrapConfiguration.java
+++ b/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/GcpSecretManagerBootstrapConfiguration.java
@@ -18,8 +18,6 @@ package org.springframework.cloud.gcp.autoconfigure.secretmanager;
 
 import java.io.IOException;
 
-import javax.annotation.PostConstruct;
-
 import com.google.api.gax.core.CredentialsProvider;
 import com.google.cloud.secretmanager.v1beta1.SecretManagerServiceClient;
 import com.google.cloud.secretmanager.v1beta1.SecretManagerServiceSettings;
@@ -29,7 +27,6 @@ import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
-import org.springframework.boot.convert.ApplicationConversionService;
 import org.springframework.cloud.bootstrap.config.PropertySourceLocator;
 import org.springframework.cloud.gcp.core.DefaultCredentialsProvider;
 import org.springframework.cloud.gcp.core.DefaultGcpProjectIdProvider;
@@ -38,7 +35,7 @@ import org.springframework.cloud.gcp.core.UserAgentHeaderProvider;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.core.convert.converter.Converter;
-import org.springframework.core.convert.converter.ConverterRegistry;
+import org.springframework.core.env.ConfigurableEnvironment;
 
 /**
  * Bootstrap Autoconfiguration for GCP Secret Manager which enables loading secrets as
@@ -69,25 +66,30 @@ public class GcpSecretManagerBootstrapConfiguration {
 				: new DefaultGcpProjectIdProvider();
 	}
 
-	@PostConstruct
-	public void init() {
-		// In this method, we register the converters which convert SecretPayload objects to
-		// Strings and byte[].
-		ConverterRegistry converterRegistry = (ConverterRegistry) ApplicationConversionService.getSharedInstance();
+	/**
+	 * Registers {@link ByteString} type converters to convert to String and byte[].
+	 */
+	@Bean
+	public ConfigurableEnvironment configurableEnvironment(
+			ConfigurableEnvironment configurableEnvironment) {
 
-		converterRegistry.addConverter(new Converter<ByteString, String>() {
-			@Override
-			public String convert(ByteString source) {
-				return source.toStringUtf8();
-			}
-		});
+		configurableEnvironment.getConversionService().addConverter(
+				new Converter<ByteString, String>() {
+					@Override
+					public String convert(ByteString source) {
+						return source.toStringUtf8();
+					}
+				});
+		
+		configurableEnvironment.getConversionService().addConverter(
+				new Converter<ByteString, byte[]>() {
+					@Override
+					public byte[] convert(ByteString source) {
+						return source.toByteArray();
+					}
+				});
 
-		converterRegistry.addConverter(new Converter<ByteString, byte[]>() {
-			@Override
-			public byte[] convert(ByteString source) {
-				return source.toByteArray();
-			}
-		});
+		return configurableEnvironment;
 	}
 
 	@Bean
--- a/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/GcpSecretManagerBootstrapConfiguration.java
+++ b/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/GcpSecretManagerBootstrapConfiguration.java
@@ -56,23 +56,17 @@ public class GcpSecretManagerBootstrapConfiguration {
 
 	private final GcpProjectIdProvider gcpProjectIdProvider;
 
-	public GcpSecretManagerBootstrapConfiguration(GcpSecretManagerProperties properties)
-			throws IOException {
+	public GcpSecretManagerBootstrapConfiguration(
+			GcpSecretManagerProperties properties,
+			ConfigurableEnvironment configurableEnvironment) throws IOException {
 
 		this.properties = properties;
 		this.credentialsProvider = new DefaultCredentialsProvider(properties);
 		this.gcpProjectIdProvider = properties.getProjectId() != null
 				? properties::getProjectId
 				: new DefaultGcpProjectIdProvider();
-	}
 
-	/**
-	 * Registers {@link ByteString} type converters to convert to String and byte[].
-	 */
-	@Bean
-	public ConfigurableEnvironment configurableEnvironment(
-			ConfigurableEnvironment configurableEnvironment) {
-
+		// Registers {@link ByteString} type converters to convert to String and byte[].
 		configurableEnvironment.getConversionService().addConverter(
 				new Converter<ByteString, String>() {
 					@Override
@@ -80,7 +74,7 @@ public class GcpSecretManagerBootstrapConfiguration {
 						return source.toStringUtf8();
 					}
 				});
-		
+
 		configurableEnvironment.getConversionService().addConverter(
 				new Converter<ByteString, byte[]>() {
 					@Override
@@ -88,8 +82,6 @@ public class GcpSecretManagerBootstrapConfiguration {
 						return source.toByteArray();
 					}
 				});
-
-		return configurableEnvironment;
 	}
 
 	@Bean
--- a/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/GcpSecretManagerBootstrapConfiguration.java
+++ b/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/GcpSecretManagerBootstrapConfiguration.java
@@ -18,13 +18,18 @@ package org.springframework.cloud.gcp.autoconfigure.secretmanager;
 
 import java.io.IOException;
 
+import javax.annotation.PostConstruct;
+
 import com.google.api.gax.core.CredentialsProvider;
 import com.google.cloud.secretmanager.v1beta1.SecretManagerServiceClient;
 import com.google.cloud.secretmanager.v1beta1.SecretManagerServiceSettings;
+import com.google.protobuf.ByteString;
 
+import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.boot.convert.ApplicationConversionService;
 import org.springframework.cloud.bootstrap.config.PropertySourceLocator;
 import org.springframework.cloud.gcp.core.DefaultCredentialsProvider;
 import org.springframework.cloud.gcp.core.DefaultGcpProjectIdProvider;
@@ -32,10 +37,20 @@ import org.springframework.cloud.gcp.core.GcpProjectIdProvider;
 import org.springframework.cloud.gcp.core.UserAgentHeaderProvider;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
+import org.springframework.core.convert.converter.Converter;
+import org.springframework.core.convert.converter.ConverterRegistry;
 
+/**
+ * Bootstrap Autoconfiguration for GCP Secret Manager which enables loading secrets as
+ * properties into the application {@link org.springframework.core.env.Environment}.
+ *
+ * @author Daniel Zou
+ * @since 1.3
+ */
 @Configuration
 @EnableConfigurationProperties(GcpSecretManagerProperties.class)
-@ConditionalOnProperty("spring.cloud.gcp.secretmanager.enabled")
+@ConditionalOnClass(SecretManagerServiceClient.class)
+@ConditionalOnProperty(value = "spring.cloud.gcp.secretmanager.enabled", matchIfMissing = true)
 public class GcpSecretManagerBootstrapConfiguration {
 
 	private final GcpSecretManagerProperties properties;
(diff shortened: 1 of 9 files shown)


spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/SecretManagerPropertySource.java
```diff
@@ -0,0 +1,110 @@
...
+
+	/**
+	 * Extracts the Secret ID from the {@link Secret}. The secret ID refers to the unique ID
+	 * given to the secret when it is saved under a GCP project.
+	 *
+	 * <p>
+	 * The secret ID is extracted from the full secret name of the form:
+	 * projects/${PROJECT_ID}/secrets/${SECRET_ID}
+	 */
+	private static String extractSecretId(Secret secret) {
```
sethvargo: In general, what we've been recommending to integrators downstream is the following transformation:

```text
:pid/:sid => projects/:pid/secrets/:sid/versions/latest
:pid/:sid/:vid => projects/:pid/secrets/:sid/versions/:vid
projects/:pid/secrets/:sid => projects/:pid/secrets/:sid/versions/latest
projects/:pid/secrets/:sid/versions/:vid => projects/:pid/secrets/:sid/versions/:vid
```

That way, the result of the parsing is always a fully-qualified secret version reference which you can pass directly to AccessSecretVersion without needing to complicate that method.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/SecretManagerPropertySource.java
```diff
@@ -0,0 +1,110 @@
...
+			SecretManagerServiceClient client, String projectId, String secretId) {
+
+		SecretVersionName secretVersionName = SecretVersionName.newBuilder()
+				.setProject(projectId)
+				.setSecret(secretId)
+				.setSecretVersion(LATEST_VERSION_STRING)
+				.build();
+
+		AccessSecretVersionResponse response = client.accessSecretVersion(secretVersionName);
+		return response.getPayload().getData().toStringUtf8();
```
sethvargo: Should probably document it in SpringBoot that only "string" secrets are accepted. Secret Manager accepts a wide character set for payload values.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/SecretManagerPropertySource.java
```diff
@@ -0,0 +1,110 @@
...
+import com.google.cloud.secretmanager.v1beta1.ProjectName;
+import com.google.cloud.secretmanager.v1beta1.Secret;
+import com.google.cloud.secretmanager.v1beta1.SecretManagerServiceClient;
+import com.google.cloud.secretmanager.v1beta1.SecretManagerServiceClient.ListSecretsPagedResponse;
+import com.google.cloud.secretmanager.v1beta1.SecretVersionName;
+
+import org.springframework.cloud.gcp.core.GcpProjectIdProvider;
+import org.springframework.core.env.EnumerablePropertySource;
+
+public class SecretManagerPropertySource extends EnumerablePropertySource<SecretManagerServiceClient> {
```
meltsufin: javadoc pls

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/SecretManagerPropertySource.java
```diff
@@ -0,0 +1,110 @@
...
+
+	private static final String SECRETS_NAMESPACE = "spring-cloud-gcp.secrets.";
```
meltsufin: Allow this to be configurable in `GcpSecretManagerProperties`?
Maybe some people just want to load the properties directly without any prefix. Anything wrong with that?

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/SecretManagerPropertySourceLocator.java
```diff
@@ -0,0 +1,46 @@
...
+			SecretManagerServiceClient client,
+			GcpProjectIdProvider projectIdProvider) {
+		this.client = client;
+		this.projectIdProvider = projectIdProvider;
+	}
+
+	@Override
+	public PropertySource<?> locate(Environment environment) {
+		return new SecretManagerPropertySource(
+				"spring-cloud-gcp-secret-manager",
```
meltsufin: Maybe make this a constant?

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/SecretManagerPropertySource.java
```diff
@@ -0,0 +1,110 @@
...
+	private final String[] propertyNames;
+
+	public SecretManagerPropertySource(
+			String propertySourceName,
+			SecretManagerServiceClient client,
+			GcpProjectIdProvider projectIdProvider) {
+
+		super(propertySourceName, client);
+
+		Map<String, Object> propertiesMap = initializePropertiesMap(client, projectIdProvider.getProjectId());
```
meltsufin: You're loading properties in the constructor. Would it make sense to allow users to reload the properties by providing a public "refresh" method to do it?

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/SecretManagerPropertySource.java
meltsufin: `public Object getProperty(String name)`
 So, I don't think we really need to convert all values to Strings. @dzou WDYT?

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/it/SecretManagerIntegrationTests.java
```diff
@@ -0,0 +1,131 @@
...
+						+ "Please use '-Dit.secretmanager=true' to enable them.")
+				.isEqualTo("true");
+	}
+
+	@Before
+	public void setupSecretManager() {
+		this.context = new SpringApplicationBuilder()
+				.sources(GcpContextAutoConfiguration.class, GcpSecretManagerBootstrapConfiguration.class)
+				.web(WebApplicationType.NONE)
+				.properties("spring.cloud.gcp.secretmanager.enabled=true")
```
meltsufin: Can we have a test for when it's `false` too?

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/it/SecretManagerIntegrationTests.java
```diff
@@ -0,0 +1,131 @@
...
+	@Test
+	public void testConfiguration() {
+		assertThat(this.context.getEnvironment().getProperty("spring-cloud-gcp.secrets.my-secret"))
+				.isEqualTo("the secret data.");
+	}
+
+	/**
+	 * Creates the secret with the specified payload if the secret does not already exist.
+	 */
+	private void createSecret(String secretId, String payload) {
```
meltsufin: This reminds me, do we want to create a `SecretManagerTemplate`? You would also put the secret reading in there.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/SecretManagerPropertySource.java
dzou: Ah good point. I could just leave it as the [SecretPayload](https://cloud.google.com/secret-manager/docs/reference/rest/v1beta1/SecretPayload) which would be more flexible. Users would then be injecting/reading `SecretPayload` from the environment properties (instead of us forcing strings) and can convert the values to string or their preferred payload.

Do you feel this is a reasonable approach?

meltsufin: Well, I think there might be problems with mapping to `@ConfigurationProperties` classes that have `Strings` and other basic Java types. It's something that we should test.

spencergibb: You wouldn't want `SecretPayload` leaking into user code.

spencergibb: Spring does the conversion automatically. If there isn't a `SecretPayload` encode/decoder it would fail.

dzou: Ohh ok, this `SecretPayload` object can really just be converted to two options: `String` or `byte[]`.

@spencergibb - How do I hook into this encode/decoder concept? Like is it possible for me to override `getProperty(String propertyName, Class<T> targetType)` such that if targetType is `String` I decode to string, and if the target type is `byte[]` I decode to byte[]?

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/SecretManagerPropertySource.java
elefeint: More generally, test behavior of a`@RefreshScope` property in the sample. Ideally, triggering a refresh event should get those properties to expire.

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/it/SecretManagerIntegrationTests.java
elefeint: +1, and this would also go into the new module.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/SecretManagerPropertySourceLocator.java
```diff
@@ -0,0 +1,46 @@
...
+package org.springframework.cloud.gcp.autoconfigure.secretmanager;
+
+import com.google.cloud.secretmanager.v1beta1.SecretManagerServiceClient;
+
+import org.springframework.cloud.bootstrap.config.PropertySourceLocator;
+import org.springframework.cloud.gcp.core.GcpProjectIdProvider;
+import org.springframework.core.env.Environment;
+import org.springframework.core.env.PropertySource;
+
+public class SecretManagerPropertySourceLocator implements PropertySourceLocator {
```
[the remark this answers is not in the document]
spencergibb: Shouldn't it just be conditional on class of the secret manager api rather than setting a property?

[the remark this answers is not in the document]
elefeint: +1

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/GcpSecretManagerBootstrapConfiguration.java
```diff
@@ -0,0 +1,72 @@
...
+import org.springframework.cloud.gcp.core.DefaultCredentialsProvider;
+import org.springframework.cloud.gcp.core.DefaultGcpProjectIdProvider;
+import org.springframework.cloud.gcp.core.GcpProjectIdProvider;
+import org.springframework.cloud.gcp.core.UserAgentHeaderProvider;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+@EnableConfigurationProperties(GcpSecretManagerProperties.class)
+@ConditionalOnProperty("spring.cloud.gcp.secretmanager.enabled")
```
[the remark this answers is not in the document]
dzou: I think this should be disabled by default like our config management; otherwise all applications would be making RPCs to secret manager to list the secrets even if none exist.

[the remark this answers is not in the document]
spencergibb: Are there any required properties in `GcpSecretManagerProperties`?

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/SecretManagerPropertySource.java
dzou: Noted, I simplified the method a bit more, but the `.split()` I am doing is somewhat inevitable because I need to get the shortened secret name which is used as the Key of the Spring environment property.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/GcpSecretManagerBootstrapConfiguration.java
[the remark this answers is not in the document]
dzou: There are no required properties.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/SecretManagerPropertySource.java
dzou: String and byte[] properties are now supported through registering custom SecretPayload converters. See: https://github.com/spring-cloud/spring-cloud-gcp/pull/2168/files#diff-04e47cc423dee80f0eb313f527e9f5b2R74-R90

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/GcpSecretManagerBootstrapConfiguration.java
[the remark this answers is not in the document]
dzou: Added `matchIfMIssing= true` to enable by default and then also added `@ConditionalOnClass(SecretManagerServiceClient.class)` to make conditional on the addition of secret manager class to classpath.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/SecretManagerPropertySource.java
dzou: Done.

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/it/SecretManagerIntegrationTests.java
dzou: Done.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/SecretManagerPropertySourceLocator.java
dzou: Done.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/SecretManagerPropertySource.java
dzou: Regarding the refresh functionality, can I verify this in a separate PR? This would be something to verify through the addition of the sample app and a Properties class.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/SecretManagerPropertySourceLocator.java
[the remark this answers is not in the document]
dzou: Done.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/SecretManagerPropertySource.java
```diff
@@ -0,0 +1,112 @@
...
+
+	private static Map<String, Object> initializePropertiesMap(
+			SecretManagerServiceClient client, String projectId, String secretsNamespace) {
+
+		ListSecretsPagedResponse response = client.listSecrets(ProjectName.of(projectId));
+
+		HashMap<String, Object> secretsMap = new HashMap<>();
+		for (Secret secret : response.iterateAll()) {
+			String secretId = extractSecretId(secret);
+			ByteString secretPayload = getSecretPayload(client, projectId, secretId);
```
sethvargo: Non-blocking, but a potential future optimization would be to parallelize this.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/SecretManagerPropertySource.java
```diff
@@ -0,0 +1,112 @@
...
+		return secretsMap;
+	}
+
+	private static ByteString getSecretPayload(
+			SecretManagerServiceClient client, String projectId, String secretId) {
+
+		SecretVersionName secretVersionName = SecretVersionName.newBuilder()
+				.setProject(projectId)
+				.setSecret(secretId)
+				.setSecretVersion(LATEST_VERSION_STRING)
```
sethvargo: I think people will want to access specific secret versions. Defaulting to "latest" is not a production best practice.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/GcpSecretManagerBootstrapConfiguration.java
```diff
@@ -0,0 +1,109 @@
...
+		this.gcpProjectIdProvider = properties.getProjectId() != null
+				? properties::getProjectId
+				: new DefaultGcpProjectIdProvider();
+	}
+
+	@PostConstruct
+	public void init() {
+		// In this method, we register the converters which convert SecretPayload objects to
+		// Strings and byte[].
+		ConverterRegistry converterRegistry = (ConverterRegistry) ApplicationConversionService.getSharedInstance();
```
spencergibb: Asking the boot team if there is a better way than using a static method.

spencergibb: If I'm not mistaken, you should just be able to register these as beans.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/SecretManagerPropertySource.java
dzou: Noted, added to #2176 to track.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/SecretManagerPropertySource.java
dzou: Ah good point, we'll definitely need an answer for this. I think we'll have to discuss a bit more in-depth with our team what is the best way to let users specify which secret versions they want to load into the environment; I guess we wouldn't want to load all versions of all secrets by default.

I added this point to #2176 to track; we will definitely implement this before releasing to users. I just want this PR to setup the initial scaffolding.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/GcpSecretManagerBootstrapConfiguration.java
dzou: Unfortunately this does not work; I did something like this:

```
public class GcpSecretManagerBootstrapConfiguration {

  ... other beans ...

	@Bean
	public Converter<ByteString, String> getConverterA() {
		return new Converter<ByteString, String>() {
			@Override
			public String convert(ByteString source) {
				return source.toStringUtf8();
			}
		};
	}

	@Bean
	public Converter<ByteString, byte[]> getConverterB() {
		return new Converter<ByteString, byte[]>() {
			@Override
			public byte[] convert(ByteString source) {
				return source.toByteArray();
			}
		};
	}
}
```

spencergibb: Hmm, maybe injecting `ConfigurableEnvironment` and using `ConfigurableEnvironment.getConversionService()` rather than the static method then?

elefeint: I think the beans have to be annotated with `@ConfigurationPropertiesBinding` to be discoverable.

spencergibb: that's specific to binding and `ByteString` isn't used in binding.

dzou: The `ConfigurableEnvironment` method worked for me!

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/GcpSecretManagerBootstrapConfiguration.java
```diff
@@ -0,0 +1,111 @@
...
+		this.gcpProjectIdProvider = properties.getProjectId() != null
+				? properties::getProjectId
+				: new DefaultGcpProjectIdProvider();
+	}
+
+	/**
+	 * Registers {@link ByteString} type converters to convert to String and byte[].
+	 */
+	@Bean
+	public ConfigurableEnvironment configurableEnvironment(
```
spencergibb: I wouldn't redefine a `ConfigurableEnvironment` bean. Just put it as a constuctor arg and call a method.

dzou: Done.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/SecretManagerPropertySource.java
```diff
@@ -0,0 +1,112 @@
...
+	public String[] getPropertyNames() {
+		return propertyNames;
+	}
+
+	@Override
+	public Object getProperty(String name) {
+		return properties.get(name);
+	}
+
+	private static Map<String, Object> initializePropertiesMap(
```
meltsufin: This doesn't actually "initialize" the properties map because it has no side effect.
```suggestion
	private static Map<String, Object> readSecretsPropertiesMap(
```

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/SecretManagerPropertySourceLocator.java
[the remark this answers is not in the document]
meltsufin: Can we add a tracking issue to add a template?

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/it/SecretManagerIntegrationTests.java
```diff
@@ -0,0 +1,148 @@
...
+	@Test
+	public void testConfiguration() {
+		assertThat(this.context.getEnvironment().getProperty("my-secret"))
+				.isEqualTo("the secret data.");
+
+		byte[] byteArraySecret = this.context.getEnvironment().getProperty("my-secret", byte[].class);
+
+		// This byte array is the equivalent representation to "the secret data."
+		assertThat(byteArraySecret).isEqualTo(
+				new byte[] { 116, 104, 101, 32, 115, 101, 99, 114, 101, 116, 32, 100, 97, 116, 97, 46 });
```
meltsufin: `"the secret data.".getBytes()` or something like that instead?

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/SecretManagerPropertySourceLocator.java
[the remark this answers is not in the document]
dzou: Yes, this is tracked in #2176

dzou:
@elefeint @meltsufin - PTAL when you guys get a chance :)

PR #2173, opened 2020-02-04
Add Javadoc Configuration for Project

dzou:
Adds Javadoc generation scripts for spring-cloud-gcp.

This provides the Kokoro scripts which upload the javadocs to:
https://googleapis.dev/java/spring-cloud-gcp/latest/index.html

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -216,6 +216,23 @@
 				<artifactId>appengine-maven-plugin</artifactId>
 				<version>${app-engine-maven-plugin.version}</version>
 			</plugin>
+			<plugin>
+				<groupId>org.apache.maven.plugins</groupId>
+				<artifactId>maven-javadoc-plugin</artifactId>
+				<configuration>
+					<excludePackageNames>
+						com.example,
+						com.example.*,
+						<!-- Exclude samples and classes in the Firestore Google Client library packages -->
+						com.google.cloud.firestore
+					</excludePackageNames>
+					<links>
+						<link>https://googleapis.dev/java/google-cloud-vision/latest/</link>
+						<link>https://googleapis.dev/java/google-cloud-spanner/latest/</link>
+						<link>https://googleapis.dev/java/google-cloud-clients/latest/</link>
+					</links>
+				</configuration>
+			</plugin>
 		</plugins>
 	</build>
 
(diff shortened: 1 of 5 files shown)


pom.xml
```diff
@@ -216,6 +216,23 @@
 				<artifactId>appengine-maven-plugin</artifactId>
 				<version>${app-engine-maven-plugin.version}</version>
 			</plugin>
+			<plugin>
+				<groupId>org.apache.maven.plugins</groupId>
+				<artifactId>maven-javadoc-plugin</artifactId>
+				<configuration>
+					<excludePackageNames>
+						com.example,
```
meltsufin: Instead of excluding by package name, can we just disable the javadoc plugin for all samples?

pom.xml
```diff
@@ -216,6 +216,23 @@
...
+						com.example.*,
+						<!-- Exclude samples and classes in the Firestore Google Client library packages -->
+						com.google.cloud.firestore
+					</excludePackageNames>
+					<links>
+						<link>https://googleapis.dev/java/google-cloud-vision/latest/</link>
```
meltsufin: What's the purpose of these links?

pom.xml
dzou: I tried disabling the javadoc plugin `<skip>true</skip>` from the samples root directory; unfortunately it does not work.

pom.xml
dzou: It makes it so that any client library javadoc classes referenced in the spring-cloud-gcp javadoc will link successfully to the client library javadocs.

meltsufin: Cool! How do we know that we covered all of such references?

pom.xml
meltsufin: You might have to do it on each individual sample?

dzou: Unfortunately it still does not work, there is something different about running this command from root versus in the individual directories. For example, when I applied the `skip` to the sample root, the samples get skipped when I run the javadoc command from the samples root, however it does not get skipped when running from project root.

pom.xml
dzou: It's a best effort guess :(

dzou:
Hey @meltsufin, I feel pretty good about merging this in now; PTAL.

We learned that successive runs of the kokoro jobs does indeed update the docs, so I feel the configuration in this PR is correct. The docs do not get updated on the `/latest` url but on the url with the version specified.

So you can see that:

This gets refreshed (and does not have the com.example packages)
https://googleapis.dev/java/spring-cloud-gcp/1.3.0.BUILD-SNAPSHOT/index.html

But this does not get refreshed:
https://googleapis.dev/java/spring-cloud-gcp/latest/index.html

The reason for this appears to be an issue with the googleapis.dev infrastructure which we are following up on.

Issue #2165, opened 2020-01-31
@Interleaved(lazy=false) is still lazy for SqlSpannerQuery


s13o:
There is one more place where `@Interleaved(lazy=false)` still not supported even after the merge of  #2110. The method `SqlSpannerQuery#executeReadSql` will execute a raw SQL without injection of `ARRAY (SELECT AS STRUCT ...)` for Interleaved lists. But we can do it more or less safely in a method `SpannerStatementQueryExecutor.applySortingPagingQueryOptions` line 151 by replacing the
```
 new StringBuilder("SELECT * FROM (").append(sql).append(")"), (o) -> {
```
with something like
```
 new StringBuilder("SELECT *").append(<inject Interleaved here>).appenr(" FROM (").append(sql).append(")"), (o) -> {
```

I'm afraid that it caused a bug with `@Query` annotation now - `@Interleaved(lazy=false)` fields are not "proxied" anymore so they will be not loaded from the database, never, am I right?

dmitry-s:
`@Query` annotation is intended to be used with custom projection queries. I don't think we want to modify a user provided query, because users might want to retrieve a limited set of fields for performance reasons.

@s13o could you clarify your use case? Is it possible to use name-based query method instead of `@Query` annotated one?

Thanks

s13o:
1. We do modify the SQL provided by the `@Query`. Please, take a look on `SqlSpannerQuery#executeReadSql` line 245. 
2. We can at least add a "fetchEagerInterleaved" option into the `@Query`. But, as for me, more natural will be to have `skipInterleaved` with default value `true` to skip fetching any Interleaved at all. But if `skipnterleaved=false` we should fetch them as expected from the definition of `@Interleaved` annotation.
3. Let's have a look on an entity like this
```
@Table(name = "Entity")
public class Entity {
    final UUID id;
    
    @Interleaved(lazy = true)
    private List<ApocProviderHist> lazyInterleaved;
    
    @Interleaved(lazy = false)
    private List<ApocProviderHist> eagerInterleaved;
}
```
I'm afraid today with `@Query("select * from Entity")` we will have a list of Entities with a correctly proxied `lazyInterleaved` fields but `eagerInterleaved` will always be empty

dmitry-s:
[...]

PR #2174, opened 2020-02-04
Eager Interleaved support added to a @Query annotation (#2165)

s13o:
a new param `fetchInterleaved` with defaults `false` was added to the `@Query`


spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/Query.java
```diff
@@ -51,4 +52,11 @@
 	 * query method is executed as a DML query.
 	 */
 	boolean dmlStatement() default false;
+
+	/**
+	 * When {@code false} then any {@link Interleaved} lists of the fetched entity will be lazy-loaded.
+	 * Otherwise an SQL will be wrapped to eager-load lists with {@link Interleaved#lazy()} equals fo "false".
```
dmitry-s:
```suggestion
	 * Otherwise an SQL will be wrapped to eager-load lists with {@link Interleaved#lazy()} equals to {@code false} .
```

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryExecutor.java
```diff
@@ -303,18 +308,23 @@ private static void bindParameter(ValueBinder<Statement.Builder> bind,
...
 			if (spannerPersistentProperty.isEagerInterleaved()) {
 				Class childType = spannerPersistentProperty.getColumnInnerType();
 				SpannerPersistentEntity childPersistentEntity = mappingContext.getPersistentEntity(childType);
 				joiner.add(getChildrenStructsQuery(
-					childPersistentEntity, spannerPersistentEntity, mappingContext, spannerPersistentProperty.getColumnName()));
+						childPersistentEntity, spannerPersistentEntity, mappingContext, spannerPersistentProperty.getColumnName()));
 			}
 		});
-		String childrenSubquery = joiner.toString();
-		return String.join(", ", spannerPersistentEntity.columns())
```
dmitry-s: why was `String.join(", ", spannerPersistentEntity.columns())` removed?

dmitry-s:
@s13o overall looks good, a couple of small issues. Could you also add an integration test for this new functionality? SpannerRepositoryIntegrationTests is the right class for that.

Thanks!

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/Query.java
```diff
@@ -51,4 +52,11 @@
...
+	 * @return {@code false} if {@link Interleaved} lists should be skipped.
+	 */
+	boolean fetchInterleaved() default false;
```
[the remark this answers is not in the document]
s13o: Well, it still not a perfect name. I hear it like "fetchAnyInterleavedAlwaysEagerly". And the `fetchInterleavedEagerly` is still a very long name.
In the worst situation, any interleaved will be proxied and lazy-fetched in any case.
How about to rename it to `eagerFetch`? it sounds like - by default we have a lazy-fetch but we can enable eager-fetch when needed.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryExecutor.java
s13o: It is not removed - I've split an old method `getColumnsStringForSelect` on two - a new method `getChildrenSubquery` + old method `getColumnsStringForSelect`
So the old method now looks like
```
	public static String getColumnsStringForSelect(
			SpannerPersistentEntity<?> spannerPersistentEntity, SpannerMappingContext mappingContext) {
		return String.join(", ", spannerPersistentEntity.columns())
				+ getChildrenSubquery(spannerPersistentEntity, mappingContext);
	}
```

and all "joins" are in place

s13o:
> @s13o overall looks good, a couple of small issues. Could you also add an integration test for this new functionality? SpannerRepositoryIntegrationTests is the right class for that.
> 
> Thanks!

With my pleasure.
I have tested it as a snapshot dependency on a real project - it works well.
I just don't know where these tests are. Are they inherited from `AbstractSpannerIntegrationTest` in the same repo?

s13o:
> @s13o overall looks good, a couple of small issues. Could you also add an integration test for this new functionality? SpannerRepositoryIntegrationTests is the right class for that.
> 
> Thanks!

With my pleasure

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/Query.java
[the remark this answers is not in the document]
meltsufin: I'm OK with `fetchInterleaved`. I would just ask to create an integration test that uses `@Query(.... fetchInterleaved=true)` directly on an actual repository.

spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/repository/it/SpannerRepositoryIntegrationTests.java
```diff
@@ -246,6 +255,28 @@ public void queryMethodsTest() {
...
+		Mockito.verify(spannerTemplate, Mockito.times(1))
+				.executeQuery(any(Statement.class), any(SpannerQueryOptions.class));
+
+		Mockito.clearInvocations(spannerTemplate);
+		assertThat(tradeRepository.fetchByIdLazy(aTrade.getId()))
+				.isNotEmpty()
+				.hasValueSatisfying(t -> assertThat(t.getId()).isEqualTo(aTrade.getId()))
+				.hasValueSatisfying(t -> assertThat(t.getTraderId()).isEqualTo(aTrade.getTraderId()))
+				.hasValueSatisfying(t -> assertThat(t.getSymbol()).isEqualTo(aTrade.getSymbol()))
+				.hasValueSatisfying(t -> assertThat(t.getSubTrades()).hasSize(aTrade.getSubTrades().size()));
```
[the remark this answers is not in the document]
s13o: the `Trade` has default value for `subTrades` field like
```java
	@Interleaved
	private List<SubTrade> subTrades = Collections.emptyList();
```
Than is why such non-null validation in class `SpannerTemplate` line 588 is not enouph
```java
					//an interleaved property can only be List
					List propertyValue = (List) accessor.getProperty(spannerPersistentProperty);
					if (propertyValue != null) {
						resolveChildEntities(propertyValue, null);
						return;
					}
```
It was added after the rev: 26e9adf

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryExecutor.java
dmitry-s: I see, makes sense

spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/repository/it/SpannerRepositoryIntegrationTests.java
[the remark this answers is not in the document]
dmitry-s: Good catch. How do you think that can be fixed?
Maybe we should nullify the laze loaded fields on the parent entity retrieval?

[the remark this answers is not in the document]
s13o: I've tried to tell about this problem we could potentially have before but my Eng is not so fluent to be clear, sorry.
I have no clever idea yet :(
I think the most strong solution could be to have some additional flag in a`spannerPersistentProperty` that it has been populated already. 

"nulling" of fields is not a good idea - the field could be created by Lombok builder and it could be a business logic (like enty-to-json transformers) that expect the field can never be null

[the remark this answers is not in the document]
meltsufin: Discussing with @dmitry-s offline, he pointed out that it would just be null until someone tries to access it, at which point it will be lazy-loaded. So, no one should ever see it as null.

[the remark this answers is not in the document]
s13o: It sounds like a hack.
What the next step should be? Are we going to merge this PR and will do a fix later or we can collaborate somehow in another way?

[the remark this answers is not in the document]
dmitry-s: I will add a fix in this branch.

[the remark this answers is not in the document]
s13o: so little I know about the GitHub : (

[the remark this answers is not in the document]
dmitry-s: Actually, it has nothing to do with the default value. The thing is, `subTrades` is not a lazy loaded property. That's why when we convert result rows to entities, we expect the field to be present. If it is not, we assume that there are no children and assign an empty list.

meltsufin:
@s13o We've discussed the `fetchInterleaved` on `@Query` offline with @dmitry-s, and basically we decided that because it can conflict with specification in `@Interleaved`, we should just remove it and simply respect whatever is specified in `@Interleaved` regarding lazy loading or not.

The only complication is in what to do with projection queries. Here we will inspect the properties chosen for the projection, and lookup the corresponding `@Interleaved` annotation in the Entity for those properties to decide weather to eagerly or lazy-load them.

So for this PR, let's remove `fetcInterleaved` and simply assume that it's always true, and then have follow-up PRs for whatever else needs to be fixed up.

spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/repository/it/SpannerRepositoryIntegrationTests.java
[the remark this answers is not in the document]
dmitry-s: Sorry, my test was wrong. You are right, it is because of the default value.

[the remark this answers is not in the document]
s13o: in theory, we can to some "magic proxy" for each interleaved ... but it seems to be too complex to be stable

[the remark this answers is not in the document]
s13o: we can do a very simple decision - is the code reached line 593 of SpannerTemplate line it means that any eager-interleaved is already fetched so any other should be a proxy for "lazy-interleaved"
I think the only scenario when it is not true - it is @Query(fetchInterleaved=false)
Is not it? 
Probably I'm wrong...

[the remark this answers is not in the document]
s13o: in my implementation of PR #2132 I built an SQL-query (ResultSet) and a function to process it at the same time in parallel - it was a "pair". So when I'm started to process RS I had only to invoke a function.
Your PR is much simple so it is better, but there is a single place here where it is difficult to understand which fields are populated already. 

I don't understand how this magic method works (ConverterAwareMappingSpannerEntityProcessor#mapToList(com.google.cloud.spanner.ResultSet, java.lang.Class<T>, java.util.Set<java.lang.String>, boolean))
```
entityReader.read(entityClass,
	resultSet.getCurrentRowAsStruct(),
	includeColumns,
	allowMissingColumns)
```
And I can't propose any solution yet :(
But this method is a good place to mark any Interleaved as 'populated' at processing of it

s13o:
> @s13o We've discussed the `fetchInterleaved` on `@Query` offline with @dmitry-s, and basically we decided that because it can conflict with specification in `@Interleaved`, we should just remove it and simply respect whatever is specified in `@Interleaved` regarding lazy loading or not.
> 
> The only complication is in what to do with projection queries. Here we will inspect the properties chosen for the projection, and lookup the corresponding `@Interleaved` annotation in the Entity for those properties to decide weather to eagerly or lazy-load them.
> 
> So for this PR, let's remove `fetcInterleaved` and simply assume that it's always true, and then have follow-up PRs for whatever else needs to be fixed up.

Well, sounds reasonable
Ok, I can finish it as you told me but I.m not really familiar with "projections" yet. So I'm not sure that I can help with them enough fast

s13o:
Ops, another problem with @Query ("select count (*) from XXX") - we can have "interleaved" in the entity XXX but there is no interleaved subqueries at all in this SQL. Integration tests fail...

dmitry-s:
Hi @s13o,

It seems reasonable that we do not need to augment query for anything but select. Let's put Projections aside for now, we will handle that in a separate issue.

I see this issue grew in complexity significantly. Do you want to keep working on it or do you need help? Let us know.

Thanks!

s13o:
> I see this issue grew in complexity significantly. Do you want to keep working on it or do you need help? Let us know.
> 
> Thanks!

I'm going to do a PR today soon. At least to discuss it

s13o:
Done. I hope it close to expected by you, gentlemen.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryExecutor.java
```diff
@@ -162,6 +171,27 @@ private SpannerStatementQueryExecutor() {
...
+		final String subquery = getChildrenSubquery(persistentEntity, mappingContext);
+		if (subquery.isEmpty()) {
+			return sql;
+		}
+
+		final Pair<String, String> parts = splitSelectAll(sql);
+		return parts.getSecond().isEmpty() ? sql : parts.getFirst() + subquery + parts.getSecond();
+	}
+
+	private static Pair<String, String> splitSelectAll(String sql) {
```
dmitry-s: Could you clarify what is this method for?

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
```diff
@@ -352,25 +352,25 @@ private void applyDeleteMutations(Iterable<?> objects, List<Mutation> mutations)
 	}
 
 	@Override
-	public void delete(Class entityClass, Key key) {
+	public <T> void delete(Class<T> entityClass, Key key) {
```
dmitry-s: T is never used, please remove

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
```diff
@@ -352,25 +352,25 @@ private void applyDeleteMutations(Iterable<?> objects, List<Mutation> mutations)
...
 		applyDeleteMutations(entityClass, KeySet.newBuilder().addKey(key).build(),
 				Collections.singletonList(this.mutationFactory.delete(entityClass, key)));
 	}
 
 	@Override
-	public void delete(Class entityClass, KeySet keys) {
+	public <T> void delete(Class<T> entityClass, KeySet keys) {
```
dmitry-s: T is never used, please remove

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
```diff
@@ -352,25 +352,25 @@ private void applyDeleteMutations(Iterable<?> objects, List<Mutation> mutations)
...
 		applyDeleteMutations(entityClass, keys, Collections
 				.singletonList(this.mutationFactory.delete(entityClass, keys)));
 	}
 
-	private void applyDeleteMutations(Class entityClass, KeySet keys, List<Mutation> mutations) {
+	private <T> void applyDeleteMutations(Class<T> entityClass, KeySet keys, List<Mutation> mutations) {
```
dmitry-s: T is never used, please remove

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
```diff
@@ -352,25 +352,25 @@ private void applyDeleteMutations(Iterable<?> objects, List<Mutation> mutations)
...
 		maybeEmitEvent(new BeforeDeleteEvent(mutations, null, keys, entityClass));
 		applyMutations(mutations);
 		maybeEmitEvent(new AfterDeleteEvent(mutations, null, keys, entityClass));
 	}
 
 	@Override
-	public long count(Class entityClass) {
+	public <T> long count(Class<T> entityClass) {
```
dmitry-s: T is never used, please remove

s13o: np, but it was declared in SpannerOperations.
It was done mainly to remove the warning with code inspection about "Raw use of parametrized class 'Class'".
Still, remove?

meltsufin: Matching what's in `SpannerOperations` makes sense to me.

dmitry-s: Makes sense, let's keep it.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
s13o: Inherited from the SpannerOperations

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
s13o: Inherited from the SpannerOperations

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryExecutor.java
s13o: The method name is probably bad, but...
The initial idea was to use wrapped query `SELECT * FROM (<initial query>)` in the same way as the method `applySort` at line 160. It was fine until we decided to use the `fetchInterleaved` as permanently `true`. After a number of failed integration tests and attempts to fix them, I understood that the wrapped query will not work. Mainly because of Spanner does not preserve sort ordering of embedded queries.
Then I concluded - when the user has created an SQL with a complete list of fields to select it is not really good idea to add some more fields with additional sub-selects - it could really affect performance to select fields that were not requested. But when the query is like `select * from Table ...` - it is the best candidate to split it into two parts an add interleaved fields into the middle. Like `select *, ` + <interlieved> + ` from Table ...` and it does not break any skip\limit or sorting conditions.

dmitry-s: I don't really like that we have to rely on `SELECT_ALL_PATTERN`, there could be cases where it doesn't work and that would lead to inconsistent behavior. It is not transparent and would be hard to detect. 

Is there a way to avoid regex matching? What if we use joins instead of sub-queries?

s13o:
Guys, any update with the review?

meltsufin:
@s13o Did you see @dmitry-s's concern about the use of regex?

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryExecutor.java
dmitry-s: To clarify, the query I'm thinking about would be something like that:
```
 SELECT     *
FROM       (SELECT * FROM   singers WHERE  singerid = 1) AS singers
INNER JOIN
           (
                    SELECT   singerid,
                             array_agg(struct(albumid, albumtitle, array
                             (
                                    select as struct singerid,
                                           albumid,
                                           albumtitle,
                                           array
                                           (
                                                  SELECT AS struct *
                                                  FROM   songs
                                                  WHERE  songs.singerid = albums.singerid
                                                  AND    songs.albumid = albums.albumid)))) AS alb
                    FROM     albums
                    GROUP BY albums.singerid) AS albums
ON         albums.singerid = singers.singerid
LIMIT      10
OFFSET     30 
```
where `SELECT * FROM   singers WHERE  singerid = 1` is the user-provided query.

What do you think about that approach?

s13o:
> @s13o Did you see @dmitry-s's concern about the use of regex?

No, but now I see.
I didn't like the regex too, but this solution works with all the tests we have now.
I'm not sure I can find another way.
I think soon or later we will have to parse the SQL from the "@Query". We already have a hidden problem when the order/slip/limit exists in SQL and we getting it from `SpannerPageableQueryOptions` at the same time. It's tricky to understand what should be in the result.
Are you strongly against the regex or we can discuss it and cover by better tests?

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryExecutor.java
s13o: I understood
It looks more complex but the main - it is not clear how to process such resultset

s13o: the new version is better :)

s13o: here we have original SQL as a source for a new query - it is pointed in "... FROM (SELECT * FROM   singers WHERE  singerid = 1)". 
But the sort order of such SQL is not preserved

dmitry-s: Does it preserve order in current implementation? I believe we wrap the user-provided query in `SELECT * FROM (<query>) LIMIT .. OFFSET ..`

s13o: No, it does not.
That is a major problem. :(
We have an integration test that will fail.
Some test from `SpannerRepositoryIntegrationTests#queryMethodsTest` line ~ 192 fails unpredictable on assertions `isGreaterThan` like
```
		List<Trade> buyTradesRetrieved = this.tradeRepository
				.annotatedTradesByAction("BUY");
		assertThat(buyTradesRetrieved).containsExactlyInAnyOrderElementsOf(trader1BuyTrades);
		assertThat(buyTradesRetrieved.get(0).getId()).isGreaterThan(buyTradesRetrieved.get(1).getId());
		assertThat(buyTradesRetrieved.get(1).getId()).isGreaterThan(buyTradesRetrieved.get(2).getId());
```

s13o: We really wrap the user-provided queries by `SELECT * FROM (<query>) LIMIT .. OFFSET ..`. We doing it ONLY when we have a `SpannerPageableQueryOptions` parameter - in such case any "sorting" from it will overwrite the pointed by the SQL itself. And we should at least specify it in doc documentation I believe.

meltsufin: I think we should just accept the fact that the sort order in custom `@Query`s will not be preserved. The solutions is to pass in the sort order in the `Pageable` argument. This is definitely something that needs to be in our refdoc.

dmitry-s: @s13o I think it's ok if ordering is not preserved. We apply ordering from `Pageable`, and that's how it should be passed. 

Could you roll back changes that introduced regexp?
I think after that we should be able to merge this PR.

Thanks

s13o: oops, will try
but I'll have to change the test `SpannerRepositoryIntegrationTests#queryMethodsTest` line ~ 192

s13o: May I ask you, why you don't like the idea to inject child-subquery after "select *"? It seems to be simple enough. I can't really imagine a case when it could fail. Right now - the only leading spaces can affect it. Ok, the regex could be like
```java
Pattern.compile("^\\s+?select(\\s+?all|\\s+?distinct)*(?<star>\\s+?\\*)", Pattern.CASE_INSENSITIVE);
```
The query is parsed only once, it is from an annotation, so we can cache the result and it should not affect performance.

dmitry-s: Sounds good. The sort order should be determined by the `Pageable` that is passed to the method. That is: 
```
PageRequest
				.of(2, 2, org.springframework.data.domain.Sort.by(Order.asc("id")))
```
lines 188-189

s13o: but you didn't answer me about the regex. 
You probably don't like Pearl :)

s13o: I need to have it merged so I'll revert the regex, but, please your opinion is very important to me.

meltsufin: @s13o We avoid parsing the SQL in principle. Experience has taught us that it's a rather risky option. Even augmenting queries is risky, but we decided to go with it in the absence of better alternatives.
For example, note that Spanner SQL may contain [query hints](https://cloud.google.com/spanner/docs/query-syntax). That's one situation where parsing can fall apart.

dmitry-s: @s13o You don't have to merge it to undo a commit. I think `git revert <commit hash>` should do the trick.

s13o: You goddamn right. I also hated the idea to parse the SQL from the beginning but saw no other alternative to satisfy failed IT tests.
Many thanks for the beautiful explanation.

s13o: > @s13o You don't have to merge it to undo a commit. I think `git revert <commit hash>` should do the trick.

Finally, I found a time to complete the PR!
Sorry for the delay, please review.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/Query.java
```diff
@@ -40,6 +41,9 @@
 	/**
 	 * Takes a Cloud Spanner SQL string to define the actual query to be executed. This one will
 	 * take precedence over the method name then.
+	 * <p/>Please, pay attention that sorting conditions should be passed by a {@link Pageable} parameter
```
dmitry-s:
```suggestion
	 * <p/>Please note that sorting options should be passed by a {@link Pageable} parameter
```

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/Query.java
```diff
@@ -40,6 +41,9 @@
...
+	 * instead of the query part. Otherwise, with the current implementation, the sorting could be broken
```
dmitry-s:
```suggestion
	 * instead of the query. Otherwise the sorting could be broken
```

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/Query.java
```diff
@@ -40,6 +41,9 @@
...
+	 * by a logic that fetches eager-interleaved fields when they are in the query result.
```
dmitry-s:
```suggestion
	 * by logic that fetches eager-interleaved fields.
```

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryExecutor.java
```diff
@@ -131,24 +131,33 @@ private SpannerStatementQueryExecutor() {
 	 * @param sql the sql that will be wrapped with sorting and paging options.
 	 * @param mappingContext a mapping context to convert between Cloud Spanner column names
 	 *     and underlying property names.
+	 * @param fetchInterleaved when {@code true} additional subqueries will be added
+	 *     to fetch eager-Interleaved lists with a single query.
+	 *     Please, pay attention there is no sense to pass it as {@code true} when the {@param sql} already contains
```
dmitry-s:
```suggestion
	 *     Please note there is no sense to pass it as {@code true} when the {@param sql} already contains
```

dmitry-s:
There are some weird CI tests issues. I'm trying to fix them.

s13o:
> There are some weird CI tests issues. I'm trying to fix them.

Weird, I saw the build green

meltsufin:
> > There are some weird CI tests issues. I'm trying to fix them.
> 
> Weird, I saw the build green
> Oh, no, It seems to be related with Spring test context :(

We don't run integration tests on PRs in Travis.

s13o:
> > > There are some weird CI tests issues. I'm trying to fix them.
> > 
> > 
> > Weird, I saw the build green
> > Oh, no, It seems to be related with Spring test context :(
> 
> We don't run integration tests on PRs in Travis.

I see
But I did it locally before the push.
It always works in IntelliJ but crashed in a command line with slightly different exceptions on different JDK

s13o:
nice

dmitry-s:
fixes #2165

PR #2188, opened 2020-02-07
Fix Datastore broken test that requires an index

dmitry-s:


dzou:
Datastore indices can be added using the `gcloud` CLI:
https://cloud.google.com/appengine/docs/standard/python/config/indexconfig

spring-cloud-gcp-data-datastore/src/test/java/org/springframework/cloud/gcp/data/datastore/it/DatastoreIntegrationTests.java
```diff
@@ -845,15 +850,34 @@ public void testPageableGqlEntityProjectionsSlice() {
...
 
-	//An index containing both "color" and "size" is required
-	@Ignore
-	@Test
-	public void testSliceString() {
-		Slice<String> slice =
-				this.testEntityRepository.getSliceStringBySize(2L, PageRequest.of(0, 3));
+	@Test(timeout = 10000L)
+	public void testSliceString() throws IOException, URISyntaxException, InterruptedException {
+		URL resource = this.getClass().getResource("/index.yaml");
```
dzou: Can you put the index creating logic in a helper function and then put it in the `@Before` of the test? So that in the future if we have another test that needs an index it will be clear how to add it.

spring-cloud-gcp-data-datastore/src/test/java/org/springframework/cloud/gcp/data/datastore/it/DatastoreIntegrationTests.java
```diff
@@ -845,15 +850,34 @@ public void testPageableGqlEntityProjectionsSlice() {
...
+		}
+
+		Slice<String> slice;
+		try {
+			slice = this.testEntityRepository.getSliceStringBySize(2L, PageRequest.of(0, 3));
+		}
+		catch (DatastoreException e) {
+			if (e.getMessage().contains("no matching index found")) {
+				throw new RuntimeException(
+						"The required index is not found. "
```
dzou: I don't think you need the try/catch to print an extra message for the user; it would be just as clear to add a 1-liner comment that the test may fail if running for the first time since it takes time to create the index.

PR #2192, opened 2020-02-12
remove index creation to fix a Datastore test

dmitry-s:


spring-cloud-gcp-data-datastore/src/test/java/org/springframework/cloud/gcp/data/datastore/it/DatastoreIntegrationTests.java
```diff
@@ -851,41 +845,14 @@ public void testPageableGqlEntityProjectionsSlice() {
...
-
-		if (process.waitFor() != 0) {
-			throw new RuntimeException("Error while creating index.");
-		}
-
-		try {
-			runnable.run();
-		}
-		catch (DatastoreException e) {
-			if (e.getMessage().contains("no matching index found")) {
```
meltsufin: Should we keep this though?

Issue #2176, opened 2020-02-05
Secret Manager Support Tasks Tracking


dzou:
This thread will be used to track the tasks/feature requests associated with https://github.com/spring-cloud/spring-cloud-gcp/issues/2115 .

The tasks include:
- [x] Add Refdocs
- [x] Add Sample App
- [x] Enable Secret Manager on travis.
- [x] Verify that `@RefreshScope` works --- This doesn't work...
- [x] Add Secret Manager Template (if need is present)
- [x] Split code into separate module from autoconfig if template is added
- [ ] Consider [how to make property loading parallelized/non-blocking](https://github.com/spring-cloud/spring-cloud-gcp/pull/2168#discussion_r375371939)
- [x] Secret Manager should allow users to specify versions; [defaulting to "latest" is not a production best practice.](https://github.com/spring-cloud/spring-cloud-gcp/pull/2168#discussion_r375371939)

dzou:
The tasks in this bug have all been completed (with the exception of the parallelized-loading feature).

Not 100% the parallel-loading idea is possible, I was doing some thinking and I think for this to work, there would need to be a way for the properties to be mapped to `ListenableFutures` and then the user to access these in a call-back oriented way. For now to keep things simple lets just keep things synchronous and revisit the async feature if demand for it emerges.

PR #2195, opened 2020-02-13
Add basic template for Secret Manager

dzou:
Add template class for interfacing with Secret Manager which supports creating and reading secrets.

Contributes to #2176.


spring-cloud-gcp-secretmanager/src/main/java/org/springframework/cloud/gcp/secretmanager/SecretManagerTemplate.java
```diff
@@ -0,0 +1,196 @@
...
+	/**
+	 * Gets the secret payload of the specified {@code secretId} at version
+	 * {@code versionName}.
+	 *
+	 * @param secretId unique identifier of your secret in Secret Manager.
+	 * @param versionName which version of the secret to load. The version can be a version
+	 *     number as a string (e.g. "5") or an alias (e.g. "latest").
+	 * @return The secret payload as byte[]
+	 */
+	public byte[] getSecretPayload(String secretId, String versionName) {
```
meltsufin: `getSecretBytes` perhaps?

spring-cloud-gcp-secretmanager/src/main/java/org/springframework/cloud/gcp/secretmanager/SecretManagerTemplate.java
```diff
@@ -0,0 +1,196 @@
...
+		return getSecretVersion(secretId, versionName).toByteArray();
+	}
+
+	private ByteString getSecretVersion(String secretId, String versionName) {
```
meltsufin: I know it's a private method, but maybe even make it public?
Also for naming consistency: `getSecretByteString`?

spring-cloud-gcp-secretmanager/src/main/java/org/springframework/cloud/gcp/secretmanager/SecretManagerTemplate.java
```diff
@@ -0,0 +1,196 @@
...
+	 * Returns the lower-level {@link SecretManagerServiceClient} client object for making API
+	 * calls to Secret Manager service.
+	 *
+	 * <p>
+	 * Useful for executing more advanced use-cases that are not covered by
+	 * {@link SecretManagerTemplate}.
+	 *
+	 * @return the {@link SecretManagerServiceClient} client object.
+	 */
+	public SecretManagerServiceClient getSecretManagerServiceClient() {
```
meltsufin: Why do we need to expose it?
It's a bean in the application context anyway.

spring-cloud-gcp-secretmanager/src/main/java/org/springframework/cloud/gcp/secretmanager/SecretManagerTemplate.java
```diff
@@ -0,0 +1,196 @@
...
+	 * <p>
+	 * If there is already a secret saved in SecretManager with the specified
+	 * {@code secretId}, then it simply creates a new version under the secret with
+	 * the secret {@code payload}.
+	 *
+	 * @param secretId the secret ID of the secret to create.
+	 * @param payload the secret payload; supported payload types: (UTF-8 encoded) String and
+	 *     byte[].
+	 */
+	public void createSecret(String secretId, Object payload) {
```
meltsufin: Why not just have two overloaded methods: one with `String` and one with `byte[]`?

spring-cloud-gcp-secretmanager/src/main/java/org/springframework/cloud/gcp/secretmanager/SecretManagerTemplate.java
```diff
@@ -0,0 +1,196 @@
...
+	 */
+	private boolean secretExists(String secretId) {
+		ProjectName projectName = ProjectName.of(this.projectIdProvider.getProjectId());
+		ListSecretsPagedResponse listSecretsResponse = this.secretManagerServiceClient.listSecrets(projectName);
+
+		return StreamSupport.stream(listSecretsResponse.iterateAll().spliterator(), false)
+				.anyMatch(secret -> secret.getName().contains(secretId));
+	}
+
+	private ByteString convertToByteString(Object rawPayload) {
```
meltsufin: You can get rid of this method if instead you rely on the two possible types: `String` and `byte[]`.

spring-cloud-gcp-secretmanager/src/main/java/org/springframework/cloud/gcp/secretmanager/SecretManagerTemplate.java
```diff
@@ -0,0 +1,196 @@
...
+import org.springframework.cloud.gcp.core.GcpProjectIdProvider;
+
+/**
+ * Offers convenience methods for performing common operations on Secret Manager including
+ * creating and reading secrets.
+ *
+ * @author Daniel Zou
+ * @since 1.3
+ */
+public class SecretManagerTemplate {
```
meltsufin: Typically we would create an interface as well called `SecretManagerOperations`.

spring-cloud-gcp-secretmanager/src/main/java/org/springframework/cloud/gcp/secretmanager/SecretManagerTemplate.java
dzou: Done.

spring-cloud-gcp-secretmanager/src/main/java/org/springframework/cloud/gcp/secretmanager/SecretManagerTemplate.java
dzou: Done.

spring-cloud-gcp-secretmanager/src/main/java/org/springframework/cloud/gcp/secretmanager/SecretManagerTemplate.java
dzou: Done. Removed.

spring-cloud-gcp-secretmanager/src/main/java/org/springframework/cloud/gcp/secretmanager/SecretManagerTemplate.java
dzou: Done.

spring-cloud-gcp-secretmanager/src/main/java/org/springframework/cloud/gcp/secretmanager/SecretManagerTemplate.java
dzou: Done.

spring-cloud-gcp-secretmanager/src/main/java/org/springframework/cloud/gcp/secretmanager/SecretManagerTemplate.java
dzou: Done.

spring-cloud-gcp-secretmanager/src/main/java/org/springframework/cloud/gcp/secretmanager/SecretManagerOperations.java
```diff
@@ -0,0 +1,89 @@
...
+	 * {@code payload}.
+	 *
+	 * @param secretId the secret ID of the secret to create.
+	 * @param payload the secret payload string.
+	 */
+	void createSecret(String secretId, String payload);
+
+	/**
+	 * Creates a new secret using the provided {@code secretId} and creates a new version of
+	 * the secret with the provided {@code payload}.
```
elefeint: "Creates a new secret or a new version of existing secret"?

spring-cloud-gcp-secretmanager/src/main/java/org/springframework/cloud/gcp/secretmanager/SecretManagerOperations.java
```diff
@@ -0,0 +1,89 @@
...
+	 *
+	 * <p>
+	 * If there is already a secret saved in SecretManager with the specified
+	 * {@code secretId}, then it simply creates a new version under the secret with the secret
+	 * {@code payload}.
+	 *
+	 * @param secretId the secret ID of the secret to create.
+	 * @param payload the secret payload string.
+	 */
+	void createSecret(String secretId, String payload);
```
elefeint: I wonder if `setSecret` is more precise, since the secret could be created _or_ updated.

spring-cloud-gcp-secretmanager/src/main/java/org/springframework/cloud/gcp/secretmanager/SecretManagerOperations.java
```diff
@@ -0,0 +1,89 @@
...
+	/**
+	 * Gets the secret payload of the specified {@code secretId} at version
+	 * {@code versionName}.
+	 *
+	 * @param secretId unique identifier of your secret in Secret Manager.
+	 * @param versionName which version of the secret to load. The version can be a version
+	 *     number as a string (e.g. "5") or an alias (e.g. "latest").
+	 * @return The secret payload as String
+	 */
+	String getSecretString(String secretId, String versionName);
```
elefeint: Does it make sense to just ask for `getSecretString(String secretId)` without specifying a version and getting back `latest` label by default? It feels like the most common usecase.

spring-cloud-gcp-secretmanager/src/main/java/org/springframework/cloud/gcp/secretmanager/SecretManagerTemplate.java
```diff
@@ -0,0 +1,147 @@
...
+
+	public SecretManagerTemplate(
+			SecretManagerServiceClient secretManagerServiceClient,
+			GcpProjectIdProvider projectIdProvider) {
+		this.secretManagerServiceClient = secretManagerServiceClient;
+		this.projectIdProvider = projectIdProvider;
+	}
+
+	@Override
+	public void createSecret(String secretId, String payload) {
```
elefeint: since we know whether a secret is new or existing, would it make sense to return a boolean flag, an enum or a version number indicating the result?

spring-cloud-gcp-secretmanager/src/main/java/org/springframework/cloud/gcp/secretmanager/SecretManagerTemplate.java
```diff
@@ -0,0 +1,147 @@
...
+		if (!secretExists(secretId)) {
```
elefeint: This is not a large amount of logic, but it is repeated between the different overloaded versions of this method.
I think I'd call `createSecret(String, byte[])` right away and not do any version determination in this method.

spring-cloud-gcp-secretmanager/src/main/java/org/springframework/cloud/gcp/secretmanager/SecretManagerTemplate.java
```diff
@@ -0,0 +1,147 @@
...
+	 * Note that the {@link Secret} object does not contain the secret payload. You must
+	 * create versions of the secret which stores the payload of the secret.
+	 */
+	private void createSecret(String secretId) {
+		ProjectName projectName = ProjectName.of(projectIdProvider.getProjectId());
+
+		Secret secret = Secret.newBuilder()
+				.setReplication(
+						Replication.newBuilder()
+								.setAutomatic(Replication.Automatic.newBuilder().build())
```
elefeint: You can remove the intermediate `build()` calls -- proto builders usually take ether a final object or its builder.

spring-cloud-gcp-secretmanager/src/main/java/org/springframework/cloud/gcp/secretmanager/SecretManagerTemplate.java
```diff
@@ -0,0 +1,147 @@
...
+		this.secretManagerServiceClient.createSecret(request);
+	}
+
+	/**
+	 * Returns true if there already exists a secret under the GCP project with the
+	 * {@code secretId}.
+	 */
+	private boolean secretExists(String secretId) {
+		ProjectName projectName = ProjectName.of(this.projectIdProvider.getProjectId());
+		ListSecretsPagedResponse listSecretsResponse = this.secretManagerServiceClient.listSecrets(projectName);
```
elefeint: This can be a lot of secrets.
Can [getSecret()](https://github.com/googleapis/java-secretmanager/blob/d4f471f8b3ad5d2d24f848a886a4a6dd40231500/google-cloud-secretmanager/src/main/java/com/google/cloud/secretmanager/v1beta1/SecretManagerServiceClient.java#L546) be used instead to retrieve the specific secret?

spring-cloud-gcp-secretmanager/src/main/java/org/springframework/cloud/gcp/secretmanager/SecretManagerOperations.java
dzou: Ah, so in Secret Manager previous versions of secrets are never deleted; i.e. the `Secret` is like a container/parent, and then you create `SecretVersions` under the Secret. I think "create" would be correct, but not sure if there may be better choices for what is being created; like do you feel `createSecretVersion` may be more clear?

elefeint: I think it's okay to leave it as is.
I thought about `createSecretVersion` but that takes focus off "secret" and onto "version", which is also not ideal.

dzou: Sounds good, will leave as is.

spring-cloud-gcp-secretmanager/src/main/java/org/springframework/cloud/gcp/secretmanager/SecretManagerOperations.java
dzou: Done.

spring-cloud-gcp-secretmanager/src/main/java/org/springframework/cloud/gcp/secretmanager/SecretManagerOperations.java
dzou: I added an overloaded method which omits versionName to return the latest.

According to a secret manager dev defulting to accessing the latest version may not be a "production best practice" so I think we ought to keep a method that accepts versions. https://github.com/spring-cloud/spring-cloud-gcp/pull/2168#discussion_r375372256

spring-cloud-gcp-secretmanager/src/main/java/org/springframework/cloud/gcp/secretmanager/SecretManagerTemplate.java
dzou: I prefer `void` just to keep it simple for now; but definitely open to implementing this if users request for it in the future.

spring-cloud-gcp-secretmanager/src/main/java/org/springframework/cloud/gcp/secretmanager/SecretManagerTemplate.java
dzou: Done.

spring-cloud-gcp-secretmanager/src/main/java/org/springframework/cloud/gcp/secretmanager/SecretManagerTemplate.java
dzou: Done.

spring-cloud-gcp-secretmanager/src/main/java/org/springframework/cloud/gcp/secretmanager/SecretManagerTemplate.java
dzou: Done! The getSecret approach is now being used; requires a try/catch style.

spring-cloud-gcp-secretmanager/src/main/java/org/springframework/cloud/gcp/secretmanager/SecretManagerTemplate.java
```diff
@@ -0,0 +1,151 @@
...
+				.setSecret(secret)
+				.build();
+		this.secretManagerServiceClient.createSecret(request);
+	}
+
+	/**
+	 * Returns true if there already exists a secret under the GCP project with the
+	 * {@code secretId}.
+	 */
+	private boolean secretExists(String secretId) {
```
meltsufin: Could it be useful as a public method?

dzou: Done; made public.

dzou:
> It would be nice to get 100% test coverage and some integration tests.
> 
> Also, are you planning to update auto-configuration in a separate PR?

Done, added integration tests and additional test coverage in unit tests.

Ya I will update the auto-config in a followup, the need will arise in the sample application updates.

dzou:
PTAL - should be good now; I think the subsequent build failures are flakes.

dzou:
> Can you just replace `@since 1.3` with `@since 1.2.2`?

Done.

spring-cloud-gcp-secretmanager/pom.xml
```diff
@@ -0,0 +1,29 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <parent>
+    <artifactId>spring-cloud-gcp</artifactId>
+    <groupId>org.springframework.cloud</groupId>
+    <version>1.3.0.BUILD-SNAPSHOT</version>
```
meltsufin:
```suggestion
    <version>1.2.2.BUILD-SNAPSHOT</version>
```

dzou: Nice catch, thanks.

PR #2196, opened 2020-02-14
document collections and lists

### LINKED ISSUE #2189: Add tests and documentation for child entities collection in Firestore
Currently it is not obvious if we support that feature. We should clarify that we do in docs. Also, we should have tests and/or samples.

dmitry-s:
fixes #2189

### DIFF
--- a/docs/src/main/asciidoc/firestore.adoc
+++ b/docs/src/main/asciidoc/firestore.adoc
@@ -113,7 +113,7 @@ NOTE: Internally we use Firestore client library object mapping. See https://dev
 
 ==== Embedded entities and lists
 Spring Data Cloud Firestore supports embedded properties of custom types and lists. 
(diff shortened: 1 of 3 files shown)


docs/src/main/asciidoc/firestore.adoc
```diff
@@ -112,6 +111,18 @@ This annotation is optional, by default the collection name is derived from the
 
 NOTE: Internally we use Firestore client library object mapping. See https://developers.google.com/android/reference/com/google/firebase/firestore/package-summary[the documentation] for supported annotations.
 
+==== Embedded entities and lists
+Spring Data Cloud Firestore supports embedded properties of custom types and lists. 
+Given a custom POJO defenition, you can have properties of this type or lists of this type in your entities.
```
meltsufin:
```suggestion
Given a custom POJO definition, you can have properties of this type or lists of this type in your entities.
```

Issue #2203, opened 2020-02-18
Enhance support of the PENDING_COMMIT_TIMESTAMP 


s13o:
The support of "PENDING_COMMIT_TIMESTAMP" is enough primitive in SpringData for Spanner today. We actually can't use it with only a Spring Data in a business logic when we need to have several timestamp fields in different objects be consistently filled by an actual commit timestamp value on reaching some business condition. It is possible only with DML statements like `UPDATE Singers SET LastUpdated = PENDING_COMMIT_TIMESTAMP() WHERE SingerId = 1` or Mutations like 
```java
Mutation.newInsertBuilder("Performances")
            .set("SingerId")
            .to(performance.singerId)
            .set("VenueId")
            .to(performance.venueId)
            .set("EventDate")
            .to(performance.eventDate)
            .set("Revenue")
            .to(performance.revenue)
            .set("LastUpdateTime")
            .to(Value.COMMIT_TIMESTAMP)
            .build());
```
A simple solution might be to declare a constant value like the `Value.COMMIT_TIMESTAMP` does. Unfortunately. It is impossible today because the 3-party developer can't overwrite some "default converters" like `JavaToSpannerTimestampConverter` used at `SpannerConverters#JAVA_SQL_TO_SPANNER_DATE_CONVERTER`.

meltsufin:
Have you tried passing in epoch for the timestamp? It looks like that's what `Value.COMMIT_TIMESTAMP` is.

s13o:
> Have you tried passing in epoch for the timestamp? It looks like that's what `Value.COMMIT_TIMESTAMP` is.

I did my investigation a half of year ago. I know that `SpannerConverters` has been changed since that. Even when the option to use `Value.COMMIT_TIMESTAMP` exists now it is still not clear, undocumented and untrivial I believe.

If I'm wrong (it will be nice btw), please, close this issue and provide a link on the documentation or some example how to use it.

s13o:
I think I can do a PR for it enough fast at least to discuss what I'm talking about but, please, after the merge of #2165 :)

Issue #2202, opened 2020-02-18
"Soft-delete" filter support for Interleaved collections


s13o:
The "soft-delete" is a very common practice in relational databases when records just marked "as deleted" but remain in the database for historical or audit needs. The Spanner has`@Interleaved` annotation as good support of one-to-many relation. Unfortunately, there is no way exists to exclude "soft-deleted" children from Interleaved lists at fetching from the database.

It will be nice to add support of `@Where` and\or `@Filter` annotations that Hibernate has.

s13o:
Well, I have no worked example for this one but I believe we can do it enough fast after all improvements of eager-interleaved we have gotten.

PR #2208, opened 2020-02-19
Enhance support of the PENDING_COMMIT_TIMESTAMP

s13o:
Fixes: #2203.


s13o:
> Overall looks good, but I have a question.
> Why don't we introduce an annotation (let's say `@CommitTimestamp`)? In this way users won't need to manually set value to their date/time fields. When they save an object we could put `Value.COMMIT_TIMESTAMP` in those fields.

We have already an attribute in a `@Column(spannerCommitTimestamp=true)` annotation. It is not enough. Annotations work in statics - so it works always when exists, at any commit.
But we need a business logic that will say: "well, we are going to mark the entity as `soft-deleted` now (see #2202 pls). It means that the column `Timestamp::deletetAt` should be updated to store the `PENDING_COMMIT_TIMESTAMP` value. At the same time, this entity already has a column `Timestamp::updatedAt` that already marked by the annotation `@Column(spannerCommitTimestamp=true)`. And we need to have both these fields absolutely consistent when the commit has gone"

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/CommitTimestamp.java
```diff
@@ -0,0 +1,62 @@
+/*
+ * Copyright 2017-2018 the original author or authors.
```
meltsufin:
```suggestion
 * Copyright 2017-2020 the original author or authors.
```

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/CommitTimestamp.java
```diff
@@ -0,0 +1,62 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
+import com.google.cloud.spanner.Value;
+
+/**
+ * The factory method {@link #of(Class)} should be used to create a specific value of a pointed class
+ * that will be converted then to the {@link Value#COMMIT_TIMESTAMP} by {@link SpannerConverters}.
+ * It should be used when the database column has a Timestamp type with option "allow_commit_timestamp"
+ * and the business logic needs to manage when the PENDING_COMMIT_TIMESTAMP should be stored into this field.
```
meltsufin: Please add `@author`.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/SpannerConverters.java
```diff
@@ -143,17 +146,44 @@ public LocalDateTime convert(Timestamp timestamp) {
...
+			return toTimestamp(date);
 		}
 	};
 
+	/**
+	 * A converter from {@link java.util.Date} to the Spanner timestamp type.
+	 */
+	// @formatter:off
+	public static final Converter<java.sql.Date, Timestamp>
+			SQL_DATE_TIMESTAMP_CONVERTER = new Converter<java.sql.Date, Timestamp>() {
```
meltsufin: Do we really need this new converter? Seems like it would be the subject of another PR.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/SpannerConverters.java
```diff
@@ -167,17 +197,32 @@ public Timestamp convert(java.util.Date date) {
 					return timestamp.toDate();
 				}
 			};
+	/**
+	 * A converter from the Spanner timestamp type to {@link java.util.Date}.
+	 */
+	// @formatter:off
+	public static final Converter<Timestamp, java.sql.Date> TIMESTAMP_SQL_DATE_CONVERTER =
```
meltsufin: Same for this one.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/SpannerConverters.java
s13o: There is some problem with such "code" converters - no one can overwrite them because of initialization restrictions...
Basically I agree, I'll think about how to improve it better.

s13o:
> LGTM, but please add integration tests

I didn't find a good template for it. I've added some unit tests instead. 
Can you propose a scenario for an integration test in common words?

s13o:
The integration test added.
But the database should be re-created because schema has changed.

s13o:
Hello, any progress with review of this PR of when I can expect it?

dmitry-s:
Hi @s13o,

there is a test failure:
 ConverterAwareMappingSpannerEntityWriterTests.testCommitTimestampsType:380->lambda$testCommitTimestampsType$0:381 Â» IllegalArgument

I was waiting until it is fixed.

Thanks

s13o:
I can't reproduce it. Just have it re-tested locally again, with all IT tests - all tests are passed fine.
Can you re-test it?

dmitry-s:
sure, restarting CI

s13o:
I've added one more IT test, it needs additional table - I didn't want to affect existing tests
How we can recreate test database schema?

dmitry-s:
@s13o you can create tables. Take a look at `AbstractSpannerIntegrationTest.createDatabaseWithSchema()`

dmitry-s:
@s13o,  the test failed again:
```
[ERROR] testCommitTimestampsType  Time elapsed: 0.003 s  <<< ERROR!
java.lang.IllegalArgumentException: "CommitTimestamp" is not supported for the type class [Z
	at org.springframework.cloud.gcp.data.spanner.core.convert.ConverterAwareMappingSpannerEntityWriterTests.lambda$testCommitTimestampsType$0(ConverterAwareMappingSpannerEntityWriterTests.java:381)
	at org.springframework.cloud.gcp.data.spanner.core.convert.ConverterAwareMappingSpannerEntityWriterTests.testCommitTimestampsType(ConverterAwareMappingSpannerEntityWriterTests.java:380)
```

s13o:
The only reason why the test `ConverterAwareMappingSpannerEntityWriterTests` could fail is when the class `SpannerConverters` was not loaded into the classloader before the test. But I can't really imagine how it could be possible? I can't reproduce it locally

s13o:
> class [Z

what is it? what a class is it?

s13o:
well, it is probably some synthetic field, JaCoCo can add some mess I remember...

it should be like this
```java
		doWithFields(CommitTimestamps.class,
				f -> setField(f, entity, CommitTimestamp.of(f.getType())),
				ff -> !ff.isSynthetic() && Objects.isNull(ff.getAnnotation(PrimaryKey.class)));
```

Has sense?

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/CommitTimestamp.java
```diff
@@ -0,0 +1,118 @@
...
+		});
+		return converter;
+	}
+
+	/**
+	 * A specific template of "to timestamp" converter that any custom converter should follow
+	 * to support "CommitTimestamp" feature.
+	 * @param <S> a source type of converter.
+	 */
+	public static abstract class CommitTimestampSupplier<S> implements Converter<S, Timestamp> {
```
dmitry-s: I don't think this class should be abstract, you never use it for inheritance.
Also, it should not be named `CommitTimestampSupplier` as it does not supply anything. 
It seems that `CommitTimestampDecorator` would be a more appropriate name.

Does it make sense?

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/SpannerConverters.java
```diff
@@ -112,16 +116,12 @@ public LocalDate convert(com.google.cloud.Date date) {
...
-		public Timestamp convert(LocalDateTime dateTime) {
-			return JAVA_TO_SPANNER_TIMESTAMP_CONVERTER.convert(java.sql.Timestamp.valueOf(dateTime));
-		}
-	};
+	public static final Converter<LocalDateTime, Timestamp> LOCAL_DATE_TIME_TIMESTAMP_CONVERTER =
+			CommitTimestamp.register(
+					new CommitTimestampSupplier<LocalDateTime>(
+							Value.COMMIT_TIMESTAMP.toSqlTimestamp().toLocalDateTime(),
+							SpannerConverters::toTimestamp
+					) { });
```
dmitry-s: Is it really necessary to use anonymous class?

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/SpannerConverters.java
```diff
@@ -273,4 +257,46 @@ public ByteArray convert(byte[] bytes) {
...
+	public static Timestamp toTimestamp(@NonNull LocalDateTime dateTime) {
+		return toTimestamp(java.sql.Timestamp.valueOf(dateTime));
+	}
+
+	/**
+	 * An utility function to convert an {@link Instant} value to the {@link Timestamp}.
+	 * @param instant the value to convert
+	 * @return the equivalent Timestamp value
+	 */
+	public static Timestamp toTimestamp(@NonNull Instant instant) {
```
dmitry-s: If any of those methods are used only once, please just inline them.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryExecutor.java
```diff
@@ -133,7 +133,7 @@ private SpannerStatementQueryExecutor() {
 	 *     and underlying property names.
 	 * @param fetchInterleaved when {@code true} additional subqueries will be added
 	 *     to fetch eager-Interleaved lists with a single query.
-	 *     Please note there is no sense to pass it as {@code true} when the {@param sql} already contains
+	 *     Please note there is no sense to pass it as {@code true} when the {@code sql} already contains
```
dmitry-s:
```suggestion
	 *     Please note, it doesn't make sense to pass it as {@code true} when the {@code sql} already contains
```

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/SpannerConverters.java
s13o: unfortunately, yes, it is necessary.
Today any `Converter` should be mandatory a class, not a lambda-function - types of "source" and "target" are resolved from the class as TypeArguments :(

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/CommitTimestamp.java
s13o: a better name is `CommitTimestampConverter`
It should be abstract, yes, I've answered about it in the next comment

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/SpannerConverters.java
s13o: well, I had a hope to re-use as `public static <D extends java.util.Date> Timestamp toTimestamp(@NonNull D date)` for a sql-data converter in my project :)

Can we move them to `ConversionUtils` class?

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/CommitTimestamp.java
dmitry-s: I would prefer to call it a `Decorator` because it follows the Decorator design pattern (https://en.wikipedia.org/wiki/Decorator_pattern)

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/SpannerConverters.java
```diff
@@ -273,4 +262,15 @@ public ByteArray convert(byte[] bytes) {
 					SPANNER_TO_JAVA_SQL_DATE_CONVERTER,
 					TIMESTAMP_LOCAL_DATE_CONVERTER,
 					TIMESTAMP_LOCAL_DATE_TIME_CONVERTER));
+
+	/**
+	 * An utility function to convert a {@link java.sql.Timestamp} value to the {@link Timestamp}.
```
dmitry-s:
```suggestion
	 * A utility function to convert a {@link java.sql.Timestamp} value to the {@link Timestamp}.
```

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/convert/CommitTimestamp.java
```diff
@@ -0,0 +1,118 @@
...
+	 * @param <S> a source type of converter.
+	 */
+	public static abstract class CommitTimestampDecorator<S> implements Converter<S, Timestamp> {
+
+		final S commitTimestamp;
+		private final Function<S, Timestamp> converter;
+
+		protected CommitTimestampDecorator(S commitTimestamp, Function<S, Timestamp> converter) {
+			this.commitTimestamp = commitTimestamp;
+			this.converter = converter;
```
dmitry-s: is it possible call `CommitTimestamp.register(this)` here and avoid calling it from `SpannerComvertors` class?

s13o: It was like you telling me from the very beginning.
Then I changed it because I thought you will not like when constructor doing something not tightly related to the construction of the object

s13o: changing back...

PR #2219, opened 2020-02-26
"Soft-delete" filter support for Interleaved collections

s13o:
Fixes: #2202


spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/mapping/Where.java
```diff
@@ -0,0 +1,44 @@
...
+
+import static java.lang.annotation.ElementType.FIELD;
+import static java.lang.annotation.ElementType.TYPE;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+/**
+ * Where clause to add to the element Entity or target entity of a collection.
+ * The clause is written in SQL. A common use case here is for soft-deletes.
+ * It can be used on class level or on interleaved list as well.
+ * It overwrites the class annotation when used on an interleaved list of the same type.
```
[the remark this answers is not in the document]
s13o: It means that English is not my native, sorry for that.
I wanted to say that IF the annotation is used on an interleaved list column on some "parent" entity it will overwrite the same annotation on the child entity class.

For instance - a "child" class is annotated like 

```java
@Table(name = "CHILDREN")
@Where("deleted=false")
public class Child  {
   ...
}
```
But we can create some "patent" with a column like
```java
    @Column
    @Where("deleted=true")
    private List<Child> children;
```
In such case the `@Where("deleted=true")` will be used when we fetch a "parent" entity.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryExecutor.java
```diff
@@ -162,6 +165,11 @@ private SpannerStatementQueryExecutor() {
 							.getPersistentProperty(o.getProperty());
 					return (property != null) ? property.getColumnName() : o.getProperty();
 				});
+		String whereCause = Optional.ofNullable(persistentEntity.findAnnotation(Where.class))
```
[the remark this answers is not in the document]
s13o: ok, will check, thanks

s13o:
I'm going to add more integration tests.
It will be nice to get more comments from you, guys, before that. 
Did I miss any repository that should also support the `@Where`?

s13o:
Can someone assign @dmitry-s  to review the PR?
Unfortunately, I can't :(

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/mapping/Where.java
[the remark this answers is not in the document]
meltsufin:
```suggestion
 * It overwrites the class-level `@Where` annotation when used on an interleaved list property of the same type.
```

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
```diff
@@ -213,9 +216,11 @@ public long executePartitionedDmlStatement(Statement statement) {
 			SpannerPersistentEntity<T> persistentEntity) {
 		//an entity is eligible if all of the following true:
 		//1. entity has eager loaded properties
-		//2. there are no read options, as they can't be applied to a query
-		//3. key set does not have ranges, as they can't be used in a query
+		//2. entity has "Where" annotation
+		//3. there are no read options, as they can't be applied to a query
+		//4. key set does not have ranges, as they can't be used in a query
 		return persistentEntity.hasEagerlyLoadedProperties() &&
+				persistentEntity.findAnnotation(Where.class) == null &&
```
dmitry-s: create `getWhere()` method in `SpannerPersistentEntity`

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
```diff
@@ -213,9 +216,11 @@ public long executePartitionedDmlStatement(Statement statement) {
 			SpannerPersistentEntity<T> persistentEntity) {
 		//an entity is eligible if all of the following true:
 		//1. entity has eager loaded properties
-		//2. there are no read options, as they can't be applied to a query
-		//3. key set does not have ranges, as they can't be used in a query
+		//2. entity has "Where" annotation
```
dmitry-s: Could you explain why can't an entity be eagerly loaded if it is annotated with `@Where`?

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryExecutor.java
```diff
@@ -218,27 +227,45 @@ private SpannerStatementQueryExecutor() {
...
+		String condition = join(or.size() == 0 ? "" : or.size() == 1 ? or.get(0)
+				: or.stream().collect(Collectors.joining(") OR (", "(", ")")), whereCause);
 		String sb = "SELECT " + getColumnsStringForSelect(persistentEntity, mappingContext, true) + " FROM "
-				+ persistentEntity.tableName() + (cond.isEmpty() ? "" : " WHERE " + cond);
+				+ persistentEntity.tableName() + (condition.isEmpty() ? "" : " WHERE " + condition);
 		return buildStatementFromSqlWithArgs(sb, tags, null, writeConverter,
 				keyParts.toArray(), null);
 	}
 
+	private static String join(String cond1, String cond2) {
```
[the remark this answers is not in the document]
s13o: ok

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryExecutor.java
```diff
@@ -218,27 +227,45 @@ private SpannerStatementQueryExecutor() {
...
 				tagNum++;
 			}
-			orJoiner.add(andJoiner.toString());
+			or.add(andJoiner.toString());
 		}
-		String cond = orJoiner.toString();
+		String whereCause = Optional.ofNullable(persistentEntity.findAnnotation(Where.class))
+				.map(Where::value).orElse("");
+		String condition = join(or.size() == 0 ? "" : or.size() == 1 ? or.get(0)
+				: or.stream().collect(Collectors.joining(") OR (", "(", ")")), whereCause);
```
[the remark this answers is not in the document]
s13o: I did it as a first attempt. The stream always returns some value and when we have only one element it became wrapped by bracket - it looks dummy. We need a custom "Joiner" to prevent extra brackets when the stream has a single entry.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryExecutor.java
```diff
@@ -218,27 +227,45 @@ private SpannerStatementQueryExecutor() {
...
 	}
 
+	private static String join(String cond1, String cond2) {
+		if (cond1.isEmpty()) {
+			return cond2;
+		}
+		if (cond2.isEmpty()) {
+			return cond1;
+		}
+		StringJoiner where = new StringJoiner(") AND (", "(", ")");
```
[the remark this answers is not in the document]
s13o: no, some of them or even both can be an empty string

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryExecutor.java
```diff
@@ -324,7 +351,11 @@ private static String getChildrenSubquery(
 				Class childType = spannerPersistentProperty.getColumnInnerType();
 				SpannerPersistentEntity childPersistentEntity = mappingContext.getPersistentEntity(childType);
 				joiner.add(getChildrenStructsQuery(
-						childPersistentEntity, spannerPersistentEntity, mappingContext, spannerPersistentProperty.getColumnName()));
+						childPersistentEntity, spannerPersistentEntity, mappingContext, spannerPersistentProperty.getColumnName(),
+						Optional.ofNullable(spannerPersistentProperty.getWhere().map(Where::value)
+								.orElseGet(() -> Optional.ofNullable(AnnotatedElementUtils
+										.findMergedAnnotation(childType, Where.class)).map(Where::value).orElse(null))))
```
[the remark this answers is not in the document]
s13o: ok, will try
Thanks!

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
s13o: It can. But when the interleaved property with a list of such entries has an own `@Where` - it should have higher priority than an annotation on a class level

s13o: I think the `isEligibleForEagerFetch` should be renamed

dmitry-s: `isEligibleForEagerFetch` is intended to determine if we can and should run `executeReadQueryAndResolveChildren` instead of `readContext.read`, hence the name.

If someone want's to use read and pass read options - we need to call `readContext.read`, we can't just replace it with a call to `executeReadQueryAndResolveChildren` and ignore options. The interleaved children will have to be loaded lazily.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryExecutor.java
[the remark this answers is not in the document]
dmitry-s: yes, but you check if they are empty in lines 243-248
after that we know we have 2 parts and they are not empty

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryExecutor.java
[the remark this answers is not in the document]
dmitry-s: That's ok if the only element it's wrapped in parentheses. I'm ok with that if that's the price to pay for more readable code :)

```
  static String orJoin(List<String> parts){
    return parts.stream().map(s -> "("+s+")").collect(Collectors.joining(" OR "));
  }
.....
  System.out.println(orJoin(Arrays.asList()));
  System.out.println(orJoin(Arrays.asList("one")));
  System.out.println(orJoin(Arrays.asList("one", "two")));
```
gives this:
```

(one)
(one) OR (two)
```

It looks totally acceptable to me.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryExecutor.java
```diff
@@ -201,14 +209,15 @@ private SpannerStatementQueryExecutor() {
 	public static <T> Statement buildQuery(KeySet keySet,
 			SpannerPersistentEntity<T> persistentEntity, SpannerCustomConverter writeConverter,
 			SpannerMappingContext mappingContext) {
-		StringJoiner orJoiner = new StringJoiner(" OR ");
+		List<String> or = new ArrayList<>();
```
dmitry-s: rename to `orParts`

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryExecutor.java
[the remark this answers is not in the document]
s13o: it's better to use `Collectors.joining(") OR (", "(", ")")` instead of `map(s -> "("+s+")")`.

Ok, I'll follow your advice but I'll update some tests to satisfy additional brackets - I tried not to update any test that not related directly to my change

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryExecutor.java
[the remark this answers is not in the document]
s13o: will check it again, thanks

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
s13o: clear

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
s13o: ok, I gave a wrong explanation, sorry.
It is not related to interleaved fields.
What if the entity itself has a class annotation `@Where`?
In such a case, we can't use the `readContext.read` to fetch it but instead, we should build an SQL query with "SQL where condition"  from the annotation we have.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryExecutor.java
s13o: ok

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
dmitry-s: I see your point. 
In that case will need to modify the logic a little bit.

What we need to do is throw an exception when we have `@Interleved` (eager) children or `@Where` present and options are not null.

We can't just silently ignore the options if we can't pass them. Instead, we should let the user know that something is wrong.

Does it make sense?

s13o: Options? what options?
How they affect each other?
we have a code
```java
	private <T> boolean isEligibleForEagerFetch(KeySet keys, SpannerReadOptions options,
			SpannerPersistentEntity<T> persistentEntity) {
		return persistentEntity.hasEagerlyLoadedProperties() &&
				persistentEntity.findAnnotation(Where.class) == null &&
				options == null && !keys.getRanges().iterator().hasNext();
	}
```
do you propose to throw an exception here like this?
```java
...
		if (options!=null && (persistentEntity.hasEagerlyLoadedProperties() ||  persistentEntity.getWhere() != null )){
			throw new IllegalArgumentException("bla-bla-bla");
		}

		return persistentEntity.hasEagerlyLoadedProperties() &&
				options == null && !keys.getRanges().iterator().hasNext();
...
```

well, can we better convert the SpannerReadOptions to the SpannerQueryOption and throw an exception when it is impossible?

s13o: Another idea - the method `readContext.read` is very simple, It has completely another design and another idea with "query" methods and related feature-annotations (Interleaved, Where, whatever else). Can we just completely ignore these annotations for `readContext.read` - I mean for `SpannerOperations#read`? We should just stop the usage of this method in repositories

s13o: or we can design the `@Where` to be easily converted to the `KeySet`.
@Where` - is a new feature and we can do it as we like.

dmitry-s: There are 3 separate branches
1. no need for eager or `@Where` - we call `readContext.read`
2. we need eager or `@Where` and options are null - in this case we call `executeReadQueryAndResolveChildren`
3. we need eager or `@Where` and options are not null - we throw an exception

What do you think about that approach?

s13o: deal

meltsufin:
@s13o @dmitry-s If you guys want this in the next release going out Monday, we need this PR in together with the reference documentation by EOD.

s13o:
> @s13o @dmitry-s If you guys want this in the next release going out Monday, we need this PR in together with the reference documentation by EOD.

Sorry, I was too busy on Monday. Actually the PR was updated but without Javadoc, it has been just added :(

docs/src/main/asciidoc/spanner.adoc
```diff
@@ -420,6 +420,9 @@ If a property marked for lazy fetching is never retrieved, then it is also skipp
 
 If used inside a transaction, subsequent operations on lazily-fetched properties use the same transaction context as that of the original parent entity.
 
+===== Soft Delete
```
meltsufin: Soft Delete is just one application of this. The section should probably be called something like "Declarative Filtering with `@Where`"

docs/src/main/asciidoc/spanner.adoc
```diff
@@ -420,6 +420,9 @@ If a property marked for lazy fetching is never retrieved, then it is also skipp
...
+The `@Where` annotation could be applied to the entity class or to the interleaved property. This annotation provides an SQL where clause
```
meltsufin: Our conventions is one sentence per line.

docs/src/main/asciidoc/spanner.adoc
s13o: ok
Should I add more information like examples? For me, it is a trivial use-case that Hirberhane has for years.

docs/src/main/asciidoc/spanner.adoc
s13o: ok, thanks

docs/src/main/asciidoc/spanner.adoc
meltsufin: Yeah a small example of soft-delete would be great!

s13o: will add one

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryExecutor.java
```diff
@@ -162,6 +164,9 @@ private SpannerStatementQueryExecutor() {
 							.getPersistentProperty(o.getProperty());
 					return (property != null) ? property.getColumnName() : o.getProperty();
 				});
+		if (!persistentEntity.getWhere().isEmpty()) {
```
dmitry-s: This construction `persistentEntity.getWhere().isEmpty()` is used in multiple places.  Please create `hasWhere()` method to improve readability.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
```diff
@@ -209,14 +213,35 @@ public long executePartitionedDmlStatement(Statement statement) {
...
-				options == null && !keys.getRanges().iterator().hasNext();
+	/**
+	 * In many cases {@link KeySet} with {@link SpannerReadOptions} are compatible with
+	 * {@link SpannerReadOptions}. The method throws exception when it is impossible.
+	 * @param options read-parameters
+	 * @return query-parameters
+	 * @throws IllegalArgumentException when {@link SpannerQueryOptions} can't be converted to {@link SpannerQueryOptions}
+	 * 	or {@code keys} have "ranges".
+	 */
+	private static SpannerQueryOptions toQueryOption(KeySet keys, SpannerReadOptions options) throws IllegalArgumentException {
```
dmitry-s: Move this method to `SpannerQueryOptions` class and rename to `fromReadOptions`

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryExecutor.java
```diff
@@ -218,32 +250,45 @@ private SpannerStatementQueryExecutor() {
...
 		List<SpannerPersistentProperty> parentKeyProperties = parentPersistentEntity
 				.getFlattenedPrimaryKeyProperties();
-		String condition = parentKeyProperties.stream()
+		String keyCause = parentKeyProperties.stream()
 				.map(keyProp -> tableName + "." + keyProp.getColumnName()
 						+ " = "
 						+ parentPersistentEntity.tableName() + "." + keyProp.getColumnName())
 				.collect(Collectors.joining(" AND "));
-
+		String condition = combine(keyCause, whereCause);
```
dmitry-s: typos, should be `keyClause`, `whereClause`

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryExecutor.java
```diff
@@ -218,32 +250,45 @@ private SpannerStatementQueryExecutor() {
 				keyParts.add(parentKeyParts.next());
 				tagNum++;
 			}
-			orJoiner.add(andJoiner.toString());
+			orParts.add(andJoiner.toString());
 		}
-		String cond = orJoiner.toString();
+		String keyCause = orParts.stream().collect(
+				() -> new StringJoiner(") OR (", "(", ")").setEmptyValue(""),
```
dmitry-s: Can we use `orParts.stream().map(s -> "("+s+")").collect(Collectors.joining(" OR "))` instead?
That would be shorter and more readable.

s13o:
> Overall looks good, except for a couple of issues.
> 
> BTW, it looks like we are going to need to add`@SQLDelete` and possibly `@PreRemove` in order to have fully functional soft delete support.
> 
> I'm curious, are you planning to work on those?

Thank you for your comments -  they are incredibly helpful, as usual.
These two annotations are little out of my project's needs right now.
But in a common - it is very cool to have and I'm ready to participate.

Have you seen that I did a new method `SpannerQueryOptions toQueryOption(KeySet keys, SpannerReadOptions options)`?  It essentially changes the previous implementation. Are you OK with it?

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryExecutor.java
s13o: I thought I found a way to not wrap an item by parentheses when the stream has a single value only. But I was wrong. I'll switch to your implementation, thanks.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
dmitry-s: Another option is to move this method to `SpannerReadOptions`, so you could call it like this:
`options.toQueryOptions()`

dmitry-s:
@s13o Yes, I've seen it and left a comment. Looks good, but needs to me moved.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
s13o: Ah, well, clear

s13o:
> @s13o Yes, I've seen it and left a comment. Looks good, but needs to me moved.
Clear, thanks.
Unfortunately, I should switch to urgent google dataflow tasks for some time. #BrainSplitProgramming

meltsufin:
Hi @s13o, Dmitry is currently out of office, so it may take us a few days to get back to you.
Thanks for your patience.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/core/SpannerTemplate.java
```diff
@@ -253,10 +264,10 @@ public long executePartitionedDmlStatement(Statement statement) {
 	@Override
 	public <T> List<T> queryAll(Class<T> entityClass,
 			SpannerPageableQueryOptions options) {
-		SpannerPersistentEntity<?> persistentEntity = this.mappingContext
-				.getPersistentEntity(entityClass);
+		SpannerPersistentEntity<?> persistentEntity = this.mappingContext.getPersistentEntity(entityClass);
+		String condition = persistentEntity.hasWhere() ? " WHERE " + persistentEntity.getWhere() : "";
```
dmitry-s: I see this logic is repeated multiple times. We should introduce a method (let's call it `buildWhere`) that checks if a string is empty or null and if not, returns `" WHERE " + string`, otherwise returns `""`

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryExecutor.java
```diff
@@ -218,32 +250,44 @@ private SpannerStatementQueryExecutor() {
 				keyParts.add(parentKeyParts.next());
 				tagNum++;
 			}
-			orJoiner.add(andJoiner.toString());
+			orParts.add(andJoiner.toString());
 		}
-		String cond = orJoiner.toString();
+		String keyClause = orParts.size() == 1
+				? orParts.get(0) : orParts.stream().map(s -> "(" + s + ")").collect(Collectors.joining(" OR "));
+		String condition = combine(keyClause, whereClause != null ? whereClause : "");
```
dmitry-s: let's move this ` whereClause != null ? whereClause : ""` to the `combine` method and apply it to both arguments

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryExecutor.java
```diff
@@ -218,32 +250,44 @@ private SpannerStatementQueryExecutor() {
...
 						+ parentPersistentEntity.tableName() + "." + keyProp.getColumnName())
 				.collect(Collectors.joining(" AND "));
-
+		String condition = combine(keylCause, whereClause);
 		return "ARRAY (SELECT AS STRUCT " + getColumnsStringForSelect(childPersistentEntity, mappingContext, true) + " FROM "
-				+ tableName + " WHERE " + condition + ") as " + columnName;
+				+ tableName + " WHERE " + condition + ") AS " + columnName;
 	}
 
+	private static String combine(String cond1, String cond2) {
```
dmitry-s: rename to `combineWithAnd` for clarity

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryExecutor.java
```diff
@@ -218,32 +250,44 @@ private SpannerStatementQueryExecutor() {
 				keyParts.add(parentKeyParts.next());
 				tagNum++;
 			}
-			orJoiner.add(andJoiner.toString());
+			orParts.add(andJoiner.toString());
 		}
-		String cond = orJoiner.toString();
+		String keyClause = orParts.size() == 1
```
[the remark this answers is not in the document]
s13o: I do not agree with your argument. The code is already complex enough and this small update will not simplify it - it will just brake some IT tests. But, sure, I'll follow your advice.

s13o:
Anything else?

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryExecutor.java
```diff
@@ -201,14 +210,37 @@ private SpannerStatementQueryExecutor() {
 	public static <T> Statement buildQuery(KeySet keySet,
 			SpannerPersistentEntity<T> persistentEntity, SpannerCustomConverter writeConverter,
 			SpannerMappingContext mappingContext) {
-		StringJoiner orJoiner = new StringJoiner(" OR ");
+		return buildQuery(keySet, persistentEntity, writeConverter, mappingContext, persistentEntity.getWhere());
+	}
+
+	/**
+	 * Builds a query that returns the rows associated with a key set with additional SQL-where.
+	 * But the {@link org.springframework.cloud.gcp.data.spanner.core.mapping.Where} will be ignored.
```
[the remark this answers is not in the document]
s13o: the method is needed for better compatibility with the previous version. Probably someone already uses the previous method - it was "public static"

Java doc updated, check it, pls.

s13o:
Did I still have something missed?

s13o:
Thanks a lot!

PR #2222, opened 2020-02-27
Migrate 1.2.x version back to master branch

meltsufin:
This PR is effectively making the 1.2.x branch master.
The reason for this is that the current master (1.3.x) track is based on the Hoxton release train just like the 1.2.x branch. So, there is no reason for a version bump. The features we've been slating for 1.3.x will instead be released as part of 1.2.x service releases, starting with the upcoming 1.2.2 release.

The plan is to up the project version to 2.0 when we start developing for the Ilford release train later this year.


spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/core/GcpContextAutoConfiguration.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017-2018 the original author or authors.
```
elefeint: revert?

spring-cloud-gcp-samples/spring-cloud-gcp-integration-pubsub-sample/pom.xml
```diff
@@ -5,12 +5,12 @@
...
 		<artifactId>spring-cloud-gcp-samples</artifactId>
 		<groupId>org.springframework.cloud</groupId>
-		<version>1.3.0.BUILD-SNAPSHOT</version>
+		<version>1.2.2.BUILD-SNAPSHOT</version>
 	</parent>
 
 	<artifactId>spring-cloud-gcp-integration-pubsub-sample</artifactId>
 	<groupId>org.springframework.cloud</groupId>
-	<version>1.3.0.BUILD-SNAPSHOT</version>
+	<version>1.2.2.BUILD-SNAPSHOT</version>
```
elefeint: This line can be removed; it matches the parent version.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/core/GcpContextAutoConfiguration.java
meltsufin: seems more correct

PR #2223, opened 2020-02-28
Create AutoConfiguration for Secret Manager Template and extend sample

dzou:
This adds a non-bootstrap phase autoconfiguration class to autowire the `SecretManagerTemplate` and then extends the existing sample application to use the template class.

Contributes to #2176.

**Refdoc** is prepared in the PR here: #2224.

### DIFF
--- a/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/GcpSecretManagerTemplateConfiguration.java
+++ b/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/GcpSecretManagerTemplateConfiguration.java
@@ -21,25 +21,22 @@ import java.io.IOException;
 import com.google.api.gax.core.CredentialsProvider;
 import com.google.cloud.secretmanager.v1beta1.SecretManagerServiceClient;
 import com.google.cloud.secretmanager.v1beta1.SecretManagerServiceSettings;
-import com.google.protobuf.ByteString;
 
 import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
-import org.springframework.cloud.bootstrap.config.PropertySourceLocator;
 import org.springframework.cloud.gcp.core.DefaultCredentialsProvider;
 import org.springframework.cloud.gcp.core.DefaultGcpProjectIdProvider;
 import org.springframework.cloud.gcp.core.GcpProjectIdProvider;
 import org.springframework.cloud.gcp.core.UserAgentHeaderProvider;
+import org.springframework.cloud.gcp.secretmanager.SecretManagerTemplate;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
-import org.springframework.core.convert.converter.Converter;
-import org.springframework.core.env.ConfigurableEnvironment;
 
 /**
- * Bootstrap Autoconfiguration for GCP Secret Manager which enables loading secrets as
- * properties into the application {@link org.springframework.core.env.Environment}.
+ * Autoconfiguration for GCP Secret Manager which provides an instance of the
+ * {@link SecretManagerTemplate}.
  *
  * @author Daniel Zou
  * @since 1.2.2
@@ -47,8 +44,8 @@ import org.springframework.core.env.ConfigurableEnvironment;
 @Configuration
 @EnableConfigurationProperties(GcpSecretManagerProperties.class)
 @ConditionalOnClass(SecretManagerServiceClient.class)
-@ConditionalOnProperty(value = "spring.cloud.gcp.secretmanager.bootstrap.enabled", matchIfMissing = true)
-public class GcpSecretManagerBootstrapConfiguration {
+@ConditionalOnProperty(value = "spring.cloud.gcp.secretmanager.template.enabled", matchIfMissing = true)
+public class GcpSecretManagerTemplateConfiguration {
 
 	private final GcpSecretManagerProperties properties;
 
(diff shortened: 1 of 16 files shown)


spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/GcpSecretManagerTemplateConfiguration.java
```diff
@@ -0,0 +1,82 @@
...
+ * {@link SecretManagerTemplate}.
+ *
+ * @author Daniel Zou
+ * @since 1.2.2
+ */
+@Configuration
+@EnableConfigurationProperties(GcpSecretManagerProperties.class)
+@ConditionalOnClass(SecretManagerServiceClient.class)
+@ConditionalOnProperty(value = "spring.cloud.gcp.secretmanager.template.enabled", matchIfMissing = true)
+public class GcpSecretManagerTemplateConfiguration {
```
meltsufin: I'm having second thoughts on the separation of the template from the bootstrap configuration. The main reason being that we wanted to re-use the template in the property source implementation, and currently we don't. If we do create the `SecretManagerTemplate` in the boostrap, does it remain accessible to the application context? If so, I'm not sure of what the benefit would be of this separate configuration class.

meltsufin: Also, do we want to support the use-case where someone doesn't want to use bootstrap properties but wants to use the template?
Since loading secrets into the application context inadvertently may be risky, maybe we should even disable it by default, and the only thing you get with the starter by default, would be the template.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/GcpSecretManagerTemplateConfiguration.java
```diff
@@ -0,0 +1,82 @@
...
+
+	private final GcpSecretManagerProperties properties;
+
+	private final CredentialsProvider credentialsProvider;
+
+	private final GcpProjectIdProvider gcpProjectIdProvider;
+
+	public GcpSecretManagerTemplateConfiguration(
```
elefeint: `GcpSecretManagerTemplateConfiguration` and `GcpSecretManagerBootstrapConfiguration` end up being somewhat similar. 
I guess it's related to Mike's comment above, but possibly we should have one autoconfiguration with an optional conversion bean/property source additions that are blocked by an extra, off-by-default, property?

Is there any harm in doing regular bean initialization during bootstrap phase if someone does opt for template-only without enabling property initialization?

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/GcpSecretManagerTemplateConfiguration.java
dzou: Right, I created two separate configurations in order to support the use-case where someone doesn't want to use bootstrap properties but wants to use the template.

If you want the template in the property source impl, I can add an additional `@ConditionalOnMissingBean SecretManagerTemplate` to the bootstrap config. Then disable the bootstrap config by default?

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/GcpSecretManagerTemplateConfiguration.java
dzou: Ohh I see, yes we can keep all beans in one configuration. I just created two because I wasn't sure if you could activate/deactivate beans by property as well.

So I could add a `@ConditionalOnProperty` to a bean definition?

spring-cloud-gcp-samples/spring-cloud-gcp-secretmanager-sample/src/main/resources/templates/index.html
```diff
@@ -0,0 +1,98 @@
...
+  <link type="text/css" href="css/style.css" rel="stylesheet"/>
+</head>
+
+<body>
+<h1>Secret Manager Demo with Spring Cloud GCP</h1>
+
+<div class="panel">
+  <h3>Secret Manager Property Source</h3>
+  At the bootstrap phase, we loaded the following secret into the application context:
+  <br/>application-secret: [[${applicationSecret}]]
```
meltsufin: The secret blends into the sentence. Maybe emphasize it a bit more?
```suggestion
  <br/><b>application-secret:</b> <i>[[${applicationSecret}]]</i>
```

dzou:
@meltsufin @elefeint - Discussed with @meltsufin and we decided to go with the 2 config approach because if a user disables the bootstrap then we would want them to acquire the template instance at non-bootstrap time. This is enabled with the `spring.cloud.gcp.secretmanager.bootstrap.enabled` setting which defaults to false.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/GcpSecretManagerTemplateConfiguration.java
dzou: Discussed offline with @meltsufin; see: https://github.com/spring-cloud/spring-cloud-gcp/pull/2223#issuecomment-593570686

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/GcpSecretManagerTemplateConfiguration.java
dzou: Resolved offline; see https://github.com/spring-cloud/spring-cloud-gcp/pull/2223#issuecomment-593570686

spring-cloud-gcp-samples/spring-cloud-gcp-secretmanager-sample/src/main/java/com/example/SecretManagerWebController.java
```diff
@@ -18,30 +18,50 @@
...
+	@ResponseBody
+	public String getSecret(@RequestParam String secretId, ModelMap map) {
+		String secretPayload = this.secretManagerTemplate.getSecretString(secretId);
+		return "Secret ID: " + secretId + " | Value: " + secretPayload;
+	}
+
+	@PostMapping("/createSecret")
+	public String createSecret(@RequestParam String secretId, @RequestParam String secretPayload) {
+		this.secretManagerTemplate.createSecret(secretId, secretPayload);
+		return "redirect:/";
```
meltsufin: The redirect makes it hard to tell that something actually happened.
Maybe just return a message like "Secret created".

You can of course also just use the `ModelAndView` with an additional "message" argument for this endpoint and the one above, but that's a bit more fancy.

spring-cloud-gcp-samples/spring-cloud-gcp-secretmanager-sample/src/main/resources/templates/index.html
dzou: Done.

spring-cloud-gcp-samples/spring-cloud-gcp-secretmanager-sample/src/main/java/com/example/SecretManagerWebController.java
dzou: Done. Went with the additional param approach.

PR #2226, opened 2020-03-02
Making Pub/Sub reactive pull non-blocking

mzeijen:
_Please note that this pull request is a proposal and it is not ready yet to be merged._ 

The current `PubSubReactiveFactory` is based on blocking calls to `PubSubTemplate`, to be able to reactively pull messages from Pub/Sub. This isn't as efficient as it could be. Because, the Google Pub/Sub client does offer the ability to do non-blocking asynchronous pulling off messages. However, the current PubSubTemplate doesn't offer any pull methods which use this ability and thus the `PubSubReactiveFactory` can't make use of it.

In this draft pull request I propose changes that makes the `PubSubReactiveFactory.poll(...)` non-blocking, without breaking its backwards compatibility (except for the exception handling, see note below). I make this possible by adding a `PubSubTemplate.pullFuture` method, which returns a `ListenableFuture` instead of blocking while it requests the messages.

I did some testing with this code against the Pub/Sub service, processing millions of messages, and comparing it to the `blocking` variant. It doesn't seem to have any performance downside. It does reduces the number of threads that need to be used (as there is one less thread pool that comes in to play), and doesn't do any blocking of threads.

If you think this change makes sense, then I will be happy to finish the work so that it is ready to be merged. If you have any concerns then I will happy to discuss them.

The following things are currently missing or may need to change, before this pull request is ready:

- I am not a big fan of the method name `pullFuture`. It is not in line with the other methods, but I need a different name for `pull` as it gives a different return type. The `future` in the name does clearly state what you will get...
- There are other `pull` related methods in `PubSubTemplate` for which a version should be offered that return a `ListenableFuture` and that don't block.
- In `PubSubTemplate.pullFuture` the `ApiFutures.addCallback` call now uses the `MoreExecutors.directExecutor()` executor. Because if it, the callback will be handled by the same thread that handles the response from the PubSub client.  I am wondering if this is ok or that we should make this configurable like it is already configurable for the `ack`, `nack` and `modifyAckDeadline` methods. I don't see much benefit in having another thread pool that must be managed by default, though. If it needs to go into a different thread pool then it would be great if we could use something generic, for CPU bound processing, like the Reactor `parallel` thread pool. Something to discuss.

Note: In the current implementation of the PubSubReactiveFactory, what I can see and what I have tested, the implementation doesn't correctly deal with exceptions being thrown by the `PubSubTemplate` or the Google PubSub client. The exceptions aren't passed upstream to the sink. Also, when doing a backpressure based poll, an exception will result in a "dead" flux, as any exception in the BlockingLimitedDemandPullTask will kill the task and no messages will ever be put in the sink anymore. The Flux doesn't know that it is dead and will simply keep on waiting for messages that will never arrive. This is because the BlockingLimitedDemandPullTask doesn't catch and forward the exception to the sink. 
In the implementation of this pull request, we forward all exceptions to the sink. A `retry` can then be used on the `Flux`, returned by `poll`, to make sure exceptions don't result in a dead stream.
The issue in the current implementation `PubSubReactiveFactory` should probably be solved independent of this pull request, if there is a chance that this pull request will take a long time or will never be merged. I can create a ticket for it and a separate pull request, if desired.


pivotal-issuemaster:
@mzeijen Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-gcp&pullRequestId=2226)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-cloud/spring-cloud-gcp&pullRequestId=2226) to manually synchronize the status of this Pull Request.

mzeijen:
Oops, accidentally didn't create it as a draft pull request.

PR #2227, opened 2020-03-02
Making Pub/Sub reactive pull non-blocking

mzeijen:
_Please note that this pull request is a proposal and it is not ready yet to be merged._ 

The current `PubSubReactiveFactory` is based on blocking calls to `PubSubTemplate`, to be able to reactively pull messages from Pub/Sub. This isn't as efficient as it could be. Because, the Google Pub/Sub client does offer the ability to do non-blocking asynchronous pulling off messages. However, the current PubSubTemplate doesn't offer any pull methods which use this ability and thus the `PubSubReactiveFactory` can't make use of it.

In this draft pull request I propose changes that makes the `PubSubReactiveFactory.poll(...)` non-blocking, without breaking its backwards compatibility (except for the exception handling, see note below). I make this possible by adding a `PubSubTemplate.pullFuture` method, which returns a `ListenableFuture` instead of blocking while it requests the messages.

I did some testing with this code against the Pub/Sub service, processing millions of messages, and comparing it to the `blocking` variant. It doesn't seem to have any performance downside. It does reduces the number of threads that need to be used (as there is one less thread pool that comes in to play), and doesn't do any blocking of threads.

If you think this change makes sense, then I will be happy to finish the work so that it is ready to be merged. If you have any concerns then I will happy to discuss them.

The following things are currently missing or may need to change, before this pull request is ready:

- I am not a big fan of the method name `pullFuture`. It is not in line with the other methods, but I need a different name for `pull` as it gives a different return type. The `future` in the name does clearly state what you will get...
- There are other `pull` related methods in `PubSubTemplate` for which a version should be offered that return a `ListenableFuture` and that don't block.
- In `PubSubTemplate.pullFuture` the `ApiFutures.addCallback` call now uses the `MoreExecutors.directExecutor()` executor. Because if it, the callback will be handled by the same thread that handles the response from the PubSub client.  I am wondering if this is ok or that we should make this configurable like it is already configurable for the `ack`, `nack` and `modifyAckDeadline` methods. I don't see much benefit in having another thread pool that must be managed by default, though. If it needs to go into a different thread pool then it would be great if we could use something generic, for CPU bound processing, like the Reactor `parallel` thread pool. Something to discuss.

Note: In the current implementation of the PubSubReactiveFactory, what I can see and what I have tested, the implementation doesn't correctly deal with exceptions being thrown by the `PubSubTemplate` or the Google PubSub client. The exceptions aren't passed upstream to the sink. Also, when doing a backpressure based poll, an exception will result in a "dead" flux, as any exception in the BlockingLimitedDemandPullTask will kill the task and no messages will ever be put in the sink anymore. The Flux doesn't know that it is dead and will simply keep on waiting for messages that will never arrive. This is because the BlockingLimitedDemandPullTask doesn't catch and forward the exception to the sink. 
In the implementation of this pull request, we forward all exceptions to the sink. A `retry` can then be used on the `Flux`, returned by `poll`, to make sure exceptions don't result in a dead stream.
The issue in the current implementation `PubSubReactiveFactory` should probably be solved independent of this pull request, if there is a chance that this pull request will take a long time or will never be merged. I can create a ticket for it and a separate pull request, if desired.


pivotal-issuemaster:
@mzeijen Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-gcp&pullRequestId=2227)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-cloud/spring-cloud-gcp&pullRequestId=2227) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@mzeijen Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-gcp&pullRequestId=2227)!

elefeint:
Yes, please file a separate issue for the error propagation bug, sorry about that!

The naming convention could be `pullAsync`; it's a common convention.

As far as Guava, Spring projects try to stay away from using it due to having had bad experiences with versioning/api evolution. Now that Google libraries come with a BOM, this should be less of an issue, though, so perhaps we can consider it.

mzeijen:
@elefeint @meltsufin Great that you like it. I will finish this contribution than with the suggested changes.

I'll create a separate issue for the error propagation bug then. I'll probably do a merge request for it. 

Regarding Guava and the `MoreExecutors` usage: I understand the reluctance in using it. We do the same for our internal libraries and frameworks. However, the `ApiFutures.addCallback` already heavily relies on it, so I thought it shouldn't be a big deal.

Regarding the configurable executor: what should the default executor be in that case? Could it still be the `MoreExecutors.directExecutor()`? The thread with which the flux is then executed is the `grpc-default-executor`. I could also set the `publisherExecutorProvider` as the default.

mzeijen:
The bug report for the exception handling is https://github.com/spring-cloud/spring-cloud-gcp/pull/2227.

meltsufin:
@mzeijen Thanks for the error-handling fix.

Regarding executors, for the default, we can definitely use `MoreExecutors.directExecutor()` or `Executor executor = Runnable::run` to avoid Guava imports. We just want to make sure it's also configurable.

mzeijen:
That is a great tip to use the `Executor executor = Runnable::run` instead of `MoreExecutors.directExecutor()`. I hadn't really looked at the implementation of the `directExecutor` and didn't know how simple it actually is! I'll go for the `Runnable::run` then, as it saves us on a checkstyle suppression.

mzeijen:
I believe the pull request is ready now for the thorough review and for merging if it is ok as it is now. 

What I did is:

- all pull related methods in the `PubSubSubscriberOperations` now have async versions, with the names ending with `Async`. All methods have corresponding unit tests.
- the executor, used for processing the async callbacks of those pull methods, is now configurable via the `setAsyncPullExecutor`. The autoconfiguration has been updated so you can configure this by providing the `pubSubAsynchronousPullExecutor` bean. By default `Runnable::run` is used as implementation.
- I refactored the `PubSubReactiveFactory` a bit and added more javadoc.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberTemplate.java
```diff
@@ -173,13 +210,56 @@ public Subscriber subscribe(String subscription,
...
 
 		PullResponse pullResponse = this.subscriberStub.pullCallable().call(pullRequest);
-		return pullResponse.getReceivedMessagesList().stream()
-						.map((message) -> new PulledAcknowledgeablePubsubMessage(
-								PubSubSubscriptionUtils.toProjectSubscriptionName(pullRequest.getSubscription(),
-										this.subscriberFactory.getProjectId()),
-								message.getMessage(),
-								message.getAckId()))
-						.collect(Collectors.toList());
+		return toAcknowledgeablePubsubMessageList(
```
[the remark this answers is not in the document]
mzeijen: Good catch. I applied your suggestion.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
```diff
@@ -153,6 +153,12 @@ public PubSubPublisherTemplate pubSubPublisherTemplate(PublisherFactory publishe
 		return pubSubPublisherTemplate;
 	}
 
+	@Bean
+	@ConditionalOnMissingBean(name = "pubSubAsynchronousPullExecutor")
```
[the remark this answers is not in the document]
mzeijen: When writing this I was in doubt what to do. Normally I would also use an `Optional` or `ObjectProvider`. But somehow I thought that wouldn't be in line with how this autoconfiguration class worked. Looking at it again now, I see many usages of `ObjectProvider`... 
So, I applied your suggestion, but instead of using an `Optional` I used a `ObjectProvider`, as it is used all over the place, instead of `Optional`. So, its more in line with the rest of the class.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubReactiveAutoConfiguration.java
```diff
@@ -49,30 +47,14 @@
...
 
-		Scheduler scheduler = null;
-		if (userProvidedScheduler.isPresent()) {
-			scheduler = userProvidedScheduler.get();
-		}
-		else {
-			this.defaultPubSubReactiveScheduler = Schedulers.newElastic("pubSubReactiveScheduler");
-			scheduler = this.defaultPubSubReactiveScheduler;
-		}
+		Scheduler scheduler = userProvidedScheduler.orElseGet(() -> Schedulers.parallel());
```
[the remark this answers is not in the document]
mzeijen: The reason why I opted for using `Schedulers.parallel()` is because it is optimised for non-blocking executions and this new code it is fully non-blocking, as you noted. 

I didn't use a named scheduler because IMHO in a non-blocking reactive environment, it is bad practice to use separate thread pools for separate tasks. Building reactive applications is about getting the most out of system resources. Separate thread pools would go against that. Having many more threads, meant for CPU work, then CPU cores is waste because they aren't doing anything for most of the time. They only cost memory and create more context-switching overhead.

At my company we have some applications that have thousands of threads because of the practice of using separate named thread pools for different tasks. Those applications are reaching their limits due to this and it is actually a problem.

I do understand your concern with regards to trouble shooting the bugs you describe. Without having named thread pools, it sometimes can be really difficult to figure out where the culprit is. But, TBH, in the reactive world dealing with those kind of issues is already especially difficult due to the nature of reactive programming. It would be nice if there was a "reactor native way" to dynamically enhance the name of thread, to add troubleshooting meta information...

These are just my 2 cents. I am not a specialist and maybe I am wrong about how important it is to do these kinds of optimisations. So, I will defer to your opinion on this. In the end the scheduler is configurable, so users can always optimise this however they want it to be.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/reactive/PubSubReactiveFactory.java
```diff
@@ -77,116 +86,62 @@ public PubSubReactiveFactory(PubSubSubscriberOperations subscriberOperations, Sc
...
+		this.subscriberOperations.pullAsync(subscriptionName, intDemand, false).addCallback(
+				messages -> {
+					if (!sink.isCancelled()) {
+						messages.forEach(sink::next);
 					}
-					catch (DeadlineExceededException e) {
+					if (!sink.isCancelled()) {
+						long numToPull = numRequested - messages.size();
+						if (numToPull > 0) {
+							backpressurePull(subscriptionName, numToPull, sink);
```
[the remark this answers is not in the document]
mzeijen: Indeed, this method seems to be recursive, but it is not because the method calls itself within the context of a different thread than itself was executed. There is no thread stack build up.

Regarding your second point: is this actually a problem? PubSub should be able to handle it just fine. It is made for concurrent outstanding calls. It even recommends to do concurrent calls, to be able to get the highest throughput. It's just a bit unpredictable how many outstanding calls there can be at any given time.

If this is a problem then I think we can change this so that it is guaranteed that there is only on process retrieving the messages at any give time. When a new request comes in for new messages, it will then pass this on to this process and it will add it to the existing number of other messages that it already tries to pull. There is probably an elegant way of doing this in reactor...

spring-cloud-gcp-pubsub/src/test/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberTemplateTests.java
```diff
@@ -158,21 +163,33 @@ public void setUp() {
...
+		when(this.pullCallable.futureCall(any(PullRequest.class))).thenReturn(this.pullApiFuture);
+
+		doAnswer((invocation) -> {
+			Runnable runnable = invocation.getArgument(0);
+			runnable.run();
+			return null;
+		}).when(this.pullApiFuture).addListener(any(Runnable.class), any(Executor.class));
+		when(this.pullApiFuture.isDone()).thenReturn(true);
+		when(this.pullApiFuture.get()).thenReturn(PullResponse.newBuilder()
+				.addReceivedMessages(ReceivedMessage.newBuilder().setMessage(this.pubsubMessage).build()).build());
```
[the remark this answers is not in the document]
mzeijen: Fixed

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubReactiveAutoConfiguration.java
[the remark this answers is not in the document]
mzeijen: Never mind about what I said in the post above, because it isn't that relevant in this case. After look at it again, I noticed that the  `Schedulers.parallel()` scheduler is only used for doing the interval. It is not being used to handle the future, and the resulting Flux, because that is handled by the PubSub client thread that received the response. The `PubSubSubscriberTemplate.asyncPullExecutor` can be used to change the thread that handles that.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
```diff
@@ -167,10 +167,12 @@ public Executor pubSubAcknowledgementExecutor() {
 	@ConditionalOnMissingBean
 	public PubSubSubscriberTemplate pubSubSubscriberTemplate(SubscriberFactory subscriberFactory,
 			ObjectProvider<PubSubMessageConverter> pubSubMessageConverter,
-			@Qualifier("pubSubAcknowledgementExecutor") Executor executor) {
+			@Qualifier("pubSubAsynchronousPullExecutor") ObjectProvider<Executor> asyncPullExecutor,
+			@Qualifier("pubSubAcknowledgementExecutor") Executor ackExecutor) {
```
meltsufin: If we're using the `ObjectProvider` for the `asyncPullExecutor`, why doesn't it apply to `ackExecutor`?
Both approaches might be valid, but we should try to be consistent.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubReactiveAutoConfiguration.java
```diff
@@ -49,30 +47,14 @@
...
 		return new PubSubReactiveFactory(subscriberTemplate, scheduler);
 	}
 
-	@PreDestroy
-	public void closeScheduler() {
-		if (this.defaultPubSubReactiveScheduler != null) {
-			this.defaultPubSubReactiveScheduler.dispose();
```
meltsufin: Does our new default `Scheduler.parallel()` scheduler not need closing anymore?

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberTemplate.java
```diff
@@ -214,19 +314,63 @@ public Subscriber subscribe(String subscription,
...
+		if (maxMessages != null) {
+			Assert.isTrue(maxMessages > 0, "The maxMessages must be greater than 0.");
+		}
+
+		PullRequest pullRequest = this.subscriberFactory.createPullRequest(
+				subscription, maxMessages, returnImmediately);
+
+		final SettableListenableFuture<List<PubsubMessage>> settableFuture = new SettableListenableFuture<>();
+
+		this.pullAsync(pullRequest).addCallback(
```
meltsufin: All of the code above this line in the method is shared with `pullAndAck`. Can we try to reduce the code duplication?

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/reactive/PubSubReactiveFactory.java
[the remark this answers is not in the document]
meltsufin: Look fine to me. Good point regarding two different threads being involved which avoids the stack buildup.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
mzeijen: I did not change that into a `ObjectProvider` because there is always a `pubSubAcknowledgementExecutor` bean in the spring context (see the `pubSubAcknowledgementExecutor()` bean definition method above this method). So, having a  `ObjectProvider` is pointless as it will always have a bean. Agree?

Side note: you can wonder why we need the `pubSubAcknowledgementExecutor` bean at all, and why the default ackExecutor, defined in `PubSubSubscriberTemplate`, doesn't suffice. Currently it is a `Executors.newSingleThreadExecutor()`, which doesn't scale, I guess, and that is why the `pubSubAcknowledgementExecutor` bean defines a ThreadPoolTaskExecutor that can have more threads, depending on the `...subscriber.max-acknowledgement-threads` property. 

We could also just change the default `Executors.newSingleThreadExecutor()` in the `PubSubSubscriberTemplate` in to  a `Runnable::run` and the `pubSubAcknowledgementExecutor` bean could only be created if a specific `...subscriber.max-acknowledgement-threads` is defined in the properties (that way we stay somewhat backwards compatible). That way it is  in line how we do callback processing for both ack responses and message pull responses. I haven't done any such changes though, as that would be a big change in how the PubSubSubscriberTemplate works by default for ack responses. Also it is unrelated to my ticket.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubReactiveAutoConfiguration.java
mzeijen: The `Schedulers.parallel()` is a global scheduler from the Reactor library and it has the responsibility to manage its lifecycle and close it when it is appropriate. Or at least something that is managing the Reactor library is responsible for all that.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubReactiveAutoConfiguration.java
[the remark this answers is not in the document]
bsideup: @elefeint if you make it non-blocking then a custom scheduler here does not make sense at all and can be removed (`interval` will implicitly use `Schedulers.parallel()`) :)

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberTemplate.java
mzeijen: I moved the validation into the `SubscriberFactory.createPullRequest`, where the `subscription` validation also already was done. This removed the duplication.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/DefaultSubscriberFactory.java
```diff
@@ -227,6 +227,8 @@ public PullRequest createPullRequest(String subscriptionName, Integer maxMessage
 						PubSubSubscriptionUtils.toProjectSubscriptionName(subscriptionName, this.projectId).toString());
 
 		if (maxMessages != null) {
```
mzeijen: Side note: I think it is a mistake that maxMessage can be set to `null`, because according to the Google PubSub API docs it must always have a positive number (see: https://cloud.google.com/pubsub/docs/reference/rpc/google.pubsub.v1#pullrequest). I believe that if it isn't set, then a default value of `0` is used.

Shall I report it as a bug?

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/reactive/PubSubReactiveFactory.java
```diff
@@ -77,116 +86,62 @@ public PubSubReactiveFactory(PubSubSubscriberOperations subscriberOperations, Sc
...
 	}
 
-	/**
-	 * Runnable task issuing blocking Pub/Sub Pull requests until the specified number of
-	 * messages has been retrieved.
-	 */
-	private class BlockingLimitedDemandPullTask extends PubSubPullTask {
+	private Flux<AcknowledgeablePubsubMessage> pullAll(String subscriptionName) {
+		CompletableFuture<List<AcknowledgeablePubsubMessage>> pullResponseFuture = this.subscriberOperations
+				.pullAsync(subscriptionName, Integer.MAX_VALUE, true).completable();
```
mzeijen: side note: use `true` for the third argument, defines the `return_immediately` as `true`. However according to the Google PubSub API docs, this is not recommended and the argument is actually deprecated (see https://cloud.google.com/pubsub/docs/reference/rpc/google.pubsub.v1#pullrequest). 

What to do with it? Separate bug?

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
meltsufin: I think you're right, this raises more questions about the introduction of the `pubSubAcknowledgementExecutor` bean rather than the use of the `ObjectProvider` for `asyncPullExecutor`. The change was made in #1349 over a year ago and the `...subscriber.max-acknowledgement-threads` was never really documented in the refdocs. So, it might be OK to just remove it, and replace with the `Runnable::run` default in the template itself. However, it looks like there were some performance issues with the direct executor. In any case, like you said, it's irrelevant to your ticket. @elefeint Should we file a separate ticket to investigate the `pubSubAcknowledgementExecutor`?

elefeint: Yes, let's create a different issue, and I agree that using `Runnable::run` by default is the right thing to do here.

elefeint: #2245

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubReactiveAutoConfiguration.java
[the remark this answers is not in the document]
elefeint: Excellent, thank you, this resolves my concern. It seems I have to unlearn the habit of creating a custom scheduler for every dedicated task.
The implication of `Flux.interval(Duration, Duration)` using `Schedulers.parallel()` by default is that we don't need a custom scheduler in `PubSubReactiveFactory` at all. On the other hand, it does not hurt anything to have extra flexibility in `pubSubReactiveScheduler` that the client application could control through overriding beans. Clearly I have not unlearned the habit of custom schedulers yet...

#2246 to track; this PR is large enough already.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/reactive/PubSubReactiveFactory.java
elefeint: Thank you for catching this; we actually did not realize this field recently got deprecated.
Let's keep the behavior consistent with the past in this PR.

#2247 to track.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/DefaultSubscriberFactory.java
elefeint: Yes, please. Zero is disallowed, too, so we should assert that the parameter is neither null nor zero.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/SubscriberFactory.java
```diff
@@ -53,7 +53,8 @@
 	 * Create a {@link PullRequest} for synchronously pulling a number of messages from
 	 * a Google Cloud Pub/Sub subscription.
 	 * @param subscriptionName the name of the subscription
-	 * @param maxMessages the maximum number of pulled messages
+	 * @param maxMessages the maximum number of pulled messages,
+	 * which can be null or must be a positive number
```
elefeint: I'd skip the "can be null". It's a required field in the proto, so the request will fail. You are right that our validation exempting `null` is a bug.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/DefaultSubscriberFactory.java
mzeijen: I created the bug #2248

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/support/SubscriberFactory.java
mzeijen: Ok, I removed it. Bug was created under #2248.

mzeijen:
I added myself as author to all classes I changed, correct the @since and made sure all copyright years are correctly set now. Should be good now.

mzeijen:
Awesome that we got this feature merged! My colleagues and I really appreciate this. 

Also kudos for the way you handle these kinds of contributions. Although it took a while, mostly probably due to the time difference, it was a very pleasant process. I really appreciate the positive attitude with the constructive feedback. This makes me want to do more contributions :).

Thanks!

elefeint:
Thank you -- this was really open source at its best; you've invested a lot of time but now all reactive Pub/Sub users will benefit from the lower overhead.

This change will make it into the next Spring Cloud patch release, currently scheduled for [mid-May](https://github.com/spring-cloud/spring-cloud-release/milestone/81). 

If you or your team would like to suggest features or discuss your use-cases for any GCP/Spring integrations (or Spanner/Hibernate or Spanner/R2DBC), we can set up a video conference. I'll ping the team e-mail over gitter.

mzeijen:
@elefeint Great! I will contact you if we want to take you up for your offer.

PR #2251, opened 2020-03-16
Interpret missing maxMessages as MAX_INT in Pub/Sub PullRequests

### LINKED ISSUE #2248: The maxMessage argument in the PubSubTemplate pull method must be a positive number
The PubSubTemplate pull methods have a nullable maxMessage argument, which allows `null`. However, according to the Google PubSub API docs the pull request maxMessage must always have a positive number (see: https://cloud.google.com/pubsub/docs/reference/rpc/google.pubsub.v1#pullrequest). 
Currently if null is provided via PubSubTemplate then 0 is send to PubSub, which is wrong.

dzou:
If the `maxMessages` parameter is provided as `null` in the PubSubTemplate, interpret this as `Integer.MAX_INTEGER` which means there should be no limit on the number of messages pulled.

Fixes #2248.

### DIFF
--- a/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberOperations.java
+++ b/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberOperations.java
@@ -86,8 +86,8 @@ public interface PubSubSubscriberOperations {
 	 * Pull and auto-acknowledge a number of messages from a Google Cloud Pub/Sub subscription.
 	 * @param subscription canonical subscription name, e.g., "subscriptionName", or the fully-qualified
 	 * subscription name in the {@code projects/<project_name>/subscriptions/<subscription_name>} format
-	 * @param maxMessages the maximum number of pulled messages. If this value is null then there is no limit
-	 * on the number of pulled messages.
+	 * @param maxMessages the maximum number of pulled messages. If this value is null then
+	 * up to Integer.MAX_VALUE messages will be requested.
 	 * @param returnImmediately returns immediately even if subscription doesn't contain enough
 	 * messages to satisfy {@code maxMessages}
 	 * @return the list of received messages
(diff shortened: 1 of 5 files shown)


spring-cloud-gcp-pubsub/src/test/java/org/springframework/cloud/gcp/pubsub/support/DefaultSubscriberFactoryTests.java
```diff
@@ -70,4 +71,19 @@ public void testNewDefaultSubscriberFactory_nullProject() {
...
+	@Test
+	public void testCreatePullRequest_nonZeroMaxMessages() {
+		DefaultSubscriberFactory factory = new DefaultSubscriberFactory(() -> "project");
+		factory.setCredentialsProvider(this.credentialsProvider);
+
+		// If a maxMessages is null (it was omitted), then set max to MAX_INT
+		PullRequest request = factory.createPullRequest("test", null, true);
+		assertThat(request.getMaxMessages()).isEqualTo(Integer.MAX_VALUE);
+
+		// If maxMessages < 0, should throw an error.
```
elefeint: This is a nit, but can this be a different test? Null and negative numbers are distinct scenarios.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberOperations.java
```diff
@@ -86,7 +86,8 @@
 	 * Pull and auto-acknowledge a number of messages from a Google Cloud Pub/Sub subscription.
 	 * @param subscription canonical subscription name, e.g., "subscriptionName", or the fully-qualified
 	 * subscription name in the {@code projects/<project_name>/subscriptions/<subscription_name>} format
-	 * @param maxMessages the maximum number of pulled messages
+	 * @param maxMessages the maximum number of pulled messages. If this value is null then there is no limit
```
elefeint: Would it make sense to replace "there is no limit" with "up to `Integer.MAX_VALUE` messages will be requested"? I know that you documented it in refdoc, but for people who read javadoc it would be better to be explicit.

spring-cloud-gcp-pubsub/src/test/java/org/springframework/cloud/gcp/pubsub/support/DefaultSubscriberFactoryTests.java
dzou: Done.

spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberOperations.java
dzou: Done.

PR #2264, opened 2020-03-20
Fix build after dependency upgrade

dzou:
Fixes the build after #2256 .

Resolves some errors related to mockito version upgrade and also spring security upgrade.


dzou:
Hmm still something weird with the dependencies... lets give it another day or two, maybe still something not completely right on the Spring side.

dzou:
I think there is a bug in spring-cloud-stream which is why our emulator tests are breaking; will followup with them and post the issue here.

dzou:
Filed https://github.com/spring-cloud/spring-cloud-stream-starters/issues/103 to ask if it's possible to upgrade Ivyland dependency version to 3.1.0.B-S.

dzou:
Also filed https://github.com/spring-cloud/spring-cloud-gcp/issues/2269, I will remove the version number override that I hardcoded for the spring-cloud-stream-test dependency once the version in the spring-cloud-stream bom is upgraded to `3.1.0.B-S` instead of 3.0.4

PR #2265, opened 2020-03-20
Use correct channel in Firestore autoconfiguration

### LINKED ISSUE #2254: FirestoreReactiveRepository yields GRPC-errors after enabling Trace
**Describe the bug**
Spring boot version: 2.2.5
Spring Cloud version: Hoxton.SR3

Queries via FirestoreReactiveRepository results in error with root cause "UNIMPLEMENTED: GRPC target method can't be resolved" if using trace. Enabling/disabling trace in configuration will change between expected/normal behavior and this error. 

Expected behavior would be to get the document back and tracing-information to be present on the google cloud project. 

**Sample**
Simple sample included, hopefully all that is needed to reproduce.
[tracingfirestore.zip](https://github.com/spring-cloud/spring-cloud-gcp/files/4349732/tracingfirestore.zip)

elefeint:
Fixes #2254.


elefeint:
I don't think so; trace triggering it was just incidental -- any dependency with its own grpc channel that happens to be autoconfigured earlier would trigger it.

I'll wait for the build to be green before merging.

Issue #2266, opened 2020-03-21
The `@Where` is applied in wrong order at SpannerStatementQueryExecutor#applySortingPagingQueryOptions


s13o:
**Describe the bug**
When an Entity has `@Where` and fetched from `@Query` with `Sort` or `Pageable` parameters the `@Where` will be applied after `ORDER BY` and the SpannerStatementQueryExecutor will generate a wrong SQL query

**Sample**
```java
	@Query("SELECT * FROM :org.springframework.cloud.gcp.data.spanner.test.domain.SubTrade:"
			+ " WHERE id = @id AND trader_id = @trader_id")
	List<SubTrade> getList(@Param("id") String id, @Param("trader_id") String tradeId, Sort sort);
```
will be crushed with an exception like
```
com.google.cloud.spanner.SpannerException: INVALID_ARGUMENT: io.grpc.StatusRuntimeException: INVALID_ARGUMENT: Syntax error: Unexpected keyword WHERE [at 1:604]
...sub_trades_1584821944625 ORDER BY subTradeId DESC WHERE deleted = false
```

PR #2267, opened 2020-03-21
Fix @Where with ORDER BY query generation

s13o:
Integration test added.
Fixes: #2266.

### DIFF
--- a/spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/test/domain/SubTradeRepository.java
+++ b/spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/test/domain/SubTradeRepository.java
@@ -16,15 +16,32 @@
 
 package org.springframework.cloud.gcp.data.spanner.test.domain;
 
+
+import java.util.List;
+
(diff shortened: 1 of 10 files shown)


spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/test/domain/SubTradeRepository.java
```diff
@@ -44,4 +44,8 @@
 			+ " WHERE id = @id AND trader_id = @trader_id")
 	List<SubTrade> getList(@Param("id") String id, @Param("trader_id") String tradeId, Sort sort);
 
+	@Query("SELECT count(1) FROM :org.springframework.cloud.gcp.data.spanner.test.domain.SubTrade:"
```
dmitry-s: please test that the final sql string is what we expect

s13o: added.
I found that two tests were absolutely identical - noPageableParamNotWrappedQueryTest & noPageableParamWrappedQueryTest.
So I left only one.

PR #2270, opened 2020-03-24
Polish Secret Manager Code

dzou:
Followup to secret manager version feature changes.

Contents of PR:
- Fixes integration tests
- Rename 'version' property to 'default-version' : I feel this more accurately conveys that when a user specifies a specific version of a secret it will override the default.
- Fixes bug where if the default version doesn't exist for a particular secret it will now correctly be skipped instead of throw error.
- General polish.

### DIFF
--- a/docs/src/main/asciidoc/secretmanager.adoc
+++ b/docs/src/main/asciidoc/secretmanager.adoc
@@ -51,11 +51,7 @@ Spring Cloud GCP Secret Manager offers several configuration properties to custo
 | Name | Description | Required | Default value
 | `spring.cloud.gcp.secretmanager.bootstrap.enabled` | Enables loading secrets from Secret Manager as a bootstrap property source. Set this to **true** to enable the feature. | No | `false`
 | `spring.cloud.gcp.secretmanager.secret-name-prefix` | A prefix string to prepend to the property names of secrets read from Secret Manager | No | "" (empty string)
-| `spring.cloud.gcp.secretmanager.default-version`
-| The default version that secrets will be read at, if a specific version for the secret is not configured.
-
-Note: If you set this to a value other than "latest" and a secret has not reached that version yet, then the secret **will not** be injected into the environment. | No | "latest"
-| `spring.cloud.gcp.secretmanager.versions.<secret-id>` | Defines a version for a specific secret-id to read from Secret Manager. It will take precedence over `spring.cloud.gcp.secretmanager.version` if both are specified | No | "" (empty string)
+| `spring.cloud.gcp.secretmanager.versions.<secret-id>` | Defines a version for a specific secret-id to read from Secret Manager instead of using the latest version. | No | "" (empty string)
 |===
 
 See the Authentication Settings section below for information on how to set properties to authenticate to Secret Manager.
--- a/docs/src/main/asciidoc/secretmanager.adoc
+++ b/docs/src/main/asciidoc/secretmanager.adoc
@@ -52,7 +52,7 @@ Spring Cloud GCP Secret Manager offers several configuration properties to custo
 | `spring.cloud.gcp.secretmanager.bootstrap.enabled` | Enables loading secrets from Secret Manager as a bootstrap property source. Set this to **true** to enable the feature. | No | `false`
 | `spring.cloud.gcp.secretmanager.secret-name-prefix` | A prefix string to prepend to the property names of secrets read from Secret Manager | No | "" (empty string)
 | `spring.cloud.gcp.secretmanager.default-version`
-| The default version that secrets will be read at, if a specific version for the secret is not configured.version
+| The default version that secrets will be read at, if a specific version for the secret is not configured.
 
 Note: If you set this to a value other than "latest" and a secret has not reached that version yet, then the secret **will not** be injected into the environment. | No | "latest"
 | `spring.cloud.gcp.secretmanager.versions.<secret-id>` | Defines a version for a specific secret-id to read from Secret Manager. It will take precedence over `spring.cloud.gcp.secretmanager.version` if both are specified | No | "" (empty string)
--- a/docs/src/main/asciidoc/secretmanager.adoc
+++ b/docs/src/main/asciidoc/secretmanager.adoc
@@ -51,7 +51,10 @@ Spring Cloud GCP Secret Manager offers several configuration properties to custo
 | Name | Description | Required | Default value
 | `spring.cloud.gcp.secretmanager.bootstrap.enabled` | Enables loading secrets from Secret Manager as a bootstrap property source. Set this to **true** to enable the feature. | No | `false`
 | `spring.cloud.gcp.secretmanager.secret-name-prefix` | A prefix string to prepend to the property names of secrets read from Secret Manager | No | "" (empty string)
-| `spring.cloud.gcp.secretmanager.version` | Defines a specific version of secrets to read from Secret Manager | No | "latest"
+| `spring.cloud.gcp.secretmanager.default-version`
+| The default version that secrets will be read at, if a specific version for the secret is not configured.version
+
+Note: If you set this to a value other than "latest" and a secret has not reached that version yet, then the secret **will not** be injected into the environment. | No | "latest"
 | `spring.cloud.gcp.secretmanager.versions.<secret-id>` | Defines a version for a specific secret-id to read from Secret Manager. It will take precedence over `spring.cloud.gcp.secretmanager.version` if both are specified | No | "" (empty string)
 |===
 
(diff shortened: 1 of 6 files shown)


meltsufin:
@eddumelendez FYI

docs/src/main/asciidoc/secretmanager.adoc
```diff
@@ -51,7 +51,10 @@ Spring Cloud GCP Secret Manager offers several configuration properties to custo
 | Name | Description | Required | Default value
 | `spring.cloud.gcp.secretmanager.bootstrap.enabled` | Enables loading secrets from Secret Manager as a bootstrap property source. Set this to **true** to enable the feature. | No | `false`
 | `spring.cloud.gcp.secretmanager.secret-name-prefix` | A prefix string to prepend to the property names of secrets read from Secret Manager | No | "" (empty string)
-| `spring.cloud.gcp.secretmanager.version` | Defines a specific version of secrets to read from Secret Manager | No | "latest"
+| `spring.cloud.gcp.secretmanager.default-version`
+| The default version that secrets will be read at, if a specific version for the secret is not configured.version
```
meltsufin:
```suggestion
| The default version that secrets will be read at, if a specific version for the secret is not configured.
```

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/SecretManagerPropertySource.java
```diff
@@ -87,43 +86,49 @@ public Object getProperty(String name) {
...
 				.setSecretVersion(version)
 				.build();
 
-		AccessSecretVersionResponse response = client.accessSecretVersion(secretVersionName);
-		return response.getPayload().getData();
+		try {
+			AccessSecretVersionResponse response = client.accessSecretVersion(secretVersionName);
+			return response.getPayload().getData();
+		}
+		catch (Exception e) {
```
meltsufin: Why not just catch `NotFoundException` here?

eddumelendez:
thanks for let me know @meltsufin

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/SecretManagerPropertySource.java
```diff
@@ -87,43 +86,49 @@ public Object getProperty(String name) {
...
+			if (e instanceof NotFoundException) {
+				LOGGER.debug("Skipped loading secret " + secretId + " because it does not have version " + defaultVersion);
```
meltsufin: Notice that this will also happen when a specific secret.version is specified.
I fear that this make it hard to debug mistakes in the client code.
Consider adding another property to chose whether to skip silently or fail.
Something like `failOnMissingSecretVersion`.

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/it/SecretManagerIntegrationTests.java
```diff
@@ -202,19 +185,21 @@ private void createSecretPayload(String data) {
...
-					return this.client.accessSecretVersion(secretVersionName) != null;
-				});
+				.anyMatch(secret -> secret.getName().contains(secretId));
+	}
+
+	private void deleteSecret(String secretId) {
+		try {
+			this.client.deleteSecret(SecretName.of(this.projectIdProvider.getProjectId(), secretId));
+		}
+		catch (Exception e) {
```
meltsufin: Again, why not just catch `NotFoundException`?

docs/src/main/asciidoc/secretmanager.adoc
dzou: Done.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/SecretManagerPropertySource.java
dzou: Ooops forgot I could do this; done.

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/it/SecretManagerIntegrationTests.java
dzou: Done.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/SecretManagerPropertySource.java
dzou: For this one, this is what I think makes the most sense:
- If a user-specified version doesn't exist, throw an error.
- If a default version doesn't exist for a secret, ignore the secret.

I think this makes the most sense for the default version handling because most users will have multiple secrets in their project, and they might all be at different versions (like some you'll be at version 5 and maybe some at version 1 or 2).

In practice, for any value of `default-version` other than "latest", there is a low chance that every secret in the project is up to that version. For example, if a choose the version `2`, this bootstrap loading would fail if any secret in the project was only at version `1`. So I think the `default-version` option is only useful if you skip the secret that isn't up to the default version.

meltsufin: Your reasoning makes sense, but I still feel that it's just too complicated and will leave room for some surprises for people who do not carefully read the documentation. You can argue that the default version setting should never be used, unless the project keeps all of the secrets at the same version. Maybe they have a convention where each version of the application has a corresponding version of the secrets. In that case, as a developer I would like a hard fail if I make a mistake and forget to include a secret with the right version that follows that convention.
Of course there could be exceptions. Therefore, I'm suggesting to add a property to allow skipping secrets with non-existing versions.
I guess we're struggling here partly because we're not sure what is the common use-case for `default-version`.
Maybe @eddumelendez can help us out here.

dzou: I see. I just was not a fan of the additional setting because I imagine most projects will have secrets of different versions.

I don't see it being likely that each version of an application will have its own version of secrets - like a password to your MySQL database might stay the same across all of your application versions but some other secret might be updated more frequently. And what if you have to add a new secret to an application? If you're at version 5 for your current secrets, the new secret must be updated 5 times to hit version 5. (The only way to increment a secret version is to create a new version of it in the APIs; you can't create a new secret and set it to 5 directly)

Hmm, the more I think about it, the less useful overriding the default-version is. It feels like `latest` makes the most sense and overriding secret versions should be done on an individual level per secret. What do you guys think? @meltsufin @eddumelendez

meltsufin: I didn't realize that the version cannot be set directly.
I happy to remove the default version feature.
@saturnism @eddumelendez WDYT?

eddumelendez: > It feels like defaulting to latest makes the most sense and overriding secret versions should be done on an individual level per secret.

I'm in favor of this. 

There is just two options for every secret which is use the latest or a specific version. Given that, `default-version` doesn't make much sense. 

Also, regarding to:

> If a user-specified version doesn't exist, throw an error.
If a default version doesn't exist for a secret, ignore the secret

So, if the secret is ignored then `@Value` will not resolve it, right? I think the docs are pretty clear about the context but not sure if will be hard to figure it out is we are missing to read the docs.

meltsufin: Sounds like we're all leaning towards just removal of `default-version` and hard fail in case a specified version doesn't exist.

dzou:
@meltsufin - done, just removed the `default-version`; PTAL.

PR #2280, opened 2020-03-26
Move Pub/Sub health autoconfiguration under main one

### LINKED ISSUE #2271: Pub/Sub: 1.2.2 RELEASE autoconfigure failure with spring.cloud.gcp.pubsub.enabled=false
and yet again as with #2037 now the `pubSubTemplate` is broken when `spring.cloud.gcp.pubsub.enabled=false`. This was working with 1.2.1.

error message:
```
***************************
APPLICATION FAILED TO START
***************************

Description:

Parameter 0 of method pubSubHealthIndicator in org.springframework.cloud.gcp.autoconfigure.pubsub.health.PubSubHealthIndicatorAutoConfiguration required a bean of type 'org.springframework.cloud.gcp.pubsub.core.PubSubTemplate' that could not be found.

The following candidates were found but could not be injected:
	- Bean method 'pubSubTemplate' in 'GcpPubSubAutoConfiguration' not loaded because @ConditionalOnProperty (spring.cloud.gcp.pubsub.enabled) found different value in property 'spring.cloud.gcp.pubsub.enabled'


Action:

Consider revisiting the entries above or defining a bean of type 'org.springframework.cloud.gcp.pubsub.core.PubSubTemplate' in your configuration.
```

@elefeint I have tried using `spring.cloud.gcp.pubsub.reactive.enabled=false`, but this time it did not help.

elefeint:
Prevent Pub/Sub actuator healthcheck autoconfiguration from triggering if overall Pub/Sub is disabled.

Fixes #2271.


elefeint:
@artembilan Would you mind taking a look at the autoconfiguration changes? This implements your original recommendation from #2040.

elefeint:
I was going to send a PR to change Datastore after this.
It's true that this is a breaking change for someone who excluded the autoconfiguration by class name. Should I ice this PR until we are ready to do a major release, likely after moving github orgs?

artembilan:
I think that's OK to do this in the minor release with an appropriate "Migration Guide".
Doesn't look like such a class should be used too much in the target projects...

meltsufin:
> I was going to send a PR to change Datastore after this.
> It's true that this is a breaking change for someone who excluded the autoconfiguration by class name. Should I ice this PR until we are ready to do a major release, likely after moving github orgs?

What do you think of just adding `@ConditionalOnBean(value = PubSubTemplate.class)` and the reference documentation for now? The subclass can wait till a major release.

elefeint:
Yeah, that's what I meant. But I might as well open a new PR for that, since this one is a larger change.

elefeint:
@artembilan I'd be okay with documenting this really well -- the change for end users would be trivial: removing their existing exclusion. Spring Boot documentation say somewhere that all autoconfiguration is an implementation detail except for class name.

@meltsufin What do you think about proceeding with original PR with good release note warnings?

meltsufin:
> Spring Boot documentation say somewhere that all autoconfiguration is an implementation detail except for class name.

That sounds like a vote on the side of keeping the class as is and adding the condition I suggested.
I'm OK with either, as long as we get it consistent with the Datastore Health Indicator autoconfig.

docs/src/main/asciidoc/pubsub.adoc
```diff
@@ -25,6 +25,24 @@ dependencies {
 
 This starter is also available from https://start.spring.io[Spring Initializr] through the `GCP Messaging` entry.
 
+=== Spring Boot Actuator Support
+
+==== Cloud Pub/Sub Health Indicator
+
+If you are using Spring Boot Actuator, you can take advantage of the Cloud Pub/Sub health indicator called `pubsub`.
+The health indicator will verify whether Cloud Pub/Sub is up and accessible by your application.
+To enable it, all you need to do is add the https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#production-ready[Spring Boot Actuator] to your project.
```
meltsufin: Is there a way to disable it via the Actuator configuration?

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/pubsub/PubSubHealthIndicatorTests.java
```diff
@@ -69,4 +72,16 @@ public void healthDownGenericException() {
 		assertThat(healthIndicator.health().getStatus()).isEqualTo(Status.DOWN);
 	}
 
+	@Test
+	public void healthIndicatorDisabledWhenPubSubTurnedOff() {
+		ApplicationContextRunner contextRunner = new ApplicationContextRunner()
+				.withConfiguration(AutoConfigurations.of(
+						GcpPubSubAutoConfiguration.class, PubSubHealthIndicatorAutoConfiguration.class))
+				.withPropertyValues("spring.cloud.gcp.datastore.project-id=test-project",
+						"management.health.pubsub.enabled=true",
```
meltsufin: I think we should mention this property in the refdoc as well.

Issue #2283, opened 2020-03-26
Allow to select the source projectId when using secret


guillaumeblaquiere:
I would like to load easily secrets from more than 1 project. Today, only the default project, or the project set in the properties can be easily used by the SecretManagerTemplate bean

**Describe the solution you'd like**
The capability to create and read secrets from default project (like today) but also to have the capability to set a projectId for reading the secrets in another project

This implies changes into SecretManagerTemplate and into the bootstrap of secret. On the bootstrap, a solution can be to set the project id in the same way as the versions are defined today.
`spring.cloud.gcp.secretmanager.projectId.<secretId>`

Open to discuss about the best implementation.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

dzou:
Thanks for your PR.

> On the bootstrap, a solution can be to set the project id in the same way as the versions are defined today.
> `spring.cloud.gcp.secretmanager.projectId.<secretId>`

Yeah I like this idea a lot; would be in favor of letting users specify a custom project id.

guillaumeblaquiere:
I created a new PR for bootstrap: #2289

Integration test relevance to discuss. TODOs in code where updates are required.
Open to discuss.

dzou:
Regarding property source loading, how would you handle the case where if two projects had 2 secrets with the same secret ID?

dzou:
Hey I took a look at the PR, my bad; so my comments about name collisions are incorrect; this is like specifying the project-id similar to specifying a version over the default.

Could you describe your use-case a bit more? Just want to learn more about it; right now I guess my only concern is that the feature increases the complexity of the settings, and then also there might be unexpected interactions with if a user also sets the version override setting for the specific secret as well.

guillaumeblaquiere:
In my context, and for cost reason, we have 1 shared project for test and dev environment with 1 Cloud SQL MySQL and 1 Cloud SQL PostGreSQL database. We chose to store the passwords in the secret manager of this shared project.

Then, the developer and the other projects can use this database by using the password stored in SecretManager, in addition of other secret stored in their project (like API Keys).

But, you have right. If I have the "password" secret name in my default project and in the shared project, how can I specify both ?

The best way is to wrap the projectId and the secretId in the same key, something like `projectId/secretId`. If `projectId/` is missing, we use default project. Like this, it's compatible with the existing.

But I never saw `/` in properties name. And, I you said, it's become to be quite complex to define the properties and to understand them.

[...]

PR #2284, opened 2020-03-26
Secret manager template with project (#2283)

guillaumeblaquiere:
In relation with the issue #2283

This first step is focused on the Secret Manager Template bean. If it's on the right way, I will add the bootstrap part to load secret from another projects

### DIFF
--- a/CONTRIBUTING.adoc
+++ b/CONTRIBUTING.adoc
@@ -47,7 +47,7 @@ Install the plugin `Eclipse Code Formatter`.
 You can find it by searching in "Browse Repositories", under `Settings > Plugins` within IDEA (Once installed, you will need to reboot IDEA for it to take effect).
 
 Then navigate to `Settings > Other Settings` (this might be under `Preferences` on Mac) and select the Eclipse Code Formatter.
-Select the `eclipse-code-formatter.xml` file for the field `Eclipse Java Formatter config file` and the file `eclipse.importorder` for the field `Import order`.
+Select the `eclipse-code-formatter.xml` file for the field `Eclipse Java Formatter config file` and the file `eclipse.importorder` for the field `Import order` in the `src/eclipse` directory
 Enable the `Eclipse code formatter` by clicking `Use the Eclipse code formatter` radio button at the top of the page, then click the *OK* button.
 
 * NOTE: If you configure the `Eclipse Code Formatter` from `File > Other Settings > Default Settings`, it will set this policy across all of your IDEA projects.
--- a/CONTRIBUTING.adoc
+++ b/CONTRIBUTING.adoc
@@ -47,7 +47,7 @@ Install the plugin `Eclipse Code Formatter`.
 You can find it by searching in "Browse Repositories", under `Settings > Plugins` within IDEA (Once installed, you will need to reboot IDEA for it to take effect).
 
 Then navigate to `Settings > Other Settings` (this might be under `Preferences` on Mac) and select the Eclipse Code Formatter.
-Select the `eclipse-code-formatter.xml` file for the field `Eclipse Java Formatter config file` and the file `eclipse.importorder` for the field `Import order` in the `src/eclipse` directory
+Select the `eclipse-code-formatter.xml` file for the field `Eclipse Java Formatter config file` and the file `eclipse.importorder` for the field `Import order` in the `src/eclipse` directory.
 Enable the `Eclipse code formatter` by clicking `Use the Eclipse code formatter` radio button at the top of the page, then click the *OK* button.
 
 * NOTE: If you configure the `Eclipse Code Formatter` from `File > Other Settings > Default Settings`, it will set this policy across all of your IDEA projects.
--- a/spring-cloud-gcp-samples/spring-cloud-gcp-secretmanager-sample/src/main/java/com/example/SecretManagerWebController.java
+++ b/spring-cloud-gcp-samples/spring-cloud-gcp-secretmanager-sample/src/main/java/com/example/SecretManagerWebController.java
@@ -18,7 +18,6 @@ package com.example;
 
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
-import org.springframework.cloud.gcp.core.GcpProjectIdProvider;
 import org.springframework.cloud.gcp.secretmanager.SecretManagerTemplate;
 import org.springframework.core.env.Environment;
 import org.springframework.stereotype.Controller;
@@ -28,8 +27,6 @@ import org.springframework.web.bind.annotation.PostMapping;
 import org.springframework.web.bind.annotation.RequestParam;
 import org.springframework.web.bind.annotation.ResponseBody;
 import org.springframework.web.servlet.ModelAndView;
-
-import javax.annotation.PostConstruct;
 
 @Controller
 public class SecretManagerWebController {
@@ -65,8 +62,9 @@ public class SecretManagerWebController {
 	public String getSecret(@RequestParam String secretId, @RequestParam String version, @RequestParam String projectId,
 			ModelMap map) {
 		String resolvedVersion = version;
-		if (resolvedVersion == null || resolvedVersion.equals(""))
+		if (resolvedVersion == null || resolvedVersion.equals("")) {
 			resolvedVersion = SecretManagerTemplate.latestVersion;
+		}
 
 		String secretPayload;
 		if (projectId == null || projectId.equals("")) {
--- a/spring-cloud-gcp-samples/spring-cloud-gcp-secretmanager-sample/src/main/java/com/example/SecretManagerWebController.java
+++ b/spring-cloud-gcp-samples/spring-cloud-gcp-secretmanager-sample/src/main/java/com/example/SecretManagerWebController.java
@@ -16,6 +16,7 @@
 
 package com.example;
 
+import org.apache.commons.lang3.StringUtils;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.cloud.gcp.secretmanager.SecretManagerTemplate;
@@ -59,35 +60,42 @@ public class SecretManagerWebController {
 
 	@GetMapping("/getSecret")
 	@ResponseBody
-	public String getSecret(@RequestParam String secretId, @RequestParam String version, @RequestParam String projectId,
+	public String getSecret(
+			@RequestParam String secretId,
+			@RequestParam(required=false) String version,
+			@RequestParam(required=false) String projectId,
 			ModelMap map) {
-		String resolvedVersion = version;
-		if (resolvedVersion == null || resolvedVersion.equals("")) {
-			resolvedVersion = SecretManagerTemplate.LATEST_VERSION;
+
+		if (StringUtils.isEmpty(version)) {
+			version = SecretManagerTemplate.LATEST_VERSION;
 		}
 
 		String secretPayload;
-		if (projectId == null || projectId.equals("")) {
-			// The latest parameter "Version" is optional. If missing, the latest is returned
-			secretPayload = this.secretManagerTemplate.getSecretString(secretId, resolvedVersion);
+		if (StringUtils.isEmpty(projectId)) {
+			secretPayload = this.secretManagerTemplate.getSecretString(secretId, version);
 		}
 		else {
-			secretPayload = this.secretManagerTemplate.getSecretString(secretId, resolvedVersion, projectId);
+			secretPayload = this.secretManagerTemplate.getSecretString(secretId, version, projectId);
 		}
+
 		return "Secret ID: " + secretId + " | Value: " + secretPayload
 				+ "<br/><br/><a href='/'>Go back</a>";
 	}
 
 	@PostMapping("/createSecret")
 	public ModelAndView createSecret(
-			@RequestParam String secretId, @RequestParam String secretPayload, @RequestParam String projectId,
+			@RequestParam String secretId,
+			@RequestParam String secretPayload,
+			@RequestParam(required=false) String projectId,
 			ModelMap map) {
-		if (projectId == null || projectId.equals("")) {
+
+		if (StringUtils.isEmpty(projectId)) {
 			this.secretManagerTemplate.createSecret(secretId, secretPayload);
 		}
 		else {
 			this.secretManagerTemplate.createSecret(secretId, secretPayload, projectId);
 		}
+
 		map.put("applicationSecret", this.applicationSecretValue);
 		map.put("myApplicationSecret", this.myApplicationSecretValue);
 		map.put("message", "Secret created!");
--- a/spring-cloud-gcp-samples/spring-cloud-gcp-secretmanager-sample/src/main/java/com/example/SecretManagerWebController.java
+++ b/spring-cloud-gcp-samples/spring-cloud-gcp-secretmanager-sample/src/main/java/com/example/SecretManagerWebController.java
@@ -17,6 +17,7 @@
 package com.example;
 
 import org.apache.commons.lang3.StringUtils;
+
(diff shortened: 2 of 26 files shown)


pivotal-issuemaster:
@guillaumeblaquiere Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-gcp&pullRequestId=2284)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-cloud/spring-cloud-gcp&pullRequestId=2284) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@guillaumeblaquiere Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-gcp&pullRequestId=2284)!

dzou:
@guillaumeblaquiere - Thanks for the PR! We will take a look. Also we'll see what the result of our offline discussion is regarding whether we would like to limit the scope of the template to 1 project or allow it to query any project.

CONTRIBUTING.adoc
```diff
@@ -47,7 +47,7 @@ Install the plugin `Eclipse Code Formatter`.
 You can find it by searching in "Browse Repositories", under `Settings > Plugins` within IDEA (Once installed, you will need to reboot IDEA for it to take effect).
 
 Then navigate to `Settings > Other Settings` (this might be under `Preferences` on Mac) and select the Eclipse Code Formatter.
-Select the `eclipse-code-formatter.xml` file for the field `Eclipse Java Formatter config file` and the file `eclipse.importorder` for the field `Import order`.
+Select the `eclipse-code-formatter.xml` file for the field `Eclipse Java Formatter config file` and the file `eclipse.importorder` for the field `Import order` in the `src/eclipse` directory
```
meltsufin:
```suggestion
Select the `eclipse-code-formatter.xml` file for the field `Eclipse Java Formatter config file` and the file `eclipse.importorder` for the field `Import order` in the `src/eclipse` directory.
```

spring-cloud-gcp-secretmanager/src/main/java/org/springframework/cloud/gcp/secretmanager/SecretManagerOperations.java
```diff
@@ -40,6 +40,21 @@
...
+	 * <p>
+	 * If there is already a secret saved in SecretManager with the specified
+	 * {@code secretId}, then it simply creates a new version under the secret with the secret
+	 * {@code payload}.
+	 *
+	 * @param secretId the secret ID of the secret to create.
+	 * @param payload the secret payload string.
+	 * @param projectId unique identifier of your project.
+	 */
+	void createSecretWithProject(String secretId, String payload, String projectId);
```
meltsufin: I would just create an overloaded method `createSecret(secretId, payload, projectId)`

spring-cloud-gcp-secretmanager/src/test/java/org/springframework/cloud/gcp/secretmanager/SecretManagerTemplateTests.java
```diff
@@ -90,6 +90,35 @@ public void testCreateSecretIfMissing() {
 		verify(this.client).addSecretVersion(addSecretVersionRequest);
 	}
 
+	@Test
+	public void testCreateSecretIfMissingWithProject() {
```
dzou:
```suggestion
	public void testCreateSecretIfMissing_withProject() {
```

Minor style - if you could make modify the test name to be `_withProject` I think it is more clear that it refers to using the project parameter

guillaumeblaquiere:
I'm totally agree with you and I wanted to overload!! But I can't. Indeed, Java method signature are on the type, not on the param name and there were a conflict between these (for example):

```
	@Override
	public String getSecretString(String secretId, String versionName) {....}

	@Override
	public String getSecretString(String secretId, String projectId) {....}
```

Thus, I had to change the method name for differentiating the signatures. And for being consistent in  the whole code, I added, the `WithProject` suffix to all the methods that offer the possibility to set the `ProjectId` param.

Another idea is to require a GcpProjectIdProvider type for projectId insteand of a String
```
	@Override
	public String getSecretString(String secretId, GcpProjectIdProvider projectId) {....}
```

And to let the developer to perform this kind of call
```
	String secret = secretManagerTemplate.getSecretString("my-secret", () -> "custom-project");
```
It works, but, IMO, it's not clean and readable.

Let me know what you prefer.

dzou:
> I'm totally agree with you and I wanted to overload!! But I can't. Indeed, Java method signature are on the type, not on the param name and there were a conflict between these (for example):
> 
> ```
> 	@Override
> 	public String getSecretString(String secretId, String versionName) {....}
> 
> 	@Override
> 	public String getSecretString(String secretId, String projectId) {....}
> ```
> 
> Thus, I had to change the method name for differentiating the signatures. And for being consistent in the whole code, I added, the `WithProject` suffix to all the methods that offer the possibility to set the `ProjectId` param.

That is a good point. Maybe we could just omit adding the `String getSecretStringWithProject(String secretId, String projectId)`, and only add `String getSecretString(String secretId, String version, String projectId)`?

I.e. don't add the method that lets users specify only `secretId` and `projectId`. Maybe this is not too bad of a compromise to keep the overloading? I don't think users would mind having to specify the version in addition to the project id.

guillaumeblaquiere:
> I.e. don't add the method that lets users specify only secretId and projectId. Maybe this is not too bad of a compromise to keep the overloading? I don't think users would mind having to specify the version in addition to the project id.

Yes, it's an good option. I though about it, didn't know the best solution. Go ahead with yours.

spring-cloud-gcp-secretmanager/src/test/java/org/springframework/cloud/gcp/secretmanager/SecretManagerTemplateTests.java
guillaumeblaquiere: done in newer commit

spring-cloud-gcp-secretmanager/src/main/java/org/springframework/cloud/gcp/secretmanager/SecretManagerOperations.java
guillaumeblaquiere: done in newer commit

CONTRIBUTING.adoc
guillaumeblaquiere: done in newer commit

guillaumeblaquiere:
Updated with the bootstrap part.

I added a Static public variable for the latestVersion. I did this because the user need to specify the version in case of use of `ProjectId` param. For helping him, the static variable is here to provide right latestVersion string value.

Sometime, the convention is to set in UPPER_CASE the static value. Tell me if it's the case in your convention too.

spring-cloud-gcp-secretmanager/src/main/java/org/springframework/cloud/gcp/secretmanager/SecretManagerTemplate.java
```diff
@@ -40,6 +40,11 @@
  */
 public class SecretManagerTemplate implements SecretManagerOperations {
 
+	/**
+	 * Default value for the latest version of the secret.
+	 */
+	public static final String latestVersion = "latest";
```
dzou: Oh yup, I see. We would prefer all uppercase name here: `LATEST_VERSION`.

dzou:
Thanks Guillaume, I'm also gonna help merge in master to your branch and also push some refdoc updates to your PR branch.

spring-cloud-gcp-secretmanager/src/main/java/org/springframework/cloud/gcp/secretmanager/SecretManagerTemplate.java
guillaumeblaquiere: Pushed

guillaumeblaquiere:
Coverage result is strange!

dzou:
Hey @guillaumeblaquiere - Would you mind if we split this PR and only do the template changes in this one? And then we will address the changes to the property source loading for multiple project IDs in the second one?

Sorry to ask this and make you undo some work. Just wanted to limit the scope of the PR and also give us an opportunity to discuss how to best expose loading secrets from multiple projects as property source.

guillaumeblaquiere:
Done (with rebase into master to resolve conflicts).

guillaumeblaquiere:
New PR for bootstrap: #2289

dzou:
Thanks for the splitting up the PR. Added some minor commits and merged master into the branch. Looks good on my end.

guillaumeblaquiere:
Thanks. Travis is broken. Tell me if this PR need more work.
Best

dzou:
> Thanks. Travis is broken. Tell me if this PR need more work.
> Best

Just pushed the fix. I think the PR is in a good state; I'll take care of any remaining comments that Mike might have.

spring-cloud-gcp-secretmanager/src/main/java/org/springframework/cloud/gcp/secretmanager/SecretManagerOperations.java
```diff
@@ -73,6 +103,18 @@
...
+	 * Gets the secret payload of the specified {@code secretId} at version
+	 * {@code versionName} for a specific {@code projectId}.
+	 *
+	 * @param secretId unique identifier of your secret in Secret Manager.
+	 * @param versionName which version of the secret to load. The version can be a version
+	 *     number as a string (e.g. "5") or an alias (e.g. "latest").
+	 * @param projectId unique identifier of your project.
+	 * @return The secret payload as String
+	 */
+	String getSecretString(String secretId, String versionName, String projectId);
```
meltsufin: Are we missing `String getSecretString(String secretId, String projectId)` variation or would this be too much?

spring-cloud-gcp-secretmanager/src/main/java/org/springframework/cloud/gcp/secretmanager/SecretManagerOperations.java
```diff
@@ -111,4 +177,14 @@
...
+
+	/**
+	 * Returns true if there already exists a secret under the GCP {@code projectId} with the
+	 * {@code secretId}.
+	 *
+	 * @param secretId unique identifier of your secret in Secret Manager.
+	 * @param projectId unique identifier of your project.
+	 * @return true if the secret exists in Secret Manager; false otherwise
+	 */
+	boolean secretExists(String secretId, String projectId);
```
meltsufin: Does it make sense to have `secretExists(String secretId, String versionName, String projectId)`?
This would check if a specific version of the secret exists.

spring-cloud-gcp-secretmanager/src/main/java/org/springframework/cloud/gcp/secretmanager/SecretManagerOperations.java
guillaumeblaquiere: No, it's not missing. We discussed previously about this.

> Java method signature are on the type, not on the param name and there were a conflict between these (for example):
>
>```
>	@Override
>	public String getSecretString(String secretId, String versionName) {....}
>
>	@Override
>	public String getSecretString(String secretId, String projectId) {....}
>```

So we choose to only propose the projectId with the version. And that's why I also introduced a static public variable `LATEST_VERSION` for helping the developers when they want to get the secret without care about the version

But, we can explore another solution.

spring-cloud-gcp-secretmanager/src/main/java/org/springframework/cloud/gcp/secretmanager/SecretManagerOperations.java
guillaumeblaquiere: I guess that the primary use of this function is internal for the secret creation (create a new secret or add a version to an existing secret). Thereby, the version is useless in this context.

However, if we want to offer a convenient interface to developer, it could be interesting. If you agree, I propose these interfaces:

```
boolean secretExists(String secretId);// Check latest version in default project
boolean secretExists(String secretId, String version); // Check specific version in default project
boolean secretExists(String secretId, String version, String projectId);// // Check specific version in specific project
```
Like for your `getSecretString` comment, I won't be able to keep `boolean secretExists(String secretId, String projectId);` because of method signature conflict with `boolean secretExists(String secretId, String version);`

Let me know.

dzou: Yeah let's omit adding `secretExists(String secretId, String versionName, String projectId)`, since it would conflict with existing method signature and I think its value is relatively limited as you say with it being primarily internal.

meltsufin:
@guillaumeblaquiere @dzou I'm a little concerned with proliferation of the different method permutations. We have `getSecret`, `getSecretString`, and `getSecretByteString` multiplied by 3 variations of each. Maybe we should think about how common the use of each method will be. I'm thinking that maybe we just need to add the project and version permutations to the `getSecretByteString` because it's pretty trivial to go from `ByteString` to `byte[]` and `String`. We get the benefits of the flexibility of being able to specify the project and the version without adding too much to the cognitive load of the users of the template. You can think of the users of the `getSecretByteString` as the power users, and the users of the version and project are also power users. So, we can take care of their needs with just two additional methods.

Similarly, for the `createSecret` methods we should probably just choose one type of payload, probably `byte[]` to focus on to give the version and project flexibility on.

What do you guys think?

spring-cloud-gcp-secretmanager/src/main/java/org/springframework/cloud/gcp/secretmanager/SecretManagerOperations.java
meltsufin: Sorry for the noise, you are absolutely right about the method signature collisions. I didn't think of that.

guillaumeblaquiere:
@meltsufin, Your concern is correct. I think that your solution acceptable:

* Less methods -> easier to maintain and read
* For "exceptional" use, the template offer the capability to specify the projectId but the developer have to do few "extra" work for getting his secret (in string or byteString).

It's a good balance.

@dzou, what do you think?

dzou:
> @meltsufin, Your concern is correct. I think that your solution acceptable:
> 
> * Less methods -> easier to maintain and read
> * For "exceptional" use, the template offer the capability to specify the projectId but the developer have to do few "extra" work for getting his secret (in string or byteString).
> 
> It's a good balance.
> 
> @dzou, what do you think?

Sounds good to me. Just pushed the changes.

spring-cloud-gcp-samples/spring-cloud-gcp-secretmanager-sample/src/main/java/com/example/SecretManagerWebController.java
```diff
@@ -46,25 +48,58 @@
...
+			@RequestParam(required = false) String projectId,
+			ModelMap map) {
+
+		if (StringUtils.isEmpty(version)) {
+			version = SecretManagerTemplate.LATEST_VERSION;
+		}
+
+		String secretPayload;
+		if (StringUtils.isEmpty(projectId)) {
+			secretPayload = this.secretManagerTemplate.getSecretString(secretId, version);
```
meltsufin: Should we remove this variant of `getSecretString` as well? Same goes for `getSecretBytes`.

guillaumeblaquiere: These variants exists in the version 1.2.2. We shouldn't remove them to ensure the downward compatibility.

meltsufin: Ah, didn't realize that. Thanks!

PR #2287, opened 2020-03-27
Update trace docs now that the trace scheduler is hidden

### LINKED ISSUE #2016: Hide traceSenderThreadPool in trace autoconfiguration

elefeint:
This is a leftover documentation task from #2016. After the trace scheduler bean was removed in #2158, this PR removes the following wording: " This is especially a problem because our starter comes with an executor. To avoid this noise, please disable automatic instrumentation of executors via `spring.sleuth.scheduled.enabled=false` in your application configuration."

I left the overall warning in place, though, because this will still be a problem when combined with any other starters (for example, our Pub/Sub starter) that add scheduler beans to context.

I also updated the branding from Stackdriver to Cloud Trace where possible (per [rebranding announcement](https://cloud.google.com/blog/products/management-tools/cloud-operations-grows-with-monitoring-logging-more)). The code stays prefixed with Stackdriver.

Closes #2016.

### DIFF
--- a/docs/src/main/asciidoc/trace.adoc
+++ b/docs/src/main/asciidoc/trace.adoc
@@ -94,7 +94,7 @@ spring.sleuth.web.skipPattern=(^cleanup.*|.+favicon.*)  # Ignore some URL paths.
 spring.sleuth.scheduled.enabled=false                   # disable executor 'async' traces
 ----
 
-WARNING: By default, Spring Cloud Sleuth auto-configuration instruments executor beans, which may cause recurring traces with the name `async` to appear in Cloud Trace if your application or one of its dependencies introduces scheduler beans into Spring application context. To avoid this noise, please disable automatic instrumentation of executors via `spring.sleuth.scheduled.enabled=false` in your application configuration.
+WARNING: By default, Spring Cloud Sleuth auto-configuration instruments executor beans, which may cause recurring traces with the name `async` to appear in Stackdriver Trace if your application or one of its dependencies introduces scheduler beans into Spring application context. To avoid this noise, please disable automatic instrumentation of executors via `spring.sleuth.scheduled.enabled=false` in your application configuration.
 
 Spring Cloud GCP Trace does override some Sleuth configurations:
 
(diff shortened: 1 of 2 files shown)


docs/src/main/asciidoc/trace.adoc
```diff
@@ -89,21 +89,21 @@ For example, when you are testing to see the traces are going through, you can s
...
 [source]
 ----
-spring.sleuth.sampler.probability=1                     # Send 100% of the request traces to Stackdriver.
+spring.sleuth.sampler.probability=1                     # Send 100% of the request traces to Cloud Trace.
 spring.sleuth.web.skipPattern=(^cleanup.*|.+favicon.*)  # Ignore some URL paths.
 spring.sleuth.scheduled.enabled=false                   # disable executor 'async' traces
 ----
 
-WARNING: By default, Spring Cloud Sleuth auto-configuration instruments executor beans, which causes many traces with the name `async` to appear in Stackdriver Trace. This is especially a problem because our starter comes with an executor. To avoid this noise, please disable automatic instrumentation of executors via `spring.sleuth.scheduled.enabled=false` in your application configuration.
+WARNING: By default, Spring Cloud Sleuth auto-configuration instruments executor beans, which may cause recurring traces with the name `async` to appear in Cloud Trace if your application or one of its dependencies introduces scheduler beans into Spring application context.
```
meltsufin: Are we sure that the name is always `async`? Maybe just remove the warning altogether?
If we do keep this warning, can we mention how to disable them with `spring.sleuth.scheduled.enabled=false`?

elefeint: `async` is a [default name](https://github.com/spring-cloud/spring-cloud-sleuth/blob/ec1ada24b18dd7221e377336dbcbd73b27604f55/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/async/TraceRunnable.java#L40). I've tested it with our pubsub starter, and yep, it was async, starting after the thread pool was first used.

PR #2289, opened 2020-03-27
Secret manager autoconfigure projectId selection

guillaumeblaquiere:
Bootstrap part for using custom project when define a secret into the bootstrap file or into the @Value decorator.

TODOs to solve on integration tests: Need an additional project for performing tests of creation and access to the secrets because, currently, the tests are performed in the current default project. Not relevant.


meltsufin:
@guillaumeblaquiere Is this PR ready for review? It looks like you're adding projectId customization support, but the title of the PR doesn't reflect that.

guillaumeblaquiere:
@meltsufin Yes, it is, but only for the bootstrap part. I split the previous PR. This one for the bootstrap (and autoconfigure) part. The #2284 for the template part.

dzou:
We will move forward with #2302 which offers more elegant way for specifying project and version combinations for all secrets in properties and template. Gonna close this one.
Thanks for the contributions @guillaumeblaquiere !

Issue #2268, opened 2020-03-23
Whitelisted MDC keys in Sleuth are not present when Trace is enabled


olssonoskar:
**Describe the bug**
Cloud version: Hoxton.SR3
Boot version: 2.2.5

We configure Sleuth to include request headers in the mdc (for logs). It works well with the StackdriverJsonLayout by default, but after enabling trace, these headers are not populated anymore.

The expected behavior is that the population of MDC should work and be included in logging statements no matter if trace is enabled or not. 

Including sample project with a test to verify the behavior. Enabling/disabling trace will make the test pass or fail.

[demo.zip](https://github.com/spring-cloud/spring-cloud-gcp/files/4369124/demo.zip)

dzou:
Thanks for the report and the reproducible example @olssonoskar . I will take a look at this for you and hope to followup shortly.

dzou:
@meltsufin - Some initial first thoughts after debugging - I think this might be related to zipkin code since the traces bring in Zipkin; maybe there is some code in our Zipkin dependencies that is not forwarding the MDC headers correctly.

meltsufin:
@adriancole Do you have hints for us on how to investigate this?

adriancole:
@olssonoskar can you make a fork of this branch and add your changes to it (without gcp in use) I'd like to see if this is sleuth in general or specifically when -gcp is involved https://github.com/openzipkin/sleuth-webmvc-example#spring-webflux-tracing

when done, post your fork back here which has a commit of your changes. it makes it really easy to validate. cheers!

adriancole:
@meltsufin first trying to see if this has to do with the assembly hooks added by sleuth

meltsufin:
I tried `spring.sleuth.log.slf4j.whitelisted-mdc-keys=req-id` instead of `log.slf4j.whitelisted-mdc-keys=req-id` as recommended in the [documentation](https://cloud.spring.io/spring-cloud-static/spring-cloud-sleuth/2.1.0.RELEASE/single/spring-cloud-sleuth.html) and it worked for me.

See: https://github.com/openzipkin/sleuth-webmvc-example/pull/34

@olssonoskar Can you give a try?

meltsufin:
Never mind, my mistake. You actually have the right property in your example.
Continuing to investigate...

meltsufin:
The difference I'm seeing is in the default extractor in the `HttpServerHandler`.

[...]

PR #2290, opened 2020-03-27
Support Extra Propagation Fields with Trace

meltsufin:
Providing a Trace Propagation Factory Builder instead of just the Factory,
which allows us to rely on the `TraceAutoConfiguration.sleuthPropagation()`
rather than completely skipping it.

Fixes: #2268.


meltsufin:
@adriancole Ultimately, what we're after is the ability to configure the delegate for `TraceAutoConfiguration`, which is currently hardcoded as `B3Propagation.FACTORY`. I'm open to whatever would be the best way to do it. It will probably require some tweaks to `TraceAutoConfiguration` in Sleuth though.

adriancole:
the exception will go away with a new brave release which will immediately go into the next sleuth release.

https://github.com/openzipkin/brave/pull/1128

meltsufin:
@adriancole Thanks for such a quick turnaround!

spring-cloud-gcp-dependencies/pom.xml
```diff
@@ -16,8 +16,8 @@
 	<packaging>pom</packaging>
 
 	<properties>
-		<gcp-libraries-bom.version>4.4.1</gcp-libraries-bom.version>
-		<cloud-sql-socket-factory.version>1.0.15</cloud-sql-socket-factory.version>
+		<gcp-libraries-bom.version>5.3.0</gcp-libraries-bom.version>
```
elefeint, line 19: Does it make sense to move the version updates to a separate PR?

meltsufin:
Looks like `TraceSampleApplicationTests` fails as soon as we upgrade `spring-cloud-sleuth` to `2.2.3.BUILD-SNAPSHOT`. 😞

elefeint:
I can look into it Thursday -- I've had a lot of tracing on my mind lately.

meltsufin:
We should rework this PR based on https://github.com/spring-cloud/spring-cloud-sleuth/pull/1647 as well.

elefeint:
My first thought was that we needed a newer brave, but sleuth snapshot already pulls in the latest...

elefeint:
Ah. `HexCodec` got moved into a subpackage in openzipkin/brave#1193.

We need to update https://github.com/openzipkin/zipkin-gcp/blob/8b2462c054737259bdd707de49da0cc32aaa6806/propagation-stackdriver/src/main/java/zipkin2/propagation/stackdriver/XCloudTraceContextExtractor.java

meltsufin:
See: https://github.com/openzipkin/zipkin-gcp/issues/175

adriancole:
zipkin-gcp 0.17.0 is on the way now. shading the internal HexCodec thing so it should be safe.

Notably, this fixes the propagation package also, though former is still there. brave.propagation instead of zipkin2.propagation

Finally, this allows you to control the underlying B3PropagationFactory, as mentioned 
https://github.com/spring-cloud/spring-cloud-sleuth/pull/1647

One subtle difference 2.2.x to 3.x is the change in the primary inject format.

2.2.x is
```java
return BaggagePropagation.newFactoryBuilder(B3Propagation.newFactoryBuilder()
                         .injectFormat(B3Propagation.Format.MULTI).build());
```

3.0 is
```java
return BaggagePropagation.newFactoryBuilder(B3Propagation.newFactoryBuilder()
                         .injectFormat(B3Propagation.Format.SINGLE_NO_PARENT).build());
```

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/trace/StackdriverTraceAutoConfiguration.java
```diff
@@ -221,8 +218,8 @@ public Sender stackdriverSender(GcpTraceProperties traceProperties,
 
 	@Bean
 	@ConditionalOnMissingBean
-	public Propagation.Factory stackdriverPropagation() {
-		return StackdriverTracePropagation.FACTORY;
+	public BaggagePropagation.FactoryBuilder baggagePropagationFactoryBuilder() {
+		return BaggagePropagation.newFactoryBuilder(StackdriverTracePropagation.FACTORY);
```
adriancole, line 222: right here will change with 0.17 we should use the correct brave.propagation package and add a comment that the underlying B3Propagation.Factory needs to be configured differently for sleuth 3.0

This can be a TODO I suppose as 3.0 isn't final anyway

spring-cloud-gcp-samples/spring-cloud-gcp-trace-sample/src/test/java/brave/internal/HexCodec.java
```diff
@@ -0,0 +1,112 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package brave.internal;
+
+// code originally imported from zipkin.Util
+public final class HexCodec {
```
adriancole, line 17: as this is src/test ... maybe just use Long.whatever? HexCodec is mostly for performance

meltsufin: I will remove this class altogether, since you've added it in `0.17.0`.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/trace/StackdriverTraceAutoConfiguration.java
meltsufin: Do you mean that the `B3Propagation.Factory` would be configured differently in zipkin-gcp, because we're not configuring it here?

meltsufin:
@adriancole Can you please do a final review of this PR to make sure I configured the propagation factory correctly. Thanks!

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/trace/StackdriverTraceAutoConfiguration.java
adriancole: I mean what we are overriding will be different configuration 2.2.x vs 3.0.x. Until we some annotation or otherwise to indicate the "real trace format" factory vs the wrapped one, this project will have to build differently for 2.2.x vs 3.0 or have some logic etc. @spencergibb mentioned maybe we can change 3.0 to use the `@Primary` annotation. Will think about it. Anyway, this is fine for 2.2.x

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/trace/StackdriverTraceAutoConfiguration.java
```diff
@@ -221,8 +219,9 @@ public Sender stackdriverSender(GcpTraceProperties traceProperties,
 
 	@Bean
 	@ConditionalOnMissingBean
-	public Propagation.Factory stackdriverPropagation() {
-		return StackdriverTracePropagation.FACTORY;
+	public BaggagePropagation.FactoryBuilder baggagePropagationFactoryBuilder() {
+		return BaggagePropagation.newFactoryBuilder(StackdriverTracePropagation.newFactory(
+				B3Propagation.newFactoryBuilder().injectFormat(B3Propagation.Format.MULTI).build()));
```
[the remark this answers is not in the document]
adriancole: we don't have a way to pass this yet. so this is good enough for 2.2.x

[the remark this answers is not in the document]
meltsufin: Thanks!

PR #2294, opened 2020-03-30
Fix Secret Manager integration test flakiness

dzou:
Fix the Secret Manager Integration Test flakiness; should fix the build.


dzou:
> Ah, nice! I was wondering if it was a flake; figured I'd let my commit run through to check.

Thanks, hope this works :)

Issue #2274, opened 2020-03-24
Documentation Structure Clean Up


saturnism:
**Describe the bug**
Spring Cloud GCP documentation structure is drifting from consistency.

For example:

* Google Cloud Pub/Sub
  * Sample
  * Pub/Sub Operations & Templates

* Stackdriver Logging
  * ...
  * Sample

* Cloud IAP
  * Configuration
  * Sample

* Secret Manager
  * Dependency Setup
  * Secret Manager Property Source. (that contains configuration)
  * ...
  * Sample Application

** Proposal **

* Consistent product naming: Cloud Pub/Sub, Cloud IAP, Cloud Secret Manager
* Consistent structure:
  * Implicit Introduction
  * Configuration (includes dependency, and table of properties)
  * ... <- functionalities
  * Sample

PR #2295, opened 2020-03-31
Reference Documentation Cleanup

meltsufin:
This is an attempt to improve consistency and structure.
It's by no means making it perfect, but hopefully moves it
in the right direction to counter the drift.

Fixes: #2274.

### DIFF
--- a/docs/src/main/asciidoc/trace.adoc
+++ b/docs/src/main/asciidoc/trace.adoc
@@ -1,4 +1,4 @@
-== Spring Cloud Sleuth
+== Stackdriver Trace
 
 https://cloud.spring.io/spring-cloud-sleuth/[Spring Cloud Sleuth] is an instrumentation framework for Spring Boot applications.
 It captures trace information and can forward traces to services like Zipkin for storage and analysis.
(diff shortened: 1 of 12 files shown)


docs/src/main/asciidoc/trace.adoc
```diff
@@ -1,4 +1,4 @@
-== Spring Cloud Sleuth
+== Stackdriver Trace
```
dzou: Maybe add an additional sentence in the first paragraph that mentions stackdriver as it opens with "Spring Cloud Sleuth" while the title is Stackdriver Trace.

elefeint: Technically Cloud Trace...

meltsufin: I know, I should probably use this opportunity to rename it across the board. We just can't rename class names and module names.

meltsufin: It's actually a substantial change that needs to be done carefully.
Let's do it separately,
https://github.com/spring-cloud/spring-cloud-gcp/issues/2296

meltsufin: Done. Take a look at the commit below.

PR #2302, opened 2020-04-03
Create protocol for specifying secrets' project and versions

dzou:
This creates a protocol for specifying secrets' projects and versions.

The protocol format uses the same conventions as https://github.com/GoogleCloudPlatform/github-actions/tree/master/get-secretmanager-secrets#inputs

```
# 1. Long form
spring.property=${secret-prefix:projects/<project-id>/secrets/<secret-id>/versions/<version-id>}

# 2.  Long form - "latest" version
spring.property=${secret-prefix:projects/<project-id>/secrets/<secret-id>}

# 3. Short form
spring.property=${secret-prefix:<project-id>/<secret-id>/<version-id>}

# 4. Short form - default project; specify secret + version
spring.property=${secret-prefix:<secret-id>/<version>} 

# 5. Shortest form - "latest" version and default project.
spring.property=${secret-prefix:<secret-id>}
```

Some further questions:
- How to introduce this feature? And how to deprecate old way of specifying secrets? Can they co-exist for some time?
- Which of the two approaches for specifying the default project do you like here: https://github.com/spring-cloud/spring-cloud-gcp/issues/2283#issuecomment-608629222 The OP makes a good point about how the `#4` form above might be better replaced by `<secret-id>/<version>` with it defaulting to a project instead. However this approach comes with the drawback that it diverges from the github actions convention.

- Perhaps people need `<secret-id>/<version-id>` more based on discussion offline.


spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/SecretManagerPropertySourceLocator.java
```diff
@@ -60,11 +61,22 @@ public void setVersions(Map<String, String> versions) {
...
 	public PropertySource<?> locate(Environment environment) {
-		return new SecretManagerPropertySource(
-				SECRET_MANAGER_NAME,
-				this.client,
-				this.projectIdProvider,
-				this.secretsPrefix,
-				this.versions);
+		CompositePropertySource compositePropertySource = new CompositePropertySource(SECRET_MANAGER_NAME);
+
+		compositePropertySource.addPropertySource(
```
elefeint, line 66: Do we want to have both resolvable? The new version is more comprehensive. Or are we doing it for backwards compatibility?

Perhaps we could automatically configure only the newest property source, and provide a "legacy" property for  loading the old one?

spring-cloud-gcp-samples/spring-cloud-gcp-secretmanager-sample/src/main/resources/bootstrap.properties
```diff
@@ -11,4 +11,9 @@ spring.cloud.gcp.secretmanager.secret-name-prefix=secrets.
 #spring.cloud.gcp.secretmanager.versions.application-secret=1
 
 # You can directly load the secret into a variable like this example
-my-application-secret=${secrets.application-secret}
+# This demonstrates multiple ways of loading the same application secret using template syntax.
+my-application-secret-1=${secrets.application-secret}
+my-application-secret-2=${gcp-secret/projects/my-kubernetes-codelab-217414/secrets/application-secret}
```
dzou, line 16: Consider different prefix usage

spring-cloud-gcp-samples/spring-cloud-gcp-secretmanager-sample/src/main/resources/bootstrap.properties
```diff
@@ -11,4 +11,9 @@ spring.cloud.gcp.secretmanager.secret-name-prefix=secrets.
...
+my-application-secret-3=${gcp-secret/projects/my-kubernetes-codelab-217414/secrets/application-secret/version/latest}
```
dzou, line 17: Can also be a project number for projects/ value

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/SecretManagerPropertyIdentifier.java
```diff
@@ -0,0 +1,112 @@
...
+import org.springframework.cloud.gcp.core.GcpProjectIdProvider;
+import org.springframework.util.Assert;
+
+/**
+ * Identifiers for specifying Secret Manager secrets using (project-id, secret-id, version).
+ *
+ * @author Daniel Zou
+ * @since 1.2.3
+ */
+public class SecretManagerPropertyIdentifier {
```
meltsufin, line 28: I'm not sure you really need this class. Can you just use the existing `SecretVersionName`?
I understand you need a place to put the `parseFromProperty` method somewhere, but maybe for now you can just put it as a static method in `SecretManagerAccessPropertySource`?

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/SecretManagerPropertySourceLocator.java
dzou: Yeah I wasn't sure how to handle this. I think would prefer if we only have the new version. Maybe if it's ok, I'd like to just remove the existing functionality and replace it with the new method. @meltsufin 

I guess it doesn't make sense for both to co-exist since the legacy one already does the work of downloading all the secrets, and then if the new system exists they might redownload secrets they already loaded into their properties.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/SecretManagerPropertyIdentifier.java
```diff
@@ -0,0 +1,112 @@
...
+	private final String projectId;
+
+	private final String secretId;
+
+	private final String version;
+
+	SecretManagerPropertyIdentifier(String projectId, String secretId, String version) {
+		Assert.notNull(projectId, "Project Id of GCP Secret Manager secret must not be null");
+		Assert.notNull(projectId, "Secret Id of GCP Secret Manager secret must not be null");
+		Assert.notNull(projectId, "Version of GCP Secret Manager secret must not be null");
```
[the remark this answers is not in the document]
dzou: Nice catch :)

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/SecretManagerPropertyIdentifier.java
```diff
@@ -0,0 +1,112 @@
...
+
+	private final String projectId;
+
+	private final String secretId;
+
+	private final String version;
+
+	SecretManagerPropertyIdentifier(String projectId, String secretId, String version) {
+		Assert.notNull(projectId, "Project Id of GCP Secret Manager secret must not be null");
+		Assert.notNull(projectId, "Secret Id of GCP Secret Manager secret must not be null");
```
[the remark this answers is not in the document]
dzou: Removed file in favor of using `SecretVersionName`

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/SecretManagerPropertyIdentifier.java
dzou: Good call, done.

dzou:
Redid the PR to just remove the legacy way of doing things and replace with new. If you feel good about this, I can move forward with doc updates.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/SecretManagerPropertySource.java
```diff
@@ -16,117 +16,119 @@
...
  * @since 1.2.2
  */
 public class SecretManagerPropertySource extends EnumerablePropertySource<SecretManagerServiceClient> {
 
-	private static final Log LOGGER = LogFactory.getLog(SecretManagerPropertySource.class);
-
-	private static final String LATEST_VERSION_STRING = "latest";
-
-	private final Map<String, Object> properties;
+	private static final String GCP_SECRET_PREFIX = "gcp-secret/";
```
meltsufin, line 37: Did the Secrets Manager team comment on this prefix? Have you looked into using `sm://`?

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/SecretManagerPropertySource.java
```diff
@@ -16,117 +16,119 @@
...
 	}
 
+	/**
+	 * The {@link SecretManagerPropertySource} is not enumerable, so this always returns an empty array.
+	 * @return the empty array.
+	 */
 	@Override
 	public String[] getPropertyNames() {
-		return propertyNames;
+		return new String[0];
```
meltsufin, line 68: Interesting that this still works.
Maybe we should log a warning when this method is called?

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/GcpSecretManagerProperties.java
```diff
@@ -39,17 +36,6 @@
 	 */
 	private String projectId;
 
-	/**
-	 * Defines a prefix String that will be prepended to the environment property names
-	 * of secrets in Secret Manager.
-	 */
-	private String secretNamePrefix = "";
```
meltsufin, line 46: Would it be so bad to let users overwrite the prefix?

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/SecretManagerPropertySource.java
```diff
@@ -16,117 +16,119 @@
...
-	@Override
-	public Object getProperty(String name) {
-		return properties.get(name);
+	private ByteString getSecretPayload(SecretVersionName secretIdentifier) {
+		try {
+			AccessSecretVersionResponse response = getSource().accessSecretVersion(secretIdentifier);
+			return response.getPayload().getData();
+		}
+		catch (NotFoundException e) {
+			return null;
```
meltsufin, line 77: Wouldn't this be considered an error if we got here? Not sure we should swallow the exception.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/SecretManagerPropertySource.java
```diff
@@ -16,117 +16,119 @@
...
+				&& tokens[0].equals("projects")
+				&& tokens[2].equals("secrets")
+				&& tokens[4].equals("versions")) {
+			// property is form "gcp-secret/projects/<project-id>/secrets/<secret-id>/versions/<version>"
+			projectId = tokens[1];
+			secretId = tokens[3];
+			version = tokens[5];
+		}
+		else {
+			return null;
```
meltsufin, line 125: Log an error or throw exception? Invalid format? etc.

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/SecretPropertySourceTests.java
```diff
@@ -0,0 +1,93 @@
...
+		String property = "spring.cloud.datasource";
+		SecretVersionName secretIdentifier =
+				SecretManagerPropertySource.parseFromProperty(property, DEFAULT_PROJECT_ID_PROVIDER);
+
+		assertThat(secretIdentifier).isNull();
+	}
+
+	@Test
+	public void testShortProperty_secretId() {
+		String property = "gcp-secret/the-secret";
```
meltsufin, line 41: Maybe add a test for just `"gcp-secret/"`?

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/SecretPropertySourceTests.java
```diff
@@ -0,0 +1,93 @@
...
+package org.springframework.cloud.gcp.autoconfigure.secretmanager;
+
+import com.google.cloud.secretmanager.v1beta1.SecretVersionName;
+import org.junit.Test;
+
+import org.springframework.cloud.gcp.core.GcpProjectIdProvider;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class SecretPropertySourceTests {
```
meltsufin, line 26: The name doesn't match the class `SecretManagerPropertySource`.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/SecretManagerPropertySource.java
```diff
@@ -16,117 +16,119 @@
...
-
-		ListSecretsPagedResponse response = client.listSecrets(ProjectName.of(projectId));
-		Map<String, Object> secretsMap = new HashMap<>();
-		for (Secret secret : response.iterateAll()) {
-			String secretId = extractSecretId(secret);
-			ByteString secretPayload = getSecretPayload(client, projectId, secretId, versions);
-			if (secretPayload != null) {
-				secretsMap.put(secretsPrefix + secretId, secretPayload);
-			}
+	static SecretVersionName parseFromProperty(String property, GcpProjectIdProvider projectIdProvider) {
```
meltsufin, line 81: Visible for testing?

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/SecretPropertySourceTests.java
```diff
@@ -0,0 +1,93 @@
...
+	@Test
+	public void testLongProperty_projectSecretVersion() {
+		String property = "gcp-secret/projects/my-project/secrets/the-secret/versions/3";
+		SecretVersionName secretIdentifier =
+				SecretManagerPropertySource.parseFromProperty(property, DEFAULT_PROJECT_ID_PROVIDER);
+
+		assertThat(secretIdentifier.getProject()).isEqualTo("my-project");
+		assertThat(secretIdentifier.getSecret()).isEqualTo("the-secret");
+		assertThat(secretIdentifier.getSecretVersion()).isEqualTo("3");
+	}
```
meltsufin, line 92: Missing a test for `getSecretPayload`.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/SecretManagerPropertySource.java
dzou: Colon can't be used in a property name because it [means something special in SPEL](https://stackoverflow.com/questions/31711308/spring-value-escape-colon-in-default-value). I wouldn't want to try to escape the colon either; I think it would look worse with the back slashes.

Am open to other secret prefixes though.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/SecretManagerPropertySource.java
dzou: Can't unfortunately. This method gets called by framework code somewhere and will pollute the logs.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/GcpSecretManagerProperties.java
dzou: I think it can only hurt users if they choose a bad prefix. It needs to be a unique/special string that lets us exit early so we know that it's definitely not a property we need to parse and attempt to query.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/SecretManagerPropertySource.java
meltsufin: I wonder if we can tap into that "special meaning" and configure it for our purpose.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/SecretManagerPropertySource.java
meltsufin: Maybe at `DEBUG` or `TRACE` level? How often will this be called?

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/SecretManagerPropertySource.java
dzou: Colon is like the null-safety operator. If the left side of the colon is null, then it defaults to the value on the right side.

meltsufin:
@dzou Please make sure the sample app integration tests work as well.
Also, I think at this point we should make the bootstrap auto-config enabled by default.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/SecretManagerPropertySource.java
meltsufin: This is only the case if SPEL is used, such as with `@Value`, right?
Notice that with `@Value` on a `Resource` it's interpreted as a protocol instead.
For example, see the "gs://" implementation we have for Storage.

dzou: Ok so I found something interesting. It seems like doing the `gcp-sm://` works but has slightly unexpected behavior.

- If the secret exists, then doing the `gcp-sm://path/to/secret` works fine. It will load the secret without issue.

- If the secret property source returns null for any secret property, then it will fallback to the evaluating the left-hand side and if that is null it will show the right hand side. See in our sample, we have secrets named `application-secret`, so in our sample, if the secrets are not present, you see that it will render `//application-secret` instead of the secret value.

image

So I guess if we move forward with throwing an exception in the cases where a secret is missing or the secret URL is not formatted correctly then user will never reach this unexpected behavior.

What are your thoughts?

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/SecretManagerPropertySource.java
dzou: Hmm I don't think it's worth it; feel like it will just add noise. It is called at least 10 times per test run in the integration tests.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/SecretManagerPropertySource.java
dzou: Sure, let's throw exception, this might be good for supporting the `gcp-sm://` prefix per our discussion above.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/SecretManagerPropertySource.java
dzou: Done.

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/SecretPropertySourceTests.java
dzou: Done.

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/SecretPropertySourceTests.java
dzou: Done.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/SecretManagerPropertySource.java
dzou: Added comment `//visible for testing`; can't add annotation though, it's guava.

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/SecretPropertySourceTests.java
dzou: We'll cover this directly in your configuration tests.

I don't think it's worth making it non-private to write tests for it unlike the `parseProperty` method which had a lot more code.

dzou:
> @dzou Please make sure the sample app integration tests work as well.
> Also, I think at this point we should make the bootstrap auto-config enabled by default.

Done.
Done - I actually just deleted the normal autoconfiguration class and we'll use the bootstrap by default?

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/GcpSecretManagerBootstrapConfiguration.java
```diff
@@ -48,7 +49,7 @@
 @Configuration
 @EnableConfigurationProperties(GcpSecretManagerProperties.class)
 @ConditionalOnClass(SecretManagerServiceClient.class)
-@ConditionalOnProperty("spring.cloud.gcp.secretmanager.bootstrap.enabled")
```
meltsufin, line 51: Docs need to be updated following this change.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/SecretManagerPropertySource.java
meltsufin: I'm not following fully. Let's discuss.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/GcpSecretManagerBootstrapConfiguration.java
dzou: Done, updated docs.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/SecretManagerPropertySource.java
dzou: I'll look a little more into to understand if it's fixable.

Just to summarize our discussion this morning, using colon has the consequence where if we ever return `null` for a secret, it will interpret the colon as the null safety operator. This is why it was printing out `//application-secret` for the secret that did not exist, because the original string was `gcp-sm://application-secret`; after resolving it to null, it treated the contents of the right of the colon as the "default value".

dzou: Ran the code through the debugger, the colon is a special character; i don't think this is configurable. I think we should avoid it to avoid unexpected behavior in the future with null values.

I would be open to other choices of prefixes, maybe `sm|` or `gcp-sm|` these seem to all work.

meltsufin: Should we ever return `null` though? Is there something wrong with throwing an exception for a property that doesn't exist?

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/GcpSecretManagerBootstrapConfiguration.java
```diff
@@ -48,7 +49,7 @@
...
+@ConditionalOnProperty(value = "spring.cloud.gcp.secretmanager.enabled", matchIfMissing = true)
 public class GcpSecretManagerBootstrapConfiguration {
 
 	private final GcpSecretManagerProperties properties;
```
meltsufin, line 55: Unused.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/SecretManagerPropertySource.java
dzou: Yeah I can see the argument where if a user specifies a secret using our syntax, there is the expectation that a secret is there; if it's missing then that's most likely an error and they probably don't want null there.

I would be comfortable with either approach. We can still throw exception using `|` too by the way.

To me it seems the tradeoff here is purely cosmetic. I.e. if the cosmetic benefits of using `:` outweighs the risk that we need to change behavior to return `null` someday.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/SecretManagerPropertySource.java
```diff
@@ -16,117 +16,129 @@
...
-		ListSecretsPagedResponse response = client.listSecrets(ProjectName.of(projectId));
-		Map<String, Object> secretsMap = new HashMap<>();
-		for (Secret secret : response.iterateAll()) {
-			String secretId = extractSecretId(secret);
-			ByteString secretPayload = getSecretPayload(client, projectId, secretId, versions);
-			if (secretPayload != null) {
-				secretsMap.put(secretsPrefix + secretId, secretPayload);
-			}
+	// Visible for Testing
+	static SecretVersionName parseFromProperty(String property, GcpProjectIdProvider projectIdProvider) {
```
meltsufin, line 78: Should this move to `SecretManagerTemplate`? We probably want to reconsider all those additional methods for specifying project and version and use the fully qualified secret identifiers instead.

docs/src/main/asciidoc/secretmanager.adoc
```diff
@@ -80,12 +99,14 @@ The auto-configured `SecretManagerTemplate` bean can be customized using configu
 By default, Spring Cloud GCP Secret Manager will authenticate using Application Default Credentials.
 This can be overridden using the authentication properties.
 
+NOTE: All of the below settings must be specified in a https://cloud.spring.io/spring-cloud-commons/multi/multi__spring_cloud_context_application_context_services.html#_the_bootstrap_application_context[`bootstrap.properties`] (or `bootstrap.yaml`) file which is the properties file used to configure settings for bootstrap-phase Spring configuration.
+
 |===
 | Name | Description | Required | Default value
```
meltsufin, line 105: Since now there is a single table of properties, we should move it up to be right under the dependency section, just like other chapters we have.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/SecretManagerPropertySource.java
meltsufin: What's the prefix that the Secret Manager team suggested we use? There's value in being compatible with other libraries for this product. Let's confirm with the product team.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/GcpSecretManagerBootstrapConfiguration.java
dzou: Done.

docs/src/main/asciidoc/secretmanager.adoc
dzou: Done.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/SecretManagerPropertySource.java
dzou: Done; I added these methods for getSecret, but for `createSecret` and `secretExists` it doesn't quite fit since those methods can accept project and secretId as valid inputs but are not fixed to a version.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/secretmanager/SecretManagerPropertySource.java
dzou: Based on discussion offline, looks like `sm://` has precedent of being used, so I just changed it to that for now.

spring-cloud-gcp-secretmanager/src/main/java/org/springframework/cloud/gcp/secretmanager/SecretManagerPropertySource.java
```diff
@@ -0,0 +1,73 @@
...
+			SecretManagerServiceClient client,
+			GcpProjectIdProvider projectIdProvider) {
+		super(propertySourceName, client);
+
+		this.projectIdProvider = projectIdProvider;
+	}
+
+	@Override
+	public Object getProperty(String name) {
+		SecretVersionName secretIdentifier =
```
meltsufin, line 49: Can we wire in the template and use it here?

spring-cloud-gcp-secretmanager/src/main/java/org/springframework/cloud/gcp/secretmanager/SecretManagerOperations.java
```diff
@@ -78,15 +94,13 @@
...
 	 *
-	 * @param secretId unique identifier of your secret in Secret Manager.
-	 * @param versionName which version of the secret to load. The version can be a version
-	 *     number as a string (e.g. "5") or an alias (e.g. "latest").
+	 * @param secretUri the Uri string identifying the secret.
+	 * 		See the javadoc for {@link SecretManagerOperations} for a summary of the URI syntax.
 	 * @return The secret payload as String
 	 */
-	String getSecretString(String secretId, String versionName);
+	String getSecretStringByUri(String secretUri);
```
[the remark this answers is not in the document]
dzou: Done.

spring-cloud-gcp-secretmanager/src/main/java/org/springframework/cloud/gcp/secretmanager/SecretManagerOperations.java
```diff
@@ -97,38 +111,13 @@
...
-	 * {@code versionName} for a specific {@code projectId}.
-	 *
-	 * @param secretId unique identifier of your secret in Secret Manager.
-	 * @param versionName which version of the secret to load. The version can be a version
-	 *     number as a string (e.g. "5") or an alias (e.g. "latest").
-	 * @param projectId unique identifier of your project.
-	 * @return The secret payload as {@link ByteString}
-	 */
-	ByteString getSecretByteString(String secretId, String versionName, String projectId);
+	byte[] getSecretBytesByUri(String secretUri);
```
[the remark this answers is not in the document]
dzou: Done.

spring-cloud-gcp-secretmanager/src/main/java/org/springframework/cloud/gcp/secretmanager/SecretManagerPropertySource.java
dzou: Done.

spring-cloud-gcp-secretmanager/src/main/java/org/springframework/cloud/gcp/secretmanager/SecretManagerPropertyUtils.java
```diff
@@ -0,0 +1,102 @@
...
+			projectId = tokens[1];
+			secretId = tokens[3];
+			version = tokens[5];
+		}
+		else {
+			throw new IllegalArgumentException(
+					"Unrecognized format for specifying a GCP Secret Manager secret: " + input);
+		}
+
+		Assert.isTrue(
```
meltsufin, line 84: `Assert.hasText` instead maybe?

dzou: Gotcha, i'll change it.

Let me submit this and i'll send a followup cleanup PR; some other stuff I want to cleanup too.

gsdatta:
Apologies in advance for commenting on such an old PR! I'm currently in the process of setting up a spring boot project on GCP app engine, which requires us us to use secrets manager as a replacement for env vars.

One of my worries here is that the forced protocol prefix adds to vendor lock-in that we'd like to avoid. I was hoping I could add a custom prefix (e.g. `MY_ENV_VAR_`) to make it _look like_ an environment variable (I initially misunderstood the `secret-name-prefix` property as a way to replace the `sm://`).

I'm curious to hear why the protocol prefix is necessary. Is it just to denote to the starter that it should be looked up via secrets manager? Can this protocol prefix be configurable?

dzou:
> Apologies in advance for commenting on such an old PR! I'm currently in the process of setting up a spring boot project on GCP app engine, which requires us us to use secrets manager as a replacement for env vars.
> 
> One of my worries here is that the forced protocol prefix adds to vendor lock-in that we'd like to avoid. I was hoping I could add a custom prefix (e.g. `MY_ENV_VAR_`) to make it _look like_ an environment variable (I initially misunderstood the `secret-name-prefix` property as a way to replace the `sm://`).
> 
> I'm curious to hear why the protocol prefix is necessary. Is it just to denote to the starter that it should be looked up via secrets manager? Can this protocol prefix be configurable?

@gsdatta -- The `sm://` prefix is necessary to indicate which property values should be looked up in Secret Manager. Basically all the property values with this prefix -- we assume that the string after the prefix corresponds to the secret's ID. We don't want to look up every value in an `application.properties` file in secret manager.

Hmm we could make the prefix configurable, but I have an alternative approach which may be simpler than a custom prefix like `MY_ENV_VAR_`.

What about a default value feature? Something like:

`spring.properties.password=${sm://my-secret:PASSWORD_ENV}` or `spring.properties.password=${PASSWORD_ENV:sm://my-secret}`

So we could check if the secret is present/null; if so, default to the provided value `PASSWORD_ENV`. Or the other way around in the second example.

This way you can keep the behavior with relying on env vars to be vendor-generic. Does that work for you?

gsdatta:
Makes sense. I think the default value structure would work for us quite well! Gives us the flexibility of overriding in certain environments as required when secrets manager may not be necessary.

dzou:
@gsdatta -- Thanks for the input. We're starting version `2.0.0` of our repo under the `GoogleCloudPlatform` organization; see: https://github.com/GoogleCloudPlatform/spring-cloud-gcp

We're going to start making all of our updates there.

I filed a new feature request for this in our new repo: https://github.com/GoogleCloudPlatform/spring-cloud-gcp/issues/213

dzou:
@gsdatta -- Hey I did some research, I think we have a decent way of doing this already in the current implementation.

You can try something like below; in the example `FOOBAR` is the environmental variable:

```
spring.password=${FOOBAR:${sm://application-secret/latest}}
```

In this case, if foobar is not null, it will use the value of `FOOBAR`. Otherwise, it will retrieve the secret manager secret. Let me know if this solution solves your problem.

PR #2306, opened 2020-04-07
Add secret manager bootstrap configuration unit test

meltsufin:
@dzou It's working yet. The property source doesn't seem to get registered at all.


dzou:
Thanks! This is very helpful, I can get the test to work.

> The property source doesn't seem to get registered at all.

I'll take a look at this - did you conclude this from the sample test not working or the unit test not working? I think the unit test currently doesn't work because the `secretManagerClient` mock must be configured to return a response when you retrieve the `my-secret` secret.

dzou:
Hmm I couldn't figure it out either actually - getting the same problem as you with the property source not being registered in the application context.

meltsufin:
@artembilan Do you have any thoughts on why the property source may not be getting registered via this test setup?

artembilan:
@meltsufin ,
I probably don't see the whole picture and it is hard to assume from so short change what is going on.
Probably you need to consult with the https://github.com/spring-cloud/spring-cloud-config project to be sure how to test those external sources...

dzou:
I think it might be because we would need to add the test's configuration class to `meta-inf/spring.factories`.

Like:
```
org.springframework.cloud.bootstrap.BootstrapConfiguration=com.example.TestConfiguraiton
```

But I guess we wouldn't want to add test configuration to `spring.factories`.

@artembilan - Is there another hook into setting `org.springframework.cloud.bootstrap.BootstrapConfiguration` programmatically from our test code?

artembilan:
See here: https://github.com/spring-cloud/spring-cloud-config/blob/master/spring-cloud-config-client/src/test/java/org/springframework/cloud/config/client/ConfigServerBootstrapConfigurationTests.java

I think that `parent/child` trick should be used over here, too.

You need to be sure that you load bootstrap with custom properties source *before* the actual application context.

dzou:
I wasn't able to get that method to work unfortunately, maybe still missing thing.

I wonder if it's ok to add a `test/resources/meta-inf/spring.factories`; registering the test configuration there made the test work.

meltsufin:
@artembilan @dzou It almost seems to work, but I can't seem to override the bootstrap autoconfig with mocks:
```java
	private SpringApplicationBuilder applicationBuilder = new SpringApplicationBuilder(
			TestBootstrapConfiguration.class)
			.child(GcpSecretManagerBootstrapConfiguration.class)
			.child(TestConfiguration.class)
			.web(WebApplicationType.NONE);
```

```
Caused by: io.grpc.StatusRuntimeException: NOT_FOUND: Secret [projects/235239892118/secrets/my-secret] not found or has no versions.
	at io.grpc.Status.asRuntimeException(Status.java:533)
	... 15 more
```

dzou:
Ahh Nice! I'll pull your branch and help take a look.

dzou:
I added the `spring.factories` in the `test/resources/` and it gets wired correctly. I guess the question is if there is an alternate method or if this is OK.

meltsufin:
> I added the `spring.factories` in the `test/resources/` and it gets wired correctly. I guess the question is if there is an alternate method or if this is OK.

I think that's fine.  I cleaned it up a bit more. Let's merge this in?

dzou:
> > I added the `spring.factories` in the `test/resources/` and it gets wired correctly. I guess the question is if there is an alternate method or if this is OK.
> 
> I think that's fine. I cleaned it up a bit more. Let's merge this in?

Sounds good, will do.

I just added another test case to verify fully-qualified secret url with project.

PR #2311, opened 2020-04-10
Cleanup Secret Manager module organization

dzou:
Some cleanup for secret manager module.

* Move property source integration test from autoconfigure/ to secret-manager/
* Change to using `Assert.hasText(...)`

### DIFF
--- a/spring-cloud-gcp-secretmanager/src/test/java/org/springframework/cloud/gcp/secretmanager/it/SecretManagerPropertySourceIntegrationTests.java
+++ b/spring-cloud-gcp-secretmanager/src/test/java/org/springframework/cloud/gcp/secretmanager/it/SecretManagerPropertySourceIntegrationTests.java
@@ -70,7 +70,7 @@ public class SecretManagerPropertySourceIntegrationTests {
 	@Before
 	public void setupSecretManager() {
 		ConfigurableApplicationContext context = new SpringApplicationBuilder()
-				.child(SecretManagerTestConfiguration.class)
+				.sources(SecretManagerTestConfiguration.class)
 				.web(WebApplicationType.NONE)
 				.run();
 
@@ -87,7 +87,7 @@ public class SecretManagerPropertySourceIntegrationTests {
 		createSecret(TEST_SECRET_ID, "the secret data.");
 
 		ConfigurableApplicationContext context = new SpringApplicationBuilder()
-				.child(SecretManagerTestConfiguration.class)
+				.sources(SecretManagerTestConfiguration.class)
 				.web(WebApplicationType.NONE)
 				.run();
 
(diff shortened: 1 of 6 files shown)


spring-cloud-gcp-secretmanager/src/test/java/org/springframework/cloud/gcp/secretmanager/it/SecretManagerPropertySourceIntegrationTests.java
```diff
@@ -101,28 +99,14 @@ public void testConfiguration() {
 		assertThat(byteArraySecret).isEqualTo("the secret data.".getBytes());
 	}
 
-	@Test
-	public void testConfigurationDisabled() {
```
meltsufin, line 105: We're losing this test now.

spring-cloud-gcp-secretmanager/src/test/java/org/springframework/cloud/gcp/secretmanager/it/SecretManagerPropertySourceIntegrationTests.java
```diff
@@ -49,9 +47,9 @@
 import static org.assertj.core.api.Assumptions.assumeThat;
 import static org.awaitility.Awaitility.await;
 
-public class SecretManagerIntegrationTests {
+public class SecretManagerPropertySourceIntegrationTests {
```
meltsufin, line 50: Would it be possible to add the `@Value` test here as well like what we're doing in the other PR:

```
	@Test
	public void test() {
		try (ConfigurableApplicationContext c = applicationBuilder.run()) {
			String secret = c.getBean("secret", String.class);
			assertThat(secret).isEqualTo("hello");
		}
	}

	@Configuration
	static class TestConfiguration {

		@Value("${sm://spring-cloud-gcp-it-secret")
		private String secret;

		@Bean
		public String secret() {
			return secret;
		}
	}
```

spring-cloud-gcp-secretmanager/src/test/java/org/springframework/cloud/gcp/secretmanager/it/SecretManagerPropertySourceIntegrationTests.java
dzou: Gotcha - I moved this test to the configuration unit tests you wrote: SecretManagerBootstrapConfigurationTests

spring-cloud-gcp-secretmanager/src/test/java/org/springframework/cloud/gcp/secretmanager/it/SecretManagerPropertySourceIntegrationTests.java
dzou: Done.

dzou:
I reworked the tests a bit -

I reduced the scope of the integration tests and now relying on the unit tests more since the integration tests had the problem were deleting and then recreating the secrets to clean the test environment was a flaky process. The tests to verify getting secrets for different projects and versions from the integration tests were moved to SecretManagerBootstrapConfigurationTests.java.

PR #2317, opened 2020-04-14
Remove DocBook which is no longer used

meltsufin:
It doesn't look like DocBook is needed anymore.
@artembilan Can someone from the Spring Team confirm?
Also, I added `<useNativeGit>true</useNativeGit>` configuration without which the generate docs command doesn't work for me.
```
 mvn clean install -Pdocs -pl docs
```


docs/pom.xml
```diff
@@ -102,6 +102,9 @@
 					<plugin>
 						<groupId>pl.project13.maven</groupId>
 						<artifactId>git-commit-id-plugin</artifactId>
+						<configuration>
+							<useNativeGit>true</useNativeGit>
```
artembilan, line 106: Do we really still need this `git-commit-id-plugin`?
Do you have some logic in docs based on the current commit status?

meltsufin: Yeah I don't think we really use it. I'll take it out.

Buzzardo: This mostly looks good to me. We've been abandoning Docbook and have (I think) gotten it out of all the projects now (though there may be old files sitting around, as here).

Why did you remove the two lines in docs/src/main/asciidoc/_configprops.adoc? Have we dropped those options?

meltsufin: @Buzzardo this file seems to be auto-updated every time you run the docs build, and indeed we have removed some properties in recent PRs.

Buzzardo: I did a little digging in the code and can see that spring.cloud.gcp.secretmanager.secret-name-prefix and spring.cloud.gcp.secretmanager.versions have been removed, so I'm good with dropping those two lines.

This PR looks good to me.

Thanks for the cleanup. Always nice to remove 3,700 unneeded lines.

PR #2319, opened 2020-04-14
Suppress exception in Pub/Sub adapter in AUTO_ACK and MANUAL modes

### LINKED ISSUE #2250: In AUTO_ACK mode message is immediately nack when exception occurs.
**According to AckMode.AUTO_ACK docs:**

The framework acks the {@link com.google.pubsub.v1.PubsubMessage} after it is
sent to the channel and processed successfully.
<p>The framework does not immediately nack the message when the exception occurs,
and allows the eventual redelivery to take effect.

**Actual behaviour:**

The message is immediately redelivered when the exception occurs.
For any exception PubSubInboundChannelAdapter throws PubSubException, 
due to this message is nacked MessageDispatcher.AckHandler in onFailure.

elefeint:
The current behavior of the Pub/Sub Spring Integration inbound adapter throwing an exception in AUTO_ACK and MANUAL modes results in client library reacting to the exception by nacking in its code.

This PR removes the exception and logs a warning instead, allowing the message to be redelivered according to Subscription ackDeadline setting and the client library's automatic deadline extension setting.

Fixes #2250.
Follow-up to #2074.

### DIFF
--- a/spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/pubsub/it/PubSubChannelAdaptersIntegrationTests.java
+++ b/spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/pubsub/it/PubSubChannelAdaptersIntegrationTests.java
@@ -240,10 +240,12 @@ public class PubSubChannelAdaptersIntegrationTests {
 				assertThat(numReceivedMessages.get()).isEqualTo(1);
 
 				// Expect redelivery within 2 minutes:
-				// 30 seconds subscription ackDeadline
+				// 10 seconds subscription ackDeadline
 				// + 60 seconds https://github.com/googleapis/java-pubsub/issues/141
-				// + 30 seconds anti-flake buffer
-				Awaitility.await().atMost(120, TimeUnit.SECONDS)
+				// + 20 seconds anti-flake buffer
+				Awaitility.await()
+						.atLeast(10, TimeUnit.SECONDS)
+						.atMost(90, TimeUnit.SECONDS)
 						.until(() -> numReceivedMessages.get() > 1);
 				assertThat(numReceivedMessages.get()).isEqualTo(2);
 			}
@@ -390,7 +392,7 @@ public class PubSubChannelAdaptersIntegrationTests {
 				GcpProjectIdProvider projectIdProvider,
 				CredentialsProvider credentialsProvider) {
 			if (pubSubAdmin.getSubscription(this.subscriptionName) == null) {
-				pubSubAdmin.createSubscription(this.subscriptionName, this.topicName, 30);
+				pubSubAdmin.createSubscription(this.subscriptionName, this.topicName, 10);
 			}
 
 			DefaultSubscriberFactory factory = new DefaultSubscriberFactory(projectIdProvider);
--- a/spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/pubsub/it/PubSubChannelAdaptersIntegrationTests.java
+++ b/spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/pubsub/it/PubSubChannelAdaptersIntegrationTests.java
@@ -239,12 +239,12 @@ public class PubSubChannelAdaptersIntegrationTests {
 						.until(() -> numReceivedMessages.get() > 0);
 				assertThat(numReceivedMessages.get()).isEqualTo(1);
 
-				// Expect redelivery within 2 minutes:
+				// Expect redelivery after at least 10 seconds but within 1.5 minutes:
 				// 10 seconds subscription ackDeadline
 				// + 60 seconds https://github.com/googleapis/java-pubsub/issues/141
 				// + 20 seconds anti-flake buffer
 				Awaitility.await()
-						.atLeast(10, TimeUnit.SECONDS)
+						.atLeast(9, TimeUnit.SECONDS)
 						.atMost(90, TimeUnit.SECONDS)
 						.until(() -> numReceivedMessages.get() > 1);
 				assertThat(numReceivedMessages.get()).isEqualTo(2);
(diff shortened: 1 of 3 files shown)


elefeint:
Yes, I am going to whip #2078 into shape next.

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/pubsub/it/PubSubChannelAdaptersIntegrationTests.java
```diff
@@ -201,6 +205,57 @@ public void sendAndReceiveMessageManualAck() {
...
+				// wait for initial delivery
+				Awaitility.await().atMost(10, TimeUnit.SECONDS)
+						.until(() -> numReceivedMessages.get() > 0);
+				assertThat(numReceivedMessages.get()).isEqualTo(1);
+
+				// Expect redelivery within 2 minutes:
+				// 30 seconds subscription ackDeadline
+				// + 60 seconds https://github.com/googleapis/java-pubsub/issues/141
+				// + 30 seconds anti-flake buffer
+				Awaitility.await().atMost(120, TimeUnit.SECONDS)
```
meltsufin, line 246: >It verifies that in AUTO_ACK mode, the message is neither acked nor nacked
How does it verify that it's not nacked?

If this test in fact takes around 2 minutes to pass, I think it's too much to add to the build.
Can we just verify that the message is not redelivered immediately because that was the original problem anyway?
In fact, it doesn't look like this test is verifying that aspect at all? The message can be redelivered immediately, and it will pass.

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/pubsub/it/PubSubChannelAdaptersIntegrationTests.java
```diff
@@ -201,6 +205,57 @@ public void sendAndReceiveMessageManualAck() {
...
+						.until(() -> numReceivedMessages.get() > 1);
+				assertThat(numReceivedMessages.get()).isEqualTo(2);
+			}
+			finally {
+				context.getBean(PubSubInboundChannelAdapter.class).stop();
+				PubSubAdmin pubSubAdmin = context.getBean(PubSubAdmin.class);
+				pubSubAdmin.deleteSubscription((String) context.getBean("subscriptionName"));
+				pubSubAdmin.deleteTopic((String) context.getBean("topicName"));
```
meltsufin, line 254: Can we do this in `@After` this kind of code seems to be repeated across tests.

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/pubsub/it/PubSubChannelAdaptersIntegrationTests.java
```diff
@@ -324,7 +390,7 @@ public SubscriberFactory defaultSubscriberFactory(
 				GcpProjectIdProvider projectIdProvider,
 				CredentialsProvider credentialsProvider) {
 			if (pubSubAdmin.getSubscription(this.subscriptionName) == null) {
-				pubSubAdmin.createSubscription(this.subscriptionName, this.topicName);
+				pubSubAdmin.createSubscription(this.subscriptionName, this.topicName, 30);
```
meltsufin, line 393: In the interests of making this test pass faster, can we go lower than 30s?

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/pubsub/it/PubSubChannelAdaptersIntegrationTests.java
elefeint: In theory, `assertThat(numReceivedMessages.get()).isEqualTo(1);` could catch immediate redelivery, but I don't think it would be reliable. 

I had another check there waiting for 10 seconds and making sure the message count was still 1, but that relied on Thread.sleep(). I think invariants verfiication was added into a later version of Awaitility... going to try using that.

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/pubsub/it/PubSubChannelAdaptersIntegrationTests.java
elefeint: Not the way it is right now, since it requires spring context to find the name of topic/subscription.
I can refactor the tests in a different PR, so that before/after are both done outside of the context.

elefeint: This would allow not relying on custom integration test configuration, I think. We could just rely on autoconfig.

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/pubsub/it/PubSubChannelAdaptersIntegrationTests.java
elefeint: Yes, 10 seconds is the minimum.

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/pubsub/it/PubSubChannelAdaptersIntegrationTests.java
elefeint: I'd like to verify eventual redelivery (provided I can do it without creating a flake) since we also had a bug where the message got accidentally acked, and so never redelivered.

elefeint: Actually, await().atLeast() does the job.

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/pubsub/it/PubSubChannelAdaptersIntegrationTests.java
elefeint: #2321 to track

PR #2322, opened 2020-04-15
rely on spring for nullability constraints checking

### LINKED ISSUE #2316: Getting a single entry findBy... results in EmptyResultDataAccessException
Hi *.

**Version: 1.2.2.RELEASE
JDK: 8
Spring Boot Version: 2.2.6.RELEASE
Kotlin: 1.3.71**

I have a simple repository like this:

```
interface ApiUserRepository : DatastoreRepository<ApiUser, Long> {
    fun findByUsername(username: String): ApiUser? // notice that return type is nullable
}
```

But when that method called (**AND** there is **no data** related with the username in datastore) the exception thrown:
`org.springframework.dao.EmptyResultDataAccessException: Expecting at least 1 result, but none found`

The exception message comes from [here](https://github.com/spring-cloud/spring-cloud-gcp/blob/master/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/PartTreeDatastoreQuery.java#L179). 

I also tried to change repository function return type to non-nullable, but it didn't help.

dmitry-s:
fixes #2316


docs/src/main/asciidoc/datastore.adoc
```diff
@@ -985,6 +985,36 @@ If a method's return type is `org.springframework.data.domain.Page`, the returne
 NOTE: Methods that return `Page` execute an additional query to compute total number of pages.
 Methods that return `Slice`, on the other hand, don't execute any additional queries and therefore are much more efficient.
 
+==== Null Handling of Repository Methods
+Java `java.util.Optional` can be used to indicate the potential absence of a return value.
+
+Alternatively, query methods can return the result without a wrapper.
+In that case the absence of a query result is indicated by returning `null`.
+Repository methods returning collections are guaranteed never to return `null` but rather the corresponding empty representation.
```
meltsufin, line 993:
```suggestion
Repository methods returning collections are guaranteed never to return `null` but rather the corresponding empty collection.
```

docs/src/main/asciidoc/datastore.adoc
```diff
@@ -985,6 +985,36 @@ If a method's return type is `org.springframework.data.domain.Page`, the returne
...
+
+==== Nullability Annotations
```
meltsufin, line 995: Is this supposed to be a subsection of the above or a new section?

docs/src/main/asciidoc/datastore.adoc
```diff
@@ -985,6 +985,36 @@ If a method's return type is `org.springframework.data.domain.Page`, the returne
...
+You can express nullability constraints for repository methods by using https://docs.spring.io/spring/docs/5.2.5.RELEASE/spring-framework-reference/core.html#null-safety[Spring Framework’s nullability annotations].
+They provide an opt-in null checks during runtime, as follows:
```
meltsufin, line 997:
```suggestion
They enable null checks during runtime, as follows:
```

docs/src/main/asciidoc/datastore.adoc
```diff
@@ -985,6 +985,36 @@ If a method's return type is `org.springframework.data.domain.Page`, the returne
...
+
+ . `@NonNullApi`: Used on the package level to declare that the default behavior for parameters and return values is to not accept or produce null values.
+
+ . `@NonNull`: Used on a parameter or return value that must not be null.
+
+ . `@Nullable`: Used on a parameter or return value that can be `null` (not needed on a parameter and return value where @NonNullApi applies).
```
meltsufin, line 1003:
```suggestion
 . `@Nullable`: Used on a parameter or return value that can be `null` (not needed on a parameter and return value where `@NonNullApi` applies).
```

docs/src/main/asciidoc/datastore.adoc
```diff
@@ -985,6 +985,36 @@ If a method's return type is `org.springframework.data.domain.Page`, the returne
...
+
+To enable runtime checking of nullability constraints for query methods, you need to activate non-nullability on the package level by using Spring’s `@NonNullApi` in `package-info.java`, as shown in the following example:
```
meltsufin, line 1005: Is there anything you can use at a class level or method level?
Wouldn't `@NonNull` on  a method work?
This sentence makes it sound like you can only make it work on methods by using the package-level annotation which is extremely inflexible and cumbersome.

spring-cloud-gcp-data-datastore/src/test/java/org/springframework/cloud/gcp/data/datastore/it/package-info.java
```diff
@@ -0,0 +1,5 @@
+/**
+ * IT tests for spring-cloud-gcp-data-datastore.
+ */
+@org.springframework.lang.NonNullApi
+package org.springframework.cloud.gcp.data.datastore.it;
```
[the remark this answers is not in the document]
dmitry-s: `@org.springframework.lang.NonNullApi` enables null checks

docs/src/main/asciidoc/datastore.adoc
dmitry-s: I think it's a separate section.

docs/src/main/asciidoc/datastore.adoc
dmitry-s: I think `@NonNull` should work on a method without `@NonNullApi`, but for some reason it doesn't. I'll look into that.

spring-cloud-gcp-data-datastore/src/test/java/org/springframework/cloud/gcp/data/datastore/repository/query/PartTreeDatastoreQueryTests.java
```diff
@@ -789,19 +789,6 @@ public void usingIdField() throws NoSuchMethodException {
 		assertThat(this.partTreeDatastoreQuery.execute(params)).isEqualTo(trade);
 	}
 
-
-	@Test
-	public void nonCollectionReturnTypeNoResults() throws NoSuchMethodException {
```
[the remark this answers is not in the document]
dmitry-s: The checks are now in AOP method interceptor. It would be hard to set up it in a unit test.

docs/src/main/asciidoc/datastore.adoc
meltsufin: My question is more on "should it be"?

docs/src/main/asciidoc/datastore.adoc
meltsufin: Let's resolve this before merging.

spring-cloud-gcp-data-datastore/src/test/java/org/springframework/cloud/gcp/data/datastore/repository/query/PartTreeDatastoreQueryTests.java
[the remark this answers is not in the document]
meltsufin: Can it be moved to an integration test then?

docs/src/main/asciidoc/datastore.adoc
dmitry-s: I don't think it should. It's not only about return values, but also about (non-)nullable parameters.

docs/src/main/asciidoc/datastore.adoc
dmitry-s: So turns out the class has to be annotated with `javax.annotation.Nonnull` in order for `@NonNull` to work on methods and parameters. This is not documented. I'm not sure if it's bag or undocumented feature.

meltsufin: Can you file it as bug/question on the Spring side just to make sure we're not missing anything.

docs/src/main/asciidoc/datastore.adoc
meltsufin: Whether an exception is thrown or null is returned seems directly related to the handling of empty results. Handling null parameters can be a separate section.

spring-cloud-gcp-data-datastore/src/test/java/org/springframework/cloud/gcp/data/datastore/repository/query/PartTreeDatastoreQueryTests.java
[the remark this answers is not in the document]
dmitry-s: I added an IT test for that.

docs/src/main/asciidoc/datastore.adoc
dmitry-s: I created an issue for Spring Data: https://jira.spring.io/browse/DATACMNS-1705

docs/src/main/asciidoc/datastore.adoc
```diff
@@ -985,6 +985,37 @@ If a method's return type is `org.springframework.data.domain.Page`, the returne
...
 
+==== Empty result handling in repository methods
+Java `java.util.Optional` can be used to indicate the potential absence of a return value.
+
+Alternatively, query methods can return the result without a wrapper.
+In that case the absence of a query result is indicated by returning `null`.
+Repository methods returning collections are guaranteed never to return `null` but rather the corresponding empty collection.
+
+===== Nullability Annotations
+You can express nullability constraints for repository methods by using https://docs.spring.io/spring/docs/5.2.5.RELEASE/spring-framework-reference/core.html#null-safety[Spring Framework’s nullability annotations].
```
meltsufin, line 996: I took another look at this documentation and realized that this is part of core Spring and is not specific to Spring Data at all. Hence, I don't think we need to make a big deal out of it here. I think it's enough to mention that if you want null checks use null-safety annotations and link to them.
Also, please use the non-version-specific link: https://docs.spring.io/spring/docs/current/spring-framework-reference/core.html#null-safety

docs/src/main/asciidoc/datastore.adoc
```diff
@@ -985,6 +985,8 @@ If a method's return type is `org.springframework.data.domain.Page`, the returne
 NOTE: Methods that return `Page` execute an additional query to compute total number of pages.
 Methods that return `Slice`, on the other hand, don't execute any additional queries and therefore are much more efficient.
 
+NOTE: You can enable nullability checks. For more details please see https://docs.spring.io/spring/docs/current/spring-framework-reference/core.html#null-safety[Spring Framework’s nullability docs].
```
meltsufin, line 988: I think you deleted too much. The part about empty result handling was useful in my opinion. I would just collapse the whole nullability section into what you have here. The rest was fine.

PR #2323, opened 2020-04-15
Refactor Pub/sub integration test to reduce duplication

elefeint:
1) removed `finally` blocks; setup and teardown are now done with junit.
2) Replaced custom output capture with a rule.
3) removed custom publisher/subscriber factory configuration; now that topic/subscription creation happens before each test, the customness is not necessary.

### DIFF
--- a/spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/pubsub/it/PubSubChannelAdaptersIntegrationTests.java
+++ b/spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/pubsub/it/PubSubChannelAdaptersIntegrationTests.java
@@ -16,44 +16,37 @@
 
 package org.springframework.cloud.gcp.autoconfigure.pubsub.it;
 
-import java.io.ByteArrayOutputStream;
-import java.io.PrintStream;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.UUID;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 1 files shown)


spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/pubsub/it/PubSubChannelAdaptersIntegrationTests.java
```diff
@@ -80,103 +72,120 @@
...
+		pubSubAdmin = new PubSubAdmin(
+				new DefaultGcpProjectIdProvider(),
+				new DefaultCredentialsProvider(() -> new Credentials())
+		);
 	}
 
-	@AfterClass
-	public static void resetStdout() {
-		System.setOut(systemOut);
+	@Before
```
meltsufin, line 106: Do we want to do this per test, or once for the whole class?

elefeint: I would at least create a subscription per test; otherwise each test has to take care to never have anything redelivered.

At that point, might as well create a fresh topic.

meltsufin: fair enough... feel free to merge.

PR #2332, opened 2020-04-24
Turn hardcoded checkstyle versions into properties

elefeint:
Most of spring-cloud-gcp inherits checkstyle rules and maven plugin versions from `spring-cloud-build`, which defines them as variables `maven-checkstyle-plugin.version` and `puppycrawl-tools-checkstyle.version`. This enables checkstyle's CI to override our plugin/rules version with the latest to test for regressions prior to releasing (see checkstyle/checkstyle#8106 for additional context).

The samples directory is different -- we inherit from Spring Boot starter parent instead, and there is a hardcoded version of the checkstyle plugin.

This PR:
* removes the hardcoding
* introduces variables to control the samples' checkstyle versions, identical to the main project
* sets variable default to be the same as in `spring-cloud-build` version we are on right now

### DIFF
--- a/spring-cloud-gcp-samples/pom.xml
+++ b/spring-cloud-gcp-samples/pom.xml
@@ -63,7 +63,6 @@
 					<plugin>
 						<groupId>org.apache.maven.plugins</groupId>
 						<artifactId>maven-checkstyle-plugin</artifactId>
-						<version>3.0.0</version>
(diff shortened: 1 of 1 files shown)


elefeint:
The versions did not change, but this makes it possible to control them from command-line arguments.

PR #2333, opened 2020-04-24
Return fake zero to avoid tripping nullability checks

elefeint:
Should fix the build.

```
[ERROR] testParallelOperations  Time elapsed: 0.243 s  <<< ERROR!
org.springframework.dao.EmptyResultDataAccessException: Result must not be null!
	at org.springframework.cloud.gcp.data.spanner.repository.it.ParallelSpannerRepositoryIntegrationTests.testParallelOperations(ParallelSpannerRepositoryIntegrationTests.java:59)
```

The other way this could get fixed is to override the `performReadOnlyTransaction` in the test repository and explicitly mark it nullable.

### DIFF
--- a/spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/repository/it/ParallelSpannerRepositoryIntegrationTests.java
+++ b/spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/repository/it/ParallelSpannerRepositoryIntegrationTests.java
@@ -66,7 +66,7 @@ public class ParallelSpannerRepositoryIntegrationTests extends AbstractSpannerIn
 				assertThat(repo.count()).isZero();
 			});
 			assertThat(repo.count()).isZero();
-			return null;
+			return 0;
 		});
 
 		executeInParallel(unused -> assertThat(this.tradeRepository.countByAction("BUY")).isEqualTo(PARALLEL_OPERATIONS));


spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/repository/it/ParallelSpannerRepositoryIntegrationTests.java
```diff
@@ -66,7 +66,7 @@ public void testParallelOperations() {
 				assertThat(repo.count()).isZero();
 			});
 			assertThat(repo.count()).isZero();
-			return null;
+			return 0;
```
dmitry-s, line 69: What were the errors that we are fixing?

elefeint:
```
[ERROR] testParallelOperations  Time elapsed: 0.243 s  <<< ERROR!
org.springframework.dao.EmptyResultDataAccessException: Result must not be null!
	at org.springframework.cloud.gcp.data.spanner.repository.it.ParallelSpannerRepositoryIntegrationTests.testParallelOperations(ParallelSpannerRepositoryIntegrationTests.java:59)
```

The test entity is shared among several tests, so the `@NotNull` did its job too well.

PR #2335, opened 2020-04-27
Datastore Projections should restrict query to contain only the necessary fields

### LINKED ISSUE #2145: Datastore Projections should restrict query to contain only the necessary fields

dmitry-s:
fixes #2145

### DIFF
--- a/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/PartTreeDatastoreQuery.java
+++ b/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/PartTreeDatastoreQuery.java
@@ -16,6 +16,7 @@
 
 package org.springframework.cloud.gcp.data.datastore.repository.query;
 
+import java.beans.PropertyDescriptor;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.Iterator;
@@ -32,6 +33,8 @@ import java.util.stream.StreamSupport;
 import com.google.cloud.datastore.Cursor;
 import com.google.cloud.datastore.EntityQuery;
 import com.google.cloud.datastore.KeyValue;
+import com.google.cloud.datastore.ProjectionEntityQuery;
+import com.google.cloud.datastore.Query;
 import com.google.cloud.datastore.StructuredQuery;
 import com.google.cloud.datastore.StructuredQuery.Builder;
 import com.google.cloud.datastore.StructuredQuery.CompositeFilter;
@@ -54,6 +57,8 @@ import org.springframework.data.domain.Slice;
 import org.springframework.data.domain.SliceImpl;
 import org.springframework.data.domain.Sort;
 import org.springframework.data.mapping.PropertyPath;
+import org.springframework.data.projection.ProjectionFactory;
+import org.springframework.data.projection.ProjectionInformation;
 import org.springframework.data.repository.query.ParameterAccessor;
 import org.springframework.data.repository.query.ParametersParameterAccessor;
 import org.springframework.data.repository.query.parser.Part;
@@ -61,6 +66,7 @@ import org.springframework.data.repository.query.parser.PartTree;
 import org.springframework.data.repository.query.parser.PartTree.OrPart;
 import org.springframework.util.Assert;
 
+import static com.google.cloud.datastore.Query.newEntityQueryBuilder;
 import static org.springframework.data.repository.query.parser.Part.Type.GREATER_THAN;
 import static org.springframework.data.repository.query.parser.Part.Type.GREATER_THAN_EQUAL;
 import static org.springframework.data.repository.query.parser.Part.Type.LESS_THAN;
@@ -83,6 +89,8 @@ public class PartTreeDatastoreQuery<T> extends AbstractDatastoreQuery<T> {
 
 	private final DatastorePersistentEntity datastorePersistentEntity;
 
+	private final ProjectionFactory projectionFactory;
+
 	private List<Part> filterParts;
 
 	private  static final Map<Part.Type, BiFunction<String, Value, PropertyFilter>> FILTER_FACTORIES =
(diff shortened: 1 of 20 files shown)


spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/PartTreeDatastoreQuery.java
```diff
@@ -223,8 +233,34 @@ else if (isCountingQuery && this.tree.isDelete()) {
 		}
 	}
 
+	private Supplier<Builder<?>> getQueryBuilderSupplier(Object[] parameters) {
+		Supplier<Builder<?>> queryBuilderSupplier;
+		queryBuilderSupplier = StructuredQuery::newEntityQueryBuilder;
```
meltsufin, line 238: Combine with the line above?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/PartTreeDatastoreQuery.java
```diff
@@ -99,15 +107,17 @@
 	 * @param datastoreTemplate used to execute the given query.
 	 * @param datastoreMappingContext used to provide metadata for mapping results to objects.
 	 * @param entityType the result domain type.
+	 * @param projectionFactory the projection factory.
```
dzou, line 110: Maybe add an additional few words describing what the purpose of projection factory/what its used for here?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/PartTreeDatastoreQuery.java
```diff
@@ -223,8 +233,34 @@ else if (isCountingQuery && this.tree.isDelete()) {
 		}
 	}
 
+	private Supplier<Builder<?>> getQueryBuilderSupplier(Object[] parameters) {
```
dzou, line 236: This `parameters` variable is never used?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/PartTreeDatastoreQuery.java
```diff
@@ -223,8 +233,34 @@ else if (isCountingQuery && this.tree.isDelete()) {
...
+		Supplier<Builder<?>> queryBuilderSupplier;
+		queryBuilderSupplier = StructuredQuery::newEntityQueryBuilder;
+		ProjectionInformation projectionInformation =
+				this.projectionFactory.getProjectionInformation(this.queryMethod.getReturnedObjectType());
+
+		if (projectionInformation != null &&
```
dzou, line 242: In what cases would the returned `projectInformation` be null?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/PartTreeDatastoreQuery.java
dzou, line 236: Maybe add a javadoc comment to help describe what this helper method is doing? I guess it's looking at what fields will be needed from the return obj and then constructing a query object containing only those fields?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/PartTreeDatastoreQuery.java
```diff
@@ -223,8 +233,34 @@ else if (isCountingQuery && this.tree.isDelete()) {
...
+	private String mapToFieldName(PropertyDescriptor propertyDescriptor) {
+		String name = propertyDescriptor.getName();
+		DatastorePersistentProperty persistentProperty =
+				(DatastorePersistentProperty) this.datastorePersistentEntity.getPersistentProperty(name);
+		return persistentProperty.getFieldName();
+	}
+
 	private Slice executeSliceQuery(Object[] parameters) {
-		EntityQuery.Builder builder = StructuredQuery.newEntityQueryBuilder()
+		StructuredQuery.Builder builder = getQueryBuilderSupplier(parameters).get()
```
dzou, line 263: This is optional but I am starting to get confused with the Supplier thing - maybe if you find you're always calling `getQueryBuilderSupplier(parameters).get()` here, you can take the opportunity to just have `getQueryBuilder()` return the query builder directly rather than a supplier and then call `.get()`.

But if it's gonna be a big refactor job then just ignore this comment.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/PartTreeDatastoreQuery.java
dmitry-s: it happens in some tests

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/PartTreeDatastoreQuery.java
```diff
@@ -176,37 +185,29 @@ public Object execute(Object[] parameters) {
...
-		}
-		else if (this.tree.isExistsProjection()) {
-			collector = Collectors.collectingAndThen(Collectors.counting(), (count) -> count > 0);
-		}
-		else if (!returnedTypeIsNumber) {
-			queryBuilderSupplier = StructuredQuery::newEntityQueryBuilder;
-			mapper = this::processRawObjectForProjection;
-		}
+		StructuredQuery.Builder<?> structuredQueryBuilder =
+				!((isCountingQuery && !this.tree.isDelete()) || this.tree.isExistsProjection())
```
meltsufin, line 195: There's way too much conditional logic in this method and many fragments are repeated.
I think it needs to be refactored to be more readable and maintainable.
We can do it in a meeting, if you want.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/PartTreeDatastoreQuery.java
```diff
@@ -223,8 +224,49 @@ else if (isCountingQuery && this.tree.isDelete()) {
 		}
 	}
 
+	private Collector<?, ?, ?> getCollector(boolean isCountingQuery) {
```
meltsufin, line 227: Can we come up with a better name for the method?

PR #2352, opened 2020-04-30
pubsub emulator rule moved to a separate project

### LINKED ISSUE #2351: Make PubSub emulator rule public
Creating new module for the future emulators support.

dmitry-s:
fixes #2351

cc/ @elefeint

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -45,6 +45,7 @@
 		<module>spring-cloud-gcp-bigquery</module>
 		<module>spring-cloud-gcp-security-firebase</module>
 		<module>spring-cloud-gcp-secretmanager</module>
+		<module>spring-cloud-gcp-test-support</module>
 	</modules>
 
 	<properties>
--- a/pom.xml
+++ b/pom.xml
@@ -45,7 +45,7 @@
 		<module>spring-cloud-gcp-bigquery</module>
 		<module>spring-cloud-gcp-security-firebase</module>
 		<module>spring-cloud-gcp-secretmanager</module>
-		<module>spring-cloud-gcp-test-support</module>
+		<module>spring-cloud-gcp-test</module>
 	</modules>
 
 	<properties>
--- a/spring-cloud-gcp-test/pom.xml
+++ b/spring-cloud-gcp-test/pom.xml
@@ -10,7 +10,7 @@
     </parent>
 
 
-    <artifactId>spring-cloud-gcp-test-support</artifactId>
+    <artifactId>spring-cloud-gcp-test</artifactId>
     <name>Spring Cloud GCP Test Support</name>
     <description>Provides tools for testing Spring Cloud GCP projects</description>
     <dependencies>
--- a/spring-cloud-gcp-test/src/main/java/org/springframework/cloud/gcp/test/PubSubEmulator.java
+++ b/spring-cloud-gcp-test/src/main/java/org/springframework/cloud/gcp/test/PubSubEmulator.java
@@ -187,7 +187,6 @@ public class PubSubEmulator extends ExternalResource {
 		else {
 			createConfig();
 		}
-
 	}
 
 	/**
@@ -213,7 +212,7 @@ public class PubSubEmulator extends ExternalResource {
 	 * to fail the test.
 	 */
 	private void createConfig() throws InterruptedException {
-		int attempts = 10;
+		int attempts = 100;
 		while (!Files.exists(EMULATOR_CONFIG_PATH) && --attempts >= 0) {
 			Thread.sleep(1000);
 		}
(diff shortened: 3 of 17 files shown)


pom.xml
```diff
@@ -45,6 +45,7 @@
 		<module>spring-cloud-gcp-bigquery</module>
 		<module>spring-cloud-gcp-security-firebase</module>
 		<module>spring-cloud-gcp-secretmanager</module>
+		<module>spring-cloud-gcp-test-support</module>
```
meltsufin, line 48: Let's rename to `spring-cloud-gcp-test`.

spring-cloud-gcp-test/pom.xml
```diff
@@ -0,0 +1,29 @@
...
+    </parent>
+
+
+    <artifactId>spring-cloud-gcp-test</artifactId>
+    <name>Spring Cloud GCP Test Support</name>
+    <description>Provides tools for testing Spring Cloud GCP projects</description>
+    <dependencies>
+		<dependency>
+			<groupId>junit</groupId>
+			<artifactId>junit</artifactId>
```
meltsufin, line 19: `test` scope?

spring-cloud-gcp-test/pom.xml
```diff
@@ -0,0 +1,29 @@
...
+		</dependency>
+
+		<dependency>
+			<groupId>org.springframework</groupId>
+			<artifactId>spring-jcl</artifactId>
+			<scope>compile</scope>
```
meltsufin, line 25: Compile scope is the default, you don't have to specify it.

spring-cloud-gcp-test/src/main/java/org/springframework/cloud/gcp/test/PubSubEmulator.java
```diff
@@ -87,7 +86,7 @@ public PubSubEmulator() {
 	@Override
 	protected void before() throws IOException, InterruptedException {
 
-		assumeTrue("PubSubEmulator rule disabled. Please enable with -Dit.pubsub-emulator.", this.enableTests);
+		Assume.assumeTrue("PubSubEmulator rule disabled. Please enable with -Dit.pubsub-emulator.", this.enableTests);
```
meltsufin, line 89: I don't think that enabling it with `-Dit.pubsub-emulator` makes sense when you're offering it as a generic test utility. That flag was specific for our project.

spring-cloud-gcp-test/src/main/java/org/springframework/cloud/gcp/test/PubSubEmulator.java
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-package org.springframework.cloud.gcp.stream.binder.pubsub;
+package org.springframework.cloud.gcp.test;
```
meltsufin, line 17:
```suggestion
package org.springframework.cloud.gcp.test.pubsub;
```

spring-cloud-gcp-test/pom.xml
elefeint: Not in this case -- the test project itself is meant to be imported as test scope, but the rule extends a JUnit class, so it needs to be compile, not test scope.

spring-cloud-gcp-test/src/main/java/org/springframework/cloud/gcp/test/PubSubEmulator.java
elefeint: It's an interesting question then to test whether the rule initialization or a test's `@BeforeClass` will run first. We would not want to do emulator startup if the tests themselves are disabled.

If there is possibility of the rule running first, perhaps we can just document this flag well?

spring-cloud-gcp-test/pom.xml
meltsufin: I wonder if we should consider making these utilities independent of JUnit. People use different test frameworks.

spring-cloud-gcp-test/src/main/java/org/springframework/cloud/gcp/test/PubSubEmulator.java
meltsufin: We can also try not to implement JUnit test rule.
That's what we ended up doing for Spring Cloud Function [here](https://github.com/spring-cloud/spring-cloud-function/blob/master/spring-cloud-function-adapters/spring-cloud-function-adapter-gcp/src/test/java/org/springframework/cloud/function/adapter/gcp/integration/LocalServerTestSupport.java).

spring-cloud-gcp-test/pom.xml
eddumelendez: make it generic so people can reuse it sounds good. Can you take into account that may some people will be able to contribute creating rules for junit4 or extension for junit5 specifically? Those can be issues with `ideal for contribution` or `help wanted`. WDYT?

dmitry-s: That a good idea! We should do that.

dmitry-s:
superseded by #2357

PR #2354, opened 2020-05-04
Spring Integration doc formatting update

dzou:
Updates the level of nesting of Spring integration samples sections in refdoc.

Because these two docs are composed to make the Spring integration docs, it will make it more clear from the index that the samples refer to pub/sub and storage respectively.

See currently: https://cloud.spring.io/spring-cloud-gcp/reference/html/#spring-integration

### DIFF
--- a/docs/src/main/asciidoc/spring-integration-pubsub.adoc
+++ b/docs/src/main/asciidoc/spring-integration-pubsub.adoc
@@ -309,7 +309,7 @@ The first patterns have precedence over the following ones.
 In the previous example, the `"*"` pattern means every header is mapped.
 However, because it comes last in the list, https://docs.spring.io/spring-integration/api/org/springframework/integration/util/PatternMatchUtils.html#smartMatch-java.lang.String-java.lang.String...-[the previous patterns take precedence].
 
-==== Pub/Sub Channel Adapter Samples
+==== Samples
 
 Available examples:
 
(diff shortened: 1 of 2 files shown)


docs/src/main/asciidoc/spring-integration-storage.adoc
```diff
@@ -95,6 +95,6 @@ public MessageHandler outboundChannelAdapter(Storage gcs) {
 }
 ----
 
-=== Sample
+==== Samples
```
meltsufin, line 98: I think I need a screenshot to see what effect this has. Also, why make it plural for a single sample?

docs/src/main/asciidoc/spring-integration-pubsub.adoc
```diff
@@ -309,10 +309,10 @@ The first patterns have precedence over the following ones.
 In the previous example, the `"*"` pattern means every header is mapped.
 However, because it comes last in the list, https://docs.spring.io/spring-integration/api/org/springframework/integration/util/PatternMatchUtils.html#smartMatch-java.lang.String-java.lang.String...-[the previous patterns take precedence].
 
-=== Sample
+==== Samples
```
meltsufin, line 312: Can you show a screenshot comparing before and after?

dzou: Before:
image

After:
image

I prefer the After here because it is more clear what the sample is a sample of rather than the sample appearing at a equivalent level to the other top-level sections inside spring-integration.

docs/src/main/asciidoc/spring-integration-storage.adoc
dzou: Done and done (see below).

PR #2356, opened 2020-05-06
Add SpannerOptions auto-configuration for emulator

eddumelendez:
Currently, in order to execute against the emulator, Spanner relies
on `SPANNER_EMULATOR_HOST` environment variable. This commit introduces
new configuration properties in order to enable and set the emulator host.

### DIFF
--- a/docs/src/main/asciidoc/spanner.adoc
+++ b/docs/src/main/asciidoc/spanner.adoc
@@ -88,6 +88,8 @@ The schema for the tables will be "ON DELETE NO ACTION" if `false`. | No | `true
 | `spring.cloud.gcp.spanner.writeSessionsFraction` | Fraction of sessions to be kept prepared for write transactions | No | 0.2 - Determined by Cloud Spanner client library
 | `spring.cloud.gcp.spanner.keepAliveIntervalMinutes` | How long to keep idle sessions alive | No | 30 - Determined by Cloud Spanner client library
 | `spring.cloud.gcp.spanner.failIfPoolExhausted` |  If all sessions are in use, fail the request by throwing an exception. Otherwise, by default, block until a session becomes available. | No | `false`
+| `spring.cloud.gcp.spanner.emulator.enabled` |  Enables the usage of an emulator. If this is set to true, then you should set the `spring.cloud.gcp.spanner.host-port` to the host:port of your locally running emulator instance. | No | `false`
+| `spring.cloud.gcp.spanner.host-port` |  The host and port of the Spanner service; can be overridden to specify connecting to an already-running https://cloud.google.com/spanner/docs/emulator#installing_and_running_the_emulator[Spanner emulator] instance. | No |
 |===
 
 ==== Repository settings
@@ -1435,10 +1437,11 @@ This command can be used to create Cloud Spanner instances:
 $ gcloud spanner instances create <instance-name> --config=emulator-config --description="<description>" --nodes=1
 ----
 
-Remember that you need to set the `SPANNER_EMULATOR_HOST` environment variable to use the emulator:
+Once the Spanner emulator is running, ensure that the following properties are set in your `application.properties` of your Spring application:
 
 ----
-$ export SPANNER_EMULATOR_HOST=localhost:9010
+spring.cloud.gcp.spanner.emulator.enabled=true
+spring.cloud.gcp.spanner.host-port=${EMULATOR_HOSTPORT}
 ----
 
 === Sample
--- a/docs/src/main/asciidoc/spanner.adoc
+++ b/docs/src/main/asciidoc/spanner.adoc
@@ -88,8 +88,8 @@ The schema for the tables will be "ON DELETE NO ACTION" if `false`. | No | `true
 | `spring.cloud.gcp.spanner.writeSessionsFraction` | Fraction of sessions to be kept prepared for write transactions | No | 0.2 - Determined by Cloud Spanner client library
 | `spring.cloud.gcp.spanner.keepAliveIntervalMinutes` | How long to keep idle sessions alive | No | 30 - Determined by Cloud Spanner client library
 | `spring.cloud.gcp.spanner.failIfPoolExhausted` |  If all sessions are in use, fail the request by throwing an exception. Otherwise, by default, block until a session becomes available. | No | `false`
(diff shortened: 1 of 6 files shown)


spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/spanner/GcpSpannerEmulatorAutoConfiguration.java
```diff
@@ -0,0 +1,57 @@
...
+		this.properties = properties;
+	}
+
+	@Bean
+	@ConditionalOnMissingBean
+	public SpannerOptions spannerOptions() {
+		SpannerOptions.Builder builder = SpannerOptions.newBuilder()
+				.setProjectId(this.properties.getProjectId())
+				.setCredentials(NoCredentials.getInstance());
+		if (this.properties.getEmulatorHost() != null) {
```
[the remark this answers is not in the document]
eddumelendez: the client library resolves the `SPANNER_EMULATOR_HOST` environment variable. But, I agree it can be dismiss. I think we can validate if the environment variable or the property is set, otherwise we can fail and say `neither environment variable or property was set`. WDYT?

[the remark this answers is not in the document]
meltsufin: Since we're providing `NoCredentials`, there is no risk of connecting to the real Spanner and incurring costs.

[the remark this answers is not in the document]
dmitry-s: In regard to  ".enabled vs the property" - .enabled is the way to go.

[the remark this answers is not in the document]
elefeint: Yes, let's fail if "enabled" is set but "emulatorHost" is missing, since there is no way to make that configuration work. `Assert.notEmpty` should work.

eddumelendez:
PR is updated with docs and also in order to keep the consistency between emulator rename the previous `emulator-host` to `host-port`.

docs/src/main/asciidoc/spanner.adoc
```diff
@@ -1435,10 +1437,11 @@ This command can be used to create Cloud Spanner instances:
 $ gcloud spanner instances create <instance-name> --config=emulator-config --description="<description>" --nodes=1
 ----
 
-Remember that you need to set the `SPANNER_EMULATOR_HOST` environment variable to use the emulator:
```
dmitry-s, line 1438: Not sure if we want to remove this. 
Setting/unsetting the variable is a quick way to switch all your tests to using the emulator and back to normal. Might be still useful.

docs/src/main/asciidoc/spanner.adoc
```diff
@@ -88,6 +88,8 @@ The schema for the tables will be "ON DELETE NO ACTION" if `false`. | No | `true
 | `spring.cloud.gcp.spanner.writeSessionsFraction` | Fraction of sessions to be kept prepared for write transactions | No | 0.2 - Determined by Cloud Spanner client library
 | `spring.cloud.gcp.spanner.keepAliveIntervalMinutes` | How long to keep idle sessions alive | No | 30 - Determined by Cloud Spanner client library
 | `spring.cloud.gcp.spanner.failIfPoolExhausted` |  If all sessions are in use, fail the request by throwing an exception. Otherwise, by default, block until a session becomes available. | No | `false`
+| `spring.cloud.gcp.spanner.emulator.enabled` |  Enables the usage of an emulator. If this is set to true, then you should set the `spring.cloud.gcp.spanner.host-port` to the host:port of your locally running emulator instance. | No | `false`
+| `spring.cloud.gcp.spanner.host-port` |  The host and port of the Spanner service; can be overridden to specify connecting to an already-running https://cloud.google.com/spanner/docs/emulator#installing_and_running_the_emulator[Spanner emulator] instance. | No |
```
meltsufin, line 92: @dmitry-s @elefeint Can we pick a consistent name for the emulator `host` property?
I see `spring.cloud.gcp.datastore.host` and `spring.cloud.gcp.pubsub.emulator-host` and `spring.cloud.gcp.firestore.host-port`.

docs/src/main/asciidoc/spanner.adoc
meltsufin: I'm fine with removing it. Even though it's available, it's not consistent with the other emulators. We can leave a note about it, but I'm afraid it will only confuse people.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/spanner/GcpSpannerEmulatorAutoConfiguration.java
```diff
@@ -0,0 +1,56 @@
...
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.util.Assert;
+
+/**
+ * Provides auto-configuration to use the Spanner emulator if enabled.
+ *
+ * @author Eddú Meléndez
```
meltsufin, line 33: Please add
`@since 1.2.3`

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/spanner/GcpSpannerProperties.java
```diff
@@ -70,6 +71,8 @@
 	// Otherwise, by default, block until a session becomes available.
 	private boolean failIfPoolExhausted = false;
 
+	private String hostPort;
```
meltsufin, line 74: Please add a comment above like the other props.

eddumelendez:
don't mind at all. I'd prefer to use `emulator.host` but I think we need to align with for the other emulator properties too. WDYT?

PR #2370, opened 2020-05-18
[DO NOT MERGE] set an old version of spring-cloud-sleuth

elefeint:
`org.springframework.cloud:spring-cloud-sleuth:2.2.2.RELEASE` uses a version of brave that's compatible with the one used by `io.zipkin.gcp:zipkin-sender-stackdriver:0.16.0`.

Let's keep this as a backup plan if we can't fix the incompatibilities with the latest brave.


meltsufin:
Let's close this.

PR #2385, opened 2020-05-21
Make 403 an allowable Pub/Sub UP status

### LINKED ISSUE #2374: PubSubHealthIndicator reported false alarm when pubsub was actually UP
PubSubHealthIndicator reports DOWN but pubsub is actually UP in some situations. Error message:

`pubSub: {
status: "DOWN",
details: {
error: "com.google.api.gax.rpc.PermissionDeniedException: io.grpc.StatusRuntimeException: PERMISSION_DENIED: User not authorized to perform this action."
}
},`

According to the code here:
https://github.com/spring-cloud/spring-cloud-gcp/blob/master/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/health/PubSubHealthIndicator.java#L56

The health indicator only reports UP if pubsub returns 404 NOT_FOUND when the indicator tries to pull an non-existing subscription id.

It worked before however it failed from yesterday probably because of a recent GCP pubsub change. So now it instead of returning 404, returns a PERMISSION_DENIED 403 for a non-existing subscription. Therefore health check failed. But the connection to pubsub is actually UP.

The fix could be expanding the checking boundary so that all 4xx response codes are considered pubsub UP.

elefeint:
Starting this week, Cloud Pub/Sub accounts without _pubsub.subscriptions.list_ permission (normally granted through _Pub/Sub Viewer_ role) receive 403 Permission Denied instead of 404 Resource Not Found.

This PR makes 403 a valid healthcheck UP status and updates documentation to make actuator use more prominent.

Fixes #2374.

### DIFF
--- a/docs/src/main/asciidoc/pubsub.adoc
+++ b/docs/src/main/asciidoc/pubsub.adoc
@@ -25,6 +25,27 @@ dependencies {
 
 This starter is also available from https://start.spring.io[Spring Initializr] through the `GCP Messaging` entry.
 
+=== Spring Boot Actuator Support
+
+==== Cloud Pub/Sub Health Indicator
+
+If you are using Spring Boot Actuator, you can take advantage of the Cloud Pub/Sub health indicator called `pubsub`.
+The health indicator will verify whether Cloud Pub/Sub is up and accessible by your application.
+To enable it, all you need to do is add the https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#production-ready[Spring Boot Actuator] to your project.
+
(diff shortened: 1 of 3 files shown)


docs/src/main/asciidoc/pubsub.adoc
```diff
@@ -25,6 +25,27 @@ dependencies {
 
 This starter is also available from https://start.spring.io[Spring Initializr] through the `GCP Messaging` entry.
 
+=== Spring Boot Actuator Support
```
meltsufin, line 28: Why did you move this section up?
Our convention is to have configuration after deps in refdoc.

elefeint: I'll move it back. My thought was that configuration is really long for pubsub, and this needed to be more prominent, but maybe not.

Issue #1761, opened 2019-07-23
Stackdriver Monitoring Exporter for Micrometer


saturnism:
Explore/add idiomatic integration w/ Stackdriver Monitoring and custom metrics. Typically in GKE deployments, users can use Micrometer and export metrics via Prometheus, and have Prometheus agents scrape & propagate metrics to Stackdriver Monitoring.

However, there are cases where it's better to propagate metrics directly to the service (e.g., when running in App Engine/serverless environment) to export metrics directly to Stackdriver Monitoring.

Micrometer can have custom exporters, it's possible to have a monitoring starter that automatically configure Micrometer and export to Stackdriver.

bbzg:
Does anyone know the status of this? Is it still not provided out of the box? Thanks.

meltsufin:
I haven't tired it, but there's [Stackdriver integration in the Micrometer project](https://github.com/micrometer-metrics/micrometer/tree/master/implementations/micrometer-registry-stackdriver/src/main/java/io/micrometer/stackdriver).

bbzg:
Yes I am using those, but I can't get them to "autowire" by just having them on the class path like other spring-cloud-gcp things. Would be nice if they could be detected and autowired. Unless someone already did that and I just missed some dependency...

meltsufin:
We certainly don't have a Spring Boot Starter for it yet.
Contributions are welcomed!

meltsufin:
Related: https://github.com/micrometer-metrics/micrometer-docs/pull/115

eddumelendez:
there is currently support for `stackdriver` in `spring-boot` you can see the docs [here](https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#production-ready-metrics-export-stackdriver)

#2391 tries to solve the `project-id` and `credentials` auto-detection.

PR #2391, opened 2020-05-25
Add Micrometer AutoConfiguration

eddumelendez:
Currently, `project-id` has to be provided in order to make the app works. This auto-configuration provides the support to set a specific `project-id` and `credentials` or spin up with the project's default values.

### DIFF
--- a/docs/src/main/asciidoc/metrics.adoc
+++ b/docs/src/main/asciidoc/metrics.adoc
@@ -1,6 +1,6 @@
 == Stackdriver Monitoring
 
-Google Cloud Platform provides a managed distributed tracing service called https://cloud.google.com/monitoring/[Stackdriver Monitoring], and https://micrometer.io/docs/registry/stackdriver[Micrometer] can be used with it to easily instrument Spring Boot applications for observability.
+Google Cloud Platform provides a service called https://cloud.google.com/monitoring/[Stackdriver Monitoring], and https://micrometer.io/docs/registry/stackdriver[Micrometer] can be used with it to easily instrument Spring Boot applications for observability.
 
 Spring Boot already provides auto-configuration for Stackdriver. This module enables to auto-detect the `project-id` and `credentials`. Also, it can be customized.
 
@@ -23,7 +23,7 @@ dependencies {
 }
 ----
 
-You must enable Stackdriver Monitoring API from the Google Cloud Console in order to capture traces.
+You must enable Stackdriver Monitoring API from the Google Cloud Console in order to capture metrics.
 Navigate to the https://console.cloud.google.com/apis/api/monitoring.googleapis.com/overview[Stackdriver Monitoring API] for your project and make sure it’s enabled.
 
 === Spring Boot Starter for Stackdriver Monitoring
--- a/docs/src/main/asciidoc/metrics.adoc
+++ b/docs/src/main/asciidoc/metrics.adoc
@@ -1,0 +1,45 @@
+== Stackdriver Monitoring
+
+Google Cloud Platform provides a managed distributed tracing service called https://cloud.google.com/monitoring/[Stackdriver Monitoring], and https://micrometer.io/docs/registry/stackdriver[Micrometer] can be used with it to easily instrument Spring Boot applications for observability.
+
+Spring Boot already provides auto-configuration for Stackdriver. This module enables to auto-detect the `project-id` and `credentials`. Also, it can be customized.
+
+Maven coordinates, using <<getting-started.adoc#_bill_of_materials, Spring Cloud GCP BOM>>:
+
+[source,xml]
+----
+<dependency>
+    <groupId>org.springframework.cloud</groupId>
+    <artifactId>spring-cloud-gcp-starter-metrics</artifactId>
+</dependency>
+----
+
+Gradle coordinates:
+
+[source,subs="normal"]
+----
+dependencies {
+    compile group: 'org.springframework.cloud', name: 'spring-cloud-gcp-starter-metrics'
+}
+----
+
+You must enable Stackdriver Monitoring API from the Google Cloud Console in order to capture traces.
+Navigate to the https://console.cloud.google.com/apis/api/monitoring.googleapis.com/overview[Stackdriver Monitoring API] for your project and make sure it’s enabled.
+
+=== Spring Boot Starter for Stackdriver Monitoring
+
+Spring Boot Starter for Stackdriver Monitoring uses Micrometer.
+
+All configurations are optional:
+
+|===
+| Name | Description | Required | Default value
+| `spring.cloud.gcp.metrics.enabled` | Auto-configure Micrometer to send metrics to Stackdriver Monitoring. | No | `true`
+| `spring.cloud.gcp.metrics.project-id` | Overrides the project ID from the <<spring-cloud-gcp-core,Spring Cloud GCP Module>> | No |
+| `spring.cloud.gcp.metrics.credentials.location` | Overrides the credentials location from the <<spring-cloud-gcp-core,Spring Cloud GCP Module>> | No |
+| `spring.cloud.gcp.metrics.credentials.encoded-key` | Overrides the credentials encoded key from the <<spring-cloud-gcp-core,Spring Cloud GCP Module>> | No |
+| `spring.cloud.gcp.metrics.credentials.scopes` | Overrides the credentials scopes from the <<spring-cloud-gcp-core,Spring Cloud GCP Module>> | No |
+|===
+
+You can use core Spring Boot Actuator properties to control reporting frequency, etc.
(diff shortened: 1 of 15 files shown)


meltsufin:
@eddumelendez Thank you for the contribution. This seems related to #1761.
We'll take a look at it shortly.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/metrics/GcpStackdriverMetricsAutoConfiguration.java
```diff
@@ -0,0 +1,74 @@
...
+ * Provides auto-detection for project.
+ *
+ * @author Eddú Meléndez
+ */
+@Configuration
+@AutoConfigureBefore(StackdriverMetricsExportAutoConfiguration.class)
+@ConditionalOnClass({StepMeterRegistry.class, StackdriverConfig.class})
+@ConditionalOnBean(Clock.class)
+@EnableConfigurationProperties({GcpMetricsProperties.class, StackdriverProperties.class})
+@ConditionalOnProperty(value = "spring.cloud.gcp.metrics.enabled", matchIfMissing = true)
```
saturnism, line 49: should this have `havingValue = "true"`?

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/metrics/GcpStackdriverMetricsAutoConfiguration.java
```diff
@@ -0,0 +1,74 @@
...
+		this.stackdriverProperties = stackdriverProperties;
+		this.projectId = (gcpMetricsProperties.getProjectId() != null)
+				? gcpMetricsProperties.getProjectId() : gcpProjectIdProvider.getProjectId();
+		this.credentialsProvider = gcpMetricsProperties.getCredentials().hasKey()
+				? new DefaultCredentialsProvider(gcpMetricsProperties) : credentialsProvider;
+	}
+
+	@Bean
+	@ConditionalOnMissingBean
+	public StackdriverConfig stackdriverConfig() {
```
saturnism, line 70: do we also need to produce the `StackdriverMeterRegistry`? Or has Spring Boot Actuator been updated to produce it automatically? (It was an open issue)

saturnism:
awesome stuff! Thank you @eddumelendez.  Just a few follow up questions. Thnx.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/metrics/GcpStackdriverMetricsAutoConfiguration.java
eddumelendez: it is not required since we are performing this auto-configuration before the one provided by spring-boot. So, we just need to pre-configure the config so spring-boot takes it

eddumelendez: just added an assertion to make sure the `meterRegistry` bean is created

docs/src/main/asciidoc/metrics.adoc
```diff
@@ -0,0 +1,45 @@
...
+
+Maven coordinates, using <<getting-started.adoc#_bill_of_materials, Spring Cloud GCP BOM>>:
+
+[source,xml]
+----
+<dependency>
+    <groupId>org.springframework.cloud</groupId>
+    <artifactId>spring-cloud-gcp-starter-metrics</artifactId>
+</dependency>
+----
```
elefeint, line 15: Users would also need a dependency on actuator starter; otherwise `Clock` bean will not be autoconfigured, preventing our autoconfig from triggering.

```
    <dependency>
      <artifactId>spring-boot-starter-actuator</artifactId>
    </dependency>
```

Or should this be in the starter, since `spring-cloud-gcp-starter-metrics` is not useful without actuator?

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/metrics/GcpStackdriverMetricsAutoConfiguration.java
```diff
@@ -0,0 +1,74 @@
...
+
+/**
+ * Provides auto-detection for project.
+ *
+ * @author Eddú Meléndez
+ */
+@Configuration
+@AutoConfigureBefore(StackdriverMetricsExportAutoConfiguration.class)
+@ConditionalOnClass({StepMeterRegistry.class, StackdriverConfig.class})
+@ConditionalOnBean(Clock.class)
```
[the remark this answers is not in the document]
eddumelendez: Don't think so. Since all of them are provided for `spring-boot` once `actuator`, `micrometer` and `micrometer-stackdriver` are declared. When the conditions meet we this auto-config triggers.

docs/src/main/asciidoc/metrics.adoc
eddumelendez: That's right, `actuator` is required. Even though the new starter will not make sense without the `actuator` dependency, I would prefer to let this clear in the docs so the developers are aware about what dependencies are declaring for this case. WDYT?

elefeint: Yes, I agree. Just add a mention of it to the refdoc.
Otherwise, looks great -- I tested the starter in a tiny sample app.

docs/src/main/asciidoc/metrics.adoc
```diff
@@ -26,6 +26,26 @@ dependencies {
 You must enable Stackdriver Monitoring API from the Google Cloud Console in order to capture metrics.
 Navigate to the https://console.cloud.google.com/apis/api/monitoring.googleapis.com/overview[Stackdriver Monitoring API] for your project and make sure it’s enabled.
 
+This starter depends on `org.springframework.boot:spring-boot-starter-actuator` dependency. Make sure the dependency is being declared.
```
elefeint, line 29:
```suggestion
This starter requires `org.springframework.boot:spring-boot-starter-actuator` dependency to activate. Make sure the dependency is being declared.
```

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/metrics/GcpMetricsProperties.java
```diff
@@ -0,0 +1,51 @@
...
+package org.springframework.cloud.gcp.autoconfigure.metrics;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.boot.context.properties.NestedConfigurationProperty;
+import org.springframework.cloud.gcp.core.Credentials;
+import org.springframework.cloud.gcp.core.CredentialsSupplier;
+import org.springframework.cloud.gcp.core.GcpScope;
+
+@ConfigurationProperties(prefix = "spring.cloud.gcp.metrics")
+public class GcpMetricsProperties implements CredentialsSupplier {
```
meltsufin, line 26: Need to add `@since 1.2.4`. Same goes for all other new class files.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/metrics/GcpStackdriverMetricsAutoConfiguration.java
saturnism: last time i used this, i needed to create my own registry bean, and documented here: https://micrometer.io/docs/registry/stackdriver

https://github.com/micrometer-metrics/micrometer-docs/pull/115#pullrequestreview-380850792

I'm unsure if actuator has been updated to autoconfigure the registry bean

eddumelendez: this is the current auto-configuration for stackdriver in spring-booot https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/stackdriver/StackdriverMetricsExportAutoConfiguration.java

saturnism: Oh nice!! Fixed in 2.3 👏👏

Issue #1682, opened 2019-05-13
Update App Engine app.yaml to use App Engine Java 11


saturnism:
There are currently some examples that are using App Engine Flex for deployment. App Engine Java 11 can deploy fat JARs going forward. It would be great to update & test the samples to use App Engine Java 11 when it's beta.

https://github.com/spring-cloud/spring-cloud-gcp/search?q=flex&unscoped_q=flex

ttomsu:
Clarified with Ray: the request is also to move from GAE _Flex_ on Java 8 to GAE _Standard_ on Java 11.

PR #2395, opened 2020-05-28
Updates GAE samples and juggles some plugin bits

ttomsu:
Fixes https://github.com/spring-cloud/spring-cloud-gcp/issues/1682

Two commits here:

1. Based on my preliminary research and a long discussion with the team yesterday around Maven plugins, I pulled the `spring-boot-maven-plugin` into the `spring-cloud-gcp-samples` parent POM. It looked like all modules under `spring-cloud-gcp-samples` were Spring Boot applications.
    * I also pulled up the `appengine-maven-plugin` into the parent POM's `pluginManagement` now that I understand what that element does. In this case, I'm setting the version in one spot. It also appears this plugin is also used in the top-level `spring-cloud-gcp` POM, where I also updated it to the latest version. Based on my understanding now of `plugin` vs `pluginManagement`, I wonder why this plugin is needed here?
    * I removed a few `maven-deploy-plugin`s too - it's my understanding that this is only needed if these are to be published, which I don't think our samples need to be. To add to that, the configurations all had `skip=true`, which means they weren't doing anything anyway (maybe this was done to explicitly _disable_ publishing?). I checked up the POM hierarchy and it doesn't appear that any parent POM applies this plugin (please doublecheck me on this though).

1. The second commit updates the 3 GAE Flex runnable samples to use GAE Standard with Java 11.  This was actually the driving force behind the first commit, as the JAR needs to be executable for GAE Standard, and the `spring-boot-maven-plugin` is what makes that happen (through a hook into the `package` goal).

### DIFF
--- a/spring-cloud-gcp-samples/pom.xml
+++ b/spring-cloud-gcp-samples/pom.xml
@@ -25,6 +25,7 @@
 		<spring-cloud-build-tools.version>3.0.0-SNAPSHOT</spring-cloud-build-tools.version>
 		<puppycrawl-tools-checkstyle.version>8.32</puppycrawl-tools-checkstyle.version>
 		<maven-checkstyle-plugin.version>3.1.1</maven-checkstyle-plugin.version>
+		<app-engine-maven-plugin.version>2.2.0</app-engine-maven-plugin.version>
 	</properties>
 
 	<modules>
@@ -116,12 +117,26 @@
 	</profiles>
 
 	<build>
+		<pluginManagement>
+			<plugins>
+				<plugin>
+					<groupId>com.google.cloud.tools</groupId>
+					<artifactId>appengine-maven-plugin</artifactId>
+					<version>${app-engine-maven-plugin.version}</version>
+				</plugin>
+			</plugins>
+		</pluginManagement>
 		<plugins>
 			<plugin>
 				<artifactId>maven-deploy-plugin</artifactId>
 				<configuration>
 					<skip>true</skip>
 				</configuration>
+			</plugin>
+			<plugin>
+				<!-- Enables repackaging the JAR into an executable. -->
+				<groupId>org.springframework.boot</groupId>
+				<artifactId>spring-boot-maven-plugin</artifactId>
 			</plugin>
 		</plugins>
 	</build>
--- a/spring-cloud-gcp-samples/spring-cloud-gcp-logging-sample/README.adoc
+++ b/spring-cloud-gcp-samples/spring-cloud-gcp-logging-sample/README.adoc
@@ -25,9 +25,17 @@ $ curl --header "x-cloud-trace-context: [Some-Trace-ID]" localhost:8080/log
 To see the logged lines, navigate to the https://console.cloud.google.com/logs/viewer[Logging section of the Cloud Console].
 Select the `Global` resource and look under the `spring.log` channel to see the logged text.
 
+=== Running on Google Cloud Platform
+
+Deploy to Google AppEngine:
+
+----
+$ mvn clean package appengine:deploy -Dapp.deploy.projectId=$PROJECT_ID -Dapp.deploy.version=$VERSION
+----
+
 If your application is deployed to Google App Engine (GAE) or Google Kubernetes Engine (GKE), you will instead find the logs under the deployment's resource in the Stackdriver UI under the `spring.log` channel.
 You can further narrow down the log by request by using the following filtering text that takes advantage of the trace ID label in each logged line.
-For example, if deployed to App Engine (by using `mvn appengine:deploy` in this sample) you can use the following to filter by the trace ID:
+For example, if deployed to App Engine, you can use the following to filter by the trace ID:
 
 ----
  resource.type="gae_app"
@@ -38,4 +46,4 @@ For example, if deployed to App Engine (by using `mvn appengine:deploy` in this 
 Expand the log entry labels and note that `appengine.googleapis.com/trace_id` label is set to the same value for both log messages.
 The trace ID header `x-cloud-trace-context` is set automatically when deployed to App Engine.
 
-NOTE: When running on GAE or GKE, also remember to remove the `CONSOLE` appender, otherwise logs will appear on Cloud Logging in duplicate: once in the `spring.log` stream and once in the `stdout` stream.
+NOTE: When running on GAE or GKE, remember to remove the `CONSOLE` appender. Otherwise, logs are duplicated in Cloud Logging (once in the `spring.log` stream and once in the `stdout` stream).
--- a/spring-cloud-gcp-samples/spring-cloud-gcp-logging-sample/README.adoc
+++ b/spring-cloud-gcp-samples/spring-cloud-gcp-logging-sample/README.adoc
@@ -27,7 +27,7 @@ Select the `Global` resource and look under the `spring.log` channel to see the 
 
 === Running on Google Cloud Platform
 
-Deploy to Google AppEngine:
+Deploy to App Engine:
 
 ----
 $ mvn clean package appengine:deploy -Dapp.deploy.projectId=$PROJECT_ID -Dapp.deploy.version=$VERSION
--- a/spring-cloud-gcp-samples/spring-cloud-gcp-logging-sample/pom.xml
+++ b/spring-cloud-gcp-samples/spring-cloud-gcp-logging-sample/pom.xml
@@ -69,7 +69,6 @@
             <plugin>
                 <groupId>com.google.cloud.tools</groupId>
                 <artifactId>appengine-maven-plugin</artifactId>
-                <version>1.3.2</version>
             </plugin>
             <plugin>
                 <artifactId>maven-deploy-plugin</artifactId>
--- a/spring-cloud-gcp-samples/spring-cloud-gcp-logging-sample/src/main/appengine/app.yaml
+++ b/spring-cloud-gcp-samples/spring-cloud-gcp-logging-sample/src/main/appengine/app.yaml
@@ -1,9 +1,5 @@
-runtime: java
-env: flex
+runtime: java11
+instance_class: F1
 
-handlers:
-- url: /.*
-  script: this field is required, but ignored
-
-manual_scaling:
-  instances: 1
+env_variables:
+  JAVA_TOOL_OPTIONS: "-XX:MaxRAM=256m -XX:ActiveProcessorCount=2 -Xmx32m"
--- a/spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-stream-binder-functional-sample/pom.xml
+++ b/spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-stream-binder-functional-sample/pom.xml
@@ -32,4 +32,16 @@
     </dependencies>
   </dependencyManagement>
 
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.springframework.boot</groupId>
+        <artifactId>spring-boot-maven-plugin</artifactId>
+        <configuration>
+          <skip>true</skip>
+        </configuration>
+      </plugin>
+    </plugins>
+  </build>
+
 </project>
--- a/spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-stream-binder-functional-sample/spring-cloud-gcp-pubsub-stream-binder-functional-sample-test/pom.xml
+++ b/spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-stream-binder-functional-sample/spring-cloud-gcp-pubsub-stream-binder-functional-sample-test/pom.xml
@@ -47,6 +47,4 @@
     </dependency>
 
   </dependencies>
-
-
 </project>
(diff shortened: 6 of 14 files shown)


dzou:
@ttomsu 
> I also pulled up the appengine-maven-plugin into the parent POM's pluginManagement now that I understand what that element does. In this case, I'm setting the version in one spot. It also appears this plugin is also used in the top-level spring-cloud-gcp POM, where I also updated it to the latest version. Based on my understanding now of plugin vs pluginManagement, I wonder why this plugin is needed here?

Hmm good point, I also don't think the `appengine-maven-plugin` would be needed in the root-level but am not completely sure. @meltsufin @elefeint - maybe you two might know

> I removed a few maven-deploy-plugins too - it's my understanding that this is only needed if these are to be published, which I don't think our samples need to be.

Yup good call on removing those. Yes, the `skip=true` thing was put there to override the parent's plugin configuration so it wouldn't get deployed. However, because the samples parent has the `maven-deploy-plugin` with `skip=true` then the children of the samples shouldn't need to redeclare that.

Very nice!

@dmitry-s 
> I'm not sure if we want to remove the plugins from the samples' pom files. Currently the samples demonstrate which dependencies and plugins are required and after this PR this won't be true anymore.

I think it's ok, it seems the plugins being removed here are really just the extra the `maven-deploy-plugin` declarations.

spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-stream-binder-functional-sample/spring-cloud-gcp-pubsub-stream-binder-functional-sample-test/pom.xml
```diff
@@ -48,5 +48,16 @@
 
   </dependencies>
 
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.springframework.boot</groupId>
+        <artifactId>spring-boot-maven-plugin</artifactId>
+        <configuration>
+          <skip>true</skip>
```
dzou, line 57: Was there an issue with excluding this from the pom?

ttomsu: Yeah - since this module only has a `src/test` folder, the `spring-boot-maven-plugin` complained it couldn't find a `Main` class.

dzou: Gotcha I see.

dzou:
LGTM pending [resolution of the Travis CI errors](https://travis-ci.org/github/spring-cloud/spring-cloud-gcp/builds/692235401). Looks like something related to `spring-cloud-gcp-pubsub-stream-binder-functional-sample-test`.

spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-stream-binder-functional-sample/spring-cloud-gcp-pubsub-stream-binder-functional-sample-test/pom.xml
ttomsu: I pulled this one level up, because I think that build error was caused by the JAR not looking right after being repackaged.

spring-cloud-gcp-samples/pom.xml
```diff
@@ -116,12 +117,20 @@
...
+					<artifactId>appengine-maven-plugin</artifactId>
+					<version>${app-engine-maven-plugin.version}</version>
+				</plugin>
+			</plugins>
+		</pluginManagement>
 		<plugins>
 			<plugin>
-				<artifactId>maven-deploy-plugin</artifactId>
-				<configuration>
-					<skip>true</skip>
```
meltsufin, line 123: I believe this was necessary to skip publishing the module when the Spring Team runs `mvn deploy`. The default of the plugin is to not skip.

spring-cloud-gcp-samples/spring-cloud-gcp-logging-sample/README.adoc
```diff
@@ -25,9 +25,17 @@ $ curl --header "x-cloud-trace-context: [Some-Trace-ID]" localhost:8080/log
 To see the logged lines, navigate to the https://console.cloud.google.com/logs/viewer[Logging section of the Cloud Console].
 Select the `Global` resource and look under the `spring.log` channel to see the logged text.
 
+=== Running on Google Cloud Platform
+
+Deploy to Google AppEngine:
```
meltsufin, line 30:
```suggestion
Deploy to App Engine:
```

spring-cloud-gcp-samples/spring-cloud-gcp-logging-sample/README.adoc
```diff
@@ -25,9 +25,17 @@ $ curl --header "x-cloud-trace-context: [Some-Trace-ID]" localhost:8080/log
...
+
+----
+$ mvn clean package appengine:deploy -Dapp.deploy.projectId=$PROJECT_ID -Dapp.deploy.version=$VERSION
```
meltsufin, line 33: What's the purpose of `clean` here?
Also, why do we have to specify project and version? I thought those were optional.

spring-cloud-gcp-samples/spring-cloud-gcp-logging-sample/pom.xml
```diff
@@ -69,13 +69,6 @@
             <plugin>
                 <groupId>com.google.cloud.tools</groupId>
                 <artifactId>appengine-maven-plugin</artifactId>
-                <version>1.3.2</version>
-            </plugin>
-            <plugin>
-                <artifactId>maven-deploy-plugin</artifactId>
-                <configuration>
-                    <skip>true</skip>
```
meltsufin, line 77: Do we know for sure that the skip deploy configuration is inherited from the parent?
We really don't want to accidentally deploy samples to maven central.

spring-cloud-gcp-samples/spring-cloud-gcp-logging-sample/src/main/appengine/app.yaml
```diff
@@ -1,9 +1,5 @@
...
+instance_class: F1
 
-handlers:
-- url: /.*
-  script: this field is required, but ignored
-
-manual_scaling:
-  instances: 1
+env_variables:
+  JAVA_TOOL_OPTIONS: "-XX:MaxRAM=256m -XX:ActiveProcessorCount=2 -Xmx32m"
```
meltsufin, line 5: Is this based on experimentation or some guidance somewhere?

spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-stream-binder-functional-sample/pom.xml
```diff
@@ -32,4 +32,16 @@
     </dependencies>
   </dependencyManagement>
 
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.springframework.boot</groupId>
+        <artifactId>spring-boot-maven-plugin</artifactId>
+        <configuration>
+          <skip>true</skip>
```
meltsufin, line 41: What's the effect of this skip configuration on Spring Boot Maven plugin?

spring-cloud-gcp-samples/pom.xml
ttomsu: I went looking for the `deploy` plugin up the chain of POMs - maybe I'm confused and it's just built-in? In that case, yes, I need to add this back in at the `spring-cloud-gcp-samples/pom.xml`, but I think then, it would cascade down and all the samples would _not_ need it, correct?

spring-cloud-gcp-samples/spring-cloud-gcp-logging-sample/README.adoc
ttomsu: `clean` wipes out anything prebuilt in the `target` folder - more just to ensure you're really deploying what you think you're deploying. It's not entirely _necessary_ but doesn't hurt anything either.

The new 2.2.0 plugin requires both the project ID and version to be specified in the config or passed via `-Dapp.deploy.[projectId | version]` - you get errors without it. I figured it was easier to do this in the command line vs an instruction on setting it in the config.

spring-cloud-gcp-samples/spring-cloud-gcp-logging-sample/pom.xml
ttomsu: No, I don't know for sure if this works as intended. Maven n00b here :smile:

spring-cloud-gcp-samples/spring-cloud-gcp-logging-sample/src/main/appengine/app.yaml
ttomsu: Lifted unapologetically from https://github.com/GoogleCloudPlatform/java-docs-samples/blob/master/appengine-java11/springboot-helloworld/src/main/appengine/app.yaml#L21

spring-cloud-gcp-samples/spring-cloud-gcp-pubsub-stream-binder-functional-sample/pom.xml
ttomsu: This sample is a bit funky - the `-sink` and `-source` modules are supposed to be independent, so in theory, they should have been able to be `repackaged` just fine. 

However, the third module (`-sample-test`) only contains tests. The test imports both the -sink and -source modules as _libraries_, which (apparently) don't play well with the Boot `repackaging` goal. 

So, at first I only disabled `spring-boot-maven-plugin` on the `-sample-test`, and it `mvn compile`d, but failed at runtime with a ClassNotFound exception from one of the other modules (or a related exception, I don't remember exactly). To workout the failing test, I just disabled it on the whole thing.

ttomsu:
@meltsufin @dmitry-s - On the topic of standalone samples: we have to decide to be fully standalone or not, because it looks (to me) to still be currently mixed.

Evidence that it's still mixed are things like:
```
<!-- Your own application should inherit from spring-boot-starter-parent -->
```
and 
```
<version>2.2.5.BUILD-SNAPSHOT</version>
```
in `spring-cloud-gcp-samples/pom.xml`

Guiding thoughts on this evidence:
1. Standalone examples should be copy/paste (even with comments you have to know how Maven works to know _what_ to change, and what to change it _to_)
2. Examples should use released versions (which support the copy/paste idea above), and are considered "production ready."

That being said, I see why it was done this way: make a change in a library, and update the example all in one PR. It makes sense from a developer workflow.

Maybe to support this, it would make sense to split the examples out into their own repo, so it would be explicitly be a two-step process to update the library _then_ update the example (once released).

@saturnism for his thoughts on this, too.

spring-cloud-gcp-samples/pom.xml
meltsufin: It is built in, but I don't know about cascading.

spring-cloud-gcp-samples/spring-cloud-gcp-logging-sample/pom.xml
meltsufin: I'm as much a n00b as you are on this.  :)
Maybe just run `mvn deploy` and see what happens?

meltsufin:
@ttomsu Regarding standalone samples, we've struggled with this question in the past and decided to make them semi-stadalone in a sense that it should be easy to detach them from the samples parent and replace it with the spring boot parent.

The reasoning is that these samples are used largely as integration tests for the project, and we need them to build from the root. Also, like you said, keeping them in sync with the latest code changes is easier when they're part of the multi-module project build.

For samples that we truly want to offer as standalone, we already have a home in java-docs-samples. See for example this one: https://github.com/GoogleCloudPlatform/java-docs-samples/tree/master/spanner/spring-data

dzou:
Hey guys, I would say don't get too hung up about the standalone thing. The ultimate goal is just to make it so if someone wants to copy/paste a sample into their own project, they can do this with minimal work.

So I think Travis's PR is fine; it still allows users to copy/paste the sample and change the parent to `spring-boot-starter-parent` and it will work out of the box.

I think there is some confusion (from the reviewers) because you saw that in this PR the `spring-boot-maven-plugin` was moved to the samples parent and my guess is you thought that this would prevent users from running `mvn spring-boot:run` if someone just copy/pasted the sample. But you don't need to add that plugin to run `mvn spring-boot:run`; that is how all our samples worked before. The plugin was added to the parent in this PR so that the JAR gets packaged correctly if the user deploys the app to AppEngine. It doesn't stop the sample from being standalone.

So I think if you want to be technical, you could just add the `spring-boot-maven-plugin` defs only to the samples that are deployable to AppEngine and remove it from the samples parent. (But I think it's fine to leave it as is.) I just wanted to clarify this since I think there is some misunderstanding because this PR still keeps all the samples "stand-alone" at the same level as before.

meltsufin:
I think this PR is fine to go in. My only remaining concern is the adjustments to the `maven-deploy-plugin` and the potential of accidentally publishing samples. We should either confirm that it's not a danger, or leave this to another PR. After all, it doesn't seem to be related to GAE support anyway. (It's a different kind of deploy :))

ttomsu:
Added deploy skip back in, as I want to ensure our samples are not actually deployed.

PR #2399, opened 2020-05-29
Updates to boot 2.3 and back to 2.x build

spencergibb:


elefeint:
The checkstyle version enforcement was in the samples root project, so that checkstyle project can run their snapshot regression tests against `spring-cloud-gcp`. Is it causing issues?

spencergibb:
Yeah, first there's a check for snapshots, now I'm getting a weird compile error in samples. I think it has to do with the mismatched boot version.

spencergibb:
Can't rely on 3.x stuff since it's not released

elefeint:
Build fails with `Failed during checkstyle execution: Unable to find configuration file at location: checkstyle.xml: Could not find resource 'checkstyle.xml'`.

I recall that checkstyle moved somewhere else in the last couple of months in spring cloud build; perhaps we need to provide a customized location? 

Note that the samples do not inherit from `spring-cloud-gcp`, and instead inherit from `spring-boot-starter-parent`. That's why `spring-cloud-build-tools` was needed as a checkstyle dependency in samples pom.

spencergibb:
Ah, then they need to be restored to the 2.3.x branch of spring cloud build

dzou:
@spencergibb - Do you still need these changes? Or are they no longer necessary after the release?

spencergibb:
I think they were taken care of elsewhere

PR #2405, opened 2020-05-30
Allow `projectId` override in Firebase authentication

### LINKED ISSUE #2178: Allow `projectId` override in Firebase authentication
The new Firebase authentication should allow users to override the default `projectId`, similarly to [Pub/Sub](https://github.com/spring-cloud/spring-cloud-gcp/blob/8d8adbe6a1fa32a17058ad9e45f22e3bb15c7018/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubProperties.java#L50), since it's possible to need Firebase in one project and, say, Datastore, in another (since they can't live in the same project).

Follow up to #2157.

femrtnz:
Hi All,

This is my first PR on spring-cloud. Please let me know if you have any questions or suggestions.

Fixes #2178, I'm enabling the project ID to be set on `FirebaseAuthenticationProperties.java` so you can use a different project ID


spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/security/FirebaseAuthenticationProjectIdEvaluatorTests.java
```diff
@@ -0,0 +1,101 @@
...
+				.run(context -> {
+					FirebaseAuthenticationProjectIdEvaluator evaluator = context.getBean(FirebaseAuthenticationProjectIdEvaluator.class);
+					assertThat(evaluator.getProjectId()).isEqualTo(GCP_PROVIDER_PROJECT_ID);
+				});
+	}
+
+	@Test
+	public void testFirebaseProjectIdSet() throws Exception {
+		new ApplicationContextRunner()
+				.withPropertyValues("spring.cloud.gcp.security.firebase=" + FIREBASE_PROJECT_ID)
```
eddumelendez, line 62: missing `.project-id`

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/security/FirebaseAuthenticationProjectIdEvaluator.java
```diff
@@ -0,0 +1,42 @@
...
+ */
+
+package org.springframework.cloud.gcp.autoconfigure.security;
+
+import org.springframework.cloud.gcp.core.GcpProjectIdProvider;
+
+/**
+ *
+ * @author Felipe M Amaral
+ * @since 1.2.2
```
eddumelendez, line 24: `1.2.4`

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/security/FirebaseAuthenticationProjectIdEvaluatorTests.java
```diff
@@ -0,0 +1,101 @@
...
+		}
+
+		@Bean
+		public FirebaseAuthenticationProperties firebaseAuthenticationProperties() {
+			return mock(FirebaseAuthenticationProperties.class);
+		}
+
+	}
+
+	static class TestConfigWithFirebaseProjectId {
```
eddumelendez, line 85: I think, you can use `@EnableConfigurationProperties(FirebaseAuthenticationProperties .class` so you can remove the bean mocking `FirebaseAuthenticationProperties`

femrtnz: Great Idea, thanks. Updated.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/security/FirebaseAuthenticationProjectIdEvaluator.java
ttomsu: Doesn't look resolved?

femrtnz: I'm so sorry, you are right. Just pushed the fix

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/security/FirebaseAuthenticationProjectIdEvaluator.java
```diff
@@ -0,0 +1,41 @@
...
+package org.springframework.cloud.gcp.autoconfigure.security;
+
+import org.springframework.cloud.gcp.core.GcpProjectIdProvider;
+
+/**
+ *
+ * @author Felipe M Amaral
+ * @since 1.2.4
+ */
+public class FirebaseAuthenticationProjectIdEvaluator {
```
meltsufin, line 26: Do we really need this class? In every other module we implement this feature directly in the auto-configuration class with code that looks something like this:
```java
	public GcpPubSubAutoConfiguration(GcpPubSubProperties gcpPubSubProperties,
			GcpProjectIdProvider gcpProjectIdProvider,
			CredentialsProvider credentialsProvider) throws IOException {
		this.gcpPubSubProperties = gcpPubSubProperties;
		this.finalProjectIdProvider = (gcpPubSubProperties.getProjectId() != null)
				: gcpProjectIdProvider;
```

femrtnz: Hi @meltsufin, 

I thought about this option when I started, however, I also noticed there are no unit tests for all similar places, so I applied a testable class.

I agree that this option would be simpler though.

Please let me know your thoughts

meltsufin: Since we have many modules in the project, I generally encourage consistency to make things more readable and maintainable. If something can be improved, then I would do it across the board. In this particular case, I would go with the existing approach, however flawed.
Even though we might be missing some tests, I believe it should still be possible to test the existing approach.

femrtnz: Fair enough. I will change to the simpler approach then. Thanks for the feedback.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/security/FirebaseAuthenticationAutoConfiguration.java
```diff
@@ -58,13 +58,19 @@
 
 	private static final String ISSUER_TEMPLATE = "https://securetoken.google.com/%s";
 
+	private final String projectId;
+
+	public FirebaseAuthenticationAutoConfiguration(GcpProjectIdProvider gcpProjectIdProvider, FirebaseAuthenticationProperties properties) {
+		this.projectId = (properties.getProjectId() != null) ? properties.getProjectId() : gcpProjectIdProvider.getProjectId();
```
eddumelendez, line 64: I think parenthesis are not needed 😬

femrtnz: Indeed! I was just following the same pattern as suggested :)
Removed :)

PR #2408, opened 2020-06-01
Bump current master version to 1.2.4.BUILD-SNAPSHOT

meltsufin:


dzou:
Looks good.

I think you can also update our frontpage readme from 1.2.2.RELEASE to 1.2.3.RELEASE here: https://github.com/spring-cloud/spring-cloud-gcp#spring-cloud-gcp-bill-of-materials-bom

ttomsu:
Is there a way to do this where the version is set in one place and referenced in the other 80+ places (by my recent replace-all count)?

meltsufin:
> Is there a way to do this where the version is set in one place and referenced in the other 80+ places (by my recent replace-all count)?

Not that I know of, but I would be eager to, if it was an option.
Generally, I do it with the built-in `mvn versions:set -DnewVersion=...`. So, it not too tedius.

PR #2426, opened 2020-06-11
Add delete secret operation to SecretManagerTemplate

kioie:
Adding a delete Secrets method to simplify the process of deleting secrets. This is an additional method to beef up features for the spring secret manager service.


dzou:
Thanks for contributing, looks good! I added some minor edits. Will give others a chance to review.

kioie:
> Looks good! Thanks for the contribution @kioie!
> 
> Note that there is also the option of deleting just a specific version of a secret, but I guess that's less common.

Thanks @meltsufin @dzou I can add a couple of more as we go along including version delete

dzou:
Sounds good thanks, I'll merge this in for now then, feel free to send out another PR when you're ready.

PR #2427, opened 2020-06-12
Add delete secret operation to SecretManagerTemplate

kioie:
Follow up to  #2426

Allows delete operation to specific secret version.

### DIFF
--- a/spring-cloud-gcp-secretmanager/src/main/java/org/springframework/cloud/gcp/secretmanager/SecretManagerOperations.java
+++ b/spring-cloud-gcp-secretmanager/src/main/java/org/springframework/cloud/gcp/secretmanager/SecretManagerOperations.java
@@ -101,14 +101,6 @@ public interface SecretManagerOperations {
 	void deleteSecret(String secretId, String projectId);
 
 	/**
-	 * Deletes the specified secret version under the default-configured project.
-	 *
-	 * @param secretId the secret ID of the secret to delete.
-	 * @param version the version to delete
-	 */
-	void deleteSecretVersion(String secretId, String version);
-
-	/**
 	 * Deletes the specified secret.
 	 *
 	 * @param secretId the secret ID of the secret to delete.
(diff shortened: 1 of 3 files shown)


dzou:
@kioie - Thanks for your contribution again! Looks good.

I ended up deciding on removing the `void deleteSecretVersion(String secretId, String version);` overload - one thing we were discussing was that we want to limit the size of `SecretManagerOperations` if possible. Also I was worried users might get mixed up between `deleteSecret(String secretId, String projectId)` and `deleteSecretVersion(String secretId, String versionId)` since both accept strings so I hope this is more clear.

@meltsufin - Yeah let's keep it this way for now just because as we discussed in the meeting there could be confusion resulting from using `sm://...` such as users specifying a secret and it being ambigious whether they want to simply delete the latest version or the entire secret.

kioie:
> @kioie - Thanks for your contribution again! Looks good.
> 
> I ended up deciding on removing the `void deleteSecretVersion(String secretId, String version);` overload - one thing we were discussing was that we want to limit the size of `SecretManagerOperations` if possible. Also I was worried users might get mixed up between `deleteSecret(String secretId, String projectId)` and `deleteSecretVersion(String secretId, String versionId)` since both accept strings so I hope this is more clear.
> 
> @meltsufin - Yeah let's keep it this way for now just because as we discussed in the meeting there could be confusion resulting from using `sm://...` such as users specifying a secret and it being ambigious whether they want to simply delete the latest version or the entire secret.

Thanks guys. @dzou makes sense to keep things clear. Do you think it will affect consistency, particularly for users who'd like to autoload projectID?

Also, just to chime in on this discourse regarding use of `sm://...`, I can see how a user trying to delete a secret version would end up deleting the whole secret, by simply missing to append the version.

dzou:
> Thanks guys. @dzou makes sense to keep things clear. Do you think it will affect consistency, particularly for users who'd like to autoload projectID?

Yeah that is a good point I was considering too. I figure we can always add things in the future but it's difficult to remove something if it's unclear. If we see a large need for it we will definitely add it!

PR #2430, opened 2020-06-16
Add enable secret version operation to SecretManagerTemplate

kioie:
Follow up to #2427

Allows enable operation to specific secret version

### DIFF
--- a/spring-cloud-gcp-secretmanager/src/main/java/org/springframework/cloud/gcp/secretmanager/SecretManagerTemplate.java
+++ b/spring-cloud-gcp-secretmanager/src/main/java/org/springframework/cloud/gcp/secretmanager/SecretManagerTemplate.java
@@ -101,6 +101,21 @@ public class SecretManagerTemplate implements SecretManagerOperations {
 		return true;
 	}
 
+	//@Override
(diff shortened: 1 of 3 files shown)


spring-cloud-gcp-secretmanager/src/main/java/org/springframework/cloud/gcp/secretmanager/SecretManagerTemplate.java
```diff
@@ -101,6 +101,21 @@ public boolean secretExists(String secretId, String projectId) {
 		return true;
 	}
 
+	//@Override
```
meltsufin, line 104: Uncomment?

kioie:
> For completeness, we would need `disableSecretVersion()` as well.

Absolutely, coming through on a separate PR

spring-cloud-gcp-secretmanager/src/main/java/org/springframework/cloud/gcp/secretmanager/SecretManagerTemplate.java
kioie: Done!

PR #2432, opened 2020-06-16
Add disable secret version operation to SecretManagerTemplate

kioie:
Follow up to #2430

Allows disable operation to specific secret version


dzou:
@kioie - Looks good, thank you!

PR #2434, opened 2020-06-16
Add subcollection support for FirestoreTemplate

### LINKED ISSUE #2415: Implement subcollection support in Firestore
User request: https://stackoverflow.com/questions/62043573/how-to-fetch-documents-from-subcollection-in-firestore-using-spring-data/

Currently we support embedded collections, but Firestore has one more type of associated collections, called [subcollections](https://firebase.google.com/docs/firestore/data-model#subcollections).  That feature allows loading the collection independently from the parent document.

dmitry-s:
Partially addresses #2415


spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreReactiveOperations.java
```diff
@@ -129,6 +129,22 @@
 	 */
 	<T> Mono<Void> deleteById(Publisher<String> idPublisher, Class entityClass);
 
+	/**
+	 * Executes a query represented as query builder and returns results of the given domain type.
+	 * @param <T> the type param of the domain type.
+	 * @param builder the query builder.
+	 * @param entityClass the domain type of entities.
+	 * @return {@link Mono} signaling when operation has completed.
```
meltsufin, line 137: The actual return type is `Flux` though?

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreReactiveOperations.java
```diff
@@ -129,6 +129,22 @@
...
+	 */
+	<T> Flux execute(StructuredQuery.Builder builder, Class<T> entityClass);
+
+	/**
+	 * Creates FirestoreReactiveOperations object with a provided parent.
+	 * The parent doesn't have to exist in Firestore, but should have a non-empty id field.
+	 * @param <T> the type param of the parent.
+	 * @param parent the query builder.
+	 * @return {@link Mono} signaling when operation has completed.
```
meltsufin, line 146: No `Mono` returned.

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/Subcollection.java
```diff
@@ -0,0 +1,38 @@
...
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Annotation for Datastore property that indicates it holds child entities that are
+ * linked to this entity by the Cloud Datastore Ancestor relationship.
+ *
+ * @author Chengyuan Zhao
```
meltsufin, line 29: And version don't seem accurate.

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreReactiveOperations.java
```diff
@@ -129,6 +129,22 @@
 	 */
 	<T> Mono<Void> deleteById(Publisher<String> idPublisher, Class entityClass);
 
+	/**
+	 * Executes a query represented as query builder and returns results of the given domain type.
+	 * @param <T> the type param of the domain type.
+	 * @param builder the query builder.
+	 * @param entityClass the domain type of entities.
+	 * @return {@link Mono} signaling when operation has completed.
+	 */
```
meltsufin, line 138: Add `@since 1.2.4`

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreReactiveOperations.java
```diff
@@ -129,6 +129,22 @@
...
+	<T> Flux execute(StructuredQuery.Builder builder, Class<T> entityClass);
+
+	/**
+	 * Creates FirestoreReactiveOperations object with a provided parent.
+	 * The parent doesn't have to exist in Firestore, but should have a non-empty id field.
+	 * @param <T> the type param of the parent.
+	 * @param parent the query builder.
+	 * @return {@link Mono} signaling when operation has completed.
+	 */
```
meltsufin, line 147: Add `@since 1.2.4`

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/Subcollection.java
```diff
@@ -0,0 +1,38 @@
...
+ *
+ * @since 1.1
+ */
+@Documented
+@Target(ElementType.FIELD)
+@Retention(RetentionPolicy.RUNTIME)
+public @interface Subcollection {
```
meltsufin, line 36: Where is this used? I don't see it.

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreReactiveOperations.java
```diff
@@ -129,6 +129,22 @@
...
 	<T> Mono<Void> deleteById(Publisher<String> idPublisher, Class entityClass);
 
+	/**
+	 * Executes a query represented as query builder and returns results of the given domain type.
+	 * @param <T> the type param of the domain type.
+	 * @param builder the query builder.
+	 * @param entityClass the domain type of entities.
+	 * @return {@link Mono} signaling when operation has completed.
+	 */
+	<T> Flux execute(StructuredQuery.Builder builder, Class<T> entityClass);
```
meltsufin, line 139: How is this related to subcollections?

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreTemplate.java
```diff
@@ -101,6 +101,10 @@ public FirestoreTemplate(FirestoreStub firestore, String parent, FirestoreClassM
 		this.mappingContext = mappingContext;
 	}
 
+	public <T> FirestoreReactiveOperations operationsWithParent(T parent) {
+		return new FirestoreTemplate(this.firestore, buildResourceName(parent), this.classMapper, this.mappingContext);
```
dzou, line 105: I think you may need to invent a new abstraction here which implements `FirestoreReactiveOperations` and just has a reference to the subcollection and the `FirestoreTemplate` and then delegates calls to the firestoretemplate.

I think copying the whole `FirestoreTemplate` object is too heavyweight for just setting the root; feels like there is too much redundant information being copied each time the method is called. Another reason why it doesn't feel quite right is because you are also going to need to call all the `setter` methods to copy over the settings that were set outside the constructor to do this right.

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreTemplate.java
```diff
@@ -101,6 +101,10 @@ public FirestoreTemplate(FirestoreStub firestore, String parent, FirestoreClassM
 		this.mappingContext = mappingContext;
 	}
 
+	public <T> FirestoreReactiveOperations operationsWithParent(T parent) {
```
dzou, line 104: I think you should add to docs how you intend users to use this method. I am imagining something like this:

```
@Autowired
FirestoreTemplate template;

template.withRoot(user1).delete(...)
template.withRoot(user2).find(...)
```

Also I think we should consider renaming it from `operationsWithParent` to something like: `withCollectionRoot(...)`, `withRoot(...)`, or `withParent(..)`, etc. The method name should be short and also be clear that we are doing our operations from a different parent other than `/`.

dzou, line 104: This also needs a `@Override`.

docs/src/main/asciidoc/firestore.adoc
```diff
@@ -255,6 +255,29 @@ Now when you call the methods annotated with `@Transactional` on your service ob
 If an error occurs during the execution of a method annotated with `@Transactional`, the transaction will be rolled back. 
 If no error occurs, the transaction will be committed.
 
+==== Subcollections
```
dzou, line 258: Can you rename your pull request to "Add subcollection support for FirestoreTemplate" to be more specific?

docs/src/main/asciidoc/firestore.adoc
```diff
@@ -255,6 +255,29 @@ Now when you call the methods annotated with `@Transactional` on your service ob
...
+A subcollection is a collection associated with a specific entity.
+Documents in subcollections can contain subcollections as well, allowing you to further nest data. You can nest data up to 100 levels deep.
+
+WARNING: Deleting a document does not delete its subcollections!
```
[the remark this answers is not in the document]
dmitry-s: see https://firebase.google.com/docs/firestore/data-model

> Warning: Deleting a document does not delete its subcollections!

> When you delete a document that has subcollections, those subcollections are not deleted. For example, there may be a document located at coll/doc/subcoll/subdoc even though the document coll/doc no longer exists. If you want to delete documents in subcollections when deleting a parent document, you must do so manually, as shown in Delete Collections.

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreTemplate.java
dmitry-s: It only "copies" references. What do you `setter` do you mean?

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreTemplate.java
dmitry-s: the docs are there in the docs submodule

dmitry-s:
@meltsufin Subcollections do not map to entities naturally. The reason is they are designed to stay and be accessible (and completely independent) even if the parent entity was removed. 

The Template approach allows you to use it these way.

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreTemplate.java
dzou: `FirestoreTemplate` has settings like buffer or useTokens which are not set through constructor but setters, these would have to be copied as well

dzou: I guess for now you can just call the setters so that you set the properties of the child template in the same way as the parent; and if the future if we need to optimize we can do it later.

meltsufin:
> @meltsufin Subcollections do not map to entities naturally. The reason is they are designed to stay and be accessible (and completely independent) even if the parent entity was removed.
> 
> The Template approach allows you to use it these way.

This being said, should we clarify with the original poster what functionality they want, and wether something like this template support would be what they want?

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreReactiveOperations.java
```diff
@@ -129,6 +129,24 @@
...
+   * Executes a query represented as query builder and returns results of the given domain type.
+   * @param <T> the type param of the domain type.
+   * @param builder the query builder.
+   * @param entityClass the domain type of entities.
+   * @return {@link Flux} emitting the found entities.
+   * @since 1.2.4
+   */
+  <T> Flux<T> execute(StructuredQuery.Builder builder, Class<T> entityClass);
+
+  /**
```
dzou, line 142: indents need to be fixed to tabs instead of spaces.

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreReactiveOperations.java
```diff
@@ -129,6 +129,24 @@
...
+   * Creates FirestoreReactiveOperations object with a provided parent.
+   * The parent doesn't have to exist in Firestore, but should have a non-empty id field.
```
dzou, line 144: This javadoc needs a little bit more description; I think you have to define "parent" more clearly.

There needs to be something that says that by default the template operates at the root level collection `/` and then once you pass an entity in here, a new template is created with that entity at the root collection and that all operations/queries will be scoped to that document's subcollections.

Also need to clarify what happens if you pass in an entity in here that is not a child of the current parent with `withParent(..)` - does it still work?

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreTemplate.java
```diff
@@ -101,6 +101,17 @@ public FirestoreTemplate(FirestoreStub firestore, String parent, FirestoreClassM
 		this.mappingContext = mappingContext;
 	}
 
+  @Override
```
meltsufin, line 104: tabs vs spaces

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreReactiveOperations.java
dmitry-s: The "scope" actually explains it. When you save any entity, it is saved within the scope of this parent's subcollections.

PR #2437, opened 2020-06-17
Replaced DatastoreTemplate with DatastoreOperations

### LINKED ISSUE #2436: Use DatastoreOperations instead of DatastoreTemplate
I was trying to proxy the DataStoreOperations by using kotlin delegate "by" method. But I found out that in several places on the DataStoreRepositoryFactory and associated classes they use a DatastoreTemplate and not the interface DatastoreOperations.

I started a refactor to push a PR, but noticed there are several methods that only exist on `DatastoreTemplate` it wasn't clear why those methods are not also present on the interface as many of them use the same types as the ones present on DataStoreOperations.

Is there a reason why all the classes are using an implementation instead of the interface contract?

Any chances this could be changed to use DataStoreOperations?

viniciusccarvalho:
Fixes #2436 

Replaced ocurrences of DatastoreTemplate to use interface DatastoreOperations. Had to pull up 3 methods into interface, but all of them are using the same types as any other method in the interface.

Note: Master build is failing, therefore I could not validate all tests, for what I can tell the data-datastore tests passes along with checkstyle and so does the autoconfigure tests and integration with emulator. But master is failing as of now.

### DIFF
--- a/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreOperations.java
+++ b/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreOperations.java
@@ -24,12 +24,14 @@ import com.google.cloud.datastore.BaseEntity;
 import com.google.cloud.datastore.Key;
 import com.google.cloud.datastore.Query;
 
+import org.springframework.cloud.gcp.data.datastore.core.convert.DatastoreEntityConverter;
 import org.springframework.data.domain.Example;
 
 /**
  * An interface of operations that can be done with Cloud Datastore.
  *
  * @author Chengyuan Zhao
+ * @author Vinicius Carvalho
  *
  * @since 1.1
  */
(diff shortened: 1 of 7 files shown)


spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreOperations.java
```diff
@@ -260,4 +262,23 @@
...
+	 * returned.
+	 * @param query the query to execute.
+	 * @param entityClass the type of object to retrieve.
+	 * @param <T> the type of object to retrieve.
+	 * @return a list of the objects found. If no keys could be found the list will be
+	 * empty.
+	 */
+	<T> DatastoreResultsIterable<?> queryKeysOrEntities(Query query, Class<T> entityClass);
+
+	<A, T> DatastoreResultsIterable<T> queryIterable(Query<A> query, Function<A, T> entityFunc);
```
meltsufin, line 283: We should javadoc here, since it's the interface.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/AbstractDatastoreQuery.java
```diff
@@ -40,12 +40,12 @@
...
 	final DatastoreQueryMethod queryMethod;
 
-	final DatastoreTemplate datastoreTemplate;
+	final DatastoreOperations datastoreTemplate;
 
 	final Class<T> entityType;
 
 	public AbstractDatastoreQuery(DatastoreQueryMethod queryMethod,
-			DatastoreTemplate datastoreTemplate,
+							DatastoreOperations datastoreTemplate,
```
meltsufin, line 48: Unintentional spacing change?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/AbstractDatastoreQuery.java
```diff
@@ -79,7 +79,7 @@ Object processRawObjectForProjection(Object object) {
 		return this.queryMethod.getResultProcessor().processResult(object);
 	}
 
-	public DatastoreTemplate getDatastoreTemplate() {
+	public DatastoreOperations getDatastoreTemplate() {
```
meltsufin, line 82: Shouldn't it be `getDatastoreOperations` then?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/support/DatastoreRepositoryFactory.java
```diff
@@ -66,7 +66,7 @@
 	 * repositories.
 	 */
 	DatastoreRepositoryFactory(DatastoreMappingContext datastoreMappingContext,
-			DatastoreTemplate datastoreTemplate) {
+							DatastoreOperations datastoreTemplate) {
```
meltsufin, line 69: Unintentional spacing change?

PR #2438, opened 2020-06-17
Adjust SecretManagerOperations Parameter Ordering

dzou:
For the new methods introduced `enableSecretVersion`, `disableSecretVersion`, and `deleteSecretVersion` I adjusted the parameter order to move the `projectId` to the end of the parameter list. This makes things consistent with our other methods which introduced `projectId`.

cc/ @kioie

### DIFF
--- a/spring-cloud-gcp-secretmanager/src/main/java/org/springframework/cloud/gcp/secretmanager/SecretManagerOperations.java
+++ b/spring-cloud-gcp-secretmanager/src/main/java/org/springframework/cloud/gcp/secretmanager/SecretManagerOperations.java
@@ -97,10 +97,10 @@ public interface SecretManagerOperations {
 	 * Enables the secret version under the specified project.
 	 *
 	 * @param secretId the secret ID of the secret to delete.
-	 * @param projectId unique identifier of your project.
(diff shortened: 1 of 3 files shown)


spring-cloud-gcp-secretmanager/src/main/java/org/springframework/cloud/gcp/secretmanager/SecretManagerOperations.java
```diff
@@ -97,10 +97,10 @@
 	 * Enables the secret version under the specified project.
 	 *
 	 * @param secretId the secret ID of the secret to delete.
```
kioie, line 99: Typo here. Should be "secret ID of the secret to enable"

kioie:
@dzou order looks great. Thanks for incorporating the changes.

spring-cloud-gcp-secretmanager/src/main/java/org/springframework/cloud/gcp/secretmanager/SecretManagerOperations.java
dzou: Nice catch, done.

PR #2440, opened 2020-06-18
Add the "spring.cloud.gcp.storage.project-id" autoconfig

### LINKED ISSUE #2435: Let Spring Cloud GCP Storage select a different project id than the core module
Hello guys.

Many modules of Spring Cloud GCP let the user configure a different project-id. Example: 
image

Spring Cloud GCP Storage doesn't have this option.  What about to let users always configure the project id for each module?

kioie:
Fixes #2435 

@meltsufin @dzou hey guys, this is an attempt to add the feature as requested on the issue. I'm really not sure if I'm on the right path, but, please review and help or maybe point the right direction.

### DIFF
--- a/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/storage/GcpStorageAutoConfiguration.java
+++ b/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/storage/GcpStorageAutoConfiguration.java
@@ -28,6 +28,7 @@ import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.cloud.gcp.autoconfigure.core.GcpProperties;
 import org.springframework.cloud.gcp.core.DefaultCredentialsProvider;
+import org.springframework.cloud.gcp.core.DefaultGcpProjectIdProvider;
 import org.springframework.cloud.gcp.core.GcpProjectIdProvider;
 import org.springframework.cloud.gcp.core.UserAgentHeaderProvider;
 import org.springframework.cloud.gcp.storage.GoogleStorageProtocolResolver;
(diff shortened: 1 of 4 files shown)


spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/storage/GcpStorageAutoConfiguration.java
```diff
@@ -53,20 +54,22 @@
...
 @Import(GoogleStorageProtocolResolver.class)
-public abstract class GcpStorageAutoConfiguration { //NOSONAR squid:S1610 must be a class for Spring
+public abstract class GcpStorageAutoConfiguration {
+	private GcpProjectIdProvider gcpProjectIdProvider; //NOSONAR squid:S1610 must be a class for Spring
 
 	@Bean
 	@ConditionalOnMissingBean
-	public static Storage storage(CredentialsProvider credentialsProvider,
-			GcpStorageProperties gcpStorageProperties,
-			GcpProjectIdProvider projectIdProvider) throws IOException {
```
meltsufin, line 62: Why are you removing the `projectIdProvider` parameter?

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/storage/GcpStorageAutoConfiguration.java
```diff
@@ -53,20 +54,22 @@
...
+	public Storage storage(CredentialsProvider credentialsProvider, GcpStorageProperties gcpStorageProperties) throws IOException {
+		this.gcpProjectIdProvider = gcpStorageProperties.getProjectId() != null
+				? gcpStorageProperties::getProjectId
+				: new DefaultGcpProjectIdProvider();
```
meltsufin, line 65: Should be `projectIdProvider` that was passed in.

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/storage/GcpStorageAutoConfigurationTests.java
```diff
@@ -46,9 +45,11 @@
  * @author Elena Felder
  */
 public class GcpStorageAutoConfigurationTests {
+	private static final String PROJECT_NAME = "hollow-light-of-the-sealed-land";
 
 	private ApplicationContextRunner contextRunner = new ApplicationContextRunner()
 		.withConfiguration(AutoConfigurations.of(GcpStorageAutoConfiguration.class))
+			.withPropertyValues("spring.cloud.gcp.storage.project-id=" + PROJECT_NAME)
```
meltsufin, line 52: I would just add another test with and without the property.

dzou:
Thank you @kioie ! I committed some polish changes for you.

@meltsufin - I was thinking about how one might write a test for this, it doesn't seem straightforward since the storage projectID is not exposed through the storage autoconfiguration. maybe it would be sufficient to just manually verify that the project setting works on a sample.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/storage/GcpStorageAutoConfiguration.java
kioie: Forgot to replace it, but I can see @dzou has setup a replacement already.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/storage/GcpStorageAutoConfiguration.java
kioie: Yea, thanks for the catch!

kioie:
Thanks @dzou for your help with the polish! @meltsufin thanks for your input and review.

dzou:
Just did some local testing and both the project id and credentials get overridden properly. Will merge.

PR #2441, opened 2020-06-19
Add maxMessages to PubSubReactiveFactory.poll

### LINKED ISSUE #2072: PubSubReactiveFactory should allow overriding the pull batch size
Add an option controlling batch size, so that `poll()` does not [always request a max int](https://github.com/spring-cloud/spring-cloud-gcp/blob/8df2a6bb91af4a8d6517d62b2b23387296670157/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/reactive/PubSubReactiveFactory.java#L174) number of messages from Pub/Sub for unbounded demand.

It was a follow up task from original implementation.

mturbe:
Fixes #2072

### DIFF
--- a/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/reactive/PubSubReactiveFactory.java
+++ b/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/reactive/PubSubReactiveFactory.java
@@ -54,16 +54,35 @@ public final class PubSubReactiveFactory {
 
 	private final Scheduler scheduler;
 
+	private final int maxMessages;
+
 	/**
 	 * Instantiate `PubSubReactiveFactory` capable of generating subscription-based streams.
+	 * <p>{@code maxMessages} is set to {@code Integer.MAX_VALUE}.</p>
 	 * @param subscriberOperations template for interacting with GCP Pub/Sub subscriber operations.
 	 * @param scheduler scheduler to use for asynchronously retrieving Pub/Sub messages.
 	 */
 	public PubSubReactiveFactory(PubSubSubscriberOperations subscriberOperations, Scheduler scheduler) {
+		this(subscriberOperations, scheduler, Integer.MAX_VALUE);
+	}
+
+	/**
+	 * Instantiate `PubSubReactiveFactory` capable of generating subscription-based streams.
+	 * @param subscriberOperations template for interacting with GCP Pub/Sub subscriber operations.
+	 * @param scheduler scheduler to use for asynchronously retrieving Pub/Sub messages.
+	 * @param maxMessages max number of messages that may be pulled from the source
+	 * subscription in case of unlimited demand.
+	 */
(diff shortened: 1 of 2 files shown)


spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/reactive/PubSubReactiveFactory.java
```diff
@@ -72,23 +72,29 @@ public PubSubReactiveFactory(PubSubSubscriberOperations subscriberOperations, Sc
...
+	 * @param pollingPeriodMs how frequently to poll the source subscription in case of
+	 * unlimited demand, in milliseconds.
+	 * @param maxMessages max number of messages that may be pulled from the source
+	 * subscription in case of unlimited demand.
 	 * @return infinite stream of {@link AcknowledgeablePubsubMessage} objects.
+	 * @throws IllegalArgumentException if {@code maxMessages < 0}
 	 */
-	public Flux<AcknowledgeablePubsubMessage> poll(String subscriptionName, long pollingPeriodMs) {
-
+	public Flux<AcknowledgeablePubsubMessage> poll(String subscriptionName, long pollingPeriodMs, int maxMessages) {
```
elefeint, line 90: Methods with multiple numeric parameters can get confusing (we have `String`, `long`, `int` here). I think I'd prefer to set `maxMessages` in an overloaded `PubSubReactiveFactory` constructor. 

Upsides:
* we get to keep a single `poll()` method. 
* `maxMessages` can be autoconfigured to current max int value by default and overridden with a property, keeping it backwards compatible and consistent with how we configure other functionality.
* `maxMessages` won't need to be propagated from `poll()` to `pollingPull()` to `pullAll()`.

Downside:
* different instances of the factory have to be created if different `maxMessages` settings are required in a single application. This is mitigated by how lightweight `PubSubReactiveFactory` is; it's just a wrapper around `PubSubTemplate` and a reactor scheduler.

@meltsufin What do you think?

dzou: Makes sense, I see the `maxMessages` setting being something not changed too frequently so setting in constructor seems appropriate.

meltsufin: Overloaded constructor or overloaded `poll` methods would work. Since it affects multiple methods, constructor is probably preferred.
Either way, the way it is now, is not good because it would causes a breaking change.

mturbe:
@elefeint, @dzou, @meltsufin: thank you for the feedback. I have reverted my previous commit and added a new overloaded constructor for PubSubReactiveFactory that has maxMessages as a parameter.

elefeint:
As follow-up, I've filed #2451 to autoconfigure the new property.

PR #2445, opened 2020-06-19
Allow null bindings in Spanner Repository Queries

### LINKED ISSUE #2443: NullPointerException when passing the null value for nullable column
## Version
both spring-cloud-gcp-data-spanner-1.2.0.RELEASE and 1.2.1.RELEASE. Only I checked.

## What I did
I used `@Query` annotation and DML statement, then pass the null value for the nullable column.

## What was occurred
NullPointerException at SqlSpannerQuery.

## What I was expected
I want to input null value in the table if I passed null value to the method and the column is nullable.

## Situation the phenomena occurs
Assume that `lastName` is the nullable column at the singer table.

Define a method like below at the repository with extends of SpannerRepository
```java
@Query(dmlStatement = true, value = "INSERT INTO singer (singer_id, birth_date, first_name, last_name) VALUES (@singerId, @birthDay, @firstName, @lastName)")
public void insert(Integer singerId, Date birthDay, String firstName, String lastName);
```
Then use it like below
```java
xxxRepository.insert(3, Date.valueOf(LocalDate.of(1994, 12, 27)), "first_name", null);
```

dzou:
Allows for the value of `null` to be bound as a variable in a Spanner query.

Fixes #2443.

### DIFF
--- a/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SqlSpannerQuery.java
+++ b/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SqlSpannerQuery.java
@@ -199,13 +199,12 @@ public class SqlSpannerQuery<T> extends AbstractSpannerQuery<T> {
 		Sort sort = null;
 
 		for (Object param : parameters) {
-			Class<?> paramClass = param.getClass();
-			if (isPageableOrSort(paramClass)) {
+			if (param != null && isPageableOrSort(param.getClass())) {
 				if (pageable != null || sort != null) {
 					throw new SpannerDataException(
 							"Only a single Pageable or Sort param is allowed.");
 				}
-				else if (Pageable.class.isAssignableFrom(paramClass)) {
+				else if (Pageable.class.isAssignableFrom(param.getClass())) {
 						pageable = (Pageable) param;
 				}
 				else {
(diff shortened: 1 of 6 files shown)


spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SqlSpannerQuery.java
```diff
@@ -199,13 +199,12 @@ public List executeRawResult(Object[] parameters) {
 		Sort sort = null;
 
 		for (Object param : parameters) {
-			Class<?> paramClass = param.getClass();
-			if (isPageableOrSort(paramClass)) {
+			if (param != null && isPageableOrSort(param.getClass())) {
```
dmitry-s, line 202: what if `param` is `null`, but also a `Pageable` or `Sort`?

meltsufin: NPE would be thrown I guess?
Is this PR changing that behavior?

dzou:
Discussed offline and @dmitry-s said he has an alternate solution; we can wait for his PR and merge the solutions.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SqlSpannerQuery.java
dzou: Yeah I think that is outside the scope of this PR; here I am just addressing passing in `null` as a param in a parameterized query.

dzou:
Closing in favor of #2448.

PR #2448, opened 2020-06-22
Handle null params in Spring Data Spanner

### LINKED ISSUE #2443: NullPointerException when passing the null value for nullable column

dmitry-s:
fixes #2443

### DIFF
--- a/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryExecutor.java
+++ b/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryExecutor.java
@@ -361,6 +361,8 @@ public final class SpannerStatementQueryExecutor {
 	private static void bindParameter(ValueBinder<Statement.Builder> bind,
 			Function<Object, Struct> paramStructConvertFunc, SpannerCustomConverter spannerCustomConverter,
 			Object originalParam, Parameter paramMetadata) {
+
+		// Gets the type of the bind parameter; if null then infer the type from the parameter metadata.
 		Class propType = originalParam != null ? originalParam.getClass() : paramMetadata.getType();
 		if (ConversionUtils.isIterableNonByteArrayType(propType)) {
 			if (!ConverterAwareMappingSpannerEntityWriter.attemptSetIterableValueOnBinder((Iterable) originalParam,
--- a/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryExecutor.java
+++ b/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryExecutor.java
@@ -361,7 +361,8 @@ public final class SpannerStatementQueryExecutor {
 	private static void bindParameter(ValueBinder<Statement.Builder> bind,
 			Function<Object, Struct> paramStructConvertFunc, SpannerCustomConverter spannerCustomConverter,
 			Object originalParam, Parameter paramMetadata) {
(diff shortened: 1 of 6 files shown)


spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/repository/it/SpannerRepositoryInsertIntegrationTests.java
```diff
@@ -0,0 +1,79 @@
...
+		}
+		this.singerRepository.deleteAll();
+	}
+
+	@After
+	public void clearData() {
+		this.singerRepository.deleteAll();
+	}
+
+		@Test
```
dzou, line 73:
```suggestion
	@Test
```

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryExecutor.java
```diff
@@ -361,7 +361,8 @@ public static Statement buildStatementFromSqlWithArgs(String sql, List<String> t
 	private static void bindParameter(ValueBinder<Statement.Builder> bind,
 			Function<Object, Struct> paramStructConvertFunc, SpannerCustomConverter spannerCustomConverter,
 			Object originalParam, Parameter paramMetadata) {
-		if (ConversionUtils.isIterableNonByteArrayType(originalParam.getClass())) {
+		Class propType = originalParam != null ? originalParam.getClass() : paramMetadata.getType();
```
dzou, line 364: Can you add a comment here about how you need to know the type of the parameter (even if you bind `null`) which is why the type information is relevant for null?

dzou:
Gonna close my PR in favor of this one.

spring-cloud-gcp-data-spanner/src/test/java/org/springframework/cloud/gcp/data/spanner/repository/query/SqlSpannerQueryTests.java
```diff
@@ -426,72 +365,18 @@ public void compoundNameConventionTest() throws NoSuchMethodException {
...
 		when(this.queryMethod.getMethod()).thenReturn(method);
-
+		Mockito.<Parameters>when(this.queryMethod.getParameters()).thenReturn(new DefaultParameters(method));
 		sqlSpannerQuery.execute(params);
 
 		verify(this.spannerTemplate, times(1)).executeQuery(any(), any());
 	}
 
 	@Test
-	public void multiplePageableSortTest() {
```
meltsufin, line 440: What's replacing these 3 tests that you deleted?

dzou:
LGTM - only thing left is I think the build is broken.

s13o:
Finally!
What is a planned release time?

dmitry-s:
@s13o 
The next Spring Cloud release is scheduled for next week, but it can be delayed.

Issue #2403, opened 2020-05-29
Add a sample app for the micrometer/Stackdriver monitoring integration


elefeint:
Include an integration test for the sample app.

Follow up to #2391.

eddumelendez:
I think this issue can be closed.

elefeint:
Yes! Fixed in #2455.

PR #2455, opened 2020-07-08
Add metrics sample

eddumelendez:
See gh-2403


eddumelendez:
@elefeint thanks for the feedback! really good points. I have made the changes so test can reproduce every time without trusting in previous executions.

eddumelendez:
@elefeint PR updated. I got it wrong the previous time. Now, the test will hit just once the endpoint and will look for the metrics every 5 seconds in a lifetime of 4 minutes. Thanks for the feedback!

Issue #957, opened 2018-08-16
Remove protobuf-java exclusion from MySQL driver


meltsufin:
Pending: https://bugs.mysql.com/bug.php?id=91999
Exclusion introduced in #955.

PR #2457, opened 2020-07-09
Remove exclusion of protobuf-java from MySQL driver

mkatircioglu:
Closes #gh-957

### DIFF
--- a/spring-cloud-gcp-autoconfigure/pom.xml
+++ b/spring-cloud-gcp-autoconfigure/pom.xml
@@ -89,14 +89,6 @@
             <groupId>mysql</groupId>
             <artifactId>mysql-connector-java</artifactId>
             <optional>true</optional>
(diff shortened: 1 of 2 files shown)


mkatircioglu:
Hi @meltsufin. I'm not totally sure how I can verify the issue therefore I have created PR as draft. After seeing there is no error on CI pipeline, I have also tested with spring-cloud-gcp-sql-mysql-sample whether it still works or not. Is it enough for you or would you like me to do something else? I would be very happy if you can guide me.

meltsufin:
@mkatircioglu This testing is sufficient. However, checking `mysql-connector-java` pom.xml, I see that it still includes `protobuf-java` as a non-optional dependency. The version of the dependency is `3.6.1` and is probably no longer incompatible to cause problems, but it can cause problems in the future, potentially.
@saturnism What do you think? I don't have access to the original bug on mysql.com. Can you see what the outcome was there? https://bugs.mysql.com/bug.php?id=91999

PR #2466, opened 2020-07-15
Firestore - automatic id generation

### LINKED ISSUE #2428: Firestore - DocumentId autogeneration
**Describe the bug**
If a `@DocumentId`field is null, the save operation fails with:
```
Caused by: java.lang.NullPointerException: null
	at org.springframework.cloud.gcp.data.firestore.FirestoreTemplate.buildResourceName(FirestoreTemplate.java:405) ~[spring-cloud-gcp-data-firestore-1.2.3.RELEASE.jar:1.2.3.RELEASE]
	at org.springframework.cloud.gcp.data.firestore.FirestoreTemplate.createUpdateWrite(FirestoreTemplate.java:392) ~[spring-cloud-gcp-data-firestore-1.2.3.RELEASE.jar:1.2.3.RELEASE]
```

In regular Firestore API, document ID is automatically generated.

**Sample**
```
@Document
class Order {

  @DocumentId
  private String id;
 ...
}
```

```
@Repository
interface OrderRepository extends FirestoreReactiveRepository<Order> {}
```

```
Order order = new Order();
repository.save(order);
```

dmitry-s:
fixes #2428

### DIFF
--- a/spring-cloud-gcp-data-firestore/pom.xml
+++ b/spring-cloud-gcp-data-firestore/pom.xml
@@ -23,6 +23,7 @@
         <dependency>
             <groupId>com.google.cloud</groupId>
             <artifactId>google-cloud-firestore</artifactId>
+			<version>1.35.2-SNAPSHOT</version>
         </dependency>
 
         <dependency>
@@ -54,11 +55,6 @@
         <dependency>
             <groupId>org.springframework.cloud</groupId>
             <artifactId>spring-cloud-gcp-core</artifactId>
-        </dependency>
-
-        <dependency>
-            <groupId>io.projectreactor</groupId>
-            <artifactId>reactor-core</artifactId>
         </dependency>
 
         <dependency>
--- a/spring-cloud-gcp-data-firestore/pom.xml
+++ b/spring-cloud-gcp-data-firestore/pom.xml
@@ -23,7 +23,6 @@
         <dependency>
             <groupId>com.google.cloud</groupId>
             <artifactId>google-cloud-firestore</artifactId>
-			<version>1.35.2-SNAPSHOT</version>
         </dependency>
 
         <dependency>
--- a/spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/AutoId.java
+++ b/spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/AutoId.java
@@ -24,7 +24,7 @@ import java.util.Random;
  * Copied from com.google.cloud.firestore.FirestoreImpl
  *
  * @author Dmitry Solomakha
- * @since 1.3
+ * @since 1.2.4
  */
 
 final class AutoId {
--- a/spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/AutoId.java
+++ b/spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/AutoId.java
@@ -1,0 +1,43 @@
+/*
+ * Copyright 2019-2020 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.gcp.data.firestore;
+
+import java.security.SecureRandom;
+import java.util.Random;
+
+//Copied from com.google.cloud.firestore.FirestoreImpl
+final class AutoId {
+	private static final String AUTO_ID_ALPHABET =
+					"ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789";
+
+	private static final Random RANDOM = new SecureRandom();
+
+	private static final int AUTO_ID_LENGTH = 20;
+
+	private AutoId() {
+	}
+
+	//Creates a pseudo-random 20-character ID that can be used for Firestore documents.
+	static String autoId() {
+		StringBuilder builder = new StringBuilder();
+		int maxRandom = AUTO_ID_ALPHABET.length();
+		for (int i = 0; i < AUTO_ID_LENGTH; i++) {
+			builder.append(AUTO_ID_ALPHABET.charAt(RANDOM.nextInt(maxRandom)));
+		}
+		return builder.toString();
+	}
+}
--- a/spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/AutoId.java
+++ b/spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/AutoId.java
@@ -19,7 +19,14 @@ package org.springframework.cloud.gcp.data.firestore;
 import java.security.SecureRandom;
 import java.util.Random;
 
-//Copied from com.google.cloud.firestore.FirestoreImpl
+/**
+ * This class is used to generate Id values.
+ * Copied from com.google.cloud.firestore.FirestoreImpl
+ *
+ * @author Dmitry Solomakha
+ * @since 1.3
+ */
+
 final class AutoId {
 	private static final String AUTO_ID_ALPHABET =
 					"ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789";
--- a/spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreTemplate.java
+++ b/spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreTemplate.java
@@ -431,7 +431,7 @@ public class FirestoreTemplate implements FirestoreReactiveOperations {
 				throw new FirestoreDataException("Automatic ID generation only supported for String type");
 			}
 
-			//TODO: replace with Internal.autoId() when it is available
+			//TODO: replace with com.google.cloud.firestore.Internal.autoId() when it is available
 			idVal = autoId();
 			persistentEntity.getPropertyAccessor(entity).setProperty(idProperty, idVal);
 		}
@@ -454,7 +454,8 @@ public class FirestoreTemplate implements FirestoreReactiveOperations {
 		return this.classMapper;
 	}
 
-	/** Creates a pseudo-random 20-character ID that can be used for Firestore documents. */
+	//Creates a pseudo-random 20-character ID that can be used for Firestore documents.
+	//Copied from com.google.cloud.firestore.FirestoreImpl
 	private static String autoId() {
 		StringBuilder builder = new StringBuilder();
 		int maxRandom = AUTO_ID_ALPHABET.length();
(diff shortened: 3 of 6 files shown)


spring-cloud-gcp-data-firestore/pom.xml
```diff
@@ -23,6 +23,7 @@
         <dependency>
             <groupId>com.google.cloud</groupId>
             <artifactId>google-cloud-firestore</artifactId>
+			<version>1.35.2-SNAPSHOT</version>
```
dzou, line 26: Fix indentation here; tab/space mismatch.

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreTemplate.java
```diff
@@ -189,6 +193,26 @@ public boolean isUsingStreamTokens() {
 		return saveAll(Mono.just(entity)).next();
 	}
 
+	private <T> Flux<T> create(Publisher<T> entities) {
+		return Flux.from(entities).flatMap(this::create);
+	}
+
+	private <T> Mono<T> create(T entity) {
```
dzou, line 200: Would be nice to add a javadoc for this one describing purpose of method something like "Creates a single (non-batched) document in Firestore; ID generation is only available for non-batched created documents."

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreTemplate.java
```diff
@@ -207,12 +231,18 @@ public boolean isUsingStreamTokens() {
 				return Flux.from(instances).doOnNext(t -> writes.add(createUpdateWrite(t)));
 			}
 
-			Flux<List<T>> inputs = Flux.from(instances).bufferTimeout(this.writeBufferSize, this.writeBufferTimeout);
-			return ObservableReactiveUtil.streamingBidirectionalCall(
-					this::openWriteStream, inputs, this::buildWriteRequest);
+			return  Flux.from(instances)
+							.groupBy(t -> getIdValue(t) == null).flatMap(groupedFlux ->
```
dzou, line 235: Minor nit, would prefer if you can format this section so each operator gets its own line like this:

```
return  Flux.from(instances)
							.groupBy(t -> getIdValue(t) == null)
              .flatMap(groupedFlux -> groupedFlux.key() ? create(groupedFlux) : upsert(groupedFlux));
```

Clever usage of Group By!

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreTemplate.java
```diff
@@ -400,34 +430,65 @@ private void doIfTransaction(Context ctx, Consumer<ReactiveFirestoreResourceHold
...
 
-		return idVal.toString();
+		return idVal;
 	}
 
 	public FirestoreClassMapper getClassMapper() {
 		return this.classMapper;
 	}
+
+	private static class ResourceName {
```
dzou, line 476: I see; I would avoid introducing a new class here if possible -

I think if you include it, you will have to write a lot more docs to describe what it means for future readers of this code - i.e. it would at least require a class-level javadoc, and also - `generated` is too general (i.e. generated what?); you may have to rename to `needsGeneratedId` or add comment; similarly `name` is too general as well - i.e. (name of what?)

It's not worth the burden of adding this class in my opinion if it is avoidable.

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreTemplate.java
```diff
@@ -400,34 +430,65 @@ private void doIfTransaction(Context ctx, Consumer<ReactiveFirestoreResourceHold
 	}
 
 	private <T> Write createUpdateWrite(T entity) {
-		String documentResourceName = buildResourceName(entity);
-		Document document = getClassMapper().entityToDocument(entity, documentResourceName);
-		return Write.newBuilder()
-				.setUpdate(document)
-				.build();
+		ResourceName resourceName = buildResourceName(entity);
```
dzou, line 433: I would avoid introducing the `ResourceName` class to avoid coupling the resource name with whether the `id` was generated or not. Also mentioned below; I think it's not worth the documentation burden.

Instead, maybe refactor so it's like:

```
Object idVal = getOrGenerateEntityId(entity);
String resourceName = buildResourceName(Class<?> entityType, Object idVal);
```

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreTemplate.java
```diff
@@ -400,34 +411,56 @@ private void doIfTransaction(Context ctx, Consumer<ReactiveFirestoreResourceHold
...
-		String documentResourceName = buildResourceName(entity);
-		Document document = getClassMapper().entityToDocument(entity, documentResourceName);
-		return Write.newBuilder()
-				.setUpdate(document)
-				.build();
+		boolean needsAutoId = getIdValue(entity) == null;
+		String resourceName = buildResourceName(entity);
+		Document document = getClassMapper().entityToDocument(entity, resourceName);
+		Builder builder = Write.newBuilder().setUpdate(document);
+		if (needsAutoId) {
```
meltsufin, line 418: I don't really see the point of having this `needsAutoId` local variable, since it's only used once.
I would just put `getIdValue(entity) == null` directly in the `if`.

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreTemplate.java
```diff
@@ -400,34 +411,56 @@ private void doIfTransaction(Context ctx, Consumer<ReactiveFirestoreResourceHold
...
 		FirestorePersistentEntity<?> persistentEntity =
 				this.mappingContext.getPersistentEntity(entity.getClass());
 		FirestorePersistentProperty idProperty = persistentEntity.getIdPropertyOrFail();
 		Object idVal = persistentEntity.getPropertyAccessor(entity).getProperty(idProperty);
+		if (idVal == null) {
+			if (idProperty.getType() != String.class) {
+				throw new FirestoreDataException("Automatic ID generation only supported for String type");
+			}
 
+			//TODO: replace with Internal.autoId() when it is available
```
meltsufin, line 434: Indicate the package of `Internal`? Otherwise, it's ambiguous.

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreTemplate.java
```diff
@@ -400,34 +411,56 @@ private void doIfTransaction(Context ctx, Consumer<ReactiveFirestoreResourceHold
...
 	}
 
 	private <T> String buildResourceName(T entity) {
 		FirestorePersistentEntity<?> persistentEntity =
 				this.mappingContext.getPersistentEntity(entity.getClass());
 		FirestorePersistentProperty idProperty = persistentEntity.getIdPropertyOrFail();
 		Object idVal = persistentEntity.getPropertyAccessor(entity).getProperty(idProperty);
+		if (idVal == null) {
+			if (idProperty.getType() != String.class) {
+				throw new FirestoreDataException("Automatic ID generation only supported for String type");
```
meltsufin, line 431: Maybe also say that id was `null` and what its type was to make the exception more clear.

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreTemplate.java
```diff
@@ -400,34 +411,56 @@ private void doIfTransaction(Context ctx, Consumer<ReactiveFirestoreResourceHold
...
 
-		return idVal.toString();
+		return persistentEntity.getPropertyAccessor(entity).getProperty(idProperty);
 	}
 
 	public FirestoreClassMapper getClassMapper() {
 		return this.classMapper;
 	}
+
+	/** Creates a pseudo-random 20-character ID that can be used for Firestore documents. */
```
meltsufin, line 457: Can you indicate where it was copied from?

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/mapping/FirestoreDefaultClassMapper.java
```diff
@@ -56,9 +57,12 @@ public FirestoreDefaultClassMapper() {
 	public <T> Document entityToDocument(T entity, String documentResourceName) {
 		DocumentSnapshot documentSnapshot = INTERNAL.snapshotFromObject(NOT_USED_PATH, entity);
 		Map<String, Value> valuesMap = INTERNAL.protoFromSnapshot(documentSnapshot);
-		return Document.newBuilder()
-				.putAllFields(valuesMap)
-				.setName(documentResourceName).build();
+		Builder builder = Document.newBuilder().putAllFields(valuesMap);
+		if (documentResourceName != null) {
```
dzou, line 61: Why is it now possible for `documentResourceName` to be null?

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/FirestoreTemplate.java
```diff
@@ -400,34 +411,56 @@ private void doIfTransaction(Context ctx, Consumer<ReactiveFirestoreResourceHold
...
+		return persistentEntity.getPropertyAccessor(entity).getProperty(idProperty);
 	}
 
 	public FirestoreClassMapper getClassMapper() {
 		return this.classMapper;
 	}
+
+	//Creates a pseudo-random 20-character ID that can be used for Firestore documents.
+	//Copied from com.google.cloud.firestore.FirestoreImpl
+	private static String autoId() {
```
dzou, line 458: Could you move the Id generating helper method `autoId()` to a package-private class? maybe something like `FirestoreIdGenerator`.

This way you can move all the code and fields in there related to ID generation; which allows you to remember what to delete once `com.google.cloud.firestore.Internal.autoId()` becomes available. Otherwise you will have to comb through the code a little bit more in the future to do this removal.

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/AutoId.java
```diff
@@ -0,0 +1,43 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.gcp.data.firestore;
+
+import java.security.SecureRandom;
+import java.util.Random;
+
+//Copied from com.google.cloud.firestore.FirestoreImpl
```
dzou, line 22: For new classes, add a top level Javadoc :) i.e. with the author tag and brief description etc.

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/mapping/FirestoreDefaultClassMapper.java
```diff
@@ -56,9 +57,10 @@ public FirestoreDefaultClassMapper() {
 	public <T> Document entityToDocument(T entity, String documentResourceName) {
 		DocumentSnapshot documentSnapshot = INTERNAL.snapshotFromObject(NOT_USED_PATH, entity);
 		Map<String, Value> valuesMap = INTERNAL.protoFromSnapshot(documentSnapshot);
-		return Document.newBuilder()
-				.putAllFields(valuesMap)
-				.setName(documentResourceName).build();
+		Builder builder = Document.newBuilder().putAllFields(valuesMap);
```
dzou, line 60: Might as well just revert this file then if there are no changes.

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/AutoId.java
dmitry-s: That's why I didn't want to introduce a new class :)

dzou: Lol sorry for the inconvenience. At least it will be deleted shortly though! (Once they expose auto id).

PR #2471, opened 2020-07-20
fix null handling for id in query by example

### LINKED ISSUE #2470: Query by example API doesn't follow documentation
According to these documentations for the Query by Example API, parameters set to null are suppose to be excluded from query by default. However, one of the parameters (an entity's datastore id) was set to null and was still included in the query. The application threw an error because ids cannot be null. 

Using version 2.3.1
https://cloud.spring.io/spring-cloud-gcp/multi/multi__spring_data_cloud_datastore.html#_query_by_example
https://docs.spring.io/spring-data/jpa/docs/current/reference/html/#query-by-example

Reproduce issue: 
- `git clone --single-branch --branch filtered-search git@github.com:googleinterns/step90-2020.git`
- `git checkout 6e64536705c950297783ec53a343dc9493528c6f`
- Go to http://localhost:8080/
- Log in with **username**: student@uni.org and **password**: test
- Click "Submit" button

Error Message:
image

dmitry-s:
fixes #2470

### DIFF
--- a/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
+++ b/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
@@ -772,8 +772,10 @@ public class DatastoreTemplate implements DatastoreOperations, ApplicationEventP
 				Value<?> value;
 				if (persistentProperty.isIdProperty()) {
 					PersistentPropertyAccessor accessor = persistentEntity.getPropertyAccessor(probe);
-					value = KeyValue.of(
-							createKey(persistentEntity.kindName(), accessor.getProperty(persistentProperty)));
+					Object property = accessor.getProperty(persistentProperty);
+					value = property != null
+									? KeyValue.of(createKey(persistentEntity.kindName(), property))
+									: NullValue.of();
 				}
 				else {
 					value = probeEntity.getValue(fieldName);
--- a/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
+++ b/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
@@ -50,6 +50,7 @@ import com.google.cloud.datastore.PathElement;
 import com.google.cloud.datastore.Query;
 import com.google.cloud.datastore.QueryResults;
 import com.google.cloud.datastore.StructuredQuery;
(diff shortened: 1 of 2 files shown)


spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
```diff
@@ -765,31 +767,13 @@ private DatastoreReaderWriter getDatastoreReadWriter() {
...
-				}
-				if (value instanceof NullValue
-						&& example.getMatcher().getNullHandler() != ExampleMatcher.NullHandler.INCLUDE) {
-					//skip null value
-					return;
-				}
-				filters.add(StructuredQuery.PropertyFilter.eq(fieldName, value));
+			if (notIgnoredProperty(example, persistentProperty)) {
+				Value<?> value = getValue(example, probeEntity, persistentEntity, persistentProperty);
+				NullHandler nullHandler = example.getMatcher().getNullHandler();
```
elefeint, line 772: So the check for `value instanceof NullValue` is not needed?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
```diff
@@ -799,6 +783,43 @@ private DatastoreReaderWriter getDatastoreReadWriter() {
...
+		if (persistentProperty.isIdProperty()) {
+			value = getIdValue(example, persistentEntity, persistentProperty);
+		}
+		else {
+			value = probeEntity.getValue(persistentProperty.getFieldName());
+		}
+		return value;
+	}
+
+	private <T> boolean notIgnoredProperty(Example<T> example, DatastorePersistentProperty persistentProperty) {
```
elefeint, line 798: It may be easier to read as `isIgnoredProperty()`, removing the negation here and putting the negation in `getDatastoreReadWriter()` method instead. Our brains are very used to the `isBlah` pattern.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
dmitry-s: It is still required. NullHandler just indicates if we need to include null valued fields or not, and `value instanceof NullValue` is a check to see if a field holds a null value.

elefeint: Oh, it's in addFilter now. I missed it.

PR #2473, opened 2020-07-22
Control sync/async publish in Spring Cloud Stream binder

elefeint:
Expose the Spring Integration `PubSubMessageHandler`'s `sync` property via extended binder properties. Turning async mode off is useful when errors must be exposed to sender.

There is currently no way to force synchronous sending through Cloud Pub/Sub Spring Cloud Stream binder.

### DIFF
--- a/spring-cloud-gcp-pubsub-stream-binder/src/main/java/org/springframework/cloud/gcp/stream/binder/pubsub/properties/PubSubProducerProperties.java
+++ b/spring-cloud-gcp-pubsub-stream-binder/src/main/java/org/springframework/cloud/gcp/stream/binder/pubsub/properties/PubSubProducerProperties.java
@@ -24,4 +24,13 @@ package org.springframework.cloud.gcp.stream.binder.pubsub.properties;
  * @author Chengyuan Zhao
  */
 public class PubSubProducerProperties extends PubSubCommonProperties {
+	private boolean sync;
+
+	public boolean isSync() {
+		return sync;
+	}
+
+	public void setSync(boolean sync) {
+		this.sync = sync;
+	}
 }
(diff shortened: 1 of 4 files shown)


spring-cloud-gcp-pubsub-stream-binder/src/test/java/org/springframework/cloud/gcp/stream/binder/pubsub/PubSubMessageChannelBinderTests.java
```diff
@@ -71,4 +94,25 @@ public void testAfterUnbindConsumer() {
 		verify(this.channelProvisioner).afterUnbindConsumer(this.consumerDestination);
 	}
 
+	@Test
+	public void producerSyncPropertyFalseByDefault() {
```
dzou, line 98: If it's possible, I think it would be better to set the property via `spring.cloud.stream.gcp.pubsub.default.producer.sync` rather than accessing the properties class directly. Mainly to verify that `spring.cloud.stream.gcp.pubsub.default.producer.sync` is indeed the correct format of the property name (as I am not sure what the correct form is myself).

elefeint: Yes, I've been getting increasingly uncomfortable with the amount of mocking in this test. I'll either move this into an integration test, or turn this test into a light integration test with only a mocked pubsub template.

spring-cloud-gcp-pubsub-stream-binder/src/main/java/org/springframework/cloud/gcp/stream/binder/pubsub/properties/PubSubProducerProperties.java
```diff
@@ -24,4 +24,13 @@
  * @author Chengyuan Zhao
  */
 public class PubSubProducerProperties extends PubSubCommonProperties {
+	private boolean sync;
```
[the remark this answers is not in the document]
elefeint: Will do.

spring-cloud-gcp-pubsub-stream-binder/src/test/java/org/springframework/cloud/gcp/stream/binder/pubsub/PubSubMessageChannelBinderTests.java
elefeint: This is the best I could do without testing SCS implementation details. There is still one mocked transition from our properties into a SCS wrapper, but at least it tests the property mapping correctly.

PR #2480, opened 2020-07-29
add FirestoreTemplate docs

### LINKED ISSUE #1795: Add documentation for FirestoreTemplate

dmitry-s:
fixes #1795

### DIFF
--- a/docs/src/main/asciidoc/firestore.adoc
+++ b/docs/src/main/asciidoc/firestore.adoc
@@ -109,8 +109,10 @@ include::{project-root}/spring-cloud-gcp-data-firestore/src/test/java/org/spring
 `@Document(collectionName = "usersCollection")` annotation configures the collection name for the documents of this type.
 This annotation is optional, by default the collection name is derived from the class name.
 
-`@DocumentId` annotation marks a field to be used as document id. This annotation is required.
-NOTE: If the property annotated with `@DocumentId` is `null` and the property type is `String`, the document id is generated automatically when the entity is saved.
+`@DocumentId` annotation marks a field to be used as document id.
+This annotation is required and the annotated field can only be of `String` type.
+
+NOTE: If the property annotated with `@DocumentId` is `null` the document id is generated automatically when the entity is saved.
 
 NOTE: Internally we use Firestore client library object mapping. See https://developers.google.com/android/reference/com/google/firebase/firestore/package-summary[the documentation] for supported annotations.
 
@@ -173,19 +175,19 @@ Using the auto-configuration provided by Spring Data Cloud Firestore, your Sprin
 public class FirestoreTemplateExample {
 
 	@Autowired
-	FirestoreTemplate firestoreTemplate;
+	FirestoreOperations firestoreOperations;
 
 	public Mono<User> createUsers() {
-		return this.firestoreTemplate.save(new User("Alice", 29))
-			.then(this.firestoreTemplate.save(new User("Bob", 60)));
+		return this.firestoreOperatons.save(new User("Alice", 29))
+			.then(this.firestoreOperatons.save(new User("Bob", 60)));
 	}
 
 	public Flux<User> findUsers() {
-		return this.firestoreTemplate.findAll(User.class);
+		return this.firestoreOperatons.findAll(User.class);
 	}
 
 	public Mono<Long> removeAllUsers() {
-		return this.firestoreTemplate.deleteAll(User.class);
+		return this.firestoreOperatons.deleteAll(User.class);
 	}
 }
 ----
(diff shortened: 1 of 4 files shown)


docs/src/main/asciidoc/firestore.adoc
```diff
@@ -110,6 +110,7 @@ include::{project-root}/spring-cloud-gcp-data-firestore/src/test/java/org/spring
 This annotation is optional, by default the collection name is derived from the class name.
 
 `@DocumentId` annotation marks a field to be used as document id. This annotation is required.
+If the value is null, and the property type is `String`, the Id is generated automatically.
```
dzou, line 113:
```suggestion
NOTE: If the property annotated with `@DocumentId` is `null` and the property type is `String`, the document id is generated automatically when the entity is saved.
```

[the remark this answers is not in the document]
dmitry-s: I don't think we should do that. It is implementation detail and could change.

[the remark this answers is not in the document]
dzou: I see, but I think there still should be some general description of this generated ID. How would the user know if it is a sequence-based generation vs. random-based?

docs/src/main/asciidoc/firestore.adoc
```diff
@@ -175,13 +175,17 @@ public class FirestoreTemplateExample {
 	@Autowired
 	FirestoreTemplate firestoreTemplate;
 
-	public void doSomething() {
+	public Mono<User> createUsers() {
+		return this.firestoreTemplate.save(new User("Alice", 29))
+		.then(this.firestoreTemplate.save(new User("Bob", 60)));
```
dzou, line 180:
```suggestion
			.then(this.firestoreTemplate.save(new User("Bob", 60)));
```

docs/src/main/asciidoc/firestore.adoc
```diff
@@ -110,7 +110,7 @@ include::{project-root}/spring-cloud-gcp-data-firestore/src/test/java/org/spring
 This annotation is optional, by default the collection name is derived from the class name.
 
 `@DocumentId` annotation marks a field to be used as document id. This annotation is required.
-If the value is null, and the property type is `String`, the Id is generated automatically.
+NOTE: If the property annotated with `@DocumentId` is `null` and the property type is `String`, the document id is generated automatically when the entity is saved.
```
dzou, line 113: Add a new line above the `NOTE:` so that it gets rendered correctly as a note in the asciidoc.

docs/src/main/asciidoc/firestore.adoc
```diff
@@ -110,6 +110,7 @@ include::{project-root}/spring-cloud-gcp-data-firestore/src/test/java/org/spring
 This annotation is optional, by default the collection name is derived from the class name.
 
 `@DocumentId` annotation marks a field to be used as document id. This annotation is required.
+NOTE: If the property annotated with `@DocumentId` is `null` and the property type is `String`, the document id is generated automatically when the entity is saved.
```
meltsufin, line 113: @dmitry-s Could the id field be an `Integer`? In that case, what happens if it's `null`?
It just seems odd that it's so specific to `String`.

docs/src/main/asciidoc/firestore.adoc
```diff
@@ -159,6 +160,43 @@ Repositories allow you to define custom Query Methods (detailed in the following
 
 NOTE: Custom queries with `@Query` annotation are not supported since there is no query language in Cloud Firestore
 
+
+=== Firestore Operations & Template
+
+`FirestoreOperations` and its implementation, `FirestoreTemplate`, provides the Template pattern familiar to Spring developers.
+
+Using the auto-configuration provided by Spring Data Cloud Firestore, your Spring application context will contain a fully configured `FirestoreTemplate` object that you can autowire in your application:
```
meltsufin, line 168: What not recommend autowiring `FirestoreOperations`?

docs/src/main/asciidoc/firestore.adoc
dmitry-s: It can't be an `Integer`. The client library can generate `String` ids only.

docs/src/main/asciidoc/firestore.adoc
meltsufin: @dmitry-s This is still not addressed. It won't render correctly otherwise.

docs/src/main/asciidoc/firestore.adoc
meltsufin: Well, _we're_ actually generating the IDs, but yes, because they need to be very random, it only works for strings.

I would maybe clarifying by adding that ID auto-generation does not work for types other than `String`.

dmitry-s: Actually we only support `String` ids. I'll document this.

docs/src/main/asciidoc/firestore.adoc
dmitry-s: switched to `FirestoreOperations`

Issue #2459, opened 2020-07-11
NullPointerException when a configuration is empty


pijpijpij:
**Describe the bug**
At start-up, with GCP config enabled, the GCP code requires a configuration (`myapp_myprofile`). It is probably a design decision to require the existence of the configuration in GCP Config. Fair enough.
In that particular `myprofile` profile, `myapp` does not need any extra property, so I did not create one.
That results in an NPE when your framework retrieves the configuration.

As a workaround, I created a property named `placeholder` with a value of `boff!`. My app now starts.

**Sample**
I don't think there's a need for that as the code is pretty clear in version 1.2.2.RELEASE which I am using. The NPE occurs on line 50 of `GoogleConfigEnvironment`. Either `variables` hadn't been set or it had been set to `null`. Defensive coding here would address the problem (use `getVariables()` and have it return an empty list instead of `null`).
Should you also want to address the value returned by the API, I am including the stack trace here:
> 2020-07-11 10:26:02 ERROR o.s.b.SpringApplication:826 - Application run failed
java.lang.RuntimeException: Error loading configuration for firmwareupdate-9d157/nurvv-api_firmware
        at org.springframework.cloud.gcp.autoconfigure.config.GoogleConfigPropertySourceLocator.locate(GoogleConfigPropertySourceLocator.ja
va:146)
        at org.springframework.cloud.bootstrap.config.PropertySourceBootstrapConfiguration.initialize(PropertySourceBootstrapConfiguration.
java:98)
        at org.springframework.boot.SpringApplication.applyInitializers(SpringApplication.java:626)
        at org.springframework.boot.SpringApplication.prepareContext(SpringApplication.java:370)
        at org.springframework.boot.SpringApplication.run(SpringApplication.java:314)
        at org.springframework.boot.web.servlet.support.SpringBootServletInitializer.run(SpringBootServletInitializer.java:152)
        at org.springframework.boot.web.servlet.support.SpringBootServletInitializer.createRootApplicationContext(SpringBootServletInitiali
zer.java:132)
        at org.springframework.boot.web.servlet.support.SpringBootServletInitializer.onStartup(SpringBootServletInitializer.java:92)
        at org.springframework.web.SpringServletContainerInitializer.onStartup(SpringServletContainerInitializer.java:172)
        at org.apache.catalina.core.StandardContext.startInternal(StandardContext.java:5135)
        at org.apache.catalina.util.LifecycleBase.start(LifecycleBase.java:183)
[...]

PR #2486, opened 2020-08-11
Add release notes for 1.2.4.RELEASE

meltsufin:


meltsufin:
> One more under General:
> 
> * Fixed NPE on empty/missing configuration ([https://github.com/spring-cloud/spring-cloud-gcp/issues/2459[#2459]](https://github.com/spring-cloud/spring-cloud-gcp/issues/2459%5B#2459%5D), thanks to @kioie)

This is part of the runtime config support, which is BETA, and we still don't publish it with the releases. 

> Do we want to mention the new metrics sample app? (#2455)

I thought the general announcement about the new module was enough, but sure we can add specific mention of the sample.

PR #2493, opened 2020-08-21
Support composite actuator contributor for multiple Pub/Sub templates.

### LINKED ISSUE #2491: CompositeHealthContributor instead of simple HealthIndicator
**Is your feature request related to a problem? Please describe.**
I am using spring-cloud-gcp 1.2.3 with spring boot 2.2.3.
My application instanciate two PubsubTemplate beans because I need to communicate with two PubSub from two Google Cloud projects.
If I enable the PubSub Health Indicator, my application can not start because of this error :
```

Parameter 0 of method pubSubHealthIndicator in org.springframework.cloud.gcp.autoconfigure.pubsub.health.PubSubHealthIndicatorAutoConfiguration required a single bean, but 2 were found:
	- project1PubSubTemplate: defined by method 'getProject1PubSubTemplate' in class path resource [io/myapp/configuration/PubSubConfiguration.class]
	- project2PubSubTemplate: defined by method 'getProject1PubSubTemplate' in class path resource [io/myapp/configuration/PubSubConfiguration.class]

```

**Describe the solution you'd like**
To avoid this, I set `management.health.pubsub.enabled=false` and created a custom CompositeHealthContributor.
```
public class PubSubHealthIndicatorAutoConfiguration extends CompositeHealthContributorConfiguration<PubSubHealthIndicator, PubSubTemplate> {

    @Bean
    @ConditionalOnMissingBean
    public HealthContributor pubSubHealthContributor(Map<String, PubSubTemplate> pubSubTemplates) {
        return createContributor(pubSubTemplates);
    }
}
```

Would this change may take place directly in Spring Cloud GCP instead of having to disable standard health indicator ?

elefeint:
Fixes #2491 

Also refactored health indicator tests into their own class.

### DIFF
--- a/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/health/PubSubHealthIndicatorAutoConfiguration.java
+++ b/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/health/PubSubHealthIndicatorAutoConfiguration.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2018-2019 the original author or authors.
+ * Copyright 2018-2020 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -16,8 +16,12 @@
 
 package org.springframework.cloud.gcp.autoconfigure.pubsub.health;
 
+import java.util.Map;
+
+import org.springframework.boot.actuate.autoconfigure.health.CompositeHealthContributorConfiguration;
 import org.springframework.boot.actuate.autoconfigure.health.ConditionalOnEnabledHealthIndicator;
 import org.springframework.boot.actuate.autoconfigure.health.HealthContributorAutoConfiguration;
+import org.springframework.boot.actuate.health.HealthContributor;
 import org.springframework.boot.actuate.health.HealthIndicator;
 import org.springframework.boot.autoconfigure.AutoConfigureAfter;
 import org.springframework.boot.autoconfigure.AutoConfigureBefore;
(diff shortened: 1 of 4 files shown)


spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/health/PubSubHealthIndicatorAutoConfiguration.java
```diff
@@ -34,21 +38,23 @@
  * {@link PubSubHealthIndicator}.
  *
  * @author Vinicius Carvalho
+ * @author Elena Felder
  *
  * @since 1.2.2
  */
-@Configuration
+@Configuration(proxyBeanMethods = false)
```
meltsufin, line 45: What does this change do?

elefeint: Unrelated to the PR, but Spring Boot classes turn off bean proxying wherever possible for performance and to avoid tripping graal incompatibilities, so I tacked it on here.
`proxyBeanMethods` being on is only useful if you call bean-producing methods directly in autoconfiguration, which we definitely don't here.

elefeint:
Yes, it acquires a grouping level: `"pubSub":{"status":"UP","components":{"pubSubTemplate1":{"status":"UP"},"pubSubTemplate2":{"status":"UP"}}}`

meltsufin:
> Yes, it acquires a grouping level: `"pubSub":{"status":"UP","components":{"pubSubTemplate1":{"status":"UP"},"pubSubTemplate2":{"status":"UP"}}}`

That's very cool!

elefeint:
I have a customized pubsub sample with multiple (no-op) pubsub templates. Did not seem worth pushing those changes to `spring-cloud-gcp`, since it's definitely not the most common scenario.

I might push it to my examples repo instead.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/health/PubSubHealthIndicatorAutoConfiguration.java
elefeint: Filed #2500 to track.

PR #2499, opened 2020-08-24
Small contribution to automate tests with GitHub Actions

geertjanw:
Hope you like it, this should speed up your testing. :-)

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -1,0 +1,23 @@
+name: Java CI
+
+on: [push]
+
+jobs:
+  test:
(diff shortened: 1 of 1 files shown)


meltsufin:
@geertjanw Thank you for the contribution, but we've already done [something like this in our new repository](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/blob/master/.github/workflows/buildAndTest.yaml) which is set up for the next major version of the project. We may copy that configuration here as well.
Sorry, we haven't popularized the new repo yet, as it's not quite the primary yet.

geertjanw:
Great, no worries, that looks good.

PR #2502, opened 2020-08-26
WIP binder producer error propagation

elefeint:
This is a mix of two things right now:
1) implementation of error propagation, with a property in the sample that needs to move into documentation.
2) reproducing a blocking issue in client library, in which failures in gRPC DNS resolution are not propagated to the returned Future.

### DIFF
--- a/spring-cloud-gcp-pubsub-stream-binder/src/test/java/org/springframework/cloud/gcp/stream/binder/pubsub/TempClientLibraryTest.java
+++ b/spring-cloud-gcp-pubsub-stream-binder/src/test/java/org/springframework/cloud/gcp/stream/binder/pubsub/TempClientLibraryTest.java
@@ -1,0 +1,81 @@
+package org.springframework.cloud.gcp.stream.binder.pubsub;
+
+import com.google.api.core.ApiFutureCallback;
+import com.google.api.core.ApiFutures;
+import com.google.cloud.pubsub.v1.Publisher;
+import com.google.protobuf.ByteString;
+import com.google.pubsub.v1.PubsubMessage;
+import org.awaitility.Awaitility;
+import org.awaitility.Duration;
+import org.junit.Test;
+
+import java.io.IOException;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.atomic.AtomicBoolean;
+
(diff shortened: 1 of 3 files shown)


spring-cloud-gcp-pubsub-stream-binder/src/test/java/org/springframework/cloud/gcp/stream/binder/pubsub/TempClientLibraryTest.java
```diff
@@ -0,0 +1,81 @@
...
+import org.junit.Test;
+
+import java.io.IOException;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+public class TempClientLibraryTest {
+
+	@Test
+	public void failedPublishCallsListener() throws IOException, InterruptedException {
```
elefeint, line 19: This test fails; the future error callback is never invoked.

spring-cloud-gcp-pubsub-stream-binder/src/test/java/org/springframework/cloud/gcp/stream/binder/pubsub/TempClientLibraryTest.java
```diff
@@ -48,6 +50,47 @@ public void onSuccess(String messageId) {
...
+	public void failedPublishCallsListener_overriddenRetrySettingsAllowsAsyncPublishToProperlyFail() throws IOException, InterruptedException {
+		Publisher publisher = Publisher.newBuilder("projects/elfel-spring/topics/exampleTopic")
+				.setEndpoint("cecicestnespasunendpoint:443")
+				.setRetrySettings(
+						RetrySettings.newBuilder()
+								.setMaxAttempts(3)
+								.setTotalTimeout(org.threeten.bp.Duration.ofSeconds(10))
+								.setInitialRpcTimeout(org.threeten.bp.Duration.ofSeconds(10))
+								.setMaxRpcTimeout(org.threeten.bp.Duration.ofSeconds(20)) // TODO: bad message; try 9; longer not shorter
+								.build())
```
elefeint, line 65: This is the minimal set of retry settings that work.

PR #2505, opened 2020-08-26
check value for nullity to avoid NPE with primitive types

### LINKED ISSUE #2497: How to add property to datastore
I am using this datastore dependency: https://cloud.spring.io/spring-cloud-static/Greenwich.RC2/multi/multi__spring_data_cloud_datastore.html#_partial_update_2

I have successfully stored a POJO to datastore and retrieved it.

My problem comes when trying to add a new property to an existing entity in datastore. By updating the POJO class with the new property, attempting to retrieve the entity results in a nullPointerException.

In Google's vanilla(not using Spring) documentation [[link]](https://cloud.google.com/datastore/docs/concepts/entities#updating_an_entity) this is achieved by doing this:
```
Entity task = Entity.newBuilder(datastore.get(taskKey)).set("priority", 5).build();
datastore.update(task);
```
I don't see how I am supposed to be able to add a new property if I can't change the class and therefore can't retrieve the entity.
I see in the documentation "Partial update is not supported yet". Is this what it is referring to? If so I would like to request that feature.

I would love some help on this.

dmitry-s:
fixes #2497

### DIFF
--- a/spring-cloud-gcp-data-datastore/src/test/java/org/springframework/cloud/gcp/data/datastore/it/DatastoreIntegrationTests.java
+++ b/spring-cloud-gcp-data-datastore/src/test/java/org/springframework/cloud/gcp/data/datastore/it/DatastoreIntegrationTests.java
@@ -944,6 +944,7 @@ public class DatastoreIntegrationTests extends AbstractDatastoreIntegrationTests
 		CompanyWithBooleanPrimitive companyWithBooleanPrimitive =
 						this.datastoreTemplate.findById(1L, CompanyWithBooleanPrimitive.class);
 		assertThat(companyWithBooleanPrimitive.name).isEqualTo(company.name);
+		assertThat(companyWithBooleanPrimitive.active).isFalse();
 	}
 }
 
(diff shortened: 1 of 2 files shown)


spring-cloud-gcp-data-datastore/src/test/java/org/springframework/cloud/gcp/data/datastore/it/DatastoreIntegrationTests.java
```diff
@@ -934,6 +934,17 @@ private String generateString(int length) {
...
+
+	@Test
+	public void newFieldTest() {
+		Company company = new Company(1L, Collections.emptyList());
+		company.name = "name1";
+		this.datastoreTemplate.save(company);
+
+		CompanyWithBooleanPrimitive companyWithBooleanPrimitive =
+						this.datastoreTemplate.findById(1L, CompanyWithBooleanPrimitive.class);
+		assertThat(companyWithBooleanPrimitive.name).isEqualTo(company.name);
```
elefeint, line 946: Assert the false-by-default boolean, too?

Alos:
I'm running into the same issue as described here using Spring Cloud GCP 1.2.5.

Issue #2511, opened 2020-09-07
Pub/Sub: TransportChannelProvider MaxInboundMessageSize default value not sufficient


f-loris:
**Describe the bug**
Pub/Sub subscribers can receive up to 20MiB of data, but the auto configuration overrides the TransportChannelProvider of the Pub/Sub library without specifying `setMaxInboundMessageSize` which keeps the default value of 4 MiB. Therefore, it is very likely that an user is facing exceptions: `RESOURCE_EXHAUSTED: gRPC message exceeds maximum size 4194304` when high volumne of data is processed

Here the default settings for subscribes can be seen:
https://github.com/googleapis/java-pubsub/blob/8cf77ee3b31356f342942a9ffaa61db2d0686769/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/stub/SubscriberStubSettings.java#L282
From my point of view spring-cloud-gcp should provide correct default values here:
https://github.com/spring-cloud/spring-cloud-gcp/blob/2adf4328c819a634a3ad7a1627426badf198c4fb/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java#L387

I'm aware that https://github.com/spring-cloud/spring-cloud-gcp/issues/1808 is open and this settings topic might be solved more comprehensive (there also might be a differentiation also within pub/sub), but this might take a while and as an quick improvement I'd suggest to set a proper limit of at least 20 MiB in the auto configuration. This should improve the user experience for now. I'm open to contribute this change.

**Sample**
This 4MiB limit was already discussed in some places here. For example: https://github.com/spring-cloud/spring-cloud-gcp/issues/2476

A workaround written in Kotlin reusing the subscribers default settings:
```
@Bean
@ConditionalOnProperty(prefix = "spring.cloud.gcp.pubsub", name = ["emulator-host"], havingValue = "false", matchIfMissing = true)
fun transportChannelProvider(gcpPubSubProperties: GcpPubSubProperties): TransportChannelProvider {
   return SubscriberStubSettings.defaultGrpcTransportProviderBuilder()
               .setKeepAliveTime(Duration.ofMinutes(gcpPubSubProperties.keepAliveIntervalMinutes.toLong()))
               .build()
}
```

meltsufin:
[...]

PR #2520, opened 2020-09-11
Differentiate between Publisher and Subscriber TransportChannelProvider

meltsufin:
Ensures that the correct client library defaults are applied for `maxInboundMessageSize`.

Fixes: #2511.

### DIFF
--- a/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
+++ b/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
@@ -28,7 +28,6 @@ import com.google.api.gax.core.CredentialsProvider;
 import com.google.api.gax.core.ExecutorProvider;
 import com.google.api.gax.core.FixedExecutorProvider;
 import com.google.api.gax.core.NoCredentialsProvider;
-import com.google.api.gax.grpc.InstantiatingGrpcChannelProvider;
 import com.google.api.gax.retrying.RetrySettings;
 import com.google.api.gax.retrying.RetrySettings.Builder;
 import com.google.api.gax.rpc.HeaderProvider;
@@ -37,6 +36,8 @@ import com.google.cloud.pubsub.v1.SubscriptionAdminClient;
 import com.google.cloud.pubsub.v1.SubscriptionAdminSettings;
 import com.google.cloud.pubsub.v1.TopicAdminClient;
 import com.google.cloud.pubsub.v1.TopicAdminSettings;
+import com.google.cloud.pubsub.v1.stub.PublisherStubSettings;
+import com.google.cloud.pubsub.v1.stub.SubscriberStubSettings;
 import org.threeten.bp.Duration;
 
 import org.springframework.beans.factory.ObjectProvider;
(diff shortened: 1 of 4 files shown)


spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubAutoConfigurationTests.java
```diff
@@ -64,12 +69,33 @@ public void keepAliveValue_custom() {
...
+	public void maxInboundMessageSize_default() {
+		ApplicationContextRunner contextRunner = new ApplicationContextRunner()
+				.withConfiguration(AutoConfigurations.of(GcpPubSubAutoConfiguration.class))
+				.withUserConfiguration(TestConfig.class);
+
+		contextRunner.run(ctx -> {
+
+			TransportChannelProvider subscriberTcp = ctx.getBean("subscriberTransportChannelProvider", TransportChannelProvider.class);
+			assertThat(FieldUtils.readField(subscriberTcp, "maxInboundMessageSize", true))
+					.isEqualTo(20 << 20);
```
elefeint, line 91: define a 20MB constant maybe? Bit shifting is a bit obfuscated.

elefeint (approved):
Also need to document the potentially breaking change -- if someone defined a transport channel without annotations, it will no longer be used.

spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubAutoConfigurationTests.java
meltsufin: It's a copy-paste from the client library.

meltsufin:
> Also need to document the potentially breaking change -- if someone defined a transport channel without annotations, it will no longer be used.

We never documented that customization. So, I think it's fair to break it. I will add it to release notes though.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
```diff
@@ -352,12 +353,12 @@ public TopicAdminClient topicAdminClient(
 	@Bean
 	@ConditionalOnMissingBean
 	public TopicAdminSettings topicAdminSettings(
-			TransportChannelProvider transportChannelProvider) {
+			@Qualifier("publisherTransportChannelProvider") TransportChannelProvider publisherTransportChannelProvider) {
```
elefeint, line 356: I wonder if admin functionality should have its own. The most obvious usecase is needing different credentials for admin operations.

elefeint (approved):
LGTM, but perhaps a follow-up issue to differentiate Topic/Subscriber admin functionality channel.

spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/pubsub/GcpPubSubAutoConfiguration.java
meltsufin: I thought that too, but I saw that they use exactly the same default for admin client.
https://github.com/googleapis/java-pubsub/blob/8cf77ee3b31356f342942a9ffaa61db2d0686769/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/SubscriptionAdminSettings.java#L227

PR #2523, opened 2020-09-15
Firestore extractDatabasePath update

dmitry-s:
Updating extractDatabasePath to avoid `org.apache.commons.lang3` dependency

### DIFF
--- a/spring-cloud-gcp-data-firestore/src/test/java/org/springframework/cloud/gcp/data/firestore/UtilTests.java
+++ b/spring-cloud-gcp-data-firestore/src/test/java/org/springframework/cloud/gcp/data/firestore/UtilTests.java
@@ -14,31 +14,26 @@
  * limitations under the License.
  */
 
-package org.springframework.cloud.gcp.data.firestore.util;
+package org.springframework.cloud.gcp.data.firestore;
+
+import org.junit.Test;
+
+import org.springframework.cloud.gcp.data.firestore.util.Util;
+
+import static org.assertj.core.api.Assertions.assertThat;
 
 /**
  * @author Dmitry Solomakha
+ * @since 1.2
  */
-public final class Util {
+public class UtilTests {
 
-	private Util() {
-	}
+	@Test
+	public void extractDatabasePathTest() {
+		String actualDbPath = "projects/MY_PROJECT/databases/MY_DB_ID";
+		String extractedDbPath = Util.extractDatabasePath(actualDbPath + "/abc/def");
 
-	public static String extractDatabasePath(String parent) {
-		//the parent looks like this: projects/{project_id}/databases/{database_id}/...
-		//and the database path is the first 4 segments, separated by /
-		StringBuilder sb = new StringBuilder();
-		int count = 0;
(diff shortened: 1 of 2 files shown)


spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/util/Util.java
```diff
@@ -27,7 +25,20 @@ private Util() {
 	}
 
 	public static String extractDatabasePath(String parent) {
-		return parent.substring(0, StringUtils.ordinalIndexOf(parent, "/", 4));
```
meltsufin, line 30: What happened to `StringUtils`? Why can't we use it anymore?

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/util/Util.java
```diff
@@ -27,7 +25,20 @@ private Util() {
...
+		int count = 0;
+		for (int i = 0; i < parent.length(); i++) {
+			char c = parent.charAt(i);
+			if (c == '/') {
+				count++;
+			}
+			if (count == 4) {
+				break;
+			}
+			sb.append(c);
```
dzou, line 40: Since you're appending each character at a time anyway in a StringBuilder, might as well just do `String.split("/")` and then do return `arr[0] + "/" + arr[1] + ... + arr[3]`; would be more readable (only 2-3 lines total).

dzou: You can even make it more efficient using the limit arg in String.split: https://docs.oracle.com/javase/7/docs/api/java/lang/String.html#split(java.lang.String,%20int)

dmitry-s: I was trying to avoid unnecessary computation, since split allocates an array and then we need to combine the items. Also, split is a regex based method so the computational complexity is higher.

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/util/Util.java
dmitry-s: `org.apache.commons.lang3` was removed from dependencies and I don't want to bring it in only for this one method.

spring-cloud-gcp-data-firestore/src/main/java/org/springframework/cloud/gcp/data/firestore/util/Util.java
dzou: I see, makes sense.

meltsufin (approved):

meltsufin:
@dmitry-s Notice the build error though.

meltsufin (approved):

PR #2526, opened 2020-09-18
Fixes build by improving performance of LogEntry reads

meltsufin:
The key change to the tests is to filter on timestamp to speed up LogEntry reads from minutes to miliseconds.


dmitry-s (approved):
LGTM

dzou (approved):

PR #2529, opened 2020-09-25
dependency: bump spring-javaformat-checkstyle.version to 0.0.25

romani:
this version bump will allow checkstyle team to reuse this project in regression testing on each commit.
it will make sure there is no false positives be introduced in new version of checkstyle to this project.

PR #2530, opened 2020-09-25
Bring samples checkstyle in sync with root checkstyle version

elefeint:
#2529 updated the main spring checkstyle rules version. This PR brings samples version in line with it.

I've filed GoogleCloudPlatform/spring-cloud-gcp#57 to remove the exclusion in the new repo that's on Illford, since Illford checkstyle rules are kept up to date in Spring Cloud Build.

### DIFF
--- a/spring-cloud-gcp-samples/pom.xml
+++ b/spring-cloud-gcp-samples/pom.xml
@@ -79,7 +79,8 @@
 							<dependency>
 								<groupId>io.spring.javaformat</groupId>
 								<artifactId>spring-javaformat-checkstyle</artifactId>
-								<version>0.0.9</version>
+                                <!-- Keep in sync with spring-javaformat-checkstyle.version in pom.xml -->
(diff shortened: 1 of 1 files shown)


spring-cloud-gcp-samples/pom.xml
```diff
@@ -79,7 +79,8 @@
 							<dependency>
 								<groupId>io.spring.javaformat</groupId>
 								<artifactId>spring-javaformat-checkstyle</artifactId>
-								<version>0.0.9</version>
+                                <!-- Keep in sync with spring-javaformat-checkstyle.version in pom.xml -->
+                                <version>0.0.25</version>
```
meltsufin, line 83: I think we need tabs here. :)

elefeint: sheesh. Checkstyle could have told me, ya know?

meltsufin (approved):

PR #2531, opened 2020-09-25
Re-enable checkstyle on samples by default

elefeint:
I created the full-checkstyle profile because `-Dcheckstyle.skip` did not work for a while. It works fine now! Also, the special checkstyle profile for samples proved to have been a bad decision -- we regularly forget it's there and assume the sample code passes locally, only to be surprised when creating a pull request.

This PR:
* Removes the full-checkstyle profile from samples parent, instead running checkstyle by default.
* Removes redundant dependencies from the plugin -- `spring-cloud-build-tools` brings them in transitively.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,7 +32,7 @@ script:
       ./mvnw test -B -P codecov ${INTEGRATION_TEST_FLAGS} && bash <(curl -s https://codecov.io/bash) -F integration;
     fi;
 install:
-  - ./mvnw -T 1.5C install -DskipTests=true -Dmaven.javadoc.skip=true -Dspring.profiles.active=spring -P full-checkstyle -B -V
+  - ./mvnw -T 1.5C install -DskipTests=true -Dmaven.javadoc.skip=true -Dspring.profiles.active=spring -B -V
 before_install:
   - if [ "$TRAVIS_SECURE_ENV_VARS" == "true" ] && [ "$TRAVIS_PULL_REQUEST" == "false" ]; then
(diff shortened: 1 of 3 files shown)


meltsufin (commented):
Looking back (#2062), one of the main reasons for the `full-checkstyle` profile was to disable style checks for people who just want to play with the samples. Are you proposing to give up that concern now?

From that PR description:
> make it possible to run full checkstyle locally (so the contributor does not get surprised by new checkstyle requirements after PR is created).

Also, since `-Dcheckstyle.skip` works, can you just add that property to the default profile?

elefeint:
Right, so now that `-Dcheckstyle.skip` works, casual users can override checkstyle easily. We could make it default behavior for samples, but it's been causing development flow problems -- you _think_ that the sample code is good, create a pull request, and get surprised with avoidable checkstyle failures.

ttomsu:
Sidebar: it looks like intellij has support to auto fix some checkstyle issues, which is pretty cool: https://stackoverflow.com/questions/19952500/how-fix-all-check-style-issues-with-intellij-idea

meltsufin:
> Right, so now that `-Dcheckstyle.skip` works, casual users can override checkstyle easily. We could make it default behavior for samples, but it's been causing development flow problems -- you _think_ that the sample code is good, create a pull request, and get surprised with avoidable checkstyle failures.

Sounds good. I was always leaning towards leaving the checkstyle on for samples anyway.
In any case, the PR title should probably be more like "Re-enable checkstyle fo samples" to be clear what it's doing.

meltsufin (approved):

Issue #2538, opened 2020-10-02
Bound the type of PubSubSubscriberOperations#ack parameter


celcius112:
At the moment [PubSubSubscriberOperations](https://github.com/spring-cloud/spring-cloud-gcp/blob/master/spring-cloud-gcp-pubsub/src/main/java/org/springframework/cloud/gcp/pubsub/core/subscriber/PubSubSubscriberOperations.java)'s `#ack`, `#nack` and `#modifyAckDeadline` methods accept a Collection of `AcknowledgeablePubsubMessage`. It would be appreciated to bound this Collection so that, for instance, a List of `ConvertedAcknowledgeablePubsubMessage` could be used as parameter.

This would bring a breaking change as the signature of the public interface `PubSubSubscriberOperations` would be modified.

As a temporary workaround, we can wrap our list in a `Collections.unmodifiableList` in order to downcast the type, for instance
```java
List<ConvertedAcknowledgeablePubsubMessage<T>> messages...
pubSubTemplate.ack(Collections.unmodifiableList(messages));
```
but this is less than optimal.

If necessary, and since the modification is quite trivial, I can provide a PR.

meltsufin:
What if we changed it to `ack(Collection<? extends AcknowledgeablePubsubMessage> messages)`? I think that would work without introducing a breaking change.

celcius112:
That's what I have in mind. However this is a breaking change, albeit a minor one, since the signature of the public interface's methods does change.

meltsufin:
Callers of this method do not need to change their code. So, it's backwards compatible in that sense.
Feel free to propose the change in a PR, we can go from there.
Thanks!

celcius112:
Good enough for me :) Thank you for your time

PR #2539, opened 2020-10-03
Bound the acknowledging operation methods' type parameter

celcius112:
Modify PubSubSubscriberOperations#ack, PubSubSubscriberOperations#nack and PubSubSubscriberOperations#modifyAckDeadline.
This is a breaking change for implementations of PubSubSubscriberOperations, but not for callers.

see https://github.com/spring-cloud/spring-cloud-gcp/issues/2538


meltsufin (approved):
Thanks!

PR #2541, opened 2020-10-06
Add query methods that return Slice to FirestoreTemplate to allow pagination

### LINKED ISSUE #2535: DatastoreTemplate DatastoreResultsIterable Cursor should be null when there is no next page
I am trying to implement cursor-based pagination and found that unable to determine whether DatastoreResultsIterable has the next page. Due to the DatastoreResultsIterable Cursor has NotNull value even there is no next page.

Given:
Datastore has some data in KIND
pageLimit > rowsCount for KIND

```
EntityQuery.Builder queryBuilder = Query.newEntityQueryBuilder()
				.setKind(KIND)
				.setOrderBy(StructuredQuery.OrderBy.desc(KEY))
				.setLimit(pageLimit);
```
```
if (urlSafeCursor != null) {
    queryBuilder.setStartCursor(Cursor.fromUrlSafe(urlSafeCursor));
}
```

When:
`DatastoreResultsIterable<DO> resultList = datastoreTemplate.query(queryBuilder.build(), DO.class);`

Expected:
resultList.getCursor() == null

Actual:
resultList.getCursor() != null

dmitry-s:
fixes #2535


spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreNextPageAwareResultsIterable.java
```diff
@@ -0,0 +1,51 @@
+/*
+ * Copyright 2019-2019 the original author or authors.
```
dzou, line 2: Copyright 2017-2020

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreNextPageAwareResultsIterable.java
```diff
@@ -0,0 +1,51 @@
...
+
+import com.google.cloud.datastore.Cursor;
+
+/**
+ * An iterable that contains a cursor for the next page and can be used to determine
+ * if the next page exists.
+ *
+ * @author Dmitry Solomakha
+ */
+public class DatastoreNextPageAwareResultsIterable<T> implements Iterable<T> {
```
dzou, line 29: I don't think this is quite the right approach; it seems a bit heavyweight to create a new class to hold a boolean.

Also the only method that returns this is `nextPageAwareQuery`; however, other query methods such as `query`, `queryIterable`, and `queryByExample` also return `DatastoreResultsIterable` but are not able to take advantage of the next-page-aware feature.

I would try to get `DatastoreResultsIterable` to have access to the `hasNextPage` somehow.

I would suggest a couple alternatives:

1. Is it possible to add a method to DatastoreTemplate like `hasNextPage(Cursor c)` or `hasNextPage(DatastoreResultsIterable i)`? if so, the user can manually check for next page through usage of template.

2. If the query is cheap, you might as well pay the cost of `nextPageExists(..)` for each DatastoreResultsIterable; and build the `hasNextPage` parameter in there. no need for premature optimization

3. Add a method called `hasNextPage()` on the `DatastoreResultsIterable` which will run the query of checking if the cursor has a next page. This would require the Iterable obj have reference to the datastore client I suppose.

meltsufin: +1
Can we just extend `DatastoreResultsIterable`?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
```diff
@@ -276,6 +277,21 @@ public long count(Class<?> entityClass) {
...
 	}
 
+	@Override
+	public <T> DatastoreNextPageAwareResultsIterable<?> nextPageAwareQuery(StructuredQuery query, Class<T> entityClass) {
+		DatastoreResultsIterable resultsIterable = queryKeysOrEntities(query, entityClass);
+
+		return new DatastoreNextPageAwareResultsIterable(resultsIterable, nextPageExists(query, resultsIterable.getCursor()));
+	}
+
+	private boolean nextPageExists(StructuredQuery query, Cursor cursorAfter) {
```
meltsufin, line 287: Can we just move this code to `hasNextPage` in the `Iterable`?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/PartTreeDatastoreQuery.java
```diff
@@ -248,21 +248,19 @@ private Slice executeSliceQuery(Object[] parameters) {
...
+				StreamSupport.stream(results.spliterator(), false).collect(Collectors.toList());
 
 		return (Slice) this.processRawObjectForProjection(
-				new SliceImpl(result, pageable, nextResult.iterator().hasNext()));
+				new SliceImpl(result, getPageable(parameters, results.getCursor()), results.hasNextPage()));
+	}
+
+	private Pageable getPageable(Object[] parameters, Cursor cursor) {
+		ParameterAccessor paramAccessor = new ParametersParameterAccessor(getQueryMethod().getParameters(), parameters);
+		Pageable pageable = DatastorePageable.from(paramAccessor.getPageable(), cursor, null);
```
meltsufin, line 262: You can just return here without assigning to `pageable`.

meltsufin (changes requested):

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreResultsIterable.java
```diff
@@ -50,4 +61,20 @@ public Cursor getCursor() {
 	public Iterable<T> getIterable() {
 		return this.iterable;
 	}
+
+	public Optional<Boolean> getHasNextPage() {
```
meltsufin, line 65: why not just
`public boolean hasNextPage()`?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreResultsIterable.java
```diff
@@ -50,4 +61,20 @@ public Cursor getCursor() {
...
+		return hasNextPage.get();
+	}
+
+	public void setHasNextPageQuery(Supplier<Boolean> query) {
```
meltsufin, line 69: I don't think a `public` setter makes sense here. Can you just provide all the necessary objects to determine `hasNextPage` into the constructor?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
```diff
@@ -272,10 +273,25 @@ public long count(Class<?> entityClass) {
 			resultsIterable = new DatastoreResultsIterable<>(convertEntitiesForRead(results, entityClass),
 					results.getCursorAfter());
 		}
+		if (query instanceof StructuredQuery) {
+			resultsIterable.setHasNextPageQuery(
```
meltsufin, line 277: Instead of calling `setHasNextPageQuery` can this block of code just move to `DatastoreResultsIterable`?
Also, why does this only apply to `StructuredQuery`?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
meltsufin: Same question as before.

meltsufin (changes requested):

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreResultsIterable.java
dmitry-s: @meltsufin Because it is not always there. We can only compute it for `StructuredQuery` queries, and we can't compute it for GQL queries. So we need to represent 3 values - true, false, empty.

meltsufin: @dmitry-s I must be missing something here. Both, `GqlQuery` and `StructuredQuery`, support cursors. So, I would image that it's possible to try to retrieve the next item using the cursor to determine `hasNext()`. How are they different in this respect?
Also, if there is a fundamental difference, then we should probably have two different classes for their results. So, maybe extend `DatastoreResultsIterable` and add `hasNextPage()`?

dmitry-s: The problem with `GqlQuery` is we need to set limit to 1 to avoid fetching the whole next page, and `GqlQuery` doesn't allow to change that.

meltsufin: So, for `GqlQuery` we would need to fetch the whole page? Is there a problem with that other than the potential performance overhead?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreNextPageAwareResultsIterable.java
```diff
@@ -0,0 +1,65 @@
...
+
+import com.google.cloud.datastore.Cursor;
+import com.google.cloud.datastore.StructuredQuery;
+
+import org.springframework.data.util.Lazy;
+
+public class DatastoreNextPageAwareResultsIterable<T> implements DatastoreResultsIterable<T> {
+	private DatastoreResultsIterable datastoreResultsIterable;
+
+	private Cursor cursor;
```
meltsufin, line 29: Never used.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreNextPageAwareResultsIterable.java
```diff
@@ -0,0 +1,65 @@
...
+package org.springframework.cloud.gcp.data.datastore.core;
+
+import java.util.Iterator;
+
+import com.google.cloud.datastore.Cursor;
+import com.google.cloud.datastore.StructuredQuery;
+
+import org.springframework.data.util.Lazy;
+
+public class DatastoreNextPageAwareResultsIterable<T> implements DatastoreResultsIterable<T> {
```
meltsufin, line 26: `DatastoreStructuredQueryResultsIterable` maybe?

meltsufin, line 26: Please add class-level javadoc and `@since 1.2.6`.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreNextPageAwareResultsIterable.java
```diff
@@ -0,0 +1,65 @@
...
+	private DatastoreResultsIterable datastoreResultsIterable;
+
+	private Cursor cursor;
+
+	private StructuredQuery structuredQuery;
```
meltsufin, line 31: Never used.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
```diff
@@ -266,16 +267,28 @@ public long count(Class<?> entityClass) {
...
 		}
+		if (query instanceof StructuredQuery) {
+			resultsIterable = new DatastoreNextPageAwareResultsIterable(
+					resultsIterable, results.getCursorAfter(), (StructuredQuery) query, this);
+		}
 		maybeEmitEvent(new AfterQueryEvent(resultsIterable, query));
 		return resultsIterable;
 	}
 
+	public boolean nextPageExists(StructuredQuery query, Cursor cursorAfter) {
```
meltsufin, line 284: Javadoc missing

`queryHasNextPage` maybe?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreNextPageAwareResultsIterable.java
```diff
@@ -0,0 +1,65 @@
...
+
+import java.util.Iterator;
+
+import com.google.cloud.datastore.Cursor;
+import com.google.cloud.datastore.StructuredQuery;
+
+import org.springframework.data.util.Lazy;
+
+public class DatastoreNextPageAwareResultsIterable<T> implements DatastoreResultsIterable<T> {
+	private DatastoreResultsIterable datastoreResultsIterable;
```
meltsufin, line 27: I don't think you're gaining much by delegating to another iterable. I would just use the currently unused instance variables instead.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreSimpleResultsIterable.java
```diff
@@ -0,0 +1,50 @@
...
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.gcp.data.datastore.core;
+
+import java.util.Iterator;
+
+import com.google.cloud.datastore.Cursor;
+
+public class DatastoreSimpleResultsIterable<T> implements DatastoreResultsIterable<T> {
```
meltsufin, line 23: javadoc missing

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/PartTreeDatastoreQuery.java
```diff
@@ -245,24 +245,25 @@ private String mapToFieldName(PropertyDescriptor propertyDescriptor) {
...
 
 	private Slice executeSliceQuery(Object[] parameters) {
+		DatastoreNextPageAwareResultsIterable<?> results =
+				(DatastoreNextPageAwareResultsIterable<?>)
+						this.datastoreOperations.queryKeysOrEntities(buildSliceQuey(parameters), this.entityType);
+
+		return (Slice) this.processRawObjectForProjection(
+				new SliceImpl(results.toList(),
+						getPageable(parameters, results.getCursor()),
+						results.hasNextPage()));
```
meltsufin, line 255: Does it make sense to make `SliceImpl` also evaluate `hasNext()` lazily?

meltsufin (changes requested):

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/repository/query/PartTreeDatastoreQuery.java
dmitry-s: It is a Spring Data class and expects a boolean parameter.

meltsufin: Can we extend `SliceImpl`?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreOperations.java
```diff
@@ -272,11 +275,31 @@
 	/**
 	 * Finds objects by using a Cloud Datastore query. If the query is a key-query, then keys are
 	 * returned.
+	 * Resulting Slice can be used to get a Pageable for the next page or to determine if next page exists.
 	 * @param query the query to execute.
 	 * @param entityClass the type of object to retrieve.
```
meltsufin, line 280: Can you clarify which type to specify for key?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreOperations.java
```diff
@@ -272,11 +275,31 @@
 	/**
 	 * Finds objects by using a Cloud Datastore query. If the query is a key-query, then keys are
 	 * returned.
+	 * Resulting Slice can be used to get a Pageable for the next page or to determine if next page exists.
```
meltsufin, line 278:
```suggestion
	 * Resulting Slice can be used to get a Pageable for the next page or to determine if next the page exists.
```

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreOperations.java
```diff
@@ -272,11 +275,31 @@
...
+	<T> Slice<?> queryKeysOrEntitiesSlice(StructuredQuery query, Class<T> entityClass, Pageable pageable);
+
+	/**
+	 * Runs a query that checks if there is at least one entry after the cursor.
+	 *
+	 * @param query the query that returned the cursor
+	 * @param cursorAfter the cursor
+	 * @return true if next page exists for the given query and cursor
+	 */
+	boolean nextPageExists(StructuredQuery query, Cursor cursorAfter);
```
meltsufin, line 294: Do we still need this method, since `Slice` has `hasNextPage()`?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
```diff
@@ -261,6 +266,32 @@ public long count(Class<?> entityClass) {
 				: null;
 	}
 
+	@Override
+	public <T> Slice<?> queryKeysOrEntitiesSlice(StructuredQuery query, Class<T> entityClass, Pageable pageable) {
+		DatastoreResultsIterable<?> results = queryKeysOrEntities(applyPageable(query, pageable), entityClass);
+		return new SliceImpl<>(results.toList(),
```
meltsufin, line 272: Can we extend `SliceImpl` such that `hasNextPage()` is lazy?
Feel free to leave for another PR though.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreOperations.java
```diff
@@ -272,11 +275,31 @@
...
+	 * Resulting Slice can be used to get a Pageable for the next page or to determine if next page exists.
 	 * @param query the query to execute.
 	 * @param entityClass the type of object to retrieve.
+	 * @param pageable that indicates page number and page size
 	 * @param <T> the type of object to retrieve.
-	 * @return a list of the objects found. If no keys could be found the list will be
-	 * empty.
+	 * @return a Slice containing found objects
+	 */
+	<T> Slice<?> queryKeysOrEntitiesSlice(StructuredQuery query, Class<T> entityClass, Pageable pageable);
```
meltsufin, line 285: Can you remind me, what is the benefit of combining `queryEntities()` and `queryKeys()` into one method?
Sonar is identifying this as a "code smell".

meltsufin (changes requested):

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
dmitry-s: I don't think it is necessary because if you don't need  `hasNextPage()`, you could just use the `queryKeysOrEntities` method instead.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreOperations.java
dmitry-s: The code would be identical for these cases, so we are avoiding duplication this way.

dmitry-s: Let me see if I can refactor it to avoid duplication.

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
meltsufin: What about the use in repositories?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java
```diff
@@ -267,8 +268,21 @@ public long count(Class<?> entityClass) {
...
-	public <T> Slice<?> queryKeysOrEntitiesSlice(StructuredQuery query, Class<T> entityClass, Pageable pageable) {
-		DatastoreResultsIterable<?> results = queryKeysOrEntities(applyPageable(query, pageable), entityClass);
+	public <T> Slice<Key> queryKeysSlice(KeyQuery query, Class<T> entityClass, Pageable pageable) {
+		return buildSlice(query, pageable, Key.class);
+	}
+
+	@Override
+	public <T> Slice<T> queryEntitiesSlice(StructuredQuery query, Class<T> entityClass, Pageable pageable) {
+		if (query instanceof KeyQuery) {
+			throw new DatastoreDataException("query must be an EntityQuery or a ProjectionEntityQuery");
```
meltsufin, line 278: Why don't you check for those types in the condition then? What if other query types are added in the future?

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreOperations.java
```diff
@@ -269,14 +272,35 @@
...
 	 * Finds objects by using a Cloud Datastore query. If the query is a key-query, then keys are
 	 * returned.
 	 * @param query the query to execute.
 	 * @param entityClass the type of object to retrieve.
 	 * @param <T> the type of object to retrieve.
-	 * @return a list of the objects found. If no keys could be found the list will be
-	 * empty.
+	 * @return an iterable containing found objects and a cursor
 	 */
 	<T> DatastoreResultsIterable<?> queryKeysOrEntities(Query query, Class<T> entityClass);
```
meltsufin, line 305: Can we also deprecate this method and bifurcate it like you did for `Slice`? This can be in a separate PR.

meltsufin (changes requested):

spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreOperations.java
dmitry-s: created an issue: #2559

meltsufin (approved):

meltsufin:
@dmitry-s Did you mean `DatastoreTemplate`? 😄

Issue #2533, opened 2020-09-28
Isolate PubSubTemplateIntegrationTests.testCreatePublishPullNextAndDelete


elefeint:
Set up a randomly named topic/subscription, so two builds can run simultaneously.

meltsufin:
This method already [creates](https://github.com/spring-cloud/spring-cloud-gcp/blob/master/spring-cloud-gcp-autoconfigure/src/test/java/org/springframework/cloud/gcp/autoconfigure/pubsub/it/PubSubTemplateIntegrationTests.java#L91) a random topic and a random subscription.

We actually need this fix for `ReactiveReceiverApplicationTest`.

PR #2546, opened 2020-10-07
Make Pub/Sub integration tests less flaky

meltsufin:


elefeint (approved):
This is definitely an improvement. We should still make topic/subscription names random to prevent flakes from multiple simultaneous builds.

meltsufin:
> This is definitely an improvement. We should still make topic/subscription names random to prevent flakes from multiple simultaneous builds.

The first one already uses random topic/subscription. The second is a sample test, which we can definitely improve by making the topic and subscription configurable. The same goes for a few other samples like the `spring-cloud-gcp-integration-pubsub-json-sample`. Do we have a tracking bug for this? It would be a `good first issue` for anyone. :)

elefeint:
Yes, Travis filed #2533.

Issue #2556, opened 2020-10-14
Make UserAgentHeaderProviderIT run


elefeint:
UserAgentHeaderProviderIT is named to be matched by failsafe, but this repo does not configure failsafe.

The new repo does configure failsafe, but still does not run the test due to the custom file name pattern matcher.

PR #2557, opened 2020-10-14
Make UserAgentHeaderProviderIT run

meltsufin:
Fixes: #2556.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -253,6 +253,19 @@
 					</links>
 				</configuration>
 			</plugin>
+			<plugin>
+				<groupId>org.apache.maven.plugins</groupId>
+				<artifactId>maven-failsafe-plugin</artifactId>
+				<version>${maven-failsafe-plugin.version}</version>
+				<executions>
+					<execution>
+						<goals>
+							<goal>integration-test</goal>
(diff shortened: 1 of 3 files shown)


meltsufin:
@ttomsu @elefeint Friendly ping.

pom.xml
```diff
@@ -253,6 +253,19 @@
...
 			</plugin>
+			<plugin>
+				<groupId>org.apache.maven.plugins</groupId>
+				<artifactId>maven-failsafe-plugin</artifactId>
+				<version>${maven-failsafe-plugin.version}</version>
+				<executions>
+					<execution>
+						<goals>
+							<goal>integration-test</goal>
+							<goal>verify</goal>
```
ttomsu, line 264: This will work when someone does `mvn verify`, but because this old repo does integration tests using `surefire`, the [TravisCI config only does `mvn test`](https://github.com/spring-cloud/spring-cloud-gcp/blob/master/.travis.yml#L31-L33), so this test will continue to not run regularly.

meltsufin: Good catch! I guess I'll modify TavisCI config.

ttomsu (approved):

PR #2587, opened 2020-11-10
Update changelog for 1.2.6.RELEASE

meltsufin:


dzou (approved):

meltsufin:
> Excluding an additional header in Pub/Sub is probably worth mentioning -- #2564.
> Maybe also #2558

@elefeint Both are mentioned, thought not necessarily worded the same way. Do you see them?

elefeint (approved):

Issue #2576, opened 2020-11-05
BigQueryTemplate missing ability to set a custom schema


tnguyenpham-clgx:
BigQueryTemplate provide an abstraction to insert new data into bigquery table with a default autodetect schema option. This is great in most cases, but there are cases we want to provide specific schema to bigquery write as autodect schema sometime do the wrong thing. For example, a column in bigquery table was created to be a String but when a new row is insert into table in csv format that field may be a string of number and auto schema detection think it is a number and create a wrong field type for this col, result in error writing to bigquery table.

Would like to have an ability to set a schema when call BigQueryTemplate. writeDataToTable() or when call BigQueryTemplate. setAutoDetectSchema().

Note that BigQueryTemplate. setAutoDetectSchema() documentation below saying that when AutoDetectSchema set to false, a schema must be defined and yet there is no option to set it. 
"
/**
	 * Sets whether BigQuery should attempt to autodetect the schema of the data when loading
	 * data into an empty table for the first time. If set to false, the schema must be
	 * defined explicitly for the table before load.
	 * @param autoDetectSchema whether data schema should be autodetected from the structure
	 *     of the data. Default is true.
	 */
"

dzou:
@tnguyenpham-clgx - Thanks for the report, will help look at this shortly. My current plan is to add an additional setting to `writeDataToTable` to let you specify the schema manually.

dzou:
Hey I'm in the process of working on this here: https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/108

Right now the PR is in the new repo but I plan on backporting it here.

dzou:
For now if you want to work around this, I would suggest creating the table manually first via an autowired `BigQuery` object and then you can use the template to load as normal with `autoDetectSchema = false`.

Our BigQueryAutoconfiguration allows you to inject an instance of `BigQuery` into your code: https://docs.spring.io/spring-cloud-gcp/docs/1.2.5.RELEASE/reference/html/#bigquery-client-object

Once injected, you can create a table like this: https://cloud.google.com/bigquery/docs/tables#creating_an_empty_table_with_a_schema_definition

dzou:
You should now be able to depend on this via. the snapshot version of spring-cloud-gcp `1.2.7.BUILD-SNAPSHOT`.

PR #2588, opened 2020-11-10
(Cherrypick) Allow for manually specifying BigQuery table schema

dzou:
This adds an interface method to the BigQueryTemplate which allows you to specify an explicit schema if you don't want to rely on schema autodetection on your dataset.

Fixes https://github.com/spring-cloud/spring-cloud-gcp/issues/2576

(cherry picked from commit 283cd16effa144262c57846e55c6a6169ba0dbc1)
Source: https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/108

### DIFF
--- a/spring-cloud-gcp-bigquery/src/test/java/org/springframework/cloud/gcp/bigquery/integration/outbound/BigQueryFileMessageHandlerIntegrationTests.java
+++ b/spring-cloud-gcp-bigquery/src/test/java/org/springframework/cloud/gcp/bigquery/integration/outbound/BigQueryFileMessageHandlerIntegrationTests.java
@@ -22,9 +22,13 @@ import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 
 import com.google.cloud.bigquery.BigQuery;
+import com.google.cloud.bigquery.Field;
+import com.google.cloud.bigquery.Field.Mode;
 import com.google.cloud.bigquery.FormatOptions;
 import com.google.cloud.bigquery.Job;
 import com.google.cloud.bigquery.QueryJobConfiguration;
+import com.google.cloud.bigquery.Schema;
+import com.google.cloud.bigquery.StandardSQLTypeName;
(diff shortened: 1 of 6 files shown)


meltsufin (approved):

spring-cloud-gcp-bigquery/src/test/java/org/springframework/cloud/gcp/bigquery/integration/outbound/BigQueryFileMessageHandlerIntegrationTests.java
```diff
@@ -85,6 +89,46 @@ public void setup() {
...
+				new MessageHeaders(messageHeaders));
+
+		ListenableFuture<Job> jobFuture =
+				(ListenableFuture<Job>) this.messageHandler.handleRequestMessage(message);
+
+		// Assert that a BigQuery polling task is scheduled successfully.
+		await().atMost(Duration.FIVE_SECONDS)
+				.untilAsserted(
+						() -> assertThat(
+								this.taskScheduler.getScheduledThreadPoolExecutor().getQueue()).hasSize(1));
```
ttomsu, line 116: Is it possible that concurrent tests would mean this executor pool has size > 1? Is there something more meaningfully checked in the queue item?

ttomsu (approved):

spring-cloud-gcp-bigquery/src/test/java/org/springframework/cloud/gcp/bigquery/integration/outbound/BigQueryFileMessageHandlerIntegrationTests.java
dzou: Hmm as long as the tests in this file are not concurrent it should be okay. It should not interfere with any other tests run in other files since the pool is not shared.

Yeah the purpose of the check is just to verify a new `Runnable` was created in the task scheduler. Doesn't seem to be anything deeper I can verify about the object though.

PR #2592, opened 2020-11-13
Fix: Spanner Emulator Project ID does not take the default Spring Clo…

### LINKED ISSUE #2582: Spanner Emulator Project ID does not take the default Spring Cloud GCP project ID
**Describe the bug**
https://github.com/spring-cloud/spring-cloud-gcp/blob/master/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/spanner/GcpSpannerEmulatorAutoConfiguration.java#L53

This seems to only use project ID specified under `spring.cloud.gcp.spanner.project-id`, rather than defaulting to `spring.cloud.gcp.project-id`.

**Sample**
https://github.com/saturnism/testcontainers-gcloud-examples/blob/main/springboot/spanner-example/src/test/resources/application-test.properties#L9

kioie:
Fixes #2582


meltsufin (approved):
Thanks!

meltsufin (commented):
A test would be nice too. :)

PR #2594, opened 2020-11-17
Randomize the table names used in BigQuery integration tests

dzou:
Randomizes the table names used in BigQuery integration tests to reduce flakiness regarding the 1,500 updates per day limit: https://cloud.google.com/bigquery/quotas#tablesinsert_requests


elefeint (approved):

meltsufin (approved):
We should probably also port to the new repo.

dzou:
> We should probably also port to the new repo.

Will do.

Issue #2597, opened 2020-11-19
Small Typo on Google Cloud Storage documentation


PierrickPuimeanChieze:
**Describe the bug**
It's just a very small typo on the Google Cloud Storage documentation ([spring-cloud-gcp/docs/src/main/asciidoc/storage.adoc](https://github.com/spring-cloud/spring-cloud-gcp/blob/master/docs/src/main/asciidoc/storage.adoc)). The doc say

`It is also possible to write to a Resource, although a WriteableResource is required.` when it should be
`It is also possible to write to a Resource, although a WritableResource is required.`. The example are good, however.

PR #2598, opened 2020-11-19
Correcting small typo on Google Cloud Storage documentation (#2597)

PierrickPuimeanChieze:
Correcting the name of the class from `WriteableResource` to `WritableResource`

I know that the issue was still not approved, but given the microscopic size of the change, I figured that I could start and submit right away. If I am wrong, sorry about this.


meltsufin (approved):
Thank you!

PR #2604, opened 2020-11-30
Fix auditing when running through DatastoreTemplate.performTransaction

### LINKED ISSUE #2603: Auditing doesn't work on entities in Datastore when running through DatastoreTemplate.performTransaction()
When using `DatastoreTemplate.performTransaction()` with `spring-cloud-gcp-data-datastore` `1.2.6.RELEASE`, auditing doesn't work for entities.

That's because [the new `DatastoreTemplate` instance created in `performTransaction()`](https://github.com/spring-cloud/spring-cloud-gcp/blob/v1.2.6.RELEASE/spring-cloud-gcp-data-datastore/src/main/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplate.java#L435-L440) does not have the `ApplicationEventPublisher` of the original one, which is the mechanism by which the `AuditingHandler` eventually gets called to set the auditing properties.

I've [added a test to `DatastoreTemplateAuditingTests` in my fork](https://github.com/spring-cloud/spring-cloud-gcp/compare/master...fpavageau:bug-auditing-transaction) which demonstrates the problem.

I also have a [simple fix ready](https://github.com/spring-cloud/spring-cloud-gcp/compare/master...fpavageau:bugfix-auditing-transaction), if it's actually a bug and you want a PR.

fpavageau:
The new `DatastoreTemplate` instance should have the same
`ApplicationEventPublisher` as the original one, so that `AuditingHandler`
can be called.

Fixes #2603

### DIFF
--- a/spring-cloud-gcp-data-datastore/src/test/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplateAuditingTests.java
+++ b/spring-cloud-gcp-data-datastore/src/test/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplateAuditingTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017-2018 the original author or authors.
+ * Copyright 2017-2020 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -53,6 +53,7 @@ import static org.mockito.Mockito.when;
  * Tests the auditing features of the template.
  *
  * @author Chengyuan Zhao
+ * @author Frank Pavageau
  */
 @RunWith(SpringRunner.class)
 @ContextConfiguration
@@ -62,6 +63,8 @@ public class DatastoreTemplateAuditingTests {
 
 	@Autowired
 	DatastoreTemplate datastoreTemplate;
+	@Autowired
+	Datastore datastore;
 
 	@Test
 	public void testModifiedNullProperties() {
@@ -80,6 +83,20 @@ public class DatastoreTemplateAuditingTests {
 		testEntity.lastUser = "person";
 
 		this.datastoreTemplate.saveAll(Collections.singletonList(testEntity));
+	}
+
+	@Test
+	public void testInTransaction() {
+		when(datastore.runInTransaction(any()))
+				.thenAnswer(invocation -> {
+					Datastore.TransactionCallable<?> callable = invocation.getArgument(0);
+					return callable.run(datastore);
+				});
+
+		TestEntity testEntity = new TestEntity();
+		testEntity.id = "a";
+
+		this.datastoreTemplate.performTransaction(operations -> operations.save(testEntity));
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


spring-cloud-gcp-data-datastore/src/test/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplateAuditingTests.java
```diff
@@ -82,6 +84,20 @@ public void testModifiedPrevProperties() {
 		this.datastoreTemplate.saveAll(Collections.singletonList(testEntity));
 	}
 
+	@Test
+	public void testInTransaction() {
+		when(datastore.runInTransaction(any()))
```
meltsufin, line 89: I see mock setup all done in `datastoreTemplate()`. Would it make sense to move it there?

meltsufin (changes requested):
Generally, looks good!
Feel free to add yourself as `@author` on the files you touched.
Thanks!!

spring-cloud-gcp-data-datastore/src/test/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplateAuditingTests.java
fpavageau: Personally, I prefer keeping the specific setup where it's actually used, and Mockito could complain about unnecessary setup in the other tests, depending on its configuration. However, I've tested moving the code and it works, so Mockito is not in strict mode and the code can be arranged however you prefer.

meltsufin: I don't feel strongly either way. I'm also not sure it's good that the verification is outside the test method, but that's a problem for another time. :)

meltsufin:
@fpavageau   BTW, no need to squash or force-push in branches/PRs because we always squash commits on merge to master. Do you want to add your name to `DatastoreTemplate.java`? If not, I'm happy to merge.

spring-cloud-gcp-data-datastore/src/test/java/org/springframework/cloud/gcp/data/datastore/core/DatastoreTemplateAuditingTests.java
fpavageau: OK, I think I'll leave it that way then, not specifying every possible behavior for all the tests.

I agree the verification in an `Answer` is brittle, which is when I first ran the test, without the added behavior on the mock, it passed: the `Datastore.put()` method was never called! We could extract an assertion method which would use an `ArgumentCaptor` to verify the call and its parameter, instead. However it's not the only occurence in the project of assertions in `Answer`s, so just fixing this one instance probably doesn't really make sense.

meltsufin: Yeah, no need to address that here.

fpavageau:
I added my name to the test class, I guess that's enough.

meltsufin (approved):

PR #2607, opened 2020-12-04
fix(tests): Vision doesn't detect boston terriers anymore

ttomsu:


dzou (approved):
They're regressing! :(

PR #2624, opened 2021-01-17
fixes #2404 by upgrading error prone to 2.4.0

mac2000:
hopefully will fix the issue with error-prone, got the same issue but with a different class, did try to switch between 8, 11, 15 but the error still persists, after upgrading error-prone at least got further

PR #2625, opened 2021-01-17
fixes #2514 will use given project id if any and fallback to unused

### LINKED ISSUE #2514: Firestore Autoconfiguration emulator not setting the project id
**Describe the bug**
With #2244 , autoconfiguration for firestore emulator has been introduced. The autoconfiguration configures a hard coded project id "unused". The consequences are that anything you do in your code like adding documents to your collection will work and can be retrieved from code but will not be revealed in the Firebase UI at http://localhost:4000/firestore.

**Environment info**
spring-cloud-gcp version 1.2.5.RELEASE
$ firebase --version
8.10.0

**Steps to reproduce**

- Run firebase init on an empty folder and start the emulator with `firebase emulator:start`. This will ask you which project you want the emulator to emulate. Select the same project as you will configure in your spring boot application in the next step. Once you provide all the required settings, it will start a web application which you can then open in a browser. Navigate to the firestore emulator page.

- Create a Spring boot application and add spring-cloud-gcp-starter-data-firestore. Configure your application to use firestore emulator with 
`spring.cloud.gcp.project-id=<google-project-id>`
`spring.cloud.gcp.firestore.emulator.enabled=true`
 `spring.cloud.gcp.firestore.host-port=host:port`

- Do a write operation in your code so you can see it in the firestore emulator UI.

**Expected behavior**
Firebase UI at http://localhost:4000/firestore should show the data.

**Actual behavior**
Firebase UI at http://localhost:4000/firestore shows only an empty root.

mac2000:
fixes #2514 by using given project id, indeed it might be used in many other places, e.g. to reset database we can `curl -x delete /project/{projectId}` so instead of trying to figure out what default are we can use given option and fallback to `unused`

### DIFF
--- a/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/firestore/GcpFirestoreEmulatorAutoConfiguration.java
+++ b/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/firestore/GcpFirestoreEmulatorAutoConfiguration.java
@@ -30,6 +30,7 @@ import com.google.firestore.v1.FirestoreGrpc;
 import io.grpc.ManagedChannel;
 import io.grpc.ManagedChannelBuilder;
 import io.grpc.auth.MoreCallCredentials;
+import org.apache.commons.lang3.StringUtils;
 import reactor.core.publisher.Flux;
 
 import org.springframework.boot.autoconfigure.AutoConfigureBefore;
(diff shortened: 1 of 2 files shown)


mac2000:
The same story as [here](https://github.com/spring-cloud/spring-cloud-gcp/pull/2624) I even added one more dedicated test to the proposed change but coverage is bad :(

ttomsu:
LGTM - @dzou, mind taking a look too?

ttomsu (approved):

dzou (approved):
Looks great, thank you.

PR #2641, opened 2021-02-12
Add automatic retry steps in the case of Maven central failure (like …

ttomsu:
…in https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/298)


elefeint (approved):
Last thing I do before packing up the work setup. 
We got internet yesterday (with some drama, had to unexpectedly* dig the neighbors' house from snow and get another appointment within 24 hours by going "polite but angry velociraptor" on customer service).
* who _expects_ to need to shovel neighbors out before getting internet installed?!

Enjoy the holiday today.

PR #2644, opened 2021-02-17
Upgrade GCP Library Versions

dzou:
Upgrades the GCP Library versions.

Had to manually override the version of spanner to `4.0.0` because it includes a bug fix for user-agent headers. Once the `libraries-bom` catches up to the latest spanner version we can remove the override.

Will cherrypick to new repo and create an issue to track the dep override if satisfactory.

Also `com.google.api-client` no longer needs to be managed manually, so removed.

```
<dependency>	
	<groupId>com.google.api-client</groupId>	
	<artifactId>google-api-client</artifactId>	
	<version>${google-api-client.version}</version>	
</dependency>
```

### DIFF
--- a/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryExecutor.java
+++ b/spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryExecutor.java
@@ -34,7 +34,6 @@ import com.google.cloud.spanner.KeySet;
 import com.google.cloud.spanner.Statement;
 import com.google.cloud.spanner.Struct;
 import com.google.cloud.spanner.ValueBinder;
-import org.apache.commons.lang3.StringUtils;
 
 import org.springframework.cloud.gcp.data.spanner.core.SpannerPageableQueryOptions;
 import org.springframework.cloud.gcp.data.spanner.core.SpannerTemplate;
@@ -51,6 +50,7 @@ import org.springframework.data.domain.Sort;
 import org.springframework.data.repository.query.ParameterAccessor;
 import org.springframework.data.repository.query.parser.Part.IgnoreCaseType;
 import org.springframework.data.repository.query.parser.PartTree;
+import org.springframework.util.StringUtils;
 
 /**
  * Executes Cloud Spanner query statements using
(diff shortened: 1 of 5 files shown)


spring-cloud-gcp-dependencies/pom.xml
```diff
@@ -16,10 +16,11 @@
 	<packaging>pom</packaging>
 
 	<properties>
-		<gcp-libraries-bom.version>13.4.0</gcp-libraries-bom.version>
+		<gcp-libraries-bom.version>16.4.0</gcp-libraries-bom.version>
 		<cloud-sql-socket-factory.version>1.0.16</cloud-sql-socket-factory.version>
```
meltsufin, line 20: `1.2.1` was just released

meltsufin (changes requested):
Do we have any versions to update in root `pom.xml` as well?

spring-cloud-gcp-dependencies/pom.xml
dzou: Done.

dzou:
> Do we have any versions to update in root `pom.xml` as well?

Yeah, I will upgrade Spring versions in a separate PR. I'll keep this one GCP dependencies only so it can easily be cherrypicked to the new repo.

meltsufin (approved):

dzou:
Hm ended up not being able to get version `16.4.0` to work. I decided to upgrade to version `16.2.0` which is the latest version that does not have that breaking change with the headers. Let's see if this works.

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryExecutor.java
```diff
@@ -315,10 +315,10 @@ public static Statement getChildrenRowsQuery(Key parentKey,
 	 */
 	private static String combineWithAnd(String cond1, String cond2) {
 		if (StringUtils.isEmpty(cond1)) {
-			return StringUtils.defaultIfEmpty(cond2, StringUtils.EMPTY);
+			return !StringUtils.isEmpty(cond2) ? cond2 : "";
```
ttomsu, line 318: IIRC `isEmpty` is deprecated in favor of `hasText()`

spring-cloud-gcp-dependencies/pom.xml
```diff
@@ -16,10 +16,9 @@
 	<packaging>pom</packaging>
 
 	<properties>
-		<gcp-libraries-bom.version>13.4.0</gcp-libraries-bom.version>
-		<cloud-sql-socket-factory.version>1.0.16</cloud-sql-socket-factory.version>
-		<guava.version>29.0-jre</guava.version>
-		<google-api-client.version>1.30.7</google-api-client.version>
+		<gcp-libraries-bom.version>16.2.0</gcp-libraries-bom.version>
```
ttomsu, line 19: We probably also have to update this in the new repo

spring-cloud-gcp-data-spanner/src/main/java/org/springframework/cloud/gcp/data/spanner/repository/query/SpannerStatementQueryExecutor.java
dzou: Hmm, I tried, but it seems like `isEmpty` checks if a string is empty/null however `hasText` does that plus check if it's all whitespace. There some tests that break from the change.

I initially changed this to get rid of the `import org.apache.commons.lang3.StringUtils`; upgrading to libraries-bom 16.2.0 seems to have lost this transitive dependency.

spring-cloud-gcp-dependencies/pom.xml
dzou: Will do.

PR #2645, opened 2021-02-17
Upgrade Spring dependency versions

dzou:


meltsufin (approved):
Take a look at the parent pom versions too, like in the case of `spring-cloud-build`.

meltsufin (approved):

PR #2665, opened 2021-05-08
Backport Errorprone Config Changes for JDK 16

meltsufin:
An attempt to fix JDK 16 build due to errorprone failing.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -57,7 +57,7 @@
 		<zipkin-gcp.version>1.0.2</zipkin-gcp.version>
 		<app-engine-maven-plugin.version>2.4.0</app-engine-maven-plugin.version>
 		<kotlin.version>1.3.31</kotlin.version>
-		<errorprone.version>2.4.0</errorprone.version>
+		<errorprone.version>2.6.0</errorprone.version>
 		<java-cfenv.version>2.2.2.RELEASE</java-cfenv.version>
 	</properties>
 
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -392,26 +392,65 @@
...
+					</plugins>
+				</pluginManagement>
+			</build>
+		</profile>
+		<!-- Error prone; JDK 8 requires a different configuration than 9+ -->
 		<profile>
 			<id>errorprone_java_9_and_up</id>
 			<activation>
-				<jdk>[9,)</jdk>
+				<jdk>[9,]</jdk>
```
elefeint, line 425: Should this be 9 through 15, or 9 through latest? As is, both `errorprone_java_9_and_up` and `jdk_16_and_up` will fire on 16.

pom.xml
```diff
@@ -431,35 +470,28 @@
 			<activation>
 				<jdk>1.8</jdk>
 			</activation>
+			<properties>
+				<javac.version>9+181-r4173-1</javac.version>
```
elefeint, line 474: uh this is oddly specific. Do we have to do this?

elefeint (changes requested):
At what point should we be getting errorprone to fix issues instead of working around them? Requiring a specific java build seems a bit much.

pom.xml
meltsufin: I believe this was in the errorprone docs.

pom.xml
meltsufin: Good point. I was just blindly backporting from the new repo.

meltsufin: Actually, I think this is intentional. The two profiles combine for 16.

elefeint (approved):

PR #2672, opened 2021-07-20
Upgrade all instances of commons-io

elefeint:
The dependabot PRs got tangled with a service account key expiration. This will address them all.

### DIFF
--- a/spring-cloud-gcp-samples/pom.xml
+++ b/spring-cloud-gcp-samples/pom.xml
@@ -59,6 +59,15 @@
 		<module>spring-cloud-gcp-metrics-sample</module>
 	</modules>
 
+	<dependencies>
+		<dependency>
+			<groupId>commons-io</groupId>
+			<artifactId>commons-io</artifactId>
+			<version>2.7</version>
+			<scope>test</scope>
+		</dependency>
+	</dependencies>
+
 	<build>
 		<pluginManagement>
 			<plugins>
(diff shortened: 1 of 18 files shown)


spring-cloud-gcp-samples/pom.xml
```diff
@@ -59,6 +59,15 @@
 		<module>spring-cloud-gcp-metrics-sample</module>
 	</modules>
 
+	<dependencies>
+		<dependency>
+			<groupId>commons-io</groupId>
+			<artifactId>commons-io</artifactId>
```
meltsufin, line 65: It's a little bit of an anti-pattern to put dependencies in the parent pom, but since it's a test dep and almost every sample uses it, I guess it's fine. Would it make sense to just add it to `dependenciesManagement` to control its version?
Another thing to consider is that we previously tried to make it possible for users to disconnect the samples from the samples parent. This change will make it a bit more difficult.

meltsufin (approved):

spring-cloud-gcp-samples/pom.xml
elefeint: Hmm, that's true. I'll introduce a property and set it in parent.

meltsufin: That still kind of creates a dependency on the parent, I guess. With the way dependabot works though, it's pretty reliable in finding all of the instances. So, if we rely on it, I don't know if it's worth it to centralize the version.

meltsufin (approved):

